PR #758, opened 2014-02-20
[FIX] issue with Akeneo logo and akeneo base url having the following fo...

sylvainraye:
...rmat http://www.mydomain.com/akeneo/

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,4 @@ web/log
 uploads_test/
 .phpspec/*.tpl
 composer.lock
+.idea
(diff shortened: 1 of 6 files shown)


tdgroot:
@BitOne subdirectory is still not supported, right? 

I managed to get it working with some nginx configuration.

``` nginx
location ~ ^/akeneo {
    index app.php;
    try_files $uri $uri/ /akeneo/app.php?$args;

    location ~ \.php$ {
        echo_exec @phpfpm;
    }
}

location ~ ^/bundles/pimui {
    return 301 $scheme://$host/akeneo$request_uri;
}
```

PR #2023, opened 2015-01-19
[POC][WIP] Attempt to clean mappings declaration and storage configuration

jjanvier:


aRn0D:
@jjanvier I don't like it, singleton introduce tight coupling!

jjanvier:
don't like it neither, this is more to show what's wrong with what we have currently

PR #2064, opened 2015-01-26
Fix #2063: change the visilibity of a property to make it more extendabl...

### LINKED ISSUE #2063: resolvedFilePath should be protected instead of private in FileWriter.php (BaseConnectorBundle)
This prevents anyone to extend the class without also re-defining the `setFilePath()` method in order to correctly manage the `resolvedFilePath` property. Not sure if this should be considered as a bug or a non-feature :smile: But if it does not break anything on your side, then it would be more convenient to change the visibility of this property.

More explanations :

- `setFilePath()` initializes the `filePath` (which is protected) and set `resolvedFilePath` to null
- in our case we have a class `B` extended (among other things) the getFilePath() method to add a `%jobid%` placeholder (because we can't rely on the date to export our medias)
- `getFilePath()` uses lazy loading to retrieve the `resolvedFilePath` : if it's null, then the value is computed with the `strtr` stuff, if it's not null then the value is directly loaded from the object property
- part of our business code is unit tested and I discovered that multiple calls to `setFilePath` on our `B` class won't actually update the `resolvedFilePath` property, and finally `getPath` will always return the same string which is not what one can expect.

PR incoming

alafon:
The PR also adds a piece of phpdoc

| Q                    | A
| -------------------- | ---
| Bug fix?             | kind off
| New feature?         | kind off
| BC breaks?           | should not
| CI currently passes? | hope so
| Tests pass?          | will see
| Scenarios pass?      | can't say
| Checkstyle issues?*  | nop
| PMD issues?**        | 
| Changelog updated?   | nop
| Fixed tickets        | 12063
| DB schema updated?   | not needed
| Migration script?    | not needed
| Doc PR               | nop

### DIFF
--- a/src/Pim/Bundle/BaseConnectorBundle/Writer/File/FileWriter.php
+++ b/src/Pim/Bundle/BaseConnectorBundle/Writer/File/FileWriter.php
@@ -32,9 +32,15 @@ class FileWriter extends AbstractConfigurableStepElement implements
      */
     protected $stepExecution;
 
+    /**
+     * @var resource
+     */
     private $handler;
 
-    private $resolvedFilePath;
+    /**
+     * @var string|null
+     */
+    protected $resolvedFilePath;
 
     /**
      * Set the file path


nidup:
Thank you @alafon !

FYI, about visibility, the policy for the core team is to never use `private` to avoid integration issues (we still have few legacy private, we're replacing them)

alafon:
Perfect, thanks @nidup

PR #2072, opened 2015-01-27
PIM-3644 Allow uploading an archive to import variant group media

filipsalpe:
| Q                    | A
| -------------------- | ---
| Bug fix?             | N
| New feature?         | Y
| BC breaks?           | N
| CI currently passes? | N
| Tests pass?          | Y
| Scenarios pass?      | Y
| Checkstyle issues?*  | N
| PMD issues?**        | N
| Changelog updated?   | N
| Fixed tickets        | PIM-3644
| DB schema updated?   | N
| Migration script?    | ~
| Doc PR               | ~


src/Pim/Bundle/BaseConnectorBundle/Processor/Denormalization/VariantGroupProcessor.php
```diff
@@ -221,26 +230,15 @@ protected function denormalizeValuesFromItemData(array $rawProductValues)
...
     /**
-     * Normalize product values objects to JSON format
+     * Normalize product value objects to JSON format
      *
      * @param ProductValueInterface[] $values
      *
      * @return array
      */
-    protected function normalizeValuesToStructuredData(array $values)
+    protected function normalizeValuesToStructuredData($values)
```
solivier: typehint

spec/Pim/Bundle/BaseConnectorBundle/Processor/Denormalization/VariantGroupProcessorSpec.php
```diff
@@ -165,17 +168,26 @@ function it_updates_a_variant_group_and_its_values(
             ->willReturn(new ConstraintViolationList());
 
         $valueNormalizer
-            ->normalize($value, 'json', ['entity' => 'product'])
+            ->normalize([$value], 'json', ['entity' => 'product'])
             ->shouldBeCalled()
```
solivier: remove the souldBeCalled

filipsalpe: We want to be sure that the normalized values are set on the template, so leaving this.

src/Pim/Bundle/BaseConnectorBundle/Processor/Denormalization/VariantGroupProcessor.php
jjanvier: why removing the array typehint ?? 
according to the phpdoc, ProductTemplateInterface:getValues returns an array of ProductValueInterface

spec/Pim/Bundle/BaseConnectorBundle/Processor/Denormalization/VariantGroupProcessorSpec.php
solivier: you're calling a willReturn after that so you need to remove the shouldBeCalled as you are moking it

filipsalpe: No, why?

solivier: because you are mocking it, so you're assuming the behavior of the method, we do not use shouldBeCalled before a willReturn

src/Pim/Bundle/BaseConnectorBundle/Processor/Denormalization/VariantGroupProcessor.php
filipsalpe: ProductTemplateInterface:getValues returns an ArrayCollection in fact, that's why

spec/Pim/Bundle/BaseConnectorBundle/Processor/Denormalization/VariantGroupProcessorSpec.php
filipsalpe: It's both a promise and a prediction. We say that the method will return this if called and we also say that it should be called. We want both here.

src/Pim/Bundle/BaseConnectorBundle/Processor/Denormalization/VariantGroupProcessor.php
jjanvier: ArratCollection

spec/Pim/Bundle/BaseConnectorBundle/Processor/Denormalization/VariantGroupProcessorSpec.php
BitOne: Nop, we don't ;) We want input, no test result here.

BitOne: Nop, we don't ;) It's an input collaborator. It will return stuff and we don't care that it will be called at all. The purpose of this method is to check is a variant group is updated (according to its name).

PR #2239, opened 2015-02-27
PIM-3307: Fix filter dropdown rendering

grena:
| Q                    | A
| -------------------- | ---
| Bug fix?             | Y
| New feature?         | N
| BC breaks?           | N
| CI currently passes? | Y
| Tests pass?          | Y
| Scenarios pass?      | 
| Checkstyle issues?*  | N
| PMD issues?**        | N
| Changelog updated?   | Y
| Fixed tickets        | PIM-3307
| DB schema updated?   | N
| Migration script?    | -
| Doc PR               | -

### DIFF
--- a/src/Pim/Bundle/UIBundle/Resources/public/css/less/oro.less
+++ b/src/Pim/Bundle/UIBundle/Resources/public/css/less/oro.less
@@ -1893,7 +1893,6 @@ textarea {
   border-right: 9px solid transparent;
   content: "";
   display: inline-block;
-  left: 11px;
   position: absolute;
   top: -9px;
 }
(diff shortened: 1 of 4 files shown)


src/Pim/Bundle/UIBundle/Resources/public/css/less/oro.less
```diff
@@ -1907,6 +1907,24 @@ textarea {
   position: absolute;
   top: -8px;
 }
+.filter-criteria {
+  &.right-dropdown-menu:before {
+    left: 11px !important;
```
filipsalpe: Is it really that important?

grena: I'll check

grena: Yup, necessary for the `left` attribute because of its parent, but useless for the `right` attribute, I fix

filipsalpe: not possible with a more specific selector?

PR #2317, opened 2015-03-12
PIM-3891: rework ProductBuilder and remove deprecated methods

nidup:
- [x] changelog with BC to complete
 - [x] sed command to write for repos

| Q                    | A
| -------------------- | ---
| Bug fix?             | N
| New feature?         | N
| BC breaks?           | Y
| CI currently passes? | Y
| Tests pass?          | Y
| Scenarios pass?      | Y
| Checkstyle issues?*  | N
| PMD issues?**        | N
| Changelog updated?   | Y
| Fixed tickets        | PIM-3891
| DB schema updated?   |
| Migration script?    |
| Doc PR               |


src/Pim/Bundle/CatalogBundle/Event/ProductEvents.php
```diff
@@ -12,26 +12,6 @@
...
 {
     /**
-     * This event is thrown each time a product is created in the system.
-     *
-     * The event listener receives an
-     * Pim\Bundle\CatalogBundle\Event\ProductEvent instance.
-     *
-     * @staticvar string
-     */
-    const CREATE = 'pim_catalog.create_product';
```
jjanvier: Isn't it weird to remove it ? Can't this be useful for integrators ?

src/Pim/Bundle/CatalogBundle/Event/ProductValueEvent.php
```diff
@@ -1,55 +0,0 @@
...
-use Symfony\Component\EventDispatcher\Event;
-
-/**
- * Filter event allows to know the create product value
- *
- * @author    Nicolas Dupont <nicolas@akeneo.com>
- * @copyright 2013 Akeneo SAS (http://www.akeneo.com)
- * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
- */
-class ProductValueEvent extends Event
```
jjanvier: same here

src/Pim/Bundle/CatalogBundle/Event/ProductEvents.php
jjanvier: anyway, if you really drop them, it's a (big) BC break and should be in the changelog

nidup: Still hesitating on this one cause for now only dispatched when using the manager method and sometimes we create in a different way, i'm referencing different uses, these dispatch are legacy coming from flexible entity we used them to add logic depending on entity, we now embed the logic in the builder

nidup: still completing the BC Break list (cf my TODO list in PR description)

PR #2327, opened 2015-03-13
PIM-3825: Use the new --no-log batch option

grena:
With the `0.4` version of the BatchBundle, we should use  the `--no-log` option to hide all logs during installation.


nidup:
GTM

PR #2331, opened 2015-03-15
[POC] PIM-3886: POC which convert OroDatagrid filters to PQB filters

nidup:
@solivier @grena there is a POC to convert OroDatagrid filters to PQB filters.

Now, PQB exposes filters already applied (which is something we would move one to enhance the PQB, we could add filters in an internal array and apply them only when we call execute, it allows us to get/add/remove filters before to execute the query).

Here, we refactor a bit the MassActionDispatcher without breaking the current behavior, the extraction of filters is only doable for product datasource (as others grids are based on QB criterias and not PQB).

There is no object hydration, we prepare the grid by applying extensions, then we get the prepared PQB filters without hydrating the results.

This solution can be enough for now.


src/Pim/Bundle/CatalogBundle/Query/ProductQueryBuilder.php
```diff
@@ -42,6 +42,9 @@ class ProductQueryBuilder implements ProductQueryBuilderInterface
     /** CursorFactoryInterface */
     protected $cursorFactory;
 
+    /** @var array */
+    protected $appliedFilters;
```
jjanvier: really bad IMO
we don't apply anything as this point
maybe what you want is a real query object that can contain your filters

nidup: yep, agree, this naming sucks :)

PR #2386, opened 2015-03-24
PIM-3894: Add identifier setter

willy-ahva:
| Q                    | A
| -------------------- | ---
| Bug fix?             | n
| New feature?         | n
| BC breaks?           | n
| CI currently passes? | y
| Tests pass?          | y
| Scenarios pass?      | y
| Checkstyle issues?*  | n
| PMD issues?**        | n
| Changelog updated?   | n
| Fixed tickets        | PIM-3894
| DB schema updated?   | n
| Migration script?    | n
| Doc PR               |

### DIFF
--- a/src/Pim/Bundle/CatalogBundle/Resources/config/updaters.yml
+++ b/src/Pim/Bundle/CatalogBundle/Resources/config/updaters.yml
@@ -70,7 +70,7 @@ services:
         class: %pim_catalog.updater.setter.text_value.class%
         parent: pim_catalog.updater.setter.abstract
         arguments:
-            - ['pim_catalog_text', 'pim_catalog_textarea']
+            - ['pim_catalog_text', 'pim_catalog_textarea', 'pim_catalog_identifier']
         tags:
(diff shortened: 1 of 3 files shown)


src/Pim/Bundle/CatalogBundle/Resources/config/updaters.yml
```diff
@@ -70,7 +70,7 @@ services:
         class: %pim_catalog.updater.setter.text_value.class%
         parent: pim_catalog.updater.setter.abstract
         arguments:
-            - ['pim_catalog_text', 'pim_catalog_textarea']
+            - ['pim_catalog_text', 'pim_catalog_textarea', 'pim_catalog_identifier']
```
solivier: Are you sure an identifier is a text value only ? It can't be an Id ?

willy-ahva: I'll verifiy but I'm pretty sure it's always a string even if it's an id

PR #2387, opened 2015-03-24
PIM-3910: Add association field adder

willy-ahva:
| Q                    | A
| -------------------- | ---
| Bug fix?             | n
| New feature?         | n
| BC breaks?           | n
| CI currently passes? | y
| Tests pass?          | y
| Scenarios pass?      | y
| Checkstyle issues?*  | n
| PMD issues?**        | n
| Changelog updated?   | n
| Fixed tickets        | PIM-3910
| DB schema updated?   | n
| Migration script?    | n
| Doc PR               |

### DIFF
--- a/src/Pim/Bundle/CatalogBundle/Updater/Adder/AssociationFieldAdder.php
+++ b/src/Pim/Bundle/CatalogBundle/Updater/Adder/AssociationFieldAdder.php
@@ -1,0 +1,202 @@
+<?php
+
+namespace Pim\Bundle\CatalogBundle\Updater\Adder;
+
+use Akeneo\Bundle\StorageUtilsBundle\Repository\IdentifiableObjectRepositoryInterface;
+use Pim\Bundle\CatalogBundle\Builder\ProductBuilderInterface;
+use Pim\Bundle\CatalogBundle\Exception\InvalidArgumentException;
+use Pim\Bundle\CatalogBundle\Model\AssociationInterface;
+use Pim\Bundle\CatalogBundle\Model\ProductInterface;
+
+/**
+ * Association field adder
+ *
+ * @author    Willy Mesnage <willy.mesnage@akeneo.com>
+ * @copyright 2015 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+class AssociationFieldAdder extends AbstractFieldAdder
+{
+    /** @var IdentifiableObjectRepositoryInterface */
+    protected $productRepository;
+
+    /** @var IdentifiableObjectRepositoryInterface */
+    protected $groupRepository;
+
+    /** @var ProductBuilderInterface */
+    protected $productBuilder;
+
+    /**
+     * @param IdentifiableObjectRepositoryInterface $productRepository
+     * @param IdentifiableObjectRepositoryInterface $groupRepository
+     * @param ProductBuilderInterface               $productBuilder
+     * @param array                                 $supportedFields
+     */
+    public function __construct(
+        IdentifiableObjectRepositoryInterface $productRepository,
+        IdentifiableObjectRepositoryInterface $groupRepository,
+        ProductBuilderInterface $productBuilder,
+        array $supportedFields
+    ) {
+        $this->productRepository = $productRepository;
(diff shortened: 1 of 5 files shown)


solivier:
think you can squash some commits

src/Pim/Bundle/CatalogBundle/Updater/Adder/AssociationFieldAdder.php
```diff
@@ -0,0 +1,202 @@
...
+        $this->productBuilder->addMissingAssociations($product);
+    }
+
+    /**
+     * Add products and groups to associations
+     *
+     * @param ProductInterface $product
+     * @param mixed            $data
+     */
+    protected function addAssociations(ProductInterface $product, $data)
```
jjanvier: bad naming

src/Pim/Bundle/CatalogBundle/Updater/Adder/AssociationFieldAdder.php
```diff
@@ -0,0 +1,202 @@
...
+        $this->groupRepository   = $groupRepository;
+        $this->productBuilder    = $productBuilder;
+        $this->supportedFields   = $supportedFields;
+    }
+
+    /**
+     * {@inheritdoc}
+     *
+     * Expected data input format :
+     *     "associations": {
```
jjanvier: is it normal we have this ?
for prices, we don't have a "prices" key

src/Pim/Bundle/CatalogBundle/Updater/Adder/AssociationFieldAdder.php
```diff
@@ -0,0 +1,202 @@
...
+     *          "XSELL": {
+     *              "groups": ["group1", "group2"],
```
jjanvier: are groups and products keys mandatory ? is yes, it could be good to have examples with empty sets of groups and products

src/Pim/Bundle/CatalogBundle/Updater/Adder/AssociationFieldAdder.php
willy-ahva: It's the same that we have in AssociationFieldSetter. I kept it to be consistent but I can change if it's better.

src/Pim/Bundle/CatalogBundle/Updater/Adder/AssociationFieldAdder.php
willy-ahva: we are in AssociationFieldAdder, so ...

src/Pim/Bundle/CatalogBundle/Updater/Adder/AssociationFieldAdder.php
nidup: you can change both, protected methods in new classes, so no pb

jjanvier: you don't  add associations here
you add products and groups to associations

naming in AssociationFieldSetter is bad so

src/Pim/Bundle/CatalogBundle/Updater/Adder/AssociationFieldAdder.php
jjanvier: and ?

nidup: @jjanvier you're right, we don't do that for others, @willy-ahva you can fix in adder and setter, expected format should be only,

     *          "XSELL": {
     *              "groups": ["group1", "group2"],
     *              "products": ["AKN_TS1", "AKN_TSH2"]
     *          },
     *          "UPSELL": {
     *              "groups": ["group3", "group4"],
     *              "products": ["AKN_TS3", "AKN_TSH4"]
     *          },

filipsalpe: the field should be 'associations', this level in data is redundant

src/Pim/Bundle/CatalogBundle/Updater/Adder/AssociationFieldAdder.php
willy-ahva: Yes it's mandatory

PR #2391, opened 2015-03-25
use the new container infrastructure on Travis

jmleroux:


jmleroux:
related to #2390

.travis.yml
```diff
@@ -5,6 +5,8 @@ php:
     - "5.6"
     - "hhvm"
 
+sudo: false
```
BitOne: A comment would be nice ;)
sudo doesn't have a natural impact to using container or not usually ;)

PR #2440, opened 2015-03-31
PIM-4017: Fix bug with variant group when saving an empty media attribute

willy-ahva:
| Q                    | A
| -------------------- | ---
| Bug fix?             | YES
| New feature?         | n
| BC breaks?           | n
| CI currently passes? | 
| Tests pass?          | y
| Scenarios pass?      | 
| Checkstyle issues?*  | n
| PMD issues?**        | n
| Changelog updated?   | y
| Fixed tickets        | PIM-4017
| DB schema updated?   | n
| Migration script?    | n
| Doc PR               |

### DIFF
--- a/src/Pim/Bundle/CatalogBundle/Manager/MediaManager.php
+++ b/src/Pim/Bundle/CatalogBundle/Manager/MediaManager.php
@@ -218,8 +218,12 @@ class MediaManager
      */
     public function createFromFilePath($filePath)
     {
-        $media = $this->factory->createMedia();
-        $media->setFilename(pathinfo($filePath, PATHINFO_BASENAME));
+        $media    = $this->factory->createMedia();
+        $fileName = pathinfo($filePath, PATHINFO_BASENAME);
+
+        if ('' !== $fileName) {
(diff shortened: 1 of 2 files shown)


src/Pim/Bundle/CatalogBundle/Manager/MediaManager.php
```diff
@@ -218,8 +218,12 @@ public function createFromFilename($filename, $isUploaded = true)
      */
     public function createFromFilePath($filePath)
     {
-        $media = $this->factory->createMedia();
-        $media->setFilename(pathinfo($filePath, PATHINFO_BASENAME));
+        $media    = $this->factory->createMedia();
+        $fileName = pathinfo($filePath, PATHINFO_BASENAME);
+
+        if ('' !== $fileName) {
```
jjanvier: in what case this can happen ?
is it really the good fix ? shouldn't we prevent from being here instead ?

willy-ahva: As I see with Emilie, save an empty media attribute is a use case. And have an empty file name happens in this case when you create the ProductTemplate from valuesData denormalization.

PR #2444, opened 2015-04-01
ignore local files

jmleroux:
Another PR, related to #2403.

With this one, i can have a custom `web/app.local.php`. In my case, it's for adding a xhprof header file.
But you can also cahnge configuration for testing another bootstrap file, add the ApcClassLoader, etc.

Also, i can have a custom `app/AppKernel.local.php` and use it in the `web/app.local.php`. Can be also usefull in a MongoDB installation.

The last file that will bother git remote updates is composer.json if it is modified....

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@ web/css/*
 web/js/*
 web/app_local.php
 app/AppKernel_local.php
+app/console_local
 app/archive/*
 app/cache/*
 app/config/*_local.yml
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,8 @@ web/bundles/*
 web/media/*
 web/css/*
 web/js/*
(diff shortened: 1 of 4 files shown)


jmleroux:
And to enforce consistency, we could replace `app/config/pim_parameters.yml`with `app/config/pim_parameters_local.yml`

jmleroux:
Documentation will come if PR accepted

.gitignore
```diff
@@ -2,6 +2,9 @@ web/bundles/*
 web/media/*
 web/css/*
 web/js/*
+web/app_local.php
+app/AppKernel_local.php
+app/console_local
```
jmleroux: we must also give the ability to tweak the console, mainly to be able to require `AppKernel_local.php`

PR #2479, opened 2015-04-08
Use default phpspec formatter

jjanvier:
We have too many specs now. Displaying all of them is useless...


grena:
Could we have a before/after screenshot ? :dancer:

filipsalpe:
GTM

PR #2487, opened 2015-04-09
Fix glitches

solivier:


src/Pim/Bundle/EnrichBundle/MassEditAction/Operation/SetAttributeRequirements.php
```diff
@@ -147,4 +147,12 @@ public function getAlias()
     {
         return 'set-attribute-requirements';
     }
+
+    /**
+     * @return string
+     */
+    public function getItemsName()
+    {
+        return 'attribute';
```
grena: I think it will be `family`

PR #2514, opened 2015-04-14
Merge branch '1.3.8' into master

willy-ahva:
| Q                    | A
| -------------------- | ---
| Bug fix?             | y
| New feature?         | n
| BC breaks?           | n
| CI currently passes? | yes
| Tests pass?          | yes
| Scenarios pass?      | yes
| Checkstyle issues?*  | n
| PMD issues?**        | n
| Changelog updated?   | yes
| Fixed tickets        | 
| DB schema updated?   | 
| Migration script?    |
| Doc PR               |

### DIFF
--- a/features/Context/CommandContext.php
+++ b/features/Context/CommandContext.php
@@ -24,6 +24,7 @@ class CommandContext extends RawMinkContext
      */
     public function iLaunchedTheCompletenessCalculator()
     {
+        $this->getFixturesContext()->clearUOW();
         $this
             ->getContainer()
             ->get('pim_catalog.manager.completeness')
@@ -159,5 +160,13 @@ class CommandContext extends RawMinkContext
     protected function getContainer()
     {
         return $this->getMainContext()->getContainer();
+    }
+
+    /**
+     * @return FixturesContext
+     */
+    private function getFixturesContext()
+    {
+        return $this->getMainContext()->getSubcontext('fixtures');
     }
 }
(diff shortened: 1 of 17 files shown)


features/Context/CommandContext.php
```diff
@@ -24,6 +24,7 @@ class CommandContext extends RawMinkContext
      */
     public function iLaunchedTheCompletenessCalculator()
     {
+        $this->getFixturesContext()->clearUOW();
```
jjanvier: why ?

src/Pim/Bundle/CatalogBundle/Validator/Constraints/Range.php
```diff
@@ -16,4 +16,19 @@ class Range extends BaseRange
     public $minDateMessage     = 'This date should be {{ limit }} or after.';
     public $maxDateMessage     = 'This date should be {{ limit }} or before.';
     public $invalidDateMessage = 'This value is not a valid date.';
+
+    /**
+     * {@inheritdoc}
+     */
+    public function __construct($options = null)
```
jjanvier: don't like to initialize arrays to null
[] is better IMO

features/Context/CommandContext.php
willy-ahva: I don't exactly know how it works. @skeleton helped me for this one. It seems when the UOW is not cleared, completeness is recalculate for the previous data and not the new one.

src/Pim/Bundle/CatalogBundle/Validator/Constraints/Range.php
willy-ahva: It's to be consistent with parents. I initialized it to null because parent __construct is initialized to null.

PR #2631, opened 2015-05-08
PIM-3232 Wrap inline scripts in require modules

Demango:


src/Pim/Bundle/CommentBundle/Resources/config/requirejs.yml
```diff
@@ -0,0 +1,5 @@
+config:
+    paths:
+        pim/comment/comment-create-action: pimcomment/js/comment-create-action
+        pim/comment/comment-listing:       pimcomment/js/comment-listing
+        pim/comment/comment-reply-action:   pimcomment/js/comment-reply-action
```
filipsalpe: alignment?

PR #2678, opened 2015-05-19
[TR] PIM-4187 Add javascript execution from file

Demango:


features/Context/FeatureContext.php
```diff
@@ -372,6 +373,35 @@ public function executeScript($script)
...
+    public function executeJs($name)
+    {
+        $args = array_slice(func_get_args(), 1);
+        $filepath = sprintf('%s/scripts/%s.js', __DIR__, $name);
+        $script = file_get_contents($filepath);
+        if (!$script) {
+            throw new PendingException(
+                sprintf(
+                    'The file could not be found, you can create it in path "%s"',
+                    __DIR__ . '/scripts/' . $filepath
```
filipsalpe: looks like filepath is already the absolute path

features/Context/FeatureContext.php
```diff
@@ -372,6 +373,35 @@ public function executeScript($script)
...
+                )
+            );
+        }
+        $this->getSession()
+            ->getDriver()
+            ->getWebDriverSession()
+            ->execute(
+                array(
```
filipsalpe: [ ] is nicer than array()

PR #2681, opened 2015-05-19
PIM-4187 Add display attribute script

Demango:


features/Context/scripts/mass-edit/display-attributes.js
```diff
@@ -0,0 +1,14 @@
+function(attributes) {
+    $('button:contains("Select attributes")').click();
+    for (var i = attributes.length - 1; i >= 0; i--) {
```
filipsalpe: why reverse?

PR #2691, opened 2015-05-21
enrich italian messages updated

danielsan80:


danielsan80:
Sorry. It doesn't rock

PR #2699, opened 2015-05-22
PIM-4199: Rework the product association reader / processor / write

skeleton:


src/Pim/Bundle/BaseConnectorBundle/Reader/File/CsvAssociationReader.php
```diff
@@ -11,6 +11,8 @@
  * @author    Julien Sanchez <julien@akeneo.com>
  * @copyright 2015 Akeneo SAS (http://www.akeneo.com)
  * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ *
+ * @deprecated will be removed in 1.5
  */
 class CsvAssociationReader extends CsvReader
```
nidup: you can drop the class we added in after the 1.3 releae

src/Pim/Bundle/BaseConnectorBundle/Writer/Doctrine/ProductAssociationWriter.php
```diff
@@ -8,6 +8,8 @@
  * @author    Filips Alpe <filips@akeneo.com>
  * @copyright 2014 Akeneo SAS (http://www.akeneo.com)
  * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ *
+ * @deprecated will be removed in 1.5. Please use \Pim\Component\Connector\Writer\Doctrine\ProductAssociationWriter
  */
 class ProductAssociationWriter extends Writer
```
nidup: ok for this "old one" (introduced before 1.4, we keep it as deprecated

nidup:
okay you can add specs and GTM !

src/Pim/Bundle/BaseConnectorBundle/Reader/File/CsvAssociationReader.php
```diff
@@ -1,60 +0,0 @@
...
-use Pim\Component\Connector\ArrayConverter\Flat\ProductAttributeFieldExtractor;
-
-/**
- * Association csv reader
- *
- * @author    Julien Sanchez <julien@akeneo.com>
- * @copyright 2015 Akeneo SAS (http://www.akeneo.com)
- * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
- */
-class CsvAssociationReader extends CsvReader
```
solivier: we do not keep this file in order to set it as deprecated ?

nidup: @solivier because we introduced it during the 1.4 (during the POC) so no need of deprecation

PR #2721, opened 2015-05-26
[TR] Travis settings : execute CS fixer only for PHP 5.6

jmleroux:
Executing CS fixer only on PHP 5.6 speed up Travis build by a 50%+ ratio


jjanvier:
GTM

PR #2732, opened 2015-05-28
PIM-4167: Fallback on a default image when error occures

grena:


src/Pim/Bundle/EnrichBundle/Controller/FileDisplayController.php
```diff
@@ -16,6 +16,9 @@
  */
 class FileDisplayController extends Controller
 {
+    /** @staticvar string */
+    const FALLBACK_IMAGE_PATH = 'bundles/pimenrich/img/img_generic.png';
```
jjanvier: not good for me, you have to take the one that is in Resource/public/images

and so you have to put the default image in the Enrich bundle

src/Pim/Bundle/EnrichBundle/Controller/FileDisplayController.php
```diff
@@ -77,4 +86,23 @@ protected function getFilesystem()
...
+
+    /**
+     * @return Response
+     */
+    protected function getFallbackImageResponse()
+    {
+        $content = file_get_contents(
+            sprintf(
+                '%s/%s',
+                $this->container->getParameter('kernel.root_dir').'/../web',
```
jjanvier: use __DIR__ and find your image in the bundle from here

src/Pim/Bundle/EnrichBundle/Controller/FileDisplayController.php
```diff
@@ -77,4 +86,23 @@ protected function getFilesystem()
...
         return $this->mountManager->getFilesystem($this->filesystemName);
     }
+
+    /**
+     * @return Response
+     */
+    protected function getFallbackImageResponse()
+    {
+        $content = file_get_contents(
+            sprintf(
```
jjanvier: don't mix, either sprint, either dots
and use realpath please

PR #2758, opened 2015-06-01
PIM-4227: fix the deletion limit for categories

solivier:
behats are runnings


src/Pim/Bundle/EnrichBundle/Resources/config/pim.yml
```diff
@@ -52,3 +52,5 @@ doctrine_migrations:
     dir_name: %doctrine_migrations_dir%
     namespace: Pim\Upgrade\Schema
     name: PIM Migrations
+parameters:
```
skeleton: add a blank line before "parameters" please

PR #2773, opened 2015-06-02
[TR] PIM-4295: Rename former category things to product_category

grena:


features/Context/catalog/apparel/jobs.yml
```diff
@@ -47,7 +47,7 @@ jobs:
     category_import:
         connector: Akeneo CSV Connector
         alias:     csv_category_import
-        label:     Category import
+        label:     Product category import
```
jjanvier: revert this one

features/Context/catalog/apparel/jobs.yml
```diff
@@ -58,7 +58,7 @@ jobs:
     category_export:
         connector: Akeneo CSV Connector
         alias:     csv_category_export
-        label:     Category export
+        label:     Product category export
```
jjanvier: this one too

features/Context/catalog/apparel/jobs.yml
jjanvier: we'll do it if we need to build an "asset category import/export"

jjanvier:
GTM

just ensure we don't break all the Behat about the categories and product categories

PR #2830, opened 2015-06-10
[TR] interface_check.sh now only checks for use statements of entities

grena:


app/Resources/jenkins/interface_check.sh
```diff
@@ -7,7 +7,7 @@ OUTPUT=''
 # Channel Completeness Currency Family Group Locale Metric Product ProductMedia ProductPrice ProductValue
 
 for pattern in Attribute Category Completeness Product ProductValue; do
-    OUTPUT=`find $1 -type f -name "*.php" | grep -v "/Tests/" | xargs grep -P "[^a-zA-Z']$pattern([ ',;)]|\$)" | grep -v "*" | grep -v "namespace" | grep -v "class $pattern" | tr "\n" "#"`$OUTPUT
+    OUTPUT=`find $1 -type f -name "*.php" | grep -v "/Tests/" | xargs grep -P "^use .*[^a-zA-Z']$pattern([ ',;)]|\$)" | grep -v "*" | grep -v "namespace" | grep -v "class $pattern" | tr "\n" "#"`$OUTPUT
 done
```
jmleroux: @grena : if we check only `use` statements, then the filter can be simplified : 
```OUTPUT=`find $1 -type f -name "*.php" | grep -v "/Tests/" | xargs grep -P "^use .*[^a-zA-Z']$pattern([ ',;)]|\$)" | tr "\n" "#"`$OUTPUT```

grena: In fact it finds a lot a false positive if I remove it ^^

solivier:
GTM

jmleroux:
GTM

PR #2834, opened 2015-06-10
Some fixes & TODOs & questions

grena:


features/Context/Page/Base/Base.php
```diff
@@ -109,8 +111,7 @@ public function getTitle()
         $name      = $elt->find('css', '.product-name');
 
         if (!$subtitle || !$separator || !$name) {
-            $title = $this->getElement('Product title')
-                ->find('css', '.product-label')->getText();
+            $title = $this->getElement('Product title')->find('css', '.product-label')->getText();
```
jmleroux: we should even check that find doesn't return NULL before `getText()`

jmleroux:
Review merged.
Thx.

PR #2886, opened 2015-06-23
[POC] Attempt to use classic DI for the batch jobs

nidup:


nidup:
I close this POC, will not be done for 1.4

PR #2894, opened 2015-06-26
PIM-4236: re-work attribute option csv import

solivier:


src/Pim/Bundle/InstallerBundle/Resources/config/fixtures_jobs.yml
```diff
@@ -14,16 +14,6 @@ jobs:
             circularRefsChecked: true
             uploadAllowed: true
 
-    fixtures_category_yml:
```
skeleton: why remove this ?

solivier: because it's not used anymore

nidup: yep we get ride of all unused format ;)

nidup:
think about do a PR for EE and update the changelog (to say we removed these yml stuff there is already a line for that) then GTM

PR #2920, opened 2015-07-01
Improved the check style check

solivier:
Check style is now checked by php-cs-fixer instead of phpcs to be compliant with travis.


build.xml
```diff
@@ -163,6 +163,26 @@
         </exec>
     </target>
 
+    <target name="php-cs-fixer" description="Find coding standard violations using PHP_CodeSniffer and print human readable output. Intended for usage on the command line before committing.">
+        <exec executable="bash">
+            <arg value="-c"/>
+            <arg value="curl http://get.sensiolabs.org/php-cs-fixer.phar -o php-cs-fixer"/>
+        </exec>
+
+        <exec executable="bash">
```
BitOne: Not really useful, just add "php " before calling php-cs-fixer

build.xml
```diff
@@ -163,6 +163,26 @@
...
+            <arg value="-c"/>
+            <arg value="chmod a+x php-cs-fixer"/>
+        </exec>
+
+        <exec executable="${basedir}/php-cs-fixer">
+            <arg value="fix" />
+            <arg path="${basedir}/src" />
+            <arg value="-v" />
+            <arg value="--config-file=.php_cs.local.php" />
```
BitOne: Would have use the ${basedir} here. You're never sure from where the ant execution is done.

build.xml
```diff
@@ -163,6 +163,26 @@
         </exec>
     </target>
 
+    <target name="php-cs-fixer" description="Find coding standard violations using PHP_CodeSniffer and print human readable output. Intended for usage on the command line before committing.">
+        <exec executable="bash">
```
BitOne: The original intent for the bash call was to use the pipe. You're not using it here, so you can call directly the curl executable with the right args.

build.xml
```diff
@@ -163,6 +163,26 @@
...
+            <arg value="-c"/>
+            <arg value="curl http://get.sensiolabs.org/php-cs-fixer.phar -o php-cs-fixer"/>
```
BitOne: Add the ${basedir} as well here before the php-cs-fixer file name to be sure that's put at the right place.

PR #2964, opened 2015-07-10
[TR] Fix feature/pam behat scenarios

grena:


features/Context/FixturesContext.php
```diff
@@ -126,6 +126,17 @@ public function clearPimFilesystem()
     }
 
     /**
+     * @BeforeScenario
+     */
+    public function clearStorageFilesystem()
+    {
+        $storageDir = $this->getContainer()->getParameter('storage_dir');
```
jjanvier: ensure to have an absolute file path here to avoid armaggedon

grena: by calling `reapath`? storage_dir is a concatenation of kernel root and storage_file

nidup: agree, then GTM

src/Pim/Bundle/CatalogBundle/Doctrine/Common/Remover/ChannelRemover.php
```diff
@@ -0,0 +1,70 @@
...
+                    ClassUtils::getClass($channel)
+                )
+            );
+        }
+
+        $options = $this->optionsResolver->resolveRemoveOptions($options);
+        $this->eventDispatcher->dispatch(ChannelEvents::PRE_REMOVE, new GenericEvent($channel));
+
+        $this->objectManager->remove($channel);
+        if (true === $options['flush']) {
```
jjanvier: maybe you should also implement flush_only_object like in the base remover

jjanvier: because we have this option but it's never used here

src/Pim/Bundle/CatalogBundle/Doctrine/Common/Remover/ChannelRemover.php
```diff
@@ -0,0 +1,70 @@
...
+            $this->objectManager->flush();
```
skeleton: Maybe we should add a POST_REMOVE event ?

grena: Yep I could, but it's not used for the moment.
Could be interesting for integrators, but then we should do it for all our removers, but it's not the case for now :(

skeleton: That's true, there is not enough events on removers in the PIM, but IMHO, this is not a reason to do not add it on a very important action like remove a channel.

PR #3043, opened 2015-08-03
PIM-4261: added missing case for file comparator

solivier:
| Q                  | A
| ------------------ | ---
| Migration script?  | n
| Behats             | y
| Specs & Static     | y
| Changelog updated  | n


jjanvier:
GTM just need to add behat for EE

PR #3077, opened 2015-08-10
[TR] PIM-4705: Update controller and grid to be able to configure job to launch with quick export

willy-ahva:
| Q                  | A
| ------------------ | ---
| Migration script?  | n
| Specs & Static     | ok
| Changelog updated  | n

### DIFF
--- a/src/Pim/Bundle/DataGridBundle/Controller/ProductExportController.php
+++ b/src/Pim/Bundle/DataGridBundle/Controller/ProductExportController.php
@@ -11,6 +11,7 @@ use Pim\Bundle\ImportExportBundle\Entity\Repository\JobInstanceRepository;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\Security\Core\Authentication\Token\Storage\TokenStorageInterface;
+use Symfony\Component\Security\Core\User\UserInterface;
 
 /**
  * Products quick export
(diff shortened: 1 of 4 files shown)


src/Pim/Bundle/EnrichBundle/Connector/Reader/MassEdit/FilteredProductReader.php
```diff
@@ -68,12 +68,12 @@ public function read()
         $configuration = $this->getJobConfiguration();
 
         if (null === $configuration) {
-            return null;
+            return;
```
phaseinducer: `null` should be kept here imho

phaseinducer:
little comment otherwise GTM

src/Pim/Bundle/EnrichBundle/Connector/Reader/MassEdit/FilteredProductReader.php
willy-ahva: `return;` returns null. So we have an explicit return and we are consistent with the rest of the PIM. That's why I removed it.

src/Pim/Bundle/DataGridBundle/Controller/ProductExportController.php
```diff
@@ -95,9 +97,9 @@ public function indexAction()
     /**
      * Get a user from the Security Context
      *
-     * @return \Symfony\Component\Security\Core\User\UserInterface|null
+     * @return UserInterface|null
      *
-     * @see Symfony\Component\Security\Core\Authentication\Token\TokenInterface::getUser()
+     * @see \Symfony\Component\Security\Core\Authentication\Token\TokenInterface::getUser()
```
jmleroux: No FQCN

jmleroux:
GTM

src/Pim/Bundle/DataGridBundle/Controller/ProductExportController.php
willy-ahva: it's just for @see not for PHPDoc, don't know what is rule for that.

PR #3160, opened 2015-08-20
Fix jshint errors

phaseinducer:
Travis builds are broken because of small errors in js files, here is the fixing PR.


nidup:
GTM

PR #3187, opened 2015-08-25
Merge 1.3 into master

ClementGautier:
| Q                  | A
| ------------------ | ---
| Migration script?  | N/A
| Behats             | N/A
| Specs & Static     | N/A
| Changelog updated  | N/A

### DIFF
--- a/src/Pim/Bundle/DataGridBundle/Extension/MassAction/Util/ProductFieldsBuilder.php
+++ b/src/Pim/Bundle/DataGridBundle/Extension/MassAction/Util/ProductFieldsBuilder.php
@@ -7,7 +7,7 @@ use Pim\Bundle\CatalogBundle\Context\CatalogContext;
 use Pim\Bundle\CatalogBundle\Manager\AssociationTypeManager;
 use Pim\Bundle\CatalogBundle\Manager\CurrencyManager;
 use Pim\Bundle\CatalogBundle\Manager\LocaleManager;
-use Pim\Bundle\CatalogBundle\Manager\ProductManager;
+use Pim\Bundle\CatalogBundle\Manager\ProductManagerInterface;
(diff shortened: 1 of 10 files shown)


src/Pim/Bundle/DataGridBundle/Extension/MassAction/Util/ProductFieldsBuilder.php
```diff
@@ -40,14 +40,14 @@ class ProductFieldsBuilder
...
-     * @param CatalogContext         $catalogContext
+     * @param ProductManagerInterface $productManager
+     * @param LocaleManager           $localeManager
+     * @param CurrencyManager         $currencyManager
+     * @param AssociationTypeManager  $assocTypeManager
+     * @param CatalogContext          $catalogContext
      */
     public function __construct(
-        ProductManager $productManager,
+        ProductManagerInterface $productManager,
```
nidup: FYI we're trying to get rid of 'ProductManager' because it only provide shortcut to other services (as attribute repository) that we could inject directly, i understand why we fix like this in a patch (no BC break) but feel free to inject ProductRepositoryInterface and AttributeRepositoryInterface if you need them

ClementGautier: @nidup I know. In this case it meens a rewrites of the ProductFieldBuidler, plus we will have to create specifics services in the EE and review where it is injected. Not a big deal but we don't really have time for this right now.

nidup: yep i understand, would be sure you know the futureless of the ProductManager class :)

nidup:
GTM

PR #3219, opened 2015-08-28
PIM-4775: Added extension point for EE in order to process products d…

ClementGautier:
…epending on user permissions

| Q                  | A
| ------------------ | ---
| Migration script?  | N
| Behats             | N
| Specs & Static     | Y
| Changelog updated  | N

### DIFF
--- a/spec/Pim/Bundle/EnrichBundle/Connector/Processor/MassEdit/Product/EditCommonAttributesProcessorSpec.php
+++ b/spec/Pim/Bundle/EnrichBundle/Connector/Processor/MassEdit/Product/EditCommonAttributesProcessorSpec.php
@@ -47,7 +47,7 @@ class EditCommonAttributesProcessorSpec extends ObjectBehavior
         $this->setStepExecution($stepExecution)->shouldReturn($this);
     }
 
-    function it_does_not_sets_values_from_variant_group(
+    function it_does_not_set_values_when_attribute_is_not_editable(
         $validator,
         $propertySetter,
         AttributeInterface $attribute,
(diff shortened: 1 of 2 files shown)


spec/Pim/Bundle/EnrichBundle/Connector/Processor/MassEdit/Product/EditCommonAttributesProcessorSpec.php
```diff
@@ -47,7 +47,7 @@ function it_sets_the_step_execution(StepExecution $stepExecution)
         $this->setStepExecution($stepExecution)->shouldReturn($this);
     }
 
-    function it_does_not_sets_values_from_variant_group(
+    function it_does_not_sets_values_when_attribute_is_not_editable(
```
solivier: it does not set values...

skeleton:
GTM

src/Pim/Bundle/EnrichBundle/Connector/Processor/MassEdit/Product/EditCommonAttributesProcessor.php
```diff
@@ -147,4 +153,14 @@ protected function isProductValid(ProductInterface $product)
...
     }
+
+    /**
+     * @param ProductInterface $product
+     *
+     * @return bool
+     */
+    protected function isProductEditable(ProductInterface $product)
+    {
+        return true;
```
solivier: no logic ?

nidup:
GTM

src/Pim/Bundle/EnrichBundle/Connector/Processor/MassEdit/Product/EditCommonAttributesProcessor.php
ClementGautier: nope, only an extension point for EE, we could move the isProductValid logic in this function but this would create a BC break.

PR #3223, opened 2015-08-28
Generic features directory, can be anywhere

BitOne:
This PR allows to run Behat features wherever they are.

Main changes for CE pull request:
 - separator is now "," instead of ":"
 - paths has to be relative to the application root and not to the features directory anymore

This PR will introduce changes that will allow for EE:
 - only one execution of run_behat (so no more yellow CI, it will be either red or blue)
 - the feature list will be entered in only one spot (no separation between EE and CE), allowing easier control on what will be executed as well as easier interaction with existing CI errors analysis tools


jjanvier:
hmmmmm 

OK !

jmleroux:
GTM

PR #3344, opened 2015-09-28
[TR] PIM-4794: Display relevant thumbnails of assets (for products grid)

phaseinducer:
| Q                  | A
| ------------------ | ---
| Migration script?  | n/a
| Specs & Static     | no regression
| Changelog updated  | n/a

### DIFF
--- a/features/Context/Page/Base/Grid.php
+++ b/features/Context/Page/Base/Grid.php
@@ -375,6 +375,29 @@ class Grid extends Index
     }
 
     /**
+     * Get an image element inside a grid cell
+     *
+     * @param string $column
+     * @param string $row
+     *
+     * @throws \InvalidArgumentException
+     *
+     * @return NodeElement
+     */
+    public function getCellImage($column, $row)
+    {
+        $cell = $this->getColumnNode($column, $row);
+        $image = $cell->find('css', 'img');
+        if (null === $image) {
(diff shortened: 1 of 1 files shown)


jjanvier:
GTM

features/Context/Page/Base/Grid.php
```diff
@@ -375,6 +375,29 @@ public function getValuesInColumn($column)
...
+     * @throws \InvalidArgumentException
+     *
+     * @return NodeElement
+     */
+    public function getCellImage($column, $row)
+    {
+        $cell = $this->getColumnNode($column, $row);
+        $image = $cell->find('css', 'img');
+        if (null === $image) {
+            throw new \InvalidArgumentException(
```
willy-ahva: should be an Behat\Mink\Exception\ExpectationException imho

willy-ahva:
little comment then GTM

features/Context/Page/Base/Grid.php
phaseinducer: I don't agree because ExpectationException should be thrown only when an assertion fails, here it's just a little helper function. I don't "expect" the image to be here, I just want to find it.

willy-ahva: Don't really understand. If you try to find an image, you expect that there is one.

PR #3370, opened 2015-10-02
PIM-4955: Fixed regression on completeness computation when locales are removed from a channel

ClementGautier:
| Q                 | A
| ----------------- | ---
| Specs             | Y
| Behats            | Y 
| Blue CI           | Y
| Changelog updated | Y 
| Review and 2 GTM  | Y

### DIFF
--- a/CHANGELOG-1.4.md
+++ b/CHANGELOG-1.4.md
@@ -1,3 +1,8 @@
+# 1.4.x
+
+## Bug fixes
+- PIM-4955: Fixed regression on completeness computation when locales are removed from a channel
+
 # 1.4.2 (2015-10-01)
 
 ## Bug fixes
--- a/src/Pim/Bundle/CatalogBundle/Manager/CompletenessManager.php
+++ b/src/Pim/Bundle/CatalogBundle/Manager/CompletenessManager.php
@@ -125,6 +125,8 @@ class CompletenessManager
      * of a channel
      *
      * @param ChannelInterface $channel
+     *
+     * @deprecated To be removed in 1.5
      */
     public function scheduleForChannel(ChannelInterface $channel)
     {
--- a/src/Pim/Bundle/EnrichBundle/EventListener/Storage/ChannelLocaleSubscriber.php
+++ b/src/Pim/Bundle/EnrichBundle/EventListener/Storage/ChannelLocaleSubscriber.php
@@ -4,6 +4,7 @@ namespace Pim\Bundle\EnrichBundle\EventListener\Storage;
 
 use Akeneo\Component\StorageUtils\Saver\BulkSaverInterface;
 use Akeneo\Component\StorageUtils\StorageEvents;
+use Pim\Bundle\CatalogBundle\Doctrine\CompletenessGeneratorInterface;
 use Pim\Bundle\CatalogBundle\Model\ChannelInterface;
 use Pim\Bundle\CatalogBundle\Repository\LocaleRepositoryInterface;
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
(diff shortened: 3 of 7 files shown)


src/Pim/Bundle/EnrichBundle/EventListener/Storage/ChannelLocaleSubscriber.php
```diff
@@ -26,14 +27,22 @@ class ChannelLocaleSubscriber implements EventSubscriberInterface
...
+     * @param CompletenessGeneratorInterface $completeness
      */
-    public function __construct(LocaleRepositoryInterface $repository, BulkSaverInterface $saver)
-    {
-        $this->repository = $repository;
-        $this->saver      = $saver;
+    public function __construct(
+        LocaleRepositoryInterface $repository,
+        BulkSaverInterface $saver,
+        CompletenessGeneratorInterface $completeness
```
jjanvier: bc break...

CHANGELOG-1.4.md
```diff
@@ -1,3 +1,8 @@
+# 1.4.x
+
+## Bug fixes
+- PIM-4955: Fixed regression on completeness coputation when locale are removed from a channel
```
grena: computation*

src/Pim/Bundle/EnrichBundle/EventListener/Storage/ChannelLocaleSubscriber.php
ClementGautier: This subscriber was added in 1.4.0, I would be really surprised if someone already overrode it. Anyway its not a good practice to overrode a Subscriber/Listener as anyone can create a new one and make it executed before or after this listener. But yes its definitly a BC break as the service is public (About his, should we make our listeners and subscribers as private to reduce BC breaking news ?). 

@nidup: What do I do ? I could create a new subscriber that will do the same thing but with only this change. It will slightly make it heavier as we will do some database request twice but keep the BC between 1.4.0 and 1.4.3.

src/Pim/Bundle/CatalogBundle/Manager/CompletenessManager.php
```diff
@@ -121,23 +121,6 @@ public function scheduleForFamily(FamilyInterface $family)
     }
 
     /**
-     * Schedule recalculation of completenesses for all products
-     * of a channel
-     *
-     * @param ChannelInterface $channel
-     */
-    public function scheduleForChannel(ChannelInterface $channel)
```
nidup: we can't drop a public method in a patch, it's a BC break, you can mark it as deprecated if useless and drop it once merged on the master

src/Pim/Bundle/EnrichBundle/EventListener/Storage/ChannelLocaleSubscriber.php
nidup: agree with the fact that nobody has yet overriden it, but i would prefer hardly try to avoid BC Break.
in that case a BC free solution can be,

```php
public function __construct(LocaleRepositoryInterface $repository, BulkSaverInterface $saver, CompletenessGeneratorInterface $completeness = null) {
```


 - CompletenessGeneratorInterface $completeness = null in the construct
 -

src/Pim/Bundle/CatalogBundle/Manager/CompletenessManager.php
ClementGautier: ok then. I removed it as it is not a part of the CompletenessManagerInterface

src/Pim/Bundle/EnrichBundle/EventListener/Storage/ChannelLocaleSubscriber.php
ClementGautier: k, i'll do that

solivier:
GTM

nidup:
GTM

PR #3420, opened 2015-10-15
PIM-3071: Format the dates for the imports

skeleton:
| Q                 | A
| ----------------- | ---
| Specs             | :white_check_mark: 
| Behats            | :white_check_mark: 
| Blue CI           | ?
| Changelog updated | NO
| Review and 2 GTM  |
| Micro Demo (PO)   | :white_check_mark: 
| Migration script  | NO
| Tech Doc          | :white_check_mark:

### DIFF
--- a/src/Pim/Component/Localization/Localizer/DateLocalizer.php
+++ b/src/Pim/Component/Localization/Localizer/DateLocalizer.php
@@ -2,7 +2,8 @@
 
 namespace Pim\Component\Localization\Localizer;
 
-use Symfony\Component\OptionsResolver\OptionsResolver;
+use Pim\Component\Localization\Exception\FormatLocalizerException;
+use Symfony\Component\OptionsResolver\Exception\MissingOptionsException;
 
 /**
  * Check if date provided respects the format expected and convert it
@@ -29,7 +30,7 @@ class DateLocalizer implements LocalizerInterface
     /**
      * {@inheritdoc}
      */
-    public function isValid($date, array $options = [])
+    public function isValid($date, array $options = [], $attributeCode)
     {
         if (null === $date || '' === $date) {
             return true;
@@ -37,13 +38,9 @@ class DateLocalizer implements LocalizerInterface
 
         $this->checkOptions($options);
 
-        try {
-            $datetime = new \DateTime();
-            if(false === $datetime->createFromFormat($options['format_date'], $date)) {
-                return false;
-            }
-        } catch (\Exception $e) {
-            return false;
+        $datetime = $this->getDateTime($date, $options);
+        if (false === $datetime) {
+            throw new FormatLocalizerException($attributeCode, $options['format_date']);
         }
 
         return true;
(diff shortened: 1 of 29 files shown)


ClementGautier:
You've added `attributeCode` argument in the Localizer component, I'm not a big fan as is break the concern separation of this component. I would vote to remove this argument as it is only used to build exception messages.

src/Pim/Component/Localization/Localizer/DateLocalizer.php
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function convertLocalizedToDefault($date, array $options = [])
+    {
+        $this->checkOptions($options);
+
+        if (null === $date || '' === $date || $options['format_date'] === static::DEFAULT_DATE_FORMAT) {
```
ClementGautier: `$options['format_date'] === static::DEFAULT_DATE_FORMAT` is useless I think. You gain micro performance but you don't have the same behavior in case the date is not well formatted.

convertLocalizedToDefault('toto', ['format_date' => 'Y-m-d']) => returns toto
convertLocalizedToDefault('toto', ['format_date' => 'd-m-Y']) => throw an exception I guess during \DateTime::createFromFormat

src/Pim/Component/Localization/Localizer/DateLocalizer.php
```diff
@@ -0,0 +1,97 @@
...
+    {
+        $this->attributeTypes = $attributeTypes;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function isValid($date, array $options = [], $attributeCode)
+    {
+        if (null === $date || ''  === $date) {
```
pierallard: Extra space between '' and ===

src/Pim/Component/Localization/Localizer/DateLocalizer.php
```diff
@@ -0,0 +1,97 @@
...
+        return $datetime->createFromFormat($options['format_date'], $date);
+    }
+
+    /**
+     * @param array $options
+     */
+    protected function checkOptions(array $options)
+    {
+        if (!isset($options['format_date']) || '' === $options['format_date']) {
+            throw new MissingOptionsException('The options "format_date" do not exist.');
```
pierallard: The option without s, no ? (not sure about context)

ClementGautier:
GTM after comments fixed and CI blue

nidup:
GTM, bonus question what will happen with existing imports when a customer will migrate from 1.4 to 1.5 ? localization options will be ignored or existing import will fail ? 2 possibilities, ignore the options for old job instance configuration (with a default config?) or provide a migration script.

skeleton:
@nidup There is a default configuration defined in ProductProcessor. If an old job is not edited, this configuration is applied

nidup:
@skeleton perfect then :)

PR #3431, opened 2015-10-19
[TR] PIM-2735: Add a simple fill field method to bypass PEF override method

grena:
| Q                 | A
| ----------------- | ---
| Added Specs       | N
| Added Behats      | N
| Travis CI is ok   | Y
| Changelog updated | N

### DIFF
--- a/features/Context/Page/Base/Base.php
+++ b/features/Context/Page/Base/Base.php
@@ -82,6 +82,21 @@ class Base extends Page
     }
 
     /**
+     * @param string $locator
+     * @param string $value
+     */
+    public function simpleFillField($locator, $value)
+    {
+        $field = parent::findField($locator);
+
+        if (null === $field) {
+            throw $this->elementNotFound('form field', 'id|name|label|value', $locator);
+        }
+
+        $field->setValue($value);
+    }
+
+    /**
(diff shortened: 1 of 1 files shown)


willy-ahva:
GTM

features/Context/Page/Base/Base.php
```diff
@@ -82,6 +82,21 @@ public function fillField($locator, $value)
     }
 
     /**
+     * @param string $locator
+     * @param string $value
+     */
+    public function simpleFillField($locator, $value)
```
ClementGautier: Instead of this, can we modify the fillField method to not do the specific stuff that you dont want in your case ?

ClementGautier: Is the issue comes from the executeScript ?

grena: We override `fillField` 2 or 3 times in children classes for PEF specific stuff and other things, using a simple other method was simple to me.

ClementGautier: My 2 cents about that was to make fillField usable anywhere to not have to remember that we have a simpleFillField

solivier:
GTM

PR #3437, opened 2015-10-20
Fix 1.4

skeleton:
| Q                 | A
| ----------------- | ---
| Specs             | N
| Behats            | Fix behats
| Blue CI           | ?
| Changelog updated | Y
| Review and 2 GTM  | -

### DIFF
--- a/features/deprecated-behat/import/browse_imports.feature
+++ b/features/deprecated-behat/import/browse_imports.feature
@@ -4,6 +4,7 @@ Feature: Browse imports
   As a product manager
   I need to be able to view a list of them
 
+  @ce
   Scenario: Successfully view, sort and filter import jobs
     Given an "apparel" catalog configuration
     And I am logged in as "Julia"
--- a/src/Pim/Bundle/ReferenceDataBundle/DataGrid/Datasource/ResultRecord/MongoDbOdm/Product/ReferenceDataTransformer.php
+++ b/src/Pim/Bundle/ReferenceDataBundle/DataGrid/Datasource/ResultRecord/MongoDbOdm/Product/ReferenceDataTransformer.php
@@ -25,10 +25,11 @@ class ReferenceDataTransformer
     public function transform(array $result, array $attribute, $locale, $scope)
     {
         $attributeCode = $attribute['code'];
-        $normalizedData = $result['normalizedData'];
-        $fromNormData = ['pim_reference_data_simpleselect', 'pim_reference_data_multiselect'];
+        $properties = $attribute['properties'];
 
-        if (in_array($attribute['attributeType'], $fromNormData)) {
+        if (isset($properties['reference_data_name']) && '' !== $properties['reference_data_name']) {
+            $normalizedData = $result['normalizedData'];
+
             $fieldCode = ProductQueryUtility::getNormalizedValueField(
                 $attributeCode,
                 $attribute['localizable'],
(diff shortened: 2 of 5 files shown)


features/deprecated-behat/import/browse_imports.feature
```diff
@@ -4,6 +4,7 @@ Feature: Browse imports
   As a product manager
   I need to be able to view a list of them
 
+  @ce
```
phaseinducer: Does it work ? Why not playing this scenario in EE ?

jjanvier:
gtm

src/Pim/Bundle/ReferenceDataBundle/DataGrid/Datasource/ResultRecord/MongoDbOdm/Product/ReferenceDataTransformer.php
```diff
@@ -25,10 +25,11 @@ class ReferenceDataTransformer
     public function transform(array $result, array $attribute, $locale, $scope)
     {
         $attributeCode = $attribute['code'];
-        $normalizedData = $result['normalizedData'];
-        $fromNormData = ['pim_reference_data_simpleselect', 'pim_reference_data_multiselect'];
+        $properties = $attribute['properties'];
+
+        if (isset($properties['reference_data_name']) && '' !== $properties['reference_data_name']) {
```
grena: Simple question, what are we checking here?

grena:
Just a question, but GTM anyway

phaseinducer:
specs are broken by your fix
GTM otherwise

features/deprecated-behat/import/browse_imports.feature
skeleton: Scenario fails on `Then the grid should contain 7 elements`
We have 7 jobs in CE, but 8 in EE.
And yes, it works

src/Pim/Bundle/ReferenceDataBundle/DataGrid/Datasource/ResultRecord/MongoDbOdm/Product/ReferenceDataTransformer.php
skeleton: We check if current attribute is a reference data. 
Before we had static attribute type. But it doesn't work on mongo EE with assets.

features/deprecated-behat/import/browse_imports.feature
phaseinducer: Wouldn't it be better to remove the assertion "the grid should contain 7 elements" and keep this in EE ?

ClementGautier:
GTM after travis fixed

features/deprecated-behat/import/browse_imports.feature
skeleton: nope, this scenario : 
- count items in list
- check what jobs are return with filters

If I remove all lines which will break CI, this scenario has no sense

src/Pim/Bundle/ReferenceDataBundle/DataGrid/Datasource/ResultRecord/MongoDbOdm/Product/ReferenceDataTransformer.php
grena: Thanks :)

PR #3441, opened 2015-10-20
[TR] PIM-5037 : Do not cascade detach product associated entities

jmleroux:
| Q                 | A
| ----------------- | ---
| Specs             |no new specs
| Behats            |no new behat
| Blue CI           |yes
| Changelog updated |yes
| Review and 2 GTM  |

### DIFF
--- a/CHANGELOG-1.3.md
+++ b/CHANGELOG-1.3.md
@@ -2,6 +2,7 @@
 
 ## Technical improvements
 - PIM-5028: Fix log levels configuration for prod and dev environments
+- PIM-5037: Do not cascade detach product associated entities
 
 # 1.3.27 (2015-10-02)
 
--- a/src/Pim/Bundle/CatalogBundle/Resources/config/model/doctrine/Association.orm.yml
+++ b/src/Pim/Bundle/CatalogBundle/Resources/config/model/doctrine/Association.orm.yml
@@ -34,8 +34,6 @@ Pim\Bundle\CatalogBundle\Model\Association:
     manyToMany:
         groups:
             targetEntity: Pim\Bundle\CatalogBundle\Entity\Group
-            cascade:
-                - detach
             joinTable:
                 name: pim_catalog_association_group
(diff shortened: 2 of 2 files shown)


nidup:
Means when bulk update and detach products all associated products and groups will stay in memory (for instance when exec rules on products associated to other products) in other hand we could handle that with a specific product detacher if we need, it's weird to have only the cascade detach

jmleroux:
@nidup you're right, but it is better to trend memory in favor of stability.
The memory problem should be addressed next.

nidup:
agree, it was imho a hack to add this cascade detach, GTM

ClementGautier:
Did you look at the git history ? It is possible that this was made for a memory, probably on a SLA, that will show up again if we don't find the proper fix. We should investigate to find where we need to cascade detach and do it manually.

jmleroux:
@ClementGautier yes, ilook at the history, and i think it should remember something to you ;)

BitOne:
GTM for me.

In all cases, cascade must be done only when the target object is really a part of the source object.
For example, cascade between values and product make sense, because values doesn't exist without the products.
For associations and products, it doesn't make sense as products targeted by an association can live without the association itself.

But as @ClementGautier said, if there was a memory leak somewhere, we may need to detach manually. But this particuliar relationship (product->association->products) is really difficult to manage as we sometimes need to process products that belong to the association at the same time as the product owner of the association.

jmleroux:
@ClementGautier does that mean we must reopen PIM-4810 : #3227 ?

BitOne:
But the fix introducted in this #3441 PR doesn't break the behat done for the #3227 . So either the Behat gives a false positive, or this fix doesn't break the #3227 fix ;)

ClementGautier:
Strange. Probably something done after the fix I did on #3227 fixed it by side effect. If the behat is ok after a cache clear then GTM and no need to reopen PIM-4810 as the behat cover this issue.

PR #3444, opened 2015-10-21
[TR] Move catalog model interfaces

nidup:
Most of our Pim components rely on few interfaces located in CatalogBundle (models and repository interfaces), these interfaces have been introduced there is long time ago, when we didn't use components.

These uses make the "where i put my code rule" not clear between component and bundle, a component should never rely on a bundle and should never rely on doctrine orm/mongo stuff, cf https://github.com/nidup/php-coupling-detector (holidays homework!)

Having these "exception" make the rule unclear and this PR begin to fix this point.

To avoid a very large BC Break, we keep the old interface as  Pim/Bundle/CatalogBundle/Model/ProductInterface which extends a new Pim/Component/Catalog/Model/ProductInterface and we mark the old interface as deprecated (third party implementations will continue to work).

I also add a sed command in the upgrade.md to easily change any use of old interface in project code.

Finally, this change, discussed with Benoit, is safe and allow to reduce our identified coupling problems from 360 to 234 (identified with this https://github.com/nidup/php-coupling-detector), most of remaining coupling issues can be addressed easily with the same strategy.

| Q                 | A
| ----------------- | ---
| Specs             | Y
| Behats            | Y
| Blue CI           | Y
| Changelog updated | Y
| Review and 2 GTM  | ?
| Micro Demo (PO)   | N
| Migration script  | Y (sed cmds)
| Tech Doc          |


features/Context/FixturesContext.php
```diff
@@ -25,12 +25,12 @@
 use Pim\Bundle\CatalogBundle\Entity\GroupType;
 use Pim\Bundle\CatalogBundle\Model\Association;
 use Pim\Bundle\CatalogBundle\Model\FamilyInterface;
-use Pim\Bundle\CatalogBundle\Model\LocaleInterface;
-use Pim\Bundle\CatalogBundle\Model\ProductInterface;
 use Pim\Bundle\CommentBundle\Entity\Comment;
 use Pim\Bundle\CommentBundle\Model\CommentInterface;
 use Pim\Bundle\DataGridBundle\Entity\DatagridView;
 use Pim\Bundle\UserBundle\Entity\User;
+use Pim\Component\Catalog\LocaleInterface;
```
skeleton: normal to have LocaleInterface outside `Model` folder ?

spec/Pim/Bundle/BaseConnectorBundle/Processor/ProductToFlatArrayProcessorSpec.php
```diff
@@ -6,10 +6,10 @@
 use League\Flysystem\Filesystem;
 use PhpSpec\ObjectBehavior;
 use Pim\Bundle\CatalogBundle\Manager\ChannelManager;
-use Pim\Bundle\CatalogBundle\Model\AttributeInterface;
-use Pim\Bundle\CatalogBundle\Model\ChannelInterface;
-use Pim\Bundle\CatalogBundle\Model\ProductInterface;
-use Pim\Bundle\CatalogBundle\Model\ProductValueInterface;
+use Pim\Component\Catalog\Model\AttributeInterface;
+use Pim\Component\Catalog\ChannelInterface;
```
skeleton: idem

features/Context/FixturesContext.php
nidup: good catch, thx! (i change)

PR #3459, opened 2015-10-23
[TR] PIM-5024: Set Lato as default font

pierallard:
| Q                 | A
| ----------------- | ---
| Specs             | n
| Behats            | n
| Blue CI           | y
| Changelog updated | n
| Review and 2 GTM  | n
| Micro Demo (PO)   | y
| Migration script  | n
| Tech Doc          | n

Just ask if you want a test-dev with this PR! http://test-dev-feature-1.akeneo.com/
- it adds Lato file and css definition
- it changes every "font-size" to set a variable accessible in main variables file, and adapt it to the new font width. The next step will be to group these font sizes to only have a little set (4 or 5).
- it removes some useless css


ClementGautier:
GTM for me after rebase

BitOne:
Do we know the Lato font license ? Do we have it somewhere ?

pierallard:
@BitOne Excellent question. The licence is "SIL Open Font License v1.10"... I have no idea if we have to add it into licence file.

ronanguilloux:
* [Lato website](http://www.latofonts.com/) (_"high-quality open-source font family"_)
* [SIL Open Font License v1.1](http://scripts.sil.org/cms/scripts/page.php?site_id=nrsi&id=OFL)
* complies with [FSF](http://www.gnu.org/licenses/license-list.html#SILOFL), [DFSG](http://www.debian.org/social_contract) and [OSI](http://opensource.org/docs/definition.php)
* [term and conditions to respect](https://en.wikipedia.org/wiki/SIL_Open_Font_License#Terms).

BitOne:
Thanks @ronanguilloux, it still not quite clear if we need to provides the SIL license file with the fonts.

At least, we could add a mention to our license or maybe a new section in the README file so we can explain under which license the fonts are distributed with Akeneo.

I'd rather do a little bit more than needed on this kind of subject than not enough. ;)

ronanguilloux:
@BitOne [no need to provide the SIL license file according the SIL FAQ](http://scripts.sil.org/cms/scripts/page.php?item_id=OFL-FAQ_web#36914bf6) since we bundle the font in a program:

>Question: 1.10 Does the full OFL license text always need to accompany the font?
>Answer: The only situation in which an OFL font can be distributed without the text of the OFL (either in a separate file or in font metadata), is when a font is embedded in a document or bundled within a program. In the case of metadata included within a font, it is legally sufficient to include only a link to the text of the OFL on http://scripts.sil.org/OFL, but we strongly recommend against this. Most modern font formats include metadata fields that will accept the full OFL text, and full inclusion increases the likelihood that users will understand and properly apply the license.

pierallard:
I added a `LICENCE.txt` in the fond folders with the SIL licence (same as @BitOne it's better to provide more than less). Is it ok for you ?

BitOne:
@pierallard Okay for me !

ronanguilloux:
FYI, most modern font formats include __metadata fields__ that will accept the full OFL text, but I do not know if you use a font format compatible w/ metadata.

grena:
GTM

phaseinducer:
GTM :+1:

PR #3460, opened 2015-10-23
PIM-5051 : Bulk remove throw an exception on versionning

jmleroux:
| Q                 | A
| ----------------- | ---
| Specs             |updated
| Behats            |no new feature
| Blue CI           |running
| Changelog updated |yes
| Review and 2 GTM  |OK

### DIFF
--- a/src/Pim/Bundle/VersioningBundle/EventSubscriber/AddRemoveVersionSubscriber.php
+++ b/src/Pim/Bundle/VersioningBundle/EventSubscriber/AddRemoveVersionSubscriber.php
@@ -97,6 +97,7 @@ class AddRemoveVersionSubscriber implements EventSubscriberInterface
             ->setSnapshot(null !== $previousVersion ? $previousVersion->getSnapshot(): [])
             ->setChangeset([]);
 
-        $this->versionSaver->save($version);
+        $options = $event->getArguments();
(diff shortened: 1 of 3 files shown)


src/Pim/Bundle/VersioningBundle/EventSubscriber/AddRemoveVersionSubscriber.php
```diff
@@ -97,6 +97,7 @@ public function postRemove(RemoveEvent $event)
             ->setSnapshot(null !== $previousVersion ? $previousVersion->getSnapshot(): [])
             ->setChangeset([]);
 
-        $this->versionSaver->save($version);
+        $options = $event->getArguments();
+        $this->versionSaver->save($version, $options);
```
grena: Will it fail if the event has no argument?

grena:
Question, but GTM

solivier:
GTM

src/Pim/Bundle/VersioningBundle/EventSubscriber/AddRemoveVersionSubscriber.php
fitn: As the RemoveEvent takes an empty array as default parameter, it won't fail ;)

PR #3475, opened 2015-10-27
Extend REST API to return all products

paulclarkin:
I have extended the REST API to support Get Products. In addition to http://pim-dev.local/api/rest/products/sku-000.json you can now hit http://pim-dev.local/api/rest/products to return an array of all products.

| Q                 | A
| ----------------- | ---
| Added Specs       | Y
| Added Behats      | Y
| Travis CI is ok   | not yet
| Changelog updated | Y

Disclaimer: I'm not a PHP Developer (.NET is my home!) so please forgive any novice PHP mistakes. Any advice would be appreciated!

Thanks

### DIFF
--- a/src/Pim/Bundle/WebServiceBundle/Controller/Rest/ProductsController.php
+++ b/src/Pim/Bundle/WebServiceBundle/Controller/Rest/ProductsController.php
@@ -1,0 +1,119 @@
+<?php
+
+namespace Pim\Bundle\WebServiceBundle\Controller\Rest;
+
+use FOS\RestBundle\Controller\Annotations\NamePrefix;
+use FOS\RestBundle\Controller\Annotations\RouteResource;
+use FOS\RestBundle\Controller\FOSRestController;
+use Nelmio\ApiDocBundle\Annotation\ApiDoc;
+use Pim\Bundle\CatalogBundle\Model\ProductInterface;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\Security\Core\Exception\AccessDeniedException;
+
+/**
+ * Product API controller
+ *
+ * @author    Filips Alpe <filips@akeneo.com>
+ * @copyright 2013 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ *
+ * @RouteResource("products")
+ * @NamePrefix("oro_api_")
+ */
+class ProductsController extends FOSRestController
+{
+    /**
+     * Get products
+     *
+     * @param Request $request
+     *
+     * @ApiDoc(
+     *      description="Get a single product",
+     *      resource=true
+     * )
+     *
+     * @return Response
+     */
+    public function getAction(Request $request)
+    {
+        $userContext       = $this->get('pim_user.context.user');
+        $availableChannels = array_keys($userContext->getChannelChoicesWithUserChannel());
+        $availableLocales  = $userContext->getUserLocaleCodes();
+
+        $channels = $request->get('channels', $request->get('channel', null));
+        if ($channels !== null) {
+            $channels = explode(',', $channels);
+
+            foreach ($channels as $channel) {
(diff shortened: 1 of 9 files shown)


src/Pim/Bundle/WebServiceBundle/Controller/Rest/ProductsController.php
```diff
@@ -0,0 +1,119 @@
...
+use Nelmio\ApiDocBundle\Annotation\ApiDoc;
+use Pim\Bundle\CatalogBundle\Model\ProductInterface;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\Security\Core\Exception\AccessDeniedException;
+
+/**
+ * Product API controller
+ *
+ * @author    Filips Alpe <filips@akeneo.com>
```
filipsalpe: Humm, did I do that?

nidup:
Hello @paulclarkin,

Thank you very much for this Pull Request, one of the biggest ever made by the community!

The Web API is a very sensitive topic, because once defined, we have to maintain it for all coming versions to avoid any issues with the following upgrades.

We released this very limited API (get one product) in an early version and we never enhanced it, we plan to enhance it in the future but, for now, it's not planed for the v1.5.

Your PR brings real business value for users of the v1.3, v1.4, an coming v1.5, I think the best move is to package it as a new optional bundle, for instance, EnhancedWebAPIBundle, there is a couple of resources on how to extract this code in a bundle http://symfony.com/doc/current/cookbook/bundles/best_practices.html and http://symfony.com/doc/current/cookbook/bundles/installation.html.

With this move, anybody will be able to use this bundle and the extended API in his project, (we have a marketplace to reference these third party bundles http://marketplace.akeneo.com/ )

In the future, when we'll handle the enhanced web API we'll be able to discuss with you, if you're interested to see your original work integrated as a default feature in the PIM, we'll be able to do so.

We'll be happy to bring you our support and help to create this bundle in a way which will ensure that it will continue to work with future versions without big maintaince.

About the implementation in itself,

You could integrate a pagination by default because with a big catalog (let's say 1M of products), the call will fail.

You should avoid to update both repositories (orm and mongo) but rely on your own querying service which may encapsulate the ProductQueryBuilderInterface.

This ProductQueryBuilderInterface can provide you a full support for filters and sorters (for both orm and mongo), you can take a look on this command which is quite close of what you could achieve as an advanced API https://github.com/akeneo/pim-community-dev/blob/master/src/Pim/Bundle/CatalogBundle/Command/QueryProductCommand.php

About the serialization, you can create a service by taking inspiration from https://github.com/akeneo/pim-community-dev/blob/master/src/Pim/Bundle/WebServiceBundle/Handler/Rest/ProductHandler.php and do
```
        $data = $this->serializer->serialize(
            $products, // an array of products here
            'json',
            [
                'locales'     => $locales,
                'channels'    => $channels,
                'resource'    => $url,
                'filter_type' => 'pim.external_api.product.view'
            ]
        );
```

Don't hesitate if you need any further information.
Best regards,

paulclarkin:
Thanks for your large comment, I'll review, digest and possibly submit a stand alone module.

Further to your note about Web APIs, I've spent quite some time looking at various PIMs and many have no/bad API. In 2015, integrating with systems via a REST API is the standard, and the expected mechanism by developers.

If Akeneo were to build a good API, it would compliment the already good UI to put Akeneo into a great position in the market.

(Not everyone wants to integrate with Magento!)

src/Pim/Bundle/WebServiceBundle/Controller/Rest/ProductsController.php
paulclarkin: Ah! Good spot!

saulhudson:
@paulclarkin 

Hi Paul, did you ever manage to to create this API feature as a Bundle? I have a similar requirement and I was hoping you could share some of your work with me to aid my development.

Many Thanks

paulclarkin:
@saulhudson sorry, i didn't take this any further. I'm not working on this any more, but feel free to fork this and do it yourself.

PR #3494, opened 2015-10-28
Grid filter "in list" for text attributes

Hocdoc:
As described in the [Akeneo forum](http://www.akeneo.com/forums/topic/grid-filter-in-list-for-text-attributes/) we have to filter products by a list of Strings.
I think this feature is useful for many users, so I created a pull request for this configuration.

I also think that the "in list"-filter would be useful for the other attribute types like f.ex. numbers, media, prices...

### DIFF
--- a/src/Pim/Bundle/CatalogBundle/Resources/config/query_builders.yml
+++ b/src/Pim/Bundle/CatalogBundle/Resources/config/query_builders.yml
@@ -163,7 +163,7 @@ services:
         arguments:
             - @pim_catalog.validator.helper.attribute
             - ['pim_catalog_text', 'pim_catalog_textarea']
-            - ['STARTS WITH', 'ENDS WITH', 'CONTAINS', 'DOES NOT CONTAIN', '=', 'EMPTY']
+            - ['STARTS WITH', 'ENDS WITH', 'CONTAINS', 'DOES NOT CONTAIN', '=', 'EMPTY', 'IN']
         tags:
             - { name: 'pim_catalog.doctrine.query.filter', priority: 30 }
 
--- a/src/Pim/Bundle/DataGridBundle/Resources/config/attribute_types.yml
+++ b/src/Pim/Bundle/DataGridBundle/Resources/config/attribute_types.yml
@@ -22,6 +22,7 @@ parameters:
                 field_options:
                     attr:
                         empty_choice: true
+                        choice_list: true
         sorter:          product_value
     pim_datagrid.product.attribute_type.pim_catalog_textarea:
         column:


nidup:
Hi @Hocdoc 

This PR is very interesting, we provide this behavior for "pim_catalog_identifier" (sku by default) which use the same filter class, so from backend point of view it works well. The problem comes from the javascript behavior of our "in list" filter.

Let say i have the following products:
```
    And the following products:
      | sku     | name-en_US    |
      | BOOTBXS | Boot black xs |
      | BOOTWXS | Boot white xs |
      | BOOTBS  | Boot black s  |
      | BOOTBL  | Mug           |
      | BOOTRXS |               |
```

And I would filter in the product grid using the "in list" operator, if i begin to type "Boot black xs", it will split words "boot", "black", "xs", we don't have this issue with sku because we don't allow spaces in a sku.

in-list

An option could be to split words in the component only when typing "," for instance, but to allow this I need to double check with UX team if it makes the original feature less usable (i asked them and keep you posted soon).

There is a git diff adding a behat scenario to test that the backend behavior is ok,
```
diff --git a/features/filter/string.feature b/features/filter/string.feature
index 572576c..2b82e61 100644
--- a/features/filter/string.feature
+++ b/features/filter/string.feature
@@ -13,13 +13,14 @@ Feature: Filter on string
       | BOOTBL  | Mug           |
       | BOOTRXS |               |
     Then I should get the following results for the given filters:
-      | filter                                                                                        | result                                     |
-      | [{"field":"name", "operator":"STARTS WITH",      "value": "Boot",         "locale": "en_US"}] | ["BOOTBXS", "BOOTWXS", "BOOTBS"]           |
-      | [{"field":"name", "operator":"EMPTY",            "value": null,           "locale": "en_US"}] | ["BOOTRXS"]                                |
-      | [{"field":"name", "operator":"NOT EMPTY",        "value": null,           "locale": "en_US"}] | ["BOOTBXS", "BOOTWXS", "BOOTBS", "BOOTBL"] |
-      | [{"field":"name", "operator":"STARTS WITH",      "value": "boot",         "locale": "en_US"}] | ["BOOTBXS", "BOOTWXS", "BOOTBS"]           |
-      | [{"field":"name", "operator":"ENDS WITH",        "value": "xs",           "locale": "en_US"}] | ["BOOTBXS", "BOOTWXS"]                     |
-      | [{"field":"name", "operator":"CONTAINS",         "value": "black",        "locale": "en_US"}] | ["BOOTBXS", "BOOTBS"]                      |
-      | [{"field":"name", "operator":"DOES NOT CONTAIN", "value": "Boot",         "locale": "en_US"}] | ["BOOTBL"]                                 |
-      | [{"field":"name", "operator":"=",                "value": "Boot black s", "locale": "en_US"}] | ["BOOTBS"]                                 |
-      | [{"field":"name", "operator":"=",                "value": "Mug ",         "locale": "en_US"}] | []                                         |
+      | filter                                                                                                           | result                                     |
+      | [{"field":"name", "operator":"STARTS WITH",      "value": "Boot",         "locale": "en_US"}]                    | ["BOOTBXS", "BOOTWXS", "BOOTBS"]           |
+      | [{"field":"name", "operator":"EMPTY",            "value": null,           "locale": "en_US"}]                    | ["BOOTRXS"]                                |
+      | [{"field":"name", "operator":"NOT EMPTY",        "value": null,           "locale": "en_US"}]                    | ["BOOTBXS", "BOOTWXS", "BOOTBS", "BOOTBL"] |
+      | [{"field":"name", "operator":"STARTS WITH",      "value": "boot",         "locale": "en_US"}]                    | ["BOOTBXS", "BOOTWXS", "BOOTBS"]           |
+      | [{"field":"name", "operator":"ENDS WITH",        "value": "xs",           "locale": "en_US"}]                    | ["BOOTBXS", "BOOTWXS"]                     |
+      | [{"field":"name", "operator":"CONTAINS",         "value": "black",        "locale": "en_US"}]                    | ["BOOTBXS", "BOOTBS"]                      |
+      | [{"field":"name", "operator":"DOES NOT CONTAIN", "value": "Boot",         "locale": "en_US"}]                    | ["BOOTBL"]                                 |
+      | [{"field":"name", "operator":"=",                "value": "Boot black s", "locale": "en_US"}]                    | ["BOOTBS"]                                 |
+      | [{"field":"name", "operator":"=",                "value": "Mug ",         "locale": "en_US"}]                    | []                                         |
+      | [{"field":"name", "operator":"IN",               "value": ["Boot black xs", "Boot black s"], "locale": "en_US"}] | ["BOOTBS", "BOOTBXS"]                      |
diff --git a/src/Pim/Bundle/CatalogBundle/Resources/config/query_builders.yml b/src/Pim/Bundle/CatalogBundle/Resources/config/query_builders.yml
index 308fbfb..03a5caa 100644
--- a/src/Pim/Bundle/CatalogBundle/Resources/config/query_builders.yml
+++ b/src/Pim/Bundle/CatalogBundle/Resources/config/query_builders.yml
@@ -163,7 +163,7 @@ services:
         arguments:
             - @pim_catalog.validator.helper.attribute
             - ['pim_catalog_text', 'pim_catalog_textarea']
-            - ['STARTS WITH', 'ENDS WITH', 'CONTAINS', 'DOES NOT CONTAIN', '=', 'EMPTY', 'NOT EMPTY']
+            - ['STARTS WITH', 'ENDS WITH', 'CONTAINS', 'DOES NOT CONTAIN', '=', 'EMPTY', 'NOT EMPTY', 'IN']
         tags:
             - { name: 'pim_catalog.doctrine.query.filter', priority: 30 }
 
diff --git a/src/Pim/Bundle/DataGridBundle/Resources/config/attribute_types.yml b/src/Pim/Bundle/DataGridBundle/Resources/config/attribute_types.yml
index 89fbdf1..4fa189d 100644
--- a/src/Pim/Bundle/DataGridBundle/Resources/config/attribute_types.yml
+++ b/src/Pim/Bundle/DataGridBundle/Resources/config/attribute_types.yml
@@ -22,6 +22,7 @@ parameters:
                 field_options:
                     attr:
                         empty_choice: true
+                        choice_list: true

```

nidup:
Hi @Hocdoc 
I discussed with our product owner, the use case for the sku is to be able to copy a list of sku from an excel file and directly paste them into the search field to filter the grid. We can't change this behavior to make it usable for every text field in grid filters. Thank you again for the proposal and sorry to not be able to merge this PR.

Hocdoc:
Thanks for your review including so much helpful information.

(We used the in-list filter exactly for our users who wants to copy&paste IDs from Excel for a SKU-like attribute. Unfortunately the GUI of the filter doesn't scale well for about more than 50 IDs, so we are using now another work-around to get a list of products filtered by text).

PR #3532, opened 2015-11-03
[WIP] PIM-5118: csv product export is now creating multiple csv files

rybus:
Native CSV export now writes CSv files by chuck of 1 000 products per file. This is to avoid memory leak and keep memory usage reasonable.

By default, output will be like : 
 `/tmp/product-YYYY-MM-DD_HH:MM:SS-X.csv` where X is the number of the file.

It can be changed by using `%filenumber%` in the filename pattern.

| Q                 | A
| ----------------- | ---
| Added Specs       | No
| Added Behats      | No
| Travis CI is ok   | N/A
| Changelog updated | No


nidup:
thank you for the PR, it has been fixed by https://github.com/akeneo/pim-community-dev/commit/13b5fd743cd35d9d20d7f365f35531fe4bd28f57 to avoid BC break on the export format

PR #3599, opened 2015-11-18
[TR] PIM-5170 - Fixes MongoDB memory leak at import time.

BitOne:
This PR fixes two problems:
 - missing `detach()` calls on ProductProcessor and ProductAssociationProcessor
 - buggy `cascadeDetach()` on our MongoDB ODM version (see https://github.com/doctrine/mongodb-odm/pull/979/ ) by reimplementing a proper `cascadeDetach()` on our ObjectDetacher()

This second fix must be not ported on master. We must switch to a stable version of MongoDB ODM . (See PIM-4981).

| Q                 | A
| ----------------- | ---
| Added Specs       | No
| Added Behats      | No
| Travis CI is ok   | In progress
| Changelog updated | yes

### DIFF
--- a/src/Akeneo/Bundle/StorageUtilsBundle/Doctrine/Common/Detacher/ObjectDetacher.php
+++ b/src/Akeneo/Bundle/StorageUtilsBundle/Doctrine/Common/Detacher/ObjectDetacher.php
@@ -4,9 +4,12 @@ namespace Akeneo\Bundle\StorageUtilsBundle\Doctrine\Common\Detacher;
 
 use Akeneo\Component\StorageUtils\Detacher\BulkObjectDetacherInterface;
 use Akeneo\Component\StorageUtils\Detacher\ObjectDetacherInterface;
+use Doctrine\Common\Collections\Collection;
 use Doctrine\Common\Persistence\ManagerRegistry;
 use Doctrine\Common\Persistence\ObjectManager;
 use Doctrine\Common\Util\ClassUtils;
+use Doctrine\ODM\MongoDB\DocumentManager;
+use Doctrine\ODM\MongoDB\PersistentCollection;
 
 /**
  * Detacher, detaches an object from its ObjectManager
@@ -34,7 +37,13 @@ class ObjectDetacher implements ObjectDetacherInterface, BulkObjectDetacherInter
     public function detach($object)
     {
         $objectManager = $this->getObjectManager($object);
-        $objectManager->detach($object);
+
+        if ($objectManager instanceof DocumentManager) {
+            $visited = [];
+            $this->doDetach($object, $visited);
+        } else {
+            $objectManager->detach($object);
+        }
     }
 
     /**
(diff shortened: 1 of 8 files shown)


src/Pim/Component/Connector/Processor/Denormalization/ProductProcessor.php
```diff
@@ -122,7 +123,7 @@ public function process($item)
         }
 
         $violations = $this->validateProduct($product);
-        if ($violations->count() > 0) {
+        if ($violations && $violations->count() > 0) {
```
jjanvier: don't think we need that, a `ConstraintViolationListInterface` is always returned

src/Pim/Component/Connector/Processor/Denormalization/ProductProcessor.php
```diff
@@ -108,6 +108,7 @@ public function process($item)
             $filteredItem = $this->filterIdenticalData($product, $filteredItem);
 
             if (empty($filteredItem)) {
+                $this->detachProduct($product);
```
jjanvier: just to understand, that fixes another bug, doesn't it?

jjanvier:
would be really cool to split into 2 commits :)
to be able to directly revert the commit about the buggy doctrine cascade detach when we'll merge on master :)

src/Akeneo/Bundle/StorageUtilsBundle/Doctrine/Common/Detacher/ObjectDetacher.php
```diff
@@ -34,7 +37,13 @@ public function __construct(ManagerRegistry $registry)
     public function detach($object)
     {
         $objectManager = $this->getObjectManager($object);
-        $objectManager->detach($object);
+
+        if ($objectManager instanceof DocumentManager) {
```
nidup: +1, with that, no BC break and no re-work to do when merging in v1.5

src/Pim/Component/Connector/Processor/Denormalization/ProductProcessor.php
BitOne: Okay, it was to be coherent with what has been done on the ProductAssociationProcessor.

src/Pim/Bundle/ConnectorBundle/Resources/config/processors.yml
```diff
@@ -71,6 +71,7 @@ services:
             - '@pim_catalog.updater.product'
             - '@pim_catalog.validator.product'
             - '@pim_catalog.comparator.filter.product_association'
+            - '@akeneo_storage_utils.doctrine.object_detacher'
```
BitOne: Nope, because this memory leak is everywhere where the object detacher is used with Mongo (rules engine, mass edit, variant group template applier, etc...)
It's a trashable fix, as in 1.5 shouldn't be applied.

src/Akeneo/Bundle/StorageUtilsBundle/Doctrine/Common/Detacher/ObjectDetacher.php
BitOne: BC break if we introduce a MongoDB object detacher, as it wil not be applied automatically on specific customer code and will not be applied on all other part where the object detacher is used.

Again, this is a fix that will not be applied on master.

BitOne:
@jjanvier : Split into two commits with updated specs on Processors.

jjanvier:
that breaks my heart
but GTM regarding with explanations :)

BitOne:
@jjanvier : You will take pleasure to kill this cascadeDetach code on master once we are working on a stable version of MongoDB ODM ;)

PR #3605, opened 2015-11-19
[TR] PIM-5106: Refactor WebUser to find compared fields

pierallard:
| Q                 | A
| ----------------- | ---
| Specs             | y
| Behats            | y
| Blue CI           | nope
| Changelog updated | y
| Review and 2 GTM  | TODO
| Micro Demo (PO)   | TODO
| Migration script  | n
| Tech Doc          | n

### DIFF
--- a/features/Context/WebUser.php
+++ b/features/Context/WebUser.php
@@ -742,6 +742,7 @@ class WebUser extends RawMinkContext
     /**
      * @param string $label
      * @param string $expected
+     * @param bool   $comparePanel
      *
      * @Then /^the field ([^"]*) should contain "([^"]*)"$/
      *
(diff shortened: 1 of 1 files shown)


features/Context/WebUser.php
```diff
@@ -750,10 +751,10 @@ public function theProductFieldValueShouldBe($fieldName, $expected = '')
      *
      * TODO: should be moved to a page context and theProductFieldValueShouldBe() method should be merged with this one
      */
-    public function theFieldShouldContain($label, $expected)
+    public function theFieldShouldContain($label, $expected, $comparePanel = false)
     {
         $this->wait();
-        $field = $this->getCurrentPage()->findField($label);
+        $field = $this->getCurrentPage()->findField($label, $comparePanel);
```
skeleton: why `$comparePanel` naming ?

features/Context/WebUser.php
```diff
@@ -750,10 +751,10 @@ public function theProductFieldValueShouldBe($fieldName, $expected = '')
      *
      * TODO: should be moved to a page context and theProductFieldValueShouldBe() method should be merged with this one
      */
-    public function theFieldShouldContain($label, $expected)
+    public function theFieldShouldContain($label, $expected, $comparePanel = false)
```
phaseinducer: You should never add a parameter in a step signature that cannot be set by its regexp. It doesn't make sense.
Plus I think that what you are trying to do here already exists : https://github.com/akeneo/pim-community-dev/blob/master/features/Context/WebUser.php#L859

PR #3608, opened 2015-11-19
[TR] Merge 1 3 into 1 4

jjanvier:


src/Pim/Bundle/CatalogBundle/Doctrine/MongoDBODM/CompletenessGenerator.php
```diff
@@ -140,7 +140,10 @@ public function getCompletenesses(array $normalizedData, array $normalizedReqs)
         $normalizedData = array_filter(
             $normalizedData,
             function ($value) {
-                return (null !== $value);
+                //TODO: should be dropped on master as we don't normalize anymore empty medias in the normalized data
+                $nullNormalizedMedia = ['filename' => null, 'originalFilename' => null];
+
+                return (null !== $value && $nullNormalizedMedia !== $value);
```
ClementGautier: why don't you does:

```php
return null !== $value && null !== $value['filename'] && null !== $value['originalFilename'];
```

ClementGautier:
a little comment but as it's not your code (I assume its only a merge PR) GTM even if you don't want to change it.

jjanvier:
it was my code from the 1.3 fix

I agree with you, the var was just here to make understand this hackish fix

nidup:
GTM

PR #3611, opened 2015-11-19
Revert "[SLA][TR]PIM-5194: fix perf on json family generation"

jjanvier:
Reverts akeneo/pim-community-dev#3583

We have identifiied a new bug with this patch. The permissions are not applied correctly on Enterprise Edition (ie: attributes are not filtered by permissions). 

The Behat _Compare and copy localized fields from different sources // Successfully copy value from my draft_ fails.


grena:
:'( GTM

PR #3613, opened 2015-11-19
attribute label size extended from 100 to 255

danielsan80:
Just a suggestion. To allow a user to use a long label to describe an attribute.
It seams to not break the layout and it give more freedom to the user.


nidup:
Hi @danielsan80 
Thank you for the proposal, unfortunately, allow longer labels has impacts on the UI display on smaller resolution (product edit form, product grid). Don't hesitate to open another pull request to discuss any other topic.
Have a good day.

PR #3643, opened 2015-11-25
[WIP][SLA] PIM-5215: Improve product import with family change

damien-carcel:
Create empty product values for new family attributes

| Q                 | A
| ----------------- | ---
| Specs             | Running, pass localy
| Behats            | Running
| Blue CI           | Running
| Changelog updated | Yes
| Review and 2 GTM  |

### DIFF
--- a/src/Pim/Component/Connector/Processor/Denormalization/ProductProcessor.php
+++ b/src/Pim/Component/Connector/Processor/Denormalization/ProductProcessor.php
@@ -115,7 +115,7 @@ class ProductProcessor extends AbstractProcessor
         }
 
         try {
-            $this->updateProduct($product, $filteredItem);
+            $this->updateProduct($product, $filteredItem, $familyCode);
         } catch (\InvalidArgumentException $exception) {
             $this->detachProduct($product);
             $this->skipItemWithMessage($item, $exception->getMessage(), $exception);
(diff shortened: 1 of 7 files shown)


src/Pim/Component/Connector/Processor/Denormalization/ProductProcessor.php
```diff
@@ -352,12 +352,22 @@ protected function findOrCreateProduct($identifier, $familyCode)
...
     {
+        $originalFamily = null;
+        if (null !== $familyCode && null !== $product->getId()) {
+            $originalFamily = $product->getFamily();
+        }
+
         $this->updater->update($product, $filteredItem);
+
+        if (null !== $originalFamily && $originalFamily->getCode() !== $familyCode) {
+            $this->builder->addMissingProductValues($product);
```
nidup: I do understand the purpose of this fix, imho, this logic could be done in the updater itself, it's general business logic, as optional values, not only an import concern cf https://github.com/akeneo/pim-community-dev/blob/1.4/src/Pim/Component/Catalog/Updater/ProductUpdater.php#L192
In other hand, in updater, it would have an impact on all other features (PEF, mass edit, rules engine).
I would give a try in updater and run behat to see if it break other features.
Any opinion on this topic @jjanvier ?

damien-carcel: I try the updater in this branch: https://github.com/akeneo/pim-community-dev/tree/PIM-5215-product-updater.
However, the problem is that I need to know active scopes and locales to set properly the new empty product values, which I don't have access to in the updater, so I need to inject the ProductBuilder, which is kind of a BC.

Another approach is to handle this not in the imports, but in the exports (product export and quick-export). What is wanted here in the end it to have all exported product(s) column, even if empty. It would be easier and have no performance impact to just add the missing column on the export.

PR #3723, opened 2015-12-08
[TMP] PIM-5034

pierallard:


src/Pim/Component/Localization/Presenter/PricesPresenter.php
```diff
@@ -38,4 +48,20 @@ function ($price) use ($numberFormatter) {
...
+     * @param string $versionedAttribute
+     *
+     * @return array
+     */
+    protected function getStructuredPrice($price, $versionedAttribute)
+    {
+        $parts = preg_split('/-/', $versionedAttribute);
+        $currency = end($parts);
+
+        return ['data' => (double) $price, 'currency' => $currency];
```
skeleton: (float)

src/Pim/Component/Localization/Presenter/PricesPresenter.php
```diff
@@ -19,7 +19,17 @@ class PricesPresenter extends NumberPresenter
      */
     public function present($prices, array $options = [])
     {
+        if ('' === $prices || null === $prices) {
+            return $prices;
+        }
+
         if (!is_array($prices)) {
+            if (array_key_exists('versioned_attribute', $options)) {
```
skeleton: use isset instead of array_key_exists

src/Pim/Component/Localization/Presenter/PricesPresenter.php
```diff
@@ -19,7 +19,17 @@ class PricesPresenter extends NumberPresenter
...
+                $structuredPrice = $this->getStructuredPrice($prices, $options['versioned_attribute']);
+
+                return $this->present($structuredPrice, $options);
```
skeleton: I don't really like this syntax. and this method begin to be complicated to read.
```
if (!is_array($prices)) {
    if (!isset($options['versioned_attribute'])) {
        return parent::present($prices, $options);
    }

    $prices = $this->getStructuredPrice($prices, $options['versioned_attribute']);
}
```
this piece of code is more readable & simple IMO

PR #3734, opened 2015-12-10
[TR][SLA]PIM-5295: Fix association product grid category filter

ClementGautier:
| Q                 | A
| ----------------- | ---
| Specs             | N/A
| Behats            | Y
| Blue CI           | Running
| Changelog updated | Y
| Review and 2 GTM  | TODO

Actually when you go to a product page, throuth the grid filtered by a category, their is a bug in the association panel, the products displayed in the datagrid are filtered by the category used in the product grid but, as the filter is hidden (you dont see the category tree in this page) you are not able to see that the grid is filtered. The quick fix it to completely remove the filter on the product association datagrid so the filter from the product grid dont apply here.

Dev Note concerning Behat: I'm trying here to use only Behat methods so I overrode some native function in order to be able to use them in a SPA context, wrapping them in a spin. This way we could be able to use them in the future as it / without a page context.

### DIFF
--- a/features/Context/FeatureContext.php
+++ b/features/Context/FeatureContext.php
@@ -412,4 +412,28 @@ class FeatureContext extends MinkContext implements KernelAwareInterface
     {
         return $this->getContainer()->get('pim_enrich.mailer.mail_recorder');
     }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function clickLink($link)
+    {
+        $this->spin(function () use ($link) {
+            parent::clickLink($link);
+
+            return true;
+        });
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function assertNumElements($num, $element)
+    {
+        $this->spin(function () use ($num, $element) {
+            parent::assertNumElements($num, $element);
+
+            return true;
+        });
+    }
 }
--- a/features/product/associate_product.feature
+++ b/features/product/associate_product.feature
@@ -7,14 +7,14 @@ Feature: Associate a product
   Background:
     Given a "footwear" catalog configuration
     And the following products:
-      | sku            |
-      | charcoal-boots |
-      | black-boots    |
-      | gray-boots     |
-      | brown-boots    |
-      | green-boots    |
-      | shoelaces      |
-      | glossy-boots   |
+      | sku            | categories        |
+      | charcoal-boots | Summer_collection |
+      | black-boots    |                   |
+      | gray-boots     |                   |
+      | brown-boots    |                   |
+      | green-boots    |                   |
+      | shoelaces      |                   |
+      | glossy-boots   |                   |
     And I am logged in as "Julia"
 
   Scenario: Associate a product to another product
@@ -167,3 +167,15 @@ Feature: Associate a product
     And I edit the "red-boots" product
     When I visit the "Associations" tab
     Then I should be able to sort the rows by Is associated
+
+  @jira https://akeneo.atlassian.net/browse/PIM-5295
+  Scenario: Association product grid is not filtered by the category selected in the product grid
+    Given I am on the products page
+    When I filter by "category" with value "Summer_collection"
+    Then I should see product charcoal-boots
+    And I should not see product black-boots
+    When I click on the "charcoal-boots" row
+    And I follow "Associations"
+    Then I should see 6 "#grid-association-product-grid tbody tr" elements
+    When I follow "Upsell"
+    Then I should see 6 "#grid-association-product-grid tbody tr" elements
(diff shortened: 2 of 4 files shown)


features/Context/NavigationContext.php
```diff
@@ -725,7 +725,14 @@ public function openPage($page, array $options = [])
         $this->currentPage = $page;
 
         /** @var Base $page */
```
phaseinducer: Could you please :fire: that while you're here ? =)

features/Context/NavigationContext.php
```diff
@@ -725,7 +725,14 @@ public function openPage($page, array $options = [])
...
-        $page = $this->getCurrentPage()->open($options);
+        $page = $this->getCurrentPage();
+
+        // Page already open
+        if (preg_match('#'.$page->getUrl($options).'#', $this->getSession()->getCurrentUrl())) {
```
phaseinducer: Don't use regexps if you can avoid it.
`false !== strpos($this->getSession()->getCurrentUrl(), $page->getUrl($options))` should work here.

features/product/associate_product.feature
```diff
@@ -167,3 +167,16 @@ Feature: Associate a product
...
+  @jira https://akeneo.atlassian.net/browse/PIM-5295
+  Scenario: Association product grid is not filtered by the category selected in the product grid
+    Given I am on the products page
+    When I filter by "category" with value "Summer_collection"
+    Then I should see product charcoal-boots
+    And I should not see product black-boots
+    When I click on the "charcoal-boots" row
+    And I edit the "charcoal-boots" product
+    And I visit the "Associations" tab
+    Then I should see 6 "#grid-association-product-grid tbody tr" elements
```
phaseinducer: Why using a selector here ?

features/product/associate_product.feature
```diff
@@ -167,3 +167,16 @@ Feature: Associate a product
...
     When I visit the "Associations" tab
     Then I should be able to sort the rows by Is associated
+
+  @jira https://akeneo.atlassian.net/browse/PIM-5295
+  Scenario: Association product grid is not filtered by the category selected in the product grid
+    Given I am on the products page
+    When I filter by "category" with value "Summer_collection"
+    Then I should see product charcoal-boots
+    And I should not see product black-boots
+    When I click on the "charcoal-boots" row
```
phaseinducer: Why click before editing ?

features/Context/NavigationContext.php
```diff
@@ -724,8 +724,23 @@ public function openPage($page, array $options = [])
...
 
-        /** @var Base $page */
-        $page = $this->getCurrentPage()->open($options);
+        $page = $this->getCurrentPage();
+
+        if ($page instanceof Base) {
+            $path = parse_url($this->getSession()->getCurrentUrl(), PHP_URL_FRAGMENT);
+
+            // If their is no hashnav we use the path
+            if (null === $path && false !== strpos('url=', $path)) {
```
phaseinducer: shoudl be a `||` here

phaseinducer: Did you mean `||` ?

phaseinducer:
GTM after fix

features/Context/FeatureContext.php
```diff
@@ -412,4 +412,36 @@ public function getMailRecorder()
...
         return $this->getContainer()->get('pim_enrich.mailer.mail_recorder');
     }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function clickLink($link)
+    {
+        return $this->spin(function () use ($link) {
+            try {
```
phaseinducer: I don't think a try / catch is needed since the spin catches exceptions. So this is enough :
```
return $this->spin(function () use ($link) {
    parent::clickLink($link);
}
```

features/Context/FeatureContext.php
```diff
@@ -412,4 +412,36 @@ public function getMailRecorder()
     {
         return $this->getContainer()->get('pim_enrich.mailer.mail_recorder');
     }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function clickLink($link)
+    {
+        return $this->spin(function () use ($link) {
```
phaseinducer: Parent method returns nothing.

features/Context/FeatureContext.php
```diff
@@ -412,4 +412,36 @@ public function getMailRecorder()
...
+            }
+        });
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function assertNumElements($num, $element)
+    {
+        return $this->spin(function () use ($num, $element) {
```
phaseinducer: same

features/product/associate_product.feature
```diff
@@ -167,3 +167,15 @@ Feature: Associate a product
...
+    And I follow "Associations"
+    Then I should see 6 "#grid-association-product-grid tbody tr" elements
```
phaseinducer: Still not a huge fan of coupling steps and DOM :/

phaseinducer:
little comment & GTM

solivier:
GTM

PR #3793, opened 2015-12-22
[SLA] PIM-5387: Fix memory leak on quick export

rybus:
| Q                 | A
| ----------------- | ---
| Specs             | No
| Behats            | No
| Blue CI           | Not launched
| Changelog updated | Yes
| Review and 2 GTM  | Pending

# Memory leak
Let 270k products, with 8k attributes
Perform a quick export of ~nk products

x : number of products
y : memory usage (byte)
memory_leak_quick_export

### DIFF
--- a/src/Pim/Bundle/EnrichBundle/Connector/Processor/QuickExport/ProductToFlatArrayProcessor.php
+++ b/src/Pim/Bundle/EnrichBundle/Connector/Processor/QuickExport/ProductToFlatArrayProcessor.php
@@ -3,6 +3,7 @@
 namespace Pim\Bundle\EnrichBundle\Connector\Processor\QuickExport;
 
 use Akeneo\Bundle\BatchBundle\Item\InvalidItemException;
+use Akeneo\Component\StorageUtils\Detacher\ObjectDetacherInterface;
 use Pim\Bundle\CatalogBundle\AttributeType\AttributeTypes;
 use Pim\Bundle\CatalogBundle\Builder\ProductBuilderInterface;
 use Pim\Bundle\CatalogBundle\Exception\InvalidArgumentException;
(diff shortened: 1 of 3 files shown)


src/Pim/Bundle/EnrichBundle/Connector/Processor/QuickExport/ProductToFlatArrayProcessor.php
```diff
@@ -105,6 +119,13 @@ public function process($product)
 
         $data['product'] = $this->serializer->normalize($product, 'flat', $this->getNormalizerContext());
 
+        if (null !== $this->objectDetacher) {
+            $this->objectDetacher->detach($product);
+        }
+        if (null !== $this->entityManager) {
+            $this->entityManager->clear();
```
nidup: This is a very hardcore clear of the UOW, we did that with the old import system and we had several issues with this, we prefer use an explicit detach of the products.

```$this->objectDetacher->detach($product);``` is not enough? You have related entities kept in memory?

Comment (nidup) [inline src/Pim/Bundle/EnrichBundle/Connector/Processor/QuickExport/ProductToFlatArrayProcessor.php]: Why we don't fix the https://github.com/akeneo/pim-community-dev/blob/master/src/Pim/Bundle/BaseConnectorBundle/Processor/ProductToFlatArrayProcessor.php the classic product export is not impacted ?

**Comment by nidup** [comment]
thank you to detail your scenario in the description of the PR with,
 - which volume are you testing
 - indicate the memory usage before the fix (which page size for products?)
 - which memory usage with only the product detach
 - which memory usage with the massive clear of the UOW

thx!

File: src/Pim/Bundle/EnrichBundle/Connector/Processor/QuickExport/ProductToFlatArrayProcessor.php [hunk 2]
```diff
@@ -3,6 +3,9 @@
 namespace Pim\Bundle\EnrichBundle\Connector\Processor\QuickExport;
 
 use Akeneo\Bundle\BatchBundle\Item\InvalidItemException;
+use Akeneo\Bundle\StorageUtilsBundle\Doctrine\Common\Detacher\ObjectDetacher;
+use Akeneo\Component\StorageUtils\Detacher\ObjectDetacherInterface;
+use Doctrine\ORM\EntityManager;
```
nidup: no used anymore

src/Pim/Bundle/EnrichBundle/Resources/config/connector/processors.yml
```diff
@@ -55,3 +55,5 @@ services:
             - '@pim_catalog.manager.channel'
             - %upload_dir%
             - '@pim_catalog.builder.product'
+            - '@akeneo_storage_utils.doctrine.object_detacher'
+            - '@doctrine.orm.entity_manager'
```
nidup: entity manager is useless

nidup:
After discussion with @BitOne and @jjanvier we should detach in the FilteredProductReader, it will be more consistent with the standard product export where we detach in the ODMProductReader

In case of import, the writer detaches objects, in case of exports, the reader detaches each time its fetches another page of objects

nidup:
After re-discussion, if we do in the processor, it will have an impact on mass edit
 - in one case (quick export), we read and not update the objects
 - in anothercase (mass edit), we read and update objects (and the writer clear the UOW by detaching objects)

nidup:
Missing phpspec then GTM

PR #3820, opened 2016-01-04
PIM-4897: count on product values is now bypassed with high volume of data

rybus:
The  `count()` inside `ProductValueCounterRepository` wasn't working witth a high volume of data and lead to an error:
`exception: distinct too big, 16mb cap`

| Q                 | A
| ----------------- | ---
| Specs             | Yes
| Behats            | No
| Blue CI           | N/A
| Changelog updated | No
| Review and 2 GTM  | OK

### DIFF
--- a/src/Pim/Bundle/CatalogBundle/Collector/DatabaseCollector.php
+++ b/src/Pim/Bundle/CatalogBundle/Collector/DatabaseCollector.php
@@ -57,8 +57,8 @@ class DatabaseCollector extends DataCollector
     {
         $this->data = [
             'mongodb_enabled'     => $this->isMongoDbEnabled(),
-            'product_value_count' => $this->productValueRepository->count(),
             'require_mongodb'     => !($this->isMongoDbEnabled() || !$this->isMongoDbRequired()),
+            'product_value_count' => $this->isMongoDbEnabled() ? null : $this->getProductValueCount(),
             'version'             => [
                 'patch' => $this->versionProvider->getPatch(),
                 'minor' => $this->versionProvider->getMinor()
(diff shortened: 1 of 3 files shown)


jjanvier:
gtm

rybus:
I left the MongoDB implementation but I don't use it anymore in case MongoDB is enabled.

nidup:
GTM on this one even with nit using it to avoid future perf issue. Thx

src/Pim/Bundle/CatalogBundle/Collector/DatabaseCollector.php
```diff
@@ -123,12 +123,20 @@ public function isMongoDbRequired()
...
     }
 
     /**
+     * @return int
+     */
+    public function getProductValueCount()
+    {
+        return $this->isMongoDbEnabled() ? 0 : $this->productValueRepository->count();
+    }
+
```
nidup: i don't understand why 2 public methods productValueCount and getProductValueCount ?

src/Pim/Bundle/CatalogBundle/Collector/DatabaseCollector.php
```diff
@@ -57,8 +57,8 @@ public function collect(Request $request, Response $response, \Exception $except
     {
         $this->data = [
             'mongodb_enabled'     => $this->isMongoDbEnabled(),
-            'product_value_count' => $this->productValueRepository->count(),
             'require_mongodb'     => !($this->isMongoDbEnabled() || !$this->isMongoDbRequired()),
+            'product_value_count' => $this->getProductValueCount(),
```
nidup: i would add the following logic here ```$this->isMongoDbEnabled() ? null : $this->getProductValueCount();``` and always call the $this->productValueRepository->count(); in the getProductValueCount().

because from the outside, when I call a method named getProductValueCount i expect to have the count of values, not 0 sometimes due to a very technical reason :)

once this will be fix, it's GTM for me.

bonus question, why do we expose so much public methods in the collector? are they really used from the outside, if not they should be protected.

PR #3821, opened 2016-01-04
Return at least a minor hint, if file upload fail.

mathielen:
In my case, there was already a directory /tmp/pim/file_storage, but not the target directory "/tmp/pim/file_storage/b/f/8/0".
Thus, somehow the recursive mkdir-call in Symfony\Component\HttpFoundation\File\File->getTargetFile(...) failed and throw an exception.
After deleting the whole /tmp/pim directory, everything worked.

The best thing would be to have the real exception-message in the HTTP-400 result, which might be a security problem....

### DIFF
--- a/src/Pim/Bundle/EnrichBundle/Controller/Rest/MediaController.php
+++ b/src/Pim/Bundle/EnrichBundle/Controller/Rest/MediaController.php
@@ -47,6 +47,7 @@ class MediaController
      */
     public function postAction(Request $request)
     {
+        /** @var \Symfony\Component\HttpFoundation\File\UploadedFile $file */
         $file       = $request->files->get('file');
         $violations = $this->validator->validate($file);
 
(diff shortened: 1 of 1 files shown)


nidup:
Hi @mathielen 

Thank you for the feedback and for the pull request!

I do agree with your explaination, your proposal is not ideal but, at least, makes the failure more explicit.

I would be glad to count you as our contributors, to be able to merge your PR, I would ask you to read and sign our contributor agreement http://www.akeneo.com/contributor-license-agreement/

jmleroux:
I will run our CI on this PR while waiting for the agreement.

jmleroux:
CI is ok and we will be able to merge this PR once we merged 1.4 into master.
Thanks a lot for this contribution.

jjanvier:
GTM

nidup:
GTM

nidup:
Merged, thank you @mathielen!

PR #3822, opened 2016-01-04
[TR] Remove incoherent scenario

phaseinducer:
This scenario is useless since an export job always has a channel. It's incoherent because the job is created directly in database without using the UI.
Plus it's broken on master because the validation error is not displayed anymore on the show page.


nidup:
GTM

grena:
GTM

PR #3833, opened 2016-01-06
[SLA][TR] PIM-5331: Fix product save issue when categories having code as integer

pierallard:
| Q                 | A
| ----------------- | ---
| Specs             | n
| Behats            | y
| Blue CI           | y
| Changelog updated | y
| Review and 2 GTM  | n

The issue is the params sent to product update. When category code is an integer, it's sent as an integer, like this:
`[ "categories" => [ 0 => 123 ] ]`
This fix cast category codes to string, like this:
`[ "categories" => [ 0 => "123" ] ]`

### DIFF
--- a/CHANGELOG-1.4.md
+++ b/CHANGELOG-1.4.md
@@ -2,6 +2,7 @@
 
 ## Bug fixes
 - PIM-5405: Fix content type for stream upload
+- PIM-5331: Fix product save issue when categories have code as integer
 
 # 1.4.15 (2015-12-30)
 
--- a/features/Context/WebUser.php
+++ b/features/Context/WebUser.php
@@ -1693,6 +1693,30 @@ class WebUser extends RawMinkContext
     }
 
     /**
+     * @param string $sku
+     * @param string $categoryCode
+     *
+     * @Then /^the category of (?:the )?product "([^"]*)" should be "([^"]*)"$/
+     */
+    public function theCategoryOfProductShouldBe($sku, $categoryCode)
+    {
+        $this->clearUOW();
+        $product = $this->getFixturesContext()->getProduct($sku);
+
+        $categoryCodes = $product->getCategoryCodes();
+        assertEquals(
+            [$categoryCode],
+            $categoryCodes,
+            sprintf(
+                'Expecting the category of "%s" to be "%s", not "%s".',
+                $sku,
+                $categoryCode,
+                implode(', ', $categoryCodes)
+            )
+        );
+    }
+
+    /**
      * @param int $count
      *
      * @Then /^there should be (\d+) updates?$/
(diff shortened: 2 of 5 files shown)


features/product/save_product_when_integer_category_code.feature
```diff
@@ -0,0 +1,24 @@
+@javascript
+Feature: Create an attribute
```
jjanvier: if the feature is create an attribute, it should not be here in the product folder
plus, can't you add these scenarios in a file that already exists and that fits to your feature?

jjanvier:
gtm except comment

skeleton:
did you try to import a category with a code as an integer ?
maybe we should add a test in https://github.com/akeneo/pim-community-dev/blob/master/features/import/import_categories.feature#L17

pierallard:
@jjanvier @skeleton IMHO the 2 scenarios should be in a specific file, because the background is specific for this use case.

CHANGELOG-1.4.md
```diff
@@ -2,6 +2,7 @@
 
 ## Bug fixes
 - PIM-5405: Fix content type for stream upload
+- PIM-5331: Fix product save issue when categories having code as integer
```
phaseinducer: "have numeric code" ?

features/Context/WebUser.php
```diff
@@ -1693,6 +1693,30 @@ public function theFamilyOfProductShouldBe($sku, $expectedFamily = '')
     }
 
     /**
+     * @param string $sku
+     * @param string $categoryCode
+     *
+     * @Then /^the category of (?:the )?product "([^"]*)" should be "([^"]*)"$/
+     */
+    public function theCategoryOfProductShouldBe($sku, $categoryCode)
```
phaseinducer: Should not be in WebUser imho

src/Pim/Bundle/EnrichBundle/Resources/public/js/product/form/categories.js
```diff
@@ -164,7 +164,7 @@ define(
                     var $categoryElement = this.$('#node_' + id);
                     var $rootElement     = $categoryElement.closest('.root-unselectable');
                     this.cache[id] = {
-                        code: $categoryElement.data('code'),
+                        code: String($categoryElement.data('code')),
```
phaseinducer: What is the problem and why does this fix it ?

features/Context/WebUser.php
pierallard: Just above, there is productShouldBeDisabled, productShouldBeEnabled, theFamilyOfProductShouldBe... Where do you think this function should be ?

src/Pim/Bundle/EnrichBundle/Resources/public/js/product/form/categories.js
pierallard: Added a note in the description.

features/Context/WebUser.php
phaseinducer: Ok then, we'll move it later

phaseinducer:
GTM

phaseinducer:
GTM

PR #3859, opened 2016-01-12
Completeness calculation in Behats now uses CLI

rybus:
| Q                 | A
| ----------------- | ---
| Specs             | No
| Behats            | Yes
| Blue CI           | To be ran
| Changelog updated | N/A
| Review and 2 GTM  | Pending


skeleton:
GTM is CI CE & EE are blue

PR #3863, opened 2016-01-12
PIM-5379: Remove completeness calculation from MongoDB saver

fitn:
Just some words to explain this PR.

3 issues are fixed here:

1. Useless completeness scheduling.
Completeness is overridden each time we save the product with this direct to db saver. No need to reschedule it, it will be done.

2. Wrong completeness generation.
The real saving part is done at the end of the `saveAll` method calling `insertDocuments` and `updateDocuments`. As long these methods are not called, the MongoDB collection current state is before the product update. In this case, the completeness is done on the "old" datas.

3. Memory leak on the product saver.
The rescheduling of the completeness is done with: `CompletenessGenerator::schedule(ProductInterface)`. It calls a clear to the collection. Accordingly, this will schedule the completeness to be removed in the UOW.
But now, we don't work anymore with the UOW here and there is no flush or clear of it. So products are detached from the UOW but completenesses are not and they are linked to products. Then products will be kept in memory.

This PR consists in removing the completeness scheduling and move the completeness calculation to call it after the product has been updated.

### DIFF
--- a/src/Pim/Bundle/CatalogBundle/Doctrine/MongoDBODM/Saver/ProductSaver.php
+++ b/src/Pim/Bundle/CatalogBundle/Doctrine/MongoDBODM/Saver/ProductSaver.php
@@ -105,14 +105,6 @@ class ProductSaver extends BaseProductSaver
             } else {
                 $productsToUpdate[] = $product;
             }
-
-            if (true === $options['schedule'] || true === $options['recalculate']) {
-                $this->completenessManager->schedule($product);
-            }
-
-            if (true === $options['recalculate']) {
-                $this->completenessManager->generateMissingForProduct($product);
-            }
         }
 
         $insertDocs = $this->getDocsFromProducts($productsToInsert);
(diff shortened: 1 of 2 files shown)


nidup:
Behat CE : http://core-ci.akeneo.com/job/dci_behat/485/
Behat EE : http://core-ci.akeneo.com/job/dci_behat/486/

src/Pim/Bundle/CatalogBundle/Doctrine/MongoDBODM/Saver/ProductSaver.php
```diff
@@ -105,14 +105,6 @@ public function saveAll(array $products, array $options = [])
             } else {
                 $productsToUpdate[] = $product;
             }
-
-            if (true === $options['schedule'] || true === $options['recalculate']) {
```
jmleroux: You should not completly remove the use of `schedule`

jmleroux:
GTM, good catch.
Thanks for this very clear description. For next time, you should also add the header array. See https://github.com/akeneo/pim-community-dev/blob/master/CONTRIBUTING.md for more info.

nidup:
Rebase and GTM, thank you very much for the PR!

PR #3866, opened 2016-01-13
[TR] PIM-5021: Forbid the use of code 'category' for an attribute

pierallard:
| Q                 | A
| ----------------- | ---
| Specs             | y
| Behats            | y
| Blue CI           | TODO
| Changelog updated | n
| Review and 2 GTM  | n

* The CHANGELOG needs to be updated or not ? *

### DIFF
--- a/CHANGELOG-1.4.md
+++ b/CHANGELOG-1.4.md
@@ -4,6 +4,7 @@
 - PIM-5213: Paginate loading of attributes on Product Edit Form and Mass Edit Common Attributes action
 
 ## Bug fixes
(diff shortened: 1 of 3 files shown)


solivier:
no behats ?

grena:
GTM, but how does it behave with an existing attribute with this code? It could raise save problems no?

solivier:
GTM

pierallard:
@grena When an attribute is called "category", the grid is not loaded. So I think this can even be doable.

grena:
Ok then, thanks for your answer!

PR #3911, opened 2016-01-21
[TR] PIM-5473 + PIM-5474: Improve completeness rescheduling on family

fitn:
2 tickets in this PR (1 commit per ticket) about scalability improvements:
- PIM-5473: When launching the completeness scheduling per family, it happens there is a timeout on MongoDB cursor. The request is well done but it returns an error due to the timeout (default conf = 30 seconds). I reproduce with almost 50k products in my family.

The improvement consists in doing many little queries instead of a huge one which take too much time. As we can't use skip + limit on a MongoDB query using the "multiple" option (and we need it to impact many documents), I implemented an internal pagination to iterate on ids and then call the update.
- PIM-5474: When saving the family, the call to the completeness scheduling takes a long time and it blocks the UI.
  The solution consists in launching a dedicated command as a backend task.

| Q | A |
| --- | --- |
| Added Specs | N |
| Added Behats | N |
| Travis CI is ok | ? |
| Changelog updated | N |

### DIFF
--- a/src/Akeneo/Component/Console/CommandLauncher.php
+++ b/src/Akeneo/Component/Console/CommandLauncher.php
@@ -96,4 +96,16 @@ class CommandLauncher
 
         return $result;
     }
+
+    /**
+     * Builds logfile path from root directory
+     *
+     * @param string $logfile
+     *
+     * @return string
+     */
+    public function buildLogfilePath($logfile)
+    {
+        return $this->rootDir .'/logs/'. $logfile;
+    }
 }
--- a/src/Pim/Bundle/CatalogBundle/Command/CompletenessFamilySchedulerCommand.php
+++ b/src/Pim/Bundle/CatalogBundle/Command/CompletenessFamilySchedulerCommand.php
@@ -1,0 +1,75 @@
+<?php
+
+namespace Pim\Bundle\CatalogBundle\Command;
+
+use Pim\Bundle\CatalogBundle\Doctrine\CompletenessGeneratorInterface;
+use Symfony\Bundle\FrameworkBundle\Command\ContainerAwareCommand;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+
+/**
+ * Reschedule completeness for family command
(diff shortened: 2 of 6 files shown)


src/Pim/Bundle/CatalogBundle/Command/CompletenessFamilySchedulerCommand.php
```diff
@@ -0,0 +1,75 @@
...
+
+use Pim\Bundle\CatalogBundle\Doctrine\CompletenessGeneratorInterface;
+use Symfony\Bundle\FrameworkBundle\Command\ContainerAwareCommand;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+
+/**
+ * Reschedule completeness for family command
+ *
```
solivier: missing licence

src/Pim/Bundle/CatalogBundle/Manager/CompletenessManager.php
```diff
@@ -110,13 +117,16 @@ public function schedule(ProductInterface $product)
...
      * of a family
+     * It could be long so it's launched as a backend task
      *
      * @param FamilyInterface $family
      */
     public function scheduleForFamily(FamilyInterface $family)
     {
         if ($family->getId()) {
-            $this->generator->scheduleForFamily($family);
+            $cmd = sprintf('pim:completeness:schedule-family %s', $family->getCode());
```
jjanvier: imvho this should be the responsibility of the business object to decide if things should be launched in background or not
so we should:
1. deprecate or remove the method
2. change everywhere it's called

src/Akeneo/Component/Console/CommandLauncher.php
```diff
@@ -96,4 +96,16 @@ public function executeForeground($command)
...
     }
+
+    /**
+     * Builds logfile path from root directory
+     *
+     * @param string $logfile
+     *
+     * @return string
+     */
+    public function buildLogfilePath($logfile)
```
jjanvier: not a big fan of those of public methods that do nothing
but OK here

PR #3922, opened 2016-01-25
[TR] Fix edit_variant_group_attribute_value:115

willy-ahva:
| Q                 | A
| ----------------- | ---
| Specs             | /
| Behats            | 
| Blue CI           | running
| Changelog updated | no
| Review and 2 GTM  |

Trying to fix `pim-community-dev/features/enrich/variant-group/edit_variant_group_attribute_value.feature:115`  -> `Scenario: Change a pim_catalog_image attribute of a variant group and ensure saving`.

The scenario try to add a media in a VG, save and then uncheck a product which is link to the VG and save. Then the page is reloaded and the row should be uncheck. When the scenario fails the row is checked :
fail behat

Maybe the row is not well unchecked at the previous step so I added a spin which verifies if the row is well checked/unchecked. If the scenario continues to fail, it will give us some tips to know where it fails.

### DIFF
--- a/features/Context/DataGridContext.php
+++ b/features/Context/DataGridContext.php
@@ -755,7 +755,14 @@ class DataGridContext extends RawMinkContext implements PageObjectAwareInterface
                 throw $this->createExpectationException(sprintf('Unable to find a checkbox for row %s', $row));
             }
 
-            $checkbox->check();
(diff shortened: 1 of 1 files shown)


jmleroux:
not GTM.
Check should not be spinned. We should spin for finding elements, not to check or click on them.

PR #3924, opened 2016-01-25
PIM-5481 - Products CSV file analyzers

BitOne:
A very simple Product CSV analyzer tools that allows to get information without importing the file. The output looks like that:
```
Analyzing Product CSV /tmp/product.csv...
Columns Count:         548
Products Count:     50,000
Values Count:    4,611,007
Values per product:
  Average:    92
  Min:        66 (line 1534)
  Max:       135 (line 2051)
```
| Q                 | A
| ----------------- | ---
| Added Specs       | Y
| Added Behats      | N
| Travis CI is ok   |  Soon
| Changelog updated | Y

### DIFF
--- a/CHANGELOG-1.5.md
+++ b/CHANGELOG-1.5.md
@@ -15,6 +15,7 @@
 - A new twig extension (StyleExtension) in UIBundle now provides a "highlight" string filter
 - PIM-5450: MongoDb ODM bundle in dev requirements 
 - PIM-5380: It is now possible to group grid actions in dropdown by specifying it in the grid configuration (see quick export in the product grid for an example)
+- PIM-5481: New command to analyze products CSV files to get stats
 
 ## Bug fixes
 
--- a/src/Pim/Bundle/ImportExportBundle/Command/AnalyzeProductCsvCommand.php
+++ b/src/Pim/Bundle/ImportExportBundle/Command/AnalyzeProductCsvCommand.php
@@ -1,0 +1,97 @@
+<?php
+
+namespace Pim\Bundle\ImportExportBundle\Command;
+
+use Symfony\Bundle\FrameworkBundle\Command\ContainerAwareCommand;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Filesystem\Filesystem;
+
+/**
+ * Provides some statistics on a product CSV file
+ *
+ * @author    Benoit Jacquemont <benoit@akeneo.com>
+ * @copyright 2016 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+class AnalyzeProductCsvCommand extends ContainerAwareCommand
+{
+    /**
+     * {@inheritdoc}
+     */
+    protected function configure()
+    {
+        $this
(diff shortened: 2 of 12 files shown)


CHANGELOG-1.5.md
```diff
@@ -13,6 +13,7 @@
 - A new twig extension (StyleExtension) in UIBundle now provides a "highlight" string filter
 - PIM-5450: MongoDb ODM bundle in dev requirements 
 - PIM-5380: It is now possible to group grid actions in dropdown by specifying it in the grid configuration (see quick export in the product grid for an example)
+- PIM-5481: New command to analyze products CSV file to get stats
```
rybus: products CSV file**s**

src/Pim/Bundle/ImportExportBundle/Command/AnalyzeProductCsvCommand.php
```diff
@@ -0,0 +1,97 @@
+<?php
+
+namespace Pim\Bundle\ImportExportBundle\Command;
```
jjanvier: shouldn't we put this in the `ConnectorBundle`? 
it's quite usefull for import/export even without any UI

src/Pim/Bundle/ImportExportBundle/Command/AnalyzeProductCsvCommand.php
```diff
@@ -0,0 +1,97 @@
...
+ */
+class AnalyzeProductCsvCommand extends ContainerAwareCommand
+{
+    /**
+     * {@inheritdoc}
+     */
+    protected function configure()
+    {
+        $this
+            ->setName('pim:csv-analyzers:products')
```
jjanvier: `pim:connector:analyzer:csv-products` ?

if you put it in the connector bundle
and the fact it's a CSV is not related to the analyse IMO, we could analyse YAML the same way
that's why I suggest `analyzer:csv-products` instead of `csv-analyzers:products`

src/Pim/Component/Connector/CsvAnalyzer/ProductCsvAnalyzer.php
```diff
@@ -0,0 +1,139 @@
...
+
+    /**
+     * Display statistics about a product CSV file
+     *
+     * @param string $productCsvFile
+     * @param string $delimiter
+     *
+     * @return array
+     */
+    public function analyzeCsv($productCsvFile, $delimiter = null)
```
jjanvier: I would simply name it `analyse` so that later, if we need, we'll be able to extract a generic interface `FileAnalyser` from this class, and we'll be able to analyse YML or CSV the same way

jjanvier:
GTM once fixed

src/Pim/Bundle/ImportExportBundle/Command/AnalyzeProductCsvCommand.php
nidup: yes, agree

src/Pim/Component/Connector/CsvAnalyzer/ProductCsvAnalyzer.php
```diff
@@ -0,0 +1,139 @@
...
+     *
+     * @param array $values
+     *
+     * @return int
+     */
+    protected function countValues(array $values)
+    {
+        $valuesCount = 0;
+
+        foreach ($values as $value) {
```
nidup: what about:
 - price value in several columns, cost-EUR, cost-USD
 - metric in several columns, weight + weight-UNIT
?

nidup:
Baby disclaimer (maybe in the command description) about the fact it's an indication not the exact expected amount of value then GTM

PR #3939, opened 2016-01-27
[TR] PIM-5476: fix issue with native csv product import and product creation

nidup:
We introduced a regression with fix #3412 in 1.4.

The case was to allow to import products from a csv file with only the sku and the familly columns.

This applied fix in 1.4.4 removed the appliance of data filtering in product processor for new products and create too much values.

I have firstly submitted a PR with a more complex approach which prepares the removal of any empty values in the DB, but we'll do this only in a upcoming minor version, not in a patch.

With small catalog, before the fix:
```
mysql> select count(*) from pim_catalog_product_value;
+----------+
|   342029 |
+----------+

Import small_cat has been successfully executed.
real    21m1.744s
user    17m53.809s
sys 0m30.667s
```

With small catalog, after the fix:
```
+----------+
| count(*) |
+----------+
|   159676 |
+----------+

Import small_cat has been successfully executed.
real	10m43.969s
user	9m23.369s
sys	0m16.664s
```

| Q                 | A
| ----------------- | ---
| Specs             | Y
| Behats            | Y
| Changelog updated | Y
| Review and 2 GTM  |

### DIFF
--- a/src/Pim/Component/Connector/Processor/Denormalization/ProductProcessor.php
+++ b/src/Pim/Component/Connector/Processor/Denormalization/ProductProcessor.php
@@ -101,13 +101,11 @@ class ProductProcessor extends AbstractProcessor
 
         $familyCode    = $this->getFamilyCode($convertedItem);
         $filteredItem  = $this->filterItemData($convertedItem);
-
         $product = $this->findOrCreateProduct($identifier, $familyCode);
-
-        if ($this->enabledComparison && null !== $product->getId()) {
+        if ($this->enabledComparison) {
             $filteredItem = $this->filterIdenticalData($product, $filteredItem);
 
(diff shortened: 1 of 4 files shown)


jjanvier:
gtm

src/Pim/Component/Connector/Processor/Denormalization/ProductProcessor.php
```diff
@@ -103,8 +103,10 @@ public function process($item)
         $filteredItem  = $this->filterItemData($convertedItem);
 
         $product = $this->findOrCreateProduct($identifier, $familyCode);
+        $newProdWithOnlyFamily =  (null === $product->getId()) && (2 === count($filteredItem))
```
grena: extra space before first parenthesis

PR #3969, opened 2016-02-05
Fix behats

phaseinducer:
Fix for :
features/category/edit_a_category.feature:10
features/product/toggle_status_product.feature:12

### DIFF
--- a/features/category/edit_a_category.feature
+++ b/features/category/edit_a_category.feature
@@ -1,3 +1,4 @@
+@javascript
 Feature: Edit a category
   In order to be able to modify the category tree
   As a product manager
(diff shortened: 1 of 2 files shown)


jmleroux:
GTM

features/category/edit_a_category.feature
```diff
@@ -1,3 +1,4 @@
+@javascript
```
skeleton: why put this tag here ? Julien spent some time to remove it and add it to each necessary scenario

skeleton:
1 question, then GTM

features/category/edit_a_category.feature
phaseinducer: IRL

PR #3972, opened 2016-02-08
PIM-5318: Translate missing labels in all screens

skeleton:
| Q                 | A
| ----------------- | ---
| Specs             | Y
| Behats            | Y
| Blue CI           | ?
| Changelog updated | Y
| Review and 2 GTM  | 
| Micro Demo (PO)   | Y
| Migration script  | N
| Tech Doc          | N

- missing keys added
- fix locale fallback in config
- factorize initialization of select2 & datetimepicker
- show datetime instead of date in grids

### DIFF
--- a/src/Pim/Bundle/DataGridBundle/Resources/public/js/datafilter/filter/date-filter.js
+++ b/src/Pim/Bundle/DataGridBundle/Resources/public/js/datafilter/filter/date-filter.js
@@ -1,7 +1,7 @@
 /* global define */
 define(
-    ['jquery', 'underscore', 'oro/translator', 'oro/datafilter/choice-filter', 'pim/date-context', 'bootstrap.datetimepicker'],
-function($, _, __, ChoiceFilter, DateContext) {
+    ['jquery', 'underscore', 'oro/translator', 'oro/datafilter/choice-filter', 'datepicker'],
+function($, _, __, ChoiceFilter, Datepicker) {
     'use strict';
 
     /**
@@ -86,20 +86,7 @@ function($, _, __, ChoiceFilter, DateContext) {
          *
          * @property
          */
-        datetimepickerOptions: {
-            format: DateContext.get('date').format,
-            defaultFormat: DateContext.get('date').defaultFormat,
-            locale: DateContext.get('language'),
-            pickTime: false
-        },
-
-        /**
-         * Additional date widget options that might be passed to filter
-         * http://api.jqueryui.com/datepicker/
-         *
-         * @property
-         */
-        externalWidgetOptions: {},
+        datetimepickerOptions: {},
 
         /**
          * References to date widgets
@@ -134,7 +121,6 @@ function($, _, __, ChoiceFilter, DateContext) {
          * @inheritDoc
          */
         initialize: function () {
-            _.extend(this.datetimepickerOptions, this.externalWidgetOptions);
             // init empty value object if it was not initialized so far
             if (_.isUndefined(this.emptyValue)) {
                 this.emptyValue = {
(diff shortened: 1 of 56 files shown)


src/Pim/Bundle/DataGridBundle/Resources/public/js/datafilter/filter/date-filter.js
```diff
@@ -201,20 +187,18 @@ function($, _, __, ChoiceFilter, DateContext) {
...
-            var picker = widget.data('datetimepicker');
-            widget.on('changeDate', function(e) {
-                picker.format = this.datetimepickerOptions.defaultFormat;
+            var picker = $widget.data('datetimepicker');
+            $widget.on('changeDate', function(e) {
+                picker.format = Datepicker.options.defaultFormat;
                 this.values.value[e.target.className].defaultFormat = picker.formatDate(e.date);
 
-                picker.format = this.datetimepickerOptions.format;
+                picker.format = Datepicker.options.format;
```
filipsalpe: Don't do this, kittens will die. Modifying a data attribute by reference twice (assuming it even changes something) basically equals begging for bugs or not being able to understand this code in 2 months.

src/Pim/Bundle/UIBundle/Resources/public/js/pim-initselect2.js
```diff
@@ -5,7 +5,21 @@ define(
         'use strict';
         return {
             resultsPerPage: 20,
-            init: function ($target) {
+            defaultOptions: {},
+            init: function ($target, options) {
+                this.defaultOptions = {
+                    allowClear: false,
+                    formatSearching: function () {
```
filipsalpe: Since this message is not dynamic, you might as well remove the function and just translate the message key once instead of doing it every time select2 wants to display it. Same below.

src/Pim/Bundle/LocalizationBundle/Controller/FormatController.php
```diff
@@ -20,7 +20,7 @@ class FormatController
     protected $dateFactory;
 
     /** @var DateFactory */
-    protected $datetimeFactoty;
+    protected $datetimeFactory;
```
solivier: :+1:

solivier:
GTM for me but I'm really not sure about JS files

src/Pim/Bundle/EnrichBundle/Resources/public/js/product/field/metric-field.js
```diff
@@ -13,8 +13,9 @@ define([
         'underscore',
         'pim/fetcher-registry',
         'text!pim/template/product/field/metric',
+        'pim/initselect2',
         'jquery.select2'
```
SamirBoulil: Should we remove this one now that we load pim/initselect2 ?

src/Pim/Bundle/EnrichBundle/Resources/public/js/product/field/simple-select-field.js
```diff
@@ -16,9 +16,10 @@ define(
         'routing',
         'pim/attribute-option/create',
         'pim/security-context',
+        'pim/initselect2',
         'jquery.select2'
```
SamirBoulil: Same here ? :)

src/Pim/Bundle/EnrichBundle/Resources/public/js/product/form/meta/change-family.js
```diff
@@ -18,10 +18,22 @@ define(
         'pim/user-context',
         'pim/i18n',
         'routing',
+        'pim/initselect2',
         'backbone/bootstrap-modal',
```
SamirBoulil: Remove unused 'backbone/bootstrap-modal' and 'jquery.select2' dependency ?

src/Pim/Bundle/DataGridBundle/Resources/public/js/datafilter/filter/date-filter.js
jmleroux: @filipsalpe , do you have a more precise recommendation ?

src/Pim/Bundle/UIBundle/Resources/public/js/pim-initselect2.js
jmleroux: Hey @filipsalpe ,
According to the code (because i didn't find the docuementation :| ) of our ô so old select2 version, it seems mandatory to use a callback.
https://github.com/select2/select2/blob/3.4.1/select2.js#L1559

Would we had a more up to date, we could use a string, like we can see here : https://github.com/select2/select2/blob/3.5.1/select2.js#L1744

filipsalpe: Ah, indeed!

src/Pim/Bundle/DataGridBundle/Resources/public/js/datafilter/filter/date-filter.js
filipsalpe: No, I have no idea what this actually does. The diff in this commit is good but the actual logic is something to be refactored.

jmlrx:
GTM

PR #3996, opened 2016-02-16
Merge 1.4 into master

phaseinducer:


src/Akeneo/Bundle/StorageUtilsBundle/Doctrine/Common/Detacher/ObjectDetacher.php
```diff
@@ -20,12 +25,16 @@ class ObjectDetacher implements ObjectDetacherInterface, BulkObjectDetacherInter
     /** @var ManagerRegistry */
     protected $managerRegistry;
 
+    /** @var array */
+    protected $scheduledForDirtyCheck;
```
skeleton: break static analysis (more than 20 characters) :(

src/Akeneo/Bundle/StorageUtilsBundle/Doctrine/Common/Detacher/ObjectDetacher.php
```diff
@@ -48,6 +64,93 @@ public function detachAll(array $objects)
     }
 
     /**
+     * Detach entity living in the scheduledForDirtyCheck's
+     * unit of work property.
+     *
+     * @param mixed $entity  The entity to be detached
+     * @param array $visited Array of already detached entities
+     */
+    protected function doDetachScheduled($entity, array &$visited)
```
skeleton: should not be in a dedicated bundle ?

src/Akeneo/Component/Analytics/ChainedDataCollectorInterface.php
```diff
@@ -8,6 +8,8 @@
  * @author    Nicolas Dupont <nicolas@akeneo.com>
  * @copyright 2015 Akeneo SAS (http://www.akeneo.com)
  * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ *
+ * @deprecated Will be removed in 1.5, please use directly Akeneo\Component\Analytics\ChainedDataCollector
```
skeleton: to remove ?

phaseinducer: Will be done in another PR

src/Akeneo/Bundle/StorageUtilsBundle/Doctrine/Common/Detacher/ObjectDetacher.php
phaseinducer: I don't understand. Ping @SamirBoulil ?

SamirBoulil: @skeleton I agree it is not an optimum solution.

Instead of creating a bundle to listen for doctrine's event and *reset* the problematic array. We chose to use the same technique used in mongodb (see below in the file) and manually cascade detach objects in the array in the ObjectDetacher class.

To put the fix here allows us to precisely remove objects to be detached instead of resetting the whole array like we once did in the separate bundle.

src/Akeneo/Bundle/StorageUtilsBundle/Doctrine/Common/Detacher/ObjectDetacher.php
phaseinducer: To fix in another PR

pierallard:
GTM

src/Akeneo/Bundle/StorageUtilsBundle/Doctrine/Common/Detacher/ObjectDetacher.php
jjanvier: as already commented in the first PR, we should not do as is in master imvho
we should put this stuff in a dedicated "FixDoctrineBugDetacher" to isolate the problem

skeleton: agree with jj
you can merge like that, but we should create a card to create a dedicated bundle

skeleton:
gtm

nidup:
GTM, as discussed previously some re-work to do, imho, we should first merge this one, then re-work

PR #4006, opened 2016-02-18
PIM-5579: Add the possibility to disable the export of products medias if needed

phaseinducer:
| Q                 | A
| ----------------- | ---
| Specs             | n/a
| Behats            | n/a
| Blue CI           | 
| Changelog updated | yes
| Review and 2 GTM  |

PR for doc : https://github.com/akeneo/pim-docs/pull/237

### DIFF
--- a/CHANGELOG-1.4.md
+++ b/CHANGELOG-1.4.md
@@ -1,3 +1,8 @@
+# 1.4.x
+
+## Scalability improvements
+- PIM-5579: Add the possibility to disable the export of products media if needed
+
 # 1.4.19 (2016-02-11)
 
 ## Bug fixes
--- a/src/Pim/Bundle/BaseConnectorBundle/Writer/File/CsvProductWriter.php
+++ b/src/Pim/Bundle/BaseConnectorBundle/Writer/File/CsvProductWriter.php
@@ -19,6 +19,9 @@ class CsvProductWriter extends ContextableCsvWriter
     /** @var FileExporterInterface */
     protected $fileExporter;
 
+    /** @var bool */
+    protected $copyMedia;
+
     /** @var string */
     protected $bufferFile;
 
@@ -27,12 +30,14 @@ class CsvProductWriter extends ContextableCsvWriter
 
     /**
      * @param FileExporterInterface $fileExporter
+     * @param bool                  $copyMedia
      */
-    public function __construct(FileExporterInterface $fileExporter)
+    public function __construct(FileExporterInterface $fileExporter, $copyMedia = true)
     {
         parent::__construct();
 
         $this->fileExporter = $fileExporter;
+        $this->copyMedia    = $copyMedia;
     }
 
     /**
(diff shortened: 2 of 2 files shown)


src/Pim/Bundle/BaseConnectorBundle/Writer/File/CsvProductWriter.php
```diff
@@ -19,6 +19,9 @@ class CsvProductWriter extends ContextableCsvWriter
     /** @var FileExporterInterface */
     protected $fileExporter;
 
+    /** @var bool */
+    protected $copyMedias;
```
solivier: copyMedia*

CHANGELOG-1.4.md
```diff
@@ -1,3 +1,8 @@
+# 1.4.x
+
+## Scalability improvements
+- PIM-5579: Add the possibility to disable the export of products medias if needed
```
solivier: media

solivier:
gtm

pierallard:
GTM

jjanvier:
This fix really emphases the architecture problem around the exports and the archivers. It's a terrible hack... Basically you have a class that implements an archiver behavior and you bypass it with a boolean argument :(. I really don't like but I don't know how you could do export in a patch :(

GTM, with sadness

src/Pim/Bundle/BaseConnectorBundle/Writer/File/CsvProductWriter.php
```diff
@@ -27,12 +30,14 @@ class CsvProductWriter extends ContextableCsvWriter
 
     /**
      * @param FileExporterInterface $fileExporter
+     * @param bool                  $copyMedia
      */
-    public function __construct(FileExporterInterface $fileExporter)
+    public function __construct(FileExporterInterface $fileExporter, $copyMedia = true)
```
skeleton: why this param is not added in DI ?

phaseinducer:
Agree. We can do much better for this whole archiving stuff, in 1.6 I hope.

src/Pim/Bundle/BaseConnectorBundle/Writer/File/CsvProductWriter.php
phaseinducer: Is it useful since it has a default value ?

ClementGautier: Boolean flags are, by definition, bad solutions: http://phpmd.org/rules/cleancode.html#BooleanArgumentFlag.
I would suggest to create an other writer (CsvProductWithoutMediaWriter or something like this).

By the way, could we do that at a lower level ? I mean before the writer is called ? At processor level or reader level ? This way we could even not hydrate Product medias probably.

ClementGautier:
missing specs

BitOne:
Why disabling the product media export while the problem lies in the archiver, not in the export ?
I would do a parameter to disable the archiver (because this is the problem), not the product export.
If you have people that have the problem with the archiver but uses the standard media export to handle their media (which is the normal cases), this will not fix their problem: it will either break their export or not this the archiver...

BitOne:
Not GTM

phaseinducer:
Replaced by https://github.com/akeneo/pim-community-dev/pull/4028

PR #4026, opened 2016-02-22
Add validation in family controller

fitn:
Validate family when adding or removing attributes.

It is just a first step. The second one needs to move the hardcoded validation from the controller to dedicated constraint.

| Q                 | A
| ----------------- | ---
| Added Specs       | N/A
| Added Behats      | N
| Travis CI is ok   |
| Changelog updated | Y

### DIFF
--- a/src/Pim/Bundle/EnrichBundle/Controller/FamilyController.php
+++ b/src/Pim/Bundle/EnrichBundle/Controller/FamilyController.php
@@ -26,6 +26,7 @@ use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
 use Symfony\Component\Routing\RouterInterface;
 use Symfony\Component\Translation\TranslatorInterface;
+use Symfony\Component\Validator\Validator\ValidatorInterface;
 
 /**
  * Family controller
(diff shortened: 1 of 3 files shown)


jjanvier:
gtm

solivier:
gtm

src/Pim/Bundle/EnrichBundle/Controller/FamilyController.php
```diff
@@ -293,9 +299,15 @@ public function addAttributesAction($id)
...
-
-        $this->request->getSession()->getFlashBag()->add('success', new Message('flash.family.attributes added'));
+        $errors = $this->validator->validate($family);
+        if (count($errors) > 0) {
+            foreach ($errors as $error) {
+                $this->request->getSession()->getFlashBag()->add('error', new Message($error->getMessage()));
+            }
+        } else {
+            $this->familySaver->save($family);
+            $this->request->getSession()->getFlashBag()->add('success', new Message('flash.family.attributes added'));
```
aRn0D: You should inject the `Request` into your method instead of inject it into your constructor, you change its signature so I guest you can manage that.

fitn: I agree but that's not the purpose of my PR ^^

solivier: I don't see why you can't do it in this PR ;)

PR #4027, opened 2016-02-22
Preparing 1.5.0-BETA1

pierallard:


jjanvier:
gtm

src/Pim/Bundle/TransformBundle/Transformer/ProductTransformer.php
```diff
@@ -1,357 +0,0 @@
...
-/**
- * Specialized transformer for products
- *
- * @author    Antoine Guigan <antoine@akeneo.com>
- * @copyright 2013 Akeneo SAS (http://www.akeneo.com)
- * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
- *
- * @deprecated will be removed in 1.6
- */
-class ProductTransformer extends EntityTransformer
```
nidup: Means we drop the old import products system

src/Pim/Bundle/TransformBundle/Resources/config/transformers.yml
```diff
@@ -37,21 +36,6 @@ services:
             - '@pim_transform.transformer.guesser'
             - '@pim_transform.transformer.column_info_transformer'
 
-    # Deprecated: This service is deprecated and will be removed in 1.5
-    pim_transform.transformer.product:
```
nidup: maybe used by pim_base_connector.processor.product via the 'pim_transform.transformer.registry'

CHANGELOG-1.5.md
```diff
@@ -1,12 +1,20 @@
-# 1.5.x
```
grena: Don't we keep the 1.5.x for later?

nidup:
GTM except the remove of the product transformer, if you run the "deprecated build" on this branch you'll see all old import scenario fail, we can assume to drop it but if we do so, we need to drop the related product processor and all related deprecated scenario + add a notice in the changelog.
Other strategy, don't drop this "deprecated product transformer" for now, merge this branch, continue to re-work via another branch to remove as much deprecated as possible

CHANGELOG-1.5.md
pierallard: @grena The 1.5.x is used to automatically change it with tag script. Now we tag the official 1.5, we don't want to tag it with an "1.5.x" in header without any thing in it.

grena: :+1:

grena:
GTM to me, but for this kind of PR, maybe another one is more secure :p

PR #4028, opened 2016-02-22
PIM-5579: Enable FileWriterArchiver even for exports with several files

phaseinducer:
This PR replaces https://github.com/akeneo/pim-community-dev/pull/4006.

| Q                 | A
| ----------------- | ---
| Specs             | yes
| Behats            | yes
| Blue CI           | running
| Changelog updated | yes
| Review and 2 GTM  | yes

Currently there are two native archivers for exports. Their role is to copy exported files (csv, media, etc.) from the export location to the archive directory (usually under app/archive/export/...). These files are used especially to be downloaded from the job page (button "Download generated file").

**How it worked before**
- Only one file exported (usually just the csv and no media) : FileWriterArchiver is used, csv is archived
- Several files exported (csv + media) : ArchivableFileWriterArchiver is used, all files are archived in a zip

**How it works now**
- Only one file exported : same as before
- Several files exported : both archivers are used, the zip is created and the csv is archived too separately.

The consequence is that now you have two download buttons, one for each archiver :
export_archives

**Why ?**
The goal is that you can easily disable ArchivableFileWriterArchiver if you experience problems with media archiving. Also :
- the media export is untouched (not the case with the previous PR), only the archiving is disabled
- the csv file can still be downloaded from the job page without any additional tweaking

(Benoît & Delphine approved)

### DIFF
--- a/spec/Pim/Bundle/BaseConnectorBundle/Archiver/FileWriterArchiverSpec.php
+++ b/spec/Pim/Bundle/BaseConnectorBundle/Archiver/FileWriterArchiverSpec.php
@@ -15,14 +15,7 @@ use Prophecy\Argument;
 
 class FileWriterArchiverSpec extends ObjectBehavior
 {
-    function let(
-        Filesystem $filesystem,
-        CsvWriter $writer,
-        JobExecution $jobExecution,
-        JobInstance $jobInstance,
-        Job $job,
-        ItemStep $step
-    ) {
+    function let(Filesystem $filesystem) {
         $this->beConstructedWith($filesystem);
     }
 
@@ -31,8 +24,14 @@ class FileWriterArchiverSpec extends ObjectBehavior
         $this->shouldHaveType('Pim\Bundle\BaseConnectorBundle\Archiver\FileWriterArchiver');
     }
 
-    function it_creates_a_file_when_writer_is_valid($filesystem, $writer, $jobExecution, $jobInstance, $job, $step)
-    {
+    function it_creates_a_file_when_writer_is_valid(
+        $filesystem,
+        CsvWriter $writer,
+        JobExecution $jobExecution,
+        JobInstance $jobInstance,
+        Job $job,
+        ItemStep $step
+    ) {
         $pathname = tempnam(sys_get_temp_dir(), 'spec');
         $filename = basename($pathname);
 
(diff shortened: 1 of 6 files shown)


spec/Pim/Bundle/BaseConnectorBundle/Archiver/FileWriterArchiverSpec.php
```diff
@@ -31,8 +24,14 @@ function it_is_initializable()
         $this->shouldHaveType('Pim\Bundle\BaseConnectorBundle\Archiver\FileWriterArchiver');
     }
 
-    function it_creates_a_file_when_writer_is_valid($filesystem, $writer, $jobExecution, $jobInstance, $job, $step)
-    {
+    function it_creates_a_file_when_writer_is_valid(
+        $filesystem,
+        CsvWriter $writer,
```
aRn0D: This should be reverted, it already done in the method let.

phaseinducer: No, I removed it.

src/Pim/Bundle/ImportExportBundle/Resources/translations/messages.en.yml
```diff
@@ -18,7 +18,7 @@ pim_import_export:
         8: UNKNOWN
 
     download_archive:
-        archive: download generated file|download generated files
+        archive: download generated archive|download generated archives
```
BitOne: Download full generated archive ?

jjanvier:
GTM if behat added to check the presence of the 2 buttons

spec/Pim/Bundle/BaseConnectorBundle/Archiver/FileWriterArchiverSpec.php
aRn0D: my bad...

src/Pim/Bundle/ImportExportBundle/Resources/translations/messages.en.yml
phaseinducer: I understand what you mean but it sounds weird :/

nidup:
GTM

grena:
GTM

PR #4037, opened 2016-02-23
[DX] always use savers in behat context

nidup:
This PR allows to always use savers in our Behat context to continue to remove all direct use of persist / flush in the stack.

This PR also add missing savers definition and declare the base saver class in the storage utils to avoid to use the base definition of the catalog everywhere.

| ----------------------------- | ---
| Added Specs              | N
| Added Behats             | Y (only changes in contexts)
| Changelog updated    | N
| Review and 2 GTM     |


jmleroux:
GTM
Nice !

jjanvier:
Great,! GTM!

nidup:
thank you! next step -> do the same with removers :)

PR #4040, opened 2016-02-24
PIM-5589: Remove association_types YML import

pierallard:

Remove all the old import system for association_types. Now we can only call CSV import.

```
```


src/Pim/Bundle/BaseConnectorBundle/Resources/config/processors.yml
```diff
@@ -92,11 +92,6 @@ services:
             - %pim_catalog.entity.category.class%
             - '@pim_transform.cache.doctrine_cache'
 
-    pim_base_connector.processor.association_type:
```
nidup: i would keep it for now but mark it as deprecated to avoid to break "deprecated builds", once we have all new import we'll drop all of them

nidup:
fix comment about processor, add a line in changelog to explain yml is not anymore supported for association type then GTM

solivier:
gtm

PR #4055, opened 2016-02-25
Fix static analysis

phaseinducer:
Fix 21 PMD warnings (reduce from 170 to 149, threshold is 160)

### DIFF
--- a/src/Pim/Component/Catalog/Comparator/Attribute/FileComparator.php
+++ b/src/Pim/Component/Catalog/Comparator/Attribute/FileComparator.php
@@ -46,35 +46,21 @@ class FileComparator implements ComparatorInterface
         $default   = ['locale' => null, 'scope' => null, 'data' => ['filePath' => null]];
         $originals = array_merge($default, $originals);
 
+        if (!isset($data['data']['filePath']) && !isset($originals['data']['filePath']) ||
+            $this->filesMatch($data, $originals)
+        ) {
+            return null;
+        }
+
         // compare a local file and a stored file (can happen during an import for instance)
-        if (isset($data['data']['filePath']) &&
-            isset($originals['data']['filePath']) &&
+        if (isset($originals['data']['filePath']) &&
+            isset($data['data']['filePath']) &&
             is_file($data['data']['filePath'])
         ) {
             $originalFile = $this->repository->findOneByIdentifier($originals['data']['filePath']);
             if (null !== $originalFile && $originalFile->getHash() === $this->getHashFile($data['data']['filePath'])) {
                 return null;
             }
-        }
-
-        // compare the two stored files by their filepaths
-        if (isset($data['data']['filePath']) &&
-            isset($originals['data']['filePath']) &&
-            $data['data']['filePath'] === $originals['data']['filePath']
-        ) {
-            return null;
-        }
-
-        // compare the two stored files by their hashes
-        if (isset($data['data']['hash']) &&
-            isset($originals['data']['hash']) &&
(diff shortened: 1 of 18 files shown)


src/Pim/Component/Catalog/Comparator/Attribute/FileComparator.php
```diff
@@ -89,4 +75,24 @@ protected function getHashFile($filePath = null)
...
+     *
+     * @return bool
+     */
+    protected function filesMatch($data, $originals)
+    {
+        return
+            isset($data['data']['filePath']) &&
+            isset($originals['data']['filePath']) &&
+            $data['data']['filePath'] === $originals['data']['filePath']
+            ||
```
grena: Is doesn't need parenthesis for the 2 "blocks" because of the `||` ?
```
(..&&..&&...) || (..&&..&&...)
```

grena:
Cool :+1: ! Just a question then GTM.

src/Pim/Component/Catalog/Comparator/Attribute/FileComparator.php
phaseinducer: In php `&&` has a higher priority than `||` (http://php.net/manual/en/language.operators.precedence.php) so no. But it's possible to add the parenthesis for readability if needed.

src/Pim/Component/Catalog/Updater/AssociationTypeUpdater.php
```diff
@@ -22,15 +22,15 @@ class AssociationTypeUpdater implements ObjectUpdaterInterface
     protected $accessor;
 
     /** @var IdentifiableObjectRepositoryInterface */
-    protected $associationTypeRepository;
+    protected $assocTypeRepository;
```
jmleroux: I would use `$associationTypeRepo`

src/Pim/Component/Connector/ArrayConverter/Flat/Product/AssociationColumnsResolver.php
```diff
@@ -20,18 +20,18 @@ class AssociationColumnsResolver
     const PRODUCT_ASSOCIATION_SUFFIX = '-products';
 
     /** @var AssociationTypeRepositoryInterface */
-    protected $associationTypeRepo;
+    protected $assocTypeRepository;
```
jmleroux: same

src/Pim/Component/Catalog/Comparator/Attribute/FileComparator.php
jjanvier: yes please add them to avoid confusion
good catch pipou

jmleroux:
GTM, even with a small comment

jjanvier:
GTM
but this limit of the lenght of a var is imvho counter effective and breaks readability, which is the most important for the code...

src/Pim/Component/Catalog/Updater/AssociationTypeUpdater.php
phaseinducer: Me too but consistency is more important

phaseinducer:
I agree but we already discussed about that with @BitOne.
And if we change that it should be decided collectively.

src/Pim/Component/Catalog/Comparator/Attribute/FileComparator.php
grena: @phaseinducer Ok, I wasn't sure :) Maybe adding parenthesis could be nice for readability then, thanks!

PR #4067, opened 2016-02-29
POC for XLSX imports/exports

skeleton:


nidup:
@skeleton don't hesitate to drop this PR if we already re-integrated all interesting parts as new features

PR #4096, opened 2016-03-04
Split variant group add products scenario

nidup:
Split the variant group scenario to multiple feature files to avoid issues with stateful versioning listener which is not reproducible with a classic use of the PIM (aka passing by the UI without updating x time the same product in the same process).

This versioning listener based on doctrine event is a real pack of :wolf: we plan to re-write the versioning using our storage events (first we need to always use saver/remover api, we're getting closer)


solivier:
gtm

pierallard:
Are you sure it works even if the features runs on the same worker?

nidup:
Would say yes because it's an issue with the stateful versioning in the behat context even with the same worker different processes will run the behat.

pierallard:
So... Gtm, but you run the CI first :troll:

nidup:
It passed several time for real ;)

PR #4130, opened 2016-03-10
Add a registerProjectBundles method in the AppKernel

fitn:
Here is a way to ease the registration of the project bundles. A dedicated method has been extracted and put at the top of the AppKernel.

### DIFF
--- a/app/AppKernel.php
+++ b/app/AppKernel.php
@@ -17,7 +17,7 @@ class AppKernel extends Kernel
      */
     protected function registerProjectBundles()
     {
-        $bundles = [
+        return [
             // your app bundles should be registered here
             new Acme\Bundle\AppBundle\AcmeAppBundle(),
         ];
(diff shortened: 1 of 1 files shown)


app/AppKernel.php
```diff
@@ -13,14 +13,22 @@
 class AppKernel extends Kernel
 {
     /**
-     * {@inheritdoc}
+     * Registers your custom bundles
      */
-    public function registerBundles()
+    protected function registerProjectBundles()
```
phaseinducer: public methods first ; )

phaseinducer:
Godd idea, GTM

app/AppKernel.php
```diff
@@ -13,14 +13,24 @@
 class AppKernel extends Kernel
 {
     /**
-     * {@inheritdoc}
+     * Registers your custom bundles
+     * 
+     * @return array
      */
-    public function registerBundles()
+    protected function registerProjectBundles()
```
phaseinducer: Public methods first ; )

skeleton:
gtm

nidup:
Thank you, great idea. Public method first, please squash your commit then GTM.

PR #4140, opened 2016-03-11
Detach entity removal

damien-carcel:
**Description**

When having a lot of products in a MongoDB installation, one cannot remove a product related entity (attribute, family, association, group…) without freezing the UI, as it takes some time to check all products to remove all corresponding data (for instance the product values if one delete an attribute) from all the products. In some case, it will even cause a timeout, causing the operation too fail and resulting in an error 500 for products that have not been cleaned.
The same problem happen when removing a product, as every other products will be checked to remove a possible association.

The operation responsible of these UI freeze are done in 3 event subscribers:
- ProductRelatedEntityRemovalSubscriber,
- RemoveOutdatedProductsFromAssociationsSubscriber,
- UpdateNormalizedProductDataSubscriber.
  The goal of this PR is to prevent that by detaching the logic into symfony commands. The subscribers only launch the commands, and so do not freeze the UI anymore.

This behavior can also happen with small number of products but great number of channels. I already encountered a CE project with less than 5M product values, but with 8 channels, with 8 activated locale each at least! In this case, the Mongo queries are very slow because all the indexes are taken by the completeness, and deleting an attribute takes several minutes!
In this case, the "UpdateNormalizedProductDataCommand" is responsible, as the MongoDB request it performs are very slow due to the lack of indexes. A possibility (placed in the PR as commented code) is to control the Mongo time out with the "socketTimeoutMS" option.

**Definition Of Done**

| Q | A |
| --- | --- |
| Added Specs | Yes |
| Added Behats | No |
| Changelog updated | Yes |
| Review and 2 GTM | Waiting |
| Micro Demo to the PO (Story only) | No |
| Migration script | No |
|  Tech Doc | No |

### DIFF
--- a/src/Pim/Bundle/CatalogBundle/Command/ProductRelatedEntityRemovalCommand.php
+++ b/src/Pim/Bundle/CatalogBundle/Command/ProductRelatedEntityRemovalCommand.php
@@ -1,0 +1,91 @@
+<?php
+
+namespace Pim\Bundle\CatalogBundle\Command;
+
+use Pim\Bundle\CatalogBundle\Doctrine\MongoDBODM\ProductRepositoryInterface;
+use Symfony\Bundle\FrameworkBundle\Command\ContainerAwareCommand;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+
+/**
+ * Removes the values of a specific entity from all products.
+ *
+ * @author    Damien Carcel <damien.carcel@akeneo.com>
+ * @copyright 2015 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+class ProductRelatedEntityRemovalCommand extends ContainerAwareCommand
+{
+    /**
+     * {@inheritdoc}
+     */
+    protected function configure()
+    {
+        $this
+            ->setName('pim:product:remove-related-entity')
+            ->setDescription('Removes the values of a specific entity from all products')
+            ->addArgument('entityName', InputArgument::REQUIRED, 'The entity name')
+            ->addArgument(
+                'ids',
+                InputArgument::REQUIRED,
+                'The list of entity IDs to remove from the products, comma separated'
+            )
+        ;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+
+        $productRepository = $this->getContainer()->get('pim_catalog.repository.product');
+        $entityName        = $input->getArgument('entityName');
+        $ids               = explode(',', $input->getArgument('ids'));
+
+        foreach ($ids as $id) {
+            $this->remove($productRepository, $entityName, (int) $id);
+        }
+    }
+
+    /**
+     * Removes an entity value linked to products and returns the IDs the
+     * updated products.
+     *
+     * @param ProductRepositoryInterface $productRepository
+     * @param string                     $entityName
+     * @param int                        $id
+     */
+    protected function remove(ProductRepositoryInterface $productRepository, $entityName, $id)
+    {
+        switch ($entityName) {
+            case 'AssociationType':
+                $productRepository->cascadeAssociationTypeRemoval($id);
+                break;
+            case 'Attribute':
+                $productRepository->cascadeAttributeRemoval($id);
+                break;
+            case 'AttributeOption':
+                $productRepository->cascadeAttributeOptionRemoval($id);
+                break;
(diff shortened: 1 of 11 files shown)


src/Pim/Bundle/CatalogBundle/EventSubscriber/MongoDBODM/UpdateNormalizedProductDataSubscriber.php
```diff
@@ -27,27 +37,27 @@ class UpdateNormalizedProductDataSubscriber implements EventSubscriber
...
-     * @param ManagerRegistry $registry
+     * @param CommandLauncher $commandLauncher
      * @param string          $productClass
+     * @param string          $logFile
      */
-    public function __construct(ManagerRegistry $registry, $productClass)
-    {
-        $this->registry     = $registry;
-        $this->productClass = $productClass;
+    public function __construct(
```
solivier: add this in changelog

src/Pim/Bundle/CatalogBundle/EventSubscriber/MongoDBODM/ProductRelatedEntityRemovalSubscriber.php
```diff
@@ -16,38 +23,20 @@
...
+     * @param string          $logFile
      */
-    protected $pendingUpdates = [];
-
-    /**
-     * @param ManagerRegistry $registry
-     * @param string          $productClass
-     */
-    public function __construct(ManagerRegistry $registry, $productClass)
+    public function __construct(CommandLauncher $launcher, $logFile)
```
solivier: don't forget the changelog

src/Pim/Bundle/CatalogBundle/EventSubscriber/MongoDBODM/ProductRelatedEntityRemovalSubscriber.php
```diff
@@ -55,59 +44,117 @@ public function __construct(ManagerRegistry $registry, $productClass)
...
     }
 
     /**
-     * Execute pending batch updates
+     * @param object $entity
+     *
+     * @return null|string
      */
-    protected function executePendingUpdates()
+    protected function getEntityName($entity)
```
solivier: please avoid many returns in the same method (you can split it for example)

src/Pim/Bundle/CatalogBundle/Command/ProductRelatedEntityRemovalCommand.php
```diff
@@ -0,0 +1,91 @@
...
+use Symfony\Bundle\FrameworkBundle\Command\ContainerAwareCommand;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+
+/**
+ * Removes the values of a specific entity from all products.
+ *
+ * @author    Damien Carcel <damien.carcel@akeneo.com>
+ * @copyright 2015 Akeneo SAS (http://www.akeneo.com)
```
solivier: 2016

src/Pim/Bundle/CatalogBundle/Command/UpdateNormalizedProductDataCommand.php
```diff
@@ -0,0 +1,58 @@
...
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+
+/**
+ * Executes a list of queries to updates the normalized data of a
+ * product document when related entities are modified.
+ *
+ * @author    Damien Carcel <damien.carcel@akeneo.com>
+ * @copyright 2015 Akeneo SAS (http://www.akeneo.com)
```
solivier: 2016

src/Pim/Bundle/CatalogBundle/EventSubscriber/MongoDBODM/ProductRelatedEntityRemovalSubscriber.php
damien-carcel: Done

src/Pim/Bundle/CatalogBundle/EventSubscriber/MongoDBODM/ProductRelatedEntityRemovalSubscriber.php
damien-carcel: Done

src/Pim/Bundle/CatalogBundle/EventSubscriber/MongoDBODM/UpdateNormalizedProductDataSubscriber.php
damien-carcel: Done

PR #4163, opened 2016-03-17
Fix path of AttributeInterface

Hocdoc:
This reference is only used in a PHP Doc. Nonetheless the path was wrong.


nidup:
Thank you @Hocdoc !

PR #4180, opened 2016-03-18
Behat - Spin options count

jmleroux:
We need to spin the options count to avoid failures on CI.

| Added Behats                      |y
| Changelog updated                 |n

### DIFF
--- a/features/Context/WebUser.php
+++ b/features/Context/WebUser.php
@@ -498,11 +498,14 @@ class WebUser extends RawMinkContext
      */
     public function theOptionsSectionShouldContainOption($expectedCount = 1)
     {
-        if ($expectedCount != $count = $this->getCurrentPage()->countOptions()) {
-            throw $this->createExpectationException(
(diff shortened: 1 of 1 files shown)


features/Context/WebUser.php
```diff
@@ -498,11 +498,14 @@ public function iShouldSeeTheSection($title)
...
     {
-        if ($expectedCount != $count = $this->getCurrentPage()->countOptions()) {
-            throw $this->createExpectationException(
-                sprintf('Expecting to see %d option, saw %d.', $expectedCount, $count)
-            );
-        }
+        $this->spin(function () use ($expectedCount) {
+            $count = $this->getCurrentPage()->countOptions();
+            if ($expectedCount != $count) {
+                throw new \RuntimeException(sprintf('Expecting to see %d options, saw %d.', $expectedCount, $count));
```
willy-ahva: Why not ExpectactionException ?

jmleroux: It will be a TimeOutException anyway in the calling spin.

PR #4170, opened 2016-03-18
TIP-221: domain layer extraction

jjanvier:
This PR:
* removes the AttributeOptionManager
* unglues attribute type form from its definition without BC break (important for integrators that made a custom attribute type)
* makes all the code (except FileStorage component and bundle) totally agnostic from Flysystem, which is cool :)

Before: 26 coupling rules violated
Now: 18 coupling rules violated, still some work, but we're getting closer :)

| Q                                 | A
| --------------------------------- | ---
| Added Specs                       | Y
| Added Behats                      | -
| Changelog updated                 | Y
| Review and 2 GTM                  | 
| Micro Demo to the PO (Story only) | -
| Migration script                  | -
| Tech Doc                          | -

PR #4212, opened 2016-03-24
TIP-221: domain layer extraction

jjanvier:
This PR:
* removes the AttributeOptionManager
* unglues attribute type form from its definition without BC break (important for integrators that made a custom attribute type)
* makes all the code (except FileStorage component and bundle) totally agnostic from Flysystem, which is cool :)

Before: 26 coupling rules violated
Now: 18 coupling rules violated, still some work, but we're getting closer :)

| Review and 2 GTM                  | Y, in the PR I reverted https://github.com/akeneo/pim-community-dev/pull/4170


jjanvier:
merged as the failing behat fails also on master

PR #4219, opened 2016-03-26
[WIP] Install command events (any demand for this?)

koenreiniers:
This is something I've cooked up to fulfill one of my own needs. Extending the install command is impossible without extending the existing classes, copy/pasting a lot of code.  
For the problem I ran into a post-install event was enough, but I figured I'd build a bit upon this idea and add some more events, which you can view in the AssetsCommand.  

I realise that there probably won't be much demand for this feature, since a lot of you will be deploying PIM only once and be done with it. If there is though, I'd gladly continue working on it and submit a pull request.

### DIFF
--- a/src/Pim/Bundle/InstallerBundle/Command/AssetsCommand.php
+++ b/src/Pim/Bundle/InstallerBundle/Command/AssetsCommand.php
@@ -1,5 +1,5 @@
 <?php
-namespace Erp\Bundle\InstallerBundle\Command;
+namespace Pim\Bundle\InstallerBundle\Command;
 
 use Pim\Bundle\InstallerBundle\Event\CommandBatchEvent;
 use Pim\Bundle\InstallerBundle\Event\CommandEvent;
--- a/src/Pim/Bundle/InstallerBundle/Command/AssetsCommand.php
+++ b/src/Pim/Bundle/InstallerBundle/Command/AssetsCommand.php
@@ -1,11 +1,17 @@
 <?php
+namespace Erp\Bundle\InstallerBundle\Command;
 
-namespace Pim\Bundle\InstallerBundle\Command;
-
-use Pim\Bundle\InstallerBundle\CommandExecutor;
-use Symfony\Component\Console\Command\Command;
+use Pim\Bundle\InstallerBundle\Event\CommandBatchEvent;
+use Pim\Bundle\InstallerBundle\Event\CommandEvent;
+use Pim\Bundle\InstallerBundle\Event\InstallEvent;
+use Pim\Bundle\InstallerBundle\Event\InstallEvents;
+use Pim\Bundle\InstallerBundle\SimpleCommand\SimpleCommand;
+use Pim\Bundle\InstallerBundle\SimpleCommand\SimpleCommandBatch;
+use Pim\Bundle\InstallerBundle\SimpleCommand\SimpleCommandExecutor;
+use Symfony\Bundle\FrameworkBundle\Command\ContainerAwareCommand;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\EventDispatcher\EventDispatcherInterface;
 
 /**
  * Assets dump command
@@ -14,8 +20,14 @@ use Symfony\Component\Console\Output\OutputInterface;
  * @copyright 2013 Akeneo SAS (http://www.akeneo.com)
  * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
  */
-class AssetsCommand extends Command
+class AssetsCommand extends ContainerAwareCommand
 {
+    /** @var SimpleCommandExecutor */
+    protected $commandExecutor;
+
+    /** @var EventDispatcherInterface */
+    protected $eventDispatcher;
+
     /**
      * {@inheritdoc}
      */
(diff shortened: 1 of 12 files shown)


nidup:
Hi @koenreiniers 

:tada: first, thank you very much for the proposal :tada: 

Happy to see your PR especially because this bundle is not extensible and quite messy (we never took enough time to take care of it, let's say it's not our better legacy :wink: ). FYI, we recently improved it a bit by making it use the standard imports to setup data.

I do agree, there will be not so much demand for this but, for instance, in our enterprise edition, we override it to add extra command and configuration so your approach seems really interesting.
Could you give me a bit more details about what is the custom code that you need to add in your install process?

At the end, installing the PIM involves to run a bunch of commands to check requirements, install database, install assets. As each of these command are run through the CommandExecutor, I wonder if a couple of dispatches could be done inside this class (pre exec, post exec).

I'll be glad to continue to work with you on this topic, to be able to merge your contribution at the end, I'll need that you read and sign the following contributor agreement http://www.akeneo.com/contributor-license-agreement/

PS: i added few comments in the PR

src/Pim/Bundle/InstallerBundle/Command/InstallCommand.php
```diff
@@ -53,6 +60,7 @@ protected function initialize(InputInterface $input, OutputInterface $output)
      */
     protected function execute(InputInterface $input, OutputInterface $output)
     {
+        $this->eventDispatcher->dispatch(InstallEvents::PRE_INSTALL, new InstallEvent());
```
nidup: ok, interesting to have a pre and post install

src/Pim/Bundle/InstallerBundle/Event/CommandEvent.php
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    /**
+     * Sets the command
+     *
+     * @param SimpleCommandInterface $command
+     *
+     * @return $this
+     */
+    public function setCommand($command)
```
nidup: the command is set through the constructor, i would not add this setter to make immutable the CommandEvent

src/Pim/Bundle/InstallerBundle/Event/InstallEvent.php
```diff
@@ -0,0 +1,8 @@
+<?php
+namespace Pim\Bundle\InstallerBundle\Event;
+
+use Symfony\Component\EventDispatcher\GenericEvent;
+
+class InstallEvent extends GenericEvent
```
nidup: I would not add this base InstallEvent, CommandBatchEvent and CommandEvent can directly extends Symfony\Component\EventDispatcher\Event

src/Pim/Bundle/InstallerBundle/Event/CommandBatchEvent.php
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    /**
+     * Sets the command batch
+     *
+     * @param SimpleCommandBatch $commandBatch
+     *
+     * @return $this
+     */
+    public function setCommandBatch($commandBatch)
```
nidup: the $commandBatch is always set through the constructor, i would not add this setter to make immutable the CommandBatchEvent

src/Pim/Bundle/InstallerBundle/SimpleCommand/SimpleCommandInterface.php
```diff
@@ -0,0 +1,37 @@
...
+    public function getCommand();
+
+    /**
+     * Sets the command
+     *
+     * @param string$command
+     *
+     * @return $this
+     */
+    public function setCommand($command);
```
nidup: no need of setter

src/Pim/Bundle/InstallerBundle/SimpleCommand/SimpleCommandInterface.php
```diff
@@ -0,0 +1,37 @@
...
+    public function getParams();
+
+    /**
+     * Sets the params
+     *
+     * @param array $params
+     *
+     * @return $this
+     */
+    public function setParams(array $params);
```
nidup: no need of setter

src/Pim/Bundle/InstallerBundle/Command/AssetsCommand.php
```diff
@@ -31,7 +43,8 @@ protected function configure()
      */
     protected function initialize(InputInterface $input, OutputInterface $output)
     {
-        $this->commandExecutor = new CommandExecutor(
+        $this->eventDispatcher = $this->getContainer()->get("event_dispatcher");
+        $this->commandExecutor = new SimpleCommandExecutor(
```
nidup: I would rename SimpleCommandInterface to CommandDefinition.
Then assume to change the CommandExecutor as following would make the whole system simpler,
```
class CommandExecutor
    public function runCommand(CommandDefinition $definition)
    public function runAllCommand(CommandDefinition[] $definitions)
```
It would imply a BC Break (easy to explain and fix during migrations).

jjanvier: and CommandDefinitionInterface could be at the root of the InstallerBundle

jjanvier:
Hello @koenreiniers,

Thanks for contributing :tada:! We really appreciate :smiley: 

I agree with @nidup that refactoring the `CommandExecutor` is better than introducing a new `SimpleCommandExecutor`. Do you need further explanation on this topic?

Also, to be able to merge your PR could you:
- [ ] sign the contributor agreement http://www.akeneo.com/contributor-license-agreement/
- [ ] launch `php-cs-fixer` to clean your code style
- [ ] add the licence to every class you create
- [ ] add PHPdoc to every class you create
- [ ] create specs if needed

nidup:
Another PR covering a related/same need https://github.com/akeneo/pim-community-dev/pull/5722

nidup:
Another pull request improving the extensibility of the installer has been merged https://github.com/akeneo/pim-community-dev/pull/5722 

Don't hesitate to re-open an issue if you want discuss improvements on its new state

PR #4245, opened 2016-03-30
[POC][WIP] Simplify the batch jobs configuration

nidup:
Today, Batch Bundle is quite problematic for several reasons.
First, the configuration and UI concerns are totally mixed in basic batch classes.
Second, to implement the bulk processes, we introduced a database configuration class to store the config, we should avoid to use it to allow to run a job only by passing configuration as argument (no need to store then run).


src/Akeneo/Component/Batch/Exception/UndefinedParameterException.php
```diff
@@ -0,0 +1,14 @@
+<?php
+
+namespace Akeneo\Component\Batch\Exception;
+
+/**
+ * Class ConfigurationException
```
pierallard: Naming

nidup: Just a POC i close the PR the new one is TIP-303-job-parameters :wink:

nidup:
Closed in favor of https://github.com/akeneo/pim-community-dev/pull/4326

PR #4259, opened 2016-04-01
PIM-5696: Fix XLSX products import with numeric values in text attributes

phaseinducer:
**What was the issue?**

When you edit an Excel file and put a numeric value in a cell, it is stored as a number in the xlsx file.
It is a problem if it concerns a text attribute or the sku. During import you will have an error like "Attribute ... expects a string as data, interger given".

**What's the fix?**

The processor use converters located in Pim\Component\Connector\ArrayConverter\Flat\Product\ValueConverter to transform the data coming from the reader into the standard format. One of these converters is the ScalarConverter, which does nothing but build the standard format array (locale + scope + value) and handles several attribute types : texts, dates, numbers and booleans.
I created a separated TextConverter that allow to cast the value to string during conversion.

I also had to modify the ProductAssociationProcessor to convert the data properly and make it work with a full numeric sku.

| Changelog updated                 | n/a

### DIFF
--- a/features/Pim/Behat/Context/Domain/Collect/ImportProfilesContext.php
+++ b/features/Pim/Behat/Context/Domain/Collect/ImportProfilesContext.php
@@ -37,7 +37,14 @@ class ImportProfilesContext extends PimContext
             $writer = WriterFactory::create($extension);
             $writer->openToFile($filename);
             foreach (explode(PHP_EOL, $string) as $row) {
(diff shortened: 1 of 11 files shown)


jjanvier:
GTM
and very good explanation

ClementGautier:
GTM

SamirBoulil:
Good Work GTM !

PR #4273, opened 2016-04-03
GITHUB-4220: small behat scenario fixes

nidup:


a2xchip:
@nidup The info below csv header were not checked and i left it as was )

PR #4279, opened 2016-04-04
PIM-5725: Fix reference data name of the attribute in case it's not

jjanvier:
PIM-5725: Fix reference data name of the attribute in case this attribute is not a reference data

Add a wrong configuration to your SKU attribute. In the column "properties" set:
`a:1:{s:19:"reference_data_name";s:0:"";}
`. AFAIK, it can only be done manually. So that's impossible to add a Behat on this.

During the product normalization, `""` is returned instead of `null` here https://github.com/akeneo/pim-community-dev/blob/1.4/src/Pim/Bundle/TransformBundle/Normalizer/MongoDB/ProductValueNormalizer.php#L138.
Then the product normalization is totally wrong and the product can not be saved.

| Added Behats                      | impossible

### DIFF
--- a/CHANGELOG-1.4.md
+++ b/CHANGELOG-1.4.md
@@ -2,6 +2,7 @@
 
 ## Bug fixes
 
+- PIM-5725: Fix reference data name of the attribute in case this attribute is not a reference data
 - PIM-5699: Fix 'is equal to' operator in export / import history grid filter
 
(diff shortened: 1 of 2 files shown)


nidup:
GTM. what about add a business validation to enforce this?

SamirBoulil:
Gtm

PR #4297, opened 2016-04-06
TIP-245: Extract a DateTimeFilter from DateFilter

phaseinducer:
**Why?**

Natively, a product can have multiple dates :
- fields "updated_at" and "created_at"
- attributes of type "pim_catalog_date"

Until now in the PQB we had only a DateFilter that handled both cases. It was complicated, buggy (with the "not between" operator for example) and it didn't allow to filter precisely on "created at" and "updated at" since all inputs were dates without time.

**What this PR does?**

- It separates the two cases with one DateFilter that handles attributes only, and one DateTimeFilter that handles fields only.
- The DateTimeFilter can now filter to the second.
- In the case of filtering the product grid, it's not the responsibility of the PQB filters anymore to add times to input values. This logic has been moved to the grid filter (Pim\Bundle\FilterBundle\Filter\ProductValue\DateTimeRangeFilter).

### DIFF
--- a/features/filter/date.feature
+++ b/features/filter/date.feature
@@ -7,10 +7,10 @@ Feature: Filter on date
     Given a "footwear" catalog configuration
     And the following products:
       | sku     | destocking_date |
+      | BOOTBL  | 2013-01-25      |
+      | BOOTBS  | 2015-01-20      |
       | BOOTBXS | 2015-01-25      |
       | BOOTWXS | 2015-03-25      |
-      | BOOTBS  | 2015-01-20      |
-      | BOOTBL  | 2013-01-25      |
       | BOOTRXS |                 |
     Then I should get the following results for the given filters:
       | filter                                                                                                                                   | result                                     |
@@ -18,7 +18,7 @@ Feature: Filter on date
       | [{"field":"destocking_date", "operator":">", "value": "2015-01-19"}]                                                                     | ["BOOTBXS", "BOOTBS", "BOOTWXS"]           |
       | [{"field":"destocking_date", "operator":"<", "value": "2015-01-21"}]                                                                     | ["BOOTBS", "BOOTBL"]                       |
       | [{"field":"destocking_date", "operator":"BETWEEN", "value": ["2015-01-20", "2015-03-25"]}]                                               | ["BOOTBXS", "BOOTWXS", "BOOTBS"]           |
-      | [{"field":"destocking_date", "operator":"NOT BETWEEN", "value": ["2015-01-20", "2015-03-25"]}]                                           | ["BOOTBL", "BOOTBS"]                       |
+      | [{"field":"destocking_date", "operator":"NOT BETWEEN", "value": ["2015-01-20", "2015-03-25"]}]                                           | ["BOOTBL"]                       |
       | [{"field":"destocking_date", "operator":"NOT EMPTY", "value": null}]                                                                     | ["BOOTBXS", "BOOTWXS", "BOOTBS", "BOOTBL"] |
       | [{"field":"destocking_date", "operator":"!=", "value": "2015-01-20"}]                                                                    | ["BOOTBXS", "BOOTWXS", "BOOTBL"]           |
       | [{"field":"destocking_date", "operator":">", "value":"2015-01-19"}, {"field":"destocking_date", "operator":"!=", "value":"2015-03-25"}]  | ["BOOTBXS", "BOOTBS"]                      |
--- a/features/filter/date.feature
+++ b/features/filter/date.feature
@@ -18,7 +18,7 @@ Feature: Filter on date
       | [{"field":"destocking_date", "operator":">", "value": "2015-01-19"}]                                                                     | ["BOOTBXS", "BOOTBS", "BOOTWXS"]           |
       | [{"field":"destocking_date", "operator":"<", "value": "2015-01-21"}]                                                                     | ["BOOTBS", "BOOTBL"]                       |
       | [{"field":"destocking_date", "operator":"BETWEEN", "value": ["2015-01-20", "2015-03-25"]}]                                               | ["BOOTBXS", "BOOTWXS", "BOOTBS"]           |
-      | [{"field":"destocking_date", "operator":"NOT BETWEEN", "value": ["2015-01-20", "2015-03-25"]}]                                           | ["BOOTBL"]                       |
+      | [{"field":"destocking_date", "operator":"NOT BETWEEN", "value": ["2015-01-20", "2015-03-25"]}]                                           | ["BOOTBL"]                                 |
       | [{"field":"destocking_date", "operator":"NOT EMPTY", "value": null}]                                                                     | ["BOOTBXS", "BOOTWXS", "BOOTBS", "BOOTBL"] |
       | [{"field":"destocking_date", "operator":"!=", "value": "2015-01-20"}]                                                                    | ["BOOTBXS", "BOOTWXS", "BOOTBL"]           |
       | [{"field":"destocking_date", "operator":">", "value":"2015-01-19"}, {"field":"destocking_date", "operator":"!=", "value":"2015-03-25"}]  | ["BOOTBXS", "BOOTBS"]                      |
--- a/src/Pim/Bundle/CatalogBundle/Doctrine/MongoDBODM/Filter/DateFilter.php
+++ b/src/Pim/Bundle/CatalogBundle/Doctrine/MongoDBODM/Filter/DateFilter.php
@@ -155,7 +155,7 @@ class DateFilter extends AbstractAttributeFilter implements AttributeFilterInter
      *
      * @throws InvalidArgumentException
      *
-     * @return integer
+     * @return int|null
      */
     protected function formatSingleValue($type, $value)
     {
--- a/src/Pim/Bundle/CatalogBundle/Doctrine/MongoDBODM/Filter/DateTimeFilter.php
+++ b/src/Pim/Bundle/CatalogBundle/Doctrine/MongoDBODM/Filter/DateTimeFilter.php
@@ -4,68 +4,52 @@ namespace Pim\Bundle\CatalogBundle\Doctrine\MongoDBODM\Filter;
 
 use Pim\Bundle\CatalogBundle\Doctrine\MongoDBODM\ProductQueryUtility;
 use Pim\Component\Catalog\Exception\InvalidArgumentException;
-use Pim\Component\Catalog\Model\AttributeInterface;
-use Pim\Component\Catalog\Query\Filter\AttributeFilterInterface;
+use Pim\Component\Catalog\Query\Filter\FieldFilterInterface;
 use Pim\Component\Catalog\Query\Filter\Operators;
-use Pim\Component\Catalog\Validator\AttributeValidatorHelper;
 
 /**
- * Date filter
+ * Datetime filter
  *
- * @author    Nicolas Dupont <nicolas@akeneo.com>
- * @copyright 2014 Akeneo SAS (http://www.akeneo.com)
+ * @author    Yohan Blain <yohan.blain@akeneo.com>
+ * @copyright 2016 Akeneo SAS (http://www.akeneo.com)
  * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
  */
-class DateFilter extends AbstractAttributeFilter implements AttributeFilterInterface
+class DateTimeFilter extends AbstractFilter implements FieldFilterInterface
 {
-    const DATETIME_FORMAT = 'Y-m-d';
+    const DATETIME_FORMAT = 'Y-m-d H:i:s';
 
     /** @var array */
-    protected $supportedAttributes;
+    protected $supportedFields;
 
     /**
-     * @param AttributeValidatorHelper $attrValidatorHelper
-     * @param array                    $supportedAttributes
-     * @param array                    $supportedOperators
+     * @param array $supportedFields
+     * @param array $supportedOperators
      */
     public function __construct(
-        AttributeValidatorHelper $attrValidatorHelper,
-        array $supportedAttributes = [],
+        array $supportedFields = [],
         array $supportedOperators = []
     ) {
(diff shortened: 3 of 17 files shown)


src/Pim/Bundle/CatalogBundle/Doctrine/MongoDBODM/Filter/DateFilter.php
```diff
@@ -79,43 +66,54 @@ public function addAttributeFilter(
...
+                $this->qb->field($field)->exists(true);
+                $this->qb->field($field)->notEqual($value);
+                break;
+            case Operators::IS_EMPTY:
+                $this->qb->field($field)->exists(false);
+                break;
+            case Operators::IS_NOT_EMPTY:
+                $this->qb->field($field)->exists(true);
+                break;
         }
```
fitn: default: throw new \Exception?

src/Pim/Bundle/CatalogBundle/Doctrine/MongoDBODM/Filter/DateTimeFilter.php
```diff
@@ -0,0 +1,178 @@
...
+                $this->qb->field($field)->exists(true);
+                $this->qb->field($field)->notEqual($value);
+                break;
+            case Operators::IS_EMPTY:
+                $this->qb->field($field)->exists(false);
+                break;
+            case Operators::IS_NOT_EMPTY:
+                $this->qb->field($field)->exists(true);
+                break;
+        }
```
fitn: default Throw exception?

features/filter/date.feature
```diff
@@ -7,18 +7,18 @@ Feature: Filter on date
...
       | [{"field":"destocking_date", "operator":"=", "value": "2015-01-25"}]                                                                     | ["BOOTBXS"]                                |
       | [{"field":"destocking_date", "operator":">", "value": "2015-01-19"}]                                                                     | ["BOOTBXS", "BOOTBS", "BOOTWXS"]           |
       | [{"field":"destocking_date", "operator":"<", "value": "2015-01-21"}]                                                                     | ["BOOTBS", "BOOTBL"]                       |
       | [{"field":"destocking_date", "operator":"BETWEEN", "value": ["2015-01-20", "2015-03-25"]}]                                               | ["BOOTBXS", "BOOTWXS", "BOOTBS"]           |
-      | [{"field":"destocking_date", "operator":"NOT BETWEEN", "value": ["2015-01-20", "2015-03-25"]}]                                           | ["BOOTBL", "BOOTBS"]                       |
+      | [{"field":"destocking_date", "operator":"NOT BETWEEN", "value": ["2015-01-20", "2015-03-25"]}]                                           | ["BOOTBL"]                       |
```
skeleton: indent `|` please

src/Pim/Bundle/CatalogBundle/Doctrine/MongoDBODM/Filter/DateFilter.php
```diff
@@ -17,43 +16,31 @@
...
     /**
      * @param AttributeValidatorHelper $attrValidatorHelper
      * @param array                    $supportedAttributes
-     * @param array                    $supportedFields
      * @param array                    $supportedOperators
      */
     public function __construct(
         AttributeValidatorHelper $attrValidatorHelper,
         array $supportedAttributes = [],
-        array $supportedFields = [],
```
skeleton: bc break

src/Pim/Bundle/CatalogBundle/Doctrine/MongoDBODM/Filter/DateFilter.php
```diff
@@ -17,43 +16,31 @@
  * @copyright 2014 Akeneo SAS (http://www.akeneo.com)
  * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
  */
-class DateFilter extends AbstractAttributeFilter implements AttributeFilterInterface, FieldFilterInterface
```
skeleton: bc breaks

src/Pim/Bundle/CatalogBundle/Doctrine/MongoDBODM/Filter/DateFilter.php
```diff
@@ -155,95 +153,44 @@ protected function formatValues($type, $value)
      * @param string $type
      * @param mixed  $value
      *
-     * @return string
+     * @throws InvalidArgumentException
+     *
+     * @return integer
```
skeleton: @return integer|null

src/Pim/Bundle/CatalogBundle/Doctrine/MongoDBODM/Filter/DateFilter.php
```diff
@@ -17,43 +16,31 @@
  * @copyright 2014 Akeneo SAS (http://www.akeneo.com)
  * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
  */
-class DateFilter extends AbstractAttributeFilter implements AttributeFilterInterface, FieldFilterInterface
+class DateFilter extends AbstractAttributeFilter implements AttributeFilterInterface
 {
-    /** @var array */
-    protected $supportedAttributes;
+    const DATETIME_FORMAT = 'Y-m-d';
```
pierallard: Why this const is declared here ? In a MongoDB filter ?

src/Pim/Bundle/CatalogBundle/Doctrine/MongoDBODM/Filter/DateFilter.php
```diff
@@ -79,43 +66,54 @@ public function addAttributeFilter(
...
+     * @param string $operator
+     * @param mixed  $value
      */
-    protected function getTimestamp($data, $endOfDay = false)
+    protected function applyFilter($field, $operator, $value)
     {
-        if ($data instanceof \DateTime && true === $endOfDay) {
-            $data->setTime(23, 59, 59);
-        } elseif (!$data instanceof \DateTime && true === $endOfDay) {
-            $data = sprintf('%s 23:59:59', $data);
```
pierallard: Thank you :heart:

src/Pim/Bundle/CatalogBundle/Doctrine/MongoDBODM/Filter/DateFilter.php
nidup: yep, even if it's a very internal class, thank you to indicate it in the BC break list. 
A day we'll have a list of supported `@api` and this kind of class will not be inside ;)
For now let's indicate as much as we can :)

src/Pim/Bundle/CatalogBundle/Doctrine/MongoDBODM/Filter/DateTimeFilter.php
```diff
@@ -0,0 +1,178 @@
...
+/**
+ * Datetime filter
+ *
+ * @author    Yohan Blain <yohan.blain@akeneo.com>
+ * @copyright 2016 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+class DateTimeFilter extends AbstractFilter implements FieldFilterInterface
+{
+    const DATETIME_FORMAT = 'Y-m-d H:i:s';
```
pierallard: Same remark

src/Pim/Bundle/CatalogBundle/Doctrine/MongoDBODM/Filter/DateTimeFilter.php
```diff
@@ -0,0 +1,178 @@
...
+use Pim\Component\Catalog\Query\Filter\Operators;
+
+/**
+ * Datetime filter
+ *
+ * @author    Yohan Blain <yohan.blain@akeneo.com>
+ * @copyright 2016 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+class DateTimeFilter extends AbstractFilter implements FieldFilterInterface
```
pierallard: IMHO, DateTimeFilter should extend DateFilter.

src/Pim/Bundle/CatalogBundle/Doctrine/ORM/Filter/DateFilter.php
```diff
@@ -16,45 +15,31 @@
  * @copyright 2014 Akeneo SAS (http://www.akeneo.com)
  * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
  */
-class DateFilter extends AbstractAttributeFilter implements FieldFilterInterface, AttributeFilterInterface
+class DateFilter extends AbstractAttributeFilter implements AttributeFilterInterface
 {
-    /** @var array */
-    protected $supportedAttributes;
+    const DATETIME_FORMAT = 'Y-m-d';
```
pierallard: You see, you declare same const 2 times, for the same functionnality.

src/Pim/Bundle/CatalogBundle/Doctrine/ORM/Filter/DateTimeFilter.php
```diff
@@ -0,0 +1,142 @@
...
+        }
+
+        throw InvalidArgumentException::expected(
+            $type,
+            'array with 2 elements, string or \DateTime',
+            'filter',
+            'date',
+            print_r($value, true)
+        );
+    }
```
pierallard: A loooooot of duplicated code..... :(

src/Pim/Bundle/FilterBundle/Filter/ProductValue/DateTimeRangeFilter.php
```diff
@@ -13,18 +13,39 @@
  */
 class DateTimeRangeFilter extends AbstractDateFilter
 {
+    const DATETIME_FORMAT = 'Y-m-d H:i:s';
```
pierallard: Here should be a good place for you const, no ?

src/Pim/Bundle/CatalogBundle/Doctrine/MongoDBODM/Filter/DateFilter.php
nidup: dunno what we do in other filters, but make the API more defensive is a good idea to avoid silent failing.
in other hand, this class is not made to be used directly (btw all our PQB filters could be declared as private services), and maybe it's not even possible (when properly used, aka via the PQB) to call this method with a not supported operator.
if we add an exception here, I would use a dedicated exception declared in the catalog component, not a generic one and I would add it in all filters for consistency purpose, so imho not the scope of this PR.

pierallard:
There is a lot of duplicated code, between filters and between date/datetime. IMHO you may find a solution to prevent it.

nidup:
Nice PR, thank you for the "boy scouting", please add the BC Break in the changelog then GTM.

src/Pim/Bundle/CatalogBundle/Doctrine/MongoDBODM/Filter/DateTimeFilter.php
nidup: Imvho, inheritance is not a good way to share code here (and in general), would prefer a bit of duplicated code in very simple classes (we're talking here of calls to the query builder and exception throwing) than a X level inheritance introducing issues like "ah in fact, this operator is not supported by children but it's supported by parent". composition FTW.

src/Pim/Bundle/CatalogBundle/Doctrine/MongoDBODM/Filter/DateFilter.php
phaseinducer: Just a convenient way to define the format used here. Not related to the localization.

src/Pim/Bundle/CatalogBundle/Doctrine/MongoDBODM/Filter/DateTimeFilter.php
```diff
@@ -0,0 +1,178 @@
...
+                break;
+        }
+    }
+
+    /**
+     * Format values to string or array of strings
+     *
+     * @param string $type
+     * @param mixed  $value
+     *
```
skeleton: add @throw InvalidArgumentException

src/Pim/Bundle/CatalogBundle/Doctrine/ORM/Filter/DateFilter.php
```diff
@@ -16,45 +15,31 @@
  * @copyright 2014 Akeneo SAS (http://www.akeneo.com)
  * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
  */
-class DateFilter extends AbstractAttributeFilter implements FieldFilterInterface, AttributeFilterInterface
```
skeleton: bc break

src/Pim/Bundle/CatalogBundle/Doctrine/ORM/Filter/DateFilter.php
```diff
@@ -16,45 +15,31 @@
...
-     *
      * @param AttributeValidatorHelper $attrValidatorHelper
      * @param array                    $supportedAttributes
-     * @param array                    $supportedFields
      * @param array                    $supportedOperators
      */
     public function __construct(
         AttributeValidatorHelper $attrValidatorHelper,
         array $supportedAttributes = [],
-        array $supportedFields = [],
```
skeleton: same

src/Pim/Bundle/CatalogBundle/Doctrine/ORM/Filter/DateTimeFilter.php
```diff
@@ -0,0 +1,142 @@
...
+    {
+        return in_array($field, $this->supportedFields);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function addFieldFilter($field, $operator, $value, $locale = null, $scope = null, $options = [])
+    {
+        if (Operators::IS_EMPTY === $operator) {
```
skeleton: ternary ?

skeleton:
gtm once fixed :+1:

PR #4305, opened 2016-04-06
Batch cleanup JobLauncher, Job, Step

nidup:
- Remove useless setConfig / getConfig methods from the JobLauncherInterface
 - Remove setters setJobRepository / setEventDispatcher from Job and AbstractStep and inject these objects by constructor.,AbstractStep and Job always requires to use a JobRepo and an eventDispatcher in their internal logic, these are not optional arguments. These setters are only used in the related factories.
 - Remove deprecated BaseConnectorBundle ValidatorStep (has been replaced in 1.4 and it's deprecated since then).
 - JobLauncherInterface takes an array as configuration and not an already json encoded string
 - Move BaseConnectorBundle validation from annotations to files
 - Bump composer and fix the BatchJobHandler


composer.json
```diff
@@ -49,15 +49,15 @@
...
+        "monolog/monolog": "1.18.2",
         "nelmio/api-doc-bundle": "2.9.0",
         "oneup/flysystem-bundle": "1.1.0",
         "sensio/distribution-bundle": "~4.0",
         "sensio/framework-extra-bundle": "~3.0,>=3.0.2",
         "sensio/generator-bundle": "2.3.5",
         "symfony/assetic-bundle": "2.3.1",
         "symfony/icu": "1.1.0",
-        "symfony/monolog-bundle": "~2.4",
+        "symfony/monolog-bundle": "2.10.0",
```
jmleroux: why not `~2.10.0` ?

Issue #4310, opened 2016-04-08
Akeneo\Component\Batch\Job\Job bad dependency


jmleroux:
Batch Job seems to depend of StepInterface but it really depends of [AbstractStep](https://github.com/akeneo/pim-community-dev/blob/master/src/Akeneo/Component/Batch/Step/AbstractStep.php) because of this line:
https://github.com/akeneo/pim-community-dev/blob/master/src/Akeneo/Component/Batch/Job/Job.php#L395

Also, it maybe show a missing spec, because it should have been detected by unit tests.

skeleton:
Can we close this issue @jmleroux ?

PR #4311, opened 2016-04-08
Fix batch job dependency on AbstractStep

jmleroux:
This PR fix the batch job dependency raised in issue #4310 
But maybe the real fix should be to really rely on the interface


jmleroux:
It seems that [execute](https://github.com/akeneo/pim-community-dev/pull/4311/files#diff-3f6dd69d59c93376c1025b789d0d0956R111) is just partially covered

PR #4317, opened 2016-04-12
PIM-5653: Create a storage agnostic Product Reader using the PQB

phaseinducer:

This PR introduces a new product reader. It's storage-agnostic thanks to the PQB.
It will replace the old ORM/Mongo readers, which are now deprecated.
Inspired by what has been done on [mass edit](https://github.com/akeneo/pim-community-dev/blob/master/src/Pim/Bundle/EnrichBundle/Connector/Reader/MassEdit/FilteredProductReader.php) and [EnhancedConnectorBundle](https://github.com/akeneo-labs/EnhancedConnectorBundle).

| Added Behats                      | n/a
| Micro Demo to the PO (Story only) | n/a

### DIFF
--- a/features/filter/completeness_without_locale.feature
+++ b/features/filter/completeness_without_locale.feature
@@ -1,0 +1,17 @@
+Feature: Filter on completeness
+  In order to filter on products
+  As an internal process or any user
+  I need to be able to filter on product by completeness without specifying any locale
+
+  Scenario: Successfully filter on completeness without specifying any locale
+    Given an "apparel" catalog configuration
+    And the following products:
+      | sku     | family  | name-en_US      | name-en_GB      | price         | size   | color | manufacturer     |
+      | TSHIRTS | tshirts | T-shirt S Black | T-shirt S Black | 10 USD, 9 GBP | size_S | black | american_apparel |
+      | TSHIRTM | tshirts | T-shirt M Black |                 | 10 USD, 9 GBP | size_M | black | american_apparel |
+      | TSHIRTL | tshirts | T-shirt L Black |                 | 10 USD, 9 GBP | size_L |       |                  |
+    Then I should get the following results for the given filters:
+      | filter                                                                       | result                 |
+      | [{"field":"completeness", "operator":"=",  "value": 100, "scope": "tablet"}] | ["TSHIRTS", "TSHIRTM"] |
+      | [{"field":"completeness", "operator":"<",  "value": 100, "scope": "tablet"}] | ["TSHIRTM", "TSHIRTL"] |
+      | [{"field":"completeness", "operator":"!=", "value": 100, "scope": "tablet"}] | ["TSHIRTM", "TSHIRTL"] |
--- a/src/Pim/Bundle/CatalogBundle/Doctrine/ORM/Join/CompletenessJoin.php
+++ b/src/Pim/Bundle/CatalogBundle/Doctrine/ORM/Join/CompletenessJoin.php
@@ -17,8 +17,6 @@ class CompletenessJoin
     protected $qb;
 
     /**
-     * Instanciate the utility
-     *
      * @param QueryBuilder $qb
      */
     public function __construct(QueryBuilder $qb)
@@ -31,44 +29,55 @@ class CompletenessJoin
      *
      * @param string $completenessAlias the join alias
      * @param string $locale            the locale
-     * @param sting  $scope             the scope
+     * @param string $scope             the scope
      *
      * @return CompletenessJoin
      */
     public function addJoins($completenessAlias, $locale, $scope)
     {
-        $rootAlias         = $this->qb->getRootAlias();
+        $rootAlias         = $this->qb->getRootAliases()[0];
         $localeAlias       = $completenessAlias.'Locale';
         $channelAlias      = $completenessAlias.'Channel';
-        $rootEntity        = current($this->qb->getRootEntities());
+
+        $rootEntity          = $this->qb->getRootEntities()[0];
         $completenessMapping = $this->qb->getEntityManager()
             ->getClassMetadata($rootEntity)
             ->getAssociationMapping('completenesses');
         $completenessClass = $completenessMapping['targetEntity'];
 
+        $joinCondition = sprintf('%s.product = %s.id', $completenessAlias, $rootAlias);
+
         $this->qb
-            ->leftJoin(
-                'PimCatalogBundle:Locale',
-                $localeAlias,
-                'WITH',
-                $localeAlias.'.code = :cLocaleCode'
-            )
             ->leftJoin(
                 'PimCatalogBundle:Channel',
                 $channelAlias,
                 'WITH',
                 $channelAlias.'.code = :cScopeCode'
             )
+            ->setParameter('cScopeCode', $scope);
+
+        $joinCondition .= sprintf(' AND %s.channel = %s.id', $completenessAlias, $channelAlias);
+
+        if (null !== $locale) {
+            $this->qb
+                ->leftJoin(
+                    'PimCatalogBundle:Locale',
+                    $localeAlias,
+                    'WITH',
+                    $localeAlias.'.code = :cLocaleCode'
+                )
+                ->setParameter('cLocaleCode', $locale);
+
+            $joinCondition .= sprintf(' AND %s.locale = %s.id', $completenessAlias, $localeAlias);
+        }
+
+        $this->qb
             ->leftJoin(
                 $completenessClass,
                 $completenessAlias,
                 'WITH',
-                $completenessAlias.'.locale = '.$localeAlias.'.id AND '.
-                $completenessAlias.'.channel = '.$channelAlias.'.id AND '.
-                $completenessAlias.'.product = '.$rootAlias.'.id'
-            )
-            ->setParameter('cLocaleCode', $locale)
-            ->setParameter('cScopeCode', $scope);
+                $joinCondition
+            );
 
         return $this;
     }
--- a/src/Pim/Bundle/ConnectorBundle/Resources/config/batch_jobs/csv_connector.yml
+++ b/src/Pim/Bundle/ConnectorBundle/Resources/config/batch_jobs/csv_connector.yml
@@ -140,7 +140,7 @@ connector:
                 export:
                     title:     pim_connector.jobs.csv_product_export.export.title
                     services:
-                        reader:    pim_base_connector.reader.doctrine.product
+                        reader:    pim_connector.reader.product
                         processor: pim_base_connector.processor.product_to_flat_array
                         writer:    pim_connector.writer.file.csv_product
                     parameters:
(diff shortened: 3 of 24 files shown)


nidup:
Much API, such :heartbeat:

src/Pim/Component/Connector/Reader/ProductReader.php
```diff
@@ -0,0 +1,233 @@
...
+use Pim\Component\Catalog\Repository\ChannelRepositoryInterface;
+
+/**
+ * Storage-agnostic product reader using the Product Query Builder
+ *
+ * @author    Yohan Blain <yohan.blain@akeneo.com>
+ * @copyright 2016 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+class ProductReader extends AbstractConfigurableStepElement implements ItemReaderInterface, StepExecutionAwareInterface
```
nidup: Really wonder if we should put this class directly in the Reader folder or if we need a subfolder.

The BaseReader is in a Reader/Doctrine folder which is totally weird because it does not even depends on doctrine... In Processor we separate by normalization / denormalization. In Writer, we separate by Doctrine and File... and writers are not doctrine tinted...

Let's take a bit of time to think about a relevant orga here, it could be a subfolder storage (looks very generic), or database (quite generic but more precise without being doctrine, orm, mongo tinted).

Database sounds ok but if we do, could we move the base reader here too and change the writer/doctrine to writer/database (with a sed in the upgrade.md, service alias etc, not necessary in the scope of this PR :) )

@jjanvier any opinion?

phaseinducer: Ok to put the new reader into a "database" subfolder.
But the BaseReader relies on "Doctrine\Common\Persistence\ObjectRepository" so not so simple.

src/Pim/Component/Catalog/Query/ProductQueryBuilder.php
```diff
@@ -105,28 +105,28 @@ public function getQueryBuilder()
...
 
-        if ($filter === null) {
+        if (null === $filter) {
             throw new \LogicException(
                 sprintf('Filter on field "%s" is not supported', $field)
             );
         }
 
-        if ($filter->supportsOperator($operator) === false) {
+        if (false === $filter->supportsOperator($operator)) {
```
skeleton: `if (!`

src/Pim/Bundle/CatalogBundle/Doctrine/ORM/Join/CompletenessJoin.php
```diff
@@ -31,44 +29,55 @@ public function __construct(QueryBuilder $qb)
...
                 $channelAlias.'.code = :cScopeCode'
             )
+            ->setParameter('cScopeCode', $scope);
+
+        $joinCondition .= sprintf(' AND %s.channel = %s.id', $completenessAlias, $channelAlias);
+
+        if (null !== $locale) {
+            $this->qb
+                ->leftJoin(
+                    'PimCatalogBundle:Locale',
```
jjanvier: hmm really not sure that works if the Locale is override

solivier:
gtm

features/filter/completeness_without_locale.feature
```diff
@@ -0,0 +1,18 @@
...
+  I need to be able to filter on product by completeness without specifying any locale
+
+  Scenario: Successfully filter on completeness without specifying any locale
+    Given an "apparel" catalog configuration
+    And the following products:
+      | sku     | family  | name-en_US      | name-en_GB      | price         | size   | color | manufacturer     |
+      | TSHIRTS | tshirts | T-shirt S Black | T-shirt S Black | 10 USD, 9 GBP | size_S | black | american_apparel |
+      | TSHIRTM | tshirts | T-shirt M Black |                 | 10 USD, 9 GBP | size_M | black | american_apparel |
+      | TSHIRTL | tshirts | T-shirt L Black |                 | 10 USD, 9 GBP | size_L |       |                  |
+    And I launched the completeness calculator
```
skeleton: still needed ? as completeness is always calculated since Olivier's PR

src/Pim/Component/Connector/Reader/ProductReader.php
```diff
@@ -0,0 +1,233 @@
...
+        $this->completenessManager  = $completenessManager;
+        $this->metricConverter      = $metricConverter;
+        $this->objectDetacher       = $objectDetacher;
+        $this->generateCompleteness = $generateCompleteness;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function setChannel($channelCode)
```
jjanvier: setChannelCode

src/Pim/Component/Connector/Reader/ProductReader.php
```diff
@@ -0,0 +1,233 @@
...
+    {
+        $this->channelCode = $channelCode;
+        $this->channel     = null;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getChannel()
```
jjanvier: getChannelCode
and I don't understand why you need a getter/setter on the channel

something smells bad here

src/Pim/Component/Connector/Reader/ProductReader.php
```diff
@@ -0,0 +1,233 @@
...
+    protected $objectDetacher;
+
+    /** @var bool */
+    protected $generateCompleteness;
+
+    /** @var StepExecution */
+    protected $stepExecution;
+
+    /** @var CursorInterface */
+    protected $products;
```
skeleton: seems not used

src/Pim/Component/Connector/Reader/ProductReader.php
phaseinducer: Yep, it's the smell of decomposing batch bundle conf =)
See https://github.com/akeneo/pim-community-dev/blob/master/src/Akeneo/Component/Batch/Item/AbstractConfigurableStepElement.php#L66

features/filter/completeness_without_locale.feature
phaseinducer: Maybe not, you're right. But it should be removed from every scenarios then.

src/Pim/Bundle/CatalogBundle/Doctrine/ORM/Join/CompletenessJoin.php
phaseinducer: Right ! But not the scope of this PR, should be another DX card.

src/Pim/Component/Connector/Reader/ProductReader.php
jjanvier: imho this new reader should be in the folder Doctrine atm
and the folder Doctrine could be renamed to Database, but it's no big deal
and :+1: to rename also the Writer/Doctrine folder to Writer/Database, it will be much easier to understand

the dependency to Doctrine\Common\Persistence\ObjectRepository is totally acceptable (and quite insignificant)
that's why it has been added to the authorized dependencies here https://github.com/akeneo/pim-community-dev/blob/master/.php_cd_transitional.php#L26

jjanvier: but I would not move all this in this PR

jjanvier:
GTM once fixed

src/Pim/Bundle/CatalogBundle/Doctrine/ORM/Join/CompletenessJoin.php
jjanvier: why? you introduced that here

phaseinducer: Nope : https://github.com/akeneo/pim-community-dev/blob/master/src/Pim/Bundle/CatalogBundle/Doctrine/ORM/Join/CompletenessJoin.php#L51
Just moved stuff inside an if. Plus there could be other cases in the PIM and we need to decide how to handle that homogeneously.

src/Pim/Component/Connector/Reader/ProductReader.php
```diff
@@ -0,0 +1,233 @@
...
+
+    /**
+     * {@inheritdoc}
+     */
+    public function initialize()
+    {
+        $this->channel = $this->getChannelByCode($this->channelCode);
+        $pqb           = $this->getProductQueryBuilder($this->channel);
+
+        if ($this->generateCompleteness) {
```
willy-ahva: if (true = ...)

src/Pim/Component/Connector/Reader/ProductReader.php
```diff
@@ -0,0 +1,233 @@
...
+        ];
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function initialize()
+    {
+        $this->channel = $this->getChannelByCode($this->channelCode);
+        $pqb           = $this->getProductQueryBuilder($this->channel);
```
willy-ahva: why didn't you pass the channelCode with the getter ?

src/Pim/Component/Connector/Reader/ProductReader.php
```diff
@@ -0,0 +1,233 @@
...
+    /**
+     * @param string $code
+     *
+     * @throws ObjectNotFoundException
+     *
+     * @return ChannelInterface
+     */
+    protected function getChannelByCode($code)
+    {
+        $channel = $this->channelRepository->findOneByIdentifier($code);
```
willy-ahva: Or why didn't you get the channel code here directly with your getter ? Could avoid a param

willy-ahva:
GTM once comments fixed

src/Pim/Component/Connector/Reader/ProductReader.php
phaseinducer: I cast it in the constructor so not needed. ; )

src/Pim/Component/Connector/Reader/ProductReader.php
phaseinducer: This method needs a channel instance.

src/Pim/Component/Connector/Reader/ProductReader.php
phaseinducer: Because it increases the dependency between the methods and it's less readable imho.

src/Pim/Component/Connector/Reader/ProductReader.php
phaseinducer: Ok I'll do it in another PR.

src/Pim/Bundle/ConnectorBundle/Resources/config/batch_jobs/csv_connector.yml
```diff
@@ -140,7 +140,7 @@ connector:
                 export:
                     title:     pim_connector.jobs.csv_product_export.export.title
                     services:
-                        reader:    pim_base_connector.reader.doctrine.product
+                        reader:    pim_connector.reader.product
```
SamirBoulil: Yeah ;)

Issue #4356, opened 2016-04-22
Cannot remove selected value in product grid filter


s0ps:
Akeneo Version 1.5.1
After searching by Simple Select filter in product grid it is impossible to remove previously selected value.

a2xchip:
@s0ps Can you explain it more?

s0ps:
For example in your demo: http://demo.akeneo.com/
If you add Brand filter to filters and filter by Canon and then it is impossible remove the selected Canon option from filter.

a2xchip:
@s0ps See it! Selected options disappears after selection. Seems it is UI bug.

a2xchip:
@jjanvier @juliensnz I can reproduce it. As I understand the problem is in ```select2-choice-filter.js```.

First thoughts were that bug relates to dinamically added filters only. But if we do try the same on Group which is default it will act in the same way. But not the family filter. 

In case of family ```<ul class="select2-choices">``` is updated with  ```li>selected-value+li>input```
But in case with groups and brands not.

LaureBro:
Thanks a lot @a2xchip!

nidup:
cf PR comments.

PR #4350, opened 2016-04-20
PIM-5737: Fix behat on select2

momoss:
Fix behat on select2

| Q                                 | A
| --------------------------------- | ---
| Review and 2 GTM                  | Y

PR #4361, opened 2016-04-23
[Bugfix] issue4356

a2xchip:

Fixes filter select activity.

What does this Pull Request do? reference the related issue?
#4356

### DIFF
--- a/features/Context/DataGridContext.php
+++ b/features/Context/DataGridContext.php
@@ -751,6 +751,18 @@ class DataGridContext extends RawMinkContext implements PageObjectAwareInterface
     }
 
     /**
+     * @param string $filterName
+     * @param string $criteriaName
+     *
+     * @When /^I remove value "([^"]*)" on filter "([^"]*)"$/
+     * @And /^I remove value "([^"]*)" on filter "([^"]*)"$/
+     */
+    public function iRemoveValueOnFilter($selectionValue, $filterName)
+    {
+        $this->datagrid->removeFiltersSelection($filterName, $selectionValue, $this->getSession()->getDriver());
+    }
+
+    /**
      * @param string $row
      *
      * @When /^I click on the "([^"]*)" row$/
--- a/features/Context/Page/Base/Grid.php
+++ b/features/Context/Page/Base/Grid.php
@@ -612,9 +612,9 @@ class Grid extends Index
     /**
      * Removes selection from filter
      * 
-     * @param  string   $filterName
-     * @param  string   $selectionName
-     * @param  Driver   Interface|null $driver
+     * @param  string           $filterName
+     * @param  string           $selectionName
+     * @param  Interface|null   $driver
      * 
      * @return null
      */
@@ -633,10 +633,10 @@ class Grid extends Index
                 $selectionName
             )
         );
-        
-        sleep(1);
 
-        $this->getFilter($filterName)->find('css', 'button.filter-update')->click();
+        $updateBtn = $filter->find('css', 'button.filter-update');
+        $updateBtn->focus();
+        $updateBtn->click();
     }
 
     /**
--- a/features/Context/Page/Base/Grid.php
+++ b/features/Context/Page/Base/Grid.php
@@ -612,9 +612,9 @@ class Grid extends Index
     /**
      * Removes selection from filter
      * 
-     * @param string $filterName
-     * @param string $selectionName
-     * @param DriverInterface|null $driver
+     * @param  string   $filterName
+     * @param  string   $selectionName
+     * @param  Driver   Interface|null $driver
      * 
      * @return null
      */
@@ -633,10 +633,10 @@ class Grid extends Index
                 $selectionName
             )
         );
+        
         sleep(1);
 
-        
-        $filter->find('css', 'button.filter-update')->click();
+        $this->getFilter($filterName)->find('css', 'button.filter-update')->click();
     }
 
     /**
--- a/features/Context/Page/Base/Grid.php
+++ b/features/Context/Page/Base/Grid.php
@@ -610,6 +610,36 @@ class Grid extends Index
     }
 
     /**
+     * Removes selection from filter
+     * 
+     * @param string $filterName
+     * @param string $selectionName
+     * @param DriverInterface|null $driver
+     * 
+     * @return null
+     */
+    public function removeFiltersSelection($filterName, $selectionName, $driver = null)
+    {
+        if (!$driver || !$driver instanceof Selenium2Driver) {
+            throw new \InvalidArgumentException('Selenium2Driver is required to filter by a choice filter');
+        }
+
+        $filter = $this->getFilter($filterName);
+        $this->openFilter($filter);
+
+        $driver->executeScript(
+            sprintf(
+                "$('.select2-search-choice div:contains(\"%s\")').parent().children('a').click().trigger('click');",
+                $selectionName
+            )
+        );
+        sleep(1);
+
+        
+        $filter->find('css', 'button.filter-update')->click();
+    }
+
+    /**
      * Click on a filter in filter management list
      *
      * @param string $filterName
--- a/features/product/filtering/filter_products_per_group.feature
+++ b/features/product/filtering/filter_products_per_group.feature
@@ -45,3 +45,18 @@ Feature: Filter products
     When I filter by "Groups" with value "Empty"
     Then the grid should contain 0 element
     And I should not see products BOOK, MUG-1, MUG-2 and POSTIT
+
+  Scenario: Successfully update group filters
+    Given I am on the products page
+    When I filter by "Groups" with value "Postit"
+    Then the grid should contain 1 element
+    And I should see product POSTIT
+    And I filter by "Groups" with value "Mug"
+    Then the grid should contain 3 elements
+
+  Scenario: Successfully remove groups selected filter
+    Given I am on the products page
+    When I filter by "Groups" with value "Mug"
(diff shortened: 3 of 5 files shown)


features/Context/Page/Base/Grid.php
```diff
@@ -610,6 +610,36 @@ public function findView($viewLabel)
...
+        $filter = $this->getFilter($filterName);
+        $this->openFilter($filter);
+
+        $driver->executeScript(
+            sprintf(
+                "$('.select2-search-choice div:contains(\"%s\")').parent().children('a').click().trigger('click');",
+                $selectionName
+            )
+        );
+        sleep(1);
```
pierallard: We always avoid to "sleep" because it implies issues depending of who execute the test.
Please find a way with spin to check you're in an expected step (use `spin` method)

features/Context/Page/Base/Grid.php
```diff
@@ -610,6 +610,36 @@ public function findView($viewLabel)
...
+
+
```
pierallard: Double blank line :)

features/Context/Page/Base/Grid.php
```diff
@@ -610,6 +610,36 @@ public function findView($viewLabel)
     }
 
     /**
+     * Removes selection from filter
+     * 
+     * @param string $filterName
+     * @param string $selectionName
+     * @param DriverInterface|null $driver
```
pierallard: Please align variable names

features/product/filtering/filter_products_per_group.feature
```diff
@@ -45,3 +45,18 @@ Feature: Filter products
...
+    And I should see product POSTIT
+    And I filter by "Groups" with value "Mug"
+    Then the grid should contain 3 elements
+
+  Scenario: Successfully remove groups selected filter
+    Given I am on the products page
+    When I filter by "Groups" with value "Mug"
+    Then the grid should contain 2 elements
+    When I remove value "Mug" on filter "Groups"
+    Then the grid should contain 4 elements
```
pierallard: Tests! tests! tests! Love it :heart:

pierallard:
Hi @a2xchip ! Thanks a lot for this contribution ! Always a pleasure to review good-quality PR's including tests ;)
JS is not my cup of tea, so I let the others review in details this PR. I just annotate some little issues.

features/Context/Page/Base/Grid.php
```diff
@@ -610,6 +610,36 @@ public function findView($viewLabel)
...
+        $driver->executeScript(
+            sprintf(
+                "$('.select2-search-choice div:contains(\"%s\")').parent().children('a').click().trigger('click');",
+                $selectionName
+            )
+        );
+        sleep(1);
+
+        
+        $filter->find('css', 'button.filter-update')->click();
```
phaseinducer: You can try :
```
$button = $this->spin(function () use ($filter) {
    return $filter->find('css', '.filter-update');
}, 'Cannot find update filter button');
$button->click();
```
instead of `sleep(1);`

features/Context/Page/Base/Grid.php
```diff
@@ -610,6 +610,36 @@ public function findView($viewLabel)
...
+     * 
+     * @param string $filterName
+     * @param string $selectionName
+     * @param DriverInterface|null $driver
+     * 
+     * @return null
+     */
+    public function removeFiltersSelection($filterName, $selectionName, $driver = null)
+    {
+        if (!$driver || !$driver instanceof Selenium2Driver) {
```
phaseinducer: Don't know if it's really useful to pass the driver as parameter vs. calling `$this->getSession()->getDriver()` directly here.

features/Context/Page/Base/Grid.php
a2xchip: OK! @pierallard

a2xchip:
@pierallard As i know 1.5 is freezed for some reasons that is why i branched from master. 

@pierallard @phaseinducer I'am sorry about ```sleep``` and ```driver``` may be it looks like bad but I just mirrored that behat style coding i had seen in the same file. Please review following links, may be it is old file with overdue styles and there is necessity in refactoring it all. 

https://github.com/a2xchip/pim-community-dev/blob/issue4356/features/Context/Page/Base/Grid.php#L174

https://github.com/a2xchip/pim-community-dev/blob/issue4356/features/Context/Page/Base/Grid.php#L226

https://github.com/a2xchip/pim-community-dev/blob/issue4356/features/Context/Page/Base/Grid.php#L195

features/Context/Page/Base/Grid.php
a2xchip: @pierallard Answer in comments

a2xchip:
Hello @pierallard 

This message is out of issue #4356 

Javascript is not my best too. I mentioned that these files doing the same job. And "rest" one is written better than our target. All the difference is in the response format they receive from server.

https://github.com/a2xchip/pim-community-dev/blob/issue4356/src/Pim/Bundle/DataGridBundle/Resources/public/js/datafilter/filter/select2-rest-choice-filter.js

https://github.com/a2xchip/pim-community-dev/blob/issue4356/src/Pim/Bundle/DataGridBundle/Resources/public/js/datafilter/filter/select2-choice-filter.js

**The best will be to use ```select2-rest-choice-filter.js``` for all only changing server response in corresponding actions.**

I tried to find other places where we check removing of selected filters and it seems that such scenarios are missing. 

**It would be nice to add new scenarios: to ensure that we will not meet such bugs in other filter types interactions.**

Let me know what do you think about it. We may open new issues. ;-)

features/Context/Page/Base/Grid.php
a2xchip: I'll fix it soon.

features/Context/Page/Base/Grid.php
a2xchip: @phaseinducer As i remember ```sleep``` is not for **Update** button. I put sleep there because Update is pressed before javascript interaction come into force. I'll try to spin javascript execution but not sure if it will work for us.

pierallard:
@a2xchip 
Thanks for this complete answer.
- For the "sleep" issue, you're completely right, there is some sleep remaining, but we're working on it: See @momoss last PR https://github.com/akeneo/pim-community-dev/pull/4350/files#diff-9c335e887112e5db5a86eaffa7fd577fL225
- For the "driver" issue, you just looking at probably the most horrible function of the PIM :weary: It does not bother me to add it here, and I have to convice my colleagues of that.
- For the select2 rest/non-rest JS, I can not answer it, but I'm sure of one thing: the more we burn, the more the PIM is better. :fire: Maybe you should create an issue by copy/pasting this message, to avoid polluting this tiny PR.

features/Context/Page/Base/Grid.php
```diff
@@ -610,6 +610,36 @@ public function findView($viewLabel)
     }
 
     /**
+     * Removes selection from filter
+     * 
+     * @param  string   $filterName
+     * @param  string   $selectionName
+     * @param  Driver   Interface|null $driver
```
pierallard: I think you made a mistake here ;)

a2xchip: :-D

a2xchip: Autocompletion made... I'll update it anyway. Seems found decision to avoid sleep.

a2xchip:
@pierallard 

### About sleep. 

It is Selenium ChromeDriver's related issue as i understand after research. For example drupal experience it too: 

http://cgit.drupalcode.org/doobie/tree/features/bootstrap/FeatureContext.php?id=bc39260dfc65d9a8e1c344726c853cd8d829e84b#n1569 

https://github.com/seleniumhq/selenium-google-code-issue-archive/issues/2766

Workaround without sleep in case with update button is to focus on button. It works for me in chrome. May there will be yet another result in your jenkins environment. Let's see. If it will work we may create ```updateFilter($filter)``` and remove related ```sleep``` occurrences .

### About driver

I tried to reuse it from Grid.php following @phaseinducer recommendations but it fails, because it is not selenium driver. 

### About issues

Let me open some new issues: 

* select2
* sleep related to ```Element is not clickable```

:fire: :fire: ))))))

a2xchip:
@juliensnz @jjanvier @jmleroux  need reviews on this PR

features/Context/DataGridContext.php
```diff
@@ -751,6 +751,18 @@ public function iFilterBy($filterName, $value)
     }
 
     /**
+     * @param string $filterName
+     * @param string $criteriaName
+     *
+     * @When /^I remove value "([^"]*)" on filter "([^"]*)"$/
+     * @And /^I remove value "([^"]*)" on filter "([^"]*)"$/
```
ClementGautier: This line can be removed as behat will understand that you can use this step with `And` instead of `When` nativly

a2xchip: Thx! Never rescued to check it in real life ))

ClementGautier:
FYI our teams are actually trying to upgrade the select2 library for the upcoming 1.6 of the PIM and they are completely rewriting the behats about filter in datagrids so if I were you I would not invest much time on it ;)

nidup:
Hi @a2xchip 

This part as been widely re-worked since the pull request has been opened and I'm not even sure the issue is still present, I close the PR, feel free to re-open it after rebase if you want / the issue is still present.

Have a good day!

a2xchip:
Hello, @nidup! I think issue itself should be closed. But not sure about behat/selenium issues raised.

nidup:
Hello @a2xchip ! ok :)

PR #4362, opened 2016-04-23
Update translations from Crowdin

nono-akeneo:
Updated on 1.4-2016-04-23-09-00


a2xchip:
@nono-akeneo Is it translation key bug?

PR #4371, opened 2016-04-25
PIM-5742: Improve product saver

solivier:

- [x] Improved product saver to not use the double flush, bulk product saver can now flush bluk of products instead of doing it one by one.
- [x] Compleness scheduler is now performing a request directly via SQL
- [x] Measure perf

### DIFF
--- a/src/Pim/Bundle/CatalogBundle/Doctrine/Common/Saver/ProductSaver.php
+++ b/src/Pim/Bundle/CatalogBundle/Doctrine/Common/Saver/ProductSaver.php
@@ -74,9 +74,8 @@ class ProductSaver implements SaverInterface, BulkSaverInterface
 
         if (true === $options['flush']) {
             $this->objectManager->flush();
+            $this->completenessManager->generateMissingForProduct($product);
         }
-
-        $this->completenessManager->generateMissingForProduct($product);
 
         $this->eventDispatcher->dispatch(StorageEvents::POST_SAVE, new GenericEvent($product, $options));
     }
@@ -93,12 +92,17 @@ class ProductSaver implements SaverInterface, BulkSaverInterface
         $options = $this->optionsResolver->resolveSaveAllOptions($options);
         $this->eventDispatcher->dispatch(StorageEvents::PRE_SAVE_ALL, new GenericEvent($products, $options));
 
+        $itemOptions = $options;
+        $itemOptions['flush'] = false;
+
         foreach ($products as $product) {
-            $this->save($product, $options);
+            $this->save($product, $itemOptions);
         }
 
-        if (true === $options['flush']) {
-            $this->objectManager->flush();
+        $this->objectManager->flush();
+
+        foreach ($products as $product) {
+            $this->completenessManager->generateMissingForProduct($product);
         }
 
         $this->eventDispatcher->dispatch(StorageEvents::POST_SAVE_ALL, new GenericEvent($products, $options));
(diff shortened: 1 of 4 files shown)


src/Pim/Bundle/CatalogBundle/Doctrine/Common/Saver/ProductSaver.php
```diff
@@ -93,14 +92,22 @@ public function saveAll(array $products, array $options = [])
...
         }
 
         if (true === $options['flush']) {
             $this->objectManager->flush();
         }
 
+        foreach ($products as $product) {
+            $this->completenessManager->schedule($product);
+            $this->completenessManager->generateMissingForProduct($product);
+        }
```
nidup: i would move this in the `if (true === $options['flush']) {`because any way if flush = false is passed to a saveAll the completeness will be wrong (because based on database and not on memory objects).

src/Pim/Bundle/CatalogBundle/Doctrine/ORM/CompletenessGenerator.php
```diff
@@ -629,11 +629,17 @@ protected function getClassMetadata($className)
      */
     public function schedule(ProductInterface $product)
     {
-        foreach ($product->getCompletenesses() as $completeness) {
-            $this->manager->remove($completeness);
-        }
+        $sql = '
+            DELETE c FROM pim_catalog_completeness c
+              JOIN %product_table% p ON p.id = c.product_id
+             WHERE p.id = :product_id';
```
pierallard: Weird aligment

src/Pim/Bundle/CatalogBundle/Doctrine/Common/Saver/ProductSaver.php
solivier: yes I removed it

src/Pim/Bundle/CatalogBundle/Doctrine/ORM/CompletenessGenerator.php
```diff
@@ -629,11 +629,17 @@ protected function getClassMetadata($className)
      */
     public function schedule(ProductInterface $product)
     {
-        foreach ($product->getCompletenesses() as $completeness) {
-            $this->manager->remove($completeness);
-        }
+        $sql = '
+            DELETE c FROM pim_catalog_completeness c
```
jjanvier: don't hardcode the table

src/Pim/Bundle/CatalogBundle/Doctrine/MongoDBODM/Saver/ProductSaver.php
```diff
@@ -117,7 +117,7 @@ public function saveAll(array $products, array $options = [])
         if (count($updateDocs) > 0) {
             $this->updateDocuments($updateDocs);
         }
-        
+
```
pierallard: Mmmmh, interesting

pierallard:
GTM

grena:
GTM

src/Pim/Bundle/CatalogBundle/Doctrine/Common/Saver/ProductSaver.php
nei: @solivier @nidup I've overwritten this method to disable (generateMissingForProduct) because it was talking a massive amount of time to calculate the completeness on my dataset.. (400k products) 

The average in seconds was 12 products by second on the beginning and over time when the database was getting bigger the average dropped to 0.7 products by second.

Disabling that I can now run once when I have everything imported. I actually needed to update the mongdb timeout in order to be able to calculate that.. the completeness calculation is taking 20 min. 

Do you think I can have any problems on this approach?

PR #4379, opened 2016-04-25
[WIP] PIM-5685: Extract Business classes of Catalog Bundle (Part I - Managers)

aRn0D:
- [x] Remove `GroupManager` 
- [x] Remove `CategoryManager` 
- [x] Update changelog

| Migration script                  | no
| Tech Doc                          | no


spec/Akeneo/Bundle/ClassificationBundle/Doctrine/ORM/Repository/CategoryRepositorySpec.php
```diff
@@ -52,8 +52,18 @@ function let(
...
+    {
+        $this->shouldHaveType('Gedmo\Tree\Entity\Repository\NestedTreeRepository');
+    }
+
     function it_is_a_category_repository()
     {
         $this->shouldImplement('Akeneo\Component\Classification\Repository\CategoryRepositoryInterface');
     }
+
+    function it_is_a_identifiable_object_repository()
```
grena: > it_is_a**n**_identifiable(...)

src/Oro/Bundle/DataGridBundle/Provider/SystemAwareResolver.php
```diff
@@ -141,4 +133,35 @@ public function setContainer(ContainerInterface $container = null)
...
+
+    /**
+     * @param string $expression
+     * @param array  $optionsArguments
+     *
+     * @return mixed
+     */
+    protected function executeMethod($expression, $optionsArguments = [])
+    {
+        preg_match('#@([\w\._]+)->([\w\._]+)(\((.*)\))*#', $expression, $matches);
```
grena: Sad to rewrite this complex regex here (already in the switch in the other method).
No way to have a $var for this? (property or idk)

src/Pim/Bundle/CatalogBundle/Doctrine/ORM/Helper/ResultParser.php
```diff
@@ -0,0 +1,79 @@
+<?php
+
+namespace Pim\Bundle\CatalogBundle\Doctrine\ORM\Helper;
+
+/**
+ * Helper for doctrine
```
grena: Yes but what does it parse into what?

src/Pim/Bundle/CatalogBundle/Doctrine/ORM/Helper/ResultParser.php
```diff
@@ -0,0 +1,79 @@
...
+     *     ['id' => 10, 'label' => 'group fr', 'code' => 'group_code', 'locale' => 'fr_FR'],
+     *     ['id' => 10, 'label' => 'group en', 'code' => 'group_code', 'locale' => 'en_US'],
+     *     ['id' => 11, 'label' => null, 'code' => 'group_other_code', 'locale' => 'fr_FR']
+     * ]
+     *
+     * Out :
+     * [
+     *     10 => 'group en',
+    *      11 => '[group_other_code]',
+     * ]
```
grena: :+1:

src/Oro/Bundle/DataGridBundle/Provider/SystemAwareResolver.php
aRn0D: We can add an extra parameter but I think it is the same...

src/Pim/Bundle/CatalogBundle/Doctrine/ORM/Helper/ResultParser.php
```diff
@@ -0,0 +1,79 @@
...
+    /**
+     * Extract ids from a result set of doctrine query (array hydration)
+     *
+     * @param array $flatIds
+     *
+     * @return array
+     */
+    public static function parseIds(array $flatIds)
+    {
+        $ids = array_reduce($flatIds, function ($carry, $item) {
```
grena: Imho you could replace this by a `foreach` to ease readability. It's a kind of hack to use `array_reduce` to build an array :p

src/Pim/Bundle/CatalogBundle/Doctrine/ORM/Helper/ResultParser.php
```diff
@@ -0,0 +1,79 @@
...
+
+        $extractLabel = function (array $item) use ($locale) {
+            $label = isset($item['labels'][$locale]) ? $item['labels'][$locale] : $item['code'];
+
+            return $label;
+        };
+
+        $flatTranslations = array_map($extractLabel, array_reduce($flatTranslations, $formattedTranslations, []));
+
+        return $flatTranslations;
```
grena: This method is hard to read.
Imho it could be simpler with `foreach`s.

src/Pim/Bundle/EnrichBundle/Resources/config/datagrid/association_group.yml
```diff
@@ -65,4 +65,4 @@ datagrid:
                     options:
                         field_options:
                             multiple: true
-                            choices:  '@pim_catalog.manager.group->getTypeChoices'
+                            choices:  '@pim_catalog.repository.group_type->findTypes(false, @pim_user.context.user->getCurrentLocaleCode())'
```
grena: :fearful: Reading this in a .yml is creepy.. Not way to have another way? Like calling a single method which call other methods inside?

src/Pim/Component/Catalog/Repository/GroupTypeRepositoryInterface.php
```diff
@@ -30,4 +31,23 @@ public function createDatagridQueryBuilder();
...
+     * @param string $locale
+     *
+     * @return array
+     */
+    public function findTypes($isVariant, $locale);
+
+    /**
+     * Find axis technical identifiers
+     *
+     * @param bool   $isVariant
```
grena: extra space

grena:
Nice :fire: !
Some things I don't agree with, but don't know if related to this PR.
GTM after some fixes

src/Pim/Bundle/CatalogBundle/Doctrine/ORM/Helper/ResultParser.php
aRn0D: It is just a question of point of view, personally I think it is readable and clear. I just use the API provided by PHP :dancers:  but if the team prefer a foreach. I will change that.

src/Pim/Bundle/CatalogBundle/Doctrine/ORM/Helper/ResultParser.php
aRn0D: idem than my previous comment. My closures are pure functions and well named.

src/Pim/Bundle/EnrichBundle/Resources/config/datagrid/association_group.yml
aRn0D: Yes, we can rewrite the datagrid :D! I did not want to add an option dependency to a repository which should be only use a single method. The problem is not the repository but the way that the datagrid is built.

grena: Yup, just wanted to be sure. thanks

src/Akeneo/Component/Classification/Repository/CategoryRepositoryInterface.php
```diff
@@ -8,6 +8,7 @@
 use Doctrine\Common\Collections\Collection;
 use Doctrine\Common\Persistence\ObjectRepository;
 use Gedmo\Tree\RepositoryInterface as TreeRepositoryInterface;
+use Pim\Bundle\CatalogBundle\Doctrine\ORM\Helper\ResultParser;
```
nidup: useless use here

src/Oro/Bundle/DataGridBundle/Provider/SystemAwareResolver.php
nidup: i would move these patterns in constants or in protected methods getBlaBlaPattern(), imvho not a good idea to duplicate a such not maintainable pattern

src/Pim/Component/Catalog/Repository/GroupTypeRepositoryInterface.php
```diff
@@ -4,6 +4,7 @@
 
 use Akeneo\Component\StorageUtils\Repository\IdentifiableObjectRepositoryInterface;
 use Doctrine\Common\Persistence\ObjectRepository;
+use Pim\Bundle\CatalogBundle\Doctrine\ORM\Helper\ResultParser;
```
nidup: useless use

src/Pim/Bundle/CatalogBundle/Doctrine/ORM/Helper/ResultParser.php
```diff
@@ -0,0 +1,79 @@
+<?php
+
+namespace Pim\Bundle\CatalogBundle\Doctrine\ORM\Helper;
```
nidup: Not a big fan of the "Helper" folder which will quickly contains a whole set of non related classes, if it's used by only one class, we can put it near the class for instance in Doctrine\ORM\Repository? or directly in Doctrine/ORM?

src/Pim/Component/Catalog/Repository/AttributeRepositoryInterface.php
```diff
@@ -4,6 +4,7 @@
 
 use Akeneo\Component\StorageUtils\Repository\IdentifiableObjectRepositoryInterface;
 use Doctrine\Common\Persistence\ObjectRepository;
+use Doctrine\ORM\QueryBuilder;
```
nidup: we should not rely on Doctrine\ORM\QueryBuilder in a AttributeRepositoryInterface (Doctrine\Common is acceptable)
dunno if it's the scope or not of your PR but it makes appears this design flaw (to be checked with @jjanvier there is maybe another TIP card to fix the use of Doctrine\ORM in repository interfaces)

nidup:
Aaaaah static methods ... you should use some private and final in this PR to be sure to run a longest debate :trollface: (kidding)

More seriously, we historically avoid to write and use static methods in the project ... and this is absolutely not an argument because "historically" is the worst argument ever (the famous "we always made this way", also named "grep -r 'public static function' ./src/").

So I'm glad to have the opportunity to re-discuss about the use of static methods :wink: 

Personally, I always avoid to use such static methods because they create a very strong coupling (yeah static...) which totally forbid to replace the original implementation except by changing the method content itself (ok possibility to override all the callers, welcome in hell).

It's an advantage because it ensures that no unexpected override/change will occur in a project.

It's also a very strong limitation for the exact same reason.

I do agree that the function is pure from FP point of view and I'm also pretty sure that we'll don't have any issue with this.

I guess also that we have the same readings about when using static methods or not, for instance, http://verraes.net/2014/06/when-to-use-static-methods-in-php.

My concern is more about balancing the pros and cons in our own context, aka, an open source project, project being always customized by integrators (in several ways we even don't imagine) before to be used by customers, with a paying Service Level Agreement (SLA) support for the Enterprise Edition users and a best effort support for the Community Edition users.

Let's just discuss about the maintenance concern.

If a SLA bug is raised concerning the content of this method, we'll don't have any other choice than fix it and release a patch in the contractual delay, we'll don't be able to provide a hot fix patch only for the project (a simple override of the original class + DI declaration to fix the issue) before to provide the real fix for everybody.

If a community bug is raised, we'll obviously do our best to fix and release it quickly but a guy working on its own PIM project will not be able to do so in its project without crazy override of all "callers".

So, from my very own point of view, in our context, I do prefer use very small objects with instance methods and inject even very small services in the DI (in the context of a final project without extra integration, I would not have the same opinion and I would prefer use more static method to avoid over-engineering in this kind of cases).

I'm available to discuss about this on slack.

src/Pim/Bundle/CatalogBundle/Doctrine/ORM/Repository/GroupTypeRepository.php
```diff
@@ -84,4 +85,40 @@ public function getIdentifierProperties()
...
+    public function findTypeIds($isVariant)
+    {
+        $query = $this->createQueryBuilder('g')
+            ->select('g.id')
+            ->leftJoin('g.translations', 't')
+            ->andWhere('g.variant = :variant')
+            ->setParameter('variant', $isVariant)
+            ->getQuery();
+
+        $ids = ResultParser::parseIds($query->getArrayResult());
```
nidup: There is a trick,
```
        $qb = $this->_em->createQueryBuilder()
            ->select('g.id')
            ->from($this->_entityName, 'g', 'g.id'); // the third param allow to index by this fields
            //...
         return array_keys($query->getArrayResult());
```

and :tada: !
        if (is_array($codes) && !empty($codes)) {
            $qb->andWhere("att.code IN (:codes)");
            $qb->setParameter('codes', $codes);
        }

        if (is_array($groupIds) && !empty($groupIds)) {
            $qb->andWhere("att.group IN (:groupIds)");
            $qb->setParameter('groupIds', $groupIds);
        } elseif (is_array($groupIds)) {
            return [];
        }

        $qb->andWhere('att.useableAsGridFilter = :useableInGrid');
        $qb->setParameter('useableInGrid', 1);

        $result = $qb->getQuery()->execute([], AbstractQuery::HYDRATE_ARRAY);

        return array_keys($result);

aRn0D: good catch I will remove it!

src/Pim/Bundle/CatalogBundle/Doctrine/ORM/Helper/ResultParser.php
a2xchip: @aRn0D :smile: :smile: :smile:

nidup:
I double think about this PR, here, we're adding some methods in our business repositories to return an associative array of object id to label (or code if there is not label defined).

And I wonder why we need to provide this kind of very specific structure from our business repositories, what are the real use case? Is it business stuff or not? Do we do this for other entities? If yes, why we don't use the same parsing logic?

I checked and in fact these methods are only used to provide data to the select component of the UI part, and we do this for all entities at the end, cf,
```
nico@nico-laptop:~/git/pim-ce-16-orm$ grep -r '@' ./src/Pim/Bundle/EnrichBundle/Resources/config/datagrid/
./src/Pim/Bundle/EnrichBundle/Resources/config/datagrid/variant_group.yml:                            choices:  '@pim_catalog.manager.group->getAvailableAxisChoices'
./src/Pim/Bundle/EnrichBundle/Resources/config/datagrid/product_variant_group.yml:                            choices: '@pim_catalog.repository.channel->getLabelsIndexedByCode'
./src/Pim/Bundle/EnrichBundle/Resources/config/datagrid/product_group.yml:                            choices: '@pim_catalog.repository.channel->getLabelsIndexedByCode'
./src/Pim/Bundle/EnrichBundle/Resources/config/datagrid/product.yml:                            choices: '@pim_catalog.repository.channel->getLabelsIndexedByCode'
./src/Pim/Bundle/EnrichBundle/Resources/config/datagrid/family.yml:                            choices:  '@pim_catalog.repository.attribute->getAvailableAttributesAsLabelChoice'
./src/Pim/Bundle/EnrichBundle/Resources/config/datagrid/attribute.yml:                            choices:  '@pim_catalog.registry.attribute_type->getSortedAliases'
./src/Pim/Bundle/EnrichBundle/Resources/config/datagrid/attribute.yml:                            choices:  '@pim_catalog.repository.attribute_group->getAttributeGroupChoices'
./src/Pim/Bundle/EnrichBundle/Resources/config/datagrid/association_product.yml:                            choices: '@pim_catalog.repository.channel->getLabelsIndexedByCode'
./src/Pim/Bundle/EnrichBundle/Resources/config/datagrid/channel.yml:                            choices:  '@pim_catalog.manager.category->getTreeChoices'
./src/Pim/Bundle/EnrichBundle/Resources/config/datagrid/group.yml:                            choices:  '@pim_catalog.manager.group->getTypeChoices'
./src/Pim/Bundle/EnrichBundle/Resources/config/datagrid/association_group.yml:                            choices:  '@pim_catalog.manager.group->getTypeChoices'
```
Most of these methods are named differently but most of these return the same structured data only used for the UI components (grid, filters).

This logic is sometimes implemented with php code, sometimes we do a  ```addSelect('COALESCE(ft.label, CONCAT(\'[\', f.code, \']\')) as label') ``` **not need of any parser if we use this trick everywhere**.

We should be more consistent with all these methods.

If the concern is to provide this structured array data for all entities we should use some dedicated services to do so and not put this code in our business repositories, these services should follow a common interface, for instance, "ChoicesProviderInterface" declaring a "getChoices" (maybe possible to DRY some code for all the simple entity with a SimpleProvider configured with entityName etc through the DI).

Where to put this interface and these implementations, Catalog vs Enrich?
I maybe go to far with this PR, it's maybe no the scope of this one but I've the feeling that we're missing something by just moving couple of methods from managers to repositories without looking around.
At least for matter of consistency I would use the "coalesce" stuff everywhere.

a2xchip:
+1 for consistency.

aRn0D:
I think you're right about moving those methods to business repository, this is not the right choice... We can create news ones to retrieve data for the UI, they should be in EnrichBundle too.

`SimpleProvider` is be a good idea! I found repositories in this bundle which implement `Pim\Bundle\EnrichBundle\Form\DataTransformer\ChoicesProviderInterface` (not sure about the namespace). This interface is use every `catalog component/bundle`, `filter bundle`, and `enrich bundle`, this is not good...

I am not sure that `addSelect('COALESCE(ft.label, CONCAT(\'[\', f.code, \']\')) as label')` is as simple as that because data looks like in database: 

```
1 |  label en           | code
2 |  label fr             | code               # different rows
3 |  other label en  | other_code
```

It means that we need to do a sub request but it could be done.

So @nidup what is the boundary context of this card ?

src/Pim/Bundle/CatalogBundle/Doctrine/ORM/Helper/ResultParser.php
```diff
@@ -0,0 +1,72 @@
...
+     * [
+     *     ['id' => 10, 'label' => 'group fr', 'code' => 'group_code', 'locale' => 'fr_FR'],
+     *     ['id' => 10, 'label' => 'group en', 'code' => 'group_code', 'locale' => 'en_US'],
+     *     ['id' => 11, 'label' => null, 'code' => 'group_other_code', 'locale' => 'fr_FR']
+     * ]
+     *
+     * Out :
+     * [
+     *     10 => 'group en',
+    *      11 => '[group_other_code]',
```
willy-ahva: indentation

src/Pim/Bundle/CatalogBundle/Doctrine/ORM/Helper/ResultParser.php
```diff
@@ -0,0 +1,72 @@
+<?php
+
+namespace Pim\Bundle\CatalogBundle\Doctrine\ORM\Helper;
+
+class ResultParser
```
willy-ahva: You should add a baby description

src/Pim/Bundle/CatalogBundle/Doctrine/ORM/Helper/ResultParser.php
willy-ahva: you should have `10 => 'group fr'` no ? Or I didn't understand what is it about.

nidup:
Coalesce works only if you pass a locale as argument.

About the boundary of this card, I would love be able to sanitize this part by extracting this from our business repository (we historically put everything here because we used a single repository per entity).

If the team needs helps on other stuff I would change the minimal stuff, avoid to introduce new concept as parser (even by duplicating a bit of code in repository) and assume to only move code from manager to repository. I you have the time, extract the code in new classes and services and explain the BC break.

Take care because we maybe use different format for the frontend id to label, code to label, code to label with paginate stuff etc so don't hesitate to be very precise in the contract, maybe "IdToLabelProviderInterface" to implement only the format you want to change without pulling the whole refacto of all the weird data format we use for the UI.

src/Pim/Bundle/CatalogBundle/Doctrine/ORM/Helper/ResultParser.php
willy-ahva: Agree with pipou, this method is absolutely not maintainable ... It should be simpler.

src/Pim/Bundle/CatalogBundle/Doctrine/ORM/Helper/ResultParser.php
a2xchip: The bigest confusion for me is flatTranslations/in => flatTranslation/out naming

src/Pim/Bundle/CatalogBundle/Doctrine/ORM/Helper/ResultParser.php
willy-ahva: Agree with pipou

spec/Akeneo/Bundle/ClassificationBundle/Doctrine/ORM/Repository/CategoryRepositorySpec.php
```diff
@@ -16,7 +16,7 @@
 
 class CategoryRepositorySpec extends ObjectBehavior
 {
-    function let(
+        function let(
```
willy-ahva: indentation

PR #4384, opened 2016-04-26
(Proposition) PIM-5364: Take the full step name for warning base title

grena:
## Explanation
Giving a job:

```yml
update_products:
    title: pim.jobs.update_products.title
    steps:
        perform:
            title: pim.jobs.update_products.steps.perform.title
            (...)
```
Currently, if there is a warning during a step, the warning translation key is based on the first word of the step translation key. So if a warning is raised for the step **perform**, it will be `PIM.STEPS.EXCLUDED.TITLE`. 

process tracker show job

This PR aims to keep the step name to build a consistent warning title. Taking the same example, the warning title will be `PIM.JOBS.UPDATE_PRODUCTS.STEPS.PERFORM.STEPS.EXCLUDED.TITLE`.

## DOD

### DIFF
--- a/spec/Akeneo/Component/Batch/Model/StepExecutionSpec.php
+++ b/spec/Akeneo/Component/Batch/Model/StepExecutionSpec.php
@@ -11,7 +11,7 @@ class StepExecutionSpec extends ObjectBehavior
 {
     function let(JobExecution $jobExecution)
     {
-        $this->beConstructedWith('myStepName', $jobExecution);
+        $this->beConstructedWith('my.step.title', $jobExecution);
     }
 
     function it_is_properly_instanciated()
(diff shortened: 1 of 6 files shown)


nidup:
Nice you spotted this issue!

All these "titles" in the Batch component are really messy, in fact, we use the 'name' property which come from spring batch and we use it as a translation key for UI. (in batch_jobs.yml we configure with a title field but the DI fulfill the name property of Jobs and Steps).

We should store proper name like,
```
update_products: # <- doing this in fact, we can see that this is already a better name for the step
    name: update_products
    steps:
        perform:  # <- doing this in fact, we can see that this is already a better name for the step
            name: perform
            (...)
```

Issue doing this is BC Break etc, to take in consideration but it's not too hard to be backward compatible by keeping the title field in the DI but not using it and using the 'root' field, here 'update_products' or 'perform' (or adding an optional 'name' field).

I noted this as a task of this already too big PR https://github.com/akeneo/pim-community-dev/pull/4326 Re-working this in // and merge the title re-work on master should be great!

The fact to build translation keys from the name of a Job, Step, StepElement should be the responsibility of the UI part which use it at the end.

In ImportExportBundle, we do the following,
```
'jobExecution' => $this->serializer->normalize($jobExecution, 'json', $context),
```

The JobExecutionNormalizer and StepExecutionNormalizer should ask to a new dedicated service to build the "conventioned" translation key from the Job name, Step name or StepElement name, for instance.

Maybe a bit large re-work compared to the scope of the original question ? :smile:

grena:
Oh thank @nidup , I've just seen you comment!

Indeed, my PR is a mix of concern, the `StepExecution` should not be aware of any UI "title" thing.
If I have enough time at the end of my card (and I should), I'll try to push a PR to handle this stuff!

PR #4410, opened 2016-05-02
[Support-branch] PIM-5728, PIM-5762, PIM-5666, PIM-5738, PIM-5727, PIM-5754, PIM-5763, PIM-5622, PIM-5756

pierallard:
Commit fixing
- PIM-5728
- PIM-5762
- PIM-5666
- PIM-5727
- PIM-5754
- PIM-5763
- PIM-5622
- PIM-5756
- PIM-5655

All of these are GTMed on their own PR's.

PR #4413, opened 2016-05-03
PIM-5756: Fix empty min and max numbers on attributes import

pierallard:
<3 Thanks for taking the time to contribute! You're awesome! <3

De nada

When you import attributes and set min_number and max_number as "empty" column, these properties are set to 0.
This PR fix it by checking before assign variable if column is empty or not (and fix it too for some other fields)

### DIFF
--- a/src/Pim/Component/Connector/ArrayConverter/Flat/AttributeStandardConverter.php
+++ b/src/Pim/Component/Connector/ArrayConverter/Flat/AttributeStandardConverter.php
@@ -103,11 +103,11 @@ class AttributeStandardConverter implements StandardArrayConverterInterface
             case 'number_min':
             case 'number_max':
             case 'max_file_size':
-                $convertedItem[$field] = (float) $data;
+                $convertedItem[$field] = ('' === $data) ? null : (float) $data;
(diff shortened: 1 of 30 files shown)


ClementGautier:
Add a unit test then GTM

src/Pim/Component/Connector/ArrayConverter/Flat/AttributeStandardConverter.php
```diff
@@ -103,11 +103,11 @@ protected function convertFields($field, $booleanFields, $data, $convertedItem)
             case 'number_min':
             case 'number_max':
             case 'max_file_size':
-                $convertedItem[$field] = (float) $data;
+                $convertedItem[$field] = ('' === $data) ? null : (float) $data;
                 break;
             case 'sort_order':
```
momoss: 'sort_order' cannot be null, look at the DB config, you may return an int value :smiley:

momoss:
GTM

pierallard:
Merged in https://github.com/akeneo/pim-community-dev/pull/4410

PR #4421, opened 2016-05-06
TIP-395: Move the specs to their Bundle/Component

aRn0D:
Before :
phpspec-master

After :

after


phpspec.yml
```diff
@@ -1,2 +1,188 @@
...
+        spec_path: src/Pim/Bundle/VersioningBundle
+        src_path: src/Pim/Bundle/VersioningBundle
+    WebServiceBundle:
+        namespace: Pim\Bundle\WebServiceBundle
+        psr4_prefix: Pim\Bundle\WebServiceBundle
+        spec_path: src/Pim/Bundle/WebServiceBundle
+        src_path: src/Pim/Bundle/WebServiceBundle
+
+    ## Oro bundle
+    WebServiceBundle:
```
nidup: Copy / paste issue, FilterBundle the pb is that we have a Pim and a Oro FilterBundle we could prefix

phpspec.yml
```diff
@@ -1,2 +1,188 @@
...
+        psr4_prefix: Pim\Component\Connector
+        spec_path: src/Pim/Component/Connector
+        src_path: src/Pim/Component/Connector
+    Catalog:
+        namespace: Pim\Component\Catalog
+        psr4_prefix: Pim\Component\Catalog
+        spec_path: src/Pim/Component/Catalog
+        src_path: src/Pim/Component/Catalog
+
+    ## Akeneo Bundle
```
nidup: PIM bundles

phpspec.yml.dist
```diff
@@ -0,0 +1,8 @@
+suites:
+    ReferenceData:
```
nidup: Not sure about keeping this file I understand that it allows to define your own suite as dev and not version it, we could at least add a comment inside the file to explain that. Imho it's overkill and we can drop it everybody runs the whole suite :)

src/Pim/Bundle/BaseConnectorBundle/spec/Reader/File/YamlReaderSpec.php
```diff
@@ -81,7 +81,7 @@ function it_reads_entities_from_a_yml_file_one_by_one_incrementing_summary_info_
         $stepExecution->incrementSummaryInfo('read_lines')->shouldBeCalledTimes(3);
 
         $this->setFilePath(
-            realpath(__DIR__ . '/../../../../../../features/Context/fixtures/fake_products_with_code.yml')
+            realpath(__DIR__ . '/../../../../../../../features/Context/fixtures/fake_products_with_code.yml')
```
nidup: I would duplicate the test file in the spec folder because in case if subtree split you can't run it directly in the readonly repo and also because it's weird to depends on a behat test file

nidup:
Baby comments to fix, excluded specs from PHP cs fixer then GTM

phpspec.yml.dist
aRn0D: it can be usefull for extension !

nidup: Ok add a comment inside to explain and ok for me

phpspec.yml
```diff
@@ -1,2 +1,188 @@
+suites:
```
grena: you should not track this file :100:

grena:
GTM

PR #4436, opened 2016-05-10
PIM-5714: Enforce coupling rules on PIM Catalog component - Part I - Reference Data

damien-carcel:
**Description**

Reference Data is considered as extended feature of the PIM. In this current form, we should be able to toggle or not this feature (this should work already actually). It also means that the code related to reference data should remain in the reference data folders. We have a few classes that does not respect this rule.

To check if the rules are enforced, use the following command:
```bash
bin/php-coupling-detector detect src/Pim/Component/Catalog --config-file=.php_cd_transitional.php
```

This PR intends to fix reference data violations (**24** violations before, **21** in this PR)

**Definition Of Done**

| Migration script                  | Yes

### DIFF
--- a/src/Pim/Bundle/CatalogBundle/Resources/config/validators.yml
+++ b/src/Pim/Bundle/CatalogBundle/Resources/config/validators.yml
@@ -16,6 +16,7 @@ parameters:
     pim_catalog.validator.constraint.scopable_value.class:                 Pim\Component\Catalog\Validator\Constraints\ScopableValueValidator
     pim_catalog.validator.constraint.localizable_value.class:              Pim\Component\Catalog\Validator\Constraints\LocalizableValueValidator
     pim_catalog.validator.constraint.attribute_type_for_option.class:      Pim\Component\Catalog\Validator\Constraints\AttributeTypeForOptionValidator
+    pim_catalog.validator.constraint.is_reference_data_configured.class:   Pim\Component\Catalog\Validator\Constraints\IsReferenceDataConfiguredValidator
     pim_catalog.validator.constraint_guesser.chained.class:                Pim\Component\Catalog\Validator\ChainedAttributeConstraintGuesser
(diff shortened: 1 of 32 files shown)


src/Pim/Bundle/CatalogBundle/Resources/config/validators.yml
```diff
@@ -145,6 +146,14 @@ services:
         tags:
             - { name: validator.constraint_validator, alias: pim_attribute_type_for_option_validator }
 
+    pim_catalog.validator.constraint.is_reference_data_configured:
```
jjanvier: should go in reference data :)

src/Pim/Component/Catalog/Validator/Constraints/IsReferenceDataConfigured.php
```diff
@@ -0,0 +1,34 @@
+<?php
+
+namespace Pim\Component\Catalog\Validator\Constraints;
```
jjanvier: should go in reference Data component :)

src/Pim/Component/Catalog/Validator/Constraints/IsReferenceDataConfiguredValidator.php
```diff
@@ -0,0 +1,53 @@
+<?php
+
+namespace Pim\Component\Catalog\Validator\Constraints;
```
jjanvier: should go in reference Data component :)

jjanvier:
GTM once fixed. Very good!

src/Pim/Bundle/CatalogBundle/Resources/config/validators.yml
skeleton: Dunno, as this validator is loaded here: https://github.com/akeneo/pim-community-dev/pull/4436/files#diff-5c459100e5022b7cddcbd5c85cf88a58R5 
What's append if reference data are disabled in appKernel ?

skeleton:
1 question about validator. maybe you should disable ReferenceDataBundle appKernel and see if application still works.
If it's work, GTM :+1:

PR #4446, opened 2016-05-12
PIM-5428: Export product builder - export products according to their status

jjanvier:

This PR replaces https://github.com/akeneo/pim-community-dev/pull/4429. It's a simple and dumb implementation of the product export builder without changing the product reader behavior. As a matter of fact, this is not that easy to build something solid at the moment as:
* @nidup is currently rewriting the batch foundations https://github.com/akeneo/pim-community-dev/pull/4326
* the previous PR has some design drawbacks https://github.com/akeneo/pim-community-dev/pull/4429
* last but not least, it's kind of silly to write a brand new system that transforms UI options to PQB filters without knowing where we go. We need to do something usable both by the datagrid and the export builder. 

That's why we decided to keep it really really simple at the moment even knowing it not perfect. But it's better to stick with protected methods we already have (without changing the product reader behavior, it's purely internal), rather than introducing a whole bunch of public stuff without knowing the impacts and what we want and can achieve in the time we have.

A new folder for Behats dedicated to the export product builder. I created a dedicated simple family in the background of the Behat to make them much easy to read, and not dependant of other product filters or future changes.

### DIFF
--- a/features/export/product-export-builder/export_products_status.feature
+++ b/features/export/product-export-builder/export_products_status.feature
@@ -1,0 +1,56 @@
+@javascript
+Feature: Export products according to their status
+  In order to use the enriched product data
+  As a product manager
+  I need to be able to export the product according their status
+
+  Background:
+    Given an "footwear" catalog configuration
+    And the following family:
+      | code    | requirements-mobile |
+      | rangers | sku, name           |
+    And the following products:
+      | sku      | enabled | family  | categories        | name-en_US    |
+      | SNKRS-1B | 1       | rangers | summer_collection | Black rangers |
+      | SNKRS-1R | 0       | rangers | summer_collection | Black rangers |
+    And I am logged in as "Julia"
+
+  Scenario: Export only enabled products
+    Given the following job "csv_footwear_product_export" configuration:
+      | filePath | %tmp%/product_export/product_export.csv |
+      | enabled  | enabled                                 |
+    When I am on the "csv_footwear_product_export" export job page
+    And I launch the export job
+    And I wait for the "csv_footwear_product_export" job to finish
+    Then exported file of "csv_footwear_product_export" should contain:
+    """
+    sku;categories;enabled;family;groups;name-en_US
+    SNKRS-1B;summer_collection;1;rangers;;Black rangers
+    """
+
+  Scenario: Export only disabled products
+    Given the following job "csv_footwear_product_export" configuration:
+      | filePath | %tmp%/product_export/product_export.csv |
+      | enabled  | disabled                                |
+    When I am on the "csv_footwear_product_export" export job page
+    And I launch the export job
+    And I wait for the "csv_footwear_product_export" job to finish
+    Then exported file of "csv_footwear_product_export" should contain:
+    """
+    sku;categories;enabled;family;groups;name-en_US
+    SNKRS-1R;summer_collection;0;rangers;;Black rangers
+    """
+
+  Scenario: Export products no matter their status
+    Given the following job "csv_footwear_product_export" configuration:
+      | filePath | %tmp%/product_export/product_export.csv |
+      | enabled  | all                                     |
+    When I am on the "csv_footwear_product_export" export job page
+    And I launch the export job
+    And I wait for the "csv_footwear_product_export" job to finish
+    Then exported file of "csv_footwear_product_export" should contain:
+    """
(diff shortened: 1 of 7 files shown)


src/Pim/Bundle/ImportExportBundle/Resources/views/JobProfile/Tab/job_content.html.twig
```diff
@@ -1,2 +1,14 @@
 <div id="{{ viewElement.alias|replace({' ': '-', '.': '-'})|lower }}" class="tab-pane buffer-top">
+    {% set contentSettings %}
+        {% for step in form.job.steps %}
```
jjanvier: much be cleaned with @nidup's PR (much less foreach...)

src/Pim/Bundle/ImportExportBundle/Resources/views/JobProfile/Tab/property_edit.html.twig
```diff
@@ -7,8 +7,13 @@
 
     {% set globalSettings %}
         {% for step in form.job.steps %}
-            {% for child in step.children %}
-                {{ form_widget(child) }}
+            {% for stepElement in step.children %}
+                {% for field in stepElement.children %}
```
jjanvier: much be cleaned with @nidup's PR (much less foreach...)

phaseinducer:
GTM :+1:

features/export/product-export-builder/export_products_status.feature
```diff
@@ -0,0 +1,56 @@
+@javascript
+Feature: Export products according to their status
+  In order to use the enriched product data
+  As a product manager
+  I need to be able to export the product according their status
```
SamirBoulil: according to their

features/export/product-export-builder/export_products_status.feature
```diff
@@ -0,0 +1,56 @@
...
+    When I am on the "csv_footwear_product_export" export job page
+    And I launch the export job
+    And I wait for the "csv_footwear_product_export" job to finish
+    Then exported file of "csv_footwear_product_export" should contain:
+    """
+    sku;categories;enabled;family;groups;name-en_US
+    SNKRS-1R;summer_collection;0;rangers;;Black rangers
+    """
+
+  Scenario: Export products no matter their status
```
SamirBoulil: statuses

src/Pim/Bundle/ImportExportBundle/Resources/views/JobProfile/Tab/property_edit.html.twig
```diff
@@ -7,8 +7,12 @@
...
+                    {% if field.vars.attr['data-tab'] is not defined or 'content' != field.vars.attr['data-tab'] %}
```
SamirBoulil: Should we identify the property tab as well ? 
like `or 'property' == fields.vars.attr['data-tab']`

I mean if tomorrow we have a new tab we want to add custom fields into, we will need to change this to : 
`or ('content' != field.vars.attr['data-tab'] and 'new-tab' != field.vars.attr['data-tab']` ?

SamirBoulil:
GTM after answers ;) well done!

features/export/product-export-builder/export_products_status.feature
jjanvier: @SamirBoulil 
I need to be able to export the product*s* according their status

?

SamirBoulil: "according *to* their statuses" I'd say then

src/Pim/Bundle/ImportExportBundle/Resources/views/JobProfile/Tab/property_edit.html.twig
jjanvier: you're right, done

jjanvier: @SamirBoulil the problem is that it forces us and the integrators to change all the configurations to add the tab
maybe here we should just do 

````
 {% if field.vars.attr['data-tab'] is not defined %}
````

SamirBoulil: Agreed. a default behavior is fine :)

PR #4457, opened 2016-05-13
Remove template annotation from import/export controllers

jmleroux:
Without the `@Template` annotation, we're no longer constrained to override this bundle when creating a new profile in an external bundle.

| Added Behats                      |n
| Changelog updated                 |todo
| Tech Doc                          |todo


a2xchip:
Looks good)

jmleroux:
This PR cannot work because of inheritance inside Twig templates (not only the templates themselves).
I will try to solve this in #4461

jmleroux:
I still leave it open because the basic purpose of this PR is always relevant imho.

jmleroux:
While #4461 was not merged, i still think that removing this dependency is a good thing for future extensibility.

PR #4468, opened 2016-05-16
TIP-423: Reset permissions between each scenario

phaseinducer:
**Why this PR?**
Today, a part of our CI instability comes from the fact that permissions system is stateful between scenarios. It means that whenever a scenario modify the permissions of a role or a group, the next scenario will have these modifications too.

Let's say you test that you can't see product categories if you remove the corresponding permission, the scenario is ok. Just after, no luck, the same worker picks a scenario that needs to access these product categories. The scenario fails, apparently for no reason. But it can pass on another build depending on the execution order.

It's not clear why we did that at the time, but obviously it can't work on a distributed system where scenarios execution order is unpredictable. It has no added value (not anymore it seems), it's complicated, adds instability and quite difficult to debug. 

So the idea is to completely clear the permissions between each scenario, like all other fixtures.

**What's in this PR?**
- All tables without exceptions are cleared between each scenario using Doctrine's ORMPurger. I removed our custom one that handled excluded tables.
- ACLs tables and cache are cleared between each scenario.
- The step `I reset the "..." rights` is deleted as it shouldn't be used anymore.


features/category/list_categories.feature
```diff
@@ -24,5 +24,4 @@ Feature: List categories
     Given I am on the categories page
     Then I should not see "Please select a category on the left or Create a new category"
     When I click on the "Summer collection" category
-    Then I should not see "Server error"
-    Then I reset the "Administrator" rights
+    Then I should not see the text "Server error"
```
pierallard: Weird to check this

features/Pim/Behat/Context/SimplePurger.php
```diff
@@ -12,32 +12,32 @@
  * @copyright 2016 Akeneo SAS (http://www.akeneo.com)
  * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
  */
-class DbSessionPurger implements PurgerInterface
+class SimplePurger implements PurgerInterface
```
pierallard: As this Purger only works with ORM, does it have a name like "OrmPurger" ?

pierallard:
Little comment but GTM as it

features/Pim/Behat/Context/SimplePurger.php
phaseinducer: It's not a perfect naming but it's not related to ORM. `MysqlTablePurger` would be exact.

features/category/list_categories.feature
nidup: Yep already here before the PR but weird, bonus point i think it's a flash message

features/Pim/Behat/Context/SimplePurger.php
nidup: As you only use Dbal connection it could be a DbalTablesPurger (should work with postgresql too). The fact you use dbal or orm is not a so big deal it's more an implem detail, TablesPurger or DBTablesPurger is maybe enough

nidup:
Nice PR, baby comment on purger naming and GTM. Much :fire: such :heart:

features/Pim/Behat/Context/SimplePurger.php
nidup: In fact DbalPurger is pretty good naming too, quite close to the OrmPurger naming used for all tables managed through orm entities

nidup:
I rebased in another branch TIP-423-master-dci and running to check the status

nidup:
@phaseinducer double checked and merged, thx!

PR #4482, opened 2016-05-19
PIM-5611: Fix notification report button position

pierallard:
capture du 2016-05-19 10 53 33

Only CSS, need to be GTMed by @akeneo/ux-team


yulldesign:
GTM for UX

ClementGautier:
GTM

PR #4488, opened 2016-05-20
TIP-431: Ease the migration by using JobParametersFactory in ConfigurationToJobParametersTransformer

damien-carcel:
**Description**

For now, JobParameters is directly instantiated in ConfigurationToJobParametersTransformer. This prevents to automatically fill an incomplete rawConfiguration when editing a profile of an existing job instance.

This can occur only with the job instance created before the merge of the JobParameters PR, if a job instance exists in the database with a rawConfiguration where a parameter is missing (but can be reproduced by changing the conf directly in the database, by removing the separator configuration for instance).

This PR solves this by avoiding to do a “new JobParameters()”, but use the JobParametersFactory instead. This will ensure that the default values are added if they do not exist.

### DIFF
--- a/src/Pim/Bundle/ImportExportBundle/Form/DataTransformer/ConfigurationToJobParametersTransformer.php
+++ b/src/Pim/Bundle/ImportExportBundle/Form/DataTransformer/ConfigurationToJobParametersTransformer.php
@@ -2,12 +2,13 @@
 
 namespace Pim\Bundle\ImportExportBundle\Form\DataTransformer;
 
+use Akeneo\Component\Batch\Job\JobInterface;
 use Akeneo\Component\Batch\Job\JobParameters;
+use Akeneo\Component\Batch\Job\JobParametersFactory;
 use Symfony\Component\Form\DataTransformerInterface;
-use Symfony\Component\Form\Exception\TransformationFailedException;
 
 /**
- * Transform a configuration array to a JobParameters and conversely
+ * Transforms a configuration array to a JobParameters and conversely.
  *
  * @author    Nicolas Dupont <nicolas@akeneo.com>
  * @copyright 2016 Akeneo SAS (http://www.akeneo.com)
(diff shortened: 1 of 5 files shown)


src/Pim/Bundle/ImportExportBundle/Form/DataTransformer/ConfigurationToJobParametersTransformer.php
```diff
@@ -2,41 +2,54 @@
...
+    {
+        $this->jobParametersFactory = $jobParametersFactory;
+        $this->job                  = $job;
+    }
+
+    /**
+     * {@inheritdoc}
      *
-     * @return JobParameters
+     * Transforms an configuration (array) to a job parameters (object).
```
pierallard: a* configuration

pierallard:
GTM, continue to do descriptions like this, it's a pleasure to read.

src/Akeneo/Component/Batch/Job/JobParametersFactory.php
```diff
@@ -37,8 +37,10 @@ public function __construct(DefaultValuesProviderRegistry $registry, $jobParamet
      */
     public function create(JobInterface $job, array $parameters = [])
     {
-        $provider = $this->defaultRegistry->get($job);
-        $parameters = array_merge($provider->getDefaultValues(), $parameters);
+        if (null !== $job->getName()) {
```
nidup: nope, we should not change this contract due to a weird constraint from the form type stuff

src/Pim/Bundle/ImportExportBundle/Form/DataTransformer/ConfigurationToJobParametersTransformer.php
```diff
@@ -2,41 +2,54 @@
...
+    protected $job;
+
     /**
-     * Transforms an configuration (array) to a job parameters (object)
-     *
-     * @param  array|null $configuration
+     * @param JobParametersFactory $jobParametersFactory
+     * @param JobInterface         $job
+     */
+    public function __construct(JobParametersFactory $jobParametersFactory, JobInterface $job)
```
jjanvier: is it normal the job is here?
this class is not a symfony service?

src/Akeneo/Component/Batch/Job/JobParametersFactory.php
nidup: I don't understand how we can pass here a job with no name, can we discuss IRL?

nidup:
Interesting PR, the JobParametersFactory must not be changed, GTM once we'll have fixed this (we can discuss IRL).

src/Pim/Bundle/ImportExportBundle/Form/DataTransformer/ConfigurationToJobParametersTransformer.php
nidup: this class is not a symfony service, as it depends on a Job we avoid to declare it as a "singleton service" and imho as this class should never be used in another context or replaced, we don't even declare it as private service. so looks ok for me.

skeleton:
GTM once Nico's comment about JobParametersFactory will be fixed

src/Akeneo/Component/Batch/Job/JobParametersFactory.php
damien-carcel: OK, my bad, remnant from a previous approach.

PR #4510, opened 2016-05-25
PIM-5801: Fix save in PEF when attribute code is only numeric

damien-carcel:
**Description**

When one add to a product an attribute with a code containing only numbers values, it is not possible to save the product when this attribute has been filled.

See screenshot below:

image

As data is passed to the product updater as an array, with the attribute code as an array key, ```array_merge``` replace the numerical codes (first it encounters by 0, second by 1, etc…) so the updater doesn't understand to which attribute the value is related. Using the union operator instead of ```array_merge``` solve this.

**Definition Of Done**

### DIFF
--- a/CHANGELOG-1.5.md
+++ b/CHANGELOG-1.5.md
@@ -5,6 +5,7 @@
 - PIM-5710: Fix thumbnail display after file upload in the product edit form
 - PIM-5811: Fix family export with multiple locales activated
 - PIM-5726: Fix number of product displayed on Product Grid when category panel is withdrawn
+- PIM-5801: Fix save in product edit form when attribute code is only numeric
 
 # 1.5.3 (2016-05-13)
 
--- a/features/product/add_attributes_to_a_product.feature
+++ b/features/product/add_attributes_to_a_product.feature
@@ -39,3 +39,12 @@ Feature: Add attributes to a product
   Scenario: Successfully display unclassified attributes in group "Other"
     Given I am on the "sandals" product page
     Then I should see available attribute Comment in group "Other"
+
+  @jira https://akeneo.atlassian.net/browse/PIM-5801
+  Scenario: Successfully add an attribute with fully numeric identifier
+    Given I am on the "boots" product page
+    And I add available attribute 123
+    And I change the "Attribute 123" to "foobar"
+    When I save the product
+    Then I should not see the text "There are unsaved changes."
+    And the field Attribute 123 should contain "foobar"
--- a/src/Pim/Bundle/EnrichBundle/Controller/Rest/ProductController.php
+++ b/src/Pim/Bundle/EnrichBundle/Controller/Rest/ProductController.php
@@ -325,7 +325,7 @@ class ProductController
         $values = $this->emptyValuesFilter->filter($product, $values);
 
         unset($data['values']);
-        $data = array_merge($data, $values);
+        $data = array_replace($data, $values);
 
         $this->productUpdater->update($product, $data);
     }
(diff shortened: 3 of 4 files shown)


ClementGautier:
GTM

src/Pim/Bundle/EnrichBundle/Controller/Rest/ProductController.php
```diff
@@ -325,7 +325,7 @@ protected function updateProduct(ProductInterface $product, array $data)
         $values = $this->emptyValuesFilter->filter($product, $values);
 
         unset($data['values']);
-        $data = array_merge($data, $values);
+        $data += $values;
```
phaseinducer: Usually we use `array_replace()` for this case.

features/product/add_attributes_to_a_product.feature
```diff
@@ -39,3 +39,19 @@ Feature: Add attributes to a product
...
     Then I should see available attribute Comment in group "Other"
+
+  Scenario: Successfully add an attribute with fully numeric identifier
+    Given I am on the attributes page
+    And I create a "Text" attribute
+    Then I fill in the following information:
+      | Code            | 12345 |
+      | Attribute group | Other |
+    And I save the attribute
+    Then I should not see the text "There are unsaved changes."
```
phaseinducer: Why this step?

features/product/add_attributes_to_a_product.feature
```diff
@@ -39,3 +39,19 @@ Feature: Add attributes to a product
   Scenario: Successfully display unclassified attributes in group "Other"
     Given I am on the "sandals" product page
     Then I should see available attribute Comment in group "Other"
+
+  Scenario: Successfully add an attribute with fully numeric identifier
+    Given I am on the attributes page
+    And I create a "Text" attribute
+    Then I fill in the following information:
```
phaseinducer: And

features/product/add_attributes_to_a_product.feature
```diff
@@ -39,3 +39,19 @@ Feature: Add attributes to a product
...
+      | Code            | 12345 |
+      | Attribute group | Other |
+    And I save the attribute
+    Then I should not see the text "There are unsaved changes."
+    When I am on the "boots" product page
+    And I add available attribute 12345
+    And I fill in the following information:
+      | [12345] | foobar |
+    And I save the product
+    And I should not see the text "There are unsaved changes."
```
phaseinducer: idem
Also it should be a "Then" here

features/product/add_attributes_to_a_product.feature
```diff
@@ -39,3 +39,19 @@ Feature: Add attributes to a product
...
+    But I should see the flash message "Product successfully updated"
```
phaseinducer: Can you please check we the team about current policy around flash messages?

CHANGELOG-1.5.md
```diff
@@ -5,6 +5,7 @@
 - PIM-5710: Fix thumbnail display after file upload in the product edit form
 - PIM-5811: Fix family export with multiple locales activated
 - PIM-5726: Fix number of product displayed on Product Grid when category panel is withdrawn
+- PIM-5801: Fix save in PEF when attribute code is only numeric
```
phaseinducer: "PEF" is too technical / internal imho. Maybe use simply "product edit form"?

features/product/add_attributes_to_a_product.feature
damien-carcel: To be sure the save was really performed. But maybe it is not necessary in this case.

phaseinducer:
GTM

PR #4513, opened 2016-05-25
TIP-36: Fix test because products export order is unpredictable

grena:
By doing a product export, we can't predict the product order in the different files.
So our test must be agnostic of this.


features/Pim/Behat/Context/Domain/Spread/XlsxFileContext.php
```diff
@@ -36,6 +36,47 @@ public function exportedXlsxFileOfShouldContain($number = null, $code, TableNode
      * @param string    $code
      * @param TableNode $expectedLines
      *
+     * @Then /^exported xlsx files of "([^"]*)" should contain:$/
+     */
+    public function exportedXlsxFilesOfShouldContain($code, TableNode $expectedLines)
```
pierallard: I have a lot of pain to read what do this function... Maybe naming is bad, maybe you should move some code into sub-functions...

features/Pim/Behat/Context/Domain/Spread/XlsxFileContext.php
```diff
@@ -78,6 +119,19 @@ public function xlsxFileShouldContainRows($fileName, $rows)
     }
 
     /**
+     * @param int    $number
+     * @param string $code
+     * @param int    $itemsCount
+     *
+     * @Then /^exported xlsx file (\d+) of "([^"]*)" should contain (\d+) rows?$/
```
pierallard: I'm pinailling, but I may wrote "^the exported xlsx file..."

features/Pim/Behat/Context/Domain/Spread/XlsxFileContext.php
```diff
@@ -36,6 +36,47 @@ public function exportedXlsxFileOfShouldContain($number = null, $code, TableNode
      * @param string    $code
      * @param TableNode $expectedLines
      *
+     * @Then /^exported xlsx files of "([^"]*)" should contain:$/
```
pierallard: same, I may add "the exported xlsx files?" (with "?" and "the")

willy-ahva:
GTM

features/Pim/Behat/Context/Domain/Spread/XlsxFileContext.php
grena: It's to be consistent with the existing
```
@Given /^xlsx file "([^"]*)" should contain (\d+) rows$/
```
To avoid having : 
> **the** exported xlsx file 1 of "the_job" should contain 2 rows

and

> xlsx file 1 of "the_job" should contain 2 rows

features/Pim/Behat/Context/Domain/Spread/XlsxFileContext.php
grena: We should not call this method for 1 file only, as we should explicit the file we're looking in. There is already a dedicated method:
> Then exported xlsx file 2 of "the_job" should contain:

pierallard:
GTM

PR #4522, opened 2016-05-27
TIP-440: remove the warning translated title

nidup:
In 1.0, we introduced getName()  in StepElement (reader, processor, writer), this name is only used to build a translated title which is used to build warning messages during a step execution (cf screenshot).

warning

The translation key is always different but we almost always indicate the same name than the step, after discussion with PO, we should always indicate "warning" as warning title, the reason is enouch to detail the encountered issue.

This PR remove the now useless StepElement getName() and Warning getName().

| Micro Demo to the PO (Story only) | Y (replaced translated step element label by "warning"


src/Akeneo/Component/Batch/Item/StepElementInterface.php
```diff
@@ -11,10 +11,4 @@
  */
 interface StepElementInterface
```
nidup: @jjanvier wonder if this is still useful

jjanvier: This one, related to ItemStep, seems useless imho.

But maybe a more generic one, `src/Akeneo/Component/Batch/Step/StepElementInterface` could be useful.
Even it it's empty at the moment, it could help to identify all the steps element that could be injected in a step.

jjanvier: for example we could check in `StepFactory::createStep`  that the services that are injected do are `StepElementInterface`. Could be useful later, IDK

jjanvier:
GTM
just the remark about the StepElementInterface

src/Akeneo/Component/Batch/Item/StepElementInterface.php
nidup: great idea, digging to see what are the real impacts :)

src/Akeneo/Component/Batch/Step/StepElementInterface.php
```diff
@@ -1,20 +1,14 @@
...
 
 /**
- * StepElementInterface
+ * Defines a step element, an element which can be injected in a step
  *
  * @author    Nicolas Dupont <nicolas@akeneo.com>
  * @copyright 2016 Akeneo SAS (http://www.akeneo.com)
  * @license   http://opensource.org/licenses/MIT MIT
  */
 interface StepElementInterface
```
damien-carcel: Looks strange to me to have an empty interface. Couldn't we find a way to remove it completely?

src/Akeneo/Component/Batch/spec/Step/ItemStepSpec.php
```diff
@@ -114,7 +114,7 @@ function it_executes_with_an_invalid_item_during_processing(
 
         // second batch
         $processor->process('r4')->shouldBeCalled()->willThrow(new InvalidItemException('my msg', ['r4']));
-        $execution->addWarning(Argument::any(), Argument::any(), Argument::any(), Argument::any())->shouldBeCalled();
+        $execution->addWarning(Argument::any(), Argument::any(), Argument::any())->shouldBeCalled();
```
damien-carcel: Replace the multiple Argument::any with Argument::cetera

damien-carcel:
GTM. Just the same comment about StepElementInterface, and one optional change in a spec.

PR #4525, opened 2016-05-27
Fix behat regressions

ClementGautier:
In the Enterprise Edition we have to set job profile accesses in order to be able to access a job profil so this scenarios can't work as it.


phaseinducer:
GTM

pierallard:
gtm

PR #4529, opened 2016-05-27
Add array type config options to connectors

gplanchat:
This PR adds the ability to specify array type option values to connectors as done here :

https://github.com/lunistore/akeneo-connector-base/blob/master/README.md#productassetsprocessor


damien-carcel:
GTM

skeleton:
GTM, thanks for this contrib :)

PR #4543, opened 2016-05-30
PIM-5824: detach product at the end of process() method

skeleton:
PR to fix a memory leak on products export.

Problem occurs when PIM is used with EnhancedConnectorBundle. This bundle has it own reader and detach the product in it (https://github.com/akeneo-labs/EnhancedConnectorBundle/blob/v1.2.0/Reader/ProductReader.php#L388). But, if an entity linked to a product (product value, metric, etc) is handled during processor, Doctrine add a reference to the product in unit of work (so in memory) and the product will never detach from it => memory leak :)

There is no problem with the original reader of the PIM because there is a `$this->entityManager->clear()` before read products

@nidup we already call `detach()` for quick export (https://github.com/akeneo/pim-community-dev/blob/master/src/Pim/Bundle/EnrichBundle/Connector/Processor/QuickExport/ProductToFlatArrayProcessor.php#L106)

Before this fix, export was impossible.
With this fix: 20K products exported in 24 minutes

| Q                 | A
| ----------------- | ---
| Specs             | Y
| Behats            | N
| Blue CI           | ?
| Changelog updated | Y

### DIFF
--- a/src/Pim/Bundle/BaseConnectorBundle/Processor/ProductToFlatArrayProcessor.php
+++ b/src/Pim/Bundle/BaseConnectorBundle/Processor/ProductToFlatArrayProcessor.php
@@ -5,6 +5,7 @@ namespace Pim\Bundle\BaseConnectorBundle\Processor;
 use Akeneo\Component\Batch\Item\AbstractConfigurableStepElement;
 use Akeneo\Component\Batch\Item\ItemProcessorInterface;
 use Akeneo\Component\Localization\Localizer\LocalizerInterface;
+use Akeneo\Component\StorageUtils\Detacher\ObjectDetacherInterface;
 use Pim\Bundle\BaseConnectorBundle\Validator\Constraints\Channel;
 use Pim\Bundle\CatalogBundle\Builder\ProductBuilder;
 use Pim\Bundle\CatalogBundle\Manager\ChannelManager;
@@ -58,13 +59,17 @@ class ProductToFlatArrayProcessor extends AbstractConfigurableStepElement implem
     /** @var ProductBuilder */
     protected $productBuilder;
 
+    /** @var ObjectDetacherInterface */
+    protected $detacher;
+
     /**
      * @param Serializer              $serializer
      * @param ChannelManager          $channelManager
+     * @param ProductBuilderInterface $productBuilder
      * @param string[]                $mediaAttributeTypes
      * @param array                   $decimalSeparators
      * @param array                   $dateFormats
-     * @param ProductBuilderInterface $productBuilder
+     * @param ObjectDetacherInterface $detacher
      */
     public function __construct(
         Serializer $serializer,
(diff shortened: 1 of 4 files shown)


src/Pim/Bundle/BaseConnectorBundle/Processor/ProductToFlatArrayProcessor.php
```diff
@@ -58,28 +59,34 @@ class ProductToFlatArrayProcessor extends AbstractConfigurableStepElement implem
...
      */
     public function __construct(
         Serializer $serializer,
         ChannelManager $channelManager,
         ProductBuilderInterface $productBuilder,
         array $mediaAttributeTypes,
         array $decimalSeparators,
-        array $dateFormats
+        array $dateFormats,
+        ObjectDetacherInterface $detacher = null
```
damien-carcel: Not in a setter?

damien-carcel:
Just one question, otherwise GTM.

src/Pim/Bundle/BaseConnectorBundle/Processor/ProductToFlatArrayProcessor.php
```diff
@@ -58,28 +59,34 @@ class ProductToFlatArrayProcessor extends AbstractConfigurableStepElement implem
     /** @var ProductBuilder */
     protected $productBuilder;
 
+    /** @var ObjectDetacherInterface */
+    protected $detacher;
+
     /**
      * @param Serializer              $serializer
      * @param ChannelManager          $channelManager
+     * @param ProductBuilderInterface $productBuilder
```
jmleroux: :+1:

jmleroux:
GTM

src/Pim/Bundle/BaseConnectorBundle/Processor/ProductToFlatArrayProcessor.php
willy-ahva: According to the following code and this `= null` we have no BC Break, so IMO it's ok.

willy-ahva:
GTM

src/Pim/Bundle/BaseConnectorBundle/Processor/ProductToFlatArrayProcessor.php
damien-carcel: If the integrator add a new dependency, then there is a BC, even with a = null. A setter allow to avoid it.

solivier:
gtm

src/Pim/Bundle/BaseConnectorBundle/Processor/ProductToFlatArrayProcessor.php
skeleton: seen IRL

jjanvier:
GTM

PR #4551, opened 2016-05-31
PIM-5712: Keep file reference in db after media removal

damien-carcel:
**Description**

When adding a new media to a product, an entry is created in the table “akeneo_file_storage_file_info”. When removing the media, the entry in “akeneo_file_storage_file_info” must be kept so history is preserved, and revert is correctly done in EE.

If the user adds an attribute, upload a media, save the product, then directly remove the attribute field still containing the media, and save the product, in this case the entry in the database is deleted. This is due to the cascade remove perform by Doctrine on the product value.

This does not happen if one first removes the media (but keeps the attribute) and save the product as the product value is still here (as the field), just cleaned through JavaScript. Then, when the user removes the field, it is not linked to the media anymore, so the media is kept in a database.

This PR fixes the problem by removing the cascade remove on media in the product value.

**Definition Of Done**

### DIFF
--- a/features/Context/AssertionContext.php
+++ b/features/Context/AssertionContext.php
@@ -585,7 +585,7 @@ class AssertionContext extends RawMinkContext
     public function theScopableFieldShouldHaveTheFollowingColors($field, TableNode $table)
     {
         $element = $this->getCurrentPage()->find('css', sprintf('label:contains("%s")', $field))->getParent();
-        $colors  = $this->getMainContext()->getContainer()->getParameter('pim_enrich.colors');
+        $colors  = $this->getParameter('pim_enrich.colors');
         foreach ($table->getHash() as $item) {
             $style = $element->find('css', sprintf('label[title="%s"]', $item['scope']))->getAttribute('style');
             assertGreaterThanOrEqual(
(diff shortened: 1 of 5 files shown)


jjanvier:
GTM :+1:

features/Context/AssertionContext.php
```diff
@@ -840,6 +840,46 @@ public function iShouldNotSeeDefaultAvatar()
...
+     */
+    public function theFileShouldExistInDatabase($originalFilename)
+    {
+        $fileInfoRepoClass  = $this->getParameter('akeneo_file_storage.model.file_info.class');
+        $fileInfoRepository = $this->getRepository($fileInfoRepoClass);
+
+        $fileInfo = $fileInfoRepository->findOneBy(['originalFilename' => $originalFilename]);
+
+        assertNotNull($fileInfo, sprintf(
+            'Unable to find file with original filename %s  in database',
```
skeleton: add `"` around `%s` please, it's more readable. and remove extra space before `in database`

src/Pim/Bundle/CatalogBundle/Resources/config/model/doctrine/ProductValue.orm.yml
```diff
@@ -88,7 +88,6 @@ Pim\Component\Catalog\Model\ProductValue:
         media:
             targetEntity: Akeneo\Component\FileStorage\Model\FileInfoInterface
             cascade:
-                - remove
```
skeleton: I'm not sure we can do that into a fix, it's a BC break cc @nidup

nidup: We can fix this way, the problem is that if the product value is overriden in the project, the fix will maybe not be applied. We may have to apply the same fix in the published product value mapping in the EE. GTM once EE mapping checked.

damien-carcel: This not a problem as EE mapping extends CE mapping, using our override system.
CI is OK on EE MongoDB and the two tests I added run fine on local, both on CE ORM and EE ORM.
Override in projects should be fine too, as long as it is done as we explain in our documentation.

PR #4560, opened 2016-06-01
Fix requirement of families at import

rybus:

This is end of my previous PR fixing FamilyUpdater. I added this very line as as soon as you get inside this method, requirement should be true. Otherwise, if the requirement existed and equaled false then it would remain... false. 

| Added Behats                      | N/A


damien-carcel:
GTM

skeleton:
GTM

PR #4569, opened 2016-06-02
Change the product creation by UI and set it with import

pierallard:
These 2 scenario are ones of the longers in the DCI, because it creates 3 products via UI.
I replaced it with an import creation to be faster (and clearer btw)

Before: 5m34.305s and 2m12.711s
After: 1m15.102s and 1m13.562s


ClementGautier:
I'm not sure that we should do that as it no more test that the history is well updated using the UI.

pierallard:
See the changelog of the file. Before, there was no history at all!

ClementGautier:
GTM

jjanvier:
@ClementGautier 4 products were created via the UI
only one history was checked

Pierre only delete the 3 useless products of the behat

GTM

PR #4573, opened 2016-06-02
Fix PhpDoc and remove unused namespace

jmleroux:
A little PhpDoc fix. For the API and to enable IDE autocomplete.


solivier:
gtm

PR #4574, opened 2016-06-02
PIM-5687: Allow to remove a product from a variant group at first attempt

damien-carcel:
**Description**

It was not possible to remove a product from a variant group at the first attempt when the PIM is installed in MongoDB. One needs to do it twice, and the second time the product is indeed removed.

Issue was caused by the fact that at creation, the variant group doesn’t have an associated product template. The template not created during variant group import, only when saved from the UI.

This PR adds an empty ProductTemplate during group creation, directly in the GroupFactory (and only for variant groups).

**Definition Of Done**

| Review and 2 GTM                  | Pendong

### DIFF
--- a/CHANGELOG-1.4.md
+++ b/CHANGELOG-1.4.md
@@ -2,6 +2,7 @@
 
 ## Bug fixes
 - PIM-5820: Fix product value filtering by channel and locale in structured product normalizer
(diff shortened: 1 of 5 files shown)


solivier:
gtm

willy-ahva:
GTM

ClementGautier:
gtm

PR #4577, opened 2016-06-03
Fix Yaml translation to be parsed by crowdin

pierallard:
Crowdin is very strict with Yaml parsing (not like Sf parser)
This fix may avoid that:

https://crowdin.com/translate/akeneo/111/en-fr#12463


damien-carcel:
GTM

ClementGautier:
gtm

PR #4579, opened 2016-06-03
Fix random behat

pierallard:
This PR does 3 things:
- little refactor of a behat to avoid weird behavior
- drop gherkin function (unused)
- Add a check to fillTextField to avoid FATAL on DCI. I know, we should refactor this method, but it's not the goal of this PR, it's only to stabilize DCI.


willy-ahva:
gtm

jjanvier:
GTM

PR #4583, opened 2016-06-05
TIP-266: small batch enhancements

jjanvier:
What's done here:
* remove `SimpleFileWriter` that is unused
* set the number of max lines optional in the flat item buffer flusher. This way we'll be able to use the flusher even if you don't want to split the output into several files (ie: we could use the same logic in every CSV or XLSX writer)
* make the flat item buffer flusher able to output files without extension 
* fix a bug on `BatchBundle`. When a `JobExecution` was retrieved from the database, no `ExecutionContext` was attached to it.


src/Pim/Component/Connector/Writer/File/FlatItemBufferFlusher.php
```diff
@@ -44,19 +44,19 @@ public function __construct(FilePathResolverInterface $filePathResolver, ColumnS
      * Several output files are created if the buffer contains more items that maximum lines authorized per output file.
      *
      * @param FlatItemBuffer $buffer
-     * @param int            $maxLinesPerFile
-     * @param string         $baseFilePath
+     * @param int            $maxLinesPerFile by default -1, which means there is no limit of lines
```
pierallard: I prefer null value for default instead of negative integer, but don't know if it's possible in PHP.

pierallard:
GTM

src/Pim/Component/Connector/Writer/File/FlatItemBufferFlusher.php
jjanvier: I could do that
but as there are some % operations on `$maxLinesPerFile`, the first thing I'll do in the method would be
```
if (null === $maxLinesPerFile) $maxLinesPerFile = -1;
```

So I'm not sure it's better..

skeleton:
you should add specs to test the new behavior (no split on files, no extension)

GTM once specs added

jjanvier:
@skeleton for the specs, ask @willy-ahva... 

I extracted this behavior into this class `FlatItemBufferFlusher`. This behavior was previously copied/pasted in all the XLSX writers. And there was no spec at all for that...

src/Pim/Component/Connector/Writer/File/FlatItemBufferFlusher.php
pierallard: Totally understand. Second solution is worst. Let -1 :)

jjanvier:
specs added

PR #4584, opened 2016-06-06
Merge 1.5 into master

skeleton:


features/Context/FixturesContext.php
```diff
@@ -1692,6 +1685,14 @@ protected function createProductGroup($code, $label, $type, array $attributes, a
     }
 
     /**
+     * @return \Pim\Bundle\CatalogBundle\Factory\GroupFactory
```
pierallard: Return GroupFactory (use use)

features/Context/DataGridContext.php
```diff
@@ -354,13 +354,24 @@ public function iShowTheFilter($filterName)
...
     {
         $this->datagrid->hideFilter($filterName);
     }
 
     /**
+     * @param string $filterName
+     *
+     * @Then /^I expand the "([^"]*)" sidebar$/
+     */
+    public function iexpandTheCategoriesSidebar($filterName)
```
pierallard: iExpand

features/Context/Page/Base/Grid.php
```diff
@@ -209,8 +209,33 @@ public function findAction($element, $actionName)
...
+     * @throws \InvalidArgumentException
+     */
+    public function checkOptionInFilter(array $expectedOptions, $filterName)
+    {
+        $filter = $this->getFilter($filterName);
+
+        $filter->open();
+        $options = $filter->getOptions();
+
+        if ($options != $expectedOptions) {
```
pierallard: !== ? Not sure.

pierallard:
Little comments then GTM

features/Context/Page/Base/Grid.php
skeleton: with the operator `===`, when you compare 2 array, order is checked and can return false if thse 2 array are not exactly equals
with `==` data are well checked, but doesn't care about order

damien-carcel:
GTM

PR #4585, opened 2016-06-06
Fix for master

phaseinducer:
The first scenario must be run in CE only (not the same columns expected).
The second one was a false positive in EE and is now fixed.
The last two scenarios are skipped and will be fixed in PIM-5830.


skeleton:
GTM

damien-carcel:
GTM

PR #4596, opened 2016-06-10
[WIP] PIM-5843: Use ObjectManager::merge in ProductSaver on imports

damien-carcel:
<3 Thanks for taking the time to contribute! You're awesome! <3

Prevent product duplication on import when the product is present several time in the same imported file.

More details coming.

| Added Specs                       | Pending


jjanvier:
I already don't like the `flush` option (should be deleted imho, I don't understand in which case we need it... as a saver should save to dababase)

so adding another technical option is not the way to do it imho

also, even if the merge ​*do*​ is mandatory, you have to test the impacts
as maybe it's a "gros sac de noeud" in doctrine, maybe it can slow the imports...

jjanvier:
maybe what you can do is:
* find the product in DB
* if it doesn't exist, check in the current batch page if the product exits
* if it doesn't exist, then create it

src/Pim/Bundle/CatalogBundle/Doctrine/Common/Saver/ProductSavingOptionsResolver.php
```diff
@@ -22,9 +22,9 @@ public function resolveSaveOptions(array $options)
         $resolver = $this->createOptionsResolver();
         $resolver->setDefaults(
             [
-                'flush'             => true,
-                'recalculate'       => true,
-                'schedule'          => true
+                'flush'       => true,
+                'recalculate' => true,
+                'schedule'    => true,
```
nidup: i would avoid to add this kind of very internal extra option, agree with @jjanvier I would drop the flush option too, we should have only "business" option here, no "technical coupled to doctrine" stuff

damien-carcel: Removed, as it was completely useless anyway (it was more of a remnant of a previous try…)

src/Pim/Bundle/CatalogBundle/Doctrine/Common/Saver/ProductSaver.php
```diff
@@ -22,7 +24,7 @@
  */
 class ProductSaver implements SaverInterface, BulkSaverInterface
 {
-    /** @var ObjectManager */
+    /** @var EntityManager */
```
jjanvier: no
that's not what is passed to the constructor

jjanvier: this saver is agnostic of ORM/ODM

src/Pim/Bundle/CatalogBundle/Doctrine/Common/Saver/ProductSaver.php
```diff
@@ -103,6 +104,12 @@ public function saveAll(array $products, array $options = [])
         $itemOptions['flush'] = false;
 
         foreach ($products as $product) {
+            if (null !== $product->getId() &&
+                UnitOfWork::STATE_DETACHED === $this->objectManager->getUnitOfWork()->getEntityState($product)
```
jjanvier: `getUnitOfWork()` is not available from the interface `ObjectManager`
we can accept it as a hotfix, but maybe we should find another way to do it for master

src/Pim/Bundle/CatalogBundle/Doctrine/Common/Saver/ProductSaver.php
damien-carcel: Oups, my bad, I forgot to change that…

src/Pim/Bundle/CatalogBundle/Doctrine/Common/Saver/ProductSaver.php
damien-carcel: You are right on both subject.

I think Doctrine principle is that you should never use UOW yourself, it is Doctrine internal. But it is still a public method.

PR #4605, opened 2016-06-13
PIM-5850: Update help for akeneo:batch:job command

damien-carcel:
**Description**

Help for the "config" option is wrong. It currently displays the following message:

```-c, --config=CONFIG      Override job configuration (formatted as json. ie: php app/console akeneo:batch:job -c '[{"reader":{"filePath":"/tmp/foo.csv"}}]' acme_product_import)```

The JSON syntax is wrong and should be `"{\"filePath\":\"/tmp/foo.csv\"}"`

**Definition Of Done**

### DIFF
--- a/src/Akeneo/Bundle/BatchBundle/Command/BatchCommand.php
+++ b/src/Akeneo/Bundle/BatchBundle/Command/BatchCommand.php
@@ -38,8 +38,7 @@ class BatchCommand extends ContainerAwareCommand
                 'c',
                 InputOption::VALUE_REQUIRED,
                 'Override job configuration (formatted as json. ie: ' .
-                'php app/console akeneo:batch:job -c \'[{"reader":{"filePath":"/tmp/foo.csv"}}]\' ' .
-                'acme_product_import)'
+                'php app/console akeneo:batch:job -c "{\"filePath\":\"/tmp/foo.csv\"}" acme_product_import)'
             )
             ->addOption(
                 'email',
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/Akeneo/Bundle/BatchBundle/Command/BatchCommand.php [hunk 1]
```diff
@@ -38,8 +38,7 @@ protected function configure()
                 'c',
                 InputOption::VALUE_REQUIRED,
                 'Override job configuration (formatted as json. ie: ' .
-                'php app/console akeneo:batch:job -c \'[{"reader":{"filePath":"/tmp/foo.csv"}}]\' ' .
-                'acme_product_import)'
+                'php app/console akeneo:batch:job -c "{\"filePath\":\"/tmp/foo.csv\"}" acme_product_import)'
```
pierallard: what's this parenthesis at the end of the command ?
Why not use ' around the hash instead of " ? you will be able to remove \"

damien-carcel: “Override job configuration (formatted as json. ie: ”
The parenthesis closes the example, it is on two lines because it exceeds 120 characters.
And in my case, using ' "arg" ' in the command does not work, causing a json_decode error, where " "\arg"\ " works. Maybe related to my shell configuration, but if I have the problem, other could have it to.

jjanvier:
gtm

skeleton:
gtm

Issue #4659, opened 2016-06-27
Exception during product export


Hocdoc:
When we export our 160,000 products with Akeneo 1.5.5/MySQL we sometimes get this exception during the job execution:

```
[...]

PR #4665, opened 2016-06-27
Ignore integrity constraint violations

Hocdoc:
When running 2 completeness calculations executed at the same time you may get integrity constraint violations (see #4659).

This PR use REPLACE instead of INSERT INTO. REPLACE works exactly like INSERT, except that if an old row in the table has the same value as a new row for a PRIMARY KEY or a UNIQUE index, the old row is deleted before the new row is inserted.

There still may be errors with dropped `complete_price`, `missing_completeness` tables, but it defuse the problem of concurrent completeness calculations.

### DIFF
--- a/src/Pim/Bundle/CatalogBundle/Doctrine/ORM/CompletenessGenerator.php
+++ b/src/Pim/Bundle/CatalogBundle/Doctrine/ORM/CompletenessGenerator.php
@@ -282,7 +282,7 @@ MISSING_SQL;
     protected function getMainSqlPart()
     {
         return <<<MAIN_SQL
-            INSERT IGNORE INTO pim_catalog_completeness (
+            REPLACE pim_catalog_completeness (
                 locale_id, channel_id, product_id, ratio, missing_count, required_count
             )
             SELECT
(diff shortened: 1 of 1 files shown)


BitOne:
Hi @Hocdoc ,

Thanks for the contribution !

As the last completeness would certainly be more up to date, `IGNORE` would mean that the proper value will certainly be ignored.

What do you think about using "ON DUPLICATE KEY UPDATE" instead ? (see http://dev.mysql.com/doc/refman/5.7/en/insert-on-duplicate.html).

Thanks !

Hocdoc:
I'm now using `REPLACE` (a MySQL extension to the SQL standard) to set the completeness to the latest value.

(When exporting our 170,000 products this completeness calculation runs about 1h. I'm not sure, if it the completeness value set by this long request is always more up to date than the other concurrent completeness calculation which may start later f.ex. when a user edits a product. But either way I wish the export didn't fail with an Duplicate-entry-exception.)

fabienlem:
Hello @Hocdoc,

The new version of your fix seems to be a better solution! I updated the description in order to help our team to review this PR. We keep you updated.

Regards

ClementGautier:
@Hocdoc : I've squashed / rebased and merged this pull request here : https://github.com/akeneo/pim-community-dev/pull/5412. Thx for the contrib !

PR #4691, opened 2016-07-04
Fix failing behat on localized history

pierallard:
# Wut

- Replace methods with correct spin (removes wait)
- Fix test replacing "following products" with import to be sure history is well constructed

## Warning

If you read in details, this PR changes field language of the history. This is due to a bug in the history language. I created another card in Jira for this specific bug, but the main behavior does not change in this PR (the fields in the history are displayed with the catalogLocale, not the userLocale).

### DIFF
--- a/features/localization/product/display_localized_history.feature
+++ b/features/localization/product/display_localized_history.feature
@@ -5,30 +5,40 @@ Feature: Display the localized product history
   I need to have show localized values
 
   Background:
-    Given a "default" catalog configuration
+    Given a "footwear" catalog configuration
     And the following attributes:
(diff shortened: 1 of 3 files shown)


features/localization/product/display_localized_history.feature
```diff
@@ -23,7 +31,7 @@ Feature: Display the localized product history
     Then there should be 1 update
     And I should see history:
       | version | property      | value     |
-      | 1       | SKU           | boots     |
+      | 1       | [sku]         | boots     |
```
SamirBoulil: Why is this changed ?

SamirBoulil:
GTM after comment answered

PR #4724, opened 2016-07-13
Dispatch pre and post save events only when flush the object

iulyanp:

Dispatch pre and post save events only when flush the object in the database. 
If those changes will be approved the "PimEnterprise\Bundle\WorkflowBundle\Doctrine\Common\Saver\DelegatingProductSaver" saveAll function should dispatch the PRE_SAVE_ALL and POST_SAVE_ALL events.

…abase

### DIFF
--- a/src/Akeneo/Bundle/StorageUtilsBundle/Doctrine/Common/Saver/BaseSaver.php
+++ b/src/Akeneo/Bundle/StorageUtilsBundle/Doctrine/Common/Saver/BaseSaver.php
@@ -87,7 +87,7 @@ class BaseSaver implements SaverInterface, BulkSaverInterface
         }
 
         $allOptions = $this->optionsResolver->resolveSaveAllOptions($options);
-        $this->eventDispatcher->dispatch(StorageEvents::PRE_SAVE_ALL, new GenericEvent($objects, $allOptions));
+        $this->dispatchPreSaveEvent($objects, $allOptions);
 
         $itemOptions = $allOptions;
         $itemOptions['flush'] = false;
@@ -99,7 +99,29 @@ class BaseSaver implements SaverInterface, BulkSaverInterface
         if (true === $allOptions['flush']) {
             $this->objectManager->flush();
         }
+        $this->dispatchPostSaveEvent($objects, $allOptions);
 
-        $this->eventDispatcher->dispatch(StorageEvents::POST_SAVE_ALL, new GenericEvent($objects, $allOptions));
+    }
+
+    /**
+     * @param object $objects
+     * @param array  $allOptions
+     */
+    protected function dispatchPreSaveEvent($objects, array $allOptions)
+    {
+        if (true === $allOptions['flush']) {
+            $this->eventDispatcher->dispatch(StorageEvents::PRE_SAVE_ALL, new GenericEvent($objects, $allOptions));
+        }
+    }
+
+    /**
+     * @param object $objects
+     * @param array  $allOptions
+     */
+    protected function dispatchPostSaveEvent($objects, array $allOptions)
+    {
+        if (true === $allOptions['flush']) {
+            $this->eventDispatcher->dispatch(StorageEvents::POST_SAVE_ALL, new GenericEvent($objects, $allOptions));
+        }
     }
 }
--- a/src/Akeneo/Bundle/StorageUtilsBundle/Doctrine/Common/Saver/BaseSaver.php
+++ b/src/Akeneo/Bundle/StorageUtilsBundle/Doctrine/Common/Saver/BaseSaver.php
@@ -66,7 +66,8 @@ class BaseSaver implements SaverInterface, BulkSaverInterface
         }
 
         $options = $this->optionsResolver->resolveSaveOptions($options);
-        $this->eventDispatcher->dispatch(StorageEvents::PRE_SAVE, new GenericEvent($object, $options));
+
+        $this->dispatchPreSaveEvent($object, $options);
 
         $this->objectManager->persist($object);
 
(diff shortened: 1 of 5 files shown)


nidup:
Hi @iulyanp 

Sorry for the late answer and thank you for the pull request :tada:

FYI, about the "flush" option, we would totally drop it in an upcoming minor version for both Savers and Removers (could be in v1.6).

The "flush" option is a pure internal concern (doctrine), when I call save(), or saveAll() or remove() or removeAll(), I do expect that the change has been made in the database, no need of an option. In this case it would be better to duplicate a bit the logic of the save method in the saveAll method to avoid to pass the flush = false and to avoid to receive "erroneous" events. In the PIM, the flush = false is almost not used (except in Saver and Remover calling themselve + few cases in EE).

About your proposal on the 1.5, I would prefer a project solution, ie, check in your listener that the flush option = true to avoid to introduce any side effect (even minor) in existing projects and extensions relying on the v1.5.

Would you be interested to contribute about this flush option dropping on the master? If yes, I would be glad to help you on the PR.

BR,

iulyanp:
Hello @nidup

Thank you for your answer.
Sorry but I didn't fully understood the part with "I would prefer a project solution". Could you please be more specific.
I created this PR because we are working on a connector EventStoreBundle between Magento and Eventstore (https://geteventstore.com/). We rely our code on the events you raise from PIM after some entity it's saved into PIM database. We listen for save or saveAll events. The issue I have is that when saveAll is called I have "pre/post_save_all" and also "pre/post_save" events raised.  
I would like to know how exactly I could contribute, from which branch should I start working and create the PR? Would my PR be included in the 1.6 version?

Best regards,

nidup:
Hi @iulyanp 

Thank you for the details, I didn't fully understood your use case, it's far more clear now :+1: (you can forget my previous answer :wink: ).

No matter if you use saveAll() or save(), the "pre/post_save" events should always be raised for each saved product and each saved product should be dispatched in these events. The exact same logic is implemented in other Savers for any other entities.

The events "pre/post_save_all" notice the start and the end of a bulk saving, idea is that you you can listen this event to add a business logic which should be launched only for bulk change (for instance, you could use it in Enterprise Edition to run rules on bulk products save but you don't want run rules on each unitary product save launched from the product edit form).

**In your case, I would plug your listener only on "pre/post_save" event** to be sure to run logic on any saved entity, no matter bulk or unitary operation has been launched (it will also work with DelegatingProductSaver as it dispatches to the ProductSaver to save each product when you've permissions to update them).

I carefully re-checked the PIM savers and I noticed a design leak when using the MongoDB storage, one of our BulkSaver does not dispatch the "pre/post_save" for each object, https://github.com/akeneo/pim-community-dev/blob/1.5/src/Pim/Bundle/CatalogBundle/Doctrine/MongoDBODM/Saver/ProductSaver.php

If you're interested to contribute, you can open a new PR on the branch 1.5 with the following patch to fix this MongoDB bulk saver (if you're not interested or not using MongoDB, no pb let me now and we'll add this change in our internal tech roadmap),
```
diff --git a/src/Pim/Bundle/CatalogBundle/Doctrine/MongoDBODM/Saver/ProductSaver.php b/src/Pim/Bundle/CatalogBundle/Doctrine/MongoDBODM/Saver/ProductSaver.php
index 97e0d40..250b72a 100644
--- a/src/Pim/Bundle/CatalogBundle/Doctrine/MongoDBODM/Saver/ProductSaver.php
+++ b/src/Pim/Bundle/CatalogBundle/Doctrine/MongoDBODM/Saver/ProductSaver.php
@@ -105,6 +105,7 @@ class ProductSaver extends BaseProductSaver
             } else {
                 $productsToUpdate[] = $product;
             }
+            $this->eventDispatcher->dispatch(StorageEvents::PRE_SAVE, new GenericEvent($product, $options));
         }
 
         $insertDocs = $this->getDocsFromProducts($productsToInsert);
@@ -127,6 +128,9 @@ class ProductSaver extends BaseProductSaver
         $versions = $this->bulkVersionBuilder->buildVersions($products);
         $this->versionSaver->saveAll($versions);
 
+            $this->eventDispatcher->dispatch(StorageEvents::POST_SAVE, new GenericEvent($product, $options));
+        }
         $this->eventDispatcher->dispatch(StorageEvents::POST_SAVE_ALL, new GenericEvent($products, $options));
     }
```

With this change you should be able to develop your eventstore connector without any trouble for both ORM / MongoDB storages.

BTW, each merged PR in a minor version is also applied on the upcoming versions (the core team will be in charge to merge this 1.5 patch in the 1.6).

Don't hesitate if you need any further information, your use case is very interesting and may really help us to improve our API :smile:

iulyanp:
Hi @nidup 

Thank you for the details, everything sounds awesome.
About mongodb saver, we don't use mongodb at the moment, but I could send a PR with that fix. The only problem is that I will not be able to fix the php spec tests, please let me know if that's fine.
Also, if what you told me about the events is how it should be, maybe you should also take a look on "PimEnterprise\Bundle\WorkflowBundle\Doctrine\Common\Saver\DelegatingProductSaver", there "pre/post_save_all" events are not raised on the saveAll function.

Best regards,

nidup:
Hi @iulyanp!

Thank you for details, I'm sorry, I'm not sure to understand the changeset issue, are you talking about versionning changeset? If you get data directly from the dispatched product is seems ok (for both mass edit / product edit form save).

I have few questions to help me understand the issue, I took a quick look at http://docs.geteventstore.com/ and I guess that the subscriber will allow to send events to geteventstore, 
 - What is the format of the event you want to send? (paste sample json/xml please)
 - Do you want to put only the updated product fields in this event or the whole product data?
 - How do you plan to build the content of this event in your listener/subscriber?

I pushed here a very simple PR to do few tests, don't hesitate to add comment / snippet if you want to illustrate the pb inside https://github.com/akeneo/pim-community-dev/pull/4755/files or even better if you can show me your code :wink: 

Thanks,

PR #4754, opened 2016-07-21
PIM-5635 simple select reference data

SamirBoulil:
**Title:** Adds support for simple select reference data in the export builder

**Description:**

The filter view for simple select reference data works the same as the multi select reference data.

The simple select reference data attribute types uses the multi select reference data view in the export builder. It supports the same operators "Empty" and "In list".

**Definition Of Done:**

| Q | A |
| --- | --- |
| Added Specs |  |
| Added Behats | Y |
| Changelog updated |  |
| Review and 2 GTM | N |
| Migration script |  |
|  Tech Doc |  |

### DIFF
--- a/features/Pim/Behat/Decorator/Field/Select2Decorator.php
+++ b/features/Pim/Behat/Decorator/Field/Select2Decorator.php
@@ -22,8 +22,8 @@ class Select2Decorator extends ElementDecorator
         $values = '' !== $value ? explode(',', $value) : [];
         $this->prune();
 
-        $widget = $this->getWidget();
         foreach ($values as $value) {
+            $widget = $this->getWidget();
             $value = trim($value);
 
             $this->getSession()->executeScript(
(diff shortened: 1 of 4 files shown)


features/Pim/Behat/Decorator/Field/Select2Decorator.php
```diff
@@ -86,7 +86,7 @@ public function open()
     {
         $openerElement = $this->find('css', '.select2-arrow');
         if (null === $openerElement) {
-            $openerElement = $this->find('css', '.select2-choices');
+            $openerElement = $this->find('css', '.select2-search-field');
```
SamirBoulil: `.select2-choices` was not enough specific here. 

When an option was already selected and shown in the input and when the behat would try to click on the input to make the autocompletion appear.

The behat would inadvertedly click on the option, making the autocompletion window not appear.

PR #4760, opened 2016-07-21
[POC] backward compatibility promise

nidup:
POC PR to discuss / illustrate this hot topic on 2 components, StorageUtils and Batch.

Would love to take a large inspiration on http://symfony.com/doc/master/contributing/code/bc.html but we're very very very far to be able to engage on such amount of code, the product still evolve a lot and it's very hard to freeze the API.

We need to find our own path to semver, without locking too much innovation and changes in our still young product but ensuring smoother/clearer project upgrades for our beloved partners.

We introduce a Backward Compatibility Promise page in this PR (will be moved into only documentation but for now it allow to have the full picture in one place).

| Q | A |
| --- | --- |
| Added Specs |  |
| Added Behats |  |
| Changelog updated |  |
| Review and 2 GTM |  |
| Migration script |  |
|  Tech Doc |  |


src/Akeneo/Component/Batch/Job/JobInterface.php
```diff
@@ -13,11 +13,15 @@
  * @author    Benoit Jacquemont <benoit@akeneo.com>
  * @copyright 2013 Akeneo SAS (http://www.akeneo.com)
  * @license   http://opensource.org/licenses/MIT MIT
+ *
+ * @api
  */
 interface JobInterface
```
jjanvier: missing methods here

src/Akeneo/Component/Batch/Step/StepElementInterface.php
```diff
@@ -8,6 +8,8 @@
  * @author    Nicolas Dupont <nicolas@akeneo.com>
  * @copyright 2016 Akeneo SAS (http://www.akeneo.com)
  * @license   http://opensource.org/licenses/MIT MIT
+ *
+ * TODO, empty interface? was introduced to extract getName that we then drop because we now use convention for labels
```
jjanvier: can be removed, we never use it

src/Akeneo/Component/StorageUtils/Updater/PropertyAdderInterface.php
```diff
@@ -8,6 +8,8 @@
  * @author    Nicolas Dupont <nicolas@akeneo.com>
  * @copyright 2015 Akeneo SAS (http://www.akeneo.com)
  * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ *
+ * @internal
```
jjanvier: careful with this, only applicable if devs have a way to do the same thing without it
not sure about all the property*interface

willy-ahva:
I didn't know what is the difference between @api and @internal so I just put it here : 
@api: Contrib/custom code should feel free to depend on these APIs. We will try as hard as possible to not change this, except in cases for a critical issue such as a security issue.
@internal: We can and will change these if there is any good reason to; should not be called by contrib/custom code.

Pls, father beaver, can you explain how you choose the tag in term of the class ? I'm a little bit lost sometime to see your choice, maybe I just need to be enlightened ;)

src/Akeneo/Component/StorageUtils/Updater/PropertyAdderInterface.php
nidup: from the outside of the PIM I will push the use of ObjectUpdaterInterface, we use Property*Interface mainly for internal specific cases

src/Akeneo/Component/Batch/Job/JobInterface.php
nidup: getSteps() / getStep($name)?
If yes, you could implement a Job without steps imho. Maybe in another interface? (StepsContainerInterface or something like this?)

src/Akeneo/Component/Batch/Step/StepElementInterface.php
nidup: agree, i do this

nidup:
@willy-ahva I added a BC promise page in the PR to explain diff between @api and @internal .

nidup:
For now, we changed our strategy regarding api, the stability has been enforced through a stable web api (and standard format) documented here https://api.akeneo.com/

PR #4761, opened 2016-07-22
SDS-614: Fix CSS encoding weird chars

pierallard:
Git diff :

```
-<U+2028>// SDS-614: custom fix: allow to display correctly tree with large children count<U+2028>
-.jstree > ul > li:last-child {<U+2028>
-  padding-bottom: 80px<U+2028>;

+// SDS-614: custom fix: allow to display correctly tree with large children count
+.jstree > ul > li:last-child {
+  padding-bottom: 80px;
```


nidup:
GTM

loicguillois:
GTM

PR #4755, opened 2016-07-21
[POC] geteventstore (only used to illustrate a discussion, will be dropped)

nidup:
cf https://github.com/akeneo/pim-community-dev/pull/4724

PR #4767, opened 2016-07-22
Example of eventstore events

iulyanp:
Hello @nidup 

Thank you very much for your response.
Yes I'm talking about the versioning change set and yes we use data from the $event->getEventType() directly.

- What is the format of the event you want to send? (paste sample json/xml please)

     We use json. I pasted in some example.

- Do you want to put only the updated product fields in this event or the whole product data?

     We want to put all the product data + the change set.

- How do you plan to build the content of this event in your listener/subscriber?

     We use Normalizers to build the content of the events that we send to EventStore. We extended your Pim\Bundle\TransformBundle\Normalizer\Structured\ProductNormalizer and perform some formatting on its response so we don't send data which is only important for PIM (like validation rules).

I added some examples and comments in "Pim\Bundle\CatalogBundle\Doctrine\Common\Saver\ProductSaver" class A. B. and C.
Please let me know what do you think.

Thank you,

### DIFF
--- a/src/Acme/Bundle/AppBundle/Resources/doc/index.md
+++ b/src/Acme/Bundle/AppBundle/Resources/doc/index.md
@@ -1,0 +1,129 @@
+EventStore event examples
+--------------------------
+I have an event subscriber exactly like yours.
+
+
+#### Example of a mass edit product: action change status. 
+I disabled 2 products. As you can see the change set is the previous from when I created the product. 
+```json
+{
+  "family": "sofa_variation",
+  "groups": [],
+  "variant_group": null,
+  "categories": [],
+  "enabled": false,
+  "values": {
+    "sku": [
+      {
+        "locale": null,
+        "scope": null,
+        "data": "TESTING-1"
+      }
+    ],
+    "variation_parent_product": [
+      {
+        "locale": null,
+        "scope": null,
+        "data": "UK-SKUMAGENTO000003"
+      }
+    ],
+    "main_color": [
+      {
+        "locale": null,
+        "scope": null,
+        "data": "black"
+      }
+    ]
+  },
+  "associations": [],
+  "sku": "TESTING-1",
+  "changeset": {
+    "sku": {
+      "old": "",
+      "new": "TESTING-1"
+    },
+    "family": {
+      "old": "",
+      "new": "sofa_variation"
+    },
+    "main_color": {
+      "old": "",
+      "new": "black"
+    },
+    "variation_parent_product": {
+      "old": "",
+      "new": "UK-SKUMAGENTO000003"
+    },
+    "enabled": {
+      "old": "",
+      "new": 1
+    }
+  },
+  "author": {
+    "name": "Iulian Popa",
+    "email": "iulian.popa@made.com"
+  }
+}
+```
+In the change set I should have just one change, something like this:
+```json
+"changeset": {
+    "enabled": {
+      "old": 1,
+      "new": 0
+    }
+}
+```
+But because the flush on the product is not done I got the wrong change set.
+Or maybe because the build of the version is triggered after post_save, maybe that is a problem.
+
+## Example of a mass edit product: action "Classify products in categories".
+ Again the change set is the previous one because the object is not flushed.
+```json
+{
+  "family": "sofa_variation",
+  "groups": [],
+  "variant_group": null,
+  "categories": [
+    "sofa"
+  ],
+  "enabled": true,
+  "values": {
+    "sku": [
+      {
+        "locale": null,
+        "scope": null,
+        "data": "TESTING-1"
+      }
+    ],
+    "variation_parent_product": [
+      {
+        "locale": null,
+        "scope": null,
+        "data": "UK-SKUMAGENTO000003"
+      }
+    ],
+    "main_color": [
+      {
+        "locale": null,
+        "scope": null,
+        "data": "black"
+      }
+    ]
+  },
+  "associations": [],
+  "sku": "TESTING-1",
+  "changeset": {
+    "enabled": {
+      "old": 1,
+      "new": 0
+    }
+  },
+  "author": {
+    "name": "Iulian Popa",
+    "email": "iulian.popa@made.com"
+  }
+}
+```
+I made some comments as an example on the Pim\Bundle\CatalogBundle\Doctrine\Common\Saver\ProductSaver.
+Please let me know if you have any questions.
--- a/src/Pim/Bundle/CatalogBundle/Doctrine/Common/Saver/ProductSaver.php
+++ b/src/Pim/Bundle/CatalogBundle/Doctrine/Common/Saver/ProductSaver.php
@@ -83,7 +83,13 @@ class ProductSaver implements SaverInterface, BulkSaverInterface
         if (true === $options['recalculate']) {
             $this->completenessManager->generateMissingForProduct($product);
         }
-
+        /*
+         * A.
+         * If I listen for this event on mass edit I will not have in the $event->getEventType() the real changeset
+         * because the object was not flushed into the database.
+         * Also it's still not so clear for me why the post_save event should be dispatched if you didn't flush the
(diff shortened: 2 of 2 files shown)


nidup:
Hi @iulyanp 

Thx you for the very detailled answer, I'm merging this in my POC branch to do few tests before to answer.

BR,

src/Acme/Bundle/AppBundle/Resources/doc/index.md
```diff
@@ -0,0 +1,129 @@
+EventStore event examples
+--------------------------
+I have an event subscriber exactly like yours.
```
nidup: I pushed a new commit on the POC branch https://github.com/akeneo/pim-community-dev/pull/4755/commits/5816aa9de48009b9b2bdd061578aafa7e48b5efa to implement this with current constraint (only tested with CE, I'm diging in EE DelegatingSaver).

src/Acme/Bundle/AppBundle/Resources/doc/index.md
```diff
@@ -0,0 +1,129 @@
...
+```json
+"changeset": {
+    "enabled": {
+      "old": 1,
+      "new": 0
+    }
+}
+```
+But because the flush on the product is not done I got the wrong change set.
+Or maybe because the build of the version is triggered after post_save, maybe that is a problem.
```
nidup: right, this is a known legacy limitation (cf comments in the subscriber example)

src/Pim/Bundle/CatalogBundle/Doctrine/Common/Saver/ProductSaver.php
```diff
@@ -83,7 +83,13 @@ public function save($product, array $options = [])
         if (true === $options['recalculate']) {
             $this->completenessManager->generateMissingForProduct($product);
         }
-
+        /*
+         * A.
+         * If I listen for this event on mass edit I will not have in the $event->getEventType() the real changeset
+         * because the object was not flushed into the database.
+         * Also it's still not so clear for me why the post_save event should be dispatched if you didn't flush the
+         * object.
```
nidup: I though you only need the saved product, I better understand the issue if we need to pick the last version from the database. On all use cases I already encountered, we only used the dispatched product. I don't want to remove the `post_save` event, even for a product saved through a bulk process because for me when you listen this event you have to receive all saved products. But, no pb to discuss about from where we dispatch the event :)

src/Pim/Bundle/CatalogBundle/Doctrine/Common/Saver/ProductSaver.php
```diff
@@ -103,13 +109,34 @@ public function saveAll(array $products, array $options = [])
...
+            B.
+            Here is where the save function is called multiple times for a batch when we use mass edit.
+            The function save will dispatch post_save events without flushing the changes into the database.
+
+            If saveAll function would not base on save function, everything should work as I expect:
+            - saveAll will flush every batch and dispatch the post_save_all event, which I am listening when I do imports
+            or mass edits.
+            - save will flush every time and dispatch the events with the correct change set.
+            */
             $this->save($product, $itemOptions);
```
nidup: saveAll() may not use the save() method but,
 - dispatch a StorageEvents::PRE_SAVE for each product before the flush
 - dispatch a StorageEvents::POST_SAVE for each product after the flush

NB, not far from your original PR in fact :smile:

src/Pim/Bundle/CatalogBundle/Doctrine/Common/Saver/ProductSaver.php
```diff
@@ -103,13 +109,34 @@ public function saveAll(array $products, array $options = [])
...
 
         if (true === $options['flush']) {
             $this->objectManager->flush();
         }
-
+        /*
+         * C.
+         * 1. If we listen for both post_save and post_save_all events all the events are duplicated right now.
+         * - From post_save I have the event with a wrong change set.
+         * - From post_save_all I have the correct event.
```
nidup: based on previous comment, in case of mass edit, you'll still receive saved product in post_save_all and post_save but the change set will be ok in both case, and you'd be able to only the post_save event.

src/Pim/Bundle/CatalogBundle/Doctrine/Common/Saver/ProductSaver.php
```diff
@@ -103,13 +109,34 @@ public function saveAll(array $products, array $options = [])
...
+         *
+         * 2. If I change the save function to dispatch the event post_save only when the object is flushed (as in my
+         * previous PR) and listen for both post_save and post_save_all, everything is awesome.
+         * The only issue in that case is that on mass edit "Edit common attributes"
+         * "PimEnterprise\Bundle\WorkflowBundle\Doctrine\Common\Saver\DelegatingProductSaver" does not raise
+         * "pre/post_save_all" events at all on saveAll function.
```
nidup: Issue with the delegating saver is that, depending on your current user permissions, it will delegates the save to a product saver (the CE one), or to a product draft saver. The save all should dispatch a different event for saved drafts and for saved products. This Saver is especially complex.

nidup: Still digging in DelegatingProductSaver, doing interesting progress, I'll keep you posted monday.

src/Pim/Bundle/CatalogBundle/Doctrine/Common/Saver/ProductSaver.php
iulyanp: "I don't want to remove the post_save event, even for a product saved through a bulk process because for me when you listen this event you have to receive all saved products."
For products saved through a bulk process, you have all saved products when you listen post_save_all event. That's why I think it's fine to have post_save event only when you save a single product and post_save_all event when you save a bulk. 
But, if you want to keep also the post_save event for bulk process it should be dispatched after flush in order to have the version object real for both events (post_save, post_save_all).

src/Pim/Bundle/CatalogBundle/Doctrine/Common/Saver/ProductSaver.php
iulyanp: Yes, with this scenario I think the change set should be ok. I'll be in holiday next week but after I'll come back, I would like to help with this change, if you didn't already implement it until then.

src/Pim/Bundle/CatalogBundle/Doctrine/Common/Saver/ProductSaver.php
iulyanp: Thank you very much for your answers.
I saw about the user permissions and how DelegatingProductSaver delegates. I need only the case when the product it's saved (the CE one), so I could know that the product was updated. 
Then I can send an event to EventStore with the product changed. 
But in that case because post_save_all is not dispatched I receive only post_save event from mass edit "Edit common attributes" with the wrong change set. 
That's because the objects aren't flushed into the database before post_save is dispatched. The flush is done into the DelegatingProductSaver class in saveAll() method, without dispatching the post_save_all event.

src/Pim/Bundle/CatalogBundle/Doctrine/Common/Saver/ProductSaver.php
nidup: Ok, no pb. I'd love to work with you on this topic, you can re-work your original community edition PR to match the new scenario, I'll help you with tests and running CI to be able to merge it quickly. About contribution on EE it will be a bit more complicated but we'll find a way to discuss through "fake" PR on CE or gist files.

src/Pim/Bundle/CatalogBundle/Doctrine/Common/Saver/ProductSaver.php
nidup: once the post_save reworked to dispatch after the flush, we'll be ok with this use case too.

PR #4770, opened 2016-07-23
[TIP-435] Clean the css colors mess

pierallard:
1) Replace the #xxxxxx hardcoded colors by variables
2) Put in the variables.less or bootstrap_variables.less
3) Remove the similar colors


nidup:
such :heart: in this commit msg :laughing:

src/Pim/Bundle/DataGridBundle/Resources/public/css/filter.less
```diff
@@ -71,7 +71,7 @@
 
 .filter-select .select-filter-widget {
   font-weight: bold;
-  color: @color-333333;
+  color: @grayDark;
```
loicguillois: grey or not gray ? that is the question.

PR #4810, opened 2016-08-03
PIM-5900: Move backtogrid button and enable button to the title div, and fix css

pierallard:

capture du 2016-08-03 10 36 55

### DIFF
--- a/src/Pim/Bundle/EnrichBundle/Resources/views/CategoryTree/form.html.twig
+++ b/src/Pim/Bundle/EnrichBundle/Resources/views/CategoryTree/form.html.twig
@@ -52,13 +52,17 @@
                     (form.vars.value.parent ? 'flash.category.removed' : 'flash.tree.removed')|trans
                 ) }}
             {% endif %}
-        {% elseif parentId %}
-            {{ elements.backLink(path(route ~ '_categorytree_edit', { id: parentId }), 'btn.to parent') }}
         {% endif %}
         {{ elements.submitBtn('', 'ok') }}
     {% endset %}
 
-    {{ elements.page_header(title, buttons, null, null, elements.updated(form.vars.id)) }}
+    {% set titleButtons %}
+        {% if (form.vars.value.id is null) and parentId %}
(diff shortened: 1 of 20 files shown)


src/Pim/Bundle/EnrichBundle/Resources/views/CategoryTree/form.html.twig
```diff
@@ -52,13 +52,17 @@
...
             {% endif %}
-        {% elseif parentId %}
-            {{ elements.backLink(path(route ~ '_categorytree_edit', { id: parentId }), 'btn.to parent') }}
         {% endif %}
         {{ elements.submitBtn('', 'ok') }}
     {% endset %}
 
-    {{ elements.page_header(title, buttons, null, null, elements.updated(form.vars.id)) }}
+    {% set titleButtons %}
+        {% if (form.vars.value.id is null) and parentId %}
```
skeleton: why this button is only shown when id is null ?

src/Pim/Bundle/UIBundle/Resources/public/css/buttons.less
```diff
@@ -1,8 +1,8 @@
 @import 'less/variables';
 @import 'less/mixins';
 
-.navbar-extra .btn {
-  margin-left: 4px;
+.navbar-extra .btn+.btn {
```
skeleton: `.btn+.btn` ? can you explain it please, I didn't know this  ?

skeleton:
GTM

src/Pim/Bundle/EnrichBundle/Resources/views/CategoryTree/form.html.twig
pierallard: I moved the logic from the previous back-to-grid button, and it was a
if (form.vars.value.id)
...
else if parentId
-- it was here ---
endif

src/Pim/Bundle/UIBundle/Resources/public/css/buttons.less
pierallard: Explained IRL. For the others: 
http://www.w3schools.com/cssref/sel_element_pluss.asp

ClementGautier:
GTM

PR #4815, opened 2016-08-04
TIP-559: Rework exports: last enhancements

skeleton:
- [x] remove `mainContext` in the processors/writers and `filePathResolverOptions` in writers (it's only need for quick export)
- [x] build path file for quick export before launch the job
- [x] move Pim\Component\Connector\Filesystem\ZipFilesystemFactory to archivers as it's only used here

### DIFF
--- a/src/Pim/Bundle/DataGridBundle/Controller/ProductExportController.php
+++ b/src/Pim/Bundle/DataGridBundle/Controller/ProductExportController.php
@@ -22,6 +22,8 @@ use Symfony\Component\Security\Core\User\UserInterface;
  */
 class ProductExportController
 {
+    const DATETIME_FORMAT = 'Y-m-d_H:i:s';
+
     /** @var Request */
     protected $request;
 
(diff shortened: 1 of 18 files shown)


src/Pim/Bundle/DataGridBundle/Controller/ProductExportController.php
```diff
@@ -164,4 +161,22 @@ protected function getContextParameters()
...
+     * Build file path to replace pattern like %locale%, %scope% by real data
+     *
+     * @param string $filePath
+     * @param array  $contextParameters
+     *
+     * @return string
+     */
+    protected function buildFilePath($filePath, array $contextParameters)
+    {
+        $data = ['%datetime%' => date('Y-m-d_H:i:s')];
```
phaseinducer: It would be nice to have the format in conf somewhere, or at least in constant. What do you think?

phaseinducer:
Little comment then GTM :fire: 
I like the idea that file content and file name are two separate concerns (if I understood well).

src/Pim/Bundle/EnrichBundle/Connector/Job/JobParameters/ConstraintCollectionProvider/ProductQuickExport.php
```diff
@@ -41,9 +41,11 @@ public function getConstraintCollection()
         $baseConstraint = $this->simpleConstraint->getConstraintCollection();
         $constraintFields = $baseConstraint->fields;
         $constraintFields['filters'] = new NotBlank(['groups' => 'Execution']);
-        $constraintFields['mainContext'] = new NotBlank(['groups' => 'Execution']);
         $constraintFields['selected_properties'] = null;
         $constraintFields['with_media'] = new Type('bool');
+        $constraintFields['locale'] = new NotBlank(['groups' => 'Execution']);
+        $constraintFields['scope'] = new NotBlank(['groups' => 'Execution']);
+        $constraintFields['ui_locale'] = new NotBlank(['groups' => 'Execution']);
```
pierallard: Weird to move this.
In the other jobs, we have options like this: `{'filters': [], ..., 'context': {'locale': xxx, 'scope': yyy}}`

pierallard:
Little comment, maybe IRL discussion, then GTM

src/Pim/Bundle/EnrichBundle/Connector/Job/JobParameters/ConstraintCollectionProvider/ProductQuickExport.php
skeleton: seen IRL

PR #4823, opened 2016-08-06
Dispatch POST_SAVE event only after flush

iulyanp:
Hello @nidup 

As promised I worked on Saver classes to dispatch POST_SAVE event only after flushing the entities into the database. 
I'm not sure if this is how you expected to be but please let me know if you think I could refactor more. 
I would like to remove the "flush" option for good, but I don't know if that's ok with you. Also I think I could remove also the "schedule" option if we remove the "flush".  
For now I just added the POST_SAVE event to be dispatched after flush so we could have the correct  versioning change set when the event is raised.
If this is fine with you maybe you could merge it and I could refactor further and remove the flush option into another PR. 
I saw you have a new "akeneo_storage_utils.saver.base_options_resolver". I would like to use it for all the savers, but remove the flush option from it too.
Let me know what you think.

Regards,

### DIFF
--- a/src/Akeneo/Bundle/StorageUtilsBundle/Doctrine/Common/Saver/BaseSaver.php
+++ b/src/Akeneo/Bundle/StorageUtilsBundle/Doctrine/Common/Saver/BaseSaver.php
@@ -66,15 +66,16 @@ class BaseSaver implements SaverInterface, BulkSaverInterface
         }
 
         $options = $this->optionsResolver->resolveSaveOptions($options);
+
         $this->eventDispatcher->dispatch(StorageEvents::PRE_SAVE, new GenericEvent($object, $options));
 
         $this->objectManager->persist($object);
 
         if (true === $options['flush']) {
             $this->objectManager->flush();
-        }
 
-        $this->eventDispatcher->dispatch(StorageEvents::POST_SAVE, new GenericEvent($object, $options));
+            $this->eventDispatcher->dispatch(StorageEvents::POST_SAVE, new GenericEvent($object, $options));
+        }
     }
 
     /**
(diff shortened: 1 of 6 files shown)


nidup:
Hi @iulyanp !

:tada: Thank you very much for your pull request :tada: 

I cherry picked your commit in my own branch to do some extra tests https://github.com/akeneo/pim-community-dev/pull/4824/commits Then I've added specs (ORM & MongoDB) and I run it with success on our internal CI with all our behat builds (Enterprise Edition, Community Edition, ORM storage, MongoDB storage) :dancer: 

I do agree with you to merge this first re-work, I just merged the related PR :tada: 

You can open a second PR to remove the "flush" option in all Saver and BulkSaver, I quickly checked in the code, I'm pretty sure we can remove it without any issues as we always use it with default value "true" or by passing "true" (I will write the related PR to remove the "flush" option in Enterprise Edition).

About "schedule" and "recalculate", I'm taking a look to see the impacts or side effects we could encounter if we drop it.

We could also remove the "flush" option from the Removers in another PR.

Thank you again for your work on this topic, it really help us to improve the storage API, don't hesitate to ping me on upcoming PR :)

Have a good day!

PS: don't hesitate to subscribe on http://badger.akeneo.com/badges we're adding some badges related to contributions and I will ping you when these different badges will be available :)

iulyanp:
Hi @nidup 

That's great news! 😃

I will work on the flush option from Savers asap. In order to do this some code from save method will be duplicated in saveAll as we've discussed in my first PR.

About "schedule" and "recalculate", please let me know if it affects anything in case we drop them and use the "akeneo_storage_utils.saver.base_options_resolver". From what I saw "schedule" could be dropped, because it is always used with the default "true" option.

About Removers, I just saw that there is the same situation with POST_SAVE. I'll work on removers also to dispatch POST_SAVE after flush in another PR.

Thank you for everything.
Have a great day!

Best regards

nidup:
Hi @iulyanp 

Cool! No pb to duplicate some code between save and saveAll (or extract it in a protected method).

I double checked and the options "schedule" and "recalculate" can be totally dropped from the ProductSaver. You're right, the ProductSaver will be able to use the "akeneo_storage_utils.saver.base_options_resolver"  (as we'll drop the flush option from here in the same time).

Once your PR done, I'll re-work the DelegatingProductSaver in Enterprise Edition to make sure the whole stack works in the same way.

Don't hesitate to ping me if you need any further information :)

Best regards,

PR #4831, opened 2016-08-10
Fix quick export of all products

SamirBoulil:
**Issue:**

When we select all entities in the product grid and quick export, the job fails because the job parameter "filters" is an empty array.

The PIM fails to validate the constraint NotBlank on this job parameter.

**Solution:**

Set the same constraint as the CSV Product export for the "filters" jobParameter

___

**This PR also contains:**

- Fixes an issue with the quick export when a warning would be thrown about a media not found.
- Reverts the quick-export migrations (seen with @aRn0D) 

___

### DIFF
--- a/features/mass-action/quick_export.feature
+++ b/features/mass-action/quick_export.feature
@@ -14,7 +14,29 @@ Feature: Quick export many products from datagrid
       | pump     |          | summer_collection | Pump          | 15 EUR, 20 USD | 41   | blue  | ddd |
     And I am logged in as "Julia"
 
-  Scenario: Successfully quick export products as a CSV file
+  Scenario: Successfully quick export all products as a CSV file
+    Given I am on the products page
+    When I select all entities
+    And I press "CSV (All attributes)" on the "Quick Export" dropdown button
+    And I wait for the "csv_product_quick_export" quick export to finish
+    And I am on the dashboard page
+    Then I should have 1 new notification
+    And I should see notification:
+      | type    | message                                        |
+      | success | Quick export CSV product quick export finished |
+    When I go on the last executed job resume of "csv_product_quick_export"
+    Then I should see the text "COMPLETED"
+    And the name of the exported file of "csv_product_quick_export" should be "products_export_en_US_tablet.csv"
+    And exported file of "csv_product_quick_export" should contain:
+    """
+    sku;123;categories;color;description-en_US-tablet;enabled;family;groups;lace_color;manufacturer;name-en_US;price-EUR;price-USD;rating;side_view;size;top_view;weather_conditions
+    boots;aaa;winter_collection;black;;1;boots;;;;"Amazing boots";20;25;;;40;;
+    sneakers;bbb;summer_collection;white;;1;sneakers;;;;Sneakers;50;60;;;42;;
+    sandals;ccc;summer_collection;red;;1;sandals;;;;Sandals;5;5;;;40;;
+    pump;ddd;summer_collection;blue;;1;;;;;Pump;15;20;;;41;;
+    """
+
+  Scenario: Successfully quick export selected products as a CSV file
     Given I am on the products page
     When I select rows boots, sneakers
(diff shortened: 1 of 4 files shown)


src/Pim/Bundle/EnrichBundle/Connector/Job/JobParameters/ConstraintCollectionProvider/ProductQuickExport.php
```diff
@@ -40,7 +40,7 @@ public function getConstraintCollection()
     {
         $baseConstraint = $this->simpleConstraint->getConstraintCollection();
         $constraintFields = $baseConstraint->fields;
-        $constraintFields['filters'] = new NotBlank(['groups' => 'Execution']);
+        $constraintFields['filters'] = [];
```
ClementGautier: what about replacing it by NotNull instead ? it would ensure that we at least have an array.

ClementGautier:
GTM after comments fixed (try NotNull, if it doesn't work as expected for some reason, get back to "no validation" and merge as it)

src/Pim/Bundle/EnrichBundle/Connector/Job/JobParameters/ConstraintCollectionProvider/ProductQuickExport.php
SamirBoulil: For some reason it didn't work with NotNull (but I can search that). 

Still I'd rather stay consistent with other normal Product export constraint what do you think ?

upgrades/schema/Version_1_6_20160722155535_quick_export.php
```diff
@@ -21,34 +21,26 @@ class Version_1_6_20160722155535_quick_export extends AbstractMigration
      */
```
SamirBoulil: Revert to previous commit. Seen with @aRn0D

features/mass-action/quick_export.feature
```diff
@@ -14,7 +14,29 @@ Feature: Quick export many products from datagrid
...
+    When I select all entities
+    And I press "CSV (All attributes)" on the "Quick Export" dropdown button
+    And I wait for the "csv_product_quick_export" quick export to finish
+    And I am on the dashboard page
+    Then I should have 1 new notification
+    And I should see notification:
+      | type    | message                                        |
+      | success | Quick export CSV product quick export finished |
+    When I go on the last executed job resume of "csv_product_quick_export"
+    Then I should see "COMPLETED"
```
phaseinducer: I should see the text

src/Pim/Bundle/EnrichBundle/Connector/Processor/QuickExport/ProductProcessor.php
```diff
@@ -159,7 +160,7 @@ protected function fetchMedia(ProductInterface $product, $directory)
         $this->mediaFetcher->fetchAll($product->getValues(), $directory, $identifier);
 
         foreach ($this->mediaFetcher->getErrors() as $error) {
-            $this->stepExecution->addWarning($error['message'], [], $error['media']);
+            $this->stepExecution->addWarning($error['message'], [], new DataInvalidItem($error['media']));
```
phaseinducer: Can you explain?

phaseinducer:
GTM

src/Pim/Bundle/EnrichBundle/Connector/Processor/QuickExport/ProductProcessor.php
SamirBoulil: Yep! The interface of addWarning has changed and now waits for a DataInvalidItem instance. I reported the change that has already taken place in other classes like here https://github.com/akeneo/pim-community-dev/blob/cede6217ee8133c38264cf19b39a69980f110577/src/Pim/Component/Connector/Processor/Normalization/ProductProcessor.php#L125-L125

PR #4863, opened 2016-08-22
TIP-568: reduce time import of products

jjanvier:
This PR aims to reduce the time of product imports. There are two axes of improvements in this PR:

1. reduce the time spent in the completeness calculation. Adding the completeness calculation during the save of a product has a huge impact on computer with non SSD drives.
2. reduce the time spent in greedy methods (lot of CPU time) or methods that are called a very lot of times 

Note: this PR treats only the import of new products. We currently have a memory leak on the import of MongoDB updated products. @skeleton is taking care of that :)

Note: the benchmarks you'll see below have been realized on the _core-perf_ machine.

## The state of products import before this PR

Internally, we changed quite a lot of things between the products import 1.5 and the products import master:
* the processors now receives the standard format
* which means the readers transform flat format to standard format (and delocalize data)
* spout is used to read CSV/XLSX
* *the completeness is now calculated at each save*
* and a few more..

Those changes make difficult to analyse with blackfire what's better or worse than before, because the paths of the graphs are totally different.

mongo

Calculating the completeness at each save has no impact on the MongoDB imports. That's because, by design, the MongoDB completeness is simple to caclulate. Even better, our changes tend to improve a little bit (~10%) the import time.

orm

On ORM, that's a different story! The completeness calculation has a huge impact on the time to import. And the more the table grows, the more the completeness calculation is greedy. For 500 products, the master import is ~2.2 times slower while for 10 000 products it's ~2.7 times slower.

You can see in orange the master without the completeness calculation. We import products on the same time basis that what we had on 1.5. 

## The ORM completeness improvements
 
The ORM completeness was, by design, effective for the calculation of a lot of products at the same time. Each time the completeness was calculated we:

1. created a temporary table with all missing completeness (for the set of products we selected)
2. created a temporary table with all complete (for the set of products we selected)
3. performed a complicated query that joins on those two tables

Creating the temporary tables consumes a lot of IO waits on non SSD disks (like what we have on the core-perf server). *A lot!* The impact are much less important on machines with SSD disks.

So basically, what's done in this PR is that when we calculate the completeness for one product (ie: each time we save a product), we do not create those temporary tables. We just join on the data we need.
The temporary tables are still used when we calculate the completeness of a vast amount of products at the same time (completeness:calculate command, change attribute requirements of a family etc...). Without those temporary tables, the completeness calculations takes too much memory for a vast number of products.

## The others minor improvements

Other improvements consist of:
* caching things that are called a lot of times during an import (example: 1 000 products imported => 1 000 000 calls to the method)
   * really cache the association columns resolver
   * cache the value converter registry
* improving the code of greedy methods
   * do not use `in_array` in the product flat to standard array converter

## The state of products import after this PR

This PR (green dots, branch tweak_completeness) improves performances of products import. 

mongo

On MongoDB this branch is slightly faster than 1.5 and master due to the caching improvements and code simplifications.

orm

On ORM this branch allows us to have the same timings that the 1.5 branch, even if we do calculate the completeness. We are ~65% faster than the current master. Please keep in mind that these branch is particularly effective for non SSD drives. On SSD drives, the gain will be lower, but still significant.

### DIFF
--- a/src/Pim/Bundle/CatalogBundle/Doctrine/ORM/CompletenessGenerator.php
+++ b/src/Pim/Bundle/CatalogBundle/Doctrine/ORM/CompletenessGenerator.php
@@ -98,8 +98,8 @@ class CompletenessGenerator implements CompletenessGeneratorInterface
      */
     protected function generate(array $criteria = [])
     {
-        $this->prepareCompletePrices($criteria);
         if (!isset($criteria['productId'])) {
+            $this->prepareCompletePrices($criteria);
             $this->prepareMissingCompletenesses($criteria);
         }
 
@@ -190,7 +190,7 @@ class CompletenessGenerator implements CompletenessGeneratorInterface
     protected function getCompletePricesSQL()
     {
         return <<<COMPLETE_PRICES_SQL
-            SELECT l.id AS locale_id, c.id AS channel_id, v.id AS value_id
+            (SELECT l.id AS locale_id, c.id AS channel_id, v.id AS value_id
                 FROM pim_catalog_attribute_requirement r
                 JOIN %attribute_table% att ON att.id = r.attribute_id AND att.backend_type = "prices"
                 JOIN pim_catalog_channel c ON c.id = r.channel_id %channel_conditions%
@@ -208,7 +208,7 @@ class CompletenessGenerator implements CompletenessGeneratorInterface
                     ON price.value_id = v.id
                     AND price.currency_code = cur.code
                 GROUP BY l.id, c.id, v.id
-                HAVING COUNT(price.data) = COUNT(ccur.currency_id)
+                HAVING COUNT(price.data) = COUNT(ccur.currency_id))
 COMPLETE_PRICES_SQL;
     }
 
@@ -345,11 +345,16 @@ MISSING_SQL;
                 AND v.entity_id = p.id
             %product_value_joins%
             %extra_joins%
+            LEFT JOIN %complete_price% AS cp
+                ON cp.value_id = v.id
+                AND cp.channel_id = c.id
+                AND cp.locale_id = l.id
             LEFT JOIN pim_catalog_attribute_locale al ON al.attribute_id = v.attribute_id
 
             WHERE (
                 %product_value_conditions%
                 %extra_conditions%
+                OR cp.value_id IS NOT NULL
             )
             AND (al.locale_id = l.id OR al.locale_id IS NULL)
             AND r.required = true
@@ -374,10 +379,12 @@ MAIN_SQL;
             '%product_value_joins%'      => implode(' ', $this->getProductValueJoins()),
             '%extra_joins%'              => implode(' ', $this->getExtraJoins()),
             '%extra_conditions%'         => implode(' ', $this->getExtraConditions()),
-            '%missing_completeness%'     =>
-                isset($criteria['productId']) ?
+            '%missing_completeness%'     => isset($criteria['productId']) ?
                     $this->getMissingCompletenessesSQL() :
-                    self::MISSING_TABLE
+                    self::MISSING_TABLE,
+            '%complete_price%'           => isset($criteria['productId']) ?
+                    $this->getCompletePricesSQL() :
+                    self::COMPLETE_PRICES_TABLE,
         ];
     }
 
@@ -705,14 +712,7 @@ SQL;
      */
     protected function getExtraJoins()
     {
-        $pricesJoin = 'LEFT JOIN %s AS complete_price
-            ON complete_price.value_id = v.id
-            AND complete_price.channel_id = c.id
-            AND complete_price.locale_id = l.id';
-
-        $pricesJoin = sprintf($pricesJoin, static::COMPLETE_PRICES_TABLE);
-
-        return [$pricesJoin];
+        return [];
     }
 
     /**
@@ -720,8 +720,6 @@ SQL;
      */
     protected function getExtraConditions()
     {
-        $pricesConditions = sprintf('OR %s.value_id IS NOT NULL', static::COMPLETE_PRICES_TABLE);
-
-        return [$pricesConditions];
+        return [];
     }
 }
--- a/src/Pim/Bundle/CatalogBundle/Doctrine/ORM/CompletenessGenerator.php
+++ b/src/Pim/Bundle/CatalogBundle/Doctrine/ORM/CompletenessGenerator.php
@@ -85,14 +85,23 @@ class CompletenessGenerator implements CompletenessGeneratorInterface
 
     /**
      * Generate completeness for product where it's missing,
-     * applying the criteria if provided to reduce the product set
+     * applying the criteria if provided to reduce the product set.
+     *
+     * If the completeness is being calculated for only one product, we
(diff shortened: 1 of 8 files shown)


Milie44:
Allez là!

src/Pim/Component/Connector/spec/ArrayConverter/FlatToStandard/ProductSpec.php
```diff
@@ -150,12 +227,29 @@ function it_converts(
...
+
+            ],
+            'price' => [
+                'locale' => '',
+                'scope'  => '',
+                'data'   => [['data' => 15, 'currency' => 'EUR'], ['data' => 10, 'currency' => 'USD']]
+            ],
+            'associations' => [
+                'X_SELL' => [
+                    'groups' => ['group-A'],
```
SamirBoulil: align

src/Pim/Bundle/CatalogBundle/Doctrine/ORM/CompletenessGenerator.php
```diff
@@ -263,15 +272,19 @@ protected function applyCriteria($sql, $criteria)
     /**
      * Get the sql query to insert completeness
      *
+     * @param array
```
damien-carcel: @param array $criteria

src/Pim/Component/Connector/ArrayConverter/FlatToStandard/Product.php
```diff
@@ -292,24 +292,24 @@ protected function convertValue($column, $value)
     }
 
     /**
-     * Merge the structured value inside the passed collection
+     * Merge the structured value inside the passed item
      *
-     * @param array $collection The collection in which we add the element
-     * @param array $value      The structured value to add to the collection
+     * @param array $item   The item in which we add the element
+     * @param array $value  The structured value to add to the collection
```
damien-carcel: to add to the item ?

damien-carcel:
Small doc comments, GTM!

src/Pim/Bundle/CatalogBundle/Doctrine/ORM/CompletenessGenerator.php
```diff
@@ -685,23 +712,14 @@ public function scheduleForChannelAndLocale(ChannelInterface $channel, LocaleInt
...
     {
-        $pricesJoin = 'LEFT JOIN %s AS complete_price
-            ON complete_price.value_id = v.id
-            AND complete_price.channel_id = c.id
-            AND complete_price.locale_id = l.id';
-
-        $pricesJoin = sprintf($pricesJoin, static::COMPLETE_PRICES_TABLE);
-
-        return [$pricesJoin];
+        return [];
```
willy-ahva: Why didn't you just remove the method ?

src/Pim/Component/Connector/ArrayConverter/FlatToStandard/Product.php
```diff
@@ -337,9 +337,12 @@ protected function validateOptionalFields(array $item)
             $this->getOptionalAssociationFields()
         );
 
+        // index $optionalFields by keys to improve performances
```
willy-ahva: :+1: Good to know !

src/Pim/Component/Connector/spec/ArrayConverter/FlatToStandard/Product/AssociationColumnsResolverSpec.php
```diff
@@ -30,4 +30,11 @@ function it_resolves_association_type_field_names(
             ]
         );
     }
+
+    function it_cache_the_associations_even_without_results($assocTypeRepo)
```
willy-ahva: it_caches

src/Pim/Bundle/CatalogBundle/Doctrine/ORM/CompletenessGenerator.php
nidup: because overriden in EE...

src/Pim/Bundle/ConnectorBundle/DependencyInjection/Compiler/RegisterFlatToStandardConverterPass.php
```diff
@@ -46,23 +46,18 @@ protected function registerConverters(ContainerBuilder $container)
         $converters = $container->findTaggedServiceIds(static::CONVERTER_TAG);
 
         foreach ($converters as $serviceId => $tags) {
-            $this->registerConverter($registry, $serviceId, $tags);
+            foreach ($tags as $tag) {
+                $priority = isset($tag['priority']) ? $tag['priority'] : static::DEFAULT_PRIORITY;
```
nidup: ok

nidup:
GTM

jjanvier:
Before your comments the CI was green. As they are only related to the phpdoc, I will not relaunch the CI.

green ci

src/Pim/Component/Connector/spec/ArrayConverter/FlatToStandard/ProductSpec.php
jjanvier: done

src/Pim/Bundle/CatalogBundle/Doctrine/ORM/CompletenessGenerator.php
jjanvier: done

src/Pim/Component/Connector/ArrayConverter/FlatToStandard/Product.php
jjanvier: done

src/Pim/Component/Connector/spec/ArrayConverter/FlatToStandard/Product/AssociationColumnsResolverSpec.php
jjanvier: done

src/Pim/Bundle/CatalogBundle/Doctrine/ORM/CompletenessGenerator.php
```diff
@@ -321,7 +334,7 @@ protected function getMainSqlPart()
                         AND (al.locale_id = l.id OR al.locale_id IS NULL)
                 ) AS required_count
 
-            FROM missing_completeness m
+            FROM %missing_completeness% AS m
```
BitOne: Not required, but why not ;)

src/Pim/Component/Connector/ArrayConverter/FlatToStandard/Product.php
```diff
@@ -251,8 +251,8 @@ protected function convertItem(array $item)
                 $value = $this->convertValue($column, $value);
             }
 
-            if (null !== $value) {
-                $convertedItem = $this->mergeValueToResult($convertedItem, $value);
+            if (!empty($value)) {
```
BitOne: Are you sure about that? A string "" is considered as empty, as well as an empty array, as wel as 0, etc...

jjanvier: yes, only arrays are returned
never null values...

PR #4871, opened 2016-08-23
FileIterator: close reader only if not null

jmleroux:
The internal reader can be null if an exception occured before its instanciation.

| Review and 2 GTM                  |yes


skeleton:
gtm

PR #4873, opened 2016-08-23
PIM-5928: Export products without media

aRn0D:
When you export products without media we remove all media attributes from the generated file.

### DIFF
--- a/src/Pim/Component/Connector/Processor/Normalization/ProductProcessor.php
+++ b/src/Pim/Component/Connector/Processor/Normalization/ProductProcessor.php
@@ -186,6 +186,6 @@ class ProductProcessor implements ItemProcessorInterface, StepExecutionAwareInte
     protected function areAttributesToFilter(JobParameters $parameters)
     {
         return isset($parameters->get('filters')['structure']['attributes'])
-        && !empty($parameters->get('filters')['structure']['attributes']);
+            && !empty($parameters->get('filters')['structure']['attributes']);
     }
 }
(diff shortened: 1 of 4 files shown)


src/Pim/Component/Connector/Processor/Normalization/ProductProcessor.php
```diff
@@ -98,6 +98,15 @@ public function process($product)
                 ->get(JobInterface::WORKING_DIRECTORY_PARAMETER);
 
             $this->fetchMedia($product, $directory);
+        } else {
+            $mediaAttributes = $this->attributeRepository->findMediaAttributeCodes();
+            $productStandard['values'] = array_filter(
```
phaseinducer: What about `$productStandard['values'] = array_intersect_key($productStandard['values'], array_flip($mediaAttributes))`
?

phaseinducer: What about:
```
$productStandard['values'] = array_diff_key($productStandard['values'], array_flip($mediaAttributes))
```
?
Dunno what is the most readable :/

damien-carcel:
Same comment than @phaseinducer , otherwise GTM

src/Pim/Component/Connector/Processor/Normalization/ProductProcessor.php
aRn0D: I'd rather prefer to use `filter`, `reduce` and `map` functions to manipulate array or collection. No need to call several functions, we just need to apply a pure function on your collection.

phaseinducer:
GTM

SamirBoulil:
GTM

PR #4885, opened 2016-08-24
TIP-568: Fix memory leak on mongo

skeleton:
A memory leak has been detected on mongo when we import products which are already in database.

When the completeness is calculated, we refresh the product to have the new values in the Product entity. But when we detach the product, doctrine does not find the product values linked to the product, so product values are never detached and are kept in memory.
So we removed `refresh()` after the calcul of the completeness.
Problem, some behats failed when we removed it: when we change the requirements of a family and come back to the PEF, the panel of the completeness is not shown... because product has not the new values of the completeness... 
The real solution is to call a job after change the requirements of a family to calculate the completeness for all products linked to it, but it's too late (release is the next week :) ). So we keep the simpliest solution, call refresh when the panel is shown and only on mongo :tada: 

Last thing, we removed code added in previous versions to fix a bug in MongoDB ODM BETA12. As we use now the MongoDB ODM 1.1, this code is useless.

Result of the bench of 1000 products imported with medium catalog:

 | time | memory used
 ------------- | ------------ | -------------
before fix | 40'06" | 2300Mo 
after fix | 6'46" | 311 Mo


src/Pim/Bundle/CatalogBundle/Doctrine/Common/Saver/GroupSaver.php
```diff
@@ -218,6 +224,7 @@ protected function persistGroup($group, array $options)
 
         if ($group->getType()->isVariant() && true === $options['copy_values_to_products']) {
             $this->copyVariantGroupValues($group);
+            $this->detacher->detachAll($group->getProducts()->toArray());
```
jjanvier: will it work with a high number of products in the group?

shouldn't we detach them one by one?

src/Pim/Bundle/EnrichBundle/Controller/Rest/CompletenessController.php
```diff
@@ -77,6 +90,10 @@ public function getAction($id)
 
         $this->completenessManager->generateMissingForProduct($product);
 
+        if ('doctrine/mongodb-odm' === $this->storageDriver) {
```
jjanvier: we have a constant for that

jjanvier: maybe we should add a comment?

jjanvier:
gtm

src/Pim/Bundle/CatalogBundle/Doctrine/Common/Saver/GroupSaver.php
skeleton: they are already detach one by one in out detacher: https://github.com/akeneo/pim-community-dev/blob/master/src/Akeneo/Bundle/StorageUtilsBundle/Doctrine/Common/Detacher/ObjectDetacher.php#L61

src/Pim/Bundle/EnrichBundle/Controller/Rest/CompletenessController.php
phaseinducer: https://github.com/akeneo/pim-community-dev/blob/master/src/Akeneo/Bundle/StorageUtilsBundle/DependencyInjection/AkeneoStorageUtilsExtension.php#L23

nidup: yep for these kind of ugly stuff a comment may help the next one to understand this :hankey: could also reference a TIP related to completeness

src/Akeneo/Bundle/StorageUtilsBundle/Doctrine/Common/Detacher/ObjectDetacher.php
```diff
@@ -177,38 +177,5 @@ protected function doDetach($document, array &$visited)
...
-
-    /**
-     * Cascade detach objects to overcome MongoDB detach
-     * cascade bug on MongoDB ODM BETA12.
-     * See https://github.com/doctrine/mongodb-odm/pull/979.
-     *
-     * @param mixed $object
-     * @param array $visited Prevents infinite recursion
-     */
-    protected function cascadeDetach($document, array &$visited)
```
nidup: :fire: :+1:

nidup:
gtm

src/Pim/Bundle/EnrichBundle/Controller/Rest/CompletenessController.php
```diff
@@ -77,6 +91,12 @@ public function getAction($id)
 
         $this->completenessManager->generateMissingForProduct($product);
 
+        // Product have to be refresh to have the completeness values generated by generateMissingForProduct()
+        // (on ORM, completeness is not calculated the same way and product doesn't need to be refresh)
```
phaseinducer: refreshed?

phaseinducer:
GTM

loicguillois:
:+1:

PR #4946, opened 2016-08-30
PIM-5952: Add default message in empty family filter

damien-carcel:
**Description**

In the product export builder, the field of the family filter is empty if no family are selected. This is confusing because the user don't know what the empty field means.

This PR fixes this problem by adding two default message: one when showing the export profile in read only, another when editing the export profile.

**Definition Of Done**

### DIFF
--- a/CHANGELOG-1.6.md
+++ b/CHANGELOG-1.6.md
@@ -10,6 +10,7 @@
 - PIM-5959: Change wording of XLSX to Excel on mass actions
 - PIM-5938: Fix typos in import/export tooltips
 - PIM-5966: Fix style about category filter on product export builder
+- PIM-5952: Add the information "No condition on families" in the family field of the Export Builder
 
 # 1.6.0 (2016-08-30)
 
--- a/features/Pim/Behat/Context/JobContext.php
+++ b/features/Pim/Behat/Context/JobContext.php
@@ -8,10 +8,13 @@ use Akeneo\Component\Batch\Job\JobRegistry;
 use Akeneo\Component\Batch\Model\JobInstance;
 use Behat\Behat\Context\Step;
 use Behat\Gherkin\Node\TableNode;
+use Context\Spin\SpinCapableTrait;
 use Symfony\Component\Validator\ConstraintViolationInterface;
 
 class JobContext extends PimContext
 {
+    use SpinCapableTrait;
+
     /**
      * @param string    $code
      * @param TableNode $table
@@ -104,14 +107,14 @@ class JobContext extends PimContext
      * @param string      $action
      * @param JobInstance $job
      *
-     * @return \Behat\Behat\Context\Step\Then
+     * @return Step\Then
      *
      * @When /^I should not be able to (launch|edit) the ("([^"]*)" (export|import) job)$/
      */
     public function iShouldNotBeAbleToAccessTheJob($action, JobInstance $job)
     {
         $this->currentPage = sprintf("%s %s", ucfirst($job->getType()), $action);
-        $page              = $this->getCurrentPage()->open(['id' => $job->getId()]);
+        $this->getCurrentPage()->open(['id' => $job->getId()]);
 
         return new Step\Then('I should see "403 Forbidden"');
     }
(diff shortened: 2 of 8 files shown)


CHANGELOG-1.6.md
```diff
@@ -1,4 +1,10 @@
-# 1.6.x
+# 1.6.1
```
phaseinducer: Don't change it please, the patching script do it automatically.

CHANGELOG-1.6.md
```diff
@@ -1,4 +1,10 @@
...
+
+## Functional improvements
+
+- PIM-5952: Add the information "No condition on families" in the family field of the Export Builder
+
+# 1.6.0 (2016-08-30)
```
phaseinducer: shouldn't be here

features/Context/Page/Base/Index.php
```diff
@@ -9,7 +9,7 @@
  * @copyright 2013 Akeneo SAS (http://www.akeneo.com)
  * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
  */
-class Index extends Base
+class Index extends Form
```
phaseinducer: Can you explain please?

damien-carcel: Yes, I can. I need the method `iVisitTheTab` tab", which is in `Form`. I only `Base` is extended, then I cannot visit a tab when in "Show" mode on a job profile.

features/Pim/Behat/Context/JobContext.php
```diff
@@ -127,6 +130,24 @@ public function iTryToCreateAnUnknownJob($jobType)
     }
 
     /**
+     * @param string $label
+     * @param string $message
+     *
+     * @Then /^the export content field "(?P<label>(?:[^"]|\\")*)" should contain "(?P<message>(?:[^"]|\\")*)"$/
```
phaseinducer: Very complicated pattern :/
`/^the export content field "([^"]*)" should contain "([^"]*)"$/` should do the work.

features/Pim/Behat/Context/JobContext.php
```diff
@@ -127,6 +130,24 @@ public function iTryToCreateAnUnknownJob($jobType)
...
+     */
+    public function theExportContentFieldShouldContain($label, $message)
+    {
+        $page = $this->getCurrentPage();
+        $field = $this->spin(function () use ($label, $page) {
+            return $page->find('css', sprintf('[data-name="%s"]', $label));
+        }, sprintf('Field "%s" not found.', $label));
+
+        $hiddenInput = $field->find('css', 'input[type="hidden"]')->getOuterHtml();
```
phaseinducer: Can't we use a decorator for all that stuff?

src/Pim/Bundle/EnrichBundle/Resources/public/js/filter/product/family.js
```diff
@@ -1,16 +1,15 @@
...
 define([
-        'underscore',
-        'oro/translator',
-        'pim/filter/filter',
-        'routing',
-        'text!pim/template/filter/product/family',
-        'pim/fetcher-registry',
-        'pim/user-context',
-        'pim/i18n',
-        'jquery.select2'
```
phaseinducer: Don't remove it please, it's normal. It means that we require the lib, but since it's not a class or an instance we don't map it to a param. We always do that for jQuery plugins for example.

features/Context/Page/Base/Index.php
phaseinducer: Ok but what you're saying by doing this is basically "the index page is a form". It's not true, so it's not the solution imho.

damien-carcel: Indeed. Then I have to move all tabs methods in `Base`.

src/Pim/Bundle/EnrichBundle/Resources/public/js/filter/product/family.js
damien-carcel: Added again.

grena:
GTM

phaseinducer:
GTM, good job!

PR #4947, opened 2016-08-30
Adapt AbstractInvalidItemWriter to write xml files

SamirBoulil:
This PR is a PoC supporting the writing of the cookbook dedicated to writing an XML connector for products.

This bundle contains :
- Xml Product reader: to read products out of the xml file
- Xml File iterator: to iterate over the same file to get the invalid lines
- Xml Writer: to write the invalid items
- XmlInvalidItemWriter class: to iterate over the invalid lines found by the processor

Some simplifications may be possible. let me know if you find any :)

---

**Problems:**
- The `stepExecution->summary['read_lines']` flag is misused preventing us from getting the real invalid line number
  - https://github.com/akeneo/pim-community-dev/blob/master/src/Pim/Component/Connector/Processor/Denormalization/AbstractProcessor.php#L95)
  - https://github.com/akeneo/pim-community-dev/blob/master/src/Pim/Component/Connector/Processor/Denormalization/AbstractProcessor.php#L121
- This implementation does not support nodes having different attributes specified
  - See the array_combine of the headers and the data here https://github.com/akeneo/pim-community-dev/blob/master/src/Pim/Component/Connector/Archiver/AbstractInvalidItemWriter.php#L101
  - Although I could override the `archive()` function into my XmlInvalidItemWriter to get the headers for each nodes (in the for loop instead of before). I believe it is not clear for the integrators why we would do that.

_won't be merged_


src/Acme/Bundle/XmlConnectorBundle/Reader/File/XmlFileIterator.php
```diff
@@ -0,0 +1,129 @@
...
+        return $this->archivePath;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getHeaders()
+    {
+        $headers = [];
+        foreach ($this->xmlFileIterator->current()->attributes() as $header => $value) {
```
SamirBoulil: We suppose that all XML element have the same attributes.

src/Acme/Bundle/XmlConnectorBundle/Reader/File/XmlProductReader.php
```diff
@@ -0,0 +1,122 @@
...
+
+    public function read()
+    {
+        if (null === $this->xml) {
+            $jobParameters = $this->stepExecution->getJobParameters();
+            $filePath = $jobParameters->get('filePath');
+            // for example purpose, we should use XML Parser to read line per line
+            $this->xml = simplexml_load_file($filePath, 'SimpleXMLIterator');
+            $this->xml->rewind();
+            $this->stepExecution->addSummaryInfo('read_lines', -1);
```
SamirBoulil: Because of the `AbstractProcessor::skipItemWithConstraintViolations` as well as `AbstractProcessor:: skipItemWithMessage` play with the read_line index to guess the position of the invalid item in the file.

I guess the +1 is because of the header line that is not taken into account in the ProductReader. that's why the -1.

 https://github.com/akeneo/pim-community-dev/blob/master/src/Pim/Component/Connector/Processor/Denormalization/AbstractProcessor.php#L95

app/AppKernel.php
```diff
@@ -48,6 +48,8 @@ public function registerBundles()
             $bundles
         );
 
+        $bundles[] = new Acme\Bundle\XmlConnectorBundle\AcmeXmlConnectorBundle();
```
damien-carcel: should be in the method `registerProjectBundles`

PR #4962, opened 2016-09-01
Fix coverage of specs on CI

skeleton:
Since we updated "phpspec/phpspec" to 3.0, build on our CI to test corevage does not work.

- [x] Update phpunit to `5.4.*` (not the last version because of one of these requirements is php >= 7.0) 
- [x] Fix phpunit tests
- [x] Add all suites test to calculate the coverage

### DIFF
--- a/app/phpspec.jenkins-coverage.yml
+++ b/app/phpspec.jenkins-coverage.yml
@@ -1,7 +1,0 @@
-formatter.name: dot
-extensions:
-    - Akeneo\SkipExampleExtension
-    - PhpSpec\Extension\CodeCoverageExtension
-code_coverage:
-    format: php
-    output: app/build/coverage-parts/phpspec.php
--- a/composer.json
+++ b/composer.json
@@ -61,7 +61,7 @@
         "box/spout": "2.5.0"
     },
     "require-dev": {
-        "phpunit/phpunit": "3.7.*",
+        "phpunit/phpunit": "5.4.*",
         "squizlabs/php_codesniffer": "2.*",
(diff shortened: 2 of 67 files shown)


ClementGautier:
GTM

app/phpspec.jenkins-coverage.yml
```diff
@@ -1,7 +1,192 @@
 formatter.name: dot
```
nidup: we can't use phpspec.yml.dist to avoid to duplicate the conf

composer.json
```diff
@@ -61,7 +61,7 @@
         "box/spout": "2.5.0"
     },
     "require-dev": {
-        "phpunit/phpunit": "3.7.*",
+        "phpunit/phpunit": "5.4.*",
```
nidup: :+1:

composer.json
```diff
@@ -79,7 +79,8 @@
         "sensiolabs/behat-page-object-extension": "1.0.1",
         "phpspec/phpspec": "3.0.0",
         "akeneo/phpspec-skip-example-extension": "2.0.0",
-        "akeneo/php-coupling-detector": "dev-master"
+        "akeneo/php-coupling-detector": "dev-master",
+        "henrikbjorn/phpspec-code-coverage": "3.0.*"
```
nidup: ok like for https://github.com/akeneo/ExcelInitBundle/blob/master/composer.json#L29 for instance

nidup:
question about phpspec conf then GTM

PR #4974, opened 2016-09-02
Fix glitch on flash messages

phaseinducer:
This PR fixes a regression between 1.5 and 1.6 on flash messages display.

Before the PR:
before

After:
after


pierallard:
Gt

pierallard:
M

damien-carcel:
GTM

PR #4976, opened 2016-09-02
Remove useless test as there is already @require on spec

skeleton:
add SkipException is useless as there is already `@require Doctrine\ODM\MongoDB\MongoDBException` on spec.

Add SkipException make this build red however all specs are ok
http://core-ci.akeneo.com/view/PIM%20-%201.6/job/pim-community-dev-1-6-spec/572/console


nidup:
gtm

damien-carcel:
GTM.

PR #5007, opened 2016-09-14
Remove obsolete "behat/common-contexts"

skeleton:
I broke the PIM ( :kissing_heart: ) when I upgraded phpunit, the obsolete package "behat/common-contexts" is not compatible with our new version. I remove it and add usefull methods in `features/Context/WebApiContext`.

### DIFF
--- a/features/Context/WebApiContext.php
+++ b/features/Context/WebApiContext.php
@@ -2,10 +2,14 @@
 
 namespace Context;
 
+use Behat\Behat\Context\BehatContext;
 use Behat\Behat\Context\Step;
-use Behat\CommonContexts\WebApiContext as BehatWebApiContext;
+use Behat\Gherkin\Node\PyStringNode;
 use Behat\Gherkin\Node\TableNode;
+use Buzz\Browser;
 use Symfony\Component\Security\Core\Encoder\MessageDigestPasswordEncoder;
+
+require_once 'vendor/phpunit/phpunit/src/Framework/Assert/Functions.php';
 
 /**
  * Provides custom web API methods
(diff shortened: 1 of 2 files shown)


ClementGautier:
GTM

features/Context/WebApiContext.php
```diff
@@ -2,29 +2,54 @@
...
+
+    /** @var string */
     protected $url;
 
     /**
-     * {@inheritdoc}
+     * @param string  $baseUrl
+     * @param Browser $browser
      */
     public function __construct($baseUrl, Browser $browser = null)
```
aRn0D: Why $browser can be null ?

features/Context/WebApiContext.php
```diff
@@ -82,6 +107,97 @@ public function theResponseShouldContainTheFollowingData(TableNode $table)
...
+    /**
+     * Checks that response body contains JSON from PyString.
+     *
+     * @param PyStringNode $jsonString
+     *
+     * @Then /^(?:the )?response should contain json:$/
+     */
+    public function theResponseShouldContainJson(PyStringNode $jsonString)
+    {
+        $etalon = json_decode($this->replacePlaceHolder($jsonString->getRaw()), true);
```
aRn0D: What **etalon** means plz?

features/Context/WebApiContext.php
skeleton: I just copied the same constructor as the previous parent. Dunno why browser can be null

aRn0D: ok :)

skeleton: @aRn0D  in fact, Browser is always null https://github.com/akeneo/pim-community-dev/blob/master/features/Context/FeatureContext.php#L61
If you prefer, I can instance it directly

aRn0D: I think it could be better, no ?

features/Context/WebApiContext.php
aRn0D: I guest it copy/paste from the old context ?

aRn0D:
GTM once fixed

PR #5019, opened 2016-09-16
Add an option in Savers and Removers to define whether events are thr…

MarieMinasyan:
| Added Behats                      | N/A
| Changelog updated                 | N/A

…own from saveAll / removeAll or save / remove methods

### DIFF
--- a/src/Pim/Bundle/CatalogBundle/Doctrine/Common/Saver/GroupSaver.php
+++ b/src/Pim/Bundle/CatalogBundle/Doctrine/Common/Saver/GroupSaver.php
@@ -100,6 +100,8 @@ class GroupSaver implements SaverInterface, BulkSaverInterface
     {
         $this->validateGroup($group);
 
+        $options['unitary'] = true;
+
         $this->eventDispatcher->dispatch(StorageEvents::PRE_SAVE, new GenericEvent($group));
 
         $options = $this->optionsResolver->resolveSaveOptions($options);
@@ -119,6 +121,8 @@ class GroupSaver implements SaverInterface, BulkSaverInterface
         if (empty($groups)) {
             return;
         }
+
+        $options['unitary'] = false;
 
         $this->eventDispatcher->dispatch(StorageEvents::PRE_SAVE_ALL, new GenericEvent($groups));
 
--- a/src/Pim/Bundle/CatalogBundle/Doctrine/Common/Saver/GroupSaver.php
+++ b/src/Pim/Bundle/CatalogBundle/Doctrine/Common/Saver/GroupSaver.php
@@ -100,17 +100,15 @@ class GroupSaver implements SaverInterface, BulkSaverInterface
     {
         $this->validateGroup($group);
 
-        $options['unitary'] = true;
-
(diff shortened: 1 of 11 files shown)


BitOne:
I'm not sure about that feature. It can be very strange that not all save trigger the events and generate difficult to understand bugs.

Can you give more details about what you're trying to fix with this PR ?

If it is about performances, the fix should maybe done on the subscribers themselves.

MarieMinasyan:
The events are triggered the same way as before.
I simply added a new option, that allows to find out whether the PRE_SAVE and POST_SAVE events have been launched for a single entity update or not.

My case is: I need to call a third-party API. In case of a mass edit / an import / a rule execution, I should call a specific API with the ids of all impacted entities.

In case of a single entity update (from the UI), I should call another API.

Right now, there is no way of knowing in which case the event is launched.

BitOne:
Okay, thanks @MarieMinasyan, didn't understand. It makes sense. ;)

damien-carcel:
Rather than adding an option, wouldn't it be more efficient to add a new event?

- When dispatching `StorageEvents::PRE/POST_REMOVE` put data in a `RemoveEvent`
- When dispatching `StorageEvents::PRE/POST_REMOVE_ALL` put data in a `BulkRemoveEvent` (or `RemoveAllEvent` or whatever…)

MarieMinasyan:
@damien-carcel I'm not sure, because I don't want to change the current behaviour.
And in some cases, may be it doesn't matter whether the event was dispatched from the bulk method or not.

damien-carcel:
I didn't see your PR was on 1.6. Indeed an added option is probably better.

src/Pim/Bundle/CatalogBundle/Doctrine/Common/Saver/GroupSaver.php
```diff
@@ -100,15 +100,15 @@ public function save($group, array $options = [])
     {
         $this->validateGroup($group);
```
nidup: `$options['unitary'] = true;` here to be consistent

src/Pim/Bundle/CatalogBundle/Doctrine/Common/Saver/GroupSaver.php
```diff
@@ -120,10 +120,10 @@ public function saveAll(array $groups, array $options = [])
             return;
         }
 
-        $this->eventDispatcher->dispatch(StorageEvents::PRE_SAVE_ALL, new GenericEvent($groups));
-
         $options = $this->optionsResolver->resolveSaveAllOptions($options);
```
nidup: $options['unitary'] = false; here to be consistent

rybus:
GTM

jjanvier:
Sorry but I don't understand that 
> My case is: I need to call a third-party API. In case of a mass edit / an import / a rule execution, I should call a specific API with the ids of all impacted entities.

In that case you use the PRE_SAVE_ALL event. In the other case, you can call just PRE_SAVE. I don't understand why you need to add that option..

PR #5056, opened 2016-09-27
[TIP-638] First step

pierallard:
# What do this PR ?

It's the first PR to "BEMize" application in modules. When merged, master will be in a hybrid state, because just some elements are "BEMized". The goal is to merge PRs more often to prevent huge PRs.

- The "import profile" page is completely BEMized
- Added a step for click on visibles buttons. Needed because we change DOM of the product group tab on the product edit form to have a button per product group, instead of 1 button displayed anyhow.
- Fix unstable behat (pagination, manage filter dropdown, filters)
- Drop useless HTML (empty divs). Sorry, it renders the code harder to read because of indentation.
- Drop useless CSS (example: 'first' and 'last' classes added to main menus)
- Fix horrible steps in Oro JS like ".parent().parent().parent().parent()"
- There is a new "variables.less" files. The idea is to only use this one for BEM modules to get sizes, colors, fonts, etc to prevent the usage of 10000 colors in the application.
- Drop bootstrap responsive file because unused
- Drop a lot of CSS when it was no more used. When too much CSS was removed, we added hard rules in bemFixes.less to keep previous behavior and be able to merge.
- Finally, ends with a file just for this PR, to keep previous design "in place", but to merge quicly. It's crappy, but code is only at one place (bemFixes.less)


jjanvier:
ticket number on your commits please :)

a2xchip:
@pierallard Just curious what do you mean speaking ```bemify```? Is it about switching to yandex's bem? https://en.bem.info/

PR #5074, opened 2016-10-04
Fix multi-select attribute option search

Hocdoc:
After an update from Akeneo 1.5.8 to Akeneo 1.5.9 the multi-select attributes doesn't work as expected in the PEF anymore: When clicking on the attribute, we only get one option with the wrong translation.

This PR fixes the issue.


fabienlem:
Fixed by Akeneo's core team as explained in this issue: https://github.com/akeneo/pim-community-dev/issues/5111

Thank you again for your contribution :+1:

PR #5077, opened 2016-10-04
Pim 5984: Optimization on product import when iterating on each existing attributes

fabienlem:
Optimization requested by a customer

When importing products, the import job fetch all attributes and iterate on each of them. The problem is there is an array_merge method in this foreach loop and it's very time consuming when there are plenty of attributes in the database.

As an example, with 14000 attributes, the loop takes about 10 minutes, and only 8 second with the optimization.

| Added Specs                       | Covered by existing one
| Added Behats                      | Covered by existing ones

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,6 @@ php:
     - "5.5"
     - "5.6"
     - "7.0"
-    - "hhvm"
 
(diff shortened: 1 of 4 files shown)


jjanvier:
squash your commits please

PR #5070, opened 2016-10-03
Fixed PHP 7.1 static analysis fatal error

gplanchat:
**Description (for Contributor and Core Developer)**

Addresses an error in Travis CI on PHP 7.1 :

```
Fatal error: Declaration of Pim\Bundle\CatalogBundle\Doctrine\ORM\Repository\ChannelRepository::findBy(array $criteria, array $orderBy = Array, $limit = NULL, $offset = NULL) must be compatible with Doctrine\Common\Persistence\ObjectRepository::findBy(array $criteria, ?array $orderBy = NULL, $limit = NULL, $offset = NULL) in /home/travis/build/akeneo/pim-community-dev/src/Pim/Bundle/CatalogBundle/Doctrine/ORM/Repository/ChannelRepository.php on line 19
```

**Definition Of Done (for Core Developer only)**

| Q | A |
| --- | --- |
| Added Specs |  |
| Added Behats |  |
| Changelog updated |  |
| Review and 2 GTM |  |
| Micro Demo to the PO (Story only) |  |
| Migration script |  |
|  Tech Doc |  |

PR #5080, opened 2016-10-04
Backported PHP 7.1 compatibility from master to 1.6

gplanchat:

I cherry picked commits from #5070 for PHP 7.1 compatibility

I had to upgrade Symfony from 2.7.2 to 2.7.19 and Twig extensions from 1.2.0 to 1.4.0, the latter may create a risk of BC break

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,7 +43,4 @@ script:
 notifications:
   slack: akeneo:fDZaQeRRj1gVtXCW3f2kQAxo
 
-matrix:
-  allow_failures:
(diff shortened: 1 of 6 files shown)


nidup:
Hi @gplanchat 

Thank you for this backport but sadly, we can't change the composer dependencies in an already released version, it may have too much impacts for all customers applying the last patch. Even by bumping a patch version of a very stable dependency we already had severe issues on customer's project, that's why we're pretty conservative with this kind of change :wink:

PR #5082, opened 2016-10-05
PIM-5973: Fix association between products with new association types

ahocquard:

**Bug description :** 

- Go to the association tab in the PEF (for any product)
- Go to the association type page
- Add an association type
- Return to the association tab in the PEF (for any product)
- Try to associate a product to another one with the new association added juste before and save it : it fails

**Why this bug occurs :** 

When you open the association tab in the PEF, a javascript instance of the association type form is loaded.

This instance is bind to the mediator in order to listen when you check/uncheck checkboxes to associate a product to another one. This instance has a model, and this model contains association types.

Then, if you add a new association type, and you return to the association type PEF, you will not be able to associate the product with another one, with the new association type.

You will get a javascript error in the console.

In fact, the second time you navigate on the association tab int the PEF, a new javascript instance of the association type form is created.
This instance is bind to the mediator as well. The instance's model contains every association types + the new one created just before.

But the first instance is still bind to the mediator, and this instance's model does not contain the new association type.

So, when you check a checkbox, the mediator propagates the event at first to the old instance.
And it fails because the new association type is missing in the model of the old instance.

**Fix :** 

Unbind the old instance from the mediator when a new instance of the association type form is created.

**Behat :** 

Can't test it with Behat because this problem does not occur when the page is refreshed after the creation of the new association type (and that's the case with Behat).

| Review and 2 GTM                  | n

### DIFF
--- a/src/Pim/Bundle/EnrichBundle/Resources/public/js/product/form/associations.js
+++ b/src/Pim/Bundle/EnrichBundle/Resources/public/js/product/form/associations.js
@@ -102,10 +102,12 @@ define(
                 });
 
                 _.each(this.datagrids, function (datagrid) {
+                    mediator.clear('datagrid:selectModel:' + datagrid.name);
                     mediator.on('datagrid:selectModel:' + datagrid.name, function (model) {
                         this.selectModel(model, datagrid);
                     }.bind(this));
 
+                    mediator.clear('datagrid:unselectModel:' + datagrid.name);
                     mediator.on('datagrid:unselectModel:' + datagrid.name, function (model) {
                         this.unselectModel(model, datagrid);
                     }.bind(this));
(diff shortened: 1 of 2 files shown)


src/Pim/Bundle/EnrichBundle/Resources/public/js/product/form/associations.js
```diff
@@ -102,10 +102,12 @@ define(
                 });
 
                 _.each(this.datagrids, function (datagrid) {
+                    mediator.clear('datagrid:selectModel:' + datagrid.name);
```
SamirBoulil: is there any other method to 'refresh' instead of 'clear' ? otherwise GTM.
Btw, good description in the PR 👍

ahocquard: What do you mean by "refresh" ? You delete old instance(s) bind to the mediator with the method "clear", and then bind the new instance.

PR #5113, opened 2016-10-14
 batch EnrichBundle edit sender e-mail address

CustomBread:
The Akeneo batch from the EnrichBundle is configured to send mail pretending to be from akeneo.com mail servers. This is a bad idea from a security point of view. 

| Added Specs                       | no
| Changelog updated                 | no
| Micro Demo to the PO (Story only) | no
| Migration script                  | no
| Tech Doc                          | no

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,6 @@ php:
     - "5.5"
     - "5.6"
     - "7.0"
-    - "hhvm"
 
 # Allow to use container infrastructure
 sudo: false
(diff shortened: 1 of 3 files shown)


nidup:
Travis fails because out of memory i guess @damien-carcel fixed this for 1.5 and upper version (removed hhvm from travis i guess).

damien-carcel:
Yes, the problem arise on a PR of @fabienlem, we look at it together. The fix should be applied from 1.4 to master.
We removed HHVM which is useless, and add a configuration for travis memory (it is 1G by default, we raised it to 3G).

Issue #5119, opened 2016-10-17
Remove debug js code in 1.5 [Quick fix]


a2xchip:
**I'm reporting a Bug**

https://github.com/akeneo/pim-community-dev/blob/1.5/src/Pim/Bundle/EnrichBundle/Resources/public/js/product/form.js#L203

JS contains logging line. That writes both in dev and prod.

nidup:
The improvement will by released in upcoming v1.5.11 :tada:

a2xchip:
@nidup not so big approvement :) but anyway!

PR #5120, opened 2016-10-17
Issue #5119

a2xchip:
**Fix for issue #5119 **

What does this Pull Request do? reference the related issue?

Cleans js debug lines.

### DIFF
--- a/src/Pim/Bundle/EnrichBundle/Resources/public/js/product/form.js
+++ b/src/Pim/Bundle/EnrichBundle/Resources/public/js/product/form.js
@@ -199,8 +199,6 @@ define(
              * @return {Object}
              */
             renderExtension: function (extension) {
-                /* global console */
-                console.log(extension.parent.code, 'triggered the rendering of', extension.code);
                 this.getZone(extension.targetZone).appendChild(extension.el);
                 return extension.render();
             },


grena:
Hello @a2xchip , thanks again for this PR! :tada: 

As I deeply understand the frustration to have this log, we **only merge bug fixes on released versions**.
If someone uses this log or is used to have it, could be disturbed to see it vanish in a patch. (Even if I doubt it, but still); so sorry for that!

However, you can still remove this evil log on the master branch, we'll merge it with pleasure :+1:

a2xchip:
Ok! Not problem. But there is no any console logs in master as i know.

a2xchip:
@grena It is not just logs frustration. Those our users using FF&IE on Win10 experienced freeze during product page loading. This clearing helped. What i do not understand about master. Do you mean master as master branch? Or 1.5 I used is wrong?

grena:
@a2xchip it seems that this log has been removed in 1.6, so yes, forget my comment about master since it's not here anymore!

Quite interesting about the freeze, now I can understand you want to remove it. Do you have any before/after metrics about this log?

a2xchip:
@grena I have no exact. And i am sure i have seen it in previous 1.5 stable too but there was no any side effects. After upgrading to current 1.5 i started receiving reports about UI freezing in mentioned browsers. Testing on linux distribution could not  reproduce same freeze. After removing this log i asked users to check again their ability and they confirmed that it works. :)

nidup:
Hi @a2xchip 

Considering that this dev log has been dropped in 1.6 and regarding your tests with users and related feedback, we can bypass our standard rules "only merge bug fixes on already released versions" and merge this PR even in 1.5.

This change will be available with the upcoming v1.5 patch, thanks again for your contribution :tada: :tada:

a2xchip:
:) :)

PR #5125, opened 2016-10-17
Merge 1.5 => 1.6

grena:


features/Context/DataGridContext.php
```diff
@@ -352,6 +352,7 @@ public function iShouldNotSeeTheFilters($filters)
     public function iShowTheFilter($filterName)
     {
         $this->datagrid->showFilter($filterName);
+        $this->wait();
```
solivier: no need to be reworked ?

grena: All actions on filters are followed by a wait, this wait has been deleted, but it created a lot of random, so I put it back.

PR #5135, opened 2016-10-18
PIM-5993: Fix value display issues with simple/multi select attributes

grena:
Multiselect & Simpleselect **didn't send the locale they were searching on** to the controller.
So now they send the locale, because if you don't, the controller grabs ALL the labels translations with a LIMIT 20, it does this:

sql command adminer

Now that we filter by locale, we only have 20 attribute options.

| Q                 | A
| ----------------- | ---
| Specs             | N
| Behats            | Y
| Blue CI           |
| Changelog updated | Y
| Review and 2 GTM  |

### DIFF
--- a/src/Pim/Bundle/EnrichBundle/Resources/public/js/product/field/simple-select-field.js
+++ b/src/Pim/Bundle/EnrichBundle/Resources/public/js/product/field/simple-select-field.js
@@ -75,21 +75,30 @@ define(
                         ajax: {
                             url: choiceUrl,
                             cache: true,
-                            data: function (term) {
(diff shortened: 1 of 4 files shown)


src/Pim/Bundle/EnrichBundle/Resources/public/js/product/field/simple-select-field.js
```diff
@@ -75,22 +75,31 @@ define(
                         ajax: {
```
grena: Changes in this file are a copy/pasta from `multi-select-field.js` (just above) to be consistent on the same logic.

Issue #5129, opened 2016-10-18
[v1.6.3] Upgrade "league/flysystem-sftp" version to 1.0.10


mathewrapid:
**I'm reporting a Bug**

[akeneo-pim-system-info_2016-10-18_11-17.txt](https://github.com/akeneo/pim-community-dev/files/535590/akeneo-pim-system-info_2016-10-18_11-17.txt)

Added a new FlySystem adapter and filesystem
```
# config.yml
oneup_flysystem:
    adapters:
        magento_adapter:
            sftp:
                host: localhost
                port: ~
                username: test
                password: test
                root: ~
                timeout: ~
                privateKey: ~
                permPrivate: ~
                permPublic: ~
    filesystems:
        magento:
            adapter: magento_adapter
            mount: magento
            plugins:
                - oneup_flysystem.plugin.list_files
```

When using this filesystem for storing data an exception is thrown:
```
batch.ERROR: Encountered fatal error executing job ["exception",[{"class":"Symfony\Component\Debug\Exception\ContextErrorException","message":"Warning: unpack(): Type N: not enough input, need 4, have 1","messageParameters":[],"code":0,"trace":"#0 [internal function]: Symfony\Component\Debug\ErrorHandler->handleError(2, 'unpack(): Type ...', '/vagrant/vendor...', 2826, Array)
#1 /vagrant/vendor/phpseclib/phpseclib/phpseclib/Net/SSH2.php(2826): unpack('Nlength', '\x00')
#2 /vagrant/vendor/phpseclib/phpseclib/phpseclib/Net/SSH2.php(2771): Net_SSH2->_filter('P\x00\x00\x00\x17hostkeys-0...')
#3 /vagrant/vendor/phpseclib/phpseclib/phpseclib/Net/SSH2.php(2974): Net_SSH2->_get_binary_packet()
#4 /vagrant/vendor/phpseclib/phpseclib/phpseclib/Net/SFTP.php(441): Net_SSH2->_get_channel_packet(256)
#5 /vagrant/vendor/league/flysystem-sftp/src/SftpAdapter.php(127): Net_SFTP->login('test', 'test')
#6 /vagrant/vendor/league/flysystem-sftp/src/SftpAdapter.php(116): League\Flysystem\Sftp\SftpAdapter->login()
#7 /vagrant/vendor/league/flysystem/src/Adapter/AbstractFtpAdapter.php(572): League\Flysystem\Sftp\SftpAdapter->connect()
[...]

PR #5136, opened 2016-10-18
Remove legacy optional dependencies to flysystem #5129

nidup:
We introduced "league/flysystem-sftp" in v1.4 and we don't even use it, the namespace is not use either in CE and EE, I dropped it here and run the full CI without encountering any issue.

The real issue behing this is the following one https://github.com/akeneo/pim-community-dev/issues/5129 concerning the v1.6 where we would need to bump or remove this dep.


nidup:
Closed in favor of https://github.com/akeneo/pim-community-dev/pull/5139

Issue #5137, opened 2016-10-18
Optionally use an arbitrary precision library in the AkeneoMeasuresBundle


gplanchat:
Currently, the AkeneoMeasureBundle uses PHP's native operators. PHP at its internals uses `long int` or `double` native types for storing numbers, depending on the value. When the number is greater than the `INT_MAX` or smaller than `INT_MIN`, it is automatically casted to a `double`.

OroCommerce is using `brick/math` for high-precision taxes calculations.

Would it be reasonable to make the AkeneoMeasuresBundle use this library, either by migrating the `Akeneo\Bundle\MeasureBundle\Convert\MeasureConverter`, or by making an interface and letting live 2 implementations altogether ?

fabienlem:
Hi @gplanchat,

Thank you for this improvement suggestion. The precision issue when storing numeric values concerns   the whole PIM technical scope. We are pondering these questions but for now, we don't plan to use brick/math.

Kind regards

gplanchat:
In this case, why not create a `Akeneo\Bundle\MeasureBundle\Convert\MeasureConverterInterface` and let the integrators choose the implementation they desire ?

nidup:
Hi @gplanchat 

This is a very interesting approach, the MeasureConverter class is referenced less than 10 times in the PIM itself, so introduces an interface and rely on it should be pretty easy.

We could keep the default implem in the PIM that could be replaced in project.

Would you be interested by this contribution? (I know that you already have a lot in your plate ;) )

gplanchat:
After some thoughts, the `DECIMAL(24,12)` fields in the `pim_catalog_metric` may also have to be changed on project-specific implementations.

nidup:
Regarding decimal fields, that's why @fabienlem talked about a global topic which has large impacts regarding rounding values / storage precision, etc. Precision of the measure bundle library is a part of this big topic :)

gplanchat:
It seems to be solved in 2.0, isn't it ?

Doodoune:
I close this issue, it seems solved in 2.3.
If you still have this error, feel free to reopen it.

Regards,

PR #5145, opened 2016-10-20
Let use an arbitrary precision library in the AkeneoMeasuresBundle

gplanchat:

See #5137

By replacing the type hints of concrete class by an interface, it is possible to implement a high-precision measures conversion

| Q | A |
| --- | --- |
| Added Specs |  |
| Added Behats |  |
| Changelog updated |  |
| Review and 2 GTM |  |
| Migration script |  |
|  Tech Doc |  |

### DIFF
--- a/src/Akeneo/Bundle/MeasureBundle/Convert/MeasureConverterInterface.php
+++ b/src/Akeneo/Bundle/MeasureBundle/Convert/MeasureConverterInterface.php
@@ -1,0 +1,61 @@
+<?php
+
+namespace Akeneo\Bundle\MeasureBundle\Convert;
+
+use Akeneo\Bundle\MeasureBundle\Exception\UnknownFamilyMeasureException;
+use Akeneo\Bundle\MeasureBundle\Exception\UnknownMeasureException;
+use Akeneo\Bundle\MeasureBundle\Exception\UnknownOperatorException;
+
+/**
+ * Aims to convert measures
+ *
+ * @author    Romain Monceau <romain@akeneo.com>
+ * @copyright 2012 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/MIT MIT
+ */
+interface MeasureConverterInterface
+{
+    /**
+     * Set a family for the converter
+     * @param string $family
+     *
+     * @throws UnknownFamilyMeasureException
+     * @return MeasureConverterInterface
+     *
+     */
+    public function setFamily($family);
+
+    /**
+     * Convert a value from a base measure to a final measure
+     * @param string $baseUnit  Base unit for value
+     * @param string $finalUnit Result unit for value
+     * @param double $value     Value to convert
+     *
+     * @return double
+     */
+    public function convert($baseUnit, $finalUnit, $value);
+
+    /**
+     * Convert a value in a base unit to the standard unit
+     * @param string $baseUnit Base unit for value
+     * @param double $value    Value to convert
+     *
+     * @throws UnknownOperatorException
+     * @throws UnknownMeasureException
+     * @return double
+     *
+     */
+    public function convertBaseToStandard($baseUnit, $value);
+
+    /**
+     * Convert a value in a standard unit to a final unit
+     * @param string $finalUnit Final unit for value
+     * @param double $value     Value to convert
+     *
+     * @throws UnknownOperatorException
+     * @throws UnknownMeasureException
+     * @return double
+     *
+     */
+    public function convertStandardToResult($finalUnit, $value);
+}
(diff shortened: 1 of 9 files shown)


src/Akeneo/Bundle/MeasureBundle/Convert/MeasureConverterInterface.php
```diff
@@ -0,0 +1,61 @@
...
+     * Convert a value in a standard unit to a final unit
+     * @param string $finalUnit Final unit for value
+     * @param double $value     Value to convert
+     *
+     * @throws UnknownOperatorException
+     * @throws UnknownMeasureException
+     * @return double
+     *
+     */
+    public function convertStandardToResult($finalUnit, $value);
```
nidup: this method should not be in the interface and should be protected in the concret implem, it's never used except from the concret implementation itself (from convert method)

src/Akeneo/Bundle/MeasureBundle/Convert/MeasureConverterInterface.php
```diff
@@ -0,0 +1,61 @@
...
+{
+    /**
+     * Set a family for the converter
+     * @param string $family
+     *
+     * @throws UnknownFamilyMeasureException
+     * @return MeasureConverterInterface
+     *
+     */
+    public function setFamily($family);
```
nidup: This method should be removed, I checked in the code and each time we call convert() or convertBaseToStandard() we call the setFamily just before. I'd prefer that we pass the family as first argument of  convert() and convertBaseToStandard() to make this code more readable and solid.

src/Akeneo/Bundle/MeasureBundle/Convert/MeasureConverterInterface.php
```diff
@@ -0,0 +1,61 @@
...
+
+    /**
+     * Convert a value from a base measure to a final measure
+     * @param string $baseUnit  Base unit for value
+     * @param string $finalUnit Result unit for value
+     * @param double $value     Value to convert
+     *
+     * @return double
+     */
+    public function convert($baseUnit, $finalUnit, $value);
```
nidup: Should be `public function convert($family, $baseUnit, $finalUnit, $value);`

src/Akeneo/Bundle/MeasureBundle/Convert/MeasureConverterInterface.php
```diff
@@ -0,0 +1,61 @@
...
+     * Convert a value in a base unit to the standard unit
+     * @param string $baseUnit Base unit for value
+     * @param double $value    Value to convert
+     *
+     * @throws UnknownOperatorException
+     * @throws UnknownMeasureException
+     * @return double
+     *
+     */
+    public function convertBaseToStandard($baseUnit, $value);
```
nidup: Should be `public function convertBaseToStandard($family, $baseUnit, $value);`

src/Akeneo/Bundle/MeasureBundle/Convert/MeasureConverterInterface.php
```diff
@@ -0,0 +1,61 @@
+<?php
+
+namespace Akeneo\Bundle\MeasureBundle\Convert;
+
+use Akeneo\Bundle\MeasureBundle\Exception\UnknownFamilyMeasureException;
+use Akeneo\Bundle\MeasureBundle\Exception\UnknownMeasureException;
+use Akeneo\Bundle\MeasureBundle\Exception\UnknownOperatorException;
+
+/**
+ * Aims to convert measures
```
nidup: Could be replaced by `Convert from a measure family unit to another unit of the same family, for instance 1 KILOGRAM to 1000 GRAM for the family Weight`

src/Akeneo/Bundle/MeasureBundle/Convert/MeasureConverterInterface.php
nidup: Doulbe thinking of this API, the argument naming is not understandable and the order seems really not natural, would prefer `convert($family, $value, $fromUnit, $toUnit);`

src/Akeneo/Bundle/MeasureBundle/Convert/MeasureConverterInterface.php
nidup: Same weird naming for method and arguments, would prefer `convertToStandardUnit($family, $value, $fromUnit)`

nidup:
Thx for your PR, reviewing it I think about couple of things,

1) as we extract a new interface, we should rework a bit the naming, args and contract to make it more clear (this code is pretty old and legacy making an ugly API)
2) we could extract a Akeneo/Component/Measure from this bundle, this requires a bit more work especially from migration point of view but this could be done in a second time / second PR

Could you fix my comments that will allow us to fulfill the 1/ step? (once merged, I'll take care of documenting related BC breaks). Thx :tada:

gplanchat:
Changing the API will silently break a lot of related code in the PIM and in custom bundles, it is way out of the initial scope of this PR :)

aRn0D:
Hi, Grégory @gplanchat thank you for your contribution! I will close this PR because what we want, require more than the scope of that PR.

Issue #3783, opened 2015-12-18
Classes implementing ArchivableWriterInterface consumes too much resources


gplanchat:
I created a connector that exports to CSV using `Pim\Bundle\BaseConnectorBundle\Writer\File\CsvWriter` and its descendents for variant group and products.

When I'm exporting my products with images, a zip file seems to be generated, due to the implementation of interface `Pim\Bundle\BaseConnectorBundle\Writer\File\ArchivableWriterInterface`. This file generation seems to generate the file in-memory in the first place. I currently have an export of ~2200 products with an average of 3 HD images for each product.

The images folder weights 13Gb, the server does not have this amount of RAM and nevertheless it should not consume that much. If I remove the interface from the `CsvWriter` class, everything works fine (I do not need this zip file).

By quickly reading the sources, I can see that the zip file generation uses `thephpleague/flysystem`, with a lot of calls to `file_get_contents`. Maybe using streams interface of Flysystem instead of string interface make the memory consumption much lower.

Gregory

gplanchat:
Seems like there is the same issue here :

http://www.akeneo.com/forums/topic/issues-when-running-job-via-terminal/#post-6519

nidup:
This is a known issue, we're discussing with product owner to decide what is the best approach to better handled this case (we're preparing a set of features to enhance the export part).

It could be an extra step which zip (or copy files to a configured server), from design point of view, it should not come from the implementation of an interface (at least not natively).

gplanchat:
Maybe an optional `Job` item named `archive`, in addition to `reader`, `processor` and `writer` could do the trick, be very explicit and be easy to implement

willy-ahva:
We are currently working on refactoring the Archiver on the master branch. It'll fix your problem. More information later ;)
Thank you for the feedback !

willy-ahva:
@gplanchat can you give me the version you're working on please ? You can check this documentation to disable the archiver or create your own one : http://docs.akeneo.com/1.4/reference/scalability_guide/more_than_1GB_of_product_media_to_export.html Available since v1.4 :)
Thank you for your feedback !

gplanchat:
Hello @willy-ahva,

Did the 1.6 changes in the `AkeneoBatchBundle` fixed this issue ?

Grégory

gplanchat:
This seems to be a bit outdated now

PR #5146, opened 2016-10-20
Archiving big files causes huge memory usage

masmrlar:
| Q | A |
| --- | --- |
| Added Specs | N |
| Added Behats | N |
| Changelog updated | N |
| Review and 2 GTM | N |
| Migration script | N |
|  Tech Doc | N |

The ArchivableFileWriterArchiver is using streams to write to the zip
archive, but it reads the content of the source file using
file_get_contents.

This behaviour is changed to also use a stream to read the source file
and use putStream of flysystem to add content to the target archive.

### DIFF
--- a/src/Pim/Component/Connector/Archiver/ArchivableFileWriterArchiver.php
+++ b/src/Pim/Component/Connector/Archiver/ArchivableFileWriterArchiver.php
@@ -57,7 +57,9 @@ class ArchivableFileWriterArchiver extends AbstractFilesystemArchiver
                 );
 
                 foreach ($writer->getWrittenFiles() as $fullPath => $localPath) {
-                    $filesystem->put($localPath, file_get_contents($fullPath));
+                    $stream = fopen($fullPath, 'r');
+                    $filesystem->putStream($localPath, $stream);
+                    fclose($stream);
                 }
             }
         }


gplanchat:
It seems to be related to #3783

src/Pim/Component/Connector/Archiver/ArchivableFileWriterArchiver.php
```diff
@@ -57,7 +57,9 @@ public function archive(JobExecution $jobExecution)
                 );
 
                 foreach ($writer->getWrittenFiles() as $fullPath => $localPath) {
-                    $filesystem->put($localPath, file_get_contents($fullPath));
+                    $stream = fopen($fullPath, 'r');
```
gplanchat: Shouldn't we use Flysystem's API from the start to the end ?

phaseinducer: It's still used on the line below, @masmrlar just replaced the php method used to feed Flysystem.

src/Pim/Component/Connector/Archiver/ArchivableFileWriterArchiver.php
```diff
@@ -57,7 +57,9 @@ public function archive(JobExecution $jobExecution)
...
+                    $filesystem->putStream($localPath, $stream);
+                    fclose($stream);
```
ahocquard: From flysystem api documentation : "Some SDK's close streams after consuming them, therefore, before calling fclose on the resource, check if it's still valid using is_resource."

Please use:
```
 if (is_resource($stream)) {
    fclose($stream);
}
```

ClementGautier:
Closing as the source fork have been deleted. Please apply the patch on a new branch and be pleased to recreate a PR

PR #5132, opened 2016-10-18
Added a new measure family for food products and food supplements. Extended Duration family.

JulienDotDev:
**Description (for Contributor and Core Developer)**

This new measure unit is used to store legal informations about food products and food supplements.
This extend the duration family to add three units : week, month (conversion value could be discussed), year.

**Definition Of Done (for Core Developer only)**

| Q                                 | A
| --------------------------------- | ---
| Added Specs                       |
| Added Behats                      |
| Changelog updated                 |
| Review and 2 GTM                  |
| Micro Demo to the PO (Story only) |
| Migration script                  |
| Tech Doc                          |

PR #5148, opened 2016-10-20
Standard unit for energy

jmleroux:
The standard unit for energy is the Joule, not the kilo-calorie

### DIFF
--- a/src/Akeneo/Bundle/MeasureBundle/Resources/config/measure.yml
+++ b/src/Akeneo/Bundle/MeasureBundle/Resources/config/measure.yml
@@ -453,13 +453,19 @@ measures_config:
                 convert: [{'mul': 750.06375541921}]
                 symbol: mmHg
     Energy:
-        standard: KILOCALORIE
+        standard: JOULE
         units:
-            KILOCALORIE:
+            JOULE:
                 convert: [{'mul': 1}]
+                symbol: J
+            CALORIE:
+                convert: [{'mul': 4.184}]
+                symbol: cal
+            KILOCALORIE:
+                convert: [{'mul': 4184}]
                 symbol: kcal
             KILOJOULE:
-                convert: [{'mul': 4.1868}]
+                convert: [{'mul': 1000}]
                 symbol: kJ
     CaseBox:
         standard: PIECE


nidup:
Cf https://github.com/akeneo/pim-community-dev/pull/5132

Cf https://en.wikipedia.org/wiki/Calorie 

``̀`
Although these units are part of the metric system, they have been superseded in the International System of Units by the joule. One small calorie is approximately 4.2 joules (so one large calorie is about 4.2 kilojoules). The factor used to convert calories to joules at a given temperature is numerically equivalent to the specific heat capacity of water expressed in joules per kelvin per gram or per kilogram. The precise conversion factor depends on the definition adopted.

In spite of its non-official status, the large calorie is still widely used as a unit of food energy. The small calorie is also often used for measurements in chemistry, although the amounts involved are typically recorded in kilocalories.
``̀`

When speaking about food properties, every manufacturer use "calorie" and not "joule" even if not in ISU, there are several metrics standards and the origin of the measure bundle was zend measure which not follow ISU..

As discussed with @fitn I'd really love to discuss about defining the standard(s) we follow, could be ISU, and discuss about what we do with other common measures and units.

We need a clear rule, we already had the case with "galon" which defines different quantity depending on standard and country and that several contributors would change in different ways :)

jmleroux:
@nidup , sorry but i didn't understand which measure you agree to keep as the standard. :no_mouth:

nidup:
@jmleroux we'll have a meeting with Milie and @fitn next week to define which standard(s) we want to follow. For now, we don't follow any standard, that's the main issue, so someone can propose a change from a standard and another person change the same unit referring to another standard. We need clear rules regarding this to be able to say we follow the standard X or the standard Y and basta :smile:

jmleroux:
Energy in Wikipedia: https://en.wikipedia.org/wiki/Units_of_energy

> Because energy is defined via work, the SI unit for energy is the same as the unit of work – the joule (J), named in honor of James Prescott Joule and his experiments on the mechanical equivalent of heat.

As discussed with PO, we still use the SI as the standard.
So i think this PR is valid.

jjanvier:
actually there is something I don't understand

old code: from kcal to kJ : x4.1868
new code: from J to kcal: x4184

src/Akeneo/Bundle/MeasureBundle/Resources/config/measure.yml
```diff
@@ -453,13 +453,19 @@ measures_config:
...
     Energy:
-        standard: KILOCALORIE
+        standard: JOULE
         units:
-            KILOCALORIE:
+            JOULE:
                 convert: [{'mul': 1}]
+                symbol: J
+            CALORIE:
+                convert: [{'mul': 4.184}]
```
jjanvier: divide no?

src/Akeneo/Bundle/MeasureBundle/Resources/config/measure.yml
```diff
@@ -453,13 +453,19 @@ measures_config:
...
+                symbol: cal
+            KILOCALORIE:
+                convert: [{'mul': 4184}]
```
jjanvier: divide no?

src/Akeneo/Bundle/MeasureBundle/Resources/config/measure.yml
jmleroux: as discussed IRL, no ;)

src/Akeneo/Bundle/MeasureBundle/Resources/config/measure.yml
jmleroux: same

PR #5159, opened 2016-10-24
Media migration memory consumption

masmrlar:
With big amount of media files managed by the PIM, the media migration from version 1.3 to 1.6 is slow (and slows down during the migration process) and its quite memory consuming.

The issue is in upgrades/1.3-1.4/common/MediaMigration.php::storeLocalMedias

The foreach loop does not detach the $fileInfo entities from the entity manager.
Detaching the entities and regular flushing/clearing the entity manager helps the garbage collector to keep the memory usage low and speeds up the migration process as the entity manager does not handle the detached entities (as long as they are not reattached anymore).

| Q | A |
| --- | --- |
| Added Specs | N |
| Added Behats | N |
| Changelog updated | N |
| Review and 2 GTM | N |
| Migration script | N |
|  Tech Doc | N |

### DIFF
--- a/upgrades/1.3-1.4/common/MediaMigration.php
+++ b/upgrades/1.3-1.4/common/MediaMigration.php
@@ -93,14 +93,8 @@ class MediaMigration
      */
     public function storeLocalMedias()
     {
-        // enable the garbage collector
-        gc_enable();
-
         /** @var \Doctrine\ORM\EntityManagerInterface $em */
-        $em = null;
-        if (!$this->upgradeHelper->areProductsStoredInMongo()) {
-            $em = $this->container->get('doctrine.orm.entity_manager');
-        };
+        $em = $this->container->get('doctrine.orm.entity_manager');
 
         $this->output->writeln(sprintf('Storing medias located in <comment>%s</comment> to the catalog filesystem...', $this->mediaDirectory));
 
@@ -116,30 +110,17 @@ class MediaMigration
                 ['old_file_key' => $file->getFilename()],
                 ['id'           => $fileInfo->getId()]
             );
-            // detach and unset if products are not stored in Mongo
-            if (null !== $em) {
-                $em->detach($fileInfo);
-            }
+            $em->detach($fileInfo);
             unset($fileInfo);
             unset($file);
 
             if (0 == $count % $batch) {
-                // flush / clear / gc
-                if (null !== $em) {
-                    $em->flush();
-                    $em->clear();
-                    gc_collect_cycles();
-                }
+                $em->clear();
             }
             $count++;
         }
 
-        // flush / clear / gc finally
-        if (null !== $em) {
-            $em->flush();
-            $em->clear();
-            gc_collect_cycles();
-        }
+        $em->clear();
     }
 
     /**
--- a/upgrades/1.3-1.4/common/MediaMigration.php
+++ b/upgrades/1.3-1.4/common/MediaMigration.php
@@ -99,7 +99,7 @@ class MediaMigration
         /** @var \Doctrine\ORM\EntityManagerInterface $em */
         $em = null;
         if (!$this->upgradeHelper->areProductsStoredInMongo()) {
-            $this->container->get('doctrine.orm.entity_manager');
+            $em = $this->container->get('doctrine.orm.entity_manager');
         };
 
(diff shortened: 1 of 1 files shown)


fabienlem:
Hello @masmrlar,

Thank you a lot for your pull request :+1:

I copied and pasted the issue's description from the service desk support ticket in order to help our team to review the PR. We will do it soon!

Kind regards,
Fabien

upgrades/1.3-1.4/common/MediaMigration.php
```diff
@@ -93,18 +93,52 @@ public function createFileInfoTable()
      */
     public function storeLocalMedias()
     {
+        // enable the garbage collector
+        gc_enable();
```
jjanvier: it is useful? we didn't disable it, so it's already enabled

upgrades/1.3-1.4/common/MediaMigration.php
```diff
@@ -93,18 +93,52 @@ public function createFileInfoTable()
...
+
+        /** @var \Doctrine\ORM\EntityManagerInterface $em */
+        $em = null;
+        if (!$this->upgradeHelper->areProductsStoredInMongo()) {
```
jjanvier: fileInfo objects are always stored in MySQL
so I think you don't need this if

upgrades/1.3-1.4/common/MediaMigration.php
```diff
@@ -93,18 +93,52 @@ public function createFileInfoTable()
...
+        $count = 0;
+        $batch = 1000;
         foreach ($finder->files()->followLinks()->in($this->mediaDirectory) as $file) {
             $fileInfo = $storer->store($file, FileStorage::CATALOG_STORAGE_ALIAS);
             $this->ormConnection->update(
                 'akeneo_file_storage_file_info',
                 ['old_file_key' => $file->getFilename()],
                 ['id'           => $fileInfo->getId()]
             );
+            // detach and unset if products are not stored in Mongo
```
jjanvier: as said before, as fileInfo objects are always in MySQL, I think you can always detach them :)

upgrades/1.3-1.4/common/MediaMigration.php
```diff
@@ -93,18 +93,52 @@ public function createFileInfoTable()
...
+            if (null !== $em) {
+                $em->detach($fileInfo);
+            }
+            unset($fileInfo);
+            unset($file);
+
+            if (0 == $count % $batch) {
+                // flush / clear / gc
+                if (null !== $em) {
+                    $em->flush();
```
jjanvier: just to understand, why do you flush?
the file storer already flushes afaik

upgrades/1.3-1.4/common/MediaMigration.php
```diff
@@ -93,18 +93,52 @@ public function createFileInfoTable()
...
+                    $em->clear();
+                    gc_collect_cycles();
+                }
+            }
+            $count++;
+        }
+
+        // flush / clear / gc finally
+        if (null !== $em) {
+            $em->flush();
```
jjanvier: same comment that before

upgrades/1.3-1.4/common/MediaMigration.php
```diff
@@ -93,18 +93,52 @@ public function createFileInfoTable()
...
+            }
+            unset($fileInfo);
+            unset($file);
+
+            if (0 == $count % $batch) {
+                // flush / clear / gc
+                if (null !== $em) {
+                    $em->flush();
+                    $em->clear();
+                    gc_collect_cycles();
```
jjanvier: not sure you need to handle the gc manually
cleaning Doctrine entity manager is already a huge step :+1:

BitOne: Yeap, the GC will attempt to collect cycles by itself, so no need to call it.

upgrades/1.3-1.4/common/MediaMigration.php
masmrlar: ok, going to rework this...
Thanks for the hints

pierallard:
*info* This one is already fixed by @jmleroux in pim-community-standard https://github.com/akeneo/pim-community-standard/commit/c14e28619a578406df5f374f57c007dd063d9035

PR #5160, opened 2016-10-24
Detach read items on export job

masmrlar:
| Q | A |
| --- | --- |
| Added Specs | N |
| Added Behats | N |
| Changelog updated | N |
| Review and 2 GTM | N |
| Migration script | N |
|  Tech Doc | N |

The memory consumption of export jobs constantly increases if products
with associations are exported.

The fix detaches the read items together with their associations and
products collected in the associations.

### DIFF
--- a/src/Akeneo/Component/Batch/Step/ItemStep.php
+++ b/src/Akeneo/Component/Batch/Step/ItemStep.php
@@ -118,6 +118,7 @@ class ItemStep extends AbstractStep
             }
 
             $processedItem = $this->process($readItem);
+            unset($readItem);
             if (null !== $processedItem) {
                 $itemsToWrite[] = $processedItem;
                 $writeCount++;
--- a/src/Pim/Bundle/EnrichBundle/Connector/Processor/QuickExport/ProductProcessor.php
+++ b/src/Pim/Bundle/EnrichBundle/Connector/Processor/QuickExport/ProductProcessor.php
@@ -9,6 +9,7 @@ use Akeneo\Component\Batch\Model\StepExecution;
 use Akeneo\Component\StorageUtils\Detacher\ObjectDetacherInterface;
 use Pim\Bundle\EnrichBundle\Connector\Processor\AbstractProcessor;
 use Pim\Component\Catalog\Builder\ProductBuilderInterface;
+use Pim\Component\Catalog\Model\AssociationInterface;
 use Pim\Component\Catalog\Model\ProductInterface;
 use Pim\Component\Catalog\Repository\AttributeRepositoryInterface;
 use Pim\Component\Catalog\Repository\ChannelRepositoryInterface;
@@ -107,7 +108,7 @@ class ProductProcessor extends AbstractProcessor
             $this->fetchMedia($product, $directory);
         }
 
-        $this->detacher->detach($product);
+        $this->detachProduct($product);
 
         return $productStandard;
     }
@@ -218,5 +219,25 @@ class ProductProcessor extends AbstractProcessor
 
         $token = new UsernamePasswordToken($user, null, 'main', $user->getRoles());
         $this->tokenStorage->setToken($token);
+    }
+
+
+    /**
+     * Detach $product and its associations
+     *
+     * @param ProductInterface $product
+     *
+     * @return void
+     */
+    protected function detachProduct(ProductInterface $product)
+    {
+        $this->detacher->detach($product);
+        /** @var AssociationInterface $association */
+        $associations = $product->getAssociations();
+        foreach ($associations as $association) {
+            foreach ($association->getProducts() as $associatedProduct) {
+                $this->detacher->detach($associatedProduct);
+            }
+        }
     }
 }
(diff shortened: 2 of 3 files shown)


nidup:
Hi @masmrlar 

Thanks a lot for your pull request :tada: 

However, this change can't be merged here, the ItemStep is used in lot of contexts, it can be used with items that are objects or arrays not managed at all by the Doctrine Unit Of Work, I'd prefer a more precise fix, fixing the very specific process of product exports.

BTW, in case of product export, the products are detached here https://github.com/akeneo/pim-community-dev/blob/master/src/Pim/Component/Connector/Processor/Normalization/ProductProcessor.php#L112

When using ORM storage, the associations should be detached in cascade thanks to https://github.com/akeneo/pim-community-dev/blob/master/src/Pim/Bundle/CatalogBundle/Resources/config/model/doctrine/Product.orm.yml#L55

However you maybe found a not covered memory leak (it seems very strange as these part are well benched by our perf testing but maybe it's related to a large amount of associations for a product, or a lot of products associated).

Are you running the export in prod mode? (because when running in dev mode, the symfony profiler keeps references to all objects and cause massive memory leak).

Does it constantly consumes more memory for each batch of products and never stabilize? (usually it may consumes more an more memory on first bulks at the beginning because loading all related entities as family, attributes, options, families, etc in memory, then stabilize after few bulks).

Thank you

masmrlar:
Hi @nidup ,

you're right. This fix has really some code smells.
I'll recheck the detaching in the normalization step. We are using ORM storage and it seems that the cascading does not fully work on the chain product -> associations -> products - thats why I forcefully detach the associated items.

We are running the export in prod mode and the memory consumption increases with each product containing associated products. The products have about 5 to 10 associations.

masmrlar:
... going to rewrite this to have the foced detach of associations in the ProductProcessor...

pierallard:
Hi @masmrlar 

Thanks for this contribution!
I just checked we had a memory leak on export, and effectively the memory grow up when there are a lot of associations for each products. For example, when I set 100.000 associations for 1.000 products, I see memory leaking from 60 to 150 Mb during the export.
Next, I tried to fix the bug with your PR, and it does not stop memory leak (it's worst, it grow up to 400 Mb after 100 products).

I think your use case is different than mine because memory leak comes from un-detached associated products, and all the products of my file are attached.

Here is my test file. Simply import  (rename it to CSV) it in a standard install (with icecat catalog), then run export :)
[product_with_100_assoc.txt](https://github.com/akeneo/pim-community-dev/files/738656/product_with_100_assoc.txt)

Pierre

src/Akeneo/Component/Batch/Step/ItemStep.php
```diff
@@ -118,6 +118,7 @@ public function doExecute(StepExecution $stepExecution)
             }
 
             $processedItem = $this->process($readItem);
+            unset($readItem);
```
aRn0D: Not sure about that! This variable will be destroyed at the end of the method, so we should remove that.

src/Pim/Component/Connector/Processor/Normalization/ProductProcessor.php
```diff
@@ -188,4 +189,23 @@ protected function areAttributesToFilter(JobParameters $parameters)
...
+     *
+     * @return void
+     */
+    protected function detachProduct(ProductInterface $product)
+    {
+        $this->detacher->detach($product);
+        /** @var AssociationInterface $association */
+        $associations = $product->getAssociations();
+        foreach ($associations as $association) {
+            foreach ($association->getProducts() as $associatedProduct) {
```
aRn0D: You could use:

```php
$this->detacher->detachAll($association->getProducts()->toArray());
```

src/Pim/Bundle/EnrichBundle/Connector/Processor/QuickExport/ProductProcessor.php
```diff
@@ -219,4 +220,24 @@ protected function initSecurityContext(StepExecution $stepExecution)
...
+     *
+     * @return void
+     */
+    protected function detachProduct(ProductInterface $product)
+    {
+        $this->detacher->detach($product);
+        /** @var AssociationInterface $association */
+        $associations = $product->getAssociations();
+        foreach ($associations as $association) {
+            foreach ($association->getProducts() as $associatedProduct) {
```
aRn0D: You could use:

```php
$this->detacher->detachAll($association->getProducts()->toArray());
```

src/Pim/Bundle/EnrichBundle/Connector/Processor/QuickExport/ProductProcessor.php
```diff
@@ -219,4 +220,24 @@ protected function initSecurityContext(StepExecution $stepExecution)
...
         $this->tokenStorage->setToken($token);
     }
+
+
+    /**
+     * Detach $product and its associations
+     *
+     * @param ProductInterface $product
+     *
+     * @return void
```
aRn0D: We follow the Symfony code style convention, we don't add the return annotation for "void", we should remove it :)

pierallard:
Hi @masmrlar !
I close this PR because we have no return since 10 months. By the way, it was fixed recently on this PR : https://github.com/akeneo/pim-community-dev/pull/7414

PR #5164, opened 2016-10-25
TIP-477: call key "identifier" from the product standard format

skeleton:
use the key "identifier" in the product standard format introduced in a previous PR


SamirBoulil:
GTM

PR #5190, opened 2016-10-31
[SLA] PIM-5994: fix duplication of attributes in the product export

jjanvier:
What's here?
1. fix the way we test our export files in Behat. Yes it was wrong. We checked missing headers, but not if we had too many headers.
2. fix a few false positive Behat due to 1
3. remove a weird false positive Behat that allowed to export all labels of attributes (even the empty labels). That's not the behavior we have for the other entities. And we want to fix that, we should do it for all entities, not only for the attributes.
4. fix the duplication of attributes in the product export 


jjanvier:
missing changelog

fabienlem:
CI ok !

PR #5191, opened 2016-10-31
Exclude specs from Scrutinizer inspection

jjanvier:


.scrutinizer.yml
```diff
@@ -2,6 +2,7 @@ filter:
     paths:
         - src/*
     excluded_paths:
+        - src/*/*/*/spec/*
```
skeleton: maybe you can add `src/*/*/*/tests/*` to exclude unitary & integration tests

jjanvier: done

PR #5265, opened 2016-11-25
[Merge] 1.6 -> master

pierallard:


features/export/export_attributes.feature
```diff
@@ -43,42 +44,3 @@ Feature: Export attributes
     pim_catalog_text;123;"Attribute 123";other;;0;1;;;;;;255;;;;;;;;;;;0;0;0;0;0
     """
 
-  Scenario: Successfully export all label locales even if no value were set in csv
```
damien-carcel: Why is this test completely removed?

features/reference-data/export/export_attributes.feature
```diff
@@ -14,31 +15,31 @@ Feature: Export attributes
     And I wait for the "csv_footwear_attribute_export" job to finish
     Then exported file of "csv_footwear_attribute_export" should contain:
     """
-    type;code;label-en_US;label-fr_FR;group;unique;useable_as_grid_filter;allowed_extensions;metric_family;default_metric_unit;reference_data_name;available_locales;max_characters;validation_rule;validation_regexp;wysiwyg_enabled;number_min;number_max;decimals_allowed;negative_allowed;date_min;date_max;max_file_size;minimum_input_length;localizable;scopable;sort_order;is_read_only
```
damien-carcel: Same question here.

features/export/export_attributes.feature
jjanvier: attribute is the only entity of the PIM where empty labels where removed
we decided to remove this particularity here (the bug fix enligted this problem)
if we want to keep this behavior, we should do it globally, for all entities

PR #5292, opened 2016-12-02
[TIP-646] Remove input selectors

pierallard:
TODO

- [x] Export builder
- [x] Create/update role
- [x] Attribute options
- [x] new attribute
- [x] cross of panel not like cross of compare
- [x] options of the attributes


nidup:
Bem bem bang bang ;)

PR #5386, opened 2016-12-20
AM: fix style and clean useless stuff

grena:


grena:
Renamed & re-created ( #5392 )

PR #5397, opened 2016-12-23
PIM-6042: Fix hard cleaning of associations

anaelChardan:

| Review and 2 GTM                  | 🕐

Build : https://core-ci.akeneo.com/job/dci_behat/7761

:red_circle: Done but fail

:clock1: Done but pending

### DIFF
--- a/CHANGELOG-1.6.md
+++ b/CHANGELOG-1.6.md
@@ -3,6 +3,7 @@
 ## Bug fixes
 
 - PIM-6033: Fix validation issue when you add a blank attribute option line
+- PIM-6042: Successfully import product associations without removing already existing associations when option "compare values" is set to true
 
 # 1.6.7 (2016-12-20)
 
--- a/features/Context/FixturesContext.php
+++ b/features/Context/FixturesContext.php
@@ -1109,6 +1109,31 @@ class FixturesContext extends BaseFixturesContext
     }
 
     /**
+     * @param string    $identifier
+     * @param TableNode $table
+     *
+     * @Then /^the product "([^"]*)" should have the following associations?:$/
+     */
+    public function theProductShouldHaveTheFollowingAssociations($identifier, TableNode $table)
+    {
+        $this->getMainContext()->getSubcontext('hook')->clearUOW();
+        $filter = $this->getContainer()->get('pim_catalog.comparator.filter.product_association');
+
+        $values['associations'] = [];
+        foreach ($table->getHash() as $row) {
+            if (isset($row['products'])) {
+                $values['associations'][$row['type']]['products'] = explode(',', $row['products']);
+            }
+
+            if (isset($row['groups'])) {
+                $values['associations'][$row['type']]['groups'] = explode(',', $row['groups']);
+            }
+        }
+
+        assertEquals([], $filter->filter($this->getProduct($identifier), $values));
+    }
+
+    /**
      * @param string $productCode
      * @param string $familyCode
      *
--- a/features/import/product/import_products_with_associations.feature
+++ b/features/import/product/import_products_with_associations.feature
@@ -155,3 +155,30 @@ Feature: Execute a job
     And I launch the import job
     And I wait for the "csv_footwear_product_import" job to finish
     Then I should see the text "skipped product (no associations detected)"
+
+  @jira https://akeneo.atlassian.net/browse/PIM-6042
+  Scenario: Successfully import product associations without removing already existing associations when option "compare values" is set to true
+    Given the following product:
+      | sku     | name-en_US |
+      | SKU-001 | sku-001    |
+      | SKU-002 | sku-002    |
+      | SKU-003 | sku-003    |
+    And the following associations for the product "SKU-001":
+      | type   | product     |
+      | X_SELL | SKU-002     |
+      | UPSELL | SKU-002     |
+    And the following CSV file to import:
+      """
+      sku;UPSELL-products
+      SKU-001;SKU-003
+      """
+    And the following job "csv_footwear_product_import" configuration:
+      | filePath          | %file to import% |
+      | enabledComparison | yes              |
+    When I am on the "csv_footwear_product_import" import job page
+    And I launch the import job
+    And I wait for the "csv_footwear_product_import" job to finish
+    Then the product "SKU-001" should have the following associations:
+      | type   | products |
+      | X_SELL | SKU-002  |
+      | UPSELL | SKU-003  |
--- a/src/Pim/Component/Catalog/Model/AssociationInterface.php
+++ b/src/Pim/Component/Catalog/Model/AssociationInterface.php
@@ -100,7 +100,7 @@ interface AssociationInterface extends ReferableInterface
     /**
      * Get groups
      *
-     * @return GroupInterface[]
+     * @return GroupInterface[]|ArrayCollection
      */
     public function getGroups();
 
--- a/src/Pim/Component/Catalog/Model/ProductInterface.php
+++ b/src/Pim/Component/Catalog/Model/ProductInterface.php
@@ -135,7 +135,7 @@ interface ProductInterface extends
     /**
      * Get types of associations
      *
-     * @return AssociationInterface[]
+     * @return AssociationInterface[]|ArrayCollection
      */
     public function getAssociations();
 
--- a/src/Pim/Component/Catalog/Updater/Setter/AssociationFieldSetter.php
+++ b/src/Pim/Component/Catalog/Updater/Setter/AssociationFieldSetter.php
@@ -62,26 +62,50 @@ class AssociationFieldSetter extends AbstractFieldSetter
     public function setFieldData(ProductInterface $product, $field, $data, array $options = [])
     {
         $this->checkData($field, $data);
-        $this->clearAssociations($product);
+        $this->clearAssociations($product, $data);
         $this->addMissingAssociations($product);
         $this->setProductsAndGroupsToAssociations($product, $data);
     }
 
     /**
-     * Clear associations (remove groups and products from existing associations)
+     * Clear only concerned associations (remove groups and products from existing associations)
      *
      * @param ProductInterface $product
+     * @param array            $data
+     *
+     * Expected data input format:
+     * {
+     *     "XSELL": {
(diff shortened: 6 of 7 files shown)


CHANGELOG-1.6.md
```diff
@@ -3,6 +3,7 @@
 ## Bug fixes
 
 - PIM-6033: Fix validation issue when you add a blank attribute option line
+- PIM-6042: Successfully import product associations without remove already existing association when option "compare values" is set to true
```
SamirBoulil: without removing

features/Context/FixturesContext.php
```diff
@@ -1109,6 +1109,31 @@ public function theProductShouldHaveTheFollowingValues($identifier, TableNode $t
     }
 
     /**
+     * @param string    $identifier
+     * @param TableNode $table
+     *
+     * @Given /^the product "([^"]*)" should have the following associations?:$/
```
SamirBoulil: the product "XXX" has the following

features/Context/FixturesContext.php
```diff
@@ -1109,6 +1109,31 @@ public function theProductShouldHaveTheFollowingValues($identifier, TableNode $t
...
+     */
+    public function theProductShouldHaveTheFollowingAssociations($identifier, TableNode $table)
+    {
+        $this->getMainContext()->getSubcontext('hook')->clearUOW();
```
SamirBoulil: can you elaborate on that ?

features/Context/FixturesContext.php
anaelChardan: As it is an assertion I followed what we do on https://github.com/akeneo/pim-community-dev/blob/master/features/Context/FixturesContext.php#L1076 to be consistent :)

anaelChardan: What do you prefer ?

anaelChardan: But the instruction should be @Then

src/Pim/Component/Catalog/Updater/Setter/AssociationFieldSetter.php
```diff
@@ -62,26 +62,50 @@ public function __construct(
...
+     *         "products": ["AKN_TS1", "AKN_TSH2"]
+     *     },
+     *     "UPSELL": {
+     *         "groups": ["group3", "group4"],
+     *         "products": ["AKN_TS3", "AKN_TSH4"]
+     *     },
+     * }
      */
-    protected function clearAssociations(ProductInterface $product)
+    protected function clearAssociations(ProductInterface $product, array $data = null)
```
SamirBoulil: So in order to remove all associations of a product, you need to pass in all its associations ?

features/Context/FixturesContext.php
anaelChardan: Yup, If I don't do that my product is not updated on my behat context so I followed the same way to refresh the Unit Of Work as https://github.com/akeneo/pim-community-dev/blob/master/features/Context/FixturesContext.php#L1081

src/Pim/Component/Catalog/Updater/Setter/AssociationFieldSetter.php
anaelChardan: Yes, because in the case of a set you don't want to set the associations that you don't pass to the method. You clear only the associations you set, it works only on the scope of the setter

CHANGELOG-1.6.md
anaelChardan: +1 :)

src/Pim/Component/Catalog/Model/AssociationInterface.php
```diff
@@ -100,7 +100,7 @@ public function addGroup(GroupInterface $group);
     /**
      * Get groups
      *
-     * @return GroupInterface[]
+     * @return GroupInterface[]|ArrayCollection
```
a2xchip: There are two variants of documenting return value - type (ArrayCollection) or array of types (GroupInterface[]), both are true, at least was true before php7 version. IMHO we should use return type version to not mislead devs. It is also possible to use interface in docs as ArrayCollection implements Collection interface.

src/Pim/Component/Catalog/Model/ProductInterface.php
```diff
@@ -135,7 +135,7 @@ public function getGroupCodes();
     /**
      * Get types of associations
      *
-     * @return AssociationInterface[]
+     * @return AssociationInterface[]|ArrayCollection
```
a2xchip: Same here

src/Pim/Component/Catalog/Model/AssociationInterface.php
anaelChardan: So in my opinion it should be GroupInterface[]|Collection because GroupInterface does not provide all the methods availables on Collection but in Collection we don't have the type of collection it is ^^ :)

src/Pim/Component/Catalog/Model/ProductInterface.php
anaelChardan: Same too :D

src/Pim/Component/Catalog/Model/AssociationInterface.php
a2xchip: Ok! I assume such ```|``` means ```or``` in docblock. Dev reading docs understand he/she could typehint to ```array``` because ```GroupInterface[]``` means array of group interfaces. What will intrepreter say? imo it is return value duplication, as getter will return Collection|null if there are no groups.

a2xchip: Just checked this https://wiki.php.net/rfc/return_types and this
https://www.phpdoc.org/docs/latest/guides/types.html#keywords

Looks like such duplication could help to IDE's autocompletion as it would presume that return value has properties and methods of all documented through ```|``` types

anaelChardan: Yes, actually it helped me to write my function documenting by ArrayCollection ;), so what do you advise ? :)

a2xchip: @SamirBoulil @juliensnz @jjanvier @pierallard

anaelChardan: Yes and on the doc https://www.phpdoc.org/docs/latest/guides/types.html#multiple-types-combined they have an example for ArrayObjects :)

a2xchip: @anaelChardan It depends on how long these code could be used and who use it more: IDE's or devs. For example as a dev i do prefer to see single right return type with possible variations in face of null value. But for IDE some additional types could help to autocomplete type of value inside collection. 

P.S. ArrayCollection also implements Selectable interface

anaelChardan: Yes, I totally agree with you :) yes so it's a little bit harder ^^ with multiple implements directive :D

a2xchip: @anaelChardan Yeah, and also we have such PSR draft :-)  https://github.com/phpDocumentor/fig-standards/blob/master/proposed/phpdoc.md

features/Context/FixturesContext.php
SamirBoulil: Ok should have.

features/Context/FixturesContext.php
SamirBoulil: Ok.

src/Pim/Component/Catalog/Model/AssociationInterface.php
anaelChardan: ^^

a2xchip: :joy: At least we did little research on docblock topic @anaelChardan

anaelChardan: Yes, you're right, thanks @a2xchip  😂

CHANGELOG-1.6.md
ahocquard: associations

src/Pim/Component/Catalog/Updater/Setter/AssociationFieldSetter.php
```diff
@@ -62,26 +62,50 @@ public function __construct(
...
     {
-        foreach ($product->getAssociations() as $association) {
-            foreach ($association->getGroups() as $group) {
-                $association->removeGroup($group);
-            }
-            foreach ($association->getProducts() as $prod) {
-                $association->removeProduct($prod);
-            }
+        if ($data === null) {
```
ahocquard: wrong style it is

src/Pim/Component/Catalog/Updater/Setter/AssociationFieldSetter.php
ahocquard: No impact on the imports (except this bug) ? Because it could change the previous behavior by default.

CHANGELOG-1.6.md
anaelChardan: DONE

features/Context/FixturesContext.php
anaelChardan: DONE

features/Context/FixturesContext.php
anaelChardan: DONE

src/Pim/Component/Catalog/Updater/Setter/AssociationFieldSetter.php
anaelChardan: Nop, no impact :-) (if tests are goods :-P)

features/import/product/import_products_with_associations.feature
```diff
@@ -155,3 +155,30 @@ Feature: Execute a job
     And I launch the import job
     And I wait for the "csv_footwear_product_import" job to finish
     Then I should see the text "skipped product (no associations detected)"
+
+  @jira https://akeneo.atlassian.net/browse/PIM-6042
+  Scenario: Successfully import product associations without remove already existing association when option "compare values" is set to true
```
ahocquard: Please fix it as well, same comments as the changelog :)

PR #5404, opened 2016-12-26
PHP CS Fixer consistency

grena:

Currently we have 2 files  for the PHP CS Fixer tool:
- `.php_cs.local.php`: Used to parse all files of the projects (excepts `*Spec.php`)
- `.php_cs.php`: Used to parse only files from a pull request (based on git diff)

This PR keeps one file, named `.php_cs.php`, which parses ALL files of the PIM (`/features` and `/src`, excepts `*Spec.php`).
CS Rules are unchanged (`Symfony\CS\FixerInterface::PSR2_LEVEL`).


anaelChardan:
Will it be slower ?

grena:
@anaelChardan It parses more files than the diff of the PR, so a bit slower, but very negligible for the consistency it brings imho :)

anaelChardan:
Yeah @grena :) I agree :)

PR #5406, opened 2016-12-26
Some fixes

grena:
- Return public views in the view selector
- Fallback on entity code as id for select2 if the normalized entity doesn't have one

### DIFF
--- a/src/Pim/Bundle/EnrichBundle/Resources/public/js/grid/view-selector.js
+++ b/src/Pim/Bundle/EnrichBundle/Resources/public/js/grid/view-selector.js
@@ -446,6 +446,10 @@ define(
                 return _.map(data, function (view) {
                     view.text = view.label;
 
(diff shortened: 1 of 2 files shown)


src/Pim/Bundle/EnrichBundle/Resources/public/js/grid/view-selector.js
```diff
@@ -446,6 +446,10 @@ define(
                 return _.map(data, function (view) {
                     view.text = view.label;
 
+                    if (!_.has(view, 'id') && _.has(view, 'code')) {
```
SamirBoulil: Don't know the business logic here, but it seems wierd to put a code in an id property ?

grena: Select2 just need its objects to have a property named "id", but can be a string, a code, whatever :)

PR #5409, opened 2016-12-27
PIM-6023: Add a behat step allowing us to hover an element

ClementGautier:
| Added Specs                       | :negative_squared_cross_mark:
| Review and 2 GTM                  | In progress

### DIFF
--- a/features/Context/WebUser.php
+++ b/features/Context/WebUser.php
@@ -1443,6 +1443,23 @@ class WebUser extends RawMinkContext
     }
 
     /**
+     * @param string locator
+     *
+     * @When /^I hover over the element "([^"]*)"$/
+     */
+    public function iHoverOverTheElement($locator)
+    {
+        $session = $this->getSession();
+        $element = $session->getPage()->find('css', $locator);
+
+        if (null === $element) {
+            throw new \InvalidArgumentException(sprintf('Could not evaluate CSS selector: "%s"', $locator));
+        }
+
+        $element->mouseOver();
+    }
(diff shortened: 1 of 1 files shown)


features/Context/WebUser.php
```diff
@@ -1443,6 +1443,23 @@ public function iPressTheButton($button)
     }
 
     /**
+     * @param string locator
+     *
+     * @When /^I hover over the element "([^"]*)"$/
```
grena: It's weird to insert some kind of CSS related in Gherkin. Can't we name it explicitly "I hover the icon" or I don't know what it is, then, **internally**, use this method?

anaelChardan: I agree with pipou

ClementGautier: I personally disagree. It is easier this way to maintain: Creating specific step is counter productive IMO and lead to what we have actually (meaning more gherkin than actual code to test).

anaelChardan: Ok :-), Good explanation :-), I GTM

features/Context/WebUser.php
```diff
@@ -1443,6 +1443,23 @@ public function iPressTheButton($button)
...
+     */
+    public function iHoverOverTheElement($locator)
+    {
+        $session = $this->getSession();
+        $element = $session->getPage()->find('css', $locator);
+
+        if (null === $element) {
```
phaseinducer: Please spin instead of doing this.

PR #5416, opened 2016-12-28
TIP-667: Adds ProductValueFactory and replace hardcoded productValue instanciations

SamirBoulil:

This PR introduces a ProductValue factory which responsibility is to solely create product values.

~This factory checks for the attribute localizable, scopable properties as well as if the locale we want the product value to be created is activated or if the scope exists.~

This factory relies on the AttributeValidatorHelper in order to check if the product value can be created within the scope and locale requested.

This simplifies our codebase as those checks were made in different layers.

This factory is being injected everywhere product values were created with new ProductValue() (eg, the ProductBuilder).

| Migration script                  | :red_circle:

:red_circle: Done but fail

:clock1: Done but pending

### DIFF
--- a/CHANGELOG-1.7.md
+++ b/CHANGELOG-1.7.md
@@ -26,7 +26,7 @@
 - GITHUB-5380: Add `Pim\Component\User\Model\GroupInterface`
 - GITHUB-4696: Ping the server before updating job and step execution data to prevent "MySQL Server has gone away" issue cheers @qrz-io!
 - TIP-575: Rename FileIterator classes to FlatFileIterator and changes the reader/processor behavior to iterate over the item's position in the file instead of the item's line number in the file.
-- TIP-662: Removed the WITH_REQUIRED_IDENTIFIER option from the flatToStandard product converter.
+- TIP-662: Removed the WITH_REQUIRED_IDENTIFIER option from `Pim\Component\Connector\ArrayConverter\FlatToStandard\Product` as it was not used anymore.
 - TIP-667: Introduce a product value factory service to instanciate product values.
 
 ## BC breaks
--- a/CHANGELOG-1.7.md
+++ b/CHANGELOG-1.7.md
@@ -12,7 +12,8 @@
 - In the _Product Query Builder_, aka _PQB_, (`Pim\Component\Catalog\Query\ProductQueryBuilderInterface`), filtering products by the following filters is now deprecated: `categories.id`, `family.id`, `groups.id`. 
   Filters `categories`, `family` and `groups` have been introduced and the _PQB_ now uses them by default. The filters `categories.code`, `family.code` and `groups.code` are deprecated. 
   In the next version, the deprecated filters will be removed.
-- As it's not needed anymore to convert `codes` to `ids` in order to filter products, `Pim\Bundle\CatalogBundle\Doctrine\Common\Filter\ObjectIdResolver` and `Pim\Bundle\CatalogBundle\Doctrine\Common\Filter\ObjectIdResolverInterface` are now deprecated. 
+- As it's not needed anymore to convert `codes` to `ids` in order to filter products, `Pim\Bundle\CatalogBundle\Doctrine\Common\Filter\ObjectIdResolver` and `Pim\Bundle\CatalogBundle\Doctrine\Common\Filter\ObjectIdResolverInterface` are now deprecated.
+- Creating a product value with the ProductBuilder (`Pim\Component\Catalog\Denormalizer\Standard\ProductValueDenormalizer`) using the `createProductValue` function is now deprecated. It is advised to use the ProductValueFactory (`Pim\Component\Catalog\Factory\ProductValueFactory`) instead.
 
 ## Functional improvements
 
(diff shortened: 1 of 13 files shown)


CHANGELOG-1.7.md
```diff
@@ -28,9 +26,12 @@
 - GITHUB-5380: Add `Pim\Component\User\Model\GroupInterface`
 - GITHUB-4696: Ping the server before updating job and step execution data to prevent "MySQL Server has gone away" issue cheers @qrz-io!
 - TIP-575: Rename FileIterator classes to FlatFileIterator and changes the reader/processor behavior to iterate over the item's position in the file instead of the item's line number in the file.
+- TIP-662: Removed the WITH_REQUIRED_IDENTIFIER option from the flatToStandard product converter.
```
jjanvier: FQCN only please.
And explain we dropped it because it was not used at all :)

(yes I know, I should have noticed that in the previous PR :p)

CHANGELOG-1.7.md
```diff
@@ -28,9 +26,12 @@
...
+- TIP-667: Introduce a product value factory service to instanciate product values.
 
 ## BC breaks
-
+- Change the constructor of `Pim\Component\Catalog\Denormalizer\Standard\ProductValueDenormalizer` to add `Pim\Component\Catalog\Factory\ProductValueFactory`
+- Change the constructor of `Pim\Component\Catalog\Builder\ProductBuilder` to add `Pim\Component\Catalog\Factory\ProductValueFactory`
```
jjanvier: I think you remove something too

src/Pim/Component/Catalog/Builder/ProductBuilderInterface.php
```diff
@@ -137,6 +137,8 @@ public function addProductValue(
      * @param string             $scope
      *
      * @return ProductValueInterface
+     *
+     * @deprecated will be removed in 1.8. Please use ProductValueFactory::create instead.
```
jjanvier: add this the deprecations list in the changelog please

src/Pim/Component/Catalog/Builder/ProductBuilder.php
```diff
@@ -222,49 +225,17 @@ public function addProductValue(
...
-                throw new \InvalidArgumentException(
-                    sprintf(
-                        'A scope must be provided to create a value for the scopable attribute %s',
-                        $attribute->getCode()
-                    )
-                );
-            }
-        }
-
-        return $value;
```
nidup: :fire:

src/Pim/Component/Catalog/Builder/ProductBuilderInterface.php
nidup: :+1:

src/Pim/Component/Catalog/Factory/ProductValueFactory.php
```diff
@@ -0,0 +1,112 @@
...
+     * @param AttributeInterface $attribute
+     * @param string             $channelCode
+     * @param string             $localeCode
+     *
+     * @return ProductValueInterface
+     *
+     */
+    public function create(AttributeInterface $attribute, $channelCode, $localeCode)
+    {
+        if ($attribute->isScopable()) {
```
jjanvier: we should use the `\Pim\Component\Catalog\Validator\AttributeValidatorHelper` for all this validation

src/Pim/Component/Catalog/Factory/ProductValueFactory.php
```diff
@@ -0,0 +1,112 @@
...
+class ProductValueFactory
+{
+    /** @var ChannelRepositoryInterface */
+    protected $channelRepository;
+
+    /** @var LocaleRepositoryInterface */
+    protected $localeRepository;
+
+    /** @var string */
+    private $productValueClass;
```
nidup: private? (i don't give a :hankey: i comment here just to be sure it's on purpose, idea is not to run an infinite debate about protected vs private, anyway this kind of class should never be extended / overriden :) )

SamirBoulil: nah :) good catch :) (Phpstom autogeneration)

CHANGELOG-1.7.md
SamirBoulil: idk if we should document it, because what we removed is a key-value from an array of options passed in the constructor. What do you think ?

CHANGELOG-1.7.md
SamirBoulil: done

CHANGELOG-1.7.md
jjanvier: ok, you're absolutely right

src/Pim/Component/Catalog/Factory/ProductValueFactory.php
```diff
@@ -0,0 +1,67 @@
...
+     * This method effectively creates an empty product value while checking the provided localeCode and ChannelCode
+     * exists.
+     * The Data for this product value should be set in a second time using ProductValue::setData method.
+     *
+     * @param AttributeInterface $attribute
+     * @param string             $channelCode
+     * @param string             $localeCode
+     *
+     * @return ProductValueInterface
+     *
```
jjanvier: @throw LogicException in case locale or channel code does not fit the attribute

we have to document it imho, because this is a real important use case
we don't handle the exception that comes from attributeValidatorHelper on purpose

SamirBoulil: Done

PR #5417, opened 2016-12-28
Add locale subscriber

solivier:
This subscriber will be used for the EE Activity Manager feature: when a locale (that concern a project) is deactivated we need to delete all projects that are using this locale as the projects can't be used anymore.

| Review and 2 GTM                  |  :clock1:


src/Pim/Bundle/CatalogBundle/Entity/Locale.php
```diff
@@ -147,9 +147,6 @@ public function setChannels($channels)
     public function addChannel(ChannelInterface $channel)
     {
         $this->channels[] = $channel;
-        if ($this->channels->count() > 0) {
```
jjanvier: I don't understand why you remove this internal logic
adding a setter, that will do this logic outside here is not the way to go imho

src/Pim/Bundle/CatalogBundle/Entity/Locale.php
```diff
@@ -184,4 +178,14 @@ public function getName()
 
         return array_key_exists($this->code, $localeNames) ? $localeNames[$this->code] : null;
     }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function setActivated($activated)
```
jjanvier: I would not add that personnally

src/Pim/Bundle/CatalogBundle/EventSubscriber/LocaleSubscriber.php
```diff
@@ -0,0 +1,59 @@
...
+use Symfony\Component\EventDispatcher\EventDispatcherInterface;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\EventDispatcher\GenericEvent;
+
+/**
+ * @author    Olivier Soulet <olivier.soulet@akeneo.com>
+ * @copyright 2016 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+class LocaleSubscriber implements EventSubscriberInterface
```
jjanvier: we should be more specific
like LocaleActivationSubscriber

src/Pim/Bundle/CatalogBundle/EventSubscriber/LocaleSubscriber.php
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    /**
+     * @param GenericEvent $event
+     */
+    public function localeStatus(GenericEvent $event)
+    {
+        $locale = $event->getSubject();
+
+        if (!$locale instanceof LocaleInterface) {
```
jjanvier: the locale is (de) activated when there is a change the the channel
so we should listen on channel events

jjanvier: maybe the best is even to do that in the ChannelSaver

jjanvier: I'm available IRL for that

ClementGautier:
The code is good but I don't agree on adding an other event. Basically, listeners and subscribers on this events could've been done on the StorageEvents::PRE_SAVE sa well. In my opinion, it is 250 lines of code + same for the doc which will serve the cause of 1 or 2 edge case which could've used the other event.

So, as it is not necessary, I downvote. (again, don't take it personally: the code is great)

solivier:
@ClementGautier my bad I forget to explain the case in the PR description, in fact it's will not only be used for edge cases, we need it for the AM, in fact when the locale (that concern a project) is deactivated we need to delete all projects that are using this locale.

ClementGautier:
@solivier Then listen to `StorageEvents::PRE_SAVE` and do the logic you are doing here. Activity manager is an edge case => outside this case, I dno't know any partner / customer that change locale state and did implement a custom behavior on that. 

You can even create an abstract class `LocaleUpdatedSubscriber` that store the logic if you don't want to repeat these 3 lines on multiple Subscribers but in my opinion it is not useful.

jjanvier:
but having a business event 'LOCALE_(DE)ACTIVATED' was really good imho
even I don't know what is the best way to do it (another subscriber, directly in the ChannelSaver, or whatever)

the locale (de)activation is something that is important for the PIM and that can lead to real significant changes
and like every other catalog structure change, this is something we don't handle well at the moment (we don't handle it at all actualy)
this event, was a first step to resolve this problem

cc @ClementGautier @solivier

src/Pim/Component/Catalog/LocaleEvents.php
```diff
@@ -0,0 +1,33 @@
...
+
+    /**
+     * This event is dispatched each time a locale is activated.
+     *
+     * The event listener receives an
+     * Pim\Bundle\EnrichBundle\Event\BuildProductValueFormEvent instance.
+     *
+     * @staticvar string
+     */
+    const LOCALE_ACTIVATED = 'pim_enrich.locale_activated';
```
fitn: I think your phpdoc is wrong for these two constants. It should come from catalog and be named 'pim_catalog....' ;)

PR #5440, opened 2017-01-02
API-19: add a business exception in updaters for non existent fields

ahocquard:

In the API, we need to have specific error messages when we update non existent fields. 
For example : 

> "Property "%s" does not exist. Check the standard format documentation.

An exception "NoSuchPropertyException was already raised by Symfony. Unfortunately, this exception has just a message :

> "Cannot read property "%s" from an array. Maybe you intended to write the property path as "[%s]" instead"

If we want to get the property in a controller (for a custom message as above), it's not possible without a regex parsing the error message in NoSuchPropertyException... and that's dirty :)

This PR resolve that problem by raising business exception directly in the updaters, with a property which is accessible in the exception.

| Added Behats                      | :red_circle: 
| Changelog updated                 | :clock1:
| Review and 2 GTM                  |  :red_circle:

:red_circle: Done but fail

:clock1: Done but pending

### DIFF
--- a/src/Akeneo/Component/Classification/Updater/CategoryUpdater.php
+++ b/src/Akeneo/Component/Classification/Updater/CategoryUpdater.php
@@ -3,9 +3,13 @@
 namespace Akeneo\Component\Classification\Updater;
 
 use Akeneo\Component\Classification\Model\CategoryInterface;
+use Akeneo\Component\StorageUtils\Exception\InvalidObjectException;
+use Akeneo\Component\StorageUtils\Exception\InvalidPropertyException;
+use Akeneo\Component\StorageUtils\Exception\UnknownPropertyException;
 use Akeneo\Component\StorageUtils\Repository\IdentifiableObjectRepositoryInterface;
 use Akeneo\Component\StorageUtils\Updater\ObjectUpdaterInterface;
 use Doctrine\Common\Util\ClassUtils;
+use Symfony\Component\PropertyAccess\Exception\NoSuchPropertyException;
 use Symfony\Component\PropertyAccess\PropertyAccess;
 use Symfony\Component\PropertyAccess\PropertyAccessor;
 
@@ -39,11 +43,9 @@ class CategoryUpdater implements ObjectUpdaterInterface
     public function update($category, array $data, array $options = [])
     {
         if (!$category instanceof CategoryInterface) {
-            throw new \InvalidArgumentException(
-                sprintf(
-                    'Expects a "Akeneo\Component\Classification\Model\CategoryInterface", "%s" provided.',
-                    ClassUtils::getClass($category)
-                )
+            throw InvalidObjectException::objectExpected(
+                ClassUtils::getClass($category),
+                'Akeneo\Component\Classification\Model\CategoryInterface'
             );
         }
 
@@ -58,6 +60,9 @@ class CategoryUpdater implements ObjectUpdaterInterface
      * @param CategoryInterface $category
      * @param string            $field
      * @param mixed             $data
+     *
+     * @throws InvalidPropertyException
+     * @throws UnknownPropertyException
      */
     protected function setData(CategoryInterface $category, $field, $data)
     {
@@ -70,10 +75,21 @@ class CategoryUpdater implements ObjectUpdaterInterface
             if (null !== $categoryParent) {
                 $category->setParent($categoryParent);
             } else {
-                throw new \InvalidArgumentException(sprintf('The parent category "%s" does not exist', $data));
+                throw InvalidPropertyException::validEntityCodeExpected(
+                    'parent',
+                    'category code',
+                    'The category does not exist',
+                    'updater',
+                    'category',
+                    $data
+                );
             }
         } else {
-            $this->accessor->setValue($category, $field, $data);
+            try {
+                $this->accessor->setValue($category, $field, $data);
+            } catch (NoSuchPropertyException $e) {
+                throw UnknownPropertyException::unknownProperty($field, $e);
+            }
         }
     }
 
--- a/src/Pim/Component/Connector/Processor/Denormalization/JobInstanceProcessor.php
+++ b/src/Pim/Component/Connector/Processor/Denormalization/JobInstanceProcessor.php
@@ -10,6 +10,7 @@ use Akeneo\Component\Batch\Job\JobRegistry;
 use Akeneo\Component\Batch\Model\JobInstance;
 use Akeneo\Component\Batch\Step\StepExecutionAwareInterface;
 use Akeneo\Component\StorageUtils\Detacher\ObjectDetacherInterface;
+use Akeneo\Component\StorageUtils\Exception\ObjectUpdaterException;
 use Akeneo\Component\StorageUtils\Factory\SimpleFactoryInterface;
 use Akeneo\Component\StorageUtils\Repository\IdentifiableObjectRepositoryInterface;
 use Akeneo\Component\StorageUtils\Updater\ObjectUpdaterInterface;
@@ -85,6 +86,8 @@ class JobInstanceProcessor extends AbstractProcessor implements ItemProcessorInt
 
         try {
             $this->updater->update($entity, $item);
+        } catch (ObjectUpdaterException $exception) {
+            $this->skipItemWithMessage($item, $exception->getMessage(), $exception);
         } catch (\InvalidArgumentException $exception) {
             $this->skipItemWithMessage($item, $exception->getMessage(), $exception);
         }
--- a/src/Pim/Component/Connector/Processor/Denormalization/Processor.php
+++ b/src/Pim/Component/Connector/Processor/Denormalization/Processor.php
@@ -5,6 +5,7 @@ namespace Pim\Component\Connector\Processor\Denormalization;
 use Akeneo\Component\Batch\Item\ItemProcessorInterface;
 use Akeneo\Component\Batch\Step\StepExecutionAwareInterface;
 use Akeneo\Component\StorageUtils\Detacher\ObjectDetacherInterface;
+use Akeneo\Component\StorageUtils\Exception\ObjectUpdaterException;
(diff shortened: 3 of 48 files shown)


src/Akeneo/Component/Classification/Updater/CategoryUpdater.php
```diff
@@ -4,8 +4,10 @@
 
 use Akeneo\Component\Classification\Model\CategoryInterface;
 use Akeneo\Component\StorageUtils\Repository\IdentifiableObjectRepositoryInterface;
+use Akeneo\Component\StorageUtils\Updater\BadPropertyException;
```
nidup: what means "Bad"? "NonExistingException" should be better no ?

src/Akeneo/Component/Classification/Updater/CategoryUpdater.php
```diff
@@ -73,7 +78,11 @@ protected function setData(CategoryInterface $category, $field, $data)
                 throw new \InvalidArgumentException(sprintf('The parent category "%s" does not exist', $data));
```
nidup: This exception is too generic too, we could introduce a new business exception (maybe extending InvalidArgumentException or not). ping @jjanvier

src/Akeneo/Component/Classification/Updater/CategoryUpdater.php
```diff
@@ -73,7 +78,11 @@ protected function setData(CategoryInterface $category, $field, $data)
...
             }
         } else {
-            $this->accessor->setValue($category, $field, $data);
+            try {
+                $this->accessor->setValue($category, $field, $data);
+            } catch (NoSuchPropertyException $e) {
```
nidup: Agree on this, this exception is technical, we can use Symfony property accessor or manual stuff it's an implementation details, introducing a business exception attached to this layer is a good idea imho.

src/Akeneo/Component/StorageUtils/Updater/BadPropertyException.php
```diff
@@ -0,0 +1,38 @@
...
+namespace Akeneo\Component\StorageUtils\Updater;
+
+/**
+ * Business exception to handle Symfony exceptions thrown during updates.
+ *
+ * @author    Alexandre Hocquard <alexandre.hocquard@akeneo.com>
+ * @copyright 2016 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+class BadPropertyException extends \RuntimeException
```
nidup: Sure about RuntimeException ? Maybe a Domain Exception http://php.net/manual/en/class.domainexception.php ?

src/Akeneo/Component/Classification/Updater/CategoryUpdater.php
jjanvier: TIP-322
but it's a really generic card

src/Akeneo/Component/StorageUtils/Updater/BadPropertyException.php
nidup: just for you @jjanvier :wink:

ahocquard: I did that because it could occur only at runtime and not at compile time, and because the original error in Symfony inherit from that type.

jjanvier: as it's our business exception, it should inherit at least \LogicException
those exceptions should be treated handled in a higher level API and I suppose that's what you try to do 

RuntimeException are made for things we should not handle

src/Akeneo/Component/StorageUtils/Updater/BadPropertyException.php
```diff
@@ -0,0 +1,38 @@
...
+{
+    /** @var string */
+    protected $property;
+
+    /**
+     * @param string          $property
+     * @param \Exception|null $previous
+     * @param int             $code
+     */
+    public function __construct($property = '', \Exception $previous = null, $code = 0)
```
jjanvier: I prefer to stick the maximum possible to a normal Exception
and please do not put a default value for your property argument

something like `public function __construct($property, $message = '', $code = 0, Exception $previous = null) { }` would be better imho

src/Akeneo/Component/StorageUtils/Updater/BadPropertyException.php
ahocquard: And DomainException inherit from Logic Exception : "Exception that represents error in the program logic. This kind of exception should lead directly to a fix in your code. " And that's not the case, it's not a logic error imho (just bad input in the request).

src/Akeneo/Component/StorageUtils/Updater/BadPropertyException.php
```diff
@@ -0,0 +1,38 @@
+<?php
+
+namespace Akeneo\Component\StorageUtils\Updater;
+
+/**
+ * Business exception to handle Symfony exceptions thrown during updates.
```
jjanvier: nope
business exceptions are not made to handle technical problems
that can't be your description :) explain really what you want to achieve instead :)

the fact it comes from the Symfony Property Accessor is a detail

src/Akeneo/Component/StorageUtils/Updater/BadPropertyException.php
```diff
@@ -0,0 +1,38 @@
+<?php
+
+namespace Akeneo\Component\StorageUtils\Updater;
```
jjanvier: we are used to put our exceptions in an `Exception` folder
maybe it's not the best, I understand why you put it close to your object (personally, I prefer your way), but I think we should try to be uniform

src/Akeneo/Component/StorageUtils/Updater/UnknownPropertyException.php
```diff
@@ -0,0 +1,38 @@
+<?php
+
+namespace Akeneo\Component\StorageUtils\Updater;
+
+/**
+ * Business exception to handle Symfony exceptions thrown during updates.
```
phaseinducer: Not a good description as it could be also thrown in other cases ; )
Explain what the exception means is enough imho.

phaseinducer:
Imho we should treat it as a BC break.

src/Akeneo/Component/StorageUtils/Updater/UnknownPropertyException.php
```diff
@@ -0,0 +1,38 @@
...
+{
+    /** @var string */
+    protected $property;
+
+    /**
+     * @param string          $property
+     * @param \Exception|null $previous
+     * @param int             $code
+     */
+    public function __construct($property = '', \Exception $previous = null, $code = 0)
```
jjanvier: I prefer to stick the maximum possible to a normal Exception
and please do not put a default value for your property argument

something like public function __construct($property, $message = '', $code = 0, Exception $previous = null) { } would be better imho

src/Akeneo/Component/StorageUtils/Updater/UnknownPropertyException.php
```diff
@@ -0,0 +1,38 @@
...
+namespace Akeneo\Component\StorageUtils\Updater;
+
+/**
+ * Business exception to handle Symfony exceptions thrown during updates.
+ *
+ * @author    Alexandre Hocquard <alexandre.hocquard@akeneo.com>
+ * @copyright 2016 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+class UnknownPropertyException extends \RuntimeException
```
jjanvier: as it's our business exception, it should inherit at least \LogicException
those exceptions should be treated handled in a higher level API and I suppose that's what you try to do

RuntimeException are made for things we should not handle

src/Akeneo/Component/StorageUtils/Updater/UnknownPropertyException.php
jjanvier: business exceptions are not made to handle technical problems
that can't be your description :) explain really what you want to achieve instead :)

the fact it comes from the Symfony Property Accessor is a detail

src/Akeneo/Component/StorageUtils/Updater/UnknownPropertyException.php
```diff
@@ -0,0 +1,38 @@
+<?php
+
+namespace Akeneo\Component\StorageUtils\Updater;
```
jjanvier: we are used to put our exceptions in an `Exception` folder
maybe it's not the best, I understand why you put it close to your object (personally, I prefer your way), but I think we should try to be uniform

src/Pim/Component/Connector/Processor/Denormalization/JobInstanceProcessor.php
```diff
@@ -85,6 +86,8 @@ public function process($item)
 
         try {
             $this->updater->update($entity, $item);
+        } catch (ObjectUpdaterException $exception) {
+            $this->skipItemWithMessage($item, $exception->getMessage(), $exception);
         } catch (\InvalidArgumentException $exception) {
```
nidup: we still have this case or we keep it for backward compatibility reason? (both cases would be valid)

src/Pim/Component/Connector/Processor/Denormalization/Processor.php
```diff
@@ -63,6 +64,8 @@ public function process($item)
 
         try {
             $this->updater->update($entity, $item);
+        } catch (ObjectUpdaterException $exception) {
+            $this->skipItemWithMessage($item, $exception->getMessage(), $exception);
         } catch (\InvalidArgumentException $exception) {
```
nidup: same comment than previous, even more relevant here as this processor is the one used in 90% of cases

ahocquard:
Done in another branch : TIP-322-updaters.

PR #5454, opened 2017-01-04
PIM-6038: Update 'update' product property in the Product Updater

SamirBoulil:

**Why ?**

Today, in mongodb the products are stored using documents containing:
- the product fields
- normalizedData which contains the normalized product values of the product + the updated property

When we use the PQB to retrieve products that were updated since a date (last export for instance). We filter on the normalizedDate.updatedAt date.

However on an import, the updated date is only updated in the fields (not the normalizedData.updated property).

**What do we do from here ?**
- We could add a specific datetime filter on the PQB that filters on datetime for the field (not for the one in the normalizedData which is done by default).
- We could make sure the updatedAt date in the normalizedData is set properly. *(implemented solution)*
- How about the `\Pim\Bundle\CatalogBundle\EventSubscriber\TimestampableSubscriber` ?

**Implementation:**

- The timestampableSubscriber is used in mongo to update the dates when saving the data. It is still used for every other entities. However managing the updated date of products is a different deal because the PQB DateTime filter relies on the normalizedData.updated attribute.
- Updates the update date of the product directly in the updater.
- Updates the created date of the product directly in the updater
   - Because normalizers should not instanciate date variables
   - for consistency

| Blue CI                                    | :white_check_mark: (EE, CE, ORM, Mongo) |

:red_circle: Done but fail

:clock1: Done but pending

### DIFF
--- a/spec/Pim/Bundle/BaseConnectorBundle/Archiver/InvalidItemsCsvArchiverSpec.php
+++ b/spec/Pim/Bundle/BaseConnectorBundle/Archiver/InvalidItemsCsvArchiverSpec.php
@@ -10,6 +10,7 @@ use PhpSpec\ObjectBehavior;
 use Pim\Bundle\BaseConnectorBundle\EventListener\InvalidItemsCollector;
 use Pim\Component\Connector\Writer\File\CsvWriter;
 use Prophecy\Argument;
+use Symfony\Component\Finder\Adapter\AdapterInterface;
 
 class InvalidItemsCsvArchiverSpec extends ObjectBehavior
 {
@@ -42,12 +43,13 @@ class InvalidItemsCsvArchiverSpec extends ObjectBehavior
         $this->archive($jobExecution);
     }
 
-    function it_archives_unvalid_items(
+    function it_archives_invalid_items(
         InvalidItemsCollector $collector,
         CsvWriter $writer,
         JobExecution $jobExecution,
         JobInstance $jobInstance,
-        Filesystem $filesystem
+        Filesystem $filesystem,
+        AdapterInterface $adapter
     ) {
         $collector->getInvalidItems()->willReturn(['items']);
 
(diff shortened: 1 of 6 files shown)


spec/Pim/Bundle/BaseConnectorBundle/Archiver/InvalidItemsCsvArchiverSpec.php
```diff
@@ -10,6 +10,7 @@
 use Pim\Bundle\BaseConnectorBundle\EventListener\InvalidItemsCollector;
```
SamirBoulil: Spec broken on 1.5. So I fixed it in my PR

src/Pim/Component/Catalog/Updater/ProductUpdater.php
```diff
@@ -113,14 +113,25 @@ public function update($product, array $data, array $options = [])
...
 
+        $isProductUpdated = false;
+
         foreach ($data as $field => $values) {
             if (in_array($field, ['enabled', 'family', 'categories', 'variant_group', 'groups', 'associations'])) {
-                $this->updateProductFields($product, $field, $values);
+                $isProductUpdated |= $this->updateProductFields($product, $field, $values);
             } else {
-                $this->updateProductValues($product, $field, $values);
+                $isProductUpdated |= $this->updateProductValues($product, $field, $values);
```
solivier: not a big fan of this notation, "explicit is better than implicit" just my opinion

SamirBoulil: ;) Alright then

SamirBoulil: fixed.

src/Pim/Component/Catalog/Updater/ProductUpdater.php
```diff
@@ -131,10 +142,14 @@ public function update($product, array $data, array $options = [])
...
      * @param string           $field
      * @param mixed            $value
+     *
+     * @return boolean whether the product has been updated
      */
     protected function updateProductFields(ProductInterface $product, $field, $value)
     {
         $this->propertySetter->setData($product, $field, $value);
+
+        return true;
```
jjanvier: is it OK if nothing has been modified?
I mean, what means the updated_at? It it when the product is saved or when the product is really modified?

SamirBoulil: Ok, in reality what we do here is replace with the given value (even if it's the same).

You are right, what we need to do here is check if the product value's value is the same as the value we want to add.

PR #5490, opened 2017-01-11
TIP-671: Deprecate ProductBuilder method "addPriceForCurrencyWithData"

damien-carcel:
## Description
Right now, the ProductBuilder can add prices to the product value with two different method:
- `addPriceForCurrency` which will only add a currency, with no amount,
- `addPriceForCurrencyWithData` which add both amount and currency.

This could be done by only one method, for which the amount can be null. So `addPriceForCurrencyWithData` becomes deprecated, and only `addPriceForCurrency` is used everywhere in the PIM.

### Bonus

Since our integration tests are ran on Travis, it is mandatory to install the PIM. However, the minimal catalog is enough, and way faster to install than icecat_demo_dev, so a new `app/config/parameters_test.yml.dist` model is added and used by Travis.

## Definition Of Done


damien-carcel:
Done!

PR #5498, opened 2017-01-12
Fix CsvEncoder EOL

jmleroux:
Two small improvements:
- use binary mode for open file
- close the file handler when encoding is done

### DIFF
--- a/src/Pim/Component/Connector/Encoder/CsvEncoder.php
+++ b/src/Pim/Component/Connector/Encoder/CsvEncoder.php
@@ -150,9 +150,6 @@ class CsvEncoder implements EncoderInterface
     {
         rewind($csvResource);
 
-        $stat = fstat($csvResource);
-        ftruncate($csvResource, $stat['size']-1);
-
         return stream_get_contents($csvResource);
     }
 
--- a/src/Pim/Component/Connector/Encoder/CsvEncoder.php
+++ b/src/Pim/Component/Connector/Encoder/CsvEncoder.php
@@ -52,7 +52,7 @@ class CsvEncoder implements EncoderInterface
 
         $this->initializeContext($context);
 
-        $output = fopen('php://temp', 'r+');
+        $output = fopen('php://temp', 'r+b');
 
         $first = reset($data);
         if (isset($first) && is_array($first)) {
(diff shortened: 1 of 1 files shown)


jmleroux:
And for the record, we could move for the Symfony standard encoder when we will bump version : http://symfony.com/blog/new-in-symfony-3-2-csv-and-yaml-encoders-for-serializer

src/Pim/Component/Connector/Encoder/CsvEncoder.php
```diff
@@ -147,6 +150,9 @@ private function readCsv($csvResource)
     {
         rewind($csvResource);
 
+        $stat = fstat($csvResource);
+        ftruncate($csvResource, $stat['size']-1);
```
jjanvier: stupid quetion
are you sure it works the same on Linux/Windows/MacOS?
does a "\n" is always one byte length?

jjanvier: moreover, I don't understand the problem of having a blank line at the end
I think we have it for all files don't we?

jmleroux:
> I think we have it for all files don't we?

src/Pim/Component/Connector/Encoder/CsvEncoder.php
jmleroux: > I think we have it for all files don't we?
Yep, but this encoder is not to write in a file.
Its purpose is to encode an array into a csv string.

jmleroux: > are you sure it works the same on Linux/Windows/MacOS?
does a "\n" is always one byte length?

Far from stupid question !
I'll check.

BitOne: Yes, "\n" is always one byte length.
But if you have doubt, use `mb_strlen('\n')`, it would make your code easier to read than with the "-1".

jmleroux: but maybe the simplest solution is to use `trim` instead

jmleroux:
Improvements can be done only on master

PR #5500, opened 2017-01-12
Less magic integration tests

jjanvier:
## What's here?

This PR aims to solve a few problems we currently have with our integration tests:
- the generic  namespace  `Test\Integration` has been replaced by `Akeneo\Test\Integration`
- the technical SQL catalog has been moved to the root tests directory to be able to use it for others tests (for example we need it with Damien)
- it is now possible to load import file (CSV, YML), SQL files and Mongo (JSON) files
- now we don't have to override the properties of the `TestCase` class to have a specific behaviour. Everything is handled in a separate `Configuration` class.
- the loading of the files has been simplified. Now, we just specify the directoriales we want to load. No more magic with the catalog name, the path, the extra directories etc...
- there is an helper to get the real path of a fixture file

## The impacts for your integration tests

For those impacted by this change, now you integration tests must:
- remove the properties you have override
- extends the class `Akeneo\Test\Integration\TestCase`
- implement the method `protected function getConfiguration()` which returns a `Akeneo\Test\Integration\Configuration` object

The ```getConfiguration``` can be implemented like this:

```
    protected function getConfiguration()
    {
        return new Configuration(
            [// the list of dir you want to load],
            // do you want to purge the database for each test
        );
    }
```

If you want to load the technical catalog, you can do:

```
    protected function getConfiguration()
    {
        return new Configuration(
            [Configuration::getTechnicalCatalogPath()],
            false
        );
    }
````

| Added integration tests           | :white_check_mark: 
| Micro Demo to the PO (Story only) |-

### CODE REVIEW

File: tests/Configuration.php [hunk 1]
```diff
@@ -0,0 +1,112 @@
...
+                throw new \Exception(sprintf('The catalog directory "%s" does not exist.', $catalogDirectory));
+            }
+        }
+
+        $this->fixtureDirectories = [];
+        if (null !== $fixtureDirectories) {
+            $this->fixtureDirectories = array_merge($this->fixtureDirectories, $fixtureDirectories);
+        }
+
+        $defaultFixturePath = self::getRootDirectory() . DIRECTORY_SEPARATOR . 'tests' .
```
pierallard: for more readability, I prefer
`implode(DIRECTORY_SEPARATOR, [ self::getRootDirectory(), 'tests', 'fixtures' ])`

Ruby-style ;)

tests/TestCase.php
```diff
@@ -152,27 +183,93 @@ protected function loadCatalog($files)
...
     {
-        $directories = array_merge([$this->catalogDirectory], $this->extraDirectories);
+        $filesByType = [
+            'sql' => [],
+            'mongodb' => [],
+            'import' => [],
+        ];
+
+        foreach ($files as $filePath) {
+            //TODO: should be done in the getFilesToLoad method
```
pierallard: TODO remaining

PR #5510, opened 2017-01-13
AMS-6: Add gherkin to click on icon button

grena:

For now we have gherkin to click on buttons, but we have icon button components that can only be identified with their on-hover title.
This method allows to retrieve them.

### DIFF
--- a/features/Context/Page/Base/Base.php
+++ b/features/Context/Page/Base/Base.php
@@ -7,6 +7,7 @@ use Behat\Mink\Exception\ElementNotFoundException;
 use Behat\Mink\Exception\UnsupportedDriverActionException;
 use Context\FeatureContext;
 use Context\Spin\SpinCapableTrait;
+use Context\Traits\ClosestTrait;
 use Pim\Behat\Decorator\ElementDecorator;
 use SensioLabs\Behat\PageObjectExtension\PageObject\Element;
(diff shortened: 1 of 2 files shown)


features/Context/Page/Base/Base.php
```diff
@@ -237,6 +239,25 @@ public function getButton($locator)
...
+     * Get icon button
+     *
+     * @param string  $locator
+     *
+     * @return NodeElement|null
+     */
+    public function getIconButton($locator)
+    {
+        $button = null;
+        $icon = $this->find('css', sprintf('i[data-original-title="%s"]', $locator));
```
pierallard: I think usage of .AknIconButton is better here.

grena: Ok 👌

features/Context/Page/Base/Base.php
```diff
@@ -237,6 +239,25 @@ public function getButton($locator)
...
+     *
+     * @return NodeElement|null
+     */
+    public function getIconButton($locator)
+    {
+        $button = null;
+        $icon = $this->find('css', sprintf('.AknIconButton[data-original-title="%s"]', $locator));
+
+        if (null !== $icon) {
+            $button = $this->getClosest($icon, 'a');
```
pierallard: This will not work @grena 
"getClosest" only works with classes. It will always be null.

PR #5418, opened 2016-12-28
TIP-669: Store raw values in the product

SamirBoulil:
The goal of this PR is *_only_* to be able to store Product Values as a JSON column in MySQL.
_Only_ meaning nothing has been deleted. The values are also saved just like before. This is the first piece of a whole. 

To do that we (by order of commit):
- added a new column in the product mapping and normalized the values by subscribing to the `PRE_SAVE` event. This logic has not been put in the saver directly as it's only a technical "patapouille".  
   - an integration test about the product save has been introduced
- introduced new _Storage_ normalizers
  - they compose whenever they can with the _Standard_ normalizers
  - they are different in the way product values are indexed (we'll need it later) 
  - they normalize more data about metrics (we need to keep the base data, the base unit and the family) 
  - they have integration tests 
- allowed the _Standard_ normalizers to really be compound 

## Why this format for storing product values?

We used the format we defined in the POC. For sure we'll need something similar for the indexation. To store the values we could _almost_ have used the standard format. Almost because of the metrics (see above). By the way, maybe we should provide those extended data in the standard format of the metric. I'll write an RFC about it.

The main important things to keep in mind is that:
 - this format is based on the standard
 - this format is fully tested
 - this format can be changed later, after discussion, without any problem. It's just an internal storage format. 

## There are some ugly todos in this PR!!

Yes. You're right. Burn me.

1. it's not on master :p 
2. I'm waiting for @skeleton. I have a few questions to ask her about the integration tests.
3. I'm waiting for https://github.com/akeneo/pim-community-dev/pull/5462 to be merged to fix the metric problems

| Q                                 | A
| --------------------------------- | ---
| Added Specs                       | Y
| Added integration tests | Y
| Added Behats                      | -
| Changelog updated                 | -
| Review and 2 GTM                  | -
| Micro Demo to the PO (Story only) | -
| Migration script                  | -
| Tech Doc                          | -

PR #5521, opened 2017-01-16
TIP-682: Update to Symfony 2.7.23

damien-carcel:
## Description

This PR updates the PIM to the last version of Symfony 2.7.

This version come with a new behavior on the ChoiceType field from the Form component (https://symfony.com/doc/2.7/reference/forms/types/choice.html#choices-as-values), causing the labels of the datagrid filters operators to be replaced by their codes (simple numbers, see screenshot below).

image

This problem is fixed by the second commit.

## Definition Of Done

| Added Specs                       | :negative_squared_cross_mark:

### DIFF
--- a/app/config/config.yml
+++ b/app/config/config.yml
@@ -7,7 +7,7 @@ imports:
 framework:
     #esi:             ~
     translator:      { fallback: en }
-    secret:          %secret%
+    secret:          "%secret%"
     router:
         resource: "%kernel.root_dir%/config/routing.yml"
         strict_requirements: "%kernel.debug%"
@@ -18,8 +18,8 @@ framework:
         enabled: true
         api: 2.5
     templating:      { engines: ['twig', 'php'] } #assets_version: SomeVersionScheme
-    default_locale:           "%locale%"
-    trusted_proxies:          ~
+    default_locale:  "%locale%"
+    trusted_proxies: ~
     session:
         name:                 BAPID
         handler_id:           session.handler.pdo
--- a/composer.json
+++ b/composer.json
@@ -42,21 +42,21 @@
         "jms/serializer-bundle": "1.0.0",
         "knplabs/knp-menu": "2.0.1",
         "knplabs/knp-menu-bundle": "2.0.0",
-        "kriswallsmith/assetic": "1.1.3",
+        "kriswallsmith/assetic": "1.4.0",
         "leafo/lessphp": "0.5.0",
         "league/flysystem": "1.0.11",
         "league/flysystem-ziparchive": "1.0.2",
         "liip/imagine-bundle": "1.3.0",
         "imagine/imagine": "0.6.2",
-        "monolog/monolog": "1.18.2",
+        "monolog/monolog": "1.22.0",
         "oneup/flysystem-bundle": "1.1.0",
-        "sensio/distribution-bundle": "~4.0",
-        "sensio/framework-extra-bundle": "~3.0,>=3.0.2",
-        "sensio/generator-bundle": "2.3.5",
-        "symfony/assetic-bundle": "2.3.1",
-        "symfony/monolog-bundle": "2.10.0",
-        "symfony/swiftmailer-bundle": "2.3.8",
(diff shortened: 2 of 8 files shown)


SamirBoulil:
I'm wondering why don't we just bump it up to the symfony LTS directly ? 2.8.x I believe ?

app/config/config.yml
```diff
@@ -7,7 +7,7 @@ imports:
 framework:
     #esi:             ~
     translator:      { fallback: en }
-    secret:          %secret%
+    secret:          "%secret%"
```
jjanvier: I'm pinailling, but don't we use simple quotes usually?

jjanvier:
@SamirBoulil because it' more work actually, and it's not the topic of the single storage
here we just make sure the PIM will work without this bug https://github.com/akeneo/pim-community-dev/pull/5418#discussion_r95166453

jjanvier:
@SamirBoulil and btw, the 2.7 is already a LTS

app/config/config.yml
damien-carcel: We do, but Symfony don't, and every configuration files in "app/" is double quoted.

src/Pim/Component/Catalog/Builder/ProductBuilderInterface.php
```diff
@@ -105,6 +105,8 @@ public function addPriceForCurrencyWithData(ProductValueInterface $value, $curre
      *
      * @param ProductValueInterface $value
      * @param array                 $currencies
+     *
+     * @deprecated Will be removed in 1.8.
```
skeleton: humm, what's the link with this PR ?

composer.json
```diff
@@ -95,15 +95,11 @@
...
         "post-update-cmd": [
-            "Incenteev\\ParameterHandler\\ScriptHandler::buildParameters",
-            "Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::buildBootstrap",
-            "Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::clearCache",
-            "Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::installAssets",
-            "Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::installRequirementsFile",
-            "Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::removeSymfonyStandardFiles",
-            "Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::prepareDeploymentTarget",
-            "php app/console fos:js-routing:dump --target=web/js/routes.js"
+            "@symfony-scripts"
```
pierallard: Why not put all this in an "post-autoload-dump" to avoid 2 declarations ?

src/Oro/Bundle/FilterBundle/Filter/AbstractFilter.php
```diff
@@ -82,7 +82,7 @@ public function getMetadata()
         $operatorChoices = $formBuilderType->getOption('choices');
 
         $choices = [];
-        foreach ($operatorChoices as $key => $choice) {
+        foreach ($operatorChoices as $choice => $key) {
```
pierallard: Weeeeeeird! Are you sure than all the previous "choices" have no duplicates before flipping it ? You can erase some data doing this.

composer.json
damien-carcel: Because that is how Symfony do it. I didn't look further ;-).

src/Oro/Bundle/FilterBundle/Filter/AbstractFilter.php
damien-carcel: No, this is the real Symfony 2.7 behavior. Choices for ChoiceType fields must be as:
```php
[
    'label' => 'code',
    ...
]
```
What is wierd is that it worked at all before. For me, it was a bug of Symfony 2.7.2 that has been corrected.

The behavior I produce with the current code is the exact same behavior than in Symfony 3, so no problem the day we want to bump to it.

PR #5522, opened 2017-01-16
Tip 475 : fix an error thrown when saving a product with a media

ahocquard:

Since the standard format, there is a bug when you try to save a product which has already a media. It's due to a wrong condition check on an empty array.

Unfortunately, it hid a problem with boolean values.

When a product is saved with the UI, boolean product values are set to false, even if they are not modified and null in database (and only for attributes belonging to the family).

This is a well-known behaviour and it has some consequences on the draft.

When you make a proposal on a draft, boolean attributes belonging to the family and no set are in the proposal : the UI set them automatically to false.
Then, the proposal has unmodified boolean values, which is not normal.

In order to avoid that problem, we have two solutions : 
- set automatically the boolean values to "false" when we create a product
- fix the problem in the UI by supporting null values 

The second option will be normally done by @juliensnz . It's could be long and has a lot of impacts. Awaiting that, I did the first option.

I will open a ticket to create a migration script before the 1.7 release, in case of the second option is not done.

| Added Specs                       | :white_check_mark: :red_circle: :negative_squared_cross_mark: :clock1:
| Added Behats                      | :white_check_mark: :red_circle: :negative_squared_cross_mark: :clock1:
| Added integration tests           | :white_check_mark: :red_circle: :negative_squared_cross_mark: :clock1:
| Changelog updated                 | :white_check_mark: :red_circle: :negative_squared_cross_mark: :clock1:
| Review and 2 GTM                  | :white_check_mark: :red_circle: :negative_squared_cross_mark: :clock1:
| Micro Demo to the PO (Story only) | :white_check_mark: :red_circle: :negative_squared_cross_mark: :clock1:
| Migration script                  | :white_check_mark: :red_circle: :negative_squared_cross_mark: :clock1:
| Tech Doc                          | :white_check_mark: :red_circle: :negative_squared_cross_mark: :clock1:

:red_circle: Done but fail

:clock1: Done but pending

### DIFF
--- a/src/Pim/Component/Catalog/Builder/ProductBuilder.php
+++ b/src/Pim/Component/Catalog/Builder/ProductBuilder.php
@@ -6,6 +6,7 @@ use Pim\Component\Catalog\AttributeTypes;
 use Pim\Component\Catalog\Factory\ProductValueFactory;
 use Pim\Component\Catalog\Manager\AttributeValuesResolver;
 use Pim\Component\Catalog\Model\AttributeInterface;
+use Pim\Component\Catalog\Model\FamilyInterface;
 use Pim\Component\Catalog\Model\ProductInterface;
 use Pim\Component\Catalog\Model\ProductPriceInterface;
 use Pim\Component\Catalog\Model\ProductValueInterface;
@@ -107,6 +108,7 @@ class ProductBuilder implements ProductBuilderInterface
         if (null !== $familyCode) {
             $family = $this->familyRepository->findOneByIdentifier($familyCode);
             $product->setFamily($family);
+            $this->addBooleanToProduct($product);
         }
 
         $event = new GenericEvent($product);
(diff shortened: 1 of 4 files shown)


src/Pim/Component/Catalog/Builder/ProductBuilder.php
```diff
@@ -365,4 +367,34 @@ protected function addMissingPricesToProduct(ProductInterface $product)
...
+     *
+     * To remove when the UI will manage null values in boolean attributes.
+     *
+     * @param ProductInterface $product
+     */
+    protected function addBooleanToProduct(ProductInterface $product)
+    {
+        $family = $product->getFamily();
+
+        if (null !== $family) {
```
skeleton: add an early return to avoid triple indentation please

src/Pim/Component/Catalog/Builder/ProductBuilder.php
```diff
@@ -365,4 +367,34 @@ protected function addMissingPricesToProduct(ProductInterface $product)
             $this->addMissingPrices($value);
         }
     }
+
+    /**
+     * Set product values to "false" by default for every boolean attributes in the product's family.
```
nidup: you can do a reference to the improvement PIM-6056 that may be released for the v1.7, you can add a not in the related jira ticket to think to drop this temporary fix

src/Pim/Component/Catalog/Builder/ProductBuilder.php
```diff
@@ -365,4 +367,34 @@ protected function addMissingPricesToProduct(ProductInterface $product)
...
+     *
+     * This workaround is due to the UI that does not manage null values for boolean attributes, only false or true.
+     * It avoids to automatically submit boolean attributes belonging to the product's family in a proposal,
+     * even if those boolean attributes were not modified by the user.
+     *
+     * To remove when the UI will manage null values in boolean attributes.
```
ClementGautier: Add a `FIXME` keyword here. with that you have my GTM

PR #5546, opened 2017-01-20
Lower translation normalizer priority

jmleroux:


a2xchip:
Just curious what is for?

jmleroux:
Hello @a2xchip 
we had some problems with localizable entities originally created with the custom entity bundle.
After some investigations, we discovered this translation priority problem with the reference data. I make this PR to launch the CI and evaluate the impacts. If it goes well, i will describe the PR a little more ;)

aRn0D:
@jmleroux What is the status of this PR?

a2xchip:
is not rebased i think...

PR #5567, opened 2017-01-24
Simplify the Github pull request template

grena:

This PR aims to simplify the GitHub pull request template cartouche. Please leave comment for what you'd like :)


a2xchip:
Y/N/- ?

grena:
I like the explicit "Todo" instead of "No", because it shows that something is to do imho, if a PR is merged with "Todos" left, it should be weird. But it's a matter of tastes I think :)

a2xchip:
Sure, main purpose to not lose time on understanding statuses legend during PR formalization as i understand... how we name them is matter of tatste

PR #5569, opened 2017-01-24
API-18: Get a list of products

skeleton:
Get a list of products in the API.
- [x] be able to filter on channel
- [x] be able to filter on locales
- [x] be able to return only specific attributes
- [x] remove the identifier inside the product value (the identifier is only returned in the key "identifier")
- [x] add test on list (success)
- [x] add sanitizer for media & date
- [x] add test on failures

NOTE: apply PQB filters will be done in an other PR. Here we filter only PV with channel, locales & attributes

`OK`: Done / Validated
`-`: Not needed


src/Pim/Bundle/ApiBundle/Controller/ProductController.php
```diff
@@ -0,0 +1,220 @@
...
+     * @param ChannelRepositoryInterface          $channelRepository
+     * @param LocaleRepositoryInterface           $localeRepository
+     * @param AttributeRepositoryInterface        $attributeRepository
+     */
+    public function __construct(
+        ProductQueryBuilderFactoryInterface $pqbFactory,
+        NormalizerInterface $normalizer,
+        ChannelRepositoryInterface $channelRepository,
+        LocaleRepositoryInterface $localeRepository,
+        AttributeRepositoryInterface $attributeRepository
```
nidup: from these 3 injected repositories, we only use the method "findOneByIdentifier", I really wonder if we should typehint with IdentifiableObjectRepositoryInterface for each of them or if we should typehint with the business repository as you do :)

skeleton: yes, good question. I often typehint with the business repository to avoid BC break in the futur if we have to use other method from the repo, but dunno if it's a good reason

nidup: Yeah I do understand, i'd use the same argument ;) in other hand, typehint on the smaller contract makes a thiner dependance to the catalog code, makes easier to refactor / change the catalog part (less adherence = less impact), and push the dev who would later use another method of this repo in this class (can be you in six months ;) ) to double think about why he need to use another method, maybe he should depends on another small contract (that could be implemented by the same concret class and so inject it twice). Anyway both approaches are ok for me. I'd love to read thought of @jjanvier on this :)

jjanvier: @nidup @skeleton 

I guess it depends of what we try to achieve.

1. If we tried to develop a vertical feature, not core related, I would say that the best would be to create our own repositories interfaces. For instance, our own `MyLittleFeatureProductRepositoryInterface`, that has only the methods we need for our particular business. Nothing more. Internally, its implementation could compose, if needed, with the catalog repositories. No problem. That way of doing that forces us to have a real decoupled feature. Modification inside this feature would have no impact on the catalog. Modifications inside the catalog would have a really little impact (only a small part of the implementation) on this feature.
- Pros: no coupling between features, far easier to refactor the core features
- Cons: _seems_ more difficult to build, _seems_ more work to do

2. If the feature we release is closer to the core of the product, I guess it's acceptable to use the catalog repositories. But we really have to understand that the more the application grows, the more this way of doing is dangerous. It can become really messy in a mid term future. 
- Pros: faster to implement, less questions to think about
- Cons: short term vision, makes every single modification in the core feature more expensive

IMO, tt's always the same question. Where do you put the cursor?
And before building a new feature, we should really ask ourself, is it core oriented or not? Also we really need to understand and know where we want to go with the PIM, in a more global strategic way, functionally speaking. 

Maybe we should talk again about the ideal PIM. But, _first_, we should ask our dear PM (and POs) what's _her_ ideal PIM.

src/Pim/Bundle/ApiBundle/Controller/ProductController.php
```diff
@@ -0,0 +1,220 @@
...
+    }
+
+    /**
+     * Returns the channel and channel locales
+     *
+     * @param ChannelInterface $channel
+     *
+     * @return array
+     */
+    protected function getChannelOptions(ChannelInterface $channel)
```
jjanvier: getPqbOptions or something like these
these are not the options of the channel

src/Pim/Bundle/ApiBundle/Controller/ProductController.php
```diff
@@ -0,0 +1,220 @@
...
+
+        if ($request->query->has('channel')) {
+            $channel = $this->channelRepository->findOneByIdentifier($request->query->get('channel'));
+            if (null === $channel) {
+                throw new UnprocessableEntityHttpException(
+                    sprintf('Channel "%s" does not exist.', $request->query->get('channel'))
+                );
+            }
+
+            $options = $this->getChannelOptions($channel);
```
jjanvier: $pqbOptions

src/Pim/Bundle/ApiBundle/Controller/ProductController.php
```diff
@@ -0,0 +1,220 @@
...
+     * Set the PQB filters.
+     * If a channel is requested, add a filter to return only products linked to it category tree
+     *
+     * @param ProductQueryBuilderInterface $pqb
+     * @param array                        $pqbFilters
+     * @param ChannelInterface|null        $channel
+     */
+    protected function setPQBFilters(
+        ProductQueryBuilderInterface $pqb,
+        array $pqbFilters = [],
```
jjanvier: default arg is useless here
it's almost always useles in protected / private methods (except for objects sadly...) as you don't define a public API
default args are IMO only for public API convenience

src/Pim/Bundle/ApiBundle/Controller/ProductController.php
```diff
@@ -0,0 +1,220 @@
...
+        ChannelInterface $channel = null
+    ) {
+        if (null !== $channel) {
+            $pqbFilters['categories.code'] = [
```
jjanvier: categories

no need of `.code`, this is the default now (will be dropped in 1.8 btw)

src/Pim/Bundle/ApiBundle/Controller/ProductController.php
```diff
@@ -0,0 +1,220 @@
...
+        if (null !== $channel) {
+            if ($diff = array_diff($locales, $channel->getLocaleCodes())) {
+                $plural = sprintf(count($diff) > 1 ? 'Locales "%s" are' : 'Locale "%s" is', implode(', ', $diff));
+                throw new UnprocessableEntityHttpException(
+                    sprintf('%s not activated for the channel "%s".', $plural, $channel->getCode())
+                );
+            }
+        }
+
+        $options['locales'] = $locales;
```
jjanvier: imo, the code would be easier to understand if you'd return only the locale codes, and this option array
and could name your function `checkInputLocaleCodes` (or something much better)

src/Pim/Bundle/ApiBundle/Controller/ProductController.php
```diff
@@ -0,0 +1,220 @@
...
+                $errors[] = $attributeCode;
+            }
+        }
+
+        if (!empty($errors)) {
+            $plural = count($errors) > 1 ? 'Attributes "%s" do not exist.' : 'Attribute "%s" does not exist.';
+            throw new UnprocessableEntityHttpException(sprintf($plural, implode(', ', $errors)));
+        }
+
+        $options['attributes'] = $attributeCodes;
```
jjanvier: exactly the same comment as the previous one

src/Pim/Bundle/ApiBundle/Resources/config/normalizers.yml
```diff
@@ -27,7 +32,43 @@ services:
...
-        class: '%pim_api.normalizer.simple_collection.class%'
+    pim_api.normalizer.collection:
+        class: '%pim_api.normalizer.collection.class%'
+        tags:
+            - { name: pim_serializer.normalizer, priority: 90 }
+
+    pim_api.normalizer.product:
+        class: '%pim_api.normalizer.product.class%'
+        arguments:
+            - '@pim_catalog.normalizer.standard.product'
```
jjanvier: that won't work I think :p 
we can check IRL, it's a bit tricky

tests/DateSanitizer.php
```diff
@@ -0,0 +1,36 @@
...
+ *
+ * In integration tests, when creating a product for instance, we cannot guess created/updated dates.
+ * We can just check if the data match with the constant DATE_FIELD_PATTERN.
+ * If the pattern is checked, we return the constant DATE_FIELD_COMPARISON to be able to have an identical comparison element
+ *
+ * @author    Marie Bochu <marie.bochu@akeneo.com>
+ * @copyright 2017 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+class DateSanitizer
```
jjanvier: :+1: top

src/Pim/Bundle/ApiBundle/Controller/ProductController.php
skeleton: these options are not only for the PQB, it's also to filters product values during the normalizations.

src/Pim/Bundle/ApiBundle/Resources/config/normalizers.yml
skeleton: seen IRL, it works :)

PR #5571, opened 2017-01-24
PIM-5987: Fix import duplicated products from the same batch 

SamirBoulil:

When importing duplicated lines on a product import that live within the same batch size, the PIM creates a duplicated products for the duplicated row.

This fix keeps tracks of the processed product skus living in the same batch size and prevent duplicated rows from being imported twice.

A message is sent to the user that a duplicated line has been found and that it won't be imported.

In the next version (maybe 1.8) a new strategy will be implemented in order to have a consistent mecanism regarding those duplicated See ticket.
The POs are aware this solution is not optimum but they'd rather prevent the duplication creation and having different behaviors on import than create duplicated products.

`OK`: Done / Validated
`-`: Not needed


SamirBoulil:
After multiple discussions with the PO, we will implement a strategy to handle duplicated products homogeneously in the release after 1.7.

PR #5587, opened 2017-01-27
Split integration TestCase into smaller classes

phaseinducer:
Splitting the TestCase into smaller SRP classes will allow creating other TestCases without falling into inheritance hell.
For example we need a new ApiTestCase that will inherit from WebTestCase (so not from our main TestCase) but we still need the whole fixtures / purge logic.

### DIFF
--- a/tests/ConnectionCloser.php
+++ b/tests/ConnectionCloser.php
@@ -1,0 +1,37 @@
+<?php
+
+namespace Akeneo\Test\Integration;
+
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * @author    Yohan Blain <yohan.blain@akeneo.com>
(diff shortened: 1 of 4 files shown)


tests/ConnectionCloser.php
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    /**
+     * Ensures that all used connections are well closed.
+     *
+     * @see https://github.com/akeneo/pim-community-dev/pull/5484
+     */
+    public function closeConnections()
+    {
+        if (AkeneoStorageUtilsExtension::DOCTRINE_MONGODB_ODM === $this->container->getParameter('pim_catalog_product_storage_driver')) {
```
skeleton: I just tested, there is error "Too many connections" with mongo. Remove this condition and always call `doctrine` service will fix the error

PR #5596, opened 2017-01-30
Remove unused redirectAfterEdit()

Hocdoc:
While migrating to V. 1.6 I've noticed an error with the undefined `$result` in the `ProductController`: `switch ($request->get('action')) {` 

It seems, that this method isn't used anymore, so it can be removed.


pierallard:
Hi @Hocdoc !
Nice catch !
I will run a build on our continuous integration to see if there is no bad side effects ;)

pierallard:
Thanks @Hocdoc !
One of the fastest merge in the PIM story :+1:

a2xchip:
:tada:

PR #5598, opened 2017-01-30
Api 35 : content type header is mandatory in the API, allowing automatic JSON validation in FOS listener 

ahocquard:

Originally, the Content-type header was not mandatory in the POST/PUT/PATCH request. And if it was specified, it should be "application/json".

A lot of clients use a Content-Type by default. For example, curl use "application/x-www-form-urlencoded" if you don't specify any Content-Type header.

Therefore, it's pretty useless to allow users to not specify anything, because the client do (or rather force) that for them...
We've chosen to force the user to specify the Content-Type header.

Moreover, a FOSRest listener checks that the format is JSON when the Content-Type is "application/JSON" with a specific error message in case it's not.

https://github.com/FriendsOfSymfony/FOSRestBundle/blob/master/EventListener/BodyListener.php#L115

We've chosen to keep the message thrown by this validation (and it's better to check the format in a listener).

Unfortunately, we will keep the method getDecodedContent in the controllers, because this listener does not throw any exception if the content is empty (which is not our case : it should throw an exception).

https://github.com/FriendsOfSymfony/FOSRestBundle/blob/master/EventListener/BodyListener.php#L108

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Pim/Bundle/ApiBundle/EventSubscriber/CheckHeadersRequestSubscriber.php
+++ b/src/Pim/Bundle/ApiBundle/EventSubscriber/CheckHeadersRequestSubscriber.php
@@ -8,6 +8,7 @@ use FOS\RestBundle\Util\StopFormatListenerException;
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
 use Symfony\Component\HttpKernel\Event\GetResponseEvent;
 use Symfony\Component\HttpKernel\Exception\NotAcceptableHttpException;
+use Symfony\Component\HttpKernel\Exception\UnsupportedMediaTypeHttpException;
 use Symfony\Component\HttpKernel\HttpKernelInterface;
 use Symfony\Component\HttpKernel\KernelEvents;
 
@@ -78,10 +79,17 @@ class CheckHeadersRequestSubscriber implements EventSubscriberInterface
 
             if (in_array($request->getMethod(), ['PUT', 'PATCH', 'POST'])) {
                 $contentType = $request->headers->get('content-type');
-                if (null !== $contentType &&
-                    !in_array($contentType, ['application/x-www-form-urlencoded', $best->getValue()])
-                ) {
-                    throw new NotAcceptableHttpException(
+                if (null === $contentType) {
+                    throw new UnsupportedMediaTypeHttpException(
+                        sprintf(
+                            'The "Content-Type" header is missing. "%s" has to be specified as value.',
+                            $best->getValue()
+                        )
+                    );
+                }
+
+                if ($contentType !== $best->getValue()) {
+                    throw new UnsupportedMediaTypeHttpException(
                         sprintf(
                             '"%s" in "Content-Type" header is not valid. Only "%s" is allowed.',
                             $contentType,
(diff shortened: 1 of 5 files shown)


src/Pim/Bundle/ApiBundle/EventSubscriber/CheckHeadersRequestSubscriber.php
```diff
@@ -78,10 +79,17 @@ public function onKernelRequest(GetResponseEvent $event)
...
             if (in_array($request->getMethod(), ['PUT', 'PATCH', 'POST'])) {
                 $contentType = $request->headers->get('content-type');
-                if (null !== $contentType &&
-                    !in_array($contentType, ['application/x-www-form-urlencoded', $best->getValue()])
-                ) {
-                    throw new NotAcceptableHttpException(
+                if (null === $contentType) {
+                    throw new UnsupportedMediaTypeHttpException(
+                        sprintf(
+                            'The "Content-Type" header is missing. "%s" has to be specified.',
```
phaseinducer: "... as value" ?

src/Pim/Bundle/ApiBundle/tests/integration/Controller/Rest/Category/CreateCategoryIntegration.php
```diff
@@ -19,7 +19,7 @@ public function testHttpHeadersInResponseWhenACategoryIsCreated()
     }
 JSON;
 
-        $client->request('POST', 'api/rest/v1/categories', [], [], [], $data);
+        $client->request('POST', 'api/rest/v1/categories', [], [], ['CONTENT_TYPE' => 'application/json'], $data);
```
phaseinducer: Maybe you can use the TestCase introduced in https://github.com/akeneo/pim-community-dev/pull/5601 to centralize this.

ahocquard: Yes, I will clean that in another PR

ahocquard:
@jjanvier The attribute zone flag is set when you are in route flagged as a rest route : 
https://github.com/FriendsOfSymfony/FOSRestBundle/blob/master/EventListener/ZoneMatcherListener.php#L48

The other FOSRest listeners will then check this flag to apply their action or not.
Unfortunately, we don't have the zone set in a variable.

So, we are forced to use this try catch.

src/Pim/Bundle/ApiBundle/tests/integration/Controller/Rest/Category/CreateCategoryIntegration.php
```diff
@@ -88,15 +88,32 @@ public function testCompleteCategoryCreation()
         $this->assertSame($categoryStandard, $normalizer->normalize($category));
     }
 
+    public function testResponseWhenContentIsEmpty()
+    {
+        $client = $this->createAuthentifiedClient();
+
+        $data = '{';
```
momoss: With that, you just test that Json is not valid, if you want to test an empty content, you have to test:
 $data = '{}';

ahocquard: Actually, it's ''. It a bad copy/past I think, ty.

PR #5599, opened 2017-01-30
AMS-156: force fetching of view on columns update

grena:

**The bug:**
- Create a custom view
- Click on Columns to customise the grid
- Remove/Add columns, click on Save

**The fix:**
The problem comes from the fetchers, the promises return the View by reference, so when we update the columns, the View from the fetcher has its columns modified too, but it should not. This fix bypass the cache of the fetcher by forcing a re-fetch.

---
_Note:_
- The issue #5600 has been created to answer this problem in the whole PIM.

### DIFF
--- a/src/Pim/Bundle/EnrichBundle/Resources/public/js/grid/view-selector.js
+++ b/src/Pim/Bundle/EnrichBundle/Resources/public/js/grid/view-selector.js
@@ -236,7 +236,7 @@ define(
                         deferred.resolve(this.getDefaultView());
                     } else {
                         FetcherRegistry.getFetcher('datagrid-view')
-                            .fetch(activeViewId, {alias: this.gridAlias})
+                            .fetch(activeViewId, {alias: this.gridAlias, cached: false})
                             .then(this.postFetchDatagridView.bind(this))
                             .then(function (view) {
                                 deferred.resolve(view);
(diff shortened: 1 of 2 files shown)


src/Pim/Bundle/EnrichBundle/Resources/public/js/grid/view-selector.js
```diff
@@ -236,7 +236,7 @@ define(
                         deferred.resolve(this.getDefaultView());
                     } else {
                         FetcherRegistry.getFetcher('datagrid-view')
-                            .fetch(activeViewId, {alias: this.gridAlias})
+                            .fetch(activeViewId, {alias: this.gridAlias, cached: false})
```
a2xchip: Hah! View selector :+1: Nice catch @grena Consider my [question](https://github.com/akeneo/pim-community-dev/issues/5600#issuecomment-276324421) about example closed... Another question is if there are more same issues.

grena: @a2xchip absolutely! That's why I raised the issue, it's to keep an eye on this behavior :) Normally we don't, as you explained on your comment on the issue, we use `getFormData` etc

mickaelandrieu:
@grena do we have to close the related issue or not?

PR #5601, opened 2017-01-30
API-50: Don't bypass the firewall in integration tests

phaseinducer:
Until now the security firewall was disabled in test environment. This PR re-enables it to allow the tests to use the real OAuth workflow.

These steps are now executed for each test case:
1. create a client via the dedicated command
2. call `/api/oauth/v1/token` to get an access token

Then the token is included in every requests.
To do that we introduce a new TestCase for the API.

The good news: the integration tests are closer to the real application behavior.
The bad news: it is approximately 15% slower.

### DIFF
--- a/src/Pim/Bundle/ApiBundle/tests/integration/ApiTestCase.php
+++ b/src/Pim/Bundle/ApiBundle/tests/integration/ApiTestCase.php
@@ -1,0 +1,175 @@
+<?php
+
+namespace Pim\Bundle\ApiBundle\tests\integration;
+
+use Akeneo\Test\Integration\Configuration;
+use Akeneo\Test\Integration\ConnectionCloser;
+use Akeneo\Test\Integration\DatabasePurger;
+use Akeneo\Test\Integration\FixturesLoader;
+use Symfony\Bundle\FrameworkBundle\Client;
+use Symfony\Bundle\FrameworkBundle\Console\Application;
+use Symfony\Bundle\FrameworkBundle\Test\WebTestCase;
+use Symfony\Component\Console\Input\ArrayInput;
+use Symfony\Component\Console\Output\BufferedOutput;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * Test case dedicated to PIM API interaction including authentication handling.
+ *
+ * @author    Yohan Blain <yohan.blain@akeneo.com>
+ * @copyright 2017 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php Open Software License (OSL 3.0)
+ */
+abstract class ApiTestCase extends WebTestCase
+{
+    /** @var int Count of executed tests inside the same test class */
+    protected static $count = 0;
+
+    /** @var ContainerInterface */
+    protected $container;
+
+    /** @var string */
+    protected $accessToken;
+
+    /**
+     * {@inheritdoc}
+     */
+    public static function setUpBeforeClass()
+    {
+        self::$count = 0;
+    }
+
+    /**
+     * @return Configuration
+     */
+    abstract protected function getConfiguration();
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function setUp()
+    {
+        static::bootKernel();
+
+        $this->container = static::$kernel->getContainer();
+        $configuration = $this->getConfiguration();
+
+        self::$count++;
+
+        if ($configuration->isDatabasePurgedForEachTest() || 1 === self::$count) {
+            $databasePurger = $this->getDatabasePurger();
+            $databasePurger->purge();
+
+            $fixturesLoader = $this->getFixturesLoader($configuration);
+            $fixturesLoader->load();
+        }
+    }
+
+    /**
+     * Adds a valid access token to the client, so it is included in all its requests.
+     *
+     * @param array $options
+     * @param array $server
+     *
+     * @throws \Exception
+     *
+     * @return Client
+     */
+    protected function createAuthentifiedClient(array $options = [], array $server = [])
+    {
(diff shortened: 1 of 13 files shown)


src/Pim/Bundle/ApiBundle/tests/integration/ApiTestCase.php
```diff
@@ -0,0 +1,163 @@
...
+/**
+ * Test case dedicated to PIM API interaction including authentication handling.
+ *
+ * @author    Yohan Blain <yohan.blain@akeneo.com>
+ * @copyright 2017 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php Open Software License (OSL 3.0)
+ */
+abstract class ApiTestCase extends WebTestCase
+{
+    /** @var int Count of test inside the same test class */
```
ahocquard: "Count of executed tests"

src/Pim/Bundle/ApiBundle/tests/integration/ApiTestCase.php
```diff
@@ -0,0 +1,163 @@
...
+                'PHP_AUTH_USER' => $clientId,
+                'PHP_AUTH_PW'   => $secret,
+            ]
+        );
+
+        $response = $oauthClient->getResponse();
+        $responseBody = json_decode($response->getContent(), true);
+
+        $client->setServerParameter('HTTP_AUTHORIZATION', 'Bearer '.$responseBody['access_token']);
+    }
```
ahocquard: With this way, you have to authenticate at each call. A normal client will execute the authentification one time, and will register the access token for future use. Wouldn't be better to cache the token ?
For example : 
-  in the setUp, authenticate at the first call and keep the access token (or 
- call this method authentifyClient each time but to fill the HTTP_AUTHORIZATION header, which is in cache

src/Pim/Bundle/ApiBundle/tests/integration/Controller/Rest/Attribute/GetAttributeIntegration.php
```diff
@@ -3,14 +3,15 @@
...
+use Pim\Bundle\ApiBundle\tests\integration\ApiTestCase;
 use Symfony\Component\HttpFoundation\Response;
 
-class GetAttributeIntegration extends TestCase
+class GetAttributeIntegration extends ApiTestCase
 {
     public function testGetAnAttribute()
     {
         $client = static::createClient();
+        $this->authentifyClient($client);
```
ahocquard: a funcion $this->createAuthentifyClient could be a better option

src/Pim/Bundle/ApiBundle/tests/integration/ApiTestCase.php
```diff
@@ -0,0 +1,163 @@
...
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * Test case dedicated to PIM API interaction including authentication handling.
+ *
+ * @author    Yohan Blain <yohan.blain@akeneo.com>
+ * @copyright 2017 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php Open Software License (OSL 3.0)
+ */
+abstract class ApiTestCase extends WebTestCase
```
ahocquard: It's a copy/Past of TestCase, except for the authentification function. On the other part, it's pitty that TestCase extends WebTestCase. I understand what you did, but don't you think keeping the TestCase extending WebTestCase better, and then make ApiTestCase extending TestCase ?

tests/TestCase.php
```diff
@@ -11,7 +11,7 @@
  * @copyright 2016 Akeneo SAS (http://www.akeneo.com)
  * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
  */
-abstract class TestCase extends WebTestCase
+abstract class TestCase extends KernelTestCase
 {
     /** @var int Count of test inside the same test class */
```
ahocquard: To be consistant: "Count of executed tests" (at least the plurial for test)

src/Pim/Bundle/ApiBundle/tests/integration/ApiTestCase.php
skeleton: It's not normal, expect when you really need a client like us, to extend WebTestCase in an integration test. It's better to have multiple TestCase for each use case

phaseinducer: TestCase now extends KernelTestCase, like on master.
There is no other part of the application where we will need a web client for integration tests.
It would be a shame to add another level of inheritance (it's already the 5th) just to avoid copying 100 lines of code with almost no logic. All the logic has been extracted to small classes in https://github.com/akeneo/pim-community-dev/pull/5587 precisely to be able to do that. We want to avoid the state in which our behat stack is currently.

src/Pim/Bundle/ApiBundle/tests/integration/ApiTestCase.php
```diff
@@ -0,0 +1,175 @@
...
+{
+    /** @var int Count of executed test inside the same test class */
```
ahocquard: "tests"

tests/TestCase.php
```diff
@@ -11,9 +11,9 @@
  * @copyright 2016 Akeneo SAS (http://www.akeneo.com)
  * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
  */
-abstract class TestCase extends WebTestCase
+abstract class TestCase extends KernelTestCase
 {
-    /** @var int Count of test inside the same test class */
+    /** @var int Count of executed test inside the same test class */
```
ahocquard: "tests"

PR #5612, opened 2017-02-01
AMS-157: correctly switch grid to the user default view

grena:

Correctly switches the grid to the user default one if he has some.

### DIFF
--- a/features/datagrid/datagrid_views.feature
+++ b/features/datagrid/datagrid_views.feature
@@ -131,6 +131,9 @@ Feature: Datagrid views
       | Default product grid view | Sneakers only |
     And I press the "Save" button
     Then I should not see the text "There are unsaved changes."
+    When I am on the products page
+    Then I should see the text "Sneakers only"
+    And the grid should contain 2 elements
     When I logout
     And I am logged in as "Mary"
     And I am on the products page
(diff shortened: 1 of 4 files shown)


features/datagrid/datagrid_views.feature
```diff
@@ -131,6 +131,10 @@ Feature: Datagrid views
       | Default product grid view | Sneakers only |
     And I press the "Save" button
     Then I should not see the text "There are unsaved changes."
+    When I am on the products page
+    Then I should see the text "Sneakers only"
+    And I should see products purple-sneakers and black-sneakers
+    But I should not see product black-boots
```
pierallard: If you have only 2 products, Prefer count the number of items in the grid, it's fastest method (here you have to wait 1 min).

src/Pim/Bundle/EnrichBundle/Resources/public/js/grid/view-selector.js
```diff
@@ -228,26 +220,26 @@ define(
...
                 var activeViewId = DatagridState.get(this.gridAlias, 'view');
-                var userDefaultView = this.defaultUserView;
                 var deferred = $.Deferred();
 
-                if (activeViewId) {
-                    if ('0' === activeViewId) {
-                        deferred.resolve(this.getDefaultView());
-                    } else {
+                this.getUserDefaultView().then(function (userDefaultView) {
+                    if (userDefaultView && (!activeViewId || '0' === activeViewId)) {
```
pierallard: should be  a stupid remark, but instead of everywhere have the "0 === fooId", you should have a method "is_it_the_default_view(fooId)", it will be 100 more explicit.

grena: Not fan of a method for this, but I could create a var, would be more explicit :)

PR #5618, opened 2017-02-01
API-82: Refreshing a token

phaseinducer:
Same route as for getting a token, except we don't need username and password, just the refresh token, and the grant_type must be "refresh_token" instead of "password".
Some error messages are different.
I also did some needed refactorings in the TestCase.

`OK`: Done / Validated
`-`: Not needed


src/Pim/Bundle/ApiBundle/tests/integration/ApiTestCase.php
```diff
@@ -22,21 +22,29 @@
...
+    /** @var ContainerInterface */
+    protected $container;
 
     /**
      * {@inheritdoc}
      */
     public static function setUpBeforeClass()
     {
-        self::$count = 0;
+        static::$count = 0;
```
skeleton: why do you change to static ? this variable should not be changed on the other classes

src/Pim/Bundle/ApiBundle/tests/integration/Controller/Token/RefreshTokenIntegration.php
```diff
@@ -0,0 +1,100 @@
...
+        $response = $client->getResponse();
+        $responseBody = json_decode($response->getContent(), true);
+
+        $this->assertSame(Response::HTTP_OK, $response->getStatusCode());
+        $this->assertArrayHasKey('access_token', $responseBody);
+        $this->assertArrayHasKey('expires_in', $responseBody);
+        $this->assertArrayHasKey('token_type', $responseBody);
+        $this->assertArrayHasKey('scope', $responseBody);
+        $this->assertArrayHasKey('refresh_token', $responseBody);
+    }
```
ahocquard: Maybe we could add a test to check that you can't use two times in a row a refresh token. On the other hand, it's like if we test the internal behaviour of the bundle (which is already tested).

PR #5623, opened 2017-02-02
AMS-164: detach old view selectors in order to avoid multiple subscribers

grena:

- The problem: When switching from grid to other page multiple times, then going on a project and selecting another category, there is multiple times the message "You're leaving project scope."
- Why: We instanciate the View Selector in [_views.html.twig](https://github.com/akeneo/pim-community-dev/blob/master/src/Pim/Bundle/DataGridBundle/Resources/views/Datagrid/_views.html.twig), so, each time we go on the grid, we instanciate a new View Selector, and each instance listens to some events.
- Solution: Unbind view selector subscribers on route change

| Added Behats                      | Not possible


phaseinducer:
Wow, hardcore solution. What about something like this? https://github.com/akeneo/pim-community-dev/blob/master/src/Pim/Bundle/EnrichBundle/Resources/public/js/form/common/state.js#L65

PR #5624, opened 2017-02-02
Fix the attribute fixtures and behats due to property validation in attribute updater

ahocquard:

Since the API, we can't pass unknown properties in the updaters. In this case, it throws an exception.

A lot of behats use legacy properties or a a syntax that is not accepted anymore.
For example "metric family" instead of "metric_family".

This PR fix them.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/features/import/product/import_products_with_invalid_data.feature
+++ b/features/import/product/import_products_with_invalid_data.feature
@@ -199,7 +199,7 @@ Feature: Execute a job
   @jira https://akeneo.atlassian.net/browse/PIM-3266
   Scenario: Skip new products with non-existing media attributes during an import
     Given the following attributes:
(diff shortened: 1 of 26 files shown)


features/import/product/import_products_with_invalid_data.feature
```diff
@@ -230,7 +230,7 @@ Feature: Execute a job
       | bic-core-148        | sneakers | Bic Core 148        | 2014_collection |
       | fanatic-freewave-76 | sneakers | Fanatic Freewave 76 | 2014_collection |
     And the following attributes:
-      | label       | type  | allowed extensions |
+      | label       | type  | allowedextensions |
```
momoss: allowed_extensions

PR #5632, opened 2017-02-03
Api 106 : check locales key for the creation/updates of the labels in the API

ahocquard:

This PR resolve : 
- a fatal exception when you specified a locale which is empty (dedicated commit).
I've fixed that in every labelizable entities because it's always the same logic to get the translation, and so, always the same error.

For exampbe, before the fix, this request raised an error 500 : 
```
{
	"code": "category",
	"parent": null,
	"labels": {
		"": 12
	}
}
```

-It checks that the locale associated to the label exists when you create an entity
 For the moment, it's checked only in entities handled by the API : family, category and attributes. It does not check if the locale is activated because it implies other stuff out of the scope of the API.

For example, before the fix, this request was well processed : 
```
{
	"code": "category",
	"parent": null,
	"labels": {
		"jean_michel": 12
	}
}
```

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Pim/Bundle/CatalogBundle/Entity/AssociationType.php
+++ b/src/Pim/Bundle/CatalogBundle/Entity/AssociationType.php
@@ -142,7 +142,7 @@ class AssociationType implements AssociationTypeInterface
     public function getTranslation($locale = null)
     {
         $locale = ($locale) ? $locale : $this->locale;
-        if (!$locale) {
+        if (null === $locale) {
             return null;
         }
         foreach ($this->getTranslations() as $translation) {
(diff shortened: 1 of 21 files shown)


src/Pim/Bundle/CatalogBundle/Entity/AssociationType.php
```diff
@@ -142,7 +142,7 @@ public function getTranslations()
     public function getTranslation($locale = null)
     {
         $locale = ($locale) ? $locale : $this->locale;
```
momoss: Maybe Better with : 
(null !== $locale) ? $locale : $this->locale;

ahocquard: actually, your comment does not change my fix, but highlight another potential error 500 ! I will fix that.

PR #5645, opened 2017-02-06
Tip 322 : add business exceptions in adders, setters, removers and copiers

ahocquard:

In the API, we need the property name of the exception to be able to create specific API exceptions.

In order to do that, a reworking of the exception has been done for the updaters to throw business exception.
This is the same PR  for the adders, setters, removers, an copiers.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Pim/Component/Catalog/Updater/Adder/AssociationFieldAdder.php
+++ b/src/Pim/Component/Catalog/Updater/Adder/AssociationFieldAdder.php
@@ -2,9 +2,10 @@
 
 namespace Pim\Component\Catalog\Updater\Adder;
 
+use Akeneo\Component\StorageUtils\Exception\InvalidPropertyException;
+use Akeneo\Component\StorageUtils\Exception\InvalidPropertyTypeException;
 use Akeneo\Component\StorageUtils\Repository\IdentifiableObjectRepositoryInterface;
 use Pim\Component\Catalog\Builder\ProductBuilderInterface;
-use Pim\Component\Catalog\Exception\InvalidArgumentException;
 use Pim\Component\Catalog\Model\AssociationInterface;
 use Pim\Component\Catalog\Model\ProductInterface;
 
@@ -81,15 +82,18 @@ class AssociationFieldAdder extends AbstractFieldAdder
      *
      * @param ProductInterface $product
      * @param mixed            $data
+     *
+     * @throws InvalidPropertyException
      */
     protected function addProductsAndGroupsToAssociations(ProductInterface $product, $data)
     {
         foreach ($data as $typeCode => $items) {
             $association = $product->getAssociationForTypeCode($typeCode);
             if (null === $association) {
-                throw InvalidArgumentException::expected(
+                throw InvalidPropertyException::validEntityCodeExpected(
                     'associations',
-                    'existing association type code',
+                    'association type code',
+                    'The association type does not exist',
                     static::class,
                     $typeCode
                 );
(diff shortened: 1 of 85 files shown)


src/Pim/Component/Catalog/Updater/Setter/DateAttributeSetter.php
```diff
@@ -66,9 +66,9 @@ protected function formatData(AttributeInterface $attribute, $data)
         } elseif (is_string($data)) {
             $this->validateDateFormat($attribute, $data);
         } elseif (null !== $data && !is_string($data)) {
-            throw InvalidArgumentException::expected(
+            throw InvalidPropertyException::dateExpected(
```
skeleton: in fact, message for dateExpected method is not really correct. It says `Property "%s" expects a string with the format "%s" as data, "%s" given`. It's not really true, we accept DateTime too (but maybe not everywhere)

src/Pim/Component/Catalog/Updater/Adder/AssociationFieldAdder.php
```diff
@@ -168,30 +178,41 @@ protected function checkData($field, $data)
...
      * @param mixed  $items
      *
-     * @throws InvalidArgumentException
+     * @throws InvalidPropertyTypeException
      */
     protected function checkAssociationData($field, array $data, $assocTypeCode, $items)
     {
         if (!is_array($items) || !is_string($assocTypeCode) || !isset($items['products']) || !isset($items['groups'])) {
-            throw InvalidArgumentException::associationFormatExpected($field, $data);
+            throw InvalidPropertyTypeException::validArrayStructureExpected(
```
ahocquard: The structure should be indicated in the standard format, not in the exception itself imho. Secondly, it's bind to a PHP structure. We should expose errors that will be be language agnostic. And now, more than before with the API. That's why I deleted this message. And I have deleted the "print_r" as well in some other exceptions.

anMore over, coming from an import it's so technical

src/Pim/Component/Catalog/Updater/Adder/MultiSelectAttributeAdder.php
```diff
@@ -135,7 +138,7 @@ protected function addOptions(
      */
     protected function getOption(AttributeInterface $attribute, $optionCode)
     {
-        $identifier = $attribute->getCode() . '.' . $optionCode;
+        $identifier = $attribute->getCode().'.'.$optionCode;
```
ahocquard: nice catch, it's due to a misconfiguration of my php cs settings ;)

src/Pim/Component/Catalog/Updater/Adder/PriceCollectionAttributeAdder.php
```diff
@@ -73,58 +74,64 @@ public function addAttributeData(
...
                     'currency',
                     static::class,
-                    print_r($data, true)
+                    $data
                 );
             }
 
             if (!is_numeric($price['amount']) && null !== $price['amount']) {
-                throw InvalidArgumentException::arrayNumericKeyExpected(
+                throw new InvalidPropertyTypeException(
```
ahocquard: 'Property "attribute_code" expects a numeric as data, "EU" given.' 
is not enough explicit, you don't know that it is the currency key which is wrong.

I prefer : 
'Property "attribute_code" expects a numeric as data for the currency, "EU" given.'

src/Pim/Component/Catalog/Updater/Setter/DateAttributeSetter.php
ahocquard: You're right, but if I add Datetime, it means that you will have this error in the API and import (and they can't pass a Datetime !). It's very internal because it's a PHP object... Personally, I think we should promote the API/import messages.

I think it's a big topic !

jjanvier: in the next implementation, all product values will be created from the standard format (not from a Date object for a date product value for instance). So this message seems correct to me.

PR #5656, opened 2017-02-08
AMS-175: reset button keeps view filters & doesn't reset scope

grena:

This PR solves 2 problems about the "Reset" button on the grid:
- It sets the scope to "All", which has no sense
- When you're on a view, it resets filters, but not to the initial state of the view

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/features/Context/AssertionContext.php
+++ b/features/Context/AssertionContext.php
@@ -42,7 +42,7 @@ class AssertionContext extends RawMinkContext
     /**
      * Checks, that page does not contain specified text.
      *
-     * @Then /^I should not see the text "([^"]*)"$/
+     * @Then /^I should not see the text "(?P<text>(?:[^"]|\\")*)"$/
      */
     public function assertPageNotContainsText($text)
     {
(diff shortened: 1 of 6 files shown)


features/Context/AssertionContext.php
```diff
@@ -42,7 +42,7 @@ public function assertPageContainsText($text)
     /**
      * Checks, that page does not contain specified text.
      *
-     * @Then /^I should not see the text "([^"]*)"$/
+     * @Then /^I should not see the text "(?P<text>(?:[^"]|\\")*)"$/
```
grena: This is to be consistent with the matching done in "Then I should see the text", the method just above.
This way, we can match things with double quotes.

src/Pim/Bundle/DataGridBundle/Resources/views/macros.html.twig
```diff
@@ -77,9 +79,19 @@
                         };
 
                         var applyFilters = function (rawFilters) {
-                            var collection = new PageableCollection();
-                            var filters    = collection.decodeStateData(rawFilters);
+                            var fakeCollection = new PageableCollection();
+                            var filters = fakeCollection.decodeStateData(rawFilters);
```
phaseinducer: Maybe it's possible to call the prototype here instead of creating a useless instance?

grena: Nice catch, I'll give a try :)

PR #5661, opened 2017-02-09
[Merge 1.5->1.6]

pierallard:


upgrades/migrations/MigrateMongoDbAssociations.php
```diff
@@ -0,0 +1,168 @@
+<?php
```
jjanvier: if this script was for 1.5 (and now for 1.6 and master), should it be in this folder?

pierallard: We decided during the review this was the better place for this script.
Warning, this script is only for 1.5 and 1.6 versions, and will be deleted on master because not needed anymore.

PR #5673, opened 2017-02-10
[TIP-702] Refactor behat entity creations

pierallard:
# What do this PR ?

## tl;dr;

It fixes entity behat creation to be as close as possible as our flat format and prevent any magic usage of entity creation + entity database check.
Don't be afraid of files number nor updated lines, 95% is only behats tables.

## Moar details

This PR:
- Uses the dedicated services (convertor / processor / writer) to convert, process and write entities in database
- Only do the refacto job for attributes, attribute groups, families, channels, product groups, variant groups, association type, group types
- Update "entity checkers" ("there should be the following ...") to raise an error when a column is unchecked, it's not the case on master.
- Update "entity checkers" to allow arrays checks regardless the order (`[a,b] === [b,a]`)
- Split "the following product groups" to "the following product groups" + "the following variant groups" because it's not the same entity services and same columns
- Because every of this changes has been done with a script, it also align tables in the behat.

## Want to review ?

You can easily only read the FixtureContext.php and forgot the .features IMHO.


features/Context/FixturesContext.php
```diff
@@ -217,8 +225,20 @@ public function theFollowingAttributeGroups(TableNode $table)
...
+        $converter = $this->getContainer()->get('pim_connector.array_converter.flat_to_standard.attribute');
+        $processor = $this->getContainer()->get('pim_connector.processor.denormalization.attribute');
+        $saver     = $this->getContainer()->get('pim_catalog.saver.attribute');
+
         foreach ($table->getHash() as $data) {
-            $this->createAttribute($data);
+            $attribute = $processor->process($converter->convert($data));
+
+            if (isset($data['unique'])) {
+                // TODO Due to Pim/Component/Catalog/Updater/AttributeUpdater.php:226 (SDS-998)
```
grena: what is todo?

features/Context/FixturesContext.php
```diff
@@ -178,8 +175,12 @@ public function anEnabledOrDisabledProduct($status, $sku)
      */
     public function theFollowingFamilies(TableNode $table)
     {
+        $converter = $this->getContainer()->get('pim_connector.array_converter.flat_to_standard.family');
+        $processor = $this->getContainer()->get('pim_connector.processor.denormalization.family');
+        $saver     = $this->getContainer()->get('pim_catalog.saver.family');
+
         foreach ($table->getHash() as $data) {
-            $this->createFamily($data);
+            $saver->save($processor->process($converter->convert($data)));
```
grena: no validation needed?

features/Context/FixturesContext.php
```diff
@@ -413,27 +433,39 @@ public function createSelectAttributesWithOptions($attributeCount, $filterable,
...
+                    $requirements = [];
+                    foreach ($family->getAttributeRequirements() as $requirement) {
+                        if ($matches['channel'] === $requirement->getChannel()->getCode()) {
+                            $requirements[] = $requirement->getAttribute()->getCode();
+                        }
+                    }
+                    $this->assertArrayEquals(explode(',', $value), $requirements);
+
+                } else {
+                    throw new \InvalidArgumentException(sprintf('Can not check "%s" attribute of the family', $key));
```
grena: Cannot

features/Context/FixturesContext.php
```diff
@@ -475,42 +507,36 @@ public function thereShouldBeTheFollowingChannels(TableNode $table)
...
-            assertEquals($data['currencies'], implode(',', $currencyCodes));
-
-            if ('' !== $data['conversion_units']) {
-                $units = explode(',', $data['conversion_units']);
-                $formattedUnits = [];
-                foreach ($units as $unit) {
-                    list($key, $value) = explode(':', trim($unit));
-                    $formattedUnits[trim($key)] = trim($value);
+                } else {
+                    throw new \InvalidArgumentException(sprintf('Can not check "%s" attribute of the channel', $key));
```
grena: Cannot

features/Context/FixturesContext.php
```diff
@@ -2251,4 +2042,17 @@ protected function getDocumentManager()
...
+     * Assert than 2 arrays are equal, regardless the order of the elements.
+     *
+     * @param $array1
+     * @param $array2
+     */
+    protected function assertArrayEquals($array1, $array2)
+    {
+        sort($array1);
+        sort($array2);
+        assertEquals(join(', ', $array1), join(', ', $array2));
```
grena: `assertEquals` doesn't take order in account for arrays?

pierallard: Nope ! Example : assertEquals([1,2,3], [3,2,1]);
```     Failed asserting that two arrays are equal.
      --- Expected
      +++ Actual
      @@ @@
       Array (
      -    0 => 1
      +    0 => 3
           1 => 2
      -    2 => 3
      +    2 => 1
       )
```

features/Context/FixturesContext.php
pierallard: Validation is done in the processor 
Pim/Component/Connector/Processor/Denormalization/Processor.php:72
It throws an InvalitItemException when line is not valid.

features/Context/FixturesContext.php
grena: Ooooh ok, so it takes it in account, but we don't need it, I missread your code, thanks for the snippet 👍

features/Context/FixturesContext.php
```diff
@@ -522,11 +546,25 @@ public function thereShouldBeTheFollowingChannels(TableNode $table)
      */
     public function thereShouldBeTheFollowingGroupTypes(TableNode $table)
     {
+        assertEquals([1,2,3], [3,2,1]);
```
anaelChardan: ????

features/Context/FixturesContext.php
```diff
@@ -475,42 +505,36 @@ public function thereShouldBeTheFollowingChannels(TableNode $table)
...
         foreach ($table->getHash() as $data) {
             $channel = $this->getChannel($data['code']);
+            unset($data['code']);
 
-            if (isset($data['label-en_US'])) {
-                assertEquals($data['label-en_US'], $channel->getTranslation('en_US')->getLabel());
-            }
+            foreach ($data as $key => $value) {
+                $matches = null;
+                if ('tree' === $key) {
```
anaelChardan: Is there a way to split all these elseif conditions into checkers functions or other ? I don't know if it's really useful but I find this a little bit long (I know that it was already the case). :-)

features/Context/FixturesContext.php
anaelChardan: +1 :-)

features/Context/FixturesContext.php
pierallard: In fact, none of these functions are exactly the same. They "looks like" but in reality they don't have the same meaning.
That's 1 thing I recently learned: don't try to DRY everywhere.

PR #5679, opened 2017-02-11
Reduce memory leak with integration tests

skeleton:

Currently there is a memory leak with our integration tests. 
When we launch a test with phpunit, Symfony boot a kernel and keep the container in property `$container`.
When a test is finished, the kernel is shutdown and `$container` set to null. This operation is repeated on each test.
Problem with our current implementation is we created our own `$container` where we kept all containers in memory (we saw it with [php-meminfo](https://github.com/BitOne/php-meminfo)). So for 444 tests, we had 444 containers, 444 unit of work, etc \o/

Other baby fix: we launch now integration tests with `debug = false` (we save ~500Mo of memory)

About `gc_collect_cycles`, I don't know why we have to call it explicitly to save so much memory, if someone has an idea :) cc @BitOne ?

So, it's better but there is still a small leak :(
Jobs are never deleted from memory, I suppose it's because we create a specific entity manager (so a specific unit of work) for jobs, but I cannot clear them.
IMHO, we can investigate later, we have time before reach the memory limit.

Before fix:
phpunit-before

Fix without `gc_collect_cycles`:
phpunit-without-gccollect

Fix with `gc_collect_cycles`:
phpunit-with-gccollect

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/tests/ConnectionCloser.php
+++ b/tests/ConnectionCloser.php
@@ -33,5 +33,7 @@ class ConnectionCloser
         foreach ($doctrine->getConnections() as $connection) {
             $connection->close();
         }
+
+        gc_collect_cycles();
     }
 }
--- a/tests/TestCase.php
+++ b/tests/TestCase.php
@@ -3,8 +3,6 @@
 namespace Akeneo\Test\Integration;
 
 use Symfony\Bundle\FrameworkBundle\Test\KernelTestCase;
-use Symfony\Component\Console\Application;
-use Symfony\Component\DependencyInjection\ContainerInterface;
 
 /**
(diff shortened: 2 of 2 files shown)


tests/ConnectionCloser.php
```diff
@@ -33,5 +33,7 @@ public function closeConnections()
         foreach ($doctrine->getConnections() as $connection) {
             $connection->close();
         }
+
+        gc_collect_cycles();
```
phaseinducer: I'd rather put it in the tearDown(), it's not the job of a ConnectionCloser to clean the memory.

PR #5684, opened 2017-02-13
PIM-6153: Allow to send empty array in PriceFilter value for EMPTY and NOT EMPTY operators

willy-ahva:

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Pim/Bundle/CatalogBundle/Doctrine/ORM/Filter/PriceFilter.php
+++ b/src/Pim/Bundle/CatalogBundle/Doctrine/ORM/Filter/PriceFilter.php
@@ -54,6 +54,14 @@ class PriceFilter extends AbstractAttributeFilter implements AttributeFilterInte
         $this->checkLocaleAndScope($attribute, $locale, $scope);
 
         if (Operators::IS_EMPTY === $operator || Operators::IS_NOT_EMPTY === $operator) {
+            if (!array_key_exists('amount', $value)) {
+                $value['amount'] = null;
+            }
+            if (!array_key_exists('currency', $value)) {
+                $value['currency'] = '';
+            } else {
+                $this->checkCurrency($attribute, $value);
+            }
             $this->addEmptyTypeFilter($attribute, $value, $operator, $locale, $scope);
         } else {
             $this->checkValue($attribute, $value);
@@ -197,7 +205,18 @@ class PriceFilter extends AbstractAttributeFilter implements AttributeFilterInte
                 $data
             );
         }
+        $this->checkCurrency($attribute, $data);
+    }
 
+    /**
+     * @param AttributeInterface $attribute
+     * @param array              $data
+     *
+     * @throws InvalidPropertyTypeException
+     * @throws InvalidPropertyException
+     */
+    protected function checkCurrency(AttributeInterface $attribute, $data)
+    {
         if (!is_string($data['currency'])) {
             throw InvalidPropertyTypeException::validArrayStructureExpected(
                 $attribute->getCode(),
@@ -207,7 +226,9 @@ class PriceFilter extends AbstractAttributeFilter implements AttributeFilterInte
             );
         }
 
-        if (!in_array($data['currency'], $this->currencyRepository->getActivatedCurrencyCodes())) {
+        if (!in_array($data['currency'], $this->currencyRepository->getActivatedCurrencyCodes()) &&
+            '' !== $data['currency']
+        ) {
             throw InvalidPropertyException::validEntityCodeExpected(
                 $attribute->getCode(),
                 'currency',
--- a/src/Pim/Bundle/CatalogBundle/tests/integration/PQB/Filter/Price/PriceFilterIntegration.php
+++ b/src/Pim/Bundle/CatalogBundle/tests/integration/PQB/Filter/Price/PriceFilterIntegration.php
@@ -114,6 +114,9 @@ class PriceFilterIntegration extends AbstractFilterTestCase
 
     public function testOperatorEmpty()
     {
+        $result = $this->execute([['a_price', Operators::IS_EMPTY, []]]);
+        $this->assert($result, ['empty_product']);
+
         $result = $this->execute([['a_price', Operators::IS_EMPTY, ['amount' => '', 'currency' => '']]]);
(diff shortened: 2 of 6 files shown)


willy-ahva:
@jjanvier Request from api squad. @skeleton

src/Pim/Bundle/CatalogBundle/Doctrine/ORM/Filter/PriceFilter.php
```diff
@@ -53,6 +53,12 @@ public function addAttributeFilter(
         $this->checkLocaleAndScope($attribute, $locale, $scope, 'price');
 
         if (Operators::IS_EMPTY === $operator || Operators::IS_NOT_EMPTY === $operator) {
+            if (!array_key_exists('amount', $value)) {
```
skeleton: why this check ? why not just force value to '' even if key exists ?

src/Pim/Bundle/CatalogBundle/Doctrine/ORM/Filter/PriceFilter.php
```diff
@@ -53,6 +53,12 @@ public function addAttributeFilter(
...
+                $value['amount'] = '';
```
jjanvier: null

src/Pim/Bundle/CatalogBundle/Doctrine/ORM/Filter/PriceFilter.php
```diff
@@ -53,6 +53,12 @@ public function addAttributeFilter(
...
+            }
+            if (!array_key_exists('currency', $value)) {
+                $value['currency'] = '';
```
jjanvier: null

src/Pim/Bundle/CatalogBundle/tests/integration/PQB/Filter/Price/PriceFilterIntegration.php
```diff
@@ -114,7 +114,7 @@ public function testOperatorSuperiorOrEquals()
 
     public function testOperatorEmpty()
     {
-        $result = $this->execute([['a_price', Operators::IS_EMPTY, ['amount' => '', 'currency' => '']]]);
```
skeleton: remove keys in src/Pim/Bundle/CatalogBundle/tests/integration/PQB/Filter/Price/LocalizableFilterIntegration.php, ScopableLocalizableFilterIntegration & ScopablFilterIntegration

src/Pim/Bundle/CatalogBundle/tests/integration/PQB/Filter/Price/PriceFilterIntegration.php
```diff
@@ -126,7 +126,7 @@ public function testOperatorNotEmpty()
         $result = $this->execute([['a_price', Operators::IS_NOT_EMPTY, ['amount' => '', 'currency' => 'USD']]]);
         $this->assert($result, ['product_one']);
 
-        $result = $this->execute([['a_price', Operators::IS_NOT_EMPTY, ['amount' => '', 'currency' => '']]]);
+        $result = $this->execute([['a_price', Operators::IS_NOT_EMPTY, []]]);
```
jjanvier: maybe you should add new tests cases, and do not delete this one that can still happen?

src/Pim/Bundle/CatalogBundle/tests/integration/PQB/Filter/Price/PriceFilterIntegration.php
```diff
@@ -114,7 +114,7 @@ public function testOperatorSuperiorOrEquals()
 
     public function testOperatorEmpty()
     {
-        $result = $this->execute([['a_price', Operators::IS_EMPTY, ['amount' => '', 'currency' => '']]]);
+        $result = $this->execute([['a_price', Operators::IS_EMPTY, []]]);
```
jjanvier: maybe you should add new tests cases, and do not delete this one that can still happen?

src/Pim/Bundle/CatalogBundle/Doctrine/ORM/Filter/PriceFilter.php
willy-ahva: Because you may want to query if amount not exists on a specific currency for example

src/Pim/Bundle/CatalogBundle/tests/integration/PQB/Filter/Price/PriceFilterIntegration.php
willy-ahva: done

src/Pim/Bundle/CatalogBundle/Doctrine/ORM/Filter/PriceFilter.php
willy-ahva: PQB accepts only scalar values

jjanvier: ?

src/Pim/Bundle/CatalogBundle/Doctrine/ORM/Filter/PriceFilter.php
jjanvier: even if key DOES NOT exists

jjanvier: > Because you may want to query if amount not exists on a specific currency for example
 no you can't, that's not the purpose of the filter

jjanvier:
you added 2 if that can be combined so you should have at least 4 more tests

I still don't understand why you put '' instead of
that does not seem logical to me :/

willy-ahva:
Sorry @jjanvier I answered but it disapears. I tested with null instead of "" but it results to an error from the PQB that tell it accepts only scalar values. Moreover it was already "" values before the modif. And my if can not be combined they have to be separate, or I miss something ?

jjanvier:
the fact that it results in error
is it really something we want? or is it just because it has been coded like that and we didn't touch it?

about the tests: you added 2 consecutives `if`, which means you should have 4 more tests

PR #5687, opened 2017-02-13
Small fixes

grena:

This PR fixes 2 small things:
- Properly clean backslashes for the "I should not see the text" (copy paste from "I should see the text)
- Don't set value to the fake datepicker of the DateFormatter if the value can't be properly parsed by the Bootstrap Datepicker. This is to fix all the `WARNING: Encountered a JS error: 'TypeError: d is null'` we had in some behats.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/features/Context/AssertionContext.php
+++ b/features/Context/AssertionContext.php
@@ -46,6 +46,9 @@ class AssertionContext extends RawMinkContext
      */
     public function assertPageNotContainsText($text)
     {
+        //Remove unecessary escaped antislashes
+        $text = str_replace('\\"', '"', $text);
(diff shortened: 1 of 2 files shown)


features/Context/AssertionContext.php
```diff
@@ -46,6 +46,9 @@ public function assertPageContainsText($text)
      */
     public function assertPageNotContainsText($text)
     {
+        //Remove unecessary escaped antislashes
+        $text = str_replace('\\"', '"', $text);
+        $text = strip_tags($text);
```
grena: Copy paste from the method above.

PR #5690, opened 2017-02-13
API-101: Control access to a resource via ACLs

phaseinducer:

Apply permissions on each entity type (attributes, categories, etc.).

Not sure about how error messages are built in the AccessDeniedHandler, it looks not very robust.
Another solution would be to set hardcoded messages in controllers and access it via reflection. Let me know what you think. (edit: fixed with Doctrine Inflector)

This PR also contains a fix of the permissions edit form (inside the role form).

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Pim/Bundle/ApiBundle/Security/AccessDeniedHandler.php
+++ b/src/Pim/Bundle/ApiBundle/Security/AccessDeniedHandler.php
@@ -1,0 +1,50 @@
+<?php
+
+namespace Pim\Bundle\ApiBundle\Security;
+
+use Doctrine\Common\Inflector\Inflector;
+use Oro\Bundle\SecurityBundle\Exception\AccessDeniedException as OroAccessDeniedException;
+use Symfony\Component\HttpFoundation\JsonResponse;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\Security\Core\Exception\AccessDeniedException;
+use Symfony\Component\Security\Http\Authorization\AccessDeniedHandlerInterface;
+
+/**
+ * Handler responsible for returning a response with accurate error message when the user doesn't have the permission
+ * to access certain parts of the API.
+ *
+ * @author    Yohan Blain <yohan.blain@akeneo.com>
+ * @copyright 2017 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php Open Software License (OSL 3.0)
+ */
+class AccessDeniedHandler implements AccessDeniedHandlerInterface
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function handle(Request $request, AccessDeniedException $exception)
+    {
+        return new JsonResponse(
+            [
+                'code'    => Response::HTTP_FORBIDDEN,
+                'message' => $this->getMessage($request, $exception)
+            ],
+            Response::HTTP_FORBIDDEN
+        );
+    }
+
+    protected function getMessage(Request $request, AccessDeniedException $exception)
+    {
+        if ($exception instanceof OroAccessDeniedException) {
+            $actionName = 'GET' === $request->getMethod() ? 'list' : 'create or update';
+
+            preg_match('`\\\\(\w+)Controller`', $exception->getControllerClass(), $matches);
+            $entityName = str_replace('_', ' ', Inflector::tableize(Inflector::pluralize($matches[1])));
+
+            return sprintf('Access forbidden. You are not allowed to %s %s.', $actionName, $entityName);
+        }
+
+        return 'You are not allowed to access the web API.';
+    }
+}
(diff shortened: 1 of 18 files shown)


src/Pim/Bundle/ApiBundle/Security/AccessDeniedHandler.php
```diff
@@ -0,0 +1,61 @@
...
+            preg_match('`\\\\(\w+)Controller`', $exception->getControllerClass(), $matches);
+
+            if (isset($entityNames[$matches[1]])) {
+                return sprintf(
+                    'Access forbidden. You are not allowed to %s %s.',
+                    $actionName,
+                    $entityNames[$matches[1]]
+                );
+            }
+        }
```
ahocquard: If I understand well (and that's not maybe the case ;) ), we could not have some other controllers with the same name (AttributeController for example) but in another namespace (and handled by acl this way). 

Maybe not a problem, but it could lead to very strange behaviour later.

Note : we can have a limitation if we use POST instead of GET for some kind of request, for example a GET with too much parameters in the URI. But it's a pre-optimization.

src/Pim/Bundle/ApiBundle/Security/AccessDeniedHandler.php
```diff
@@ -0,0 +1,61 @@
...
+            'AttributeOption' => 'attribute options',
+            'Category'        => 'categories',
+            'Channel'         => 'channels',
+            'Family'          => 'families',
+        ];
+
+        if ($exception instanceof OroAccessDeniedException) {
+            $actionName = 'GET' === $request->getMethod() ? 'list' : 'administrate';
+
+            preg_match('`\\\\(\w+)Controller`', $exception->getControllerClass(), $matches);
```
pierallard: ^\\\\(w+)Controller$

src/Pim/Bundle/ApiBundle/Security/AccessDeniedHandler.php
```diff
@@ -0,0 +1,61 @@
...
+
+            if (isset($entityNames[$matches[1]])) {
+                return sprintf(
+                    'Access forbidden. You are not allowed to %s %s.',
+                    $actionName,
+                    $entityNames[$matches[1]]
+                );
```
pierallard: Possible to translate this message ?

src/Pim/Bundle/ApiBundle/Security/AccessDeniedHandler.php
```diff
@@ -0,0 +1,61 @@
...
+            }
+        }
+
+        return 'You are not allowed to access the web API.';
```
pierallard: Same

src/Pim/Bundle/ApiBundle/tests/integration/Security/AuthorizationIntegration.php
```diff
@@ -30,15 +30,317 @@ public function testAccessGranted()
...
+        "type":"pim_catalog_text",
+        "group":"attributeGroupA"
+    }
+JSON;
+
+        $client->request('POST', '/api/rest/v1/attributes', [], [], [], $data);
+
+        $expectedResponse = [
+            'code'    => 403,
+            'message' => 'Access forbidden. You are not allowed to administrate attributes.',
```
pierallard: Administrate is not a fake friend ?

src/Pim/Bundle/ApiBundle/Security/AccessDeniedHandler.php
phaseinducer: Absolutely. The more I think about it, the more I prefer the solution given in the PR description. Not super clean but simpler and more flexible.

src/Pim/Bundle/ApiBundle/Security/AccessDeniedHandler.php
phaseinducer: - The backslashes are not the beginning of the string (it is a full namespace).
- The dollar sign doesn't change anything here.
- Backslashes need to be double-escaped in regexp patterns: http://php.net/manual/en/regexp.reference.escape.php

src/Pim/Bundle/ApiBundle/Security/AccessDeniedHandler.php
phaseinducer: The purpose of this message is not to be displayed to the final user. We chose to translate none of the API messages.

src/Pim/Bundle/ApiBundle/tests/integration/Security/AuthorizationIntegration.php
phaseinducer: Looks ok: http://www.wordreference.com/enfr/administrate
But maybe we can be more specific.

src/Pim/Bundle/ApiBundle/Security/AccessDeniedHandler.php
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    protected function getMessage(Request $request, AccessDeniedException $exception)
+    {
+        $entityNames = [
+            'Attribute'       => 'attributes',
+            'AttributeOption' => 'attribute options',
+            'Category'        => 'categories',
+            'Channel'         => 'channels',
+            'Family'          => 'families',
```
skeleton: Can you add Locale (even if PR is not merged yet) ?

PR #5691, opened 2017-02-13
API-9: create a product in the API

ahocquard:

This PR add the method to create a product in the API.

`OK`: Done / Validated
`-`: Not needed


src/Pim/Component/Catalog/spec/Updater/ProductUpdaterSpec.php
```diff
@@ -17,7 +18,8 @@ function let(
         $this->beConstructedWith(
             $propertySetter,
             $templateUpdater,
-            ['enabled', 'family', 'categories', 'variant_group', 'groups', 'associations']
+            ['enabled', 'family', 'categories', 'variant_group', 'groups', 'associations'],
+            ['identifier', 'created', 'updated']
```
jjanvier: if you ignore the identifier (which in 1.7 is not a field of the product), do you ignore the corresponding product value?
should we?

src/Pim/Bundle/ApiBundle/Controller/ProductController.php
```diff
@@ -283,4 +447,46 @@ protected function checkAttributesParameters($attributes)
...
+     * not match the identifier provided in the property 'values'.
+     *
+     * The identifier provided in the property 'identifier' is mandatory.
+     * The identifier provided in the property 'values' is optional.
+     *
+     * @param array  $data body of the request already decoded
+     *
+     * @throws UnprocessableEntityHttpException
+     */
+    protected function validateBodyIdentifiers(array $data)
```
skeleton: as we don't expose the identifier in property `values` in the api, this method is useless imho, the only way to set an identifier (on a POST action) is with the key `identifier`. if user set it via the property `values`, it should be ignored

src/Pim/Bundle/ApiBundle/tests/integration/Controller/Product/CreateProductIntegration.php
```diff
@@ -0,0 +1,895 @@
...
+
+        $expectedProduct = [
+            'identifier'    => 'product_creation_family',
+            'family'        => 'familyA',
+            'groups'        => [],
+            'variant_group' => null,
+            'categories'    => [],
+            'enabled'       => true,
+            'values'        => [
+                'sku' => [
```
skeleton: nope, we remove the sku product value in the api

src/Pim/Bundle/ApiBundle/tests/integration/Controller/Product/CreateProductIntegration.php
```diff
@@ -0,0 +1,895 @@
...
+    }
+
+    /**
+     * @param array  $expectedProduct normalized data of the product that should be created
+     * @param string $identifier identifier of the product that should be created
+     */
+    protected function assertSameProducts(array $expectedProduct, $identifier)
+    {
+        $product = $this->get('pim_catalog.repository.product')->findOneByIdentifier($identifier);
+        $normalizer = $this->get('pim_catalog.normalizer.standard.product');
```
skeleton: $this->get('serializer')->normalize($product, 'external_api');

src/Pim/Bundle/ApiBundle/Controller/ProductController.php
```diff
@@ -197,6 +253,114 @@ public function deleteAction(Request $request, $code)
...
+    }
+
+    /**
+     * Get a response with HTTP code 201 when an object is created.
+     *
+     * @param ProductInterface $product
+     *
+     * @return Response
+     */
+    protected function getCreateResponse(ProductInterface $product)
```
skeleton: `getResponse` with the http code in parameter (to be more flexible with the patch action) ?

ahocquard: yep, I will refactor it in the patch

src/Pim/Bundle/ApiBundle/Controller/ProductController.php
ahocquard: Nop, it's not what we said in the API-9 ticket.

skeleton: sorry, maybe the ticket is not up to date, but we decided after the creation of this ticket to exclude te PV in the API. as it's not exposed, it can be ignored if submitted.

src/Pim/Component/Catalog/spec/Updater/ProductUpdaterSpec.php
ahocquard: Actually, the identifier is useless in the ProductUpdater, but present in the standard format. 

An updater should accept the standard format. That's why I ignore the identifier. But the product value identifier will not be ignored because it's useful to change an identifier.

src/Pim/Bundle/ApiBundle/tests/integration/Controller/Product/CreateProductIntegration.php
ahocquard: Seen with @skeleton : we keep that like this.

src/Pim/Bundle/ApiBundle/tests/integration/Controller/Product/CreateProductIntegration.php
```diff
@@ -0,0 +1,901 @@
...
+JSON;
+
+        $client->request('POST', 'api/rest/v1/products', [], [], [], $data);
+
+        $expectedContent = [
+            'code'    => 422,
+            'message' => 'Validation failed.',
+            'errors'  => [
+                [
+                    'field'   => 'values[sku].varchar',
```
skeleton: error should be like:
`{"field":"values", "attribute":"sku", "scope":null, "locale":null, "message":"This value should not be blank"}`

as defined in ticket

src/Pim/Bundle/ApiBundle/Controller/ProductController.php
```diff
@@ -197,6 +252,112 @@ public function deleteAction(Request $request, $code)
...
+     * @param ProductInterface $product category to update
+     * @param array            $data    data of the request already decoded
+     *
+     * @throws UnprocessableEntityHttpException
+     */
+    protected function updateProduct(ProductInterface $product, array $data)
+    {
+        try {
+            $this->updater->update($product, $data);
+        } catch (UnknownPropertyException $exception) {
```
jjanvier: just to understand, why this particular check for this one?

src/Pim/Bundle/ApiBundle/Controller/ProductController.php
```diff
@@ -283,4 +444,28 @@ protected function checkAttributesParameters($attributes)
...
+
+    /**
+     * Creates the identifier product value with the same identifier as the value of the identifier property.
+     * It silently overwrite the identifier product value if one is already provided in the input.
+     *
+     * @param array $data
+     *
+     * @return array
+     */
+    protected function createIdentifierProductValue(array $data)
```
jjanvier: if you want to do what suggests the name of your method, you should return only the identifier value, not the whole data

or you can rename it like `replaceIdentifierProductValue`

(personally I prefer the first solution, because in the second one, you don't really replace the value. You return a whole bunch of data. But maybe that's because my naming proposition is shitty)

src/Pim/Bundle/ApiBundle/tests/integration/Controller/Product/CreateProductIntegration.php
```diff
@@ -0,0 +1,901 @@
...
+        $client->request('POST', 'api/rest/v1/products', [], [], [], $data);
+
+        $expectedProduct = [
+            'identifier'    => 'product_creation_product_values',
+            'family'        => 'familyA',
+            'groups'        => ['groupA', 'groupB'],
+            'variant_group' => 'variantA',
+            'categories'    => ['categoryA', 'master'],
+            'enabled'       => true,
+            'values'        => [
```
jjanvier: not in your PR, but maybe we should try to mutualise of those standard values we use everywhere

maybe a simple static method `StandardProduct::values`?

src/Pim/Bundle/ApiBundle/tests/integration/Controller/Product/CreateProductIntegration.php
jjanvier: to be sure to use the same toolchain that the PIM uses

tests/FixturesLoader.php
```diff
@@ -144,6 +145,19 @@ protected function loadImportFiles(array $files)
     }
 
     /**
+     * Load the reference data into the database.
+     */
+    protected function loadReferenceData()
```
jjanvier: :+1:

src/Pim/Bundle/ApiBundle/Controller/ProductController.php
ahocquard: we have a specific message for the API (and at first, that was the reason of the whole refactoring of the exceptions !).

src/Pim/Bundle/ApiBundle/Controller/ProductController.php
ahocquard: Maybe "java" oriented, but what about populateIdentifierProductValue ?

src/Pim/Bundle/ApiBundle/tests/integration/Controller/Product/CreateProductIntegration.php
ahocquard: Discussed with @skeleton : it's not the responsibility of the POST to assert the data returned by the GET, then we keep the tests like this.

PR #5694, opened 2017-02-14
PIM-6159: Fix number comparator when removing numeric value in the PEF

rybus:

When removing a numeric value in the PEF, it might not work and leave the previous value. This was caused by a typo and worked because of a silent Warning PHP Message:
```php
$originals = ['locale' => null, "scope" => null, "data" => 2];

var_dump(null === $originals['data']['data']);
// bool(true)

$originals = ['locale' => null, "scope" => null, "data" => '2'];

var_dump(null === $originals['data']['data']);
// PHP Warning:  Illegal string offset 'data' in /home/acme/test.php on line 9
// PHP Stack trace:
// PHP   1. {main}() /home/acme/test.php:0
// bool(false)
```

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/CHANGELOG-1.6.md
+++ b/CHANGELOG-1.6.md
@@ -4,6 +4,7 @@
 
 - PIM-6149: Remove version number displayed on login page
 - PIM-6157: Improve product exports speed
+- PIM-6159: Fix number comparator when removing numeric value in the PEF
 
 ## Bug fixes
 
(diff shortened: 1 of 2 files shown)


rybus:
DCI is blue

hostep:
Would be great if this could get backported to Akeneo 1.5.x, since we also run into this using Akeneo 1.5.20.

Thanks! :)

PR #5699, opened 2017-02-14
API-110: transform doctrine fields into standard format fields

skeleton:

Sorry, there is some baby fix in this PR, but when I added tests, I found some bugs :)
The point `1` is the initial PR, points `2` & `3` concern bugs fix.

1) Currently, when the standard format is submitted but contain errors, fields returned are doctrine fields (eg. `metricFamily`), not standard format fields (eg. `metric_mamily`).

This PR fix it, it's not an ideal solution as we could have some fields in doctrine which are completely different in the standard format, but for the moment, it's work...

About the propertyPath added in LocaleValidator, without it when a label is wrong, previous message was `translations[0].locale` which means nothing for a user.
I founded this way, but if you have a better solution, I take it.
The other crappy way is to have a mapping directly in ViolationNormalizer :disappointed:

2) About Metric/Price guessers, there was a constraint duplicated for price attribute (we checked twice the value was a numeric, fist time in a isNumericValidator, second time in RangeValidator), so in the response of the API, there was the same error twice.

3) I found a little bug on the TextAttributeSetter, there was no check on type

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Pim/Bundle/ApiBundle/tests/integration/Controller/Product/CreateProductIntegration.php
+++ b/src/Pim/Bundle/ApiBundle/tests/integration/Controller/Product/CreateProductIntegration.php
@@ -771,7 +771,7 @@ JSON;
             'message' => 'Validation failed.',
             'errors'  => [
                 [
-                    'field'   => 'values[sku].varchar',
+                    'field'   => 'identifier',
                     'message' => 'This value should not be blank.',
                 ],
             ],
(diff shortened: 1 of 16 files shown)


src/Pim/Bundle/ApiBundle/tests/integration/Controller/Product/CreateProductIntegration.php
```diff
@@ -808,6 +808,50 @@ public function testResponseWhenIdentifierIsNotFilled()
         $this->assertSame(Response::HTTP_UNPROCESSABLE_ENTITY, $response->getStatusCode());
     }
 
+    public function testResponseWhenAttributeFailed()
```
ahocquard: WhenAttributeValidationFailed

src/Pim/Component/Catalog/Validator/Constraints/LocaleValidator.php
```diff
@@ -33,6 +33,15 @@ public function validate($value, Constraint $constraint)
     {
         $locale = $this->localeRepository->findOneByIdentifier($value);
         if (null === $locale) {
+            if (null !== $constraint->propertyPath) {
+                $this->context->setNode(
```
ahocquard: :/ that's pity that Symfony does not expose a clean solution to change the property path...

src/Pim/Component/Api/Normalizer/Exception/ViolationNormalizer.php
```diff
@@ -47,10 +52,72 @@ public function supportsNormalization($exception, $format = null)
...
+
         foreach ($violations as $violation) {
-            $errors[] = ['field' => $violation->getPropertyPath(), 'message' => $violation->getMessage()];
+            $error = [
+                'field'   => Inflector::tableize($violation->getPropertyPath()),
+                'message' => $violation->getMessage()
+            ];
+
+            if ($violation->getRoot() instanceof ProductInterface &&
+                1 === preg_match('|^values\[(?P<attribute>[a-z0-9-_]+)|i', $violation->getPropertyPath(), $matches)) {
```
phaseinducer: no php cs fixer warning for the end of line?

src/Pim/Component/Api/Normalizer/Exception/ViolationNormalizer.php
```diff
@@ -47,10 +52,72 @@ public function supportsNormalization($exception, $format = null)
...
+     *    "field": "identifier",
+     *    "message": "..."
+     * ]
+     *
+     * @param ConstraintViolationInterface $violation
+     * @param string                       $attributeCode
+     *
+     * @return array
+     */
+    protected function productValuesErrors(ConstraintViolationInterface $violation, $attributeCode)
```
phaseinducer: `getProductValuesError()` ?

phaseinducer: Much cleaner implem btw :+1:

src/Pim/Component/Api/Normalizer/Exception/ViolationNormalizer.php
skeleton: nope

PR #5708, opened 2017-02-15
Api 13 : add patch of a product in the api

ahocquard:

Add patch method in the API.
Patch is upsert.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Pim/Bundle/ApiBundle/Controller/ProductController.php
+++ b/src/Pim/Bundle/ApiBundle/Controller/ProductController.php
@@ -263,7 +263,46 @@ class ProductController
         $this->validateProduct($product);
         $this->saver->save($product);
 
-        $response = $this->getCreateResponse($product);
+        $response = $this->getResponse($product, Response::HTTP_CREATED);
+
+        return $response;
+    }
+
+    /**
+     * @param Request $request
+     * @param string  $code
+     *
+     * @throws BadRequestHttpException
+     *
+     * @return Response
+     */
+    public function partialUpdateAction(Request $request, $code)
+    {
+        $data = $this->getDecodedContent($request->getContent());
+
+        $isCreation = false;
+        $product = $this->productRepository->findOneByIdentifier($code);
+
+        if (null === $product) {
+            $isCreation = true;
+
+            $this->validateCodeConsistency($code, $data);
+            $data['identifier'] = $code;
+            $data = $this->populateIdentifierProductValue($data);
+
+            $product = $this->productBuilder->createProduct();
+        } else {
+            $data['identifier'] = array_key_exists('identifier', $data) ? $data['identifier'] : $code;
+            $data = $this->populateIdentifierProductValue($data);
+            $data = $this->filterEmptyValues($product, $data);
+        }
+
+        $this->updateProduct($product, $data);
+        $this->validateProduct($product);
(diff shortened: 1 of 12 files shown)


src/Pim/Bundle/ApiBundle/Controller/ProductController.php
```diff
@@ -325,6 +363,23 @@ protected function updateProduct(ProductInterface $product, array $data)
         }
     }
 
+    protected function filterValues(ProductInterface $product, array $data)
```
anaelChardan: Missing PHPDoc :-)

jjanvier: filterEmptyValues?

src/Pim/Bundle/ApiBundle/Controller/ProductController.php
```diff
@@ -270,7 +270,45 @@ public function createAction(Request $request)
...
+     * @return Response
+     */
+    public function partialUpdateAction(Request $request, $code)
+    {
+        $data = $this->getDecodedContent($request->getContent());
+
+        $isCreation = false;
+        $product = $this->productRepository->findOneByIdentifier($code);
+
+        if (null === $product) {
```
jjanvier: seems complicated to me
imho
if creation: you create a new product. point
if no identifier value: you validate the code consistency and set it. point

in all cases you can create the identifier product value and filter values

ahocquard: You don't need to filter values in the case of a creation. It's useless and time-consuming. And it could lead to different behavior between the POST and the PATCH method in case of a creation (ok it should not because the empty filter should work as designed !).

On the identifier, it's no that simple.
If you create a product, you should not be able to provide url identifier != body identifier (but body identifier is optional).
If you update a product, you should be able to update the identifier, and so, url identifier != body identifier (but url identifier is optional).

skeleton: maybe extra this code in a method and explain why it's done like that and all problematic we have with the identifier

src/Pim/Component/Catalog/Updater/ProductUpdater.php
```diff
@@ -204,4 +206,71 @@ protected function updateProductVariantValues(ProductInterface $product, array $
...
+     *
+     * @throws InvalidPropertyTypeException
+     */
+    protected function checkProductValuesData($values)
+    {
+        if (!is_array($values)) {
+            throw InvalidPropertyTypeException::arrayExpected('values', static::class, $values);
+        }
+
+        foreach ($values as $code => $productValues) {
```
skeleton: dunno the impact to iterate a second time on each PV. why not put all this validation directly in updateProductValues() ?

ahocquard: I did that to keep the same shcema as the other updaters/setters : 
- validate data
- then process the data

But you're right, it's better to do it in updateProductValues because the structure is dynamic.

src/Pim/Bundle/ApiBundle/Controller/ProductController.php
ahocquard: I will do the refactoring in the patch list PR.

PR #5721, opened 2017-02-17
API-114: create repositories for the API

skeleton:

Implements our own repositories for the API.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Pim/Bundle/ApiBundle/Doctrine/MongoDBODM/Repository/ProductRepository.php
+++ b/src/Pim/Bundle/ApiBundle/Doctrine/MongoDBODM/Repository/ProductRepository.php
@@ -1,0 +1,79 @@
+<?php
+
+namespace Pim\Bundle\ApiBundle\Doctrine\MongoDBODM\Repository;
+
+use Akeneo\Component\StorageUtils\Repository\IdentifiableObjectRepositoryInterface;
+use Doctrine\ODM\MongoDB\DocumentManager;
+use Doctrine\ODM\MongoDB\DocumentRepository;
+use Pim\Component\Api\Repository\ProductRepositoryInterface;
+use Pim\Component\Catalog\Query\ProductQueryBuilderInterface;
+
+/**
+ * @author    Marie Bochu <marie.bochu@akeneo.com>
+ * @copyright 2017 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+class ProductRepository extends DocumentRepository implements ProductRepositoryInterface
+{
+    /** @var IdentifiableObjectRepositoryInterface */
+    protected $productRepository;
+
+    /**
+     * @param DocumentManager                       $em
+     * @param string                                $className
+     * @param IdentifiableObjectRepositoryInterface $productRepository
+     */
+    public function __construct(
+        DocumentManager $em,
+        $className,
+        IdentifiableObjectRepositoryInterface $productRepository
+    ) {
+        parent::__construct($em, $em->getUnitOfWork(), $em->getClassMetadata($className));
+
+        $this->productRepository = $productRepository;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function findOneByIdentifier($identifier)
+    {
+        return $this->productRepository->findOneByIdentifier($identifier);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function searchAfterOffset(ProductQueryBuilderInterface $pqb, $limit, $offset)
+    {
+        $qb = clone $pqb->getQueryBuilder();
+
+        return $qb
+            ->limit($limit)
+            ->skip($offset)
+            ->getQuery()
+            ->execute();
+    }
+
+    /**
+     * {@inheritdoc}
+     */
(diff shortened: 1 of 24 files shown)


phaseinducer:
I reformulate what I said IRL:
- For the products: add count and pagination capabilities to the PQB.
- For other entities: create new repos implementing a single interface.

jjanvier:
all the following is under the "imho" clause

1. about adding methods to the PQB:
- If I'm right, the idea of the PQB was to be able to get all results regarding a particular search (which works perfectly for the grid we have, of for an infinite scroll we could have). That's why we choose to implement it via a cursor. If we keep this approach, the methods `setMaxResult` and `setFirstResult` does not make sense for the PQB. For me, the PQB _is not_ a facade to a product Doctrine repository.
- about the `count`. Except on your API use case, I can't imagine another use case where you need to count the number of results without applying the query. I would say the API is really a specific use case, and I wouldn't put it in the PQB.
2. if everything is _exactly_ the same for other entities, yeah, use a generic class ok. Maybe at least different services so that there is no BC break in the future.

To sum up, I'm in favor of this PR. This goes towards the direction of decoupling features (see my long answer here https://github.com/akeneo/pim-community-dev/pull/5569#discussion_r97776705). 

I find this simple, and let the door open to future possibilities.

src/Pim/Component/Api/Repository/ApiResourceRepositoryInterface.php
```diff
@@ -0,0 +1,16 @@
+<?php
+
+namespace Pim\Component\Api\Repository;
+
+use Akeneo\Component\StorageUtils\Repository\IdentifiableObjectRepositoryInterface;
+
+/**
+ * Repository interface for simple API resources
```
phaseinducer: Maybe explain what is the purpose of this interface? Why it extends the two others, explain that it's generic, suggest that the implementation should decorate other repos, etc.

src/Pim/Bundle/ApiBundle/Doctrine/ORM/Repository/AttributeRepository.php
```diff
@@ -0,0 +1,106 @@
...
+            }
+
+            return (int) $qb
+                ->select('COUNT(DISTINCT o.id)')
+                ->getQuery()
+                ->getSingleScalarResult();
+        } catch (UnexpectedResultException $e) {
+            return 0;
+        }
+    }
```
ahocquard: what's the difference with ApiResourceRepository ?

skeleton: method getIdentifierCode() is added

src/Pim/Bundle/ApiBundle/Doctrine/MongoDBODM/Repository/ProductRepository.php
```diff
@@ -0,0 +1,77 @@
...
+    {
+        return $this->productRepository->findOneByIdentifier($identifier);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function searchAfterOffset(ProductQueryBuilderInterface $pqb, $limit, $offset)
+    {
+        $qb = $pqb->getQueryBuilder();
```
ahocquard: you modify the state of the pqb. You should clone it.

src/Pim/Bundle/ApiBundle/Doctrine/MongoDBODM/Repository/ProductRepository.php
```diff
@@ -0,0 +1,77 @@
...
+            ->getQuery()
+            ->execute();
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function count(ProductQueryBuilderInterface $pqb)
+    {
+        return (int) $pqb->getQueryBuilder()->select('_id')
```
ahocquard: you modify the state of the pqb too. Is your integration test ok with mongo ? Because it should impact the call of the searchAfterOffset (which use the same pqb).

src/Pim/Bundle/ApiBundle/Doctrine/ORM/Repository/ProductRepository.php
```diff
@@ -0,0 +1,83 @@
...
+    {
+        return $this->productRepository->findOneByIdentifier($identifier);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function searchAfterOffset(ProductQueryBuilderInterface $pqb, $limit, $offset)
+    {
+        $qb = $pqb->getQueryBuilder();
```
ahocquard: you modify the state of the pqb.

src/Pim/Bundle/ApiBundle/Doctrine/MongoDBODM/Repository/ProductRepository.php
skeleton: distinct is not needed here as there is never join with other entities, so no risk to have id duplicated.

src/Pim/Bundle/ApiBundle/Doctrine/ORM/Repository/ProductRepository.php
skeleton: yep, I added on count() method but forgot it on this method. thank

PR #5727, opened 2017-02-17
API-39: Depreciate attributeType for type

phaseinducer:

This PR depreciates `attributeType` (getter & setter) in favor of `type`.
We need to do that to match the standard format. Since validation layer has lots of magic at the moment it's difficult to decouple validators/constraints from entities.

`OK`: Done / Validated
`-`: Not needed


phaseinducer:
Actually no migration is needed, in the code it's just a deprecation and the db schema is untouched.

PR #5742, opened 2017-02-21
Fix fixtures loading

aRn0D:
That validator is state full, the locale and the channel are kept in memory. For example, we try to import several catalogs for our integration tests:

* First import with ecommerce and tablet
* Second import with ecommerce and mobile. For this import channels are not reset, the validator still keeps ecommerce and tablet.

I don't know why that var is static, perhaps for performance? 

| Q | A
| Added Specs | -
| Added Behats | -
| Added integration tests | -
| Changelog updated | -
| Review and 2 GTM | -
| Migration script | -
| Tech Doc | -

### DIFF
--- a/src/Pim/Component/Catalog/Validator/AttributeValidatorHelper.php
+++ b/src/Pim/Component/Catalog/Validator/AttributeValidatorHelper.php
@@ -22,10 +22,10 @@ class AttributeValidatorHelper
     protected $scopeRepository;
 
     /** @var array */
-    protected static $localeCodes;
+    protected $localeCodes;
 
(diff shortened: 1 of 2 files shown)


jjanvier:
Yes it was related to performance problems if I remember well.
If you want to merge this PR, you have to benchmark the import execution time before and after. If you show there is not impact, no pb, you'll have my GTM with pleasure :)

Your case is weird. Why in the same request/cli, you would need to import 2 different catalogs? I think your problem comes from here. Not from this class.

nidup:
GTM if no perf issue

PR #5785, opened 2017-02-26
Api 36 : add attribute option patch in api

ahocquard:

Add attribute option PATCH in Api.

Some changes have been done in the Post method too :
The attribute is optional in the body in patch and post, because it's already provided in the url.

I've changed every attribute option tests (GET, POST, PATCH) to assert on JSON instead of PHP array, in a specific commit.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Pim/Bundle/ApiBundle/Controller/AttributeOptionController.php
+++ b/src/Pim/Bundle/ApiBundle/Controller/AttributeOptionController.php
@@ -203,39 +203,67 @@ class AttributeOptionController
      * @param Request $request
      * @param string  $attributeCode
      *
+     * @throws HttpException
+     *
      * @return Response
      *
      * @AclAncestor("pim_api_attribute_option_edit")
      */
     public function createAction(Request $request, $attributeCode)
     {
-        $data = $this->getDecodedContent($request->getContent());
-
         $attribute = $this->getAttribute($attributeCode);
 
-        $this->isAttributeSupportingOptions($attribute);
+        $data = $this->getDecodedContent($request->getContent());
 
-        if (isset($data['attribute']) && $data['attribute'] !== $attributeCode) {
-            throw new UnprocessableEntityHttpException(
-                sprintf(
-                    'Attribute code "%s" in the request body must match "%s" in the URI.',
-                    $data['attribute'],
-                    $attributeCode
-                )
-            );
-        }
+        $this->validateCodeConsistency($attributeCode, null, $data, true);
 
+        $data['attribute'] = $attributeCode;
         $attributeOption = $this->factory->create();
         $this->updateAttributeOption($attributeOption, $data);
-
-        $violations = $this->validator->validate($attributeOption);
-        if (0 !== $violations->count()) {
-            throw new ViolationHttpException($violations);
-        }
+        $this->validateAttributeOption($attributeOption);
 
         $this->saver->save($attributeOption);
 
-        $response = $this->getCreateResponse($attribute, $attributeOption);
+        $response = $this->getResponse($attribute, $attributeOption, Response::HTTP_CREATED);
+
+        return $response;
+    }
+
+    /**
+     * @param Request $request
+     * @param string  $attributeCode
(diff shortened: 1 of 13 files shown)


src/Pim/Bundle/ApiBundle/Controller/AttributeOptionController.php
```diff
@@ -282,4 +330,45 @@ protected function getCreateResponse(AttributeInterface $attribute, AttributeOpt
...
+     * When it's a creation, attribute code and option code provided in the url should match those provided in the body.
+     *
+     * @param string      $attributeCode attribute code provided in the url
+     * @param string|null $optionCode    option code provided in the url (in PATCH), null otherwise (in POST)
+     * @param array       $data          body of the request already decoded
+     * @param boolean     $isCreation    true if it's a creation, false if it's an update
+     *
+     * @throws UnprocessableEntityHttpException
+     */
+    protected function validateCodeConsistency($attributeCode, $optionCode, array $data, $isCreation)
```
skeleton: imvho, you should remove this method and call directly exceptions in create & partialUpdate method. Il will be much more readable instead of set so much condition to know if you can throw exception (personally, I prefer repeat a little bit my code instead of try to factorize too much and lose lisibility :) )

src/Pim/Bundle/CatalogBundle/Resources/config/validation/attribute.yml
```diff
@@ -188,3 +193,15 @@ Pim\Bundle\CatalogBundle\Entity\AttributeOption:
...
+            - Valid: ~
+
+Pim\Bundle\CatalogBundle\Entity\AttributeOptionValue:
+    properties:
+        value:
+            - Length:
+                max: 100
+        locale:
+            - Pim\Component\Catalog\Validator\Constraints\Locale:
+                propertyPath: labels
```
skeleton: You should check that with @phaseinducer. Dunno if it's still relevant

src/Pim/Component/Catalog/Updater/AttributeOptionUpdater.php
```diff
@@ -56,14 +56,9 @@ public function update($attributeOption, array $data, array $options = [])
...
 
-        $isNew = $attributeOption->getId() === null;
-        $readOnlyFields = ['attribute', 'code'];
         foreach ($data as $field => $value) {
             $this->validateDataType($field, $value);
-            $isReadOnlyField = in_array($field, $readOnlyFields);
-            if ($isNew || !$isReadOnlyField) {
-                $this->setData($attributeOption, $field, $value);
-            }
+            $this->setData($attributeOption, $field, $value);
```
skeleton: can you explain that ?

ahocquard: Before modification: 
When you modify the optionCode or attributeCode of an existing attribute option, it was silently ignored (and then not modified).

After modification:
When you modify the optionCode or attributeCode of an existing attribute option, validators raised an exception (because it's immutable properties).

I did that to make consistent the behaviour between the entities. Properties should not be ignored in the updaters if they are immutable.

src/Pim/Bundle/ApiBundle/Controller/AttributeOptionController.php
```diff
@@ -260,16 +291,33 @@ protected function updateAttributeOption(AttributeOptionInterface $attributeOpti
     }
 
     /**
-     * Get a response with HTTP code 201 when an object is created.
+     * Validate an attribute option. It throws an error 422 with every violated constraints if
```
phaseinducer: It's not constraints, it's violations.

src/Pim/Bundle/CatalogBundle/Resources/config/validation/attribute.yml
phaseinducer: I think you can use
```
payload:
  standardPropertyName: labels
```
instead.

PR #5800, opened 2017-02-28
TIP-705: Add tests filters do not support operator should throw

SamirBoulil:

- [X] Add one test case on the pqb filter that should throw when an operator is not supported
- [ ] Add a complex pqb query

`OK`: Done / Validated
`-`: Not needed


SamirBoulil:
PR for 1.7 instead

PR #5808, opened 2017-03-01
Api 112 : add a patch operation on a list of products/categories/attributes/families

ahocquard:

In order to update a list of products in one request, a PATCH on a list of products has been added in this PR.

Each line of the body should represent a standardized product in JSON.
```
{"identifier":"id_1"}
{"identifier":"id_2"}
```

Then, each line is processed one by one and forwarded in the unitary PATCH.
The response of each subrequest is returned line by line (and streamed) : 
```
{"code" :201, "line": 1, "identifier": "id_1"}
{"code" :201, "line": 2, "identifier": "id_2"}
```

Initially, there was no limit on the number of products to import, because it is processing line by line, and so, the content is never fully loaded in memory.

However, it implies a lot of tests and customization of the stack Apache + Php (max_execution_time, max_input_time, maybe some timeout parameters in Apache, etc).

Therefore, for the time being, we have added a limitation on the maximum of products that can be processed in one request (100 by default).

The MIME type of the format (a JSON per line) is `application/vnd.akeneopim+json`, but it's not definitive.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/app/config/pim_parameters.yml
+++ b/app/config/pim_parameters.yml
@@ -48,3 +48,9 @@ parameters:
 
     # Number of items returned by default if limit is not filled in request
     api_pagination_limit_by_default: 10
+
+    # Maximum length in bytes of the resource's json representation when updating a list of resources
+    api_input_buffer_size: 1000000
+
+    # Maximum number of resources when updating a list of resources
+    api_input_max_resources_number: 100
--- a/src/Pim/Bundle/ApiBundle/Controller/ProductController.php
+++ b/src/Pim/Bundle/ApiBundle/Controller/ProductController.php
@@ -9,6 +9,7 @@ use Akeneo\Component\StorageUtils\Repository\IdentifiableObjectRepositoryInterfa
 use Akeneo\Component\StorageUtils\Saver\SaverInterface;
 use Akeneo\Component\StorageUtils\Updater\ObjectUpdaterInterface;
 use Pim\Bundle\ApiBundle\Documentation;
+use Pim\Bundle\ApiBundle\Stream\StreamResourceResponse;
 use Pim\Component\Api\Exception\DocumentedHttpException;
 use Pim\Component\Api\Exception\PaginationParametersException;
 use Pim\Component\Api\Exception\ViolationHttpException;
@@ -29,6 +30,7 @@ use Symfony\Component\HttpFoundation\JsonResponse;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\HttpKernel\Exception\BadRequestHttpException;
+use Symfony\Component\HttpKernel\Exception\HttpException;
 use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
 use Symfony\Component\HttpKernel\Exception\UnprocessableEntityHttpException;
 use Symfony\Component\Routing\RouterInterface;
@@ -87,6 +89,9 @@ class ProductController
     /** @var ProductFilterInterface */
     protected $emptyValuesFilter;
 
+    /** @var StreamResourceResponse */
+    protected $partialUpdateStreamResource;
+
     /** @var array */
     protected $apiConfiguration;
 
@@ -106,6 +111,7 @@ class ProductController
      * @param SaverInterface                        $saver
      * @param RouterInterface                       $router
      * @param ProductFilterInterface                $emptyValuesFilter
+     * @param StreamResourceResponse                $partialUpdateStreamResource
      * @param array                                 $apiConfiguration
      */
     public function __construct(
@@ -124,6 +130,7 @@ class ProductController
         SaverInterface $saver,
         RouterInterface $router,
         ProductFilterInterface $emptyValuesFilter,
+        StreamResourceResponse $partialUpdateStreamResource,
         array $apiConfiguration
     ) {
         $this->pqbFactory = $pqbFactory;
@@ -141,6 +148,7 @@ class ProductController
         $this->saver = $saver;
         $this->router = $router;
         $this->emptyValuesFilter = $emptyValuesFilter;
+        $this->partialUpdateStreamResource = $partialUpdateStreamResource;
         $this->apiConfiguration = $apiConfiguration;
     }
 
@@ -275,7 +283,7 @@ class ProductController
      * @param Request $request
      * @param string  $code
      *
-     * @throws BadRequestHttpException
+     * @throws HttpException
      *
      * @return Response
      */
@@ -283,20 +291,18 @@ class ProductController
     {
         $data = $this->getDecodedContent($request->getContent());
 
-        $isCreation = false;
         $product = $this->productRepository->findOneByIdentifier($code);
+        $isCreation = null === $product;
 
-        if (null === $product) {
-            $isCreation = true;
-
+        if ($isCreation) {
             $this->validateCodeConsistency($code, $data);
-            $data['identifier'] = $code;
-            $data = $this->populateIdentifierProductValue($data);
-
             $product = $this->productBuilder->createProduct();
-        } else {
-            $data['identifier'] = array_key_exists('identifier', $data) ? $data['identifier'] : $code;
-            $data = $this->populateIdentifierProductValue($data);
+        }
+
+        $data['identifier'] = array_key_exists('identifier', $data) ? $data['identifier'] : $code;
+        $data = $this->populateIdentifierProductValue($data);
+
+        if (!$isCreation) {
             $data = $this->filterEmptyValues($product, $data);
         }
 
@@ -306,6 +312,21 @@ class ProductController
 
         $status = $isCreation ? Response::HTTP_CREATED : Response::HTTP_NO_CONTENT;
         $response = $this->getResponse($product, $status);
+
+        return $response;
+    }
+
+    /**
+     * @param Request $request
+     *
+     * @throws HttpException
+     *
+     * @return Response
+     */
+    public function partialUpdateListAction(Request $request)
+    {
+        $resource = $request->getContent(true);
+        $response = $this->partialUpdateStreamResource->streamResponse($resource);
 
         return $response;
     }
--- a/src/Pim/Bundle/ApiBundle/Resources/config/stream.yml
+++ b/src/Pim/Bundle/ApiBundle/Resources/config/stream.yml
@@ -19,3 +19,12 @@ services:
             - '%pim_api.configuration%'
             - 'pim_api.controller.family:partialUpdateAction'
             - 'code'
+
+    pim_api.stream.category_partial_update_stream:
+        class: '%pim_api.stream.stream_resource_response.class%'
+        arguments:
+            - '@http_kernel'
+            - '@pim_catalog.validator.unique_value_set'
+            - '%pim_api.configuration%'
+            - 'pim_api.controller.category:partialUpdateAction'
+            - 'code'
--- a/src/Pim/Bundle/ApiBundle/Resources/config/stream.yml
+++ b/src/Pim/Bundle/ApiBundle/Resources/config/stream.yml
@@ -1,0 +1,12 @@
+parameters:
+    pim_api.stream.stream_resource_response.class: Pim\Bundle\ApiBundle\Stream\StreamResourceResponse
+
+services:
+    pim_api.stream.product_partial_update_stream:
+        class: '%pim_api.stream.stream_resource_response.class%'
+        arguments:
+            - '@http_kernel'
+            - '@pim_catalog.validator.unique_value_set'
+            - '%pim_api.configuration%'
+            - 'pim_api.controller.product:partialUpdateAction'
+            - 'identifier'
--- a/src/Pim/Bundle/ApiBundle/Resources/config/stream.yml
+++ b/src/Pim/Bundle/ApiBundle/Resources/config/stream.yml
@@ -10,3 +10,12 @@ services:
             - '%pim_api.configuration%'
             - 'pim_api.controller.product:partialUpdateAction'
             - 'identifier'
+
+    pim_api.stream.family_partial_update_stream:
+        class: '%pim_api.stream.stream_resource_response.class%'
+        arguments:
+            - '@http_kernel'
+            - '@pim_catalog.validator.unique_value_set'
+            - '%pim_api.configuration%'
+            - 'pim_api.controller.family:partialUpdateAction'
+            - 'code'
--- a/src/Pim/Bundle/ApiBundle/Resources/config/stream.yml
+++ b/src/Pim/Bundle/ApiBundle/Resources/config/stream.yml
@@ -28,3 +28,12 @@ services:
             - '%pim_api.configuration%'
             - 'pim_api.controller.category:partialUpdateAction'
             - 'code'
+
+    pim_api.stream.attribute_partial_update_stream:
+        class: '%pim_api.stream.stream_resource_response.class%'
+        arguments:
+            - '@http_kernel'
+            - '@pim_catalog.validator.unique_value_set'
+            - '%pim_api.configuration%'
+            - 'pim_api.controller.attribute:partialUpdateAction'
+            - 'code'
--- a/src/Pim/Bundle/ApiBundle/Stream/StreamResourceResponse.php
+++ b/src/Pim/Bundle/ApiBundle/Stream/StreamResourceResponse.php
@@ -29,7 +29,7 @@ use Symfony\Component\HttpKernel\HttpKernelInterface;
  */
 class StreamResourceResponse
 {
-    const CONTENT_TYPE = 'application/vnd.akeneopim+json';
+    const CONTENT_TYPE = 'application/vnd.akeneo.collection+json';
 
     /** @var HttpKernelInterface */
     protected $httpKernel;
--- a/src/Pim/Bundle/ApiBundle/Stream/StreamResourceResponse.php
+++ b/src/Pim/Bundle/ApiBundle/Stream/StreamResourceResponse.php
@@ -96,7 +96,6 @@ class StreamResourceResponse
                     if (null === $data) {
                         throw new BadRequestHttpException('Invalid json message received');
                     }
-
                     if (!isset($data[$this->identifierKey]) || '' === trim($data[$this->identifierKey])) {
                         throw new UnprocessableEntityHttpException(sprintf('%s is missing.', ucfirst($this->identifierKey)));
                     }
@@ -109,15 +108,21 @@ class StreamResourceResponse
                     $subResponse = $this->forward(['code' => $data[$this->identifierKey]], $line);
 
                     if ('' !== $subResponse->getContent()) {
-                        $response =  array_merge($response, json_decode($subResponse->getContent(), true));
+                        $subResponse = json_decode($subResponse->getContent(), true);
+                        if (isset($subResponse['code'])) {
+                            $response['status_code'] = $subResponse['code'];
+                            unset($subResponse['code']);
+                        }
+
+                        $response = array_merge($response, $subResponse);
                     } else {
-                        $response['code'] = $subResponse->getStatusCode();
+                        $response['status_code'] = $subResponse->getStatusCode();
                     }
                 } catch (HttpException $e) {
                     $response = [
-                        'line'    => $lineNumber,
-                        'code'    => $e->getStatusCode(),
-                        'message' => $e->getMessage(),
+                        'line'        => $lineNumber,
+                        'status_code' => $e->getStatusCode(),
+                        'message'     => $e->getMessage(),
                     ];
                 }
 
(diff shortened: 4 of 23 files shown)


app/config/pim_parameters.yml
```diff
@@ -48,3 +48,9 @@ parameters:
 
     # Number of items returned by default if limit is not filled in request
     api_pagination_limit_by_default: 10
+
+    # Maximum length of the resource's json representation when updating a list of resources
+    api_input_buffer_size: 1000000
+
+    # Maximum number of resources when updating a list of resources
+    api_input_max_number_resources: 100
```
a2xchip: `api_input_max_resources_number` to be consistent with `api_input_buffer_size` like domain+object+representation_name

src/Pim/Bundle/ApiBundle/Resources/config/stream.yml
```diff
@@ -0,0 +1,13 @@
...
+services:
+    pim_api.stream.product_partial_update_stream:
+        class: '%pim_api.stream.stream_resource_controller.class%'
+        private: true
+        arguments:
+            - '@http_kernel'
+            - '@pim_catalog.validator.unique_value_set'
+            - 'pim_api.controller.product:partialUpdateAction'
+            - 'identifier'
+            - '%pim_api.configuration%'
```
skeleton: put conf after '@pim_catalog.validator.unique_value_set'

src/Pim/Bundle/ApiBundle/Stream/StreamResourceController.php
```diff
@@ -0,0 +1,244 @@
...
+     * @param string              $controllerName
+     * @param string              $identifierKey
+     * @param array               $configuration
+     */
+    public function __construct(
+        HttpKernelInterface $httpKernel,
+        UniqueValuesSet $uniqueValuesSet,
+        $controllerName,
+        $identifierKey,
+        $configuration
```
skeleton: array

src/Pim/Bundle/ApiBundle/Stream/StreamResourceController.php
```diff
@@ -0,0 +1,244 @@
...
+        $resourceContent = $request->getContent(true);
+
+        $this->checkLineNumberInInput($resourceContent);
+
+        $response->setCallback(function () use ($resourceContent) {
+            $lineNumber = 1;
+            $line = true;
+            rewind($resourceContent);
+
+            do {
```
skeleton: you can remove `do` and put directly ` while (false !== $line)`

src/Pim/Bundle/ApiBundle/Stream/StreamResourceController.php
```diff
@@ -0,0 +1,244 @@
...
+    /**
+     * Gets the JSON decoded content. If the content is not a valid JSON, it throws an error 400.
+     *
+     * @param string $content content of a request to decode
+     *
+     * @throws BadRequestHttpException
+     *
+     * @return array
+     */
+    protected function getDecodedContent($content)
```
skeleton: useless method imho. you can put this code directly in your public method

src/Pim/Bundle/ApiBundle/Stream/StreamResourceController.php
```diff
@@ -0,0 +1,244 @@
...
+     *
+     * @throws HttpException
+     */
+    protected function checkLineNumberInInput($streamContent)
+    {
+        $lineNumber = 0;
+        $line = $this->readInputBuffer($streamContent);
+        $maxNumberResources = $this->configuration['input']['max_number_resources'];
+
+        while (false !== $line) {
```
skeleton: why not `while (false !== fgets($streamContent)) {` 
this way you can remove the try/catch

src/Pim/Bundle/ApiBundle/tests/integration/Controller/Product/PartialUpdateListProductIntegration.php
```diff
@@ -0,0 +1,388 @@
...
+JSON;
+
+        $client->request('PATCH', 'api/rest/v1/products', [], [], [], $data);
+
+        $response = $client->getResponse();
+        $this->assertJsonStringEqualsJsonString($expectedContent, $response->getContent());
+        $this->assertSame(Response::HTTP_REQUEST_ENTITY_TOO_LARGE, $response->getStatusCode());
+    }
+
+    public function testPartialUpdateListWithTooLongLines()
```
skeleton: rename it, you don't check only the length of the lines

src/Pim/Bundle/EnrichBundle/Resources/config/bundles/fos_rest.yml
```diff
@@ -7,9 +7,12 @@ fos_rest:
         default_engine: php
         formats:
             json: true
+        mime_types:
+            'akeneo_multi_json': ['application/vnd.akeneopim+json']
```
skeleton: not a big fan of the naming, but I have no idea :disappointed: 
maybe ask to the team

a2xchip: ```akeneo_json_stream```?

src/Pim/Bundle/ApiBundle/Controller/ProductController.php
```diff
@@ -277,34 +292,33 @@ public function createAction(Request $request)
...
+        $data = $this->populateIdentifierProductValue($data);
+
+        if (!$isCreation) {
             $data = $this->filterEmptyValues($product, $data);
         }
 
         $this->updateProduct($product, $data);
         $this->validateProduct($product);
         $this->saver->save($product);
+        $this->detacher->detach($product);
```
ahocquard: Good question...
I think yes. As the allowed resource number is a parameter, if you want to modify the parameter (for the cloud for example), we should take care that it does not leak. 
It could avoid a ticket to the support by a client that put it at 5000...

src/Pim/Bundle/ApiBundle/Resources/config/stream.yml
```diff
@@ -0,0 +1,13 @@
+parameters:
+    pim_api.stream.stream_resource_controller.class: Pim\Bundle\ApiBundle\Stream\StreamResourceController
+
+services:
+    pim_api.stream.product_partial_update_stream:
+        class: '%pim_api.stream.stream_resource_controller.class%'
+        private: true
```
ahocquard: Then, maybe the naming of the service/controller is not the good one. This controller should be injected. Yes, it seems weird, but this service accepts a Request and returns a Response, exactly like a controller. And the ProductController::patchList is just a proxy to this controller.
We talked about that with @phaseinducer, maybe we should talk about this together.

src/Pim/Bundle/ApiBundle/Stream/StreamResourceController.php
ahocquard: readInputBuffer throw an exception if you have a line too long. You can't catch it if it's in the while.
Maybe transform `readInputBuffer` as a method `checkLineLength($line, $streamContent)`.

And then : 
```
while (false !== $line = stream_get_line($streamContent)) {
    $this->checkLineLength($line, $streamContent)
```

src/Pim/Bundle/ApiBundle/Resources/config/stream.yml
skeleton: this service should not accept a Request, there is no reason to do that and it will be easier to use this service in the future if you accept a content.

src/Pim/Bundle/ApiBundle/Stream/StreamResourceController.php
a2xchip: looks much clearer, but i know guys who will write smth like "No var assignments" during review :)

src/Pim/Bundle/ApiBundle/Resources/config/stream.yml
ahocquard: When you say a content, you mean a resource ?

src/Pim/Bundle/EnrichBundle/Resources/config/bundles/fos_rest.yml
ahocquard: Imho, stream is the technical description about how we consume the message (internal behaviour), not a description about the format itself.

src/Pim/Bundle/ApiBundle/Resources/config/stream.yml
skeleton: yup

src/Pim/Bundle/ApiBundle/Stream/StreamResourceResponse.php
```diff
@@ -0,0 +1,229 @@
...
+        while (false !== $line) {
+            $lineNumber++;
+            if ($lineNumber > $maxNumberResources) {
+                throw new HttpException(
+                    Response::HTTP_REQUEST_ENTITY_TOO_LARGE,
+                    sprintf('Too many resources to process, %s is the maximum allowed.', $maxNumberResources)
+                );
+            }
+
+            try {
```
skeleton: really weird to have: 
- an empty try/catch
- call a method checkLineLength() in this method while you check the number of lines.

imho, you should create two methods, one where you don't care if size line > authorized buffer (getNextLine() for instance), one where you throw an exception if size line is too big (checkLineLength()).

src/Pim/Bundle/ApiBundle/Controller/ProductController.php
```diff
@@ -87,30 +90,38 @@ class ProductController
...
-     * @param ValidatorInterface                    $productValidator
-     * @param ProductBuilderInterface               $productBuilder
-     * @param RemoverInterface                      $remover
-     * @param ObjectUpdaterInterface                $updater
-     * @param SaverInterface                        $saver
-     * @param RouterInterface                       $router
-     * @param ProductFilterInterface                $emptyValuesFilter
-     * @param array                                 $apiConfiguration
-     * @param string                                $urlDocumentation
+     * @param ValidatorInterface $productValidator
```
skeleton: align please

src/Pim/Bundle/ApiBundle/Stream/StreamResourceResponse.php
```diff
@@ -0,0 +1,245 @@
...
+    }
+
+    /**
+     * Returns the current line.
+     * If the line is too long for the buffer, consumes the rest of the line.
+     *
+     * @param resource $resourceContent
+     *
+     * @return string content of the line, truncated by the buffer size if the line is too long
+     *
```
skeleton: to remove

src/Pim/Bundle/ApiBundle/Stream/StreamResourceResponse.php
```diff
@@ -0,0 +1,241 @@
...
+                throw new HttpException(
+                    Response::HTTP_REQUEST_ENTITY_TOO_LARGE,
+                    sprintf('Too many resources to process, %s is the maximum allowed.', $maxNumberResources)
+                );
+            }
+        }
+    }
+
+    /**
+     * Forwards the request to another controller.
```
phaseinducer: We already talked about it but it would be great to write it: can you explain why you send the request through the kernel instead of injecting the controller?

src/Pim/Bundle/ApiBundle/Controller/ProductController.php
```diff
@@ -317,6 +331,21 @@ public function partialUpdateAction(Request $request, $code)
...
     /**
+     * @param Request $request
+     *
+     * @throws HttpException
+     *
+     * @return Response
+     */
+    public function partialUpdateListAction(Request $request)
+    {
+        $resourceContent = $request->getContent(true);
```
phaseinducer: `$resource` is fine imho: it's not content you get here, it's simply a resource.

src/Pim/Bundle/ApiBundle/Stream/StreamResourceResponse.php
```diff
@@ -0,0 +1,241 @@
...
+     * @param resource $resourceContent resource containing the whole data to process
+     *
+     * @throws HttpException
+     *
+     * @return StreamedResponse
+     */
+    public function streamResponse($resourceContent)
+    {
+        $response = new StreamedResponse();
+        $response->headers->set('Content-Type', StreamResourceResponse::CONTENT_TYPE);
```
phaseinducer: `static::`

src/Pim/Bundle/ApiBundle/Stream/StreamResourceResponse.php
```diff
@@ -0,0 +1,241 @@
...
+    }
+
+    /**
+     * @param resource $resourceContent resource containing the whole data to process
+     *
+     * @throws HttpException
+     *
+     * @return StreamedResponse
+     */
+    public function streamResponse($resourceContent)
```
phaseinducer: idem, don't use the term "content"

src/Pim/Bundle/ApiBundle/Stream/StreamResourceResponse.php
```diff
@@ -0,0 +1,241 @@
...
+        $bufferSize = $this->configuration['input']['buffer_size'];
+
+        $line = stream_get_line($resourceContent, $bufferSize + 1, PHP_EOL);
+        $buffer = $line;
+
+        while (strlen($buffer) > $bufferSize) {
+            $buffer = stream_get_line($resourceContent, $bufferSize + 1, PHP_EOL);
+        }
+
+        return $line;
```
phaseinducer: I don't understand why doing all these tests on `$buffer` if you return `$line` at the end.

src/Pim/Bundle/ApiBundle/Stream/StreamResourceResponse.php
```diff
@@ -0,0 +1,241 @@
...
+        $response->headers->set('Content-Type', StreamResourceResponse::CONTENT_TYPE);
+
+        $this->checkLineNumberInInput($resourceContent);
+
+        $response->setCallback(function () use ($resourceContent) {
+            rewind($resourceContent);
+
+            $lineNumber = 1;
+            $bufferSize = $this->configuration['input']['buffer_size'];
+            $line = stream_get_line($resourceContent, $bufferSize + 1, PHP_EOL);
```
phaseinducer: You don't use `getNextLine()` here?

src/Pim/Bundle/ApiBundle/Stream/StreamResourceResponse.php
```diff
@@ -0,0 +1,241 @@
...
+        }
+
+        return $line;
+    }
+
+    /**
+     * Checks the length of the line.
+     *
+     * If the line is too long for the buffer, consumes the rest of the line
+     * and throws an error 400.
```
phaseinducer: 413?

src/Pim/Bundle/ApiBundle/Stream/StreamResourceResponse.php
```diff
@@ -0,0 +1,241 @@
...
+            $bufferSize = $this->configuration['input']['buffer_size'];
+            $line = stream_get_line($resourceContent, $bufferSize + 1, PHP_EOL);
+
+            while (false !== $line) {
+                try {
+                    $this->checkLineLength($line, $resourceContent);
+
+                    $data = json_decode($line, true);
+                    if (null === $data) {
+                        throw new BadRequestHttpException('Invalid json message received');
```
phaseinducer: Isn't it inconsistent with other routes? What do we put in 400 errors usually?

src/Pim/Bundle/ApiBundle/Stream/StreamResourceResponse.php
```diff
@@ -0,0 +1,241 @@
...
+                    }
+
+                    if (!isset($data[$this->identifierKey]) || '' === trim($data[$this->identifierKey])) {
```
phaseinducer: Do you really need to check the empty string case?

src/Pim/Bundle/ApiBundle/Stream/StreamResourceResponse.php
```diff
@@ -0,0 +1,241 @@
...
+     */
+    public function __construct(
+        HttpKernelInterface $httpKernel,
+        UniqueValuesSet $uniqueValuesSet,
+        array $configuration,
+        $controllerName,
+        $identifierKey
+    ) {
+        $this->httpKernel = $httpKernel;
+        $this->uniqueValuesSet = $uniqueValuesSet;
```
phaseinducer: Seems unused

src/Pim/Bundle/ApiBundle/Stream/StreamResourceResponse.php
```diff
@@ -0,0 +1,241 @@
...
+                        'line'    => $lineNumber,
+                        'code'    => $e->getStatusCode(),
+                        'message' => $e->getMessage(),
+                    ];
+                }
+
+                $this->uniqueValuesSet->reset();
+                $this->flushOutputBuffer($response, $lineNumber);
+                $lineNumber++;
+                $line = stream_get_line($resourceContent, $bufferSize + 1, PHP_EOL);
```
phaseinducer: Same line as l. 89, looks like you want a `do... while` instead of a `while`.

src/Pim/Bundle/EnrichBundle/Resources/config/bundles/fos_rest.yml
phaseinducer: It's ugly but it's the standard way to describe custom content types.

src/Pim/Bundle/ApiBundle/Stream/StreamResourceResponse.php
```diff
@@ -0,0 +1,241 @@
...
+ * Therefore, response's headers of the different subrequests are not returned, only the content.
+ *
+ * Do note that headers of the global response can not be changed as soon as you are streaming the response's content,
+ * even if you have an error. This is due to the HTTP protocol.
+ *
+ * @author    Alexandre Hocquard <alexandre.hocquard@akeneo.com>
+ * @copyright 2017 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+class StreamResourceResponse
```
phaseinducer: The naming is very misleading imho. If you don't read the class you expect this to be a response instance, but it's absolutely not. Maybe `ResponseStreamer` ?

src/Pim/Bundle/ApiBundle/Controller/ProductController.php
ahocquard: Without it, it leaks. The question is : is it important or not ? Because we expose a parameter to modify that option. (you have my personal on that point just the comment above)

src/Pim/Bundle/ApiBundle/Stream/StreamResourceResponse.php
ahocquard: No, without it you can't create/update the same product (tested).

src/Pim/Bundle/ApiBundle/Stream/StreamResourceResponse.php
ahocquard: No because it consumes the whole line (not just limited to the buffer size).

src/Pim/Bundle/ApiBundle/Stream/StreamResourceResponse.php
ahocquard: It's 400 too and it's consistent.

src/Pim/Bundle/ApiBundle/Stream/StreamResourceResponse.php
ahocquard: Originally, it was, but I did some refactoring (@skeleton told me to use a while ;) )

You can't make a `do...while` because the first line could be wrong (and you should not enter in the loop in this case).

The lines are equals to not assign in the while : 
```
            while (false !== $line = stream_get_line($resourceContent, $bufferSize + 1, PHP_EOL)) 
```

src/Pim/Bundle/ApiBundle/Stream/StreamResourceResponse.php
ahocquard: I consume the buffer of the line (phpdoc).

src/Pim/Bundle/ApiBundle/Stream/StreamResourceResponse.php
ahocquard: `ResourceResponseStreamer` ? Because your naming is very generic.

src/Pim/Bundle/ApiBundle/Stream/StreamResourceResponse.php
ahocquard: I prefer to be sure to forward correctly (normally, you can't call the forwarded route with an empty code).

src/Pim/Bundle/ApiBundle/Stream/StreamResourceResponse.php
skeleton: you return a response, I don't see the problem to call it StreamResourceResponse

PR #5821, opened 2017-03-02
Prevent the creation of session in the API and optimize session management in listeners

ahocquard:

When you request the API, you always create a new session (GET, POST, PATCH).
Therefore, the table `pim_session` has a lot of lines for nothing.
Moreover, it's not logic because an API should be stateless and a cookie is returned in the response.

**Cause:**

It's due to the fact that there are some listeners opening a session.
These listeners are useless for the API, because the standard format is not dependent on the user or the locale. The standard format always returns the same data.

- The first commit prevents this behavior. It will not return anymore a cookie in the API response, and it will not create a session in the database for each request.

- The second commit is an optimization.
When you call `$request->getSession()`, it always creates a session in the database, and lock it.
Then, every concurrent request is waiting to get this lock, and it blocks the whole process.

Several `getSession`are done in controllers, and I think we should let it like this.
But put it in listeners means that you always lock a session, even if your request does not use session anymore later.

Then, we should close it after use in listeners. 
The price to re-open a session in some controllers is not very high compared to lock the session for each request (using a session or not).

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Pim/Bundle/EnrichBundle/EventListener/TranslateFlashMessagesSubscriber.php
+++ b/src/Pim/Bundle/EnrichBundle/EventListener/TranslateFlashMessagesSubscriber.php
@@ -2,6 +2,7 @@
 
 namespace Pim\Bundle\EnrichBundle\EventListener;
 
+use FOS\RestBundle\FOSRestBundle;
 use Pim\Bundle\EnrichBundle\Flash\Message;
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
 use Symfony\Component\HttpKernel\Event\KernelEvent;
@@ -51,7 +52,9 @@ class TranslateFlashMessagesSubscriber implements EventSubscriberInterface
      */
     public function translate(KernelEvent $event)
     {
-        if (!$event->getRequest()->hasSession()) {
+        $request = $event->getRequest();
+
+        if ($request->attributes->has(FOSRestBundle::ZONE_ATTRIBUTE)) {
             return;
         }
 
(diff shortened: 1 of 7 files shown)


jjanvier:
Honestly, I don't understand what problem you try to resolve. 

> Therefore, the table pim_session has a lot of lines for nothing.

Why is it a serious problem? I don't understand.

About the resolution of the problem. I'm really not fan of this kind of optimizations. When a new feature will be released, with the need to touch the request/session, nobody will know he'll have to do such a thing. And we'll get back to the original problem. 

To sum up my POV:
If the number of lines in `pim_session` *do* is problematic, isn't it better to remove the line before rendering the API response?
If the performance optimization *do* is significant, couldn't we find a way to apply it everywhere?

jjanvier:
Maybe something to dig:
isn't it something configurable with Symfony?
aka: when I'm these routes, use a different session handler

ahocquard:
It's just not normal to create a session for an anonymous user.
You can prevent it easily : http://symfony.com/doc/current/session/avoid_session_start.html
Imagine 10000 per hour of the API, creating 1000000 lines in 5 days for... nothing.

I don't get your point of view about deleting the line. First, you must detect where you are coming from (API or not). Then delete the line.
Why fix a problem that you can prevent ?

On the other point, it could improve the performance of the UI, because concurrent requests are not blocking each other for a session lock, even if you are not using it.

jjanvier:
> Imagine 10000 per hour of the API, creating 1000000 lines in 5 days for... nothing.
OK, I understand now.

You don't prevent the problem as you don't do it globally. The problem will re appear the next time we'll do something related (even far related) to the session. IMO, you can't assume that every developer coming in the team will be aware of this problem and will correctly check the session. This is something too tricky, and far hidden in the code stack.

PR #5824, opened 2017-03-02
TIP-666: Bem fixes !

pierallard:


src/Pim/Bundle/UIBundle/Resources/public/less/components/ColumnConfigurator.less
```diff
@@ -16,7 +16,7 @@
       border-right: 1px solid @AknBorderColor;
     }
 
-    &--autoWidth {
+    &--withAuto {
```
phaseinducer: is it a typo?

pierallard: It's a mistake I've done in the previous PR, I used "withAuto" in EE but "autoWidth" in CE... So modifier was not applied...

pierallard: @phaseinducer Here : https://github.com/akeneo/pim-enterprise-dev/pull/2609/files#diff-9949308042a60bc8714385b66a70292bL2

PR #5829, opened 2017-03-03
PIM-6151: disable the loading messages by default

solivier:
On a fresh new 1.7 install, go to System/Configuration and under menu Loading Messages, you’ll see that the yes/no is set on "No" but the Loading messages do appear.
The feature should be deactivated by default when the PIM is installed.

| Review and 2 GTM                  | 1/2

### DIFF
--- a/features/Pim/Behat/Context/NavigationContext.php
+++ b/features/Pim/Behat/Context/NavigationContext.php
@@ -369,6 +369,22 @@ class NavigationContext extends PimContext implements PageObjectAwareInterface
     }
 
     /**
+     * @Then /^I should not see a nice loading message$/
+     */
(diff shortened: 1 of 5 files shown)


features/Pim/Behat/Context/NavigationContext.php
```diff
@@ -369,6 +369,22 @@ public function iShouldSeeANiceLoadingMessage()
     }
 
     /**
+     * @Then /^I should not see a nice loading message$/
```
damien-carcel: :smile:

solivier: ahah sorry didn't find a better name :D

PR #5775, opened 2017-02-24
TIP-694: Add completeness integration tests

SamirBoulil:
(we'll squash commits before merging)

**Description (for Contributor and Core Developer)**

Add integration tests regarding the Completeness feature, it takes into account how the completeness is calculated given:
- the available attribute types in the PIM
- the localizable specificity of an attribute (as well as locale specific)
- the scopable specificity of an attribute
- the locale/scope specificities mixed
- the non required attributes do not influence the calculation

The goal is to be able to refactor the completeness without any fear of breaking something.

**Definition Of Done (for Core Developer only)**

| Q                                 | A
| --------------------------------- | ---
| Added Specs                       | -
| Added Behats                      | -
| Added integration tests           | Ok
| Changelog updated                 | -
| Review and 2 GTM                  | 
| Micro Demo to the PO (Story only) | -
| Migration script                  | -
| Tech Doc                          | -

PR #5830, opened 2017-03-03
TIP-694: isolate integration test files

jjanvier:
As a first step to make our integration tests independent, I propose, that at the beginning of each test (except if you decided to put the option `purgeDatabaseForEachTest` to false):
 - really drop the whole database schema
 - recreate the database schema 

This is just a first step, to be able to merge https://github.com/akeneo/pim-community-dev/pull/5775. With that PR, tests will be slower for sure. But this is just temporary.

What we could do after the 1.7 release is setting up a 1 day squad on this topic. Its goal would be:
- be able to launch really independent tests (with `--process-isolation` for instance, or something else). For that, we need to fix our tests.
- optimize the loading of the database which is very time consuming (see the ideas of Juju and Marie) 
- run the integration tests on Mongo
- handle EE correctly (extension points?)
    - every tests should use the PermissionsCleaner
    - schema should be totally wiped out (currently TWA needs to purge manually some tables)

(and if you're interested to work on this topic, please put a :+1: on the PR)

### DIFF
--- a/tests/DatabasePurger.php
+++ b/tests/DatabasePurger.php
@@ -2,10 +2,11 @@
 
 namespace Akeneo\Test\Integration;
 
-use Akeneo\Bundle\StorageUtilsBundle\DependencyInjection\AkeneoStorageUtilsExtension;
-use Doctrine\Common\DataFixtures\Purger\MongoDBPurger;
-use Doctrine\Common\DataFixtures\Purger\ORMPurger;
+use Symfony\Bundle\FrameworkBundle\Console\Application;
+use Symfony\Component\Console\Input\ArrayInput;
+use Symfony\Component\Console\Output\BufferedOutput;
 use Symfony\Component\DependencyInjection\ContainerInterface;
+use Symfony\Component\HttpKernel\KernelInterface;
 
 /**
  * @author    Yohan Blain <yohan.blain@akeneo.com>
@@ -14,15 +15,19 @@ use Symfony\Component\DependencyInjection\ContainerInterface;
  */
 class DatabasePurger
 {
+    /** @var KernelInterface */
+    protected $kernel;
+
     /** @var ContainerInterface */
     protected $container;
 
     /**
-     * @param ContainerInterface $container
+     * @param KernelInterface $kernel
      */
-    public function __construct(ContainerInterface $container)
+    public function __construct(KernelInterface $kernel)
     {
-        $this->container = $container;
+        $this->kernel = $kernel;
+        $this->container = $kernel->getContainer();
     }
 
     /**
@@ -30,29 +35,20 @@ class DatabasePurger
      */
     public function purge()
     {
-        if (AkeneoStorageUtilsExtension::DOCTRINE_MONGODB_ODM === $this->container->getParameter('pim_catalog_product_storage_driver')) {
-            $purgers[] = new MongoDBPurger($this->container->get('doctrine_mongodb')->getManager());
-        }
+        $application = new Application($this->kernel);
+        $application->setAutoExit(false);
 
-        $em = $this->container->get('doctrine')->getManager();
-        $purgers[] = new ORMPurger($em);
+        $input = new ArrayInput([
+            'command' => 'doctrine:schema:drop',
+            '--env' => 'test',
+            '--force' => true,
+        ]);
+        $output = new BufferedOutput();
 
-        foreach ($purgers as $purger) {
-            $purger->purge();
-        }
+        $exitCode = $application->run($input, $output);
 
-        $connection = $em->getConnection();
-
-        foreach ([
-            'acl_classes',
-            'acl_entries',
-            'acl_object_identities',
-            'acl_object_identity_ancestors',
-            'acl_security_identities'
-        ] as $aclTableName) {
-            $connection->query('SET FOREIGN_KEY_CHECKS=0');
-            $connection->executeUpdate('TRUNCATE `'.$aclTableName.'`');
-            $connection->query('SET FOREIGN_KEY_CHECKS=1');
+        if (0 !== $exitCode) {
+            throw new \Exception(sprintf('Impossible to purge the database! "%s"', $output->fetch()));
         }
     }
 }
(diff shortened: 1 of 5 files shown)


tests/FixturesLoader.php
```diff
@@ -255,4 +258,19 @@ protected function getFilesToLoad(array $directories)
...
+        $input = new ArrayInput([
+            'command' => 'doctrine:schema:create',
+            '--env' => 'test',
+        ]);
+        $output = new BufferedOutput();
+
+        $exitCode = $this->cli->run($input, $output);
+
+        if (0 !== $exitCode) {
+            throw new \Exception(sprintf('Catalog not installable! "%s"', $output->fetch()));
```
jjanvier: schema can not be created

phaseinducer:
Imho you should recreate the schema in the purger, or else there will be a dependency between the purger and the loader (if you try to call the loader without having called the purger just before it will probably crash). Plus it makes more sense to me that the loader handles data only, not structure.
Also be careful: I think you'll need to modify other test cases too.
Nice PR :+1:

jjanvier:
you're right @phaseinducer, I'll do that :)

tests/DatabasePurger.php
```diff
@@ -14,45 +15,40 @@
  */
 class DatabasePurger
```
skeleton: it's not really a purger now, you should rename it

jjanvier:
@phaseinducer @skeleton I fixed your comments

tests/DatabaseSchemaHandler.php
```diff
@@ -0,0 +1,69 @@
...
+            'command' => 'doctrine:schema:drop',
+            '--env' => 'test',
+            '--force' => true,
+        ]);
+        $output = new BufferedOutput();
+
+        $exitCode = $this->cli->run($input, $output);
+
+        if (0 !== $exitCode) {
+            throw new \Exception(sprintf('Impossible to drop the database schema! "%s"', $output->fetch()));
```
aRn0D: perhaps, we should use RuntimeException ?

tests/DatabaseSchemaHandler.php
```diff
@@ -0,0 +1,69 @@
...
+     * @param KernelInterface $kernel
+     */
+    public function __construct(KernelInterface $kernel)
+    {
+        $this->kernel = $kernel;
+        $this->cli = new Application($this->kernel);
+        $this->cli->setAutoExit(false);
+    }
+
+    public function reset()
```
aRn0D: You should add doc for each method

tests/DatabaseSchemaHandler.php
```diff
@@ -0,0 +1,69 @@
...
+
+    /** @var Application */
+    protected $cli;
+
+    /**
+     * @param KernelInterface $kernel
+     */
+    public function __construct(KernelInterface $kernel)
+    {
+        $this->kernel = $kernel;
```
aRn0D: This class attribute is not used, perhaps we should remove it.

tests/DatabaseSchemaHandler.php
```diff
@@ -0,0 +1,69 @@
...
+ * @copyright 2017 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php Open Software License (OSL 3.0)
+ */
+class DatabaseSchemaHandler
+{
+    /** @var KernelInterface */
+    protected $kernel;
+
+    /** @var Application */
+    protected $cli;
```
aRn0D: this should be private because we should not extends this class

tests/FixturesLoader.php
```diff
@@ -117,27 +126,20 @@ protected function loadImportFiles(array $files)
...
+                'code'     => $jobInstance->getCode(),
+                '--no-log' => true,
+                '-v'       => true
+            ]);
+            $output = new BufferedOutput();
+            $exitCode = $this->cli->run($input, $output);
 
             if (0 !== $exitCode) {
-                throw new \Exception(sprintf('Catalog not installable! "%s"', $command->getDisplay()));
+                throw new \Exception(sprintf('Catalog not installable! "%s"', $output->fetch()));
```
aRn0D: RuntimeException ?

tests/FixturesLoader.php
```diff
@@ -117,27 +126,20 @@ protected function loadImportFiles(array $files)
...
-                ]
-            );
+            $input = new ArrayInput([
+                'command'  => 'akeneo:batch:job',
+                'code'     => $jobInstance->getCode(),
+                '--no-log' => true,
+                '-v'       => true
+            ]);
+            $output = new BufferedOutput();
+            $exitCode = $this->cli->run($input, $output);
```
aRn0D: Definitely, this ways is cleaner !

tests/DatabaseSchemaHandler.php
```diff
@@ -0,0 +1,69 @@
...
+use Symfony\Component\Console\Input\ArrayInput;
+use Symfony\Component\Console\Output\BufferedOutput;
+use Symfony\Component\HttpKernel\KernelInterface;
+
+/**
+ * @author    Yohan Blain <yohan.blain@akeneo.com>
+ * @copyright 2017 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php Open Software License (OSL 3.0)
+ */
+class DatabaseSchemaHandler
```
aRn0D: handler sounds manager :)

aRn0D:
Don't forget the EE edition !

tests/DatabaseSchemaHandler.php
jjanvier: @aRn0D extended by yourself for the TWA

tests/DatabaseSchemaHandler.php
jjanvier: used in the override...
I agree we should take care of this problem
see the description of the PR

tests/DatabaseSchemaHandler.php
jjanvier: done

tests/FixturesLoader.php
jjanvier: done

tests/DatabaseSchemaHandler.php
jjanvier: done

tests/DatabaseSchemaHandler.php
jjanvier: completely agree, that sucks
maybe we could find something better during the task force?

PR #5838, opened 2017-03-04
Warning when installing the 1.7 version : fix variant groups spec 

ahocquard:

When you install the PIM, a warning is raised when it generates the autoloader file.
There were two classes with the same name in the spec : `VariantGroupNormalizerSpec` and `GroupNormalizerSpec`.

Actually, the VariantGroupNormalizer was not tested correctly for the versioning.
This PR fixes that problem.

Moreover, I've deleted a function that was not used.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Pim/Component/Catalog/Normalizer/Standard/VariantGroupNormalizer.php
+++ b/src/Pim/Component/Catalog/Normalizer/Standard/VariantGroupNormalizer.php
@@ -13,6 +13,9 @@ use Symfony\Component\Serializer\Normalizer\NormalizerInterface;
  */
 class VariantGroupNormalizer implements NormalizerInterface
 {
+    /** @var string[] */
+    protected $supportedFormats = ['standard'];
+
     /** @var NormalizerInterface */
     protected $translationNormalizer;
 
@@ -56,7 +59,7 @@ class VariantGroupNormalizer implements NormalizerInterface
      */
     public function supportsNormalization($data, $format = null)
     {
-        return $data instanceof GroupInterface && $data->getType()->isVariant() && 'standard' === $format;
+        return $data instanceof GroupInterface && $data->getType()->isVariant() && in_array($format, $this->supportedFormats);
     }
 
     /**
--- a/src/Pim/Bundle/VersioningBundle/Normalizer/Flat/VariantGroupNormalizer.php
+++ b/src/Pim/Bundle/VersioningBundle/Normalizer/Flat/VariantGroupNormalizer.php
@@ -64,29 +64,6 @@ class VariantGroupNormalizer extends BaseNormalizer
     }
 
     /**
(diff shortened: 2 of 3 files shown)


a2xchip:
:+1

src/Pim/Component/Catalog/Normalizer/Standard/VariantGroupNormalizer.php
```diff
@@ -13,6 +13,9 @@
  */
 class VariantGroupNormalizer implements NormalizerInterface
 {
+    /** @var string[] */
+    protected $supportedFormats = ['standard'];
```
jjanvier: I would revert that

all others standard normalizers are done this way
this avoid us the temptation to use the same normalizers for different format

if a format is based on the standard format, then we should compose (or something else, idk) the normalizer
but we should not use this normalizer for 2 different formats

jjanvier:
GTM once fixed

src/Pim/Component/Catalog/Normalizer/Standard/VariantGroupNormalizer.php
ahocquard: But this normalizer does not compose the standard normalizer. It extends it. And then , only standard format is supported. Then, how it works with flat ? Just by using the ProxyGroupNormalizer (useless since the 2.7 version upgrade because normalizer does not cache anymore the class). 

The day you delete that proxy, this normalizer is broken.
If I revert that, my spec is : 

```
   function it_only_supports_csv_and_flat_normalization_of_group(GroupInterface $group, GroupTypeInterface $groupType)
    {
        $group->getType()->willReturn($groupType);
        $groupType->isVariant()->willReturn(true, true, true, true);
        $this->supportsNormalization($group, 'csv')->shouldReturn(false);
        $this->supportsNormalization($group, 'flat')->shouldReturn(false);
        $this->supportsNormalization($group, 'json')->shouldReturn(false);

        $this->supportsNormalization($group, 'standard')->shouldReturn(true);
    }
```

Pretty weird that 'flat' is not supported, no ?

jjanvier: @ahocquard this normalizer _is_ the standard normalizer

if you talk about the `Pim\Bundle\VersioningBundle\Normalizer\Flat\VariantGroupNormalizer`, then I think we should do like the `Pim\Bundle\VersioningBundle\Normalizer\Flat\GroupNormalizer`. It composes the standard.

ahocquard: Seen IRL with @jjanvier : I will merge like that to fix the problem, but I will try to refactor it to : 
- compose with standard format normalizer instead of extending it
- delete ProxyGroupNormalizer

PR #5680, opened 2017-02-13
API-39: Enhance attributes validation

momoss:
The work on the API revealed lots of inconsistencies in attributes validation.
Done in this PR:
- Add missing constraints in attribute validation.
- Fix inconsistent constraints.
- Clean the catalogs.
- Add integration tests at validation level. For the moment only common constraints, identifier, text and textarea are covered because this PR is huge enough, other types will come soon in separated PRs.
- Introduce a way to decouple entity properties and standard format. Unfortunately there is still a mix in existing constraints, it won't be fixed in this PR. For more details see https://akeneo.atlassian.net/browse/TIP-708

Attribute form is broken and will be fixed in a separated PR.

| Q                                 | A
| --------------------------------- | ---
| Added Specs                       | OK
| Added Behats                      | -
| Added integration tests           | OK
| Changelog updated                 | Todo
| Review and 2 GTM                  | Todo
| Micro Demo to the PO (Story only) | Todo
| Migration script                  | -
| Tech Doc                          | -

`Todo`: Pending / Work in progress
`OK`: Done / Validated
`-`: Not needed

PR #5863, opened 2017-03-07
API-126: Change attribute form "scope" input to a yes/no switch

phaseinducer:

In this PR:
- I dropped the ScopableType as now we want to use a simple "switch" type for this field.
- I changed the attribute form type accordingly and changed labels of fields "scopabe" and "localizable" (as discussed with Nolwenn and Virginie).
- I deactivated the js validation for this form, as it adds almost no value and is a nightmare to maintain.
- I updated Bootstrap switch from 1.3.0 to 1.4.0 because it fixes a bug (see https://github.com/Bttstrp/bootstrap-switch/issues/67).
- I fixed behats broken (on purpose) by https://github.com/akeneo/pim-community-dev/pull/5680.

attr_form

`OK`: Done / Validated
`-`: Not needed


features/attribute/create_text_attribute.feature
```diff
@@ -24,24 +24,6 @@ Feature: Create an attribute
     And I change the "Attribute group" to "Other"
     And I save the attribute
     Then I should see validation error "Attribute code may contain only letters, numbers and underscores"
-    And the following attribute codes should not be available:
```
skeleton: why do you remove it ?

src/Pim/Component/Catalog/Validator/Constraints/SingleIdentifierAttributeValidator.php
```diff
@@ -41,7 +41,6 @@ public function validate($attribute, Constraint $constraint)
 
             if ($identifier && $identifier->getId() !== $attribute->getId()) {
                 $this->context->buildViolation($constraint->message)
-                    ->atPath($constraint->propertyPath)
```
skeleton: why do you remove it?

features/attribute/create_text_attribute.feature
phaseinducer: It became unstable with the removal of jsfv and it's now tested in integration tests.

skeleton: are you sure it's tested ? I don"t find reference to it

src/Pim/Component/Catalog/Validator/Constraints/SingleIdentifierAttributeValidator.php
phaseinducer: For the form to be iso with 1.6 this constraint needs to remain global.
For the API we'll use the payload to put the error on the "type" property.

features/attribute/create_text_attribute.feature
phaseinducer: https://github.com/akeneo/pim-community-dev/blob/api-web/src/Pim/Bundle/CatalogBundle/tests/integration/Validation/Attribute/GlobalConstraintsIntegration.php#L240
I didn't tested all the possible words but it's doable if you think it adds value.

skeleton: we should add imho, it could be really annoying if we had regression on this part.
You can use https://phpunit.de/manual/current/en/writing-tests-for-phpunit.html#writing-tests-for-phpunit.data-providers to avoid to create to much methods

PR #5871, opened 2017-03-08
docker-compose installation

mickaelandrieu:
Hello, this is the integration of Docker inside akeneo.

## How it works ?

Both ``make``, [docker](https://docs.docker.com/engine/installation/) and [docker-composer](https://docs.docker.com/compose/install/) are cross-platform,  easy to install and to configure.

> Note that on GNU/Linux and Mac OSX you are forced to launch docker as a sudoer, this is a common trick you need to use: `sudo groupadd docker && sudo gpasswd -a ${USER} docker`. Never execute a docker command as a sudoer :) More information [here](https://docs.docker.com/engine/installation/linux/linux-postinstall/#manage-docker-as-a-non-root-user)

```bash
# install docker and make if you need
# sudoer trick if on gnu/linux or Mac OS X
cp docker-compose.yml.dist docker-compose.yml
make install
# browse localhost:8000
phpstorm . # start with your favorite IDE!
```

## Why do we need that?

As core members, you don't need it: period. But Akeneo developers don't work with Akeneo every day, it depends on projects they have. Most of the time, they need to setup and bootstrap a PIM project in 1 or 2 weeks: spend half a day installing or re-installing Akeneo requirements is time consuming and many of our partners already use their own docker images.

Why not provide a simple - but functional - one? 

### Tested on Ubuntu 16.04
<details>
<summary>Working build on a clean ubuntu 16.04</summary>

ubuntu16.04-ok
</details>

But still a WIP, @nidup got errors I need to reproduce on my VM.

### Tested on Windows 10

<details>
<summary>Working build on Windows 10</summary>

windows-ok
</details>

### Tested on Mac OSX

Not yet: help me :)

## What contains your contribution?

This contribution mainly contains a ``Dockerfile`` a ``docker-compose`` configuration file and a ``Makefile`` to ease the execution of common commands through docker.

Also, I've added an environment variable to be able to set the ``installer_data`` without hardcode a configuration file(https://github.com/akeneo/pim-community-dev/pull/6115), and added some new rules on ``php.ini``, following the Symfony [performances recommendations](http://symfony.com/doc/current/performance.html#optimizing-all-the-files-used-by-symfony).

## What do you expect then?

I expect this contribution will help any developer to setup Akeneo in minutes without alter their own environment. Then I will document how to use it, and everyone will be happy!

Mickaël

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -47,5 +47,6 @@ RUN apt-get update && \
     curl https://getcomposer.org/composer.phar > /usr/local/bin/composer && chmod +x /usr/local/bin/composer
 
 COPY docker/akeneo.local.conf /etc/apache2/sites-available/000-default.conf
+USER docker
 
 CMD ["sudo", "/usr/sbin/apache2ctl", "-D", "FOREGROUND"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -47,6 +47,8 @@ RUN apt-get update && \
     curl https://getcomposer.org/composer.phar > /usr/local/bin/composer && chmod +x /usr/local/bin/composer
 
 COPY docker/akeneo.local.conf /etc/apache2/sites-available/000-default.conf
+COPY . /app
+RUN chown -R docker:docker /app
 USER docker
 
 CMD ["sudo", "/usr/sbin/apache2ctl", "-D", "FOREGROUND"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,42 +9,43 @@ WORKDIR /app
 
 RUN apt-get update && \
     apt-get -yq install \
-        sudo curl supervisor wget zip imagemagick
-RUN echo "deb http://packages.dotdeb.org jessie all" > /etc/apt/sources.list.d/dotdeb.list
-RUN wget https://www.dotdeb.org/dotdeb.gpg && apt-key add dotdeb.gpg
-RUN apt-get update && \
+        sudo curl wget zip imagemagick && \
+
+    echo "deb http://packages.dotdeb.org jessie all" > /etc/apt/sources.list.d/dotdeb.list && \
+    wget https://www.dotdeb.org/dotdeb.gpg && apt-key add dotdeb.gpg && \
+    apt-get update && \
     apt-get -yq install --no-install-recommends \
         apache2 libapache2-mod-php5 php5 php5-cli \
         php5-apcu php5-curl php5-mongo php5-intl php5-mysql \
         php5-gd php5-imagick php5-mcrypt && \
-    apt-get clean && apt-get --yes --quiet autoremove --purge
+    apt-get clean && apt-get --yes --quiet autoremove --purge && \
 
-RUN a2enmod rewrite && \
+    a2enmod rewrite && \
     php5enmod mcrypt && \
-    php5dismod xdebug
+    php5dismod xdebug && \
 
-RUN rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/* /usr/share/locale /usr/share/doc /usr/share/man
+    rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/* /usr/share/locale /usr/share/doc /usr/share/man && \
 
-RUN sudo useradd docker --shell /bin/bash --create-home \
-  && sudo usermod -a -G sudo docker \
-  && echo 'ALL ALL = (ALL) NOPASSWD: ALL' >> /etc/sudoers \
-  && echo 'docker:secret' | chpasswd
+    sudo useradd docker --shell /bin/bash --create-home && \
+    sudo usermod -a -G sudo docker && \
+    echo 'ALL ALL = (ALL) NOPASSWD: ALL' >> /etc/sudoers && \
+    echo 'docker:secret' | chpasswd && \
 
-RUN sed -i "s/;date.timezone =/date.timezone = Etc\/UTC/" /etc/php5/cli/php.ini && \
+    sed -i "s/;date.timezone =/date.timezone = Etc\/UTC/" /etc/php5/cli/php.ini && \
     sed -i "s/;date.timezone =/date.timezone = Etc\/UTC/" /etc/php5/apache2/php.ini && \
     sed -i "s/memory_limit = .*/memory_limit = 2G/" /etc/php5/cli/php.ini && \
-    sed -i "s/memory_limit = .*/memory_limit = 512M/" /etc/php5/apache2/php.ini
+    sed -i "s/memory_limit = .*/memory_limit = 512M/" /etc/php5/apache2/php.ini && \
+    sed -i "s/opcache.max_accelerated_files = .*/opcache.max_accelerated_files = 20000/" /etc/php5/apache2/php.ini && \
+    sed -i "s/realpath_cache_size = .*/realpath_cache_size=4096K/" /etc/php5/apache2/php.ini && \
+    sed -i "s/realpath_cache_ttl = .*/realpath_cache_ttl=600/" /etc/php5/apache2/php.ini && \
 
-RUN sed -i "s/export APACHE_RUN_USER=www-data/export APACHE_RUN_USER=docker/" /etc/apache2/envvars && \
+    sed -i "s/export APACHE_RUN_USER=www-data/export APACHE_RUN_USER=docker/" /etc/apache2/envvars && \
     sed -i "s/export APACHE_RUN_GROUP=www-data/export APACHE_RUN_GROUP=docker/" /etc/apache2/envvars && \
     chown -R docker: /var/lock/apache2 && \
-    echo "ServerName localhost" >> /etc/apache2/apache2.conf
+    echo "ServerName localhost" >> /etc/apache2/apache2.conf &&\
 
-RUN curl https://getcomposer.org/composer.phar > /usr/local/bin/composer && chmod +x /usr/local/bin/composer
+    curl https://getcomposer.org/composer.phar > /usr/local/bin/composer && chmod +x /usr/local/bin/composer
 
-COPY docker/supervisord.conf /etc/supervisor/conf.d/supervisord.conf
 COPY docker/akeneo.local.conf /etc/apache2/sites-available/000-default.conf
 
-USER docker
-
-CMD ["supervisord", "-c", "/etc/supervisor/conf.d/supervisord.conf"]
+CMD ["sudo", "/usr/sbin/apache2ctl", "-D", "FOREGROUND"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,48 @@
+FROM debian:jessie-slim
+MAINTAINER Akeneo Core Team <core@akeneo.com>
+
+ENV DEBIAN_FRONTEND="noninteractive"
+ENV COMPOSER_ALLOW_SUPERUSER=1
+
+EXPOSE 80 3306
+WORKDIR /app
+
+RUN apt-get update && \
+    apt-get -yq install \
+        sudo curl supervisor wget zip imagemagick
+RUN echo "deb http://packages.dotdeb.org jessie all" > /etc/apt/sources.list.d/dotdeb.list
+RUN wget https://www.dotdeb.org/dotdeb.gpg && apt-key add dotdeb.gpg
+RUN apt-get update && \
+    apt-get -yq install --no-install-recommends \
+        apache2 libapache2-mod-php5 php5 php5-cli \
+        php5-apcu php5-curl php5-mongo php5-intl php5-mysql \
+        php5-gd php5-imagick php5-mcrypt && \
+    apt-get clean && apt-get --yes --quiet autoremove --purge
+
+RUN a2enmod rewrite && \
+    php5enmod mcrypt && \
+    php5dismod xdebug
+
+RUN rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/* /usr/share/locale /usr/share/doc /usr/share/man
+
+RUN sudo useradd docker --shell /bin/bash --create-home \
+  && sudo usermod -a -G sudo docker \
+  && echo 'ALL ALL = (ALL) NOPASSWD: ALL' >> /etc/sudoers \
+  && echo 'docker:secret' | chpasswd
+
+RUN sed -i "s/;date.timezone =/date.timezone = Etc\/UTC/" /etc/php5/cli/php.ini && \
+    sed -i "s/;date.timezone =/date.timezone = Etc\/UTC/" /etc/php5/apache2/php.ini && \
+    sed -i "s/memory_limit = .*/memory_limit = 2G/" /etc/php5/cli/php.ini && \
+    sed -i "s/memory_limit = .*/memory_limit = 512M/" /etc/php5/apache2/php.ini
+
+RUN sed -i "s/export APACHE_RUN_USER=www-data/export APACHE_RUN_USER=docker/" /etc/apache2/envvars && \
+    sed -i "s/export APACHE_RUN_GROUP=www-data/export APACHE_RUN_GROUP=docker/" /etc/apache2/envvars && \
+    chown -R docker: /var/lock/apache2 && \
+    echo "ServerName localhost" >> /etc/apache2/apache2.conf
+
+RUN curl https://getcomposer.org/composer.phar > /usr/local/bin/composer && chmod +x /usr/local/bin/composer
+
+COPY docker/supervisord.conf /etc/supervisor/conf.d/supervisord.conf
+COPY docker/akeneo.local.conf /etc/apache2/sites-available/000-default.conf
+
+CMD ["supervisord", "-c", "/etc/supervisor/conf.d/supervisord.conf"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -45,4 +45,6 @@ RUN curl https://getcomposer.org/composer.phar > /usr/local/bin/composer && chmo
 COPY docker/supervisord.conf /etc/supervisor/conf.d/supervisord.conf
 COPY docker/akeneo.local.conf /etc/apache2/sites-available/000-default.conf
 
+USER docker
+
 CMD ["supervisord", "-c", "/etc/supervisor/conf.d/supervisord.conf"]
--- a/Makefile
+++ b/Makefile
@@ -5,12 +5,7 @@ else
     PLATFORM := $(shell uname -s)
 endif
 
-ifeq ($(PLATFORM), Windows)
-	APP=docker-compose exec -T app
-else
-	APP=docker-compose exec --user docker -T app
-endif
-
+APP=docker-compose exec --user docker -T app
 CONSOLE=$(APP) /usr/bin/php app/console
 
 .PHONY: help install pim-install asset-install start stop composer db-create clear-all clean
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,51 @@
+APP=docker-compose exec -T app
+CONSOLE=$(APP) /usr/bin/php app/console
+
+.PHONY: help install pim-install asset-install start stop composer db-create db-update clear-cache clear-all clean
+
+help:           ## Show this help
+	@fgrep -h "##" $(MAKEFILE_LIST) | fgrep -v fgrep | sed -e 's/\\$$//' | sed -e 's/##//'
+
+install:        ## [start composer db-create pim-install clear-all asset-install] Setup the project using Docker and docker-compose
+install: start composer db-create clear-all pim-install asset-install
+
+pim-install:    ## Install the PIM
+	$(CONSOLE) pim:install --env=prod --force
+
+asset-install:  ## Install the assets
+	$(CONSOLE) oro:requirejs:generate-config --env=prod
+	$(CONSOLE) pim:install:assets --env=prod
+
+start:          ## Start the Docker containers
+	docker-compose up -d
+
+stop:           ## Stop the Docker containers and remove the volumes
+	docker-compose down -v
+
+composer:       ## Install the project PHP dependencies
+	$(APP) composer install -o
+
+db-create:      ## Create the database and load the fixtures in it
+	$(CONSOLE) pim:installer:db
+
+db-update:      ## Update the database structure according to the last changes
+	$(CONSOLE) doctrine:schema:update --force
+
+clear-cache:    ## Clear the application cache in development
+	$(CONSOLE) cache:clear
+
+clear-all:      ## Deeply clean the application (remove all the cache, the logs, the sessions and the built assets)
+	$(APP) rm -fr app/archive/*
+	$(APP) rm -fr app/cache/*
+	$(APP) rm -fr app/file_storage/*
+	$(APP) rm -rf app/logs/*
+	$(APP) rm -rf app/sessions/*
+	$(APP) rm -rf web/bundles/*
+	$(APP) rm -rf web/css/*
+	$(APP) rm -rf web/js/*
+	$(APP) rm -rf web/media/*
+	$(APP) rm -rf supervisord.log supervisord.pid .tmp
+
+clean:          ## Removes all generated files
+	- @make clear-all
+	$(APP) rm -rf vendor
--- a/Makefile
+++ b/Makefile
@@ -6,18 +6,18 @@ endif
 
 CONSOLE=$(APP) /usr/bin/php app/console
 
-.PHONY: help install pim-install asset-install start stop composer db-create db-update clear-cache clear-all clean
+.PHONY: help install pim-install asset-install start stop composer db-create clear-all clean
 
 help:           ## Show this help
 	@fgrep -h "##" $(MAKEFILE_LIST) | fgrep -v fgrep | sed -e 's/\\$$//' | sed -e 's/##//'
 
-install:        ## [start composer db-create pim-install clear-all asset-install] Setup the project using Docker and docker-compose
+install:        ## Setup the project using Docker and docker-compose
 install: start composer clear-all pim-install asset-install
 
-pim-install:    ## Install the PIM
+pim-install:    # Install the PIM
 	$(CONSOLE) pim:install --env=prod --force
 
-asset-install:  ## Install the assets
+asset-install:  # Install the assets
 	$(CONSOLE) oro:requirejs:generate-config --env=prod
 	$(CONSOLE) pim:install:assets --env=prod
 
@@ -27,16 +27,16 @@ start:          ## Start the Docker containers
 stop:           ## Stop the Docker containers and remove the volumes
 	docker-compose down -v
 
-composer:       ## Install the project PHP dependencies
+composer:       ## Launch Composer
+	$(APP) composer $(filter-out $@,$(MAKECMDGOALS))
+
+composer-install:       # Install the project PHP dependencies
 	$(APP) composer install -o
 
-db-create:      ## Create the database and load the fixtures in it
+db-create:      # Create the database and load the fixtures in it
 	$(CONSOLE) pim:installer:db
 
-db-update:      ## Update the database structure according to the last changes
-	$(CONSOLE) doctrine:schema:update --force
-
-clear-cache:    ## Clear the application cache in development
+clear-cache:    # Clear the application cache in development
 	$(CONSOLE) cache:clear
 
 clear-all:      ## Deeply clean the application (remove all the cache, the logs, the sessions and the built assets)
@@ -50,6 +50,14 @@ clear-all:      ## Deeply clean the application (remove all the cache, the logs,
 	$(APP) rm -rf web/js/*
 	$(APP) rm -rf web/media/*
 
+phpspec:        ## Launch Phpspec tests
+	$(APP) ./bin/phpspec $(filter-out $@,$(MAKECMDGOALS))
+
+console:        ## Launch Console
+	$(CONSOLE) $(filter-out $@,$(MAKECMDGOALS))
+
 clean:          ## Removes all generated files
 	- @make clear-all
 	$(APP) rm -rf vendor
+%:
+	@:
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,11 @@
-ifeq ($(OS), Windows_NT)
+# Platform detection
+ifeq ($(OS),Windows_NT)
+    PLATFORM := Windows
+else
+    PLATFORM := $(shell uname -s)
+endif
+
+ifeq ($(PLATFORM), Windows)
 	APP=docker-compose exec -T app
 else
 	APP=docker-compose exec --user docker -T app
--- a/Makefile
+++ b/Makefile
@@ -49,7 +49,6 @@ clear-all:      ## Deeply clean the application (remove all the cache, the logs,
 	$(APP) rm -rf web/css/*
 	$(APP) rm -rf web/js/*
 	$(APP) rm -rf web/media/*
-	$(APP) rm -rf supervisord.log supervisord.pid .tmp
 
 clean:          ## Removes all generated files
 	- @make clear-all
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,9 @@
-APP=docker-compose exec -T app
+ifeq ($(OS), Windows_NT)
+	APP=docker-compose exec -T app
+else
+	APP=docker-compose exec --user docker -T app
+endif
+
 CONSOLE=$(APP) /usr/bin/php app/console
 
 .PHONY: help install pim-install asset-install start stop composer db-create db-update clear-cache clear-all clean
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,7 @@ help:           ## Show this help
 	@fgrep -h "##" $(MAKEFILE_LIST) | fgrep -v fgrep | sed -e 's/\\$$//' | sed -e 's/##//'
 
 install:        ## [start composer db-create pim-install clear-all asset-install] Setup the project using Docker and docker-compose
-install: start composer db-create clear-all pim-install asset-install
+install: start composer clear-all pim-install asset-install
 
 pim-install:    ## Install the PIM
 	$(CONSOLE) pim:install --env=prod --force
--- a/composer.json
+++ b/composer.json
@@ -135,7 +135,8 @@
                 "database_port": "PIM_DATABASE_PORT",
                 "database_name": "PIM_DATABASE_NAME",
                 "database_user": "PIM_DATABASE_USER",
-                "database_password": "PIM_DATABASE_PASSWORD"
+                "database_password": "PIM_DATABASE_PASSWORD",
+                "installer_data": "PIM_INSTALLER_DATA"
             }
         },
         "branch-alias": {
--- a/docker-compose.yml.dist
+++ b/docker-compose.yml.dist
@@ -1,0 +1,25 @@
+version: "2"
+
+services:
+    app:
+        environment:
+            PIM_DATABASE_HOST: db
+            PIM_DATABASE_PORT: 3306
+            PIM_INSTALLER_DATA: PimInstallerBundle:minimal
+        build:
+            context: .
+            dockerfile: Dockerfile
+        depends_on:
+            - db
+        volumes:
+            - .:/app
+        ports:
+            - 8000:80
+
+    db:
+        image: mysql:5.6
+        environment:
+            MYSQL_ROOT_PASSWORD: akeneo_pim
+            MYSQL_DATABASE: akeneo_pim
+            MYSQL_USER: akeneo_pim
+            MYSQL_PASSWORD: akeneo_pim
--- a/docker/akeneo.local.conf
+++ b/docker/akeneo.local.conf
@@ -1,0 +1,13 @@
+<VirtualHost *:80>
+    ServerName akeneo.dev
+
+    DocumentRoot /app/web
+    <Directory /app/web>
+        AllowOverride All
+        Require all granted
+    </Directory>
+
+    ErrorLog ${APACHE_LOG_DIR}/akeneo-pim_error.log
+    LogLevel warn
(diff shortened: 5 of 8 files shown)


app/config/parameters.yml.dist
```diff
@@ -1,7 +1,7 @@
 parameters:
     database_driver:   pdo_mysql
-    database_host:     localhost
-    database_port:     ~
+    database_host:     db
```
mickaelandrieu: note to myself: use the of way to use environement variables

composer.json
```diff
@@ -27,7 +27,7 @@
         "doctrine/cache": "1.6.0",
         "doctrine/common": "2.4.2",
         "doctrine/data-fixtures": "1.0.0",
-        "doctrine/doctrine-bundle": "1.2.0",
+        "doctrine/doctrine-bundle": "1.6.7",
```
mickaelandrieu: needed to access `--if-exists` and ``--if-not-exists`` option of doctrine create/delete commands

docker-compose.yml
```diff
@@ -0,0 +1,21 @@
...
+            - /app/var/cache
+            - /app/var/sessions
+
+    db:
+        image: mysql:5.6
+        environment:
+            MYSQL_ROOT_PASSWORD: akeneo_pim
+            MYSQL_DATABASE: akeneo_pim
+            MYSQL_USER: akeneo_pim
+            MYSQL_PASSWORD: akeneo_pim
```
mickaelandrieu: note to myself: add mongodb configuration

mickaelandrieu:
Note the installation is now Ok, I've only one error =>
trads

It's happens after I logged in to the back office

app/config/pim_parameters.yml
```diff
@@ -25,7 +25,7 @@ parameters:
     # installer_data: %kernel.root_dir%/../vendor/akeneo/catalogs/master/community/small/fixtures
     # or you can use one of the provided catalogs
     # use PimInstallerBundle:minimal for minimal data set
-    installer_data:       PimInstallerBundle:icecat_demo_dev
+    installer_data:       PimInstallerBundle:minimal
```
mickaelandrieu: Can we make it default value? This doesn't make sense to make that developer want BIG import; only sales representatives may be interested in ice at demo data.

docker/dev/php.ini
```diff
@@ -0,0 +1,8 @@
+; Add here the parameters that you want to override
+
+short_open_tag = Off
+date.timezone = Europe/Paris
```
damien-carcel: We recommand "Etc/UTC" in the doc.

Makefile
```diff
@@ -0,0 +1,54 @@
...
+
+stop:           ## Stop the Docker containers
+	docker-compose down
+
+composer:       ## Install the project PHP dependencies
+	$(APP) composer install -o
+
+db-create:      ## Create the database and load the fixtures in it
+	$(APP) php -r "for(;;){if(@fsockopen('db',3306)){break;}}" # Wait for MySQL
+	$(CONSOLE) do:da:dr --connection=default --force --if-exists
```
damien-carcel: please use full command name, it is easier to understand for newcomers.

Makefile
```diff
@@ -0,0 +1,54 @@
...
+	$(APP) rm -rf web/js/*
+	$(APP) rm -rf web/media/*
+	$(APP) rm -rf supervisord.log supervisord.pid .tmp
+
+clean:          ## Removes all generated files
+	- @make clear-all
+	$(APP) rm -rf vendor
+
+perm:           ## Fix the application cache and logs permissions
+	$(APP) chmod 777 -R app/cache app/logs app/sessions app/archive app/file_storage
```
damien-carcel: missing EOL

app/config/parameters.yml.dist
```diff
@@ -1,7 +1,7 @@
 parameters:
     database_driver:   pdo_mysql
-    database_host:     localhost
```
damien-carcel: You should not change that, as most people do not use docker. This should be done by your installation script.

app/config/pim_parameters.yml
damien-carcel: I do not agree. As dev, we want to have some data in the PIM (and it is not a big import, there is only a thousand products, it is very small).

composer.json
damien-carcel: That need to be tested on the CI, I am not sure there is not some adjustment to do in the PIM code. @jjanvier?

dev.dockerfile
```diff
@@ -0,0 +1,68 @@
...
+        php5.6-apc \
+        php5.6-mongo \
+        supervisor \
+        tzdata \
+        wget && \
+
+    apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/* && \
+
+
+    cp /usr/share/zoneinfo/Europe/Paris /etc/localtime && echo "Europe/Paris" > /etc/timezone && \
```
damien-carcel: Etc/UTC

dev.dockerfile
```diff
@@ -0,0 +1,68 @@
+FROM ubuntu:16.04
```
damien-carcel: I think it would be better to use debian-slim: smaller and only PHP 5.6.

docker/dev/php.ini
mickaelandrieu: We recommend using the right location (https://docs.akeneo.com/latest/developer_guide/installation/system_requirements/system_install_ubuntu_1604.html#php), of course this value should be overriden in docker-compose-override but I don't know (yet) how to do it

Makefile
mickaelandrieu: I agree, but I don't think this file will - or should be - edited by newcomers: wdyt?

dev.dockerfile
mickaelandrieu: yes! When you will have more time, I'll appreciate a little help on that, this pr is more a POC than a WIP :)

docker/dev/php.ini
damien-carcel: Indeed, an override is the right solution to allow user to choose its own timezone, but it should not be the default value.

Makefile
damien-carcel: Not edited, but read to understand what you do. It could be used as a learning tool/inspiration.

mickaelandrieu: fair enough, I'll update it 👍

dev.dockerfile
damien-carcel: Of course. But for debian, I think you can just change that line with `FROM debian:jessie-slim`.
Installed packages should be the same.

mickaelandrieu:
> I do not agree. As dev, we want to have some data in the PIM (and it is not a big import, there is only a thousand products, it is very small).

With minimal pack => installation time < 3 min
With icecat pack => installation time > 15 min

On my computer i7 (16 cores) 32Go de RAM SSD ... at least can we make the command pim:install accept the "fixtures" option? This solves our disagreement smoothly :)

mickaelandrieu:
"Merci" for this first review @damien-carcel: I'm still learning both Akeneo and Docker, I'm totally aware this is not good enough right now. I update my pr title from WIP to POC :)

damien-carcel:
For a POC, it is already a very nice job ;). Well done dude ;).

damien-carcel:
OK, spotted for installation time: you install xdebug, it is activated by default on Debian/Ubuntu.

You need to unactivate it with `RUN php5dismod xdebug`.

composer.json
jjanvier: no way :)

dev.dockerfile
```diff
@@ -0,0 +1,68 @@
...
+        php5.6-mbstring \
+        php5.6-mcrypt \
+        php5.6-mysql \
+        php5.6-pdo \
+        php5.6-xdebug \
+        php5.6-xml \
+        php5.6-zip \
+        php5.6-apc \
+        php5.6-mongo \
+        supervisor \
```
damien-carcel: If you use Apache+mod_php, you will not need supervisor anymore.

dev.dockerfile
mickaelandrieu: sadly not :(

damien-carcel: Then you should look at the Debian 8 section in our documentation ;-).

damien-carcel: I really think the difference is minimal and it can be quickly changed. You can look at my PHP 5.6 Dockerfile: https://github.com/damien-carcel/Dockerfiles/blob/php-5.6/php/Dockerfile#L29

composer.json
mickaelandrieu: @jjanvier as this "work" is just a POC and aims to be redone by the team, I suggest to let as it for now.

If you downgrade Doctrine you will need to override the commands ``doctrine:database:create`` and ``doctrine:database:drop`` to re-add ``if-exists|if-no-exists`` options or find a workaround to do the same checks.

It's totally up to you, I don't expect this contribution to be merged!

mickaelandrieu:
It works!! @damien-carcel feel free to play with it, improve it, delete it ...

composer.json
damien-carcel: In this case you can keep it that way. As JJ says, Doctrine should be soon updated on master with the same version than yours.

Makefile
```diff
@@ -0,0 +1,55 @@
...
+	$(APP) rm -rf web/js/*
+	$(APP) rm -rf web/media/*
+	$(APP) rm -rf supervisord.log supervisord.pid .tmp
+
+clean:          ## Removes all generated files
+	- @make clear-all
+	$(APP) rm -rf vendor
+
+perm:           ## Fix the application cache and logs permissions
+	$(APP) chmod 777 -R app/cache app/logs app/sessions app/archive app/file_storage web/
```
BitOne: I think this kind of mass destruction weapon is forbidden by the Geneva convention ;)

If you handle properly your HTTP server and cli user, `chmod 777` is not necessary. In fact, when doing a good job, it's never necessary ;)

app/config/parameters.yml.dist
BitOne: Agree with Damien. This file must not be changed with what you are trying to do.

composer.json
BitOne: This file is not in the perimeter of your PR.

dev.dockerfile
```diff
@@ -0,0 +1,65 @@
...
+    TERM="xterm" \
+    DEBIAN_FRONTEND="noninteractive" \
+    COMPOSER_ALLOW_SUPERUSER=1
+
+EXPOSE 80
+WORKDIR /app
+
+
+RUN apt-get update -q && \
+    apt-get install -qy software-properties-common language-pack-en-base && \
```
BitOne: Don't think it's necessary.

dev.dockerfile
```diff
@@ -0,0 +1,65 @@
...
+        supervisor \
+        tzdata \
+        wget && \
+
+    apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/* && \
+    phpdismod xdebug && \
+
+    echo "Etc/UTC" > /etc/timezone && \
+
+    curl -sS https://getcomposer.org/installer | php -- --install-dir=/usr/local/bin --filename=composer && \
```
BitOne: Nop ;)
Never launch an installer from an URL. It's a very big security hole.

Simply do a `curl https://getcomposer.org/composer.phar > /usr/local/bin/composer && chmod +x /usr/local/bin/composer`. Easier to read and no bad surprise ;)

docker-compose.override.yml.dist
```diff
@@ -0,0 +1,12 @@
+version: "2"
```
BitOne: If you provide a yml.dist, don't provide .yml.

docker/dev/cli/php.ini
```diff
@@ -0,0 +1,12 @@
+; Add here the parameters that you want to override
+
+short_open_tag = Off
+date.timezone = Etc/UTC
+error_log = /proc/self/fd/2
+upload_max_filesize = 20M
+post_max_size = 20M
+memory_limit= 768M
+
+xdebug.remote_autostart=0
```
BitOne: Remove xdebug related configuration.

docker/dev/nginx.conf
```diff
@@ -0,0 +1,56 @@
+daemon off;
```
BitOne: Use Apache, as defined in the documentation.

docker/dev/cli/php.ini
```diff
@@ -0,0 +1,12 @@
+; Add here the parameters that you want to override
+
+short_open_tag = Off
+date.timezone = Etc/UTC
+error_log = /proc/self/fd/2
+upload_max_filesize = 20M
```
BitOne: Doesn't make sense on a CLI env.

docker/dev/pool.conf
```diff
@@ -0,0 +1,418 @@
+[global]
```
BitOne: Please remove all commented lines.

docker/dev/supervisord.conf
```diff
@@ -0,0 +1,20 @@
+[supervisord]
```
BitOne: Useless on Apache based env.

Makefile
mickaelandrieu: You are probably right, but I can't get this work without it. Best idea I had: declare docker as nginx user, not sure it's worth when you can edit files from both server and user sides.

This need to be improved, I'm a peaceful people :)

docker-compose.override.yml.dist
mickaelandrieu: true, just forgot to remove it and add the file to .gitgnore => I'll update

docker/dev/cli/php.ini
mickaelandrieu: correct me if I'm wrong, but a lot of imports are done using jobs using Symfony commands, right? May I misunderstood the directive? How can I be sure the upload of files size using jobs will be done?

docker/dev/supervisord.conf
mickaelandrieu: not really, as all docker files using Apache I've tested for Akeneo use php fpm, and start a supervisor to ensure PHP is up.

mickaelandrieu:
@BitOne thanks for this kind review, note this pull request is a POC and not intended to be merged

Makefile
damien-carcel: OMG I didn't spot this one! Indeed, your problem here is your user does not have correct rights. Check what I did on my Apache Dockerfile.

docker/dev/supervisord.conf
damien-carcel: That's the point, do not use FPM :). We don't support it. You have to use mod_php.

docker/dev/cli/php.ini
damien-carcel: Yes, it is needed, but only on FPM or Apache configuration, not for CLI.

mickaelandrieu: @damien-carcel but uploads are mainly done with jobs, so executed in "cli" environment, isn't it ?

docker/dev/supervisord.conf
mickaelandrieu: Didn't know about fpm support, do we have a roadmap for nginx/php-fpm support? Always thought it was more performant ^^

docker/dev/cli/php.ini
damien-carcel: Indeed, you're totally right. I even did it in my own Dockerfiles, but I forgot it :stuck_out_tongue:

docker/dev/supervisord.conf
damien-carcel: You are totally right on the performance part. I can feel it on my machine. But we officially support only what our CI runs, and right now we only run Apache+mod_php with PHP 5.6.

app/config/parameters.yml.dist
gplanchat: can it be replaced by 127.0.0.1 (for mac users) ?

Makefile
gplanchat: I would have put :
```make
	$(APP) chmod 666 -R app/cache app/logs app/sessions app/archive app/file_storage
	$(APP) chmod uga+X -R app/cache app/logs app/sessions app/archive app/file_storage
```

mickaelandrieu:
Big update => ubuntu is not here anymore and replaced by debian:slim, nginx is replaced by apache2.

a2xchip:
```The difference of time install is really impressive. I think there is a problem somewhere, it is way faster on my machine, with Docker, and it is really less powerful than yours...
But Yeah, I totally agree on an option in your command. You could let the parameters.yml.dist as it is, and set the option in your makefile on "minimal" by default. This way it would be very nice.
```
@damien-carcel @mickaelandrieu the performance and diff between `minimal` and `icecat` could be because of mysql settings

app/config/parameters.yml.dist
mickaelandrieu: I've found a solution for this ;)

composer.json
mickaelandrieu: Will you agree to merge this pull request if I override the following commands to implements the missing options? This is the last comment to be adressed.

- [ ] On debian:jessie-small
- [ ] On apache
- [ ] Works on Windows/Ubuntu (need a tester for Mac OSX)
- [ ] Configurable and overridable
- [ ] Pretty good performances
- [ ] Doctrine version

jjanvier:
> I know, this need to be tested, but I need to check if the database already exists or the command will break when called a second time :/ We can also override the related commands, just tell me what to do, but I can't revert without a workaround :)

I don't understand why you need this only for Docker? why doesn't it work the same way that on Linux or Mac environments?

Dockerfile
```diff
@@ -0,0 +1,50 @@
...
+WORKDIR /app
+
+RUN apt-get update && \
+    apt-get -yq install \
+        sudo curl supervisor wget zip imagemagick
+RUN echo "deb http://packages.dotdeb.org jessie all" > /etc/apt/sources.list.d/dotdeb.list
+RUN wget https://www.dotdeb.org/dotdeb.gpg && apt-key add dotdeb.gpg
+RUN apt-get update && \
+    apt-get -yq install --no-install-recommends \
+        mysql-server apache2 libapache2-mod-php5 php5 php5-cli \
```
damien-carcel: Remove mysql-server and everything related as you provide a separate DB in the compose file.

Dockerfile
```diff
@@ -0,0 +1,50 @@
...
+MAINTAINER Akeneo Core Team <core@akeneo.com>
+
+ENV DEBIAN_FRONTEND="noninteractive"
+
+EXPOSE 80 3306
+WORKDIR /app
+
+RUN apt-get update && \
+    apt-get -yq install \
+        sudo curl supervisor wget zip imagemagick
```
damien-carcel: Remove supervisor

Dockerfile
```diff
@@ -0,0 +1,50 @@
...
+    sed -i "s/export APACHE_RUN_GROUP=www-data/export APACHE_RUN_GROUP=docker/" /etc/apache2/envvars && \
+    chown -R docker: /var/lock/apache2 && \
+    echo "ServerName localhost" >> /etc/apache2/apache2.conf
+
+RUN curl https://getcomposer.org/composer.phar > /usr/local/bin/composer && chmod +x /usr/local/bin/composer
+
+ADD docker/supervisord.conf /etc/supervisor/conf.d/supervisord.conf
+ADD docker/akeneo.local.conf /etc/apache2/sites-available/000-default.conf
+
+CMD ["supervisord", "-c", "/etc/supervisor/conf.d/supervisord.conf"]
```
damien-carcel: Launch only Apache in front mode.

Makefile
```diff
@@ -0,0 +1,60 @@
...
+	$(APP) rm -rf web/css/*
+	$(APP) rm -rf web/js/*
+	$(APP) rm -rf web/media/*
+	$(APP) rm -rf supervisord.log supervisord.pid .tmp
+
+clean:          ## Removes all generated files
+	- @make clear-all
+	$(APP) rm -rf vendor
+
+perm:           ## Fix the application cache and logs permissions
```
damien-carcel: Something is wrong here, you should not need that if Apache is correctly configured.

docker-compose.override.yml.dist
damien-carcel: And put the content of the `.yml` instead of this `.dist`. It is to light as it is.

docker/supervisord.conf
```diff
@@ -0,0 +1,12 @@
+[supervisord]
```
damien-carcel: Not needed if Apache is the main process.

docker/akeneo.local.conf
```diff
@@ -0,0 +1,14 @@
+<VirtualHost *:80>
+    ServerName localhost
+
+    DocumentRoot /app/web
+    <Directory /app/web>
+        Options Indexes FollowSymLinks MultiViews
```
damien-carcel: Useless, it is the default configuration of Apache 2.4

docker/akeneo.local.conf
```diff
@@ -0,0 +1,14 @@
+<VirtualHost *:80>
+    ServerName localhost
```
damien-carcel: akeneo.local

mickaelandrieu:
> I don't understand why you need this only for Docker? why doesn't it work the same way that on Linux or Mac environments?
this is not docker related, this is related to the fact we can't reinstall the pim without this step because the database is already here. So the issue is not at all related to this pull request in fact :)

Dockerfile
mickaelandrieu: how to do that?

Makefile
mickaelandrieu: yeah, but the issue is with apache so because this can't work without setup this perms, any idea?

docker/akeneo.local.conf
mickaelandrieu: not needed

mickaelandrieu:
> And put the content of the .yml instead of this .dist. It is to light as it is.
nope, the override should be used only if needed, the installation works even if you don't use the override

docker/supervisord.conf
mickaelandrieu: See my previous comment :)

Dockerfile
mickaelandrieu: true, to be removed

jjanvier:
so in the `pim:installer:db` command, you can drop the base if it exists
seems simple no?

docker/akeneo.local.conf
damien-carcel: It is more of a convention, and it could be useful if on want to do a revert proxy on the host.

Makefile
damien-carcel: As I said, something is wrong, I don't know what without digging, but I don't need to do that at all with my Docker images, Apache nor FPM, and they use native packages from the debian-lite image too.

Dockerfile
damien-carcel: https://github.com/damien-carcel/Dockerfiles/blob/master/apache-php/Dockerfile#L42

mickaelandrieu:
> so in the pim:installer:db command, you can drop the base if it exists
seems simple and would do the job no? 

maybe? didn't know this command let me try :)  

thank you JJ

damien-carcel:
> nope, the override should be used only if needed, the installation works even if you don't use the override
I was thinking about the database configuration that is present in your `docker-compose.yml` and not in the `.dist`. The PIM cannot work without it. Right now, it is working because you have a MySQL server launched in your Apache container (which as I said, you should not), so it is this one that is used by default.

mickaelandrieu:
@damien-carcel ok got it. Maybe I can remove the mysql image? I wanted to use a container for database, but i don't see the interest in this case: what do you think?

damien-carcel:
Yes, there is an interest: it allows developers to easily test other MySQL versions.
Anyway, like for the `parameters.yml` or `behat.yml`, a `docker-compose.yml` should never be commited and should be declared in the `.gitignore` file. Only the `.dist` should be, so the developer can customize the `.yml` as he/she wants.

mickaelandrieu:
> Anyway, like for the parameters.yml or behat.yml, a docker-compose.yml should never be commited and should be declared in the .gitignore file.

This is why I provide a ``docker-compose-override.yml.dist``: out of box the installation should works, else this bad from a DX experience (which is the main objective of this contribution).

> Yes, there is an interest: it allows developers to easily test other MySQL versions.

This is not what we want to do here: we don't want to provide developers their own development docker files (some exist, like your owns), we provide them a quick setup system to help them work with Akeneo.

The scope of this pull request is reduced and it's already take a long time to be accepted/understood/merged: we can discuss of sharing your personal images officialy, but in another pull request please :)

Makefile
mickaelandrieu: maybe it's cause of docker user, I don't know :/

damien-carcel:
In this case, it is the role of the makefile to make a `cp docker-compose.yml.dist docker-compose.yml`.

I do not want a `docker-compose.yml` in my projects other than mine, or to have to remove the existing one, replace it by the one I use, then revert the change/ignoring them before committing. This would just be a pain.

mickaelandrieu:
This will only move the "issue" to the make file, your own ``docker-compose.yml`` file will be overriden on copy: what do you have against ``docker-compose-override.yml`` file? Is there some limitations with this file?

damien-carcel:
Not necessarily, because as an advanced developer, I do not need to use the Makefile. I already know how to set the PIM and use `docker` and `compose`.

I have nothing against `docker-compose-override.yml` in itself. It is just that it goes against the accepted use of `docker-compose.yml`.

For example, on my machine, I ignore globally every `docker-compose.yml` files for `Git` (only because it is not in the PIM `.gitignore`), so I can have what I want inside it. Like my settings of `behat.yml` are different from the one of `behat.yml.dist`: `base_url` and `selenium2.wd_host`, because I use Docker and not a native Apache+PHP install.

And I think a lot of developers that are using `compose` are doing the same (I had seen it on several projects, they always provide the `.dist` and ignore the `.yml`).

Dockerfile
```diff
@@ -0,0 +1,50 @@
...
+        php5-gd php5-imagick php5-mcrypt
+
+RUN a2dismod mpm_event && \
+    a2enmod mpm_prefork && \
+    a2enmod php5 && \
+    a2enmod rewrite && \
+    php5enmod mcrypt && \
+    php5dismod xdebug
+
+RUN apt-get clean && apt-get -yq autoclean && apt-get -yq autoremove && \
```
damien-carcel: Do that in the apt-get install RUN instruction, or it will have no effect.

mickaelandrieu: done

mickaelandrieu:
@jjanvier I've removed Doctrine update, thank you :)

Makefile
```diff
@@ -0,0 +1,51 @@
+APP=docker-compose exec -T app
+CONSOLE=$(APP) /usr/bin/php app/console
+
+.PHONY: help install pim-install asset-install start stop composer db-create db-update clear-cache clear-all perm clean
```
mickaelandrieu: remove `perm` from PHONY

docker/akeneo.local.conf
mickaelandrieu: if this won't break anything I can change, let me test first

docker/akeneo.local.conf
mickaelandrieu: required in fact (not sure I have apache2.4)

docker/supervisord.conf
```diff
@@ -0,0 +1,12 @@
...
+nodaemon=true
+
+[program:mysqld]
```
mickaelandrieu: in all cases the [mysql] part is useless

mickaelandrieu:
@BitOne the contribution have been improved a lot since your last review, let me know what you think about it :)
No more nuclear weapon on rights, no more nginx, no more doctrine update, no more update on actual pim configuration: the fingerprint on project is light. This work on ubuntu and windows 10, probably on mac OS X, and the installation time depends mainly on internet connection: 7 minutes here.

Dockerfile
```diff
@@ -0,0 +1,51 @@
...
+RUN wget https://www.dotdeb.org/dotdeb.gpg && apt-key add dotdeb.gpg
+RUN apt-get update && \
+    apt-get -yq install --no-install-recommends \
+        apache2 libapache2-mod-php5 php5 php5-cli \
+        php5-apcu php5-curl php5-mongo php5-intl php5-mysql \
+        php5-gd php5-imagick php5-mcrypt && \
+    apt-get clean && apt-get --yes --quiet autoremove --purge
+
+RUN a2dismod mpm_event && \
+    a2enmod mpm_prefork && \
```
damien-carcel: Not needed. For me, you can keep mpm_event. @BitOne what do you think?

Dockerfile
```diff
@@ -0,0 +1,51 @@
...
+    a2enmod php5 && \
```
damien-carcel: To remove. It is automatically activated when you install `libapache2-mod-php5` (#DebianPowa).

docker-compose.yml.dist
```diff
@@ -0,0 +1,26 @@
...
+            - 8000:80
+
+    db:
+        image: mysql:5.6
+        environment:
+            MYSQL_ROOT_PASSWORD: akeneo_pim
+            MYSQL_DATABASE: akeneo_pim
+            MYSQL_USER: akeneo_pim
+            MYSQL_PASSWORD: akeneo_pim
+
```
damien-carcel: Missing EOL.

docker/akeneo.local.conf
damien-carcel: You are based on Jessie, so it is Apache 2.4.
I am wondering if you need it because you activated `mpm_prefork`. But I can assure you I don't need it in my images.

damien-carcel:
And you passed the 100 comments, you can claim your badge!
(Admit you did this PR only for that :trollface: )

a2xchip:
@damien-carcel Some month ago i passed 300 comments and have no badges :P

damien-carcel:
@a2xchip You need to claim the badges to have them ;-)

Dockerfile
mickaelandrieu: duly noted => I update

Dockerfile
mickaelandrieu: yes removed, and yes this fix apache configuration too

docker-compose.yml.dist
mickaelandrieu: fixed

mickaelandrieu:
@damien-carcel again, THANK YOU for all the time you spend on this review, I've addressed all your comments, except the supervisor part (still working on it).

We are really close on merging this, and I'm very happy to see docker compose installation become part of Akeneo project :)

PS: yeah I wanted this Atlas badge too :p

Makefile
```diff
@@ -0,0 +1,51 @@
...
+
+asset-install:  ## Install the assets
+	$(CONSOLE) oro:requirejs:generate-config --env=prod
+	$(CONSOLE) pim:install:assets --env=prod
+
+start:          ## Start the Docker containers
+	docker-compose up -d
+
+stop:           ## Stop the Docker containers
+	docker-compose down
```
damien-carcel: You should provide an option `docker-compose down -v` to remove volumes.

docker-compose.yml.dist
damien-carcel: Then remove useless spaces on final line :trollface:

docker/akeneo.local.conf
juliensnz: please don't use .local as it's used by mac os to search on local network. Prefer .dev or anything else

docker/akeneo.local.conf
```diff
@@ -0,0 +1,14 @@
...
+    <Directory /app/web>
+        Options Indexes FollowSymLinks MultiViews
+        AllowOverride All
+        Require all granted
+    </Directory>
+
+    ErrorLog ${APACHE_LOG_DIR}/akeneo-pim_error.log
+    LogLevel warn
+    CustomLog ${APACHE_LOG_DIR}/akeneo-pim_access.log combined
+</VirtualHost>
```
juliensnz: Missing EOF

docker/akeneo.local.conf
damien-carcel: Oh, didn't knew that, very good to know, as I use it in my own images!

Makefile
mickaelandrieu: o_O WOW I'm forced to use and reuse "docker rmi *" command everytime, why you didn't tell me before ??

docker/akeneo.local.conf
mickaelandrieu: I update, I changed for ``akeneo.dev``

docker-compose.yml.dist
mickaelandrieu: done too :)

Makefile
damien-carcel: rmi ? It is to remove images, not volumes.
For volumes it is `docker volume rm whatever`

mickaelandrieu:
@juliensnz I've adressed all your comments, thank you for the Mac OSX-specific behavior comment on vhosts 👍

juliensnz:
Why not PHP7 version?

composer.json
```diff
@@ -136,7 +136,8 @@
                 "database_port": "PIM_DATABASE_PORT",
                 "database_name": "PIM_DATABASE_NAME",
                 "database_user": "PIM_DATABASE_USER",
-                "database_password": "PIM_DATABASE_PASSWORD"
+                "database_password": "PIM_DATABASE_PASSWORD",
+                "installer_data": "PIM_INSTALLER_DATA"
```
mickaelandrieu: this can be extracted from this pull request: allow everyone to select which installer data they want to use *without* touch any configuration file

docker/akeneo.local.conf
nei: Looking forward for this PR :)

What about using the hostname as ENV variable as well? on the docker file you could add 

```yml
services:
  app:
    hostname: ${PIM_URL:-akeneo.dev}
```
and apache config to use ```ServerAlias *.dev```

so people could use nice names with it, also the port would be a nice option.

mickaelandrieu: Thank you for your interest.

Good idea, I'll update.

Dockerfile
```diff
@@ -0,0 +1,50 @@
...
+    sed -i "s/memory_limit = .*/memory_limit = 512M/" /etc/php5/apache2/php.ini
+
+RUN sed -i "s/export APACHE_RUN_USER=www-data/export APACHE_RUN_USER=docker/" /etc/apache2/envvars && \
+    sed -i "s/export APACHE_RUN_GROUP=www-data/export APACHE_RUN_GROUP=docker/" /etc/apache2/envvars && \
+    chown -R docker: /var/lock/apache2 && \
+    echo "ServerName localhost" >> /etc/apache2/apache2.conf
+
+RUN curl https://getcomposer.org/composer.phar > /usr/local/bin/composer && chmod +x /usr/local/bin/composer
+
+COPY docker/supervisord.conf /etc/supervisor/conf.d/supervisord.conf
```
flemzord: To optimize the layers of dockers, it is interesting to make only one RUN

Dockerfile
```diff
@@ -0,0 +1,50 @@
...
+COPY docker/akeneo.local.conf /etc/apache2/sites-available/000-default.conf
+
+USER docker
+
```
flemzord: Add WORKDIR

flemzord:
Why the PR does not validate?

mickaelandrieu:
@flemzord Jenkins builds are not launched automatically.

Dockerfile
mickaelandrieu: you already have a `WORKDIR` :)

mickaelandrieu:
Hello, pull request update:
* no more supervisor
* added new rules on `php.ini` file to improve performances
* added build ok on ubuntu 16.04
Have a nice day

flemzord:
It is interesting to switch from ubuntu to alpine linux to reduce the size of images :)

mickaelandrieu:
@flemzord are you on Mac OSX? would you mind to make a quick try to confirm this is working also on Mac OSX (or not)?

Dockerfile
```diff
@@ -0,0 +1,52 @@
...
+    sed -i "s/memory_limit = .*/memory_limit = 2G/" /etc/php5/cli/php.ini && \
+    sed -i "s/memory_limit = .*/memory_limit = 512M/" /etc/php5/apache2/php.ini && \
+    sed -i "s/opcache.max_accelerated_files = .*/opcache.max_accelerated_files = 20000/" /etc/php5/apache2/php.ini && \
+    sed -i "s/realpath_cache_size = .*/realpath_cache_size=4096K/" /etc/php5/apache2/php.ini && \
+    sed -i "s/realpath_cache_ttl = .*/realpath_cache_ttl=600/" /etc/php5/apache2/php.ini && \
+
+    sed -i "s/export APACHE_RUN_USER=www-data/export APACHE_RUN_USER=docker/" /etc/apache2/envvars && \
+    sed -i "s/export APACHE_RUN_GROUP=www-data/export APACHE_RUN_GROUP=docker/" /etc/apache2/envvars && \
+    chown -R docker: /var/lock/apache2 && \
+    echo "ServerName localhost" >> /etc/apache2/apache2.conf &&\
```
flemzord: no need server name

docker/akeneo.local.conf
```diff
@@ -0,0 +1,13 @@
+<VirtualHost *:80>
+    ServerName akeneo.dev
```
flemzord: no need server name

docker/akeneo.local.conf
```diff
@@ -0,0 +1,13 @@
...
+
+    DocumentRoot /app/web
+    <Directory /app/web>
+        AllowOverride All
+        Require all granted
+    </Directory>
+
+    ErrorLog ${APACHE_LOG_DIR}/akeneo-pim_error.log
```
flemzord: Please use stdout for rewrite log

nidup:
Closed in favor of https://github.com/akeneo/pim-community-dev/pull/6372 

Thx again for initiating this :)

Issue #5897, opened 2017-03-11
Shows console error in case of missing flash message


wakqasahmed:
**I'm reporting a Bug**

As shown in the snapshot, we keep getting a javascript error whenever we click on any import/export profile, in my opinion:

1) Either it should be a simple warning OR
2) If it is really critical (I don't really know why?) we should fix this by sending some message (e.g. empty string) to stop showing error on profile click.

Steps to reproduce:

1) Click Import profiles/Export profiles from Collect/Spread links in the menu.
2) Click any profile and check chrome console for error

![javascript_error](https://cloud.githubusercontent.com/assets/4486133/23820917/78918738-063d-11e7-8fad-2840da1db53c.jpg)

wakqasahmed:
Submitted a PR, please have a look and merge.
https://github.com/akeneo/pim-community-dev/pull/5898

solivier:
Thanks @wakqasahmed you PR has been merged. :+1:

PR #5898, opened 2017-03-11
Show warning instead of an error (if update message missing)

wakqasahmed:
This PR prevents the javascript to console an error if the flash message is missing (but logs a warning on the console)


a2xchip:
Hi @wakqasahmed
Have not checked it yet, just checked #5897 
Could you provide more info in description section: browser used, steps to reproduce, etc

PR #5908, opened 2017-03-12
API-39: Add integration tests on date attributes validation

phaseinducer:


src/Pim/Bundle/CatalogBundle/tests/integration/Validation/Attribute/DateIntegration.php
```diff
@@ -0,0 +1,81 @@
+<?php
+
+namespace Pim\Bundle\CatalogBundle\tests\integration\Validation\Attribute;
+
+/**
+ * @author    Yohan Blain <yohan.blain@akeneo.com>
+ * @copyright 2017 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php Open Software License (OSL 3.0)
+ */
```
ahocquard: no needed in test I think

phaseinducer: idk, we always put that lately

PR #5940, opened 2017-03-17
Fix markdown typos

mickaelandrieu:
Minor fixes in markdown rendering


mickaelandrieu:
@pierallard I don't think we need to launch all the test suite, what do you think? :)

PR #5943, opened 2017-03-17
Pim 6253 : add missing permissions on entities of the API

ahocquard:

Some permissions were missing :( 

`OK`: Done / Validated
`-`: Not needed


CHANGELOG-1.7.md
```diff
@@ -8,6 +8,7 @@
 - PIM-6242: Fix UI glitch on TWA completeness filter search field
 - PIM-6239: Translate scope with catalog locale
 - GITHUB-3435: Sort order products datagrid `Manage filter` options
+- PIM-6253: add missing permissions on entities of the API
```
phaseinducer: Uppercase please ;)

src/Pim/Bundle/ApiBundle/tests/integration/Security/AuthorizationIntegration.php
```diff
@@ -282,6 +282,32 @@ public function testAccessDeniedForListingChannels()
...
+        $client->request('GET', '/api/rest/v1/locales');
+
+        $expectedResponse = [
+            'code'    => 403,
+            'message' => 'Access forbidden. You are not allowed to list locales.',
+        ];
+
+        $response = $client->getResponse();
+        $this->assertSame(Response::HTTP_FORBIDDEN, $response->getStatusCode());
+        $this->assertSame($expectedResponse, json_decode($response->getContent(), true));
```
skeleton: use assertJsonStringEqualsJsonString and compare the expected response with the response without decode it please

src/Pim/Bundle/ApiBundle/tests/integration/Security/AuthorizationIntegration.php
```diff
@@ -282,6 +282,32 @@ public function testAccessDeniedForListingChannels()
...
+    {
+        $client = $this->createAuthenticatedClient();
+
+        $client->request('GET', '/api/rest/v1/locales');
+
+        $response = $client->getResponse();
+        $this->assertSame(Response::HTTP_OK, $response->getStatusCode());
+    }
+
+    public function testAccessDeniedForListingLocales()
```
skeleton: you test only locales routes while you add permissions on channels, categories, etc ?

phaseinducer: The idea is to test each ACL type only once.

ahocquard: We test that the ACL is well applied on one route as we did before.
And it was already tested on the other routes.

skeleton: so, we have nothing to test if route to patch a list of attribute is protected by an ACL ?
That's a real "hole in the racket" :)

ahocquard: If the developper forget to add the ACL, he will forget the test as well... (and that's what happened). If you want, I can test every routes, it can avoid some regressions. But It will just not avoid this kind of problems.

skeleton: Yes, I prefer to add them to avoid regression. I have to cover all our scope (even more when I see current discussion on #product-help on slack ;) )

PR #5945, opened 2017-03-17
PIM-6254: Fix pagination on the API when filters are applied

skeleton:

Add missing groupBy to fix products returned on list when filters are applied 

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Pim/Bundle/ApiBundle/Doctrine/ORM/Repository/ApiResourceRepository.php
+++ b/src/Pim/Bundle/ApiBundle/Doctrine/ORM/Repository/ApiResourceRepository.php
@@ -63,6 +63,7 @@ class ApiResourceRepository extends EntityRepository implements ApiResourceRepos
         }
 
         return $qb->setMaxResults($limit)
+            ->groupBy('r.id')
             ->getQuery()
             ->execute();
     }
--- a/src/Pim/Bundle/ApiBundle/Doctrine/ORM/Repository/AttributeRepository.php
+++ b/src/Pim/Bundle/ApiBundle/Doctrine/ORM/Repository/AttributeRepository.php
@@ -63,6 +63,7 @@ class AttributeRepository extends EntityRepository implements AttributeRepositor
         }
 
         return $qb->setMaxResults($limit)
+            ->groupBy('r.id')
             ->getQuery()
             ->execute();
     }
--- a/src/Pim/Bundle/ApiBundle/tests/integration/Controller/Product/ListProductWithCompletenessIntegration.php
+++ b/src/Pim/Bundle/ApiBundle/tests/integration/Controller/Product/ListProductWithCompletenessIntegration.php
@@ -1,0 +1,164 @@
+<?php
+
+namespace Pim\Bundle\ApiBundle\tests\integration\Controller\Product;
+
+use Akeneo\Test\Integration\Configuration;
+use Doctrine\Common\Collections\Collection;
+
+class ListProductWithCompletenessIntegration extends AbstractProductTestCase
+{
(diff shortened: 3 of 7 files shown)


src/Pim/Bundle/ApiBundle/Doctrine/ORM/Repository/ApiResourceRepository.php
```diff
@@ -63,6 +63,7 @@ public function searchAfterOffset(array $criteria, array $orders, $limit, $offse
         }
 
         return $qb->setMaxResults($limit)
+            ->groupBy('r.id')
```
phaseinducer: Shouldn't we use `$qb->getRootAliases()[0]` here?

src/Pim/Bundle/ApiBundle/Doctrine/ORM/Repository/AttributeRepository.php
```diff
@@ -63,6 +63,7 @@ public function searchAfterOffset(array $criteria, array $orders, $limit, $offse
         }
 
         return $qb->setMaxResults($limit)
+            ->groupBy('r.id')
```
phaseinducer: idem

src/Pim/Bundle/ApiBundle/tests/integration/Controller/Product/ListProductWithCompletenessIntegration.php
```diff
@@ -0,0 +1,157 @@
...
+
+        $this->products = $this->get('pim_catalog.repository.product')->findAll();
+    }
+
+    public function testPaginationWithCompletenessFilter()
+    {
+        $client = $this->createAuthenticatedClient();
+
+        $search = '{"completeness":[{"operator":"=","value":100,"scope":"ecommerce"}]}';
+        $client->request('GET', 'api/rest/v1/products?scope=ecommerce&locales=en_US&limit=2&search=' . $search);
```
phaseinducer: leading slash

src/Pim/Bundle/ApiBundle/Doctrine/ORM/Repository/ApiResourceRepository.php
ahocquard: Imho, this kind of group by is specific to mysql 5.5/5.6, because you do not specify all your columns of your select.
It will not pass Mysql 5.7 with "only_full_group_by" enable.

It's very strange to group by without applying any aggregation function of mysql (SUM, MEAN, etc). Group by is done for it.

skeleton: why ? alias is created on this method

skeleton: first, we do not support MySQL 5.7, and if a dev want to use this version, there is a warning on our documentation to disable this mode (see https://docs.akeneo.com/latest/developer_guide/installation/system_requirements/system_requirements.html)

second, about the `groupBy` without apply any aggregation, it's done like that on the pqb.
I can set a distinct instead, but I'm really not sure about perf

ahocquard: I say that because from a SQL point of view, it's not very readable. Using a Group by to delete row is a (bad) side effect of Mysql and it should not behave like this.

About the use of Mysql 5.7, it's like using a deprecated function and telling me that we use Symfony 2.7 for now ;)

+1 for the performance, but both function sort your data anyway.

PR #5946, opened 2017-03-17
PIM-6250: fix attribute export

solivier:

When you export attribute and reimport it, you got some errors.

Depending on the attribute type I have errors on these properties, for instance for SKU:
wysiwygEnabled: This value should be null.:
decimalsAllowed: This value should be null.:
negativeAllowed: This value should be null.:

As this values are not null but they should be null as they are not using the wysiwyg or decimals values. (like the text attribute)

### DIFF
--- a/src/Pim/Component/Connector/ArrayConverter/StandardToFlat/Attribute.php
+++ b/src/Pim/Component/Connector/ArrayConverter/StandardToFlat/Attribute.php
@@ -42,6 +42,11 @@ class Attribute extends AbstractSimpleArrayConverter implements ArrayConverterIn
                 $convertedItem[$property] = implode(',', $data);
                 break;
             case in_array($property, $this->booleanFields):
+                if (null === $data) {
+                    $convertedItem[$property] = '';
+                    break;
+                }
+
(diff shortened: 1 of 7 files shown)


src/Pim/Component/Connector/ArrayConverter/StandardToFlat/Attribute.php
```diff
@@ -42,6 +42,11 @@ protected function convertProperty($property, $data, array $convertedItem, array
                 $convertedItem[$property] = implode(',', $data);
                 break;
             case in_array($property, $this->booleanFields):
+                if (null === $data) {
```
anaelChardan: It's weird that we can have true/false/null on a boolean field :/

solivier: I agree but that's how the feature should work, if an attribute should not have a wysiwig field the value should be null not false or true.

anaelChardan: So sad :(

PR #5961, opened 2017-03-21
Fix #5941 Objects was not being detached on VariantGroupImport.

### LINKED ISSUE #5941: VariantGroupWriter Memory Leak on 1.5
**I'm reporting a Bug**

Version: 1.5.20 
Version: 1.6 (fixed already)

The variant group import wasn't detaching the objects and the memory usage was going up to 1Gb with just 100 variant groups being imported.

I've managed to run 4k variant groups and the memory kept in 218Mb.

@nidup helped to solve this issue where I will be PR soon.

nei:

Use DetachAll for VariantGroups.

It was taking more than 1Gb to process 200 variant groups. Now it is just 200Mb.

* Updated tests.

There is a failing test on the branch 1.5. @nidup


nidup:
And successful build! :tada: We had a small issue with CI build integration with Github, but the full build is ok. Thank you very much for the contribution!

PR #5981, opened 2017-03-22
PIM-6240: fix undefined channel on mass edit

solivier:

When the chosen channel doesn't has a scope for the selected locale it was displaying "undefined".
The fix now display the code instead of undefined.


a2xchip:
:tada:

PR #5984, opened 2017-03-23
Avoid to raise an exception in behats by testing the command in background

grena:
On the current CI, this behat passes but it raises an Exception (which is normal):

behat

Now the CI detects if there is an Exception, so it fails.
To avoid this, we launch the command in background and explicitly wait to test the file.

Pair fixed with @pierallard 💃

### DIFF
--- a/features/technical/security.feature
+++ b/features/technical/security.feature
@@ -6,5 +6,7 @@ Feature: Security
   @jira https://akeneo.atlassian.net/browse/PIM-6062
   Scenario: Fail when trying to inject command
     Given file "%tmp%/this_will_raise_an_error.txt" should not exist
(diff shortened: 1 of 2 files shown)


features/technical/security.feature
```diff
@@ -6,5 +6,7 @@ Feature: Security
   @jira https://akeneo.atlassian.net/browse/PIM-6062
   Scenario: Fail when trying to inject command
     Given file "%tmp%/this_will_raise_an_error.txt" should not exist
-    When I run '$(touch %tmp%/this_will_raise_an_error.txt)'
+    When I run '$(touch %tmp%/this_will_raise_an_error.txt)' in background
+    # This wait is exceptional, as otherwise we have an Exception if launched in foreground.
```
pierallard: WAIT IS FORBIDDEN

PR #5994, opened 2017-03-27
Add coupling detector

nidup:

Add coupling detector in Jenkins Build

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -103,6 +103,8 @@ if (launchUnitTests.equals("yes")) {
 
         tasks["php-cs-fixer"] = {runPhpCsFixerTest()}
 
+        tasks["php-coupling-detector"] = {runPhpCouplingDetectorTest()}
+
         tasks["grunt"] = {runGruntTest()}
 
         parallel tasks
@@ -306,6 +308,23 @@ def runBehatTest(edition, storage, features, phpVersion, mysqlVersion, esVersion
                 archiveArtifacts allowEmptyArchive: true, artifacts: 'app/build/screenshots/*.png'
                 deleteDir()
             }
+        }
+    }
+}
+
+def runPhpCouplingDetectorTest() {
+    node('docker') {
+        deleteDir()
+        try {
+            docker.image("carcel/php:7.1").inside("-v /home/akeneo/.composer:/home/akeneo/.composer -e COMPOSER_HOME=/home/akeneo/.composer") {
+                unstash "pim_community_dev"
+
+                sh "composer remove --dev --no-update doctrine/mongodb-odm-bundle;"
+                sh "composer update --ignore-platform-reqs --optimize-autoloader --no-interaction --no-progress --prefer-dist"
+                sh "./bin/php-coupling-detector detect --config-file=.php_cd src"
+            }
+        } finally {
+            deleteDir()
         }
     }
 }
--- a/composer.json
+++ b/composer.json
@@ -79,7 +79,7 @@
         "sensiolabs/behat-page-object-extension": "1.0.1",
         "phpspec/phpspec": "3.0.*",
         "akeneo/phpspec-skip-example-extension": "2.0.*",
-        "akeneo/php-coupling-detector": "dev-master"
+        "akeneo/php-coupling-detector": "0.1.0"
     },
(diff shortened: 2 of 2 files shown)


Jenkinsfile
```diff
@@ -309,3 +311,20 @@ def runBehatTest(edition, storage, features, phpVersion, mysqlVersion, esVersion
...
+
+def runPhpCouplingDetectorTest() {
+    node('docker') {
+        deleteDir()
+        try {
+            docker.image("carcel/php:7.1").inside("-v /home/akeneo/.composer:/home/akeneo/.composer -e COMPOSER_HOME=/home/akeneo/.composer") {
+                unstash "pim_community_dev"
+
+                sh "composer remove --dev --no-update doctrine/mongodb-odm-bundle;"
+                sh "composer update --ignore-platform-reqs --optimize-autoloader --no-interaction --no-progress --prefer-dist"
```
jjanvier: Why we need to update again the department?
Why can't we use what has been stashed by the master?

Jenkinsfile
```diff
@@ -309,3 +311,20 @@ def runBehatTest(edition, storage, features, phpVersion, mysqlVersion, esVersion
...
+                sh "./bin/php-coupling-detector detect --config-file=.php_cd src"
+            }
+        } finally {
```
jjanvier: Php 7 or 7.1 only no?

nidup: 7.1 only cf docker.image("carcel/php:7.1")

Jenkinsfile
nidup: I used the same stuff than for unit test, not sure it's mandatory

nidup: Checked with @damien-carcel as i don't need MongoDB in this step i can do this, when you need it you can add the alcaleus driver (only for PHP 7+)

composer.json
```diff
@@ -79,7 +79,7 @@
         "sensiolabs/behat-page-object-extension": "1.0.1",
         "phpspec/phpspec": "3.0.*",
         "akeneo/phpspec-skip-example-extension": "2.0.*",
-        "akeneo/php-coupling-detector": "dev-master"
+        "akeneo/php-coupling-detector": "0.2.0"
```
nidup: using the new release based on this merged branch

Jenkinsfile
jjanvier: sorry, it's java :/
tired this morning :/

Jenkinsfile
damien-carcel: @jjanvier because what was stashed was only for PHP 5.6. Vendors could change with PHP 7.x
Ideally, in the checkout section, at the beginning of the Jenkinsfile, we should do 3 composer update, one for each PHP version, then use those stashes in the test stages.
But I think this should be done in a dedicated PR.

composer.json
damien-carcel: Oh Yeah!
Maybe this should be done one other PIM versions too, as all from 1.5 to master, CE and EE use it, at least to get `php-cs-fixer`.

.php_cd.php
```diff
@@ -42,20 +42,22 @@
         ]),
         RuleInterface::TYPE_ONLY
     ),
-    new Rule('Akeneo\Component\Console', $cDeps, RuleInterface::TYPE_ONLY, 'have no real existence, should be merged into Batch'),
-    new Rule(
+/*    new Rule('Akeneo\Component\Console', $cDeps, RuleInterface::TYPE_ONLY, 'have no real existence, should be merged into Batch'),
```
mickaelandrieu: I'm not easy with commented code, can we remove it instead? :)

.php_cd.php
```diff
@@ -42,20 +42,22 @@
...
+      new Rule(
```
mickaelandrieu: same in all this file - except if you expect this code to work again /c @nidup

nidup: I expect to fix / re-enable it part by part ;)

nidup: But for now i would like to enforce it to be sure to not add violations :)

mickaelandrieu: 👍

PR #6043, opened 2017-04-07
[Contribution] Add subscriber to lock/unlock batch job commands

bOnepain:

Add the possibility to lock a batch job to avoid concurrency:

- Add a 'lock' attribute to the 'akeneo_batch.job' tag. It's not mandatory and false by default. Its value is boolean true or false. Batch job services tagged with the 'lock' attribute will be registered in a compiler pass to the LockSubscriber, that will handle locking and unlocking
- Add a 'no-lock' option the BatchCommand, to skip locking even if the job is configured to be locked.

| Added Specs                       | TODO

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Akeneo/Bundle/BatchBundle/Command/BatchCommand.php
+++ b/src/Akeneo/Bundle/BatchBundle/Command/BatchCommand.php
@@ -2,14 +2,14 @@
 
 namespace Akeneo\Bundle\BatchBundle\Command;
 
+use Akeneo\Component\Batch\Event\EventInterface;
 use Akeneo\Component\Batch\Item\ExecutionContext;
 use Akeneo\Component\Batch\Job\ExitStatus;
-use Akeneo\Component\Batch\Job\Job;
 use Akeneo\Component\Batch\Job\JobParametersFactory;
 use Akeneo\Component\Batch\Job\JobParametersValidator;
 use Akeneo\Component\Batch\Job\JobRegistry;
-use Akeneo\Component\Batch\Model\JobInstance;
 use Akeneo\Component\Batch\Model\StepExecution;
+use Akeneo\Component\Batch\Event\BatchCommandEvent;
 use Doctrine\ORM\EntityManager;
 use Monolog\Handler\StreamHandler;
 use Symfony\Bundle\FrameworkBundle\Command\ContainerAwareCommand;
@@ -62,6 +62,12 @@ class BatchCommand extends ContainerAwareCommand
                 null,
                 InputOption::VALUE_NONE,
                 'Don\'t display logs'
+            )
+            ->addOption(
+                'no-lock',
+                null,
+                InputOption::VALUE_NONE,
+                'Don\'t lock command (if command is lockable)'
             );
     }
 
@@ -84,6 +90,12 @@ class BatchCommand extends ContainerAwareCommand
             ->getJobManager()
             ->getRepository($jobInstanceClass)
             ->findOneByCode($code);
+
+        $event = new BatchCommandEvent($this, $input, $output, $jobInstance);
+        $this->getContainer()->get('event_dispatcher')->dispatch(EventInterface::BATCH_COMMAND_START, $event);
+        if (false === $event->commandShouldRun()) {
+            return self::EXIT_ERROR_CODE;
+        }
 
         if (!$jobInstance) {
             throw new \InvalidArgumentException(sprintf('Could not find job instance "%s".', $code));
@@ -163,6 +175,7 @@ class BatchCommand extends ContainerAwareCommand
         $job->getJobRepository()->updateJobExecution($jobExecution);
 
         $verbose = $input->getOption('verbose');
+        $exitCode = null;
         if (ExitStatus::COMPLETED === $jobExecution->getExitStatus()->getExitCode()) {
             $nbWarnings = 0;
             /** @var StepExecution $stepExecution */
@@ -184,7 +197,7 @@ class BatchCommand extends ContainerAwareCommand
                     )
                 );
 
-                return self::EXIT_SUCCESS_CODE;
+                $exitCode = self::EXIT_SUCCESS_CODE;
             } else {
                 $output->writeln(
                     sprintf(
@@ -195,7 +208,7 @@ class BatchCommand extends ContainerAwareCommand
                     )
                 );
 
-                return self::EXIT_WARNING_CODE;
+                $exitCode = self::EXIT_WARNING_CODE;
             }
         } else {
             $output->writeln(
@@ -209,8 +222,12 @@ class BatchCommand extends ContainerAwareCommand
                 $this->writeExceptions($output, $stepExecution->getFailureExceptions(), $verbose);
             }
 
-            return self::EXIT_ERROR_CODE;
+            $exitCode = self::EXIT_ERROR_CODE;
         }
+
+        $this->getContainer()->get('event_dispatcher')->dispatch(EventInterface::BATCH_COMMAND_TERMINATE, $event);
+
+        return $exitCode;
     }
 
     /**
--- a/src/Akeneo/Bundle/BatchBundle/DependencyInjection/Compiler/RegisterLockPass.php
+++ b/src/Akeneo/Bundle/BatchBundle/DependencyInjection/Compiler/RegisterLockPass.php
@@ -1,0 +1,42 @@
+<?php
+
+namespace Akeneo\Bundle\BatchBundle\DependencyInjection\Compiler;
+
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+
+/**
+ * Compiler pass to register locked jobs to the registry
+ *
+ * @author    Benoit Wannepain <bwannepain@kaliop.com>
+ * @copyright 2017 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php Open Software License (OSL 3.0)
+ */
+class RegisterLockPass implements CompilerPassInterface
+{
+    const SUBSCRIBER_ID = 'akeneo_batch.lock_subscriber';
+    const SERVICE_TAG = 'akeneo_batch.job';
+
+    /**
+     * {@inheritdoc}
+     */
+    public function process(ContainerBuilder $container)
+    {
+        if (!$container->hasDefinition(self::SUBSCRIBER_ID)) {
+            return;
+        }
+
+        $subscriberDefinition = $container->getDefinition(self::SUBSCRIBER_ID);
+        foreach ($container->findTaggedServiceIds(self::SERVICE_TAG) as $serviceId => $tags) {
+            foreach ($tags as $tag) {
+                if (isset($tag['lock']) && $tag['lock']) {
+                    $serviceDefinition = $container->getDefinition($serviceId);
+                    $subscriberDefinition->addMethodCall(
+                        'registerJobCode',
+                        [$serviceDefinition->getArgument(0)]
+                    );
+                }
+            }
+        }
+    }
+}
--- a/src/Akeneo/Bundle/BatchBundle/EventListener/LockSubscriber.php
+++ b/src/Akeneo/Bundle/BatchBundle/EventListener/LockSubscriber.php
@@ -1,0 +1,117 @@
+<?php
+
+
+namespace Akeneo\Bundle\BatchBundle\EventListener;
+
+use Akeneo\Component\Batch\Event\BatchCommandEvent;
+use Akeneo\Component\Batch\Event\EventInterface;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\Filesystem\LockHandler;
+
+/**
+ * Subscriber to lock and unlock commands
+ *
+ * @author    Benoit Wannepain <benoit.wannepain@kaliop.com>
+ * @copyright 2017 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php Open Software License (OSL 3.0)
+ */
+class LockSubscriber implements EventSubscriberInterface
+{
+    const AKENEO_BATCH_JOB = 'akeneo:batch:job';
+
+    /** @var array */
+    protected $managedJobCodes;
+
+    /** @var LockHandler */
+    protected $lockHandler;
+
+    /**
+     * LockSubscriber constructor.
+     */
+    public function __construct()
+    {
+        $this->managedJobCodes = [];
+    }
+
+    /**
+     * @param string $jobCode
+     */
+    public function registerJobCode($jobCode)
+    {
+        $this->managedJobCodes[] = $jobCode;
+    }
+
+    /**
+     * Returns an array of event names this subscriber wants to listen to.
+     *
+     * @return array The event names to listen to
+     */
+    public static function getSubscribedEvents()
+    {
+        return [
+            EventInterface::BATCH_COMMAND_START => 'onCommandStart',
+            EventInterface::BATCH_COMMAND_TERMINATE => 'onCommandTerminate',
+        ];
+    }
+
+    /**
+     * @param BatchCommandEvent $event
+     */
+    public function onCommandStart(BatchCommandEvent $event)
+    {
+        if (false === $this->supportsCommand($event)) {
+            return;
+        }
+
+        $output = $event->getOutput();
+        $this->lockHandler = new LockHandler(md5($this->getJobName($event)) . '.lock');
+        if (!$this->lockHandler->lock()) {
+            $output->writeln('<error>Command already locked</error>');
+            $event->disableCommand();
+        } else {
+            $output->writeln('<info>Command locked</info>');
+        }
+    }
+
+    /**
+     * @param BatchCommandEvent $event
+     */
+    public function onCommandTerminate(BatchCommandEvent $event)
+    {
+        if (false === $this->supportsCommand($event) || null === $this->lockHandler) {
+            return;
+        }
+
+        $this->lockHandler->release();
+        $output = $event->getOutput();
+        $output->writeln('<info>Command unlocked</info>');
+    }
+
+    /**
+     * @param BatchCommandEvent $event
+     * @return bool
+     */
+    protected function supportsCommand(BatchCommandEvent $event)
+    {
+        if ($event->getCommand()->getName() !== self::AKENEO_BATCH_JOB) {
+            return false;
+        }
+
+        $input = $event->getInput();
+        if ($input->hasOption('no-lock') && $input->getOption('no-lock')) {
+            return false;
+        }
+
+        return in_array($this->getJobName($event), $this->managedJobCodes);
+    }
+
+    /**
+     * @param BatchCommandEvent $event
+     * @return string
+     */
+    protected function getJobName(BatchCommandEvent $event)
+    {
+        return $event->getJobInstance()->getJobName();
+    }
+}
--- a/src/Akeneo/Component/Batch/Event/BatchCommandEvent.php
+++ b/src/Akeneo/Component/Batch/Event/BatchCommandEvent.php
@@ -1,0 +1,47 @@
+<?php
+
+
+namespace Akeneo\Component\Batch\Event;
+
+use Akeneo\Component\Batch\Model\JobInstance;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Event\ConsoleCommandEvent;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+
+/**
+ * @author    Benoit Wannepain <benoit.wannepain@kaliop.com>
+ * @copyright 2017 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php Open Software License (OSL 3.0)
+ */
+class BatchCommandEvent extends ConsoleCommandEvent
+{
+    /** @var JobInstance  */
+    private $jobInstance;
+
+    /**
+     * BatchCommandEvent constructor.
+     * @param Command $command
+     * @param InputInterface $input
+     * @param OutputInterface $output
+     * @param JobInstance $jobInstance
+     */
+    public function __construct(
+        Command $command,
+        InputInterface $input,
+        OutputInterface $output,
+        JobInstance $jobInstance
+    ) {
+        parent::__construct($command, $input, $output);
+
+        $this->jobInstance = $jobInstance;
+    }
+
+    /**
+     * @return JobInstance
+     */
+    public function getJobInstance()
+    {
+        return $this->jobInstance;
+    }
+}
--- a/src/Akeneo/Bundle/BatchBundle/AkeneoBatchBundle.php
+++ b/src/Akeneo/Bundle/BatchBundle/AkeneoBatchBundle.php
@@ -28,6 +28,7 @@ class AkeneoBatchBundle extends Bundle
             ->addCompilerPass(new Compiler\RegisterNotifiersPass())
             ->addCompilerPass(new Compiler\PushBatchLogHandlerPass())
             ->addCompilerPass(new Compiler\RegisterJobsPass())
+            ->addCompilerPass(new Compiler\RegisterLockPass())
             ->addCompilerPass(new Compiler\RegisterJobParametersPass('default_values_provider'))
             ->addCompilerPass(new Compiler\RegisterJobParametersPass('constraint_collection_provider'))
             ->addCompilerPass(
--- a/src/Akeneo/Bundle/BatchBundle/Resources/config/services.yml
+++ b/src/Akeneo/Bundle/BatchBundle/Resources/config/services.yml
@@ -2,6 +2,7 @@ parameters:
     akeneo_batch.job_instance_factory.class:                  Akeneo\Bundle\BatchBundle\Job\JobInstanceFactory
     akeneo_batch.logger_subscriber.class:                     Akeneo\Bundle\BatchBundle\EventListener\LoggerSubscriber
     akeneo_batch.notification_subscriber.class:               Akeneo\Bundle\BatchBundle\EventListener\NotificationSubscriber
+    akeneo_batch.lock_subscriber.class:                       Akeneo\Bundle\BatchBundle\EventListener\LockSubscriber
     akeneo_batch.logger.batch_log_handler.class:              Akeneo\Bundle\BatchBundle\Monolog\Handler\BatchLogHandler
     akeneo_batch.mail_notifier.class:                         Akeneo\Bundle\BatchBundle\Notification\MailNotifier
     akeneo_batch.set_job_execution_log_file_subscriber.class: Akeneo\Bundle\BatchBundle\EventListener\SetJobExecutionLogFileSubscriber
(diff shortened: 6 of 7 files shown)


SamirBoulil:
Hello @bOnepain  !

Thank you for your contribution! I'll make sure this is a feature we want in our PIM with the team.

I'll keep you updated regarding its decision very soon.

src/Akeneo/Bundle/BatchBundle/EventListener/LockSubscriber.php
```diff
@@ -0,0 +1,121 @@
+<?php
+
+
+namespace Akeneo\Bundle\BatchBundle\EventListener;
+
+
```
pierallard: Please avoid double empty lines

src/Akeneo/Bundle/BatchBundle/EventListener/LockSubscriber.php
```diff
@@ -0,0 +1,121 @@
...
+use Symfony\Component\Console\ConsoleEvents;
+use Symfony\Component\Console\Event\ConsoleCommandEvent;
+use Symfony\Component\Console\Event\ConsoleEvent;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\Filesystem\LockHandler;
+
+/**
+ * Subscriber to lock and unlock commands
+ *
+ * @Author Benoit Wannepain <benoit.wannepain@gmail.com>
```
pierallard: use `@author` and align with the 2 next lines please ;)

src/Akeneo/Bundle/BatchBundle/EventListener/LockSubscriber.php
```diff
@@ -0,0 +1,121 @@
...
+    {
+        $this->managedJobCodes[] = $jobCode;
+    }
+
+    /**
+     * Returns an array of event names this subscriber wants to listen to.
+     *
+     * @return array The event names to listen to
+     *
+     * @api
```
pierallard: You can drop this comment, we don't use it :)

src/Akeneo/Bundle/BatchBundle/EventListener/LockSubscriber.php
```diff
@@ -0,0 +1,121 @@
...
+        ];
+    }
+
+    /**
+     * @param ConsoleCommandEvent $event
+     */
+    public function onCommandStart(ConsoleCommandEvent $event)
+    {
+        if (false === $this->supportsCommand($event)) {
+
```
pierallard: Please remove this line

src/Akeneo/Bundle/BatchBundle/EventListener/LockSubscriber.php
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    /**
+     * @param ConsoleEvent $event
+     */
+    public function onCommandTerminate(ConsoleEvent $event)
+    {
+        if ($event->getExitCode() === ConsoleCommandEvent::RETURN_CODE_DISABLED
+            || false === $this->supportsCommand($event)) {
+
```
pierallard: Same !

src/Akeneo/Bundle/BatchBundle/EventListener/LockSubscriber.php
```diff
@@ -0,0 +1,121 @@
...
+            return;
+        }
+
+        $output = $event->getOutput();
+        $output->writeln(sprintf('<info>Command unlocked by %s</info>', get_class($this)));
+        $this->lockHandler->release();
+    }
+
+    /**
+     * @param ConsoleEvent $event
```
pierallard: And add a line between @param and @return
(I know, it's boring, it's just to respect our standards ;)

src/Akeneo/Bundle/BatchBundle/EventListener/LockSubscriber.php
```diff
@@ -0,0 +1,121 @@
...
+     * @return bool
+     */
+    protected function supportsCommand(ConsoleEvent $event)
+    {
+        $def = explode(' ', $event->getInput()->__toString());
+        if ($event->getInput()->hasOption('no-lock') || in_array('--no-lock', $def)) {
```
pierallard: I don't understand why you have to check these 2 cases ; in what case the second can happen ? 
Indeed, IMHO this could handle issues when you write some commands like `commandname --value=" no-lock "` or something like this

phaseinducer: Maybe it would be a good idea to check that the command is batch job command? (since we listen to generic events)

src/Akeneo/Bundle/BatchBundle/EventListener/LockSubscriber.php
```diff
@@ -0,0 +1,121 @@
...
+            return false;
+        }
+
+        if (isset($def[1])) {
+            $jobCode = md5($def[1]);
+            if (in_array($jobCode, $this->managedJobCodes)) {
+                $this->currentJobCode = $jobCode;
```
phaseinducer: No need to make the subscriber more stateful for that imho, since you can always retrieve the code in the event.

src/Akeneo/Bundle/BatchBundle/EventListener/LockSubscriber.php
```diff
@@ -0,0 +1,121 @@
...
+    {
+        if (false === $this->supportsCommand($event)) {
+
+            return;
+        }
+
+        $output = $event->getOutput();
+        $this->lockHandler = new LockHandler($this->currentJobCode);
+        if (!$this->lockHandler->lock()) {
+            $output->writeln(sprintf('<error>Command already locked by %s</error>', get_class($this)));
```
phaseinducer: The class is implementation detail, is it really useful to have it in the console output?

src/Akeneo/Bundle/BatchBundle/EventListener/LockSubscriber.php
```diff
@@ -0,0 +1,121 @@
...
+    {
+        if ($event->getExitCode() === ConsoleCommandEvent::RETURN_CODE_DISABLED
+            || false === $this->supportsCommand($event)) {
+
+            return;
+        }
+
+        $output = $event->getOutput();
+        $output->writeln(sprintf('<info>Command unlocked by %s</info>', get_class($this)));
+        $this->lockHandler->release();
```
phaseinducer: I know it's not supposed to happen, but if for one reason or another you pass here before passing in `onCommandStart()` you'll get a fatal error.
A simple solution would be to add `|| null === $this->lockHandler` in the check above.

bOnepain:
Hello,

I'll try to make all these changes this week, or this weekend. 
Just a question: would it be possible to use a semaphore to lock the command instead of a file ? I mean does Akeneo fulfill the PHP configuration requirements to use semaphores ?
Thanks

ahocquard:
@bOnepain the LockHandler use `flock`. Do you have any reason to use semaphore instead of that ?

src/Akeneo/Bundle/BatchBundle/DependencyInjection/Compiler/RegisterLockPass.php
```diff
@@ -0,0 +1,45 @@
...
+        }
+
+        $subscriberDefinition = $container->getDefinition(self::SUBSCRIBER_ID);
+        foreach ($container->findTaggedServiceIds(self::SERVICE_TAG) as $serviceId => $tags) {
+            foreach ($tags as $tag) {
+                if (isset($tag['lock']) && $tag['lock']) {
+                    $serviceDefinition = $container->getDefinition($serviceId);
+                    $subscriberDefinition->addMethodCall(
+                        'registerJobCode',
+                        [md5($serviceDefinition->getArgument(0))]
```
bOnepain: When ConsoleEvents::COMMAND event is dispatched, the InputInterface object has not yet been bound to the definition, meaning Arguments and Options array are not yet initialized. So we're are not able to get this information with ->getOption nor ->getArgument at that moment.

I make a md5 hash of the code in order to have a consistent string, in case some users give the command a weird code with weird characters...

src/Akeneo/Bundle/BatchBundle/EventListener/LockSubscriber.php
bOnepain: done

src/Akeneo/Bundle/BatchBundle/EventListener/LockSubscriber.php
bOnepain: done

src/Akeneo/Bundle/BatchBundle/EventListener/LockSubscriber.php
bOnepain: done

src/Akeneo/Bundle/BatchBundle/EventListener/LockSubscriber.php
bOnepain: done

src/Akeneo/Bundle/BatchBundle/EventListener/LockSubscriber.php
bOnepain: done

src/Akeneo/Bundle/BatchBundle/EventListener/LockSubscriber.php
bOnepain: done

(no problem respecting the standards)

src/Akeneo/Bundle/BatchBundle/EventListener/LockSubscriber.php
bOnepain: The problem is that options and arguments arrays are not available when tje ConsoleEvents::COMMAND event is dispatched. You're right for the possible issue, so I've the code to use $_SERVER instead. I haven't found a better way to do that.

@phaseinducer this subscriber may be used for other types of commands. It only has a list of supported commands, that is built during the compilation pass (RegisterLockPass).

src/Akeneo/Bundle/BatchBundle/DependencyInjection/Compiler/RegisterLockPass.php
```diff
@@ -0,0 +1,45 @@
...
+
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+
+/**
+ * Compiler pass to register locked jobs to the registry
+ *
+ * @author    Benoit Wannepain <bwannepain@kaliop.com>
+ * @copyright 2017 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/MIT MIT
```
bOnepain: done

src/Akeneo/Bundle/BatchBundle/DependencyInjection/Compiler/RegisterLockPass.php
bOnepain: Encoding the code to be sure to have a consistent filename without exotic character for example.

arguments and options arrays are not yet initialized when ConsoleEvents::COMMAND is dispatched.

src/Akeneo/Bundle/BatchBundle/EventListener/LockSubscriber.php
bOnepain: removed

src/Akeneo/Bundle/BatchBundle/EventListener/LockSubscriber.php
bOnepain: added

src/Akeneo/Bundle/BatchBundle/EventListener/LockSubscriber.php
```diff
@@ -0,0 +1,121 @@
...
+     * @return bool
+     */
+    protected function supportsCommand(ConsoleEvent $event)
+    {
+        $def = explode(' ', $event->getInput()->__toString());
+        if ($event->getInput()->hasOption('no-lock') || in_array('--no-lock', $def)) {
+            return false;
+        }
+
+        if (isset($def[1])) {
```
bOnepain: not available when ConsoleEvents::COMMAND is dispatched... 

Still I change the code a bit to make use of $_SERVER array

bOnepain:
@ahocquard the only reason is that with flock, the lock file is not really removed from the filesystem, but it's not a real problem either

bOnepain:
@ahocquard indeed interfacing would be a great idea. I hadn't enough time to do this. Further, I think such an abstraction layer will be integrated soon in Symfony (https://github.com/symfony/symfony/commit/a3078804efceed588fe537141f53da6dd5cc5c83)

src/Akeneo/Bundle/BatchBundle/EventListener/LockSubscriber.php
bOnepain: Indeed, I've changed the code so that it is stateless concerning this variable. But once again this information isn't available in the event yet when ConsoleEvents::COMMAND is dispatched, so I've used the $_SERVER array instead

ahocquard:
@bOnepain Nice, indeed, a Semaphore feature is coming in Symfony: https://github.com/symfony/symfony/commit/a3078804efceed588fe537141f53da6dd5cc5c83#diff-86462101f53a284b5b34cdd147a1c294R25

Unfortunately, I would like to see a PdoStore in the Symfony PR :(

But for now, I think that the LockHandler is enough, keeping the file is not a big deal, not need to use semaphore just for that imho.

src/Akeneo/Bundle/BatchBundle/DependencyInjection/Compiler/RegisterLockPass.php
phaseinducer: Oh ok, so maybe encode it only when you create the lock?

src/Akeneo/Bundle/BatchBundle/EventListener/LockSubscriber.php
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    /**
+     * @param ConsoleEvent $event
+     *
+     * @return bool
+     */
+    protected function supportsCommand(ConsoleEvent $event)
+    {
+        if (in_array('--no-lock', $_SERVER['argv'])) {
```
phaseinducer: cf my previous comment:
> Maybe it would be a good idea to check that the command is batch job command? (since we listen to generic events)

bOnepain: As I've explained, this subscriber has a array of supported commands, that is built during compilation pass.

src/Akeneo/Bundle/BatchBundle/EventListener/LockSubscriber.php
```diff
@@ -0,0 +1,112 @@
...
+            return false;
+        }
+
+        return in_array($this->getJobCode(), $this->managedJobCodes);
+    }
+
+    protected function getJobCode()
```
damien-carcel: PHPDoc ;-)

src/Akeneo/Bundle/BatchBundle/EventListener/LockSubscriber.php
```diff
@@ -0,0 +1,112 @@
...
+    {
+        if (isset($_SERVER['argv'][2])) {
```
damien-carcel: Please do not use $_SERVER here. You cannot be sure the third entry of `argv` contains the job code. For example, if I run the command `app/console -e=prod akeneo:batch:job whatever`, `$_SERVER['argv'][2]` will contain `akeneo:batch:job` instead of `whatever` as you expect.

You should use the content of the `$event` variable to retrieve the code: `$event->getInput()->getArgument('code')`.

However, those elements are only accessible the second time the Subscriber is called (no idea why it is called two times by the `akeneo:batch:job` command, it is something I need to investigate…). So you need to first test that `$event->getInput()->hasArgument('code')` do not return false.

src/Akeneo/Bundle/BatchBundle/EventListener/LockSubscriber.php
```diff
@@ -0,0 +1,112 @@
...
+            return md5($_SERVER['argv'][2]);
+        }
+    }
```
damien-carcel: Please return an empty string if there is no code in the input arguments.

src/Akeneo/Bundle/BatchBundle/EventListener/LockSubscriber.php
damien-carcel: Like for the job code below, please do not use `$_SERVER`.

You can access this parameter with `$event->getInput()->getOption('no-lock')`, and again after ensuring you indeed have some the option with `$event->getInput()->hasOption('no-lock')`.

src/Akeneo/Bundle/BatchBundle/EventListener/LockSubscriber.php
```diff
@@ -0,0 +1,112 @@
...
+        $output->writeln('<info>Command unlocked</info>');
+    }
+
+    /**
+     * @param ConsoleEvent $event
+     *
+     * @return bool
+     */
+    protected function supportsCommand(ConsoleEvent $event)
+    {
```
damien-carcel: Please first check that `$event->getCommand()->getName()` returns `akeneo:batch:job` as it is a generic console event. If the command is not the batch job command, no need to go further :-).

damien-carcel:
Hello again @bOnepain .

I also forgot to provide you our [contributor agreement](https://www.akeneo.com/contributor-license-agreement/). We need you to read and sign it to be able to merge your contribution 😃.

Regards,
Damien

bOnepain:
Hi everyone,
Thank you for your feedback, it shows that you're doing your job quite seriously, and it's a warranty of quality :+1: .

Now all the problems we're talking about are due to the fact that the ArgvInput instance has no options initialized when the ConsoleEvents::COMMAND event is dispatched, so it's not possible to get the job code/name at that moment (what made me write filthy things I must admit).

I'm quite new to Akeneo, but I had a look at the code and I think I have two possibilities:

1. I could make use of the existing hooks in \Akeneo\Component\Batch\Job\Job::execute. But we would lose the possibility to specify a --no-lock option on the command line.
2. I could hack \Akeneo\Bundle\BatchBundle\Command\BatchCommand::execute by dispatching two new events (to be created). When this method is executed, the ArgvInput instance has its options initialized so it should do the trick (and we keep the possibility to add a --no-lock option on the command line).

What do you think of that ?

gplanchat:
Can't the status and lock identifier be added to the `JobExecution` object ?

bOnepain:
@gplanchat maybe it could be possible I don't know...
But for the moment I prefer to explore the second possibility (hacking the BatchCommand), because it happens earlier in the lifecycle of the batch process. I'd prefer the locking been done as early as possible, especially before any db request, so before the Job instance is fetched from the db.
So the idea is to dispatch an event for locking the command between those two instructions:
`$code = $input->getArgument('code');
        $jobInstance = $this
            ->getJobManager()
            ->getRepository('Akeneo\Component\Batch\Model\JobInstance')
            ->findOneByCode($code);`

and to refactor a little bit the if/else structure in order to dispatch another event to unlock the command.

This is only to lock/unlock the command, and should have no impact on the current lifecycle of the command (success and exceptions management).

src/Akeneo/Bundle/BatchBundle/Command/BatchCommand.php
```diff
@@ -161,17 +172,12 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
         $job->getJobRepository()->updateJobExecution($jobExecution);
 
-        $verbose = $input->getOption('verbose');
+        $exitCode = null;
         if (ExitStatus::COMPLETED === $jobExecution->getExitStatus()->getExitCode()) {
             $nbWarnings = 0;
             /** @var StepExecution $stepExecution */
             foreach ($jobExecution->getStepExecutions() as $stepExecution) {
                 $nbWarnings += count($stepExecution->getWarnings());
-                if ($verbose) {
```
damien-carcel: Please do not remove that, we are using this to display warning in the PIM imports/exports.

damien-carcel: Please do not remove that, we are using this to display warning in the PIM imports/exports.

src/Akeneo/Bundle/BatchBundle/Command/BatchCommand.php
```diff
@@ -84,6 +89,12 @@ protected function execute(InputInterface $input, OutputInterface $output)
             ->getRepository('Akeneo\Component\Batch\Model\JobInstance')
             ->findOneByCode($code);
 
+        $event = new BatchCommandEvent($this, $input, $output, $jobInstance);
+        $this->getContainer()->get('event_dispatcher')->dispatch(EventInterface::BATCH_COMMAND_START, $event);
```
phaseinducer: I don't see where this event name is declared.

src/Akeneo/Component/Batch/Event/BatchCommandEvent.php
```diff
@@ -0,0 +1,48 @@
+<?php
+
```
phaseinducer: extra line

src/Akeneo/Component/Batch/Event/BatchCommandEvent.php
```diff
@@ -0,0 +1,48 @@
...
+
+namespace Akeneo\Component\Batch\Event;
+
+
```
phaseinducer: idem

src/Akeneo/Component/Batch/Event/BatchCommandEvent.php
```diff
@@ -0,0 +1,48 @@
...
+ * @copyright 2017 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php Open Software License (OSL 3.0)
+ */
+class BatchCommandEvent extends ConsoleCommandEvent
+{
+    /** @var JobInstance  */
+    private $jobInstance;
+
+    /**
+     * BatchCommandEvent constructor.
```
phaseinducer: useless

src/Akeneo/Component/Batch/Event/BatchCommandEvent.php
```diff
@@ -0,0 +1,48 @@
...
+     * @param Command $command
+     * @param InputInterface $input
+     * @param OutputInterface $output
+     * @param JobInstance $jobInstance
+     */
+    public function __construct(
+        Command $command,
+        InputInterface $input,
+        OutputInterface $output,
+        JobInstance $jobInstance
```
phaseinducer: What's the added value of embedding the job instance since you use it only to retrieve its name? You already have it in the input no?

src/Akeneo/Bundle/BatchBundle/DependencyInjection/Compiler/RegisterLockPass.php
phaseinducer: You don't need to register the encoded code anymore.

phaseinducer:
Looks like the big drawback compared to your previous solution is that if you get an exception during the command execution you will get a deadlock, or did I miss something?

src/Akeneo/Bundle/BatchBundle/Command/BatchCommand.php
bOnepain: Yes, I also noticed these lines had vanished, surely inattention. First thought rebasing on master would solve...

I add them back

src/Akeneo/Bundle/BatchBundle/Command/BatchCommand.php
bOnepain: done

src/Akeneo/Bundle/BatchBundle/Command/BatchCommand.php
bOnepain: done

src/Akeneo/Bundle/BatchBundle/EventListener/LockSubscriber.php
bOnepain: done

ahocquard:
@phaseinducer the process will release the lock if it's not released at the end of the script (source:from https://stackoverflow.com/questions/13522273/will-flocked-file-be-unlocked-when-the-process-die-unexpectedly).

src/Akeneo/Component/Batch/Event/BatchCommandEvent.php
bOnepain: You mean 2 empty lines ?

bOnepain:
In order to go further, I need to build a new Akeneo instance on my home computer to be able to test a little bit (I was coding blind until now), but I'm facing some configuration problems with ElasticSearch... I get a "No alive nodes found in your cluster" error message while running pim:install. As it's the first time I'm configuring an elasticsearch container, it may take some time to solve (but maybe you might help me :) )

src/Akeneo/Component/Batch/Event/BatchCommandEvent.php
bOnepain: I have to test on a running instance (I have to get that elastic thing working...). As far as I can remember, I get the job code in the input... I'm not sure to have the job name.

damien-carcel:
Hello @bOnepain .

They can be two reasons: Elasticsearch is not running, or your PIM is miss-configured.

To know if it is running, you can try to access `localhost:9200` or `127.0.0.1:9200`, and you should obtain in your browser something like the screenshot below.
image

If this works, then it means your PIM is miss-configured. You need to add this new line in your `app/config/parameters.yml` file:
```yaml
index_hosts: 'localhost: 9200'
```

We will soon update the documentation for `master` branch with explanation on how to deploy this new stack. A pull request is in progress and will soon be merged: https://github.com/akeneo/pim-docs/pull/611.

Regards,
Damien

bOnepain:
@damien-carcel thanks for the tip !

I also had to had a few environment variables in compose file. For information:
`environment:
          - cluster.name=akeneo
          - node.name=akeneo_pim
          - bootstrap.memory_lock=true
          - http.host=0.0.0.0
          - transport.host=127.0.0.1
          - "ES_JAVA_OPTS=-Xms512m -Xmx512m"
          # disable xpack
          - xpack.security.enabled=false
          - xpack.monitoring.enabled=false
          - xpack.graph.enabled=false
          - xpack.watcher.enabled=false`

damien-carcel:
Oh, my bad! I completely forgot another parameter: `index_name: akeneo_pim` to put in `app/config/parameters.yml`.
That defines the index name, so you should not have to use
```yaml
- cluster.name=akeneo
- node.name=akeneo_pim
```
in your environment variables.

Sorry for that.

Btw, remember to perform a `composer update`, as we update all Doctrine stack in our dependencies when adding Elasticsearch.

Regards,
Damien

momoss:
@phaseinducer Is ok for you now?

src/Akeneo/Component/Batch/Event/BatchCommandEvent.php
bOnepain: I've checked and that's what I said above: the input holds the job code, not the job name. 
As far as I understand Akeneo, the job code can be different from the job name, that is defined in the service definition read by the LockCompilerPass and registered at LockSubscriber. So I'm not sure I can get rid of the JobInstance.

bOnepain:
I've squashed the commits... tested a console batch command. Seems ok... ready for CI if OK for you all.

momoss:
Hello @bOnepain,

We are waiting for @phaseinducer feedback normally tomorrow.

Regards

bOnepain:
Hello, any news ?

tamarasaurus:
Hi @bOnepain, your contribution has been tested in the CI and has been merged in this PR - https://github.com/akeneo/pim-community-dev/pull/6423

Thanks so much for your contribution! Don't forget to claim your badge "Core contributor" on http://badger.akeneo.com

Issue #6061, opened 2017-04-13
1.7 CSS Sub-menu size


naolis:
Please adapt the menu size for bigger Text in the new CSS. 
Ping @pierallard 
<img width="306" alt="capture d ecran 2017-04-13 a 12 01 03" src="https://cloud.githubusercontent.com/assets/6884638/25003921/005e4bf0-2052-11e7-820d-ddd20cd599c4.png">

pierallard:
Thanks for raising this glitch Aymeric ! PR is done, waiting for reviews ;)

pierallard:
Merged ! Thanks again!

PR #6067, opened 2017-04-14
[GITHUB-6061] Fix menu display for big words

pierallard:
Solves this issue:
https://github.com/akeneo/pim-community-dev/issues/6061

No CI needed.

screenshot from 2017-04-14 10 10 31

### DIFF
--- a/src/Pim/Bundle/UIBundle/Resources/public/less/components/MainMenu.less
+++ b/src/Pim/Bundle/UIBundle/Resources/public/less/components/MainMenu.less
@@ -25,6 +25,7 @@
     display: block;
     color: white;
     transition: background 0.1s ease;
+    white-space: nowrap;
   }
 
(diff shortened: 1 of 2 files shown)


src/Pim/Bundle/UIBundle/Resources/public/less/components/MainMenu.less
```diff
@@ -39,8 +40,8 @@
   }
 
   &-level2 {
-    width: @submenuWidth;
```
tamarasaurus: does it have a max-width?

pierallard: No need of max width, it's a menu, if someone adds a title long of 1000px, it's its responsability (all menus should be very short IMHO)

PR #6090, opened 2017-04-21
Remove loop flag in favour of break statement

pamil:


nidup:
thx for your pull request @pamil :tada:

PR #6103, opened 2017-04-24
PIM-6377: add sanity check in price property formatter

LaurentPetard:

Price field does not appear on the variation datagrid.

It happens only for MongoDB storage, when there are no property data set for a price.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Pim/Bundle/DataGridBundle/Extension/Formatter/Property/ProductValue/PriceProperty.php
+++ b/src/Pim/Bundle/DataGridBundle/Extension/Formatter/Property/ProductValue/PriceProperty.php
@@ -45,8 +45,12 @@ class PriceProperty extends FieldProperty
         $data = $this->getBackendData($value);
 
         foreach ($data as $index => $price) {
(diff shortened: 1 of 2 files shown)


src/Pim/Bundle/DataGridBundle/Extension/Formatter/Property/ProductValue/PriceProperty.php
```diff
@@ -45,8 +45,12 @@ protected function convertValue($value)
         $data = $this->getBackendData($value);
 
         foreach ($data as $index => $price) {
-            $data[$index]['amount'] = $price['data'];
-            unset($price['data']);
+            if (array_key_exists('data', $price)) {
```
ahocquard: or ternary, up to you

LaurentPetard: I thougth that it would be more readable to do the "unset" in an "if" even if it's not necessary

PR #6104, opened 2017-04-24
PIM-6314: Fix parameters inversion

phaseinducer:

Parameters order was wrong.

`OK`: Done / Validated
`-`: Not needed


damien-carcel:
Good point. As both are string, spec are probably still OK, but it is "moche".

phaseinducer:
It looks like it has never been speced. But it's a deprecated method, I won't add spec for this.

PR #6107, opened 2017-04-25
PIM-6387: fix HTTP code returned when the token is invalid or expired

LaurentPetard:

When requesting the API, when the token has expired or is invalid, the HTTP status returned should be 401 instead of 422.

This is done by throwing an "UnauthorizedHttpException" during the verification of the access token. This exception need a parameter to build the HTTP header "WWW-Authenticate" which is fetched by the OAuth exception.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Pim/Bundle/ApiBundle/Security/OAuth2.php
+++ b/src/Pim/Bundle/ApiBundle/Security/OAuth2.php
@@ -4,7 +4,8 @@ namespace Pim\Bundle\ApiBundle\Security;
 
 use OAuth2\OAuth2 as BaseOAuth2;
 use OAuth2\OAuth2AuthenticateException;
-use Symfony\Component\HttpKernel\Exception\UnprocessableEntityHttpException;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\HttpKernel\Exception\HttpException;
 
 /**
  * @author    Marie Bochu <marie.bochu@akeneo.com>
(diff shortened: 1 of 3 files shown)


src/Pim/Bundle/ApiBundle/Security/OAuth2.php
```diff
@@ -21,7 +21,10 @@ public function verifyAccessToken($tokenParam, $scope = null)
         try {
             return parent::verifyAccessToken($tokenParam, $scope);
         } catch (OAuth2AuthenticateException $e) {
-            throw new UnprocessableEntityHttpException($e->getDescription(), $e);
+            $headers = $e->getResponseHeaders();
+            $challenge = isset($headers['WWW-Authenticate']) ? $headers['WWW-Authenticate'] : '';
```
ahocquard: is 'WWW-Authenticate' set in our case, or always not set ? Maybe it would be simpler to create directly the HttpException : 
```
throw new HttpException(401, $e->getDescription(), $e);
```

phaseinducer: +1 with @ahocquard.
`UnauthorizedHttpException` is useful when you're in a "browser to server" context and want the client's browser to prompt the login/password. We don't use that in the API.

LaurentPetard: You're right, it's useless for an API. It's fixed

PR #6112, opened 2017-04-26
Rework fixtures loading in integration tests & process isolation

phaseinducer:
## The current situation
When we started to write integration tests we chose to set up fixtures using the installer. It was kind of a quick win, and it was the guarantee to have exactly what we want in the database.
The major drawback was that running all import jobs is a quite long process, so we couldn't afford to run it for each test. We usually run it only at the beginning of a test case (class).

Consequences:
- Some tests that modify the data in database are heavily coupled.
- It's almost impossible to run the tests in `process-isolation` mode because it's waaaay too slow.
- The test case classes are stateful and difficult to maintain.
- We need to write homemade components to clean only specific data (e.g. permissions).

## The proposed solution
The main idea here is to **drop the database at the beginning of each test**. Then, since installing the fixtures via the installer is slow, have some sort of "cache" (a database dump) created on-the-fly after the first installer run. What typically happens is:
1. test A starts with catalog "technical"
2. the installer runs
3. the dump is created for catalog "technical"
4. test A finishes
5. test B starts with catalog "technical"
6. a dump is found so the installer is not run, the db is dropped and the dump is loaded

The advantages are:
- We can now run the tests in `process-isolation` mode, **which fixes "segmentation fault" errors** and other kind of surprises, like data kept in static class properties. It also guarantee that tests are really independent (it also means more parallelizable in the future).
- When you write tests you don't have to worry anymore about "should I reset the db for each test?", all tests have the same behavior now. The only thing that can still be configurable by test case is which catalog to use.
- The code is simpler and more maintainable.
- Still no SQL fixtures to maintain.
- Dumps are really like a cache, the tests will still run correctly if you disable it (it will just be super slow) and it's still possible to run them without the `process-isolation` option if you need to.
- It's faster. For example on the last 1.7 nightly build, integration tests ran in 36 minutes, on this PR they ran in 18 minutes. And it can still be speeded up if needed imho.

## Missing
- MongoDB implementation (but we don't run integration tests with it for the moment)
- Dump files are never cleaned. Any idea on who should do it and when?

PR for EE coming soon, after the principle is ok for everybody.

### DIFF
--- a/tests/FixturesLoader.php
+++ b/tests/FixturesLoader.php
@@ -8,6 +8,8 @@ use Symfony\Component\Console\Input\ArrayInput;
 use Symfony\Component\Console\Output\BufferedOutput;
 use Symfony\Component\DependencyInjection\ContainerInterface;
 use Symfony\Component\HttpKernel\KernelInterface;
+use Symfony\Component\Process\Exception\ProcessFailedException;
+use Symfony\Component\Process\Process;
 
 /**
  * @author    Yohan Blain <yohan.blain@akeneo.com>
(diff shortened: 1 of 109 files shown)


Jenkinsfile
```diff
@@ -232,7 +232,7 @@ def runIntegrationTest(phpVersion, storage, testSuiteName) {
                         sh "./app/console --env=test pim:install --force"
 
                         sh "mkdir -p app/build/logs/"
-                        sh "./bin/phpunit -c app/phpunit.xml.dist --testsuite ${testSuiteName} --log-junit app/build/logs/phpunit_integration.xml"
+                        sh "./bin/phpunit -c app/phpunit.xml.dist --testsuite ${testSuiteName} --log-junit app/build/logs/phpunit_integration.xml --process-isolation"
```
skeleton: to have the same behavior between CI and local, we should configure `process-isolation` in app/phpunit.xml.dist` file

damien-carcel:
> Dump files are never cleaned. Any idea on who should do it and when?
I don't think it is a big problem. On the CI, everything is trashed between each builds.
On local machine, it can be the responsibility of the developer (as long as it is clearly explained and, ideally, documented).

Jenkinsfile
damien-carcel: 👏 Oh yeah!!!

tests/FixturesLoader.php
```diff
@@ -56,12 +70,43 @@ public function __destruct()
     public function load()
     {
         $files = $this->getFilesToLoad($this->configuration->getCatalogDirectories());
```
ahocquard: If you modify the reference data, the hash will be the same, no? It could be problematic in local.

phaseinducer: Discussed IRL, it's a very little edge case.

PR #6115, opened 2017-04-27
Be able to configure installer data from environment

mickaelandrieu:
Hello,

this is an extract of https://github.com/akeneo/pim-community-dev/pull/5871.

The idea is be able to select the ``installer_data`` without have to edit a configuration file. This is also a good practice as stated in the 12 factors app manifesto (https://12factor.net/config).

What do you think?

```bash
~$: PIM_INSTALLER_DATA=PimInstallerBundle:minimal
~$: php app/console pim:install # will use the minimal set of fixtures instead of Icecat
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -136,7 +136,8 @@
                 "database_port": "PIM_DATABASE_PORT",
                 "database_name": "PIM_DATABASE_NAME",
                 "database_user": "PIM_DATABASE_USER",
-                "database_password": "PIM_DATABASE_PASSWORD"
+                "database_password": "PIM_DATABASE_PASSWORD",
+                "installer_data": "PIM_INSTALLER_DATA"
             }
         },
         "branch-alias": {


jmleroux:
The idea is interesting, but i don't feel like the installer data location should be an environment variable.
Why not make it an argument or option of the `pim:install` command?

mickaelandrieu:
@jmleroux I had this idea too, but if you take a look at `pim:install` you will see it's not managed in it but indirectly in the ``pim:installer:db`` through the fixture loader command that use the `installer_data` parameter.

So if I wanted to change this behavior, I'm forced to do a BC because I can't rely on this parameter anymore and I can't override the fixture loader after the container is frozen. More I'll make an hard bound to `pim:installer:db` into `pim:install`.

The idea with 12 factors app is that you shouldn't have to change anything in your code to change your app configuration.

Of course, if you don't use the environment variable, nothing change! We can already do that to setup the database configuration too 👍

jmleroux:
Ok, i see your point and the argumentation is very valuable.
Lets see what others have to say.

anaelChardan:
Totally agree with this PR, this is the Symfony way now to do that :)

gplanchat:
I +1 too, doing this will help to maintain development/test installer files and help maintenance when using a CI, but not limited to

damien-carcel:
@mickaelandrieu Remember to reproduce this PR on EE, and on standard editions (CE and EE) too.

damien-carcel:
(and of course to run a CI to be sure install is fine :D).

mickaelandrieu:
I will as soon as this pull request will be merged :) I can't help about EE stuff by the way :D

PR #6137, opened 2017-05-02
PIM-6395: Fix MongoDB query built to fetch products in a LazyCollection

LaurentPetard:

The problem occurs when the method "getProducts()" is called on an entity "Category" or "Group" to fetch products which are in several categories or groups.

The request built to fetch products in MongoDB on a lazy loading way is wrong. The value of the parameter used to find products must not be embedded in an array. Otherwise all the elements of the parameter should be setted, which it's not the desired behavior.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/CHANGELOG-1.5.md
+++ b/CHANGELOG-1.5.md
@@ -1,3 +1,9 @@
+# 1.5.x
+
+## Bug fixes
+
+- PIM-6395: Fix MongoDB query built to fetch products in a LazyCollection
+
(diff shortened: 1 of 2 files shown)


LaurentPetard:
No test in the spec because the call of the corrected line is done in a closure which is not directly executed. And it would be very complicated to test it in Behat because the modified method is not directly called by UI proccesses.

PR #6158, opened 2017-05-11
Pim-6376: Fix slow MongoDB queries when editing attribute options

LaurentPetard:

Update option values of attributes of type "multi-select" (from import or UI) generates slow queries on MongoDB, up to trigger timeout for big catalogs.

The problem is that the query which update products in MongoDB use the operator `$elemMatch` on a attribute without index. To add a index on each attribute of type "multi-select" is not the solution, so I used another approach.

First, I added a index on the attribute `values.optionIds` to use this one as the main condition of the update query (to retrieve all products having an option value in particular). But this doesn't allow to retrieve precisely the option value to update, because the option values are stored in an numerical array for the attributes of type "multi-select".

So, I split the query on two queries. The first add a new element with the updated value in the option values. The second remove the element having the previous value. This cannot be done in a single query on MongoDB.

I addition to fhis, I noticed that in `UpdateNormalizedProductDataSubscriber` the scheduled queries built to update products in MongoDB are not purged after they are executed. So, during the complete process of an attribute options import job, three events "postFlush" are sent, and then all of this queries are exectued three times. I fix that in the first commit.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Pim/Bundle/CatalogBundle/Doctrine/MongoDBODM/QueryGenerator/MultipleOptionValueUpdatedQueryGenerator.php
+++ b/src/Pim/Bundle/CatalogBundle/Doctrine/MongoDBODM/QueryGenerator/MultipleOptionValueUpdatedQueryGenerator.php
@@ -2,7 +2,9 @@
 
 namespace Pim\Bundle\CatalogBundle\Doctrine\MongoDBODM\QueryGenerator;
 
+use Pim\Bundle\CatalogBundle\Doctrine\MongoDBODM\NamingUtility;
 use Pim\Component\Catalog\AttributeTypes;
+use Symfony\Component\Serializer\Normalizer\NormalizerInterface;
 
 /**
  * Option value updated query generator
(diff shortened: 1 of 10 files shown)


src/Pim/Bundle/CatalogBundle/Doctrine/MongoDBODM/QueryGenerator/MultipleOptionValueUpdatedQueryGenerator.php
```diff
@@ -23,22 +45,36 @@ public function generateQuery($entity, $field, $oldValue, $newValue)
         );
 
         $queries = [];
+        $attributeOptionNormalized = $this->attributeOptionNormalizer->normalize($entity->getOption());
```
ahocquard: @jjanvier passing it with null in the constructor is generally the way we are are using to avoid taht (not 100% safe though). Could you clarify why this rule does not apply here?

jjanvier: sorry, I was wrong
forget my comment

src/Pim/Bundle/CatalogBundle/Doctrine/MongoDBODM/QueryGenerator/MultipleOptionValueUpdatedQueryGenerator.php
```diff
@@ -23,22 +45,36 @@ public function generateQuery($entity, $field, $oldValue, $newValue)
...
 
         foreach ($attributeNormFields as $attributeNormField) {
+            $optionId = (int) $entity->getOption()->getId();
```
ahocquard: put it outside of the foreach

Chemoul:
It's better yes, it takes now less than 10 minutes
Thanks @LaureBro

Issue #6086, opened 2017-04-20
Adding attributes to family from >20th group


adamrandazzo:
This may be related to issue #6054

Version 1.7.3

When adding attributes to an attribute family, any attribute that is in >20th attribute group in the system by sort_order fails and throws a javascript error.

`attributes.js:109 Uncaught TypeError: Cannot read property 'sort_order' of undefined
    at child.<anonymous> (attributes.js:109)
    at underscore.js:380
    at Function._.map._.collect (underscore.js:172)
    at Function._.sortBy (underscore.js:376)
    at child.<anonymous> (attributes.js:108)
    at Array.<anonymous> (jquery-1.10.2.js:3204)
    at fire (jquery-1.10.2.js:3048)
    at Object.add [as done] (jquery-1.10.2.js:3094)
    at Array.<anonymous> (jquery-1.10.2.js:3203)
    at Function.each (jquery-1.10.2.js:657)`

As an example:

![screen shot 2017-04-20 at 8 25 05 am](https://cloud.githubusercontent.com/assets/8340768/25231340/874431fe-25a5-11e7-9fc5-90a61eafe8fa.png)

Here is the data dump from pim_catalog_attribute_group

![screen shot 2017-04-20 at 8 31 44 am](https://cloud.githubusercontent.com/assets/8340768/25231365/a2035074-25a5-11e7-8e43-9db0028a6833.png)

Thanks!

adamrandazzo:
Changing the FETCH_LIMIT in the following file fixes the issue, but I don't want to create a pull request as I am not sure of other implications in the system. https://github.com/akeneo/pim-community-dev/blob/494b2c108427556e2818a3c2f18ede40b585c144/src/Akeneo/Component/StorageUtils/Repository/SearchableRepositoryInterface.php

LaureBro:
Hi @adamrandazzo, thanks for this feedback, our support team will take a look at it and keep you updated :)

danielsan80:
Hi.
We have the same issue.
The problem is at Pim\Bundle\EnrichBundle\Controller\Rest\AttributeGroupController:76.

```
if (empty($options)) {
    $options = $request->request->get(
        'options',
        [
            'limit' => SearchableRepositoryInterface::FETCH_LIMIT,
            'locale' => null,
        ]
    );
}
```
The javascript application asks for the attribute groups to this controller.
It expects to retrieve ALL attribute groups but the controller limits the result to 20 items if the 'limit' parameter is not set.

[...]

Issue #6054, opened 2017-04-11
Limit of 20 attribute-groups product edit form in 1.7


patrickzzz:
Hello everyone..

I'm reporting a bug on 1.7:

The product-edit-form only shows first 20 attribute-groups.. 

This is due to a new limit on the rest AttributeGroupController:
See: https://github.com/akeneo/pim-community-dev/blob/master/src/Pim/Bundle/EnrichBundle/Controller/Rest/AttributeGroupController.php#L76

Thanks for checking!

mickaelandrieu:
Hello @patrickzzz,

you're right we have identified the issue and are actually working on a patch (https://github.com/akeneo/pim-community-dev/pull/5957).

Thank you for your help!

a2xchip:
@patrickzzz @mickaelandrieu Sorry i was not reporting it anywhere, because this bug appeared after my own changes and i fixed it inside other bug ~related to the same changes.

mickaelandrieu:
@a2xchip do you confirm the issue is fixed on master branch, or can you provide the link to the correct pull request?

a2xchip:
https://github.com/akeneo/pim-community-dev/pull/5957 here it is. it is not fixed on master... i think after 1.7 it is going to be transmitted to master

mickaelandrieu:
Thanks for information Alex!

a2xchip:
@mickaelandrieu Keep you updated

tamarasaurus:
Hi @patrickzzz, this issue has been fixed in v1.7.5 (available 2017-06-02)

PR #6175, opened 2017-05-15
PIM-6146: Fix javascript error when having more than 20 attributes groups in the family page

LaurentPetard:

A javascript error occurs in the family page when there are more than 20 attribute groups, and another attribute is added through attribute groups.

This javascript function that display the attributes try to fetch all the attribute groups. But without criteria the server set a default limit to the result. The solution is to fetch only the attribute codes we need using their codes.

By fixing this bug I realize that the sorting of the attribute groups is broken. This will be fix in another ticket, and the code portion removed meanwhile.

`OK`: Done / Validated
`-`: Not needed


nidup:
It may fixes https://github.com/akeneo/pim-community-dev/issues/6086 and https://github.com/akeneo/pim-community-dev/issues/6054

LaurentPetard:
Fix already done in #5957 I close this PR.

Issue #6106, opened 2017-04-25
HTML title double encoded


Hocdoc:
**I'm reporting a Bug**

When a SKU contains f.ex. quotes, they are encoded in the HTML title again (V. 1.7.3 CE).

Steps to reproduce:

1. Create a product with quotes in the SKU like `My "Test"`.
2. Edit this product. In the tab title (HTML `<title>`-Tag) you will see `Products My &quot;Test&quot;` instead of the expected `Product My "Test"`

mickaelandrieu:
Hello @Hocdoc,

you're right. This is valid HTML so I'm wondering why is it a bug for you: does some search engine display "&quot;" as it?

I can also confirm the values are stored without encoding in the database.

Thanks for your feedback

Hocdoc:
The title that the *user* sees in his browser tab is `Products My &quot;Test&quot;`.

The title-HTML-tag must be doubble-encoded, maybe something like `Products My &amp;quot;Test&amp;quot;`

SamirBoulil:
Nice catch @Hocdoc ! This is definitely a minor bug. 

I'll label this issue and add it to the github project.

mickaelandrieu:
Indeed, I'm not sure about the best way to fix it, what do you think @SamirBoulil ?

Also, @Hocdoc, do you want to contribute to a patch?

Hocdoc:
I've add an PR #6176; 
but I haven't tested it and I'm not sure if `navTitle.params` should get escaped.

mickaelandrieu:
@Hocdoc thank you for your pull request, we'll review it as soon as possible :)

Mickaël

SamirBoulil:
The fix @Hocdoc provided has been successfully merged in 1.7 !

Thanks everyone for your help !

PR #6176, opened 2017-05-16
Fix double-escaped title (#6106)

Hocdoc:
This fixes https://github.com/akeneo/pim-community-dev/issues/6106

### DIFF
--- a/src/Pim/Bundle/EnrichBundle/Resources/views/Product/edit.html.twig
+++ b/src/Pim/Bundle/EnrichBundle/Resources/views/Product/edit.html.twig
@@ -56,13 +56,13 @@
                                     FetcherRegistry.initialize().done(function () {
                                         FetcherRegistry.getFetcher('product').fetch('{{ productId }}')
                                             .then(function (product) {
-                                                var sku      = _.escape(product.meta.label[UserContext.get('catalogLocale')]);
-                                                var newTitle = document.title.replace('%product.sku%', sku);
+                                                var label = product.meta.label[UserContext.get('catalogLocale')];
+                                                var newTitle = document.title.replace('%product.sku%', label);
 
                                                 document.title  = newTitle;
                                                 var navigation  = Navigation.getInstance();
                                                 var navTitle    = navigation.getPinButtonsData('title');
-                                                navTitle.params = {'%product.sku%': sku};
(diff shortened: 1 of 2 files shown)


SamirBoulil:
Hello @Hocdoc !

Thank you for this quick contribution :)

Let me try this fix and run the CI on it. If everything goes well, we'll have it merge by the end of the week ^^

I'll keep you updated!

SamirBoulil:
I tested out this PR and it's fixing the issue. All we need now is a rebase and then I'll run the CI on it.

Thanks for the contribution ;)

src/Pim/Bundle/EnrichBundle/Resources/views/Product/edit.html.twig
```diff
@@ -56,13 +56,13 @@
                                     FetcherRegistry.initialize().done(function () {
                                         FetcherRegistry.getFetcher('product').fetch('{{ productId }}')
                                             .then(function (product) {
-                                                var sku      = _.escape(product.meta.label[UserContext.get('catalogLocale')]);
-                                                var newTitle = document.title.replace('%product.sku%', sku);
+                                                var label = product.meta.label[UserContext.get('catalogLocale')];
```
a2xchip: i think here you need to use `pim/i18n` to get value for specific locale, not `[localeCode]` style

SamirBoulil:
Hello @hocdoc! 

The PR seems in a good way to be merged ! 
- CI is green
- We got our GTMs

I'll just add a small behat to test it all and we will be all set !

src/Pim/Bundle/EnrichBundle/Resources/views/Product/edit.html.twig
grena: @a2xchip We usually use i18n for other entities yes (because it allows to fallback on the code), but I had a look in the code and everywhere else, for product, we use the `[localeCode]` style. So imho it's ok if it's done this way to be consistent :)

SamirBoulil:
Congratulations @Hocdoc, Your PR has been merged ! 💃 

🎉  Again, thank you for your contribution ! 

You can now claim your badge "Core contributor" on http://badger.akeneo.com if you don't already have it. ^^

Issue #6153, opened 2017-05-10
Group edit triggers 414 Request URI too large with 81 selected (?) products in group


wimvds:
**I'm reporting a Bug**

Expected behavior : When I go to the edit view of a group (via Enrich, Groups, eye icon), I should see the datagrid.

Encountered behavior : When I go to the edit view of a group, I don't see the datagrid. Checking the console, I see that a call to **datagrid/product-group-grid/load** triggers a **414 Request-URI Too Large** error.

Payload of the request is the following : 
[...]

PR #6183, opened 2017-05-17
PIM-6418: Fix URL too long in variant group datagrid

LaurentPetard:

The URI generated to load the products in the variant groups Datagrid can be too long if there are a lot of products. Indeed, each product selection is added as an array parameter to the URI.

Given that this parameters are not used to load the datagrid elements (regardless of the entity). The solution is to remove them from the URI generation. It can be done in the Datagrid initialization by removing the property  `selection` from `options`. It hasn't consequences because the selection is handled in a specific attribute of the Datagrid (see line https://github.com/akeneo/pim-community-dev/blob/817d68366dcf2ed94dc465deded0d239c1fae6f1/src/Pim/Bundle/EnrichBundle/Resources/public/js/form/common/grid.js#L29)

`OK`: Done / Validated
`-`: Not needed

This will fix issue https://github.com/akeneo/pim-community-dev/issues/6153

### DIFF
--- a/src/Pim/Bundle/EnrichBundle/Resources/public/js/form/common/grid.js
+++ b/src/Pim/Bundle/EnrichBundle/Resources/public/js/form/common/grid.js
@@ -32,6 +32,13 @@ define([
                 });
                 this.options   = options;
 
+                /*
+                 * Removing to be sure that this property will not be used in URLs generated to load the data
+                 * The selection is never used back side to load the data and it can generate an URL too long.
+                 * The rightful usages of the selection are done with the property "this.selection"
+                 */
+                delete this.options.selection;
+
                 mediator.on('datagrid:selectModel:' + this.alias, function (model) {
                     this.addElement(model.get('id'));
                 }.bind(this));
--- a/CHANGELOG-1.6.md
+++ b/CHANGELOG-1.6.md
@@ -1,3 +1,9 @@
+# 1.6.x
+
+## Bug fixes
+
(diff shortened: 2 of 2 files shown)


src/Pim/Bundle/EnrichBundle/Resources/public/js/form/common/grid.js
```diff
@@ -32,6 +32,8 @@ define([
                 });
                 this.options   = options;
 
+                delete this.options.selection;
```
a2xchip: 1. why you are removing it?
2. why you are doing it in common `grid.js` if you need to fix it only for `variant group` you can implement custom `grid.js` for it.

a2xchip: Ok about first question... after re-reading PR desc
What about 2.?

LaurentPetard: 1. Because the property `options` is also used further to generate the URI to load the data. And the parameters `selection` are useless server side to fetch the data. The sub-property `options.selection` must only be used to initialize the Datagrid. It's done with the line `this.selection = options.selection`, and it's the property `this.selection` that is used to update the selected elements of the Datagrid.

2. I'm doing it here because it's true for all cases, not only for variant group. Using the property `options.selection` after the Datagrid initialization would be an error, because the changes of the selected elements are not handled in it (but only in `this.selection`)

PR #6208, opened 2017-05-23
PIM-6433: Prevent the insertion of duplicate multi-reference data code to break an import

SamirBoulil:

When importing a csv product file with a multi select reference data option code appearing twice or more for one attribute. The import job fails with an error:

`Exception SQL raised : Duplicate entry in the table that stores the values of the ref data.`

This PR intends to fix this issue by fixing the product value override to check if the reference data we want to add already exists before adding it to the multi select list.

This PR is mostly for educational purpose, so that integrators can rely on this code sample to create their on reference data correctly.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/CHANGELOG-1.6.md
+++ b/CHANGELOG-1.6.md
@@ -1,15 +1,21 @@
+# 1.6.x
+
+## Bug fixes
+
+- PIM-6433: Fix product import reference data multi with duplicate code options
+
 # 1.6.16 (2017-05-24)
 
 ## Bug fixes
 
 - PIM-6418: Fix URL too long in variant group datagrid
+- GITHUB-5451: Update akeneo_batch.yml sender email, cheers @julienquetil!
 
 # 1.6.15 (2017-05-18)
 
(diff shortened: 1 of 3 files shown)


CHANGELOG-1.6.md
```diff
@@ -1,15 +1,21 @@
...
 
 - PIM-6418: Fix URL too long in variant group datagrid
+- GITHUB-5451: Update akeneo_batch.yml sender email, cheers @julienquetil!
 
 # 1.6.15 (2017-05-18)
 
 ## Bug fixes
 
 - PIM-6376: Fix slow MongoDB queries when editing attribute options
-- GITHUB-5451: Update akeneo_batch.yml sender email, cheers @julienquetil!
```
ahocquard: Moving it in 1.6.16 is normal ?

SamirBoulil: Yep, I just realised that it was not in the right version. (I added in 1.6.16)

PR #6219, opened 2017-05-29
[PIM-6436] Fix attribute group limit in Product Edit Form

pierallard:
cf https://akeneo.atlassian.net/browse/PIM-6436

### DIFF
--- a/features/Context/WebUser.php
+++ b/features/Context/WebUser.php
@@ -972,7 +972,7 @@ class WebUser extends RawMinkContext
     /**
      * @param string $groups
      *
-     * @Then /^I add attributes by group "([^"]*)"$/
+     * @Then /^I add attributes by groups? "([^"]*)"$/
      */
     public function iAddAttributesByGroup($groups)
     {
--- a/features/attribute-group/list_attribute_groups.feature
+++ b/features/attribute-group/list_attribute_groups.feature
@@ -42,3 +42,64 @@ Feature: List attribute groups
       """
       Sizes,Marketing,Product information,Colors,Other,Media
       """
+
+  @javascript @jira https://akeneo.atlassian.net/browse/PIM-6436
+  Scenario: Sucessfully display the attribute groups in the PEF without limit
+    Given the "footwear" catalog configuration
+    And I am logged in as "Julia"
+    And the following CSV file to import:
+      """
+      code;attributes;sort_order
+      attribute_group_1;sku;101
+      attribute_group_2;name;102
+      attribute_group_3;manufacturer;103
+      attribute_group_4;weather_conditions;104
+      attribute_group_5;description;105
+      attribute_group_6;comment;106
+      attribute_group_7;price;107
+      attribute_group_8;rating;108
+      attribute_group_9;side_view;109
+      attribute_group_10;top_view;110
+      attribute_group_11;size;111
+      attribute_group_12;color;112
+      attribute_group_13;lace_color;113
+      attribute_group_14;length;114
+      attribute_group_15;volume;115
+      attribute_group_16;number_in_stock;116
+      attribute_group_17;destocking_date;117
+      attribute_group_18;handmade;118
+      attribute_group_19;heel_color;119
+      attribute_group_20;sole_color;120
+      attribute_group_21;cap_color;121
+      attribute_group_22;sole_fabric;122
+      """
+    And the following job "csv_footwear_attribute_group_import" configuration:
+      | filePath | %file to import% |
+    And I am on the "csv_footwear_attribute_group_import" import job page
+    And I launch the import job
+    And I wait for the "csv_footwear_attribute_group_import" job to finish
+    And I am on the families page
+    And I create a new family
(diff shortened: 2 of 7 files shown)


features/Context/WebUser.php
```diff
@@ -972,7 +972,7 @@ public function iShouldSeeAvailableAttributeGroup($groups)
     /**
      * @param string $groups
      *
-     * @Then /^I add attributes by group "([^"]*)"$/
+     * @Then /^I add attributes by groups? "([^"]*)"$/
      */
     public function iAddAttributesByGroup($groups)
```
mickaelandrieu: should be renamed `iAddAttributesByGroup`

src/Pim/Bundle/EnrichBundle/Controller/Rest/AttributeGroupController.php
```diff
@@ -94,13 +71,48 @@ public function indexAction(Request $request)
...
 
         return new JsonResponse($normalizedAttributeGroups);
     }
+
+    /**
+     * @param Request $request
+     *
+     * @return Array
+     */
+    protected function parseOptions(Request $request)
```
mickaelandrieu: not sure this should be here... or about an ``OptionsRequest`` class, that take Request in constructor and return the options array you need. This doesn't even need to be a service, something like `$options = OptionsRequest::buildFromRequest($request)` make more sense to me than doing stuff in controller.

features/Context/WebUser.php
pierallard: That's the name of this function, wtf ?

src/Pim/Bundle/EnrichBundle/Controller/Rest/AttributeGroupController.php
pierallard: If it disturbs you to have this function here, I can get all this back to the indexAction one.
We're applying a patch in 1.7, not the place to refactor the PIM ;)

src/Pim/Bundle/EnrichBundle/Controller/Rest/AttributeGroupController.php
```diff
@@ -94,13 +71,48 @@ public function indexAction(Request $request)
...
+    {
+        $options = $request->get('options', []);
+
+        if (!isset($options['limit'])) {
+            $options['limit'] = SearchableRepositoryInterface::FETCH_LIMIT;
+        }
+
+        if (0 > intval($options['limit'])) {
+            $options['limit'] = null;
```
mickaelandrieu: why not return the default value in this case too?

pierallard: Because we can not pass null parameters to a GET url parameter. The only option is to pass a negative integer to get it back to null in the controller.

src/Pim/Bundle/EnrichBundle/Controller/Rest/AttributeGroupController.php
mickaelandrieu: keep it please: it's better now than before 👍

features/attribute-group/list_attribute_groups.feature
```diff
@@ -42,3 +42,64 @@ Feature: List attribute groups
       """
       Sizes,Marketing,Product information,Colors,Other,Media
       """
+
+  @javascript @jira https://akeneo.atlassian.net/browse/PIM-6436
+  Scenario: Sucessfully display the attribute groups in the PEF without limit
+    Given the "footwear" catalog configuration
+    And I am logged in as "Julia"
+    And the following CSV file to import:
```
SamirBoulil: Ok, but you couldn't you just use a step to define an attribute instead of using an import ? (I mean, we don't test attribute groups import here).

You can let it as is.

src/Pim/Bundle/EnrichBundle/Controller/Rest/AttributeGroupController.php
SamirBoulil: I'd pass the result `$request->get('options', []);` instead of the whole request object.

src/Pim/Bundle/EnrichBundle/Controller/Rest/AttributeGroupController.php
SamirBoulil: Well that's a tricky one, maybe a word about it in the phpdoc ?

src/Pim/Bundle/EnrichBundle/Controller/Rest/AttributeGroupController.php
pierallard: We need the complete $request and not only options (just look above ;) )

features/attribute-group/list_attribute_groups.feature
pierallard: IMHO Import is a process near from reality than your custom "And the following..." steps. And it prevents some history bugs on DCI (which we don't test here, I know, it's just a way of life)

PR #6226, opened 2017-05-30
API-245: add the possibility to pass an username in the command

solivier:
This PR add the possibility to pass an username in the command and add it in the security token to be able to retrieve it later.

Will be used in import job rights.

### DIFF
--- a/src/Akeneo/Bundle/BatchBundle/Command/BatchCommand.php
+++ b/src/Akeneo/Bundle/BatchBundle/Command/BatchCommand.php
@@ -4,11 +4,9 @@ namespace Akeneo\Bundle\BatchBundle\Command;
 
 use Akeneo\Component\Batch\Item\ExecutionContext;
 use Akeneo\Component\Batch\Job\ExitStatus;
-use Akeneo\Component\Batch\Job\Job;
 use Akeneo\Component\Batch\Job\JobParametersFactory;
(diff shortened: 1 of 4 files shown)


src/Akeneo/Bundle/BatchBundle/Command/BatchCommand.php
```diff
@@ -78,6 +85,18 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+        $username = $input->getOption('username');
+
+        try {
+            $user = $this->getContainer()->get('security.user.provider.concrete.chain_provider')
+                ->loadUserByUsername($username);
+
+            $token = new UsernamePasswordToken($user, null, 'main', $user->getRoles());
+            $this->getContainer()->get('security.token_storage')->setToken($token);
+        } catch (UsernameNotFoundException $e) {
+            // We do nothing as we want to let the possibility to not pass an user.
```
mickaelandrieu: If `username` is optional, you should embed your ``try/catch`` inside an if condition: ``if (!is_null($username)) {}``.

But as a developer, if the username I use is wrong (I did a typo for instance) I want to be notified, so please, dont hide the error! At least, if you disagree with me, can you log or output something? 

```php
$warning = sprintf('<warning>User with username %s was not found</warning>', $username);
$output->writeln($warning);
```

PR #6249, opened 2017-06-05
API-218, API-252: Add measures families endpoints

solivier:

Add measures families endpoints

### DIFF
--- a/src/Pim/Bundle/ApiBundle/Controller/MeasureFamiliesController.php
+++ b/src/Pim/Bundle/ApiBundle/Controller/MeasureFamiliesController.php
@@ -1,0 +1,100 @@
+<?php
+
+namespace Pim\Bundle\ApiBundle\Controller;
+
+use Pim\Component\Api\Pagination\PaginatorInterface;
+use Pim\Component\Connector\ArrayConverter\ArrayConverterInterface;
+use Symfony\Component\HttpFoundation\JsonResponse;
+use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
+
+/**
+ * Measure families controller
+ *
+ * @author    Olivier Soulet <olivier.soulet@akeneo.com>
+ * @copyright 2017 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+class MeasureFamiliesController
+{
+    /** @var array */
+    protected $measuresConfig;
+
+    /** @var ArrayConverterInterface */
+    protected $measureFamilyConverter;
+
+    /** @var PaginatorInterface */
+    protected $paginator;
+
+    /** @var array */
+    protected $apiConfiguration;
+
+    /**
+     * @param ArrayConverterInterface $measureFamilyConverter
+     * @param PaginatorInterface      $paginator
+     * @param array                   $measures
+     * @param array                   $apiConfiguration
+     */
+    public function __construct(
+        ArrayConverterInterface $measureFamilyConverter,
+        PaginatorInterface $paginator,
+        array $measures,
+        array $apiConfiguration
+    ) {
+        $this->measuresConfig = $measures['measures_config'];
+        $this->measureFamilyConverter = $measureFamilyConverter;
+        $this->paginator = $paginator;
+        $this->apiConfiguration = $apiConfiguration;
+    }
+
+    /**
+     * @param string  $code
+     *
+     * @throws NotFoundHttpException
+     *
+     * @return JsonResponse
+     */
+    public function getAction($code)
+    {
+        if (!array_key_exists($code, $this->measuresConfig)) {
+            throw new NotFoundHttpException(sprintf('Measure family with code "%s" does not exist.', $code));
+        }
+
+        $normalizedFamily = $this->measureFamilyConverter->convert(
+            ['family_code' => $code, 'units' => $this->measuresConfig[$code]]
+        );
+
+        return new JsonResponse($normalizedFamily);
+    }
+
+    /**
+     * @return JsonResponse
+     */
+    public function listAction()
+    {
+        $convertedMeasureFamilies= [];
+        foreach ($this->measuresConfig as $familyCode => $units) {
+            $convertedMeasureFamilies[] = $this->measureFamilyConverter
+                ->convert(['family_code' => $familyCode, 'units' =>$units]);
+        }
+
+        $defaultParameters = [
+            'page'       => 1,
+            'limit'      => $this->apiConfiguration['pagination']['limit_by_default'],
+            'with_count' => 'false',
+        ];
+
+        $parameters = [
+            'query_parameters'    => $defaultParameters,
+            'list_route_name'     => 'pim_api_measure_families_list',
+            'item_route_name'     => 'pim_api_measure_families_get',
+        ];
+
+        $paginatedMeasureFamilies = $this->paginator->paginate(
+            $convertedMeasureFamilies,
+            $parameters,
+            null
+        );
+
+        return new JsonResponse($paginatedMeasureFamilies['_embedded']);
+    }
+}
--- a/src/Pim/Bundle/ApiBundle/Controller/MeasureFamiliesController.php
+++ b/src/Pim/Bundle/ApiBundle/Controller/MeasureFamiliesController.php
@@ -80,14 +80,12 @@ class MeasureFamiliesController
                 ->convert(['family_code' => $familyCode, 'units' => $units]);
         }
 
-        $defaultParameters = [
-            'page'       => 1,
-            'limit'      => $this->apiConfiguration['pagination']['limit_by_default'],
-            'with_count' => 'false',
-        ];
-
         $parameters = [
-            'query_parameters'    => $defaultParameters,
+            'query_parameters'    => [
+                'page'       => 1,
+                'limit'      => $this->apiConfiguration['pagination']['limit_by_default'],
+                'with_count' => 'false',
+            ],
             'list_route_name'     => 'pim_api_measure_families_list',
             'item_route_name'     => 'pim_api_measure_families_get',
         ];
--- a/src/Pim/Bundle/ApiBundle/Controller/MeasureFamiliesController.php
+++ b/src/Pim/Bundle/ApiBundle/Controller/MeasureFamiliesController.php
@@ -8,8 +8,6 @@ use Symfony\Component\HttpFoundation\JsonResponse;
 use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
 
 /**
- * Measure families controller
- *
  * @author    Olivier Soulet <olivier.soulet@akeneo.com>
  * @copyright 2017 Akeneo SAS (http://www.akeneo.com)
  * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
@@ -74,7 +72,7 @@ class MeasureFamiliesController
         $convertedMeasureFamilies= [];
         foreach ($this->measuresConfig as $familyCode => $units) {
             $convertedMeasureFamilies[] = $this->measureFamilyConverter
-                ->convert(['family_code' => $familyCode, 'units' =>$units]);
+                ->convert(['family_code' => $familyCode, 'units' => $units]);
         }
 
         $defaultParameters = [
--- a/src/Pim/Bundle/ApiBundle/Controller/MeasureFamiliesController.php
+++ b/src/Pim/Bundle/ApiBundle/Controller/MeasureFamiliesController.php
@@ -53,12 +53,17 @@ class MeasureFamiliesController
      */
     public function getAction($code)
     {
-        if (!array_key_exists($code, $this->measuresConfig)) {
+        $measuresConfig = [];
+        foreach ($this->measuresConfig as $key => $value) {
+            $measuresConfig[strtolower($key)] = $value;
+        }
+
+        if (!array_key_exists(strtolower($code), $measuresConfig)) {
             throw new NotFoundHttpException(sprintf('Measure family with code "%s" does not exist.', $code));
         }
 
         $normalizedFamily = $this->measureFamilyConverter->convert(
-            ['family_code' => $code, 'units' => $this->measuresConfig[$code]]
+            ['family_code' => $code, 'units' => $measuresConfig[strtolower($code)]]
         );
 
         return new JsonResponse($normalizedFamily);
--- a/src/Pim/Bundle/ApiBundle/tests/integration/Controller/MeasureFamilies/GetMeasureFamiliesIntegration.php
+++ b/src/Pim/Bundle/ApiBundle/tests/integration/Controller/MeasureFamilies/GetMeasureFamiliesIntegration.php
@@ -1,0 +1,181 @@
+<?php
+
+namespace Pim\Bundle\ApiBundle\tests\integration\Controller\MeasureFamilies;
+
+use Akeneo\Test\Integration\Configuration;
+use Pim\Bundle\ApiBundle\tests\integration\ApiTestCase;
+use Symfony\Component\HttpFoundation\Response;
+
+class GetMeasureFamiliesIntegration extends ApiTestCase
+{
+    public function testGetAnMeasureFamily()
+    {
+        $client = $this->createAuthenticatedClient();
+
+        $client->request('GET', 'api/rest/v1/measure-families/Area');
+
+        $standardAttribute = <<<JSON
+{
+  "code": "Area",
+  "standard": "SQUARE_METER",
+  "units": [
+    {
+      "code": "SQUARE_MILLIMETER",
+      "convert": {
+        "mul": 1.0e-6
+      },
+      "symbol": "mm\u00b2"
+    },
+    {
+      "code": "SQUARE_CENTIMETER",
+      "convert": {
+        "mul": 0.0001
+      },
+      "symbol": "cm\u00b2"
+    },
+    {
+      "code": "SQUARE_DECIMETER",
+      "convert": {
+        "mul": 0.01
+      },
+      "symbol": "dm\u00b2"
+    },
+    {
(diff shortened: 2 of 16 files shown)


src/Pim/Bundle/ApiBundle/Controller/MeasureFamiliesController.php
```diff
@@ -0,0 +1,100 @@
...
+
+namespace Pim\Bundle\ApiBundle\Controller;
+
+use Pim\Component\Api\Pagination\PaginatorInterface;
+use Pim\Component\Connector\ArrayConverter\ArrayConverterInterface;
+use Symfony\Component\HttpFoundation\JsonResponse;
+use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
+
+/**
+ * Measure families controller
```
mickaelandrieu: can be removed

src/Pim/Bundle/ApiBundle/Controller/MeasureFamiliesController.php
```diff
@@ -0,0 +1,100 @@
...
+
+    /**
+     * @return JsonResponse
+     */
+    public function listAction()
+    {
+        $convertedMeasureFamilies= [];
+        foreach ($this->measuresConfig as $familyCode => $units) {
+            $convertedMeasureFamilies[] = $this->measureFamilyConverter
+                ->convert(['family_code' => $familyCode, 'units' =>$units]);
```
mickaelandrieu:
```diff
- =>$units
+ => $units
```

src/Pim/Bundle/ApiBundle/Controller/MeasureFamiliesController.php
```diff
@@ -0,0 +1,100 @@
...
+        }
+
+        $defaultParameters = [
```
mickaelandrieu: hmm... sounds like a service configuration done in your action instead of your ``services.yml`` file :/ does it need to refactored or is it the right place to do it?

src/Pim/Bundle/ApiBundle/Controller/MeasureFamiliesController.php
```diff
@@ -0,0 +1,100 @@
...
+            'page'       => 1,
+            'limit'      => $this->apiConfiguration['pagination']['limit_by_default'],
+            'with_count' => 'false',
+        ];
+
+        $parameters = [
```
mickaelandrieu: for instance, maybe theses parameters should be declared in ``services.yml`` file

src/Pim/Bundle/ApiBundle/Controller/MeasureFamiliesController.php
```diff
@@ -0,0 +1,100 @@
...
+ * Measure families controller
+ *
+ * @author    Olivier Soulet <olivier.soulet@akeneo.com>
+ * @copyright 2017 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+class MeasureFamiliesController
+{
+    /** @var array */
+    protected $measuresConfig;
```
mickaelandrieu: why `protected` and not `private` ?

src/Pim/Component/Api/Converter/MeasureFamilyConverter.php
```diff
@@ -0,0 +1,46 @@
...
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+class MeasureFamilyConverter implements ArrayConverterInterface
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function convert(array $item, array $options = [])
+    {
+        $convertedItem = [
```
mickaelandrieu: What happens if one of this array property is undefined? I can't see any validation of a measure definitions in the workflow

src/Pim/Bundle/ApiBundle/Controller/MeasureFamiliesController.php
solivier: Because it's currently the akeneo pim convention

src/Pim/Component/Api/Converter/MeasureFamilyConverter.php
solivier: As it comes from a config file I'm sure sure we need a validation for this ?

src/Pim/Bundle/ApiBundle/Controller/MeasureFamiliesController.php
mickaelandrieu: Ok, I'm interested on why all properties of every Controllers are ``protected`` by default. It's maybe obvious or stupid question, but I want to better understand the architectural choices done here ^^

solivier: it's because it's easier to override for an integrator :)

src/Pim/Component/Api/Converter/MeasureFamilyConverter.php
mickaelandrieu: The loading is done using a [bundle extension](https://github.com/akeneo/pim-community-dev/blob/master/src/Akeneo/Bundle/MeasureBundle/DependencyInjection/AkeneoMeasureExtension.php#L35) but it is permissive. For instance if you forget the ``symbol`` key this will break your REST API, right? But I'm wondering if we should validate the measure here or in the extension. What do you think?

solivier: it's only in read mode for measure families and it comes only from a measure config file. the only way to forget the symbol value is to forget to add it in the measure config file. And I guess it will break before :)

src/Pim/Bundle/ApiBundle/Controller/MeasureFamiliesController.php
skeleton: it's not a service configuration, but a route configuration ;) imvho, we can keep like that.

src/Pim/Bundle/ApiBundle/Controller/MeasureFamiliesController.php
skeleton: The main advantage to extract these parameters is to override them easily. As we are in API context, this action has few chances to be overidden.
On the other hand, I'm not a big fan of data hardcoded like route naming (and it's my fault, I created the first action for the API and I have hardcoded routes like that :p), maybe we will extract some params for all API routes in service configuration, but not in this PR.

src/Pim/Component/Api/Converter/MeasureFamilyConverter.php
skeleton: it seems there is enough validation for the measure configuration: https://github.com/akeneo/pim-community-dev/blob/master/src/Akeneo/Bundle/MeasureBundle/DependencyInjection/Configuration.php

The only thing not checked is the `units` key, which is not required (you have to take in account that in your code @solivier )

skeleton:
thank for the review @mickaelandrieu, but it's a little bit early ;)
PR has not been sent on slack and has not the label "To review" ;)

src/Pim/Bundle/ApiBundle/Controller/MeasureFamiliesController.php
mickaelandrieu: its not really clear that the objective is to generate a route, how about ``$routeParameters`` instead?

src/Pim/Bundle/ApiBundle/Controller/MeasureFamiliesController.php
mickaelandrieu: You're right. Still, I don't really like to configure my stuff inside controllers. Not very important, but the more we will wait, the more this will become something like "Akeneo" convention. As all my questions were answered I can change my review to approve :)

src/Pim/Bundle/ApiBundle/Controller/MeasureFamiliesController.php
mickaelandrieu: please don't forget it, it's nothing but it disturbs me and will disturbs me forever :D

src/Pim/Bundle/ApiBundle/Controller/MeasureFamiliesController.php
```diff
@@ -0,0 +1,103 @@
...
+     */
+    public function listAction()
+    {
+        $convertedMeasureFamilies= [];
+        foreach ($this->measuresConfig as $familyCode => $units) {
+            $convertedMeasureFamilies[] = $this->measureFamilyConverter
+                ->convert(['family_code' => $familyCode, 'units' => $units]);
+        }
+
+        $defaultParameters = [
```
ahocquard: you can directly put it in $parameter as you don't merge anything like the other controllers

src/Pim/Bundle/ApiBundle/Controller/MeasureFamiliesController.php
```diff
@@ -0,0 +1,103 @@
...
+        }
+
+        $normalizedFamily = $this->measureFamilyConverter->convert(
+            ['family_code' => $code, 'units' => $measuresConfig[strtolower($code)]]
+        );
+
+        return new JsonResponse($normalizedFamily);
+    }
+
+    /**
```
ahocquard: same : no acl ?

src/Pim/Bundle/ApiBundle/Controller/MeasureFamiliesController.php
```diff
@@ -0,0 +1,103 @@
...
+        array $apiConfiguration
+    ) {
+        $this->measuresConfig = $measures['measures_config'];
+        $this->measureFamilyConverter = $measureFamilyConverter;
+        $this->paginator = $paginator;
+        $this->apiConfiguration = $apiConfiguration;
+    }
+
+    /**
+     * @param string  $code
```
ahocquard: no acl on this route ?

src/Pim/Bundle/ApiBundle/tests/integration/Controller/MeasureFamilies/GetMeasureFamiliesIntegration.php
```diff
@@ -0,0 +1,181 @@
...
+  ]
+}
+JSON;
+
+        $response = $client->getResponse();
+        $this->assertSame(Response::HTTP_OK, $response->getStatusCode());
+        $this->assertJsonStringEqualsJsonString($standardAttribute, $response->getContent());
+    }
+
+    public function testNotFoundAnMeasureFamily()
```
ahocquard: aMeasureFamily

src/Pim/Bundle/ApiBundle/tests/integration/Controller/MeasureFamilies/GetMeasureFamiliesIntegration.php
```diff
@@ -0,0 +1,181 @@
...
+
+namespace Pim\Bundle\ApiBundle\tests\integration\Controller\MeasureFamilies;
+
+use Akeneo\Test\Integration\Configuration;
+use Pim\Bundle\ApiBundle\tests\integration\ApiTestCase;
+use Symfony\Component\HttpFoundation\Response;
+
+class GetMeasureFamiliesIntegration extends ApiTestCase
+{
+    public function testGetAnMeasureFamily()
```
ahocquard: AMeasureFamily

src/Pim/Bundle/ApiBundle/tests/integration/Controller/MeasureFamilies/GetMeasureFamiliesIntegration.php
```diff
@@ -0,0 +1,181 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function getConfiguration()
+    {
+        return new Configuration(
+            [Configuration::getTechnicalCatalogPath()],
+            false
```
ahocquard: this parameter still exists ? Because we are in process isolation now.

src/Pim/Component/Api/Converter/MeasureFamilyConverter.php
```diff
@@ -0,0 +1,54 @@
...
+            "code" => $item['family_code'],
```
ahocquard: double quote

src/Pim/Component/Api/Converter/MeasureFamilyConverter.php
```diff
@@ -0,0 +1,54 @@
...
+     * @param array $units
+     *
+     * @return array
+     *
+     * @throws \Exception
+     */
+    protected function convertUnits(array $units)
+    {
+        if (!array_key_exists('units', $units)) {
+            throw new \Exception('The "units" array key is missing');
```
ahocquard: InvalidArgumentException ? A little "hard" to throw Exception :)

src/Pim/Component/Api/Converter/MeasureFamilyConverter.php
```diff
@@ -0,0 +1,54 @@
...
+        }
+
+        $convertedUnits = [];
+        foreach ($units['units'] as $code => $unit) {
+            $convertedUnits[] = [
+                "code" => $code,
```
ahocquard: simple quote

src/Pim/Component/Api/Converter/MeasureFamilyConverter.php
```diff
@@ -0,0 +1,54 @@
...
+                "convert" => call_user_func_array('array_merge', $unit['convert']),
```
ahocquard: could you explain please ? because phpdoc tells me that the second argument of call_user_func_array is an array.

src/Pim/Bundle/ApiBundle/Controller/MeasureFamiliesController.php
solivier: no need to have acl

src/Pim/Bundle/ApiBundle/Controller/MeasureFamiliesController.php
solivier: same no acl (as asked in the card)

src/Pim/Component/Api/Converter/MeasureFamilyConverter.php
solivier: $unit['convert'] is an array of arrays

src/Pim/Component/Api/Converter/MeasureFamilyConverter.php
solivier: it's not really an invalid argument, it's more a missing key.

src/Pim/Bundle/ApiBundle/Controller/MeasureFamiliesController.php
```diff
@@ -0,0 +1,101 @@
...
+use Pim\Component\Connector\ArrayConverter\ArrayConverterInterface;
+use Symfony\Component\HttpFoundation\JsonResponse;
+use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
+
+/**
+ * @author    Olivier Soulet <olivier.soulet@akeneo.com>
+ * @copyright 2017 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+class MeasureFamiliesController
```
skeleton: we put all our class in singulary, so MeasureFamilyController

src/Pim/Bundle/ApiBundle/Controller/MeasureFamiliesController.php
```diff
@@ -0,0 +1,101 @@
...
+            'item_route_name'     => 'pim_api_measure_families_get',
+        ];
+
+        $paginatedMeasureFamilies = $this->paginator->paginate(
+            $convertedMeasureFamilies,
+            $parameters,
+            null
+        );
+
+        return new JsonResponse($paginatedMeasureFamilies['_embedded']);
```
skeleton: please do not return only _embedded, return the full array $paginatedMeasureFamilies.
Even if we know we return all the measures, we have to have the same type of response as the other list routes. If tomorrow we change the implementation of measure and become a real resource, from the API point of view, the response will be the same, we won't have BC break.

src/Pim/Bundle/ApiBundle/tests/integration/Controller/MeasureFamilies/ListMeasureFamiliesIntegration.php
```diff
@@ -0,0 +1,1238 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function getConfiguration()
+    {
+        return new Configuration(
+            [Configuration::getTechnicalCatalogPath()],
+            false
```
skeleton: you can remove false

src/Pim/Component/Api/Converter/MeasureFamilyConverter.php
skeleton: I'm not sure we should thrown an exception here if units key is missing.
Return an empty array can be enough and match we the current implem (but you can check with PO)

src/Pim/Bundle/ApiBundle/Controller/MeasureFamiliesController.php
ahocquard: @skeleton so, is it really paginated ? Or is it just the format ?

skeleton: I want the same response between all our routes. concistency is more important than a weird behavior.
but if you really want a pagination, instead of return all measures on the first page, you can return only the top ten, I don't care, I'm not a PO ;)

ahocquard: @solivier ok, so take care about the parameters "limit" passed out to the pagination service, https://github.com/akeneo/pim-community-dev/blob/master/src/Pim/Component/Api/Pagination/OffsetHalPaginator.php#L146

It should be equal to `count($convertedMeasureFamilie) +1`

PR #6260, opened 2017-06-06
TIP-744: PHP 7.1 Compatibility

anaelChardan:
## Description

We can, finally, fully support only PHP 7.1 \o/

For that, PHP requirement in `composer.json` has been updated.

The `PimRequirements` class has been reworked too, to check MySQL version. We cannot use the old `mysql_get_client_info()` method as it does not exists anymore in PHP 7.x: https://secure.php.net/manual/fr/function.mysql-get-client-info.php
So I choose to use `PDO` (as we already check it is installed) and the PIM parameters to access MySQL and get its version.
The only other way to do it would have been to use mysqli extension, but it is not a PIM requirement, so it is not safe at all to rely on it. Some suggested to use a simple `exec(mysql --version)` but this cannot work on install where MySQL is on a different machine (distant MySQL server or docker installation with different containers).

I also updated the "storage data collector" (that did get MySQL version too), so it use PDO (and remove the MongoDB part of it).

Finally, there was some small fixes to do on behat.

Behat not green:

Behat ee: `features/workflow/unpublish_product.feature:22`

## Definition Of Done

`OK`: Done / Validated
`-`: Not needed


app/PimRequirements.php
```diff
@@ -14,36 +16,27 @@
...
-     * @staticvar string
-     */
-    const REQUIRED_MYSQL_VERSION = '5.1';
+    const REQUIRED_MYSQL_VERSION = '5.7';
 
     /**
      * {@inheritdoc}
      */
-    public function __construct(array $directoriesToCheck = array())
+    public function __construct(array $directoriesToCheck = [])
```
mickaelandrieu: this means the minimal required version for MySQL is `5.7` ?

anaelChardan: Yes :)

Jenkinsfile
```diff
@@ -300,7 +289,6 @@ def runBehatTest(edition, features, phpVersion) {
             }
 
             // Configure the PIM
-            sh "cp behat.ci.yml behat.yml"
```
ahocquard: why ? it uses directly behat.ci.yml now ?

Jenkinsfile
```diff
@@ -16,7 +16,7 @@ stage("Checkout") {
             choice(choices: 'yes\nno', description: 'Run behat tests', name: 'launchBehatTests'),
             string(defaultValue: 'ee,ce', description: 'PIM edition the behat tests should run on (comma separated values)', name: 'editions'),
             string(defaultValue: 'features,vendor/akeneo/pim-community-dev/features', description: 'Behat scenarios to build', name: 'features'),
-            choice(choices: '7.0\n7.1', description: 'PHP version to run behat with', name: 'phpVersion'),
+            choice(choices: '7.1', description: 'PHP version to run behat with', name: 'phpVersion'),
```
damien-carcel: No, because PHP 7.2 is soon to be out. This will let us the possibility to try it ;-).

Jenkinsfile
damien-carcel: No, it's because it is done twice. Look below at line 302 (314 before the removals of the PR).

app/PimRequirements.php
```diff
@@ -107,6 +100,58 @@ public function getMandatoryRequirements()
...
+     * if no connection is reached.
+     *
+     * @return string
+     */
+    protected function getMySQLVersion()
+    {
+        $parameters = Yaml::parse(file_get_contents(__DIR__.'/config/parameters.yml'));
+
+        try {
+            if (null === $parameters) {
```
damien-carcel: I'm wondering if this whole part is really usefull, as we already test for PDO in the OroRequirements, and that a check for MySQL 5.7 was added in `composer.json`.

Maybe we should be just removed it completely, with the version comparison above.

app/PimRequirements.php
```diff
@@ -14,36 +16,27 @@
...
+        $currentMySQLVersion = $this->getMySQLVersion();
+
         $this->addPimRequirement(
-            !extension_loaded('php5-mysql'),
-            'Extension php5-mysql should be installed',
-            'Install and enable <strong>php5-mysql</strong>'
+            version_compare($currentMySQLVersion, self::REQUIRED_MYSQL_VERSION, '>='),
+            sprintf('MySQL version must be at least %s', self::REQUIRED_MYSQL_VERSION),
+            sprintf(
+                'Install MySQL %s or newer (installed version is %s)',
```
damien-carcel: True enough.

app/PimRequirements.php
ahocquard: The good thing is that we could improve that later to check that ONLY_FULL_GROUP_BY is deactivated. Not possible with composer.

damien-carcel: And I said crap, it is not in composer but in `app/config`, in doctrine configuration.

mickaelandrieu:
\o/

PR #6284, opened 2017-06-12
PIM-6442: Extract values behavior from product

aRn0D:
#### Extracting behavior
Currently in our stack, only `ProductInterface` can have product values. By having "product values", we mean that a product has this kind of methods:
- `getValues`
- `setValues`
- `removeValue`
- etc.

So basically, only the product **wears the responsibility of having product values**.

This PR aims to **extract this logic** of "values aware" from the product into a new dedicated interface, temporarily named `Pim\Component\Catalog\Model\ValuesContainerInterface`.

This way, any other entity could implement this new interface to wear values. This will be used by a `ProductModel` (_kind of product template for variations_) for instance, a future class.

So now that values are no longer product-related, we also renamed the `ProductValueCollection` to `ValueCollection`.

#### So, this PR...
- **Creates** the new `ValuesContainerInterface` interface
- **Creates** the new `ValuesContainerBuilderInterface` (_its builder, obviously_)
- **Renames** the `ProductValueCollection` to `ValueCollection`
- **Replaces** the `ValuesContainerInterface` where some `ProductInterface` were used before (not everywhere, as sometimes we really want a `ProductInterface`)
- **Replaces** the `ValuesContainerBuilderInterface` where some `ProductBuilderInterface` were used before (not everywhere, as sometimes we really want a `ProductBuilderInterface`)

--------

**TODO:**

- [x] Rename ProductValueCollection
- [x] Replace some ProductInterface by ValuesContainerInterface when needed
- [x] Update changelog
- [ ] See naming of `ValuesContainer`, because there is `ValueCollection`, be clear with that


src/Pim/Component/Catalog/Model/ValueCollectionInterface.php
```diff
@@ -9,11 +9,13 @@
  *
  * Highly inspired by \Doctrine\Common\Collection.
  *
+ * TODO: should be renamed ValueCollectionInterface
```
grena: We can drop this comment now

src/Pim/Component/Catalog/Updater/Copier/AttributeCopierInterface.php
```diff
@@ -18,17 +17,17 @@
...
-     * @param AttributeInterface $toAttribute
-     * @param array              $options
+     * @param ValuesContainerInterface $fromValuesContainer
+     * @param ValuesContainerInterface $toValuesContainer
+     * @param AttributeInterface       $fromAttribute
+     * @param AttributeInterface       $toAttribute
+     * @param array                    $options
      *
-     * @throws PropertyException
+     * @return
```
grena: ?

PR #6289, opened 2017-06-14
API-245: add pim batch command to pass a username

solivier:
Here we add a new command that extends the akeneo:batch:job command to be able to pass a
username and generate a token, if no user or invalid user it raises an
exception. We added a new command pim:batch:job command to not break other
uses of the command (aka install, twa ...) that does not require an username.

`OK`: Done / Validated
`-`: Not needed


src/Pim/Bundle/ConnectorBundle/Command/BatchCommand.php
```diff
@@ -0,0 +1,145 @@
...
+            $doctrine,
+            $validator,
+            $notifier,
+            $jobParametersFactory,
+            $jobParametersValidator,
+            $jobRegistry,
+            $jobRepository
+        );
+
+        $this->logger = $logger;
```
mickaelandrieu: I'm not sure you need to do it, as you call parent constructor who fill the proterties

PR #6294, opened 2017-06-15
API-108: don't store empty product values

grena:

Currently, we store empty values for product values:
```json
{
  "sku": {
    "<all_channels>": {
      "<all_locales>": "sneakers"
    }
  },
  "date": {
    "<all_channels>": {
      "<all_locales>": null
    }
  },
  "available": {
    "<all_channels>": {
      "<all_locales>": false
    }
  }
}
```

The goal of this PR is to change this behavior by removing empty values, and never store them:
```json
{
  "sku": {
    "<all_channels>": {
      "<all_locales>": "sneakers"
    }
  },
  "available": {
    "<all_channels>": {
      "<all_locales>": false
    }
  }
}
```

**DOD**


src/Pim/Component/Catalog/Model/ProductValueCollection.php
```diff
@@ -148,6 +148,18 @@ public function removeByAttribute(AttributeInterface $attribute)
     /**
      * {@inheritDoc}
      */
+    public function removeByCodes($attributeCode, $channelCode = null, $localeCode = null)
+    {
+        $channelCode = null !== $channelCode ? $channelCode : '<all_channels>';
+        $localeCode = null !== $localeCode ? $localeCode : '<all_locales>';
+        $key = sprintf('%s-%s-%s', $attributeCode, $channelCode, $localeCode);
```
jjanvier: stupid question
there is nothing in this class to build the key?

src/Pim/Component/Catalog/Updater/ProductUpdater.php
```diff
@@ -156,28 +156,24 @@ protected function updateProductFields(ProductInterface $product, $field, $value
...
-     *  - sets optional values (not related to family's attributes) with empty data if value already exists
+     *  - removes empty values from the product
+     *  - sets optional values when a data is provided
      *
      * @param ProductInterface $product
      * @param array            $values
      */
     protected function updateProductValues(ProductInterface $product, array $values)
     {
-        $family = $product->getFamily();
```
jjanvier: should really remove the check about families here?

it's more related to the catalog remodeling

jjanvier: because currently we still have optional attributes

grena: @jjanvier The check on the family here was to **enforce empty values for attributes of the family**, so yes, we want to remove it (we don't want empty values anymore). And yes, we keep optional attributes because we only expect a non-empty value to be saved (_may it comes from the family or not_)

src/Pim/Component/Catalog/Model/ProductValueCollection.php
grena: Nope, but maybe it could be the time to create one indeed

grena: done

momoss:
It was not an expected behaviour to store null value?

grena:
@momoss apparently not!

PR #6306, opened 2017-06-17
TIP-425: fix SQL requests to be only_full_group_by compliant in Mysql >= 5.7.5

ahocquard:

PIM currently does not work with the sql mode ONLY_FULL_GROUP_BY enable. This PR aims to fix it.

### Difference between Mysql <= 5.7.4 and Mysql >= 5.7.5

From Mysql 5.7.0 to Mysql 5.7.4, the option ONLY_FULL_GROUP_BY is deactivated by default.
From Mysql >= 5.7.5, ONLY_FULL_GROUP_BY is activated by default.

It means that the default parameter in Mysql >= 5.7.5 are stricter than in Mysql <= 5.7.4!

But that's a good news, because if ONLY_FULL_GROUP_BY was activated by default in Mysql <= 5.7.4, it would be very painful. Mysql is much more "clever" in Mysql 5.7.5.

Indeed, Mysql >= 5.7.5 allows to do group by like this:
```
SELECT name, address, MAX(age) FROM t GROUP BY name;
```

In Mysql <= 5.7.4, you have to do this:
```
SELECT name, address, MAX(age) FROM t GROUP BY name, address;
```

Also, it allows alias in HAVING clause as well, not in with Mysql <= 5.7.4. And Oro uses alias in HAVING.
It could have an impact with DISTINCT + ORDER BY as well.
For more details about the impact:  https://dev.mysql.com/doc/refman/5.7/en/group-by-handling.html

So, supporting ONLY_FULL_GROUP_BY in Mysql >= 5.7.5 is enough, no need to support it in Mysql <= 5.7.4, as it's not enable by default.

### Using DISTINCT instead of GROUP BY

In the PIM, we use (far too much ^^) GROUP BY to eliminate duplicate rows.
Imho, that's a misuse of GROUP BY that intends to calculate aggregate data (COUNT, MAX, etc).

DISTINCT is better at doing this job, because it does not sort data in output.
Indeed, GROUP BY sorts data.

So, if you use `group by (name)` without any order by, the output will be sort by name. 
If you use `distinct name` without any order by, name column could be in any order.

But implicit sort with GROUP BY is a deprecated feature since Mysql 5.6 (source: https://dev.mysql.com/doc/refman/5.6/en/order-by-optimization.html). It means Mysql can change this behavior in any version (and maybe in a minor, as for the ONLY_FULL_GROUP_BY...).

To sort data, you have to use ORDER BY, whatever you use DISTINCT or GROUP BY.

So, to sum up:
- DISTINCT without ORDER BY is faster than GROUP BY to eliminate rows
- DISTINCT with ORDER BY is equals in term of performance of GROUP BY + ORDER BY (with column in order by = column of group by)

And in practice, in the PIM, I've experienced that a use of DISTINCT was 5x faster that a use of GROUP BY, on a query on the completeness, on a real customer database.

That's why I use DISTINCT most of the time instead of GROUP BY to eliminate duplicate rows.

**Important note: validated by our beloved CTO, @BitOne**  

Benchmark on group datagrid with medium catalog:

| Request  | GROUP BY  |  DISTINCT |
|---|---|---|
| Group datagrid [no order by]  | 65 ms   |  35 ms |
| Group datagrid [default order by on label]  | 69 ms  |  45 ms |
| Count (Group datagrid) | 49 ms   | 38 ms  |

### Some Oro stuff I've discovered

I've tried to optimise the total number of rows of a datagrid, without any success. I mean, the solution is too complicated to be done in this PR.

It's done here:
https://github.com/akeneo/pim-community-dev/blob/master/src/Pim/Bundle/DataGridBundle/Extension/Pager/Orm/Pager.php#L98

To calculate the total number of entities in a datagrid, we select all columns. `select(1)` is enough when counting. It avoids row-lookup. It's much more performant/scalable.

I've tried it, and it was working very well, except for two datagrids using HAVING in the filter.
For example, this request fails when counting with `count(select(1)...)` :

```
SELECT 
[columns_not_displayed_for the_example], 
CASE WHEN g.id IN (:associatedIds) THEN true ELSE false END AS is_associated
FROM Pim\Bundle\CatalogBundle\Entity\Group g LEFT JOIN g.translations translation WITH translation.locale = :dataLocale 
LEFT JOIN g.type typ 
LEFT JOIN typ.translations typTrans WITH typTrans.locale = :dataLocale 
HAVING NOT(is_associated <> false) 
ORDER BY is_associated DESC
```

When counting, it generates:
```
SELECT count(*) from (
SELECT 1
FROM Pim\Bundle\CatalogBundle\Entity\Group g 
LEFT JOIN g.translations translation WITH translation.locale = :dataLocale 
LEFT JOIN g.type typ LEFT JOIN typ.translations typTrans WITH typTrans.locale = :dataLocale 
HAVING NOT(is_associated <> false))
```

And it fails because HAVING is using an alias `is_associated`, but this alias does not exist anymore in the select. Do note that in this request we are using HAVING instead of WHERE, even if there is not GROUP BY.
Why ? Because that's the only way to filter on an alias! In a WHERE clause, alias is not possible.
Dirty hack, isn't it ?

Solution: we should replace alias in HAVING clause with the real SQL clause.
We should never use HAVING without GROUP BY as well...

And that's pitty that we could not do this easy optimisation, just because of several HAVING clauses 😞 

Trick: Want to know the datagrids using HAVING in filters ? Search `filter_by_having` in the project :)

Is the solution easy to implement (replacing alias by the real clause) ?
Actually, Oro is already replacing alias with "LIKE" operator, to fix a Doctrine <2.4 issue. So, I think it's doable. But not the purpose of this PR.

https://github.com/akeneo/pim-community-dev/blob/master/src/Oro/Bundle/FilterBundle/Datasource/Orm/OrmFilterDatasourceAdapter.php#L147

So finally, I let this count as it is, excepted that I've deleted the GROUP BY. Indeed, the count function should not eliminate duplicate rows. That's the responsibility of the original request, or it could lead to inconsistencies.

`OK`: Done / Validated
`-`: Not needed


jjanvier:
@ahocquard please keep me posted on this one
if it works without changing default MySQL config, that's really good for us
(and I need to update the documentation)

features/Context/CatalogConfigurationContext.php
```diff
@@ -58,6 +58,11 @@ public function aCatalogConfiguration($catalog)
         $this->loadCatalog($this->getConfigurationFiles($catalog));
         
         $this->getMainContext()->getContainer()->get('pim_connector.doctrine.cache_clearer')->clear();
+
+        $entityManager = $this->getContainer()->get('doctrine.orm.entity_manager');
+        $entityManager->getConnection()->exec(
+            'set global sql_mode=\'ONLY_FULL_GROUP_BY,STRICT_TRANS_TABLES,NO_ZERO_IN_DATE,NO_ZERO_DATE,ERROR_FOR_DIVISION_BY_ZERO,NO_AUTO_CREATE_USER,NO_ENGINE_SUBSTITUTION\''
```
ahocquard: again, don't be afraid, it will not be merged, just for testing

app/config/parameters.yml.dist
```diff
@@ -3,8 +3,8 @@ parameters:
     database_host:     localhost
     database_port:     ~
     database_name:     akeneo_pim
-    database_user:     akeneo_pim
-    database_password: akeneo_pim
+    database_user:     root
+    database_password: root
```
ahocquard: don't be afraid, it will not be merged, just for testing

src/Akeneo/Bundle/StorageUtilsBundle/Doctrine/ORM/Cursor/Cursor.php
```diff
@@ -120,7 +120,7 @@ protected function getEntitiesIds()
                 ->select($rootIdExpr)
                 ->resetDQLPart('from')
                 ->from($from->getFrom(), $from->getAlias(), $rootIdExpr)
-                ->groupBy($rootIdExpr);
+                ->distinct(true);
```
ahocquard: this class seems to not be used. @jjanvier do you remove it in your PR removing mongo ?

src/Pim/Bundle/CatalogBundle/Doctrine/ORM/Repository/FamilyRepository.php
```diff
@@ -118,15 +118,15 @@ public function countAll()
...
+            ->innerJoin('f.attributes', 'a')
             ->where('f.id = :id')
             ->andWhere('a.code = :code')
             ->setMaxResults(1)
             ->setParameters([
                 'id'   => $id,
                 'code' => $attributeCode,
-            ])
-            ->addGroupBy('a.id');
+            ]);
```
pierallard: You can add limit(0,1) to be more efficient

src/Pim/Bundle/EnrichBundle/Doctrine/ORM/Repository/JobExecutionRepository.php
```diff
@@ -46,10 +46,19 @@ public function createDatagridQueryBuilder()
...
+            ->groupBy('e')
+            ->addGroupBy('e.id')
+            ->addGroupBy('status')
+            ->addGroupBy('statusLabel')
+            ->addGroupBy('date')
+            ->addGroupBy('jobCode')
+            ->addGroupBy('jobLabel')
+            ->addGroupBy('jobName');
+
+
```
pierallard: Double line.
I still don't understand why you add all these groups, IMHO e.id is sufficient.

src/Pim/Bundle/CatalogBundle/Doctrine/ORM/Repository/FamilyRepository.php
ahocquard: Seen IRL: setMaxResult should do it.

src/Pim/Bundle/EnrichBundle/Doctrine/ORM/Repository/JobExecutionRepository.php
ahocquard: Seen IRL: it's a little bit "purist". Mysql 5.7.5 can detect dependencies in the column.It could be:
```
+        $qb
             ->groupBy('e')
             ->addGroupBy('jobCode')
             ->addGroupBy('jobLabel')
             ->addGroupBy('jobName');

src/Akeneo/Bundle/StorageUtilsBundle/Doctrine/ORM/Cursor/Cursor.php
jjanvier: not removed in my PR no
are you sure it's not used? we don't use cursors for other things that products?

src/Pim/Bundle/EnrichBundle/Doctrine/ORM/Repository/JobExecutionRepository.php
jjanvier: basic SQL
selected things must by grouped by

src/Akeneo/Bundle/StorageUtilsBundle/Doctrine/ORM/Cursor/Cursor.php
ahocquard: Ok, it's used by the versionning. But this class + its factory are declared 3 times with a different name. And it's used only one time for the service declaration of versioning.

PR #6308, opened 2017-06-20
Improve performance for rendering last jobs

Hocdoc:
Since updating from Akeneo 1.6 to 1.7 the rendering of the last jobs (last operations widget, Process tracker) got really slow for us. 
Our users have used the import/export feature extensively, so our MySQL job tables are huge and the new feature with counting the warnings take time. `akeneo_batch_warning` has about 4,000,000 rows.

We could use the `id` (primary key) instead of `startTime` (without any index) for the order to prevent a full table scan. 

Before this small change the query to get the last 25 jobs in the process tracker takes unbelievable 77 seconds. After the change the query takes 60ms (!) => it's now 1200x :-o faster for us.


pierallard:
IMHO, a huge better solution should be to add an index on startTime column.
In fact, nothing in the MySQL documentation implies that "id" are created in chronological order.

pierallard:
Hi @Hocdoc 
As this screens completely changed since 1.7, I will be obliged to refuse this PR and close it. I hope you understand!

Regards
Pierre

PR #6311, opened 2017-06-21
[PIM-6397] Enable search on grids

pierallard:
This PR will add a "search" input in every "simple" grid (all except product/published products/assets).

capture du 2017-06-21 11 33 42

# Why?

We made a lot of testing sessions with end-users, to check how then do a search on all the entities. And every end user searched for a word in the label. So we introduce a way to do a quicker search without clicking a lot of time, by introducing a "search input" searching for words in the label columns.
Agree with PO's, this solution is 100% sufficient and will do the job perfectly.

# How does it works ?

We just defined a new `SearchFilter`. This filter have the same behavior than the standard `TextFilter`, but it does a search without specifing operator (in, not in, equals...) and always put it as "LIKE".


src/Oro/Bundle/UserBundle/Resources/config/datagrid/roles.yml
```diff
@@ -1,62 +0,0 @@
-datagrid:
```
pierallard: This file was redundant with another one with the same content. So I dropped it. No link with this PR :)

PR #6343, opened 2017-06-28
Use official akeneo Docker images

damien-carcel:
## Description

Only PHP 5.6 is used.

## Definition Of Done

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -44,7 +44,7 @@ stage("Checkout") {
     checkouts['community'] = {
         node('docker') {
             deleteDir()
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -128,46 +120,40 @@ def runGruntTest() {
     }
 }
 
-def runPhpUnitTest(phpVersion) {
+def runPhpUnitTest() {
```
ahocquard: I would keep that parameter for the consistency through the branches.

damien-carcel: It will disappear for other branches too.

PR #6372, opened 2017-07-01
Add a "docker-compose.yml.dist" file

damien-carcel:
## Description

This PR adds the template of a Docker compose file, and its required configuration files.

It also add a helper script `bin/docker/pim-initialize.sh` to fully install the PIM (clean previous install relicates, install the database and dump the assets), and another one to prepare the front part (dump JS and CSS then run webpack).

## Definition Of Done

`OK`: Done / Validated
`-`: Not needed


src/Pim/Bundle/InstallerBundle/Command/AssetsCommand.php
```diff
@@ -18,6 +18,9 @@
  */
 class AssetsCommand extends ContainerAwareCommand
 {
+    /** @var CommandExecutor */
+    protected $commandExecutor;
```
anaelChardan: ???

bin/docker/pim-front.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+
+echo "Clean previous assets"
+
+rm -rf app/cache/*
```
ahocquard: Same comment as in EE.
Should be relative to the script.

bin/docker/pim-initialize.sh
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env bash
+
+echo "Clean previous install"
+
+rm -rf app/archive/*
```
ahocquard: same.
And be careful with "app/archive/*". It uses shell expansion. If you have too much files, it will fail. I suppose that's why you split the different lines.

You could use find.

PR #6373, opened 2017-07-01
Improve scalability of integration tests

ahocquard:

The integrations tests stack has several drawbacks:

- scalable but manually with test suites
- running time very heterogeneous on nodes (not the same number of tests between the test suites)
- some integration were not executed: adding a new directory in the tests should be declared in the phpunit.xml
- integration tests are not executed in EE => MISSION ABORTED

This PR aims to fix all of these problems.

### Scalability

There are about 250 integration files in CE, and about 1740 integration tests. Each test run in approximately 6s.
So, about 3 hours to run all the integration tests.

The first attempt was to execute in each node the same number of files. 
For example, with 10 nodes, 25 files per node.
But the result was that some nodes run with 80 tests and another one 375 tests.
So, some nodes were running in 8 minutes and the longest one in 35min.

Therefore, I choose to execute in each node approximately the same number of tests (not the same number of files).
With 10 nodes, each node run approximately 170 tests. So, approximately in 18 minutes.

### Maintenability

As said above, some integration tests were not executed because the path was not added in the phpunit.xml file.
Declare every tests in phpunit.xml is not maintainable.

It's very difficult to get a list of integration tests with phpunit. Trust me, you can Google it ;)
Moreover, it's very difficult to execute a list of class with phpunit. No command exist.

So, I've choosen to keep it simple:

- get a list of integration file and number of tests in each file, thanks to a shell command
- split it according to the number of nodes, each nodes should run approximately the same number of tests that the other => done in a groovy function

Now, the only thing to do is to add docker nodes if we want to speed up the execution time.
And change the variable `nbAvailableNode ` in the file. 

Pretty simple :)

### EE execution (aborted)

I've tried but it implies more stuff than expected.
For now, when I execute integration tests in EE with all the integration tests from CE, I have an error because the catalog could not be installed.
It could not because the path to the catalog returns the fixtures of the CE, but without the fixtures of the EE.

So, it fails with an exception from here https://github.com/akeneo/pim-community-dev/blob/master/src/Pim/Bundle/InstallerBundle/FixtureLoader/JobInstancesConfigurator.php#L79

`Exception: No replacement path for "asset_categories.csv"`

The job instances (containing job EE + CE) does not match the replace path (only CE path).

### Benchmark

|    Parallelisation strategy    | Time           |
| ------------- |:-------------:|
| Manual test suites (actual behavior) | 1h |
| Same number of files in each node | ~35m      |
| Same number of tests in each node | ~18m      |
| Same number of tests in each node + mysql in ram | ~6m |

Note: I've experienced a lot of instability on the CI's node: sometime some nodes run PHPUnit in 40 minutes (instead of 6m). Not always on the same node, not always on the same bunch of tests.

I think it's an already existing problem on the current stack, I will raise an issue.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,11 +1,15 @@
 #!groovy
 
 def editions = ["ce"]
-def phpVersion = "7.1"
-def features = "features"
-def launchUnitTests = "yes"
-def launchIntegrationTests = "yes"
-def launchBehatTests = "yes"
+String phpVersion = "7.1"
+String features = "features"
+String launchUnitTests = "yes"
+String launchIntegrationTests = "yes"
+String launchBehatTests = "yes"
+
+Integer nbAvailableNode = 8
+def testFilesCE = []
+def testFilesEE = []
 
 stage("Checkout") {
     milestone 1
@@ -32,14 +36,15 @@ stage("Checkout") {
         deleteDir()
         checkout scm
         stash "pim_community_dev"
+        deleteDir()
 
-        if (editions.contains('ee') && 'yes' == launchBehatTests) {
-           checkout([$class: 'GitSCM',
-             branches: [[name: 'master']],
-             userRemoteConfigs: [[credentialsId: 'github-credentials', url: 'https://github.com/akeneo/pim-enterprise-dev.git']]
-           ])
+        if (editions.contains('ee') && ('yes' == launchBehatTests || 'yes' == launchIntegrationTests)) {
+            checkout([$class: 'GitSCM',
+              branches: [[name: 'master']],
+              userRemoteConfigs: [[credentialsId: 'github-credentials', url: 'https://github.com/akeneo/pim-enterprise-dev.git']]
+            ])
 
-           stash "pim_enterprise_dev"
+            stash "pim_enterprise_dev"
         }
     }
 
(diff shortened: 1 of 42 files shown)


Jenkinsfile
```diff
@@ -99,6 +116,18 @@ stage("Checkout") {
                     stash "pim_enterprise_dev_full"
                 }
 
+                unstash "pim_enterprise_dev_full"
+
+                def output = sh (
+                    returnStdout: true,
+                    script: 'find src -name "*Integration.php" -exec sh -c "grep -Ho \'function test\' {} | uniq -c"  \\; | sed "s/:function test//"'
```
ahocquard: Just need to change it to `find src vendor/akeneo/pim-community-dev -name "*Integration.php" -exec sh -c "grep -Ho \'function test\' {} | uniq -c`
to execute CE tests later.

app/phpunit.xml.dist
```diff
@@ -29,59 +29,7 @@
 
         <!-- Complete suite for PIM integration tests -->
         <testsuite name="PIM_Integration_Test">
-            <directory suffix="Integration.php">../src/Pim/Component/*/tests/integration</directory>
-            <directory suffix="Integration.php">../src/Pim/Bundle/*Bundle/tests/integration</directory>
-            <directory suffix="Integration.php">../src/Akeneo/Bundle/*Bundle/tests/integration</directory>
+            <file></file>
```
damien-carcel: `<directory suffix="Integration.php">../src</directory>`? Or does it make no difference?

ahocquard: As the `<file></file>` will be replaced by the sed command with all the file, it doesn't matter.
It's just a pattern to replace.

PR #6384, opened 2017-07-10
Add a Docker compose file to the PIM

damien-carcel:
## Description

Add a `docker-compose.yml.dist` file, pre-configured. It uses `akeneo/apache-php:php-5.6` image for akeneo containers.

The commands `pim:install` and `pim:installer:assets` have been updated too, to add two new options:
- `symlink`: this will install the assets as relative symlinks instead of hard copy (present in both command as  `pim:install` and `pim:installer:assets`)
- `clean`: this will completely clean the artefacts of the previous pim install, or the previous assets install

## Definition Of Done

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/docker/akeneo-behat.conf
+++ b/docker/akeneo-behat.conf
@@ -1,0 +1,26 @@
+<VirtualHost *:80>
+    ServerName akeneo-behat.dev
+
+    DocumentRoot /srv/pim/web
+    <Directory /srv/pim/web>
+        AllowOverride None
+        Require all granted
+
+        Options -MultiViews
+        RewriteEngine On
+        RewriteCond %{REQUEST_FILENAME} !-f
+        RewriteRule ^(.*)$ app_behat.php [QSA,L]
+    </Directory>
+
+    <Directory /srv/pim>
+        Options FollowSymlinks
+    </Directory>
+
+    <Directory /srv/pim/web/bundles>
+        RewriteEngine Off
+    </Directory>
+
+    ErrorLog /var/log/akeneo-behat_error.log
+    LogLevel warn
+    CustomLog /var/log/akeneo-behat_access.log combined
+</VirtualHost>
--- a/docker/akeneo.conf
+++ b/docker/akeneo.conf
@@ -1,0 +1,26 @@
+<VirtualHost *:80>
+    ServerName akeneo.dev
+
+    DocumentRoot /srv/pim/web
+    <Directory /srv/pim/web>
+        AllowOverride None
+        Require all granted
+
+        Options -MultiViews
+        RewriteEngine On
+        RewriteCond %{REQUEST_FILENAME} !-f
+        RewriteRule ^(.*)$ app.php [QSA,L]
+    </Directory>
+
+    <Directory /srv/pim>
+        Options FollowSymlinks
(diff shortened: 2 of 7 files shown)


docker/akeneo.conf
```diff
@@ -0,0 +1,13 @@
+<VirtualHost *:80>
+    ServerName akeneo.dev
+
+    DocumentRoot /srv/pim/web
+    <Directory /srv/pim/web>
+        AllowOverride All
```
BitOne: It should be set to None, with the proper rewrite rules.

docker/akeneo-behat.conf
```diff
@@ -0,0 +1,17 @@
+<VirtualHost *:80>
+    ServerName akeneo-behat.dev
+
+    RewriteEngine On
+    RewriteCond %{DOCUMENT_ROOT}/%{REQUEST_FILENAME} !-f
+    RewriteRule ^(.*) %{DOCUMENT_ROOT}/app_behat.php [QSA,L]
+
+    DocumentRoot /srv/pim/web
+    <Directory /srv/pim/web>
+        AllowOverride All
```
BitOne: None with proper rewrite rules from .htaccess file from the PIM.

docker/akeneo.conf
damien-carcel: That is exactly what we have in the documentation, so it means the doc is wrong (and it is like that since a while) :/ .
https://docs.akeneo.com/latest/developer_guide/installation/system_requirements/system_install_debian8.html#creating-the-virtual-host-file

damien-carcel: With `AllowOverride None`, the PIM just doesn't work anymore.
Maybe it should be with `None`, and the htaccess file updated, but I think this is not the goal of this PR.

docker/akeneo-behat.conf
```diff
@@ -0,0 +1,30 @@
+<VirtualHost *:80>
+    ServerName akeneo-behat.dev
+
+    DocumentRoot /srv/pim/web
+    <Directory /srv/pim/web>
+        AllowOverride None
+        Require all granted
+
+        <IfModule mod_rewrite.c>
```
BitOne: To remove

docker/akeneo-behat.conf
```diff
@@ -0,0 +1,30 @@
...
+            RewriteRule ^(.*)$ app_behat.php [QSA,L]
+        </IfModule>
+    </Directory>
+
+    <Directory /srv/pim>
+        Options FollowSymlinks
+    </Directory>
+
+    <Directory /srv/pim/web/bundles>
+        <IfModule mod_rewrite.c>
```
BitOne: Idem

jonovate:
Unfortunately, I believe this PR forces pim:installer:assets to become symlinks only which is not good when you run php-fpm and do not want to give NGINX/proxy access to vendor folder.  Symfony doesn't seem to provide a way as its --symlink or no argument.

https://github.com/akeneo/pim-community-dev/commit/2207a3c928fcb2805386572cb2ab930034c51bd5#diff-db6bfa38ed26a4948efc67cd909edfb3R138

PR #6385, opened 2017-07-10
PIM-6526

grena:

This PR is to avoid a deadlock on completeness requests.
When we do a `INSERT INTO (...) SELECT FROM`, the `SELECT` is linked to a writing statement, so it does a  "hard" (exclusive) lock of the table. The same goes for `CREATE TEMPORARY TABLE (...) SELECT FROM`, as it's a writing operation too.

This PRs avoids this by executing the `SELECT` statement **before** doing the writing operation, this way, tables touched by the `SELECT` are not "hard locked" (soft lock).

-----
**Before:**
Duration (in milliseconds): 53366
Memory usage (in bytes): 36175872

**With this patch:**
Duration (in milliseconds): 67761 (**x1.2**)
Memory usage (in bytes): 36438016 (**x1.007**)

_Benchmark done with catalog medium, 10k products_.

-------

### DIFF
--- a/src/Pim/Bundle/CatalogBundle/Doctrine/ORM/CompletenessGenerator.php
+++ b/src/Pim/Bundle/CatalogBundle/Doctrine/ORM/CompletenessGenerator.php
@@ -42,6 +42,9 @@ class CompletenessGenerator implements CompletenessGeneratorInterface
     /** @var string */
     protected $attributeClass;
 
+    /** @var int */
+    protected $commitBatchSize;
+
     /**
      * Constructor
      *
@@ -49,14 +52,21 @@ class CompletenessGenerator implements CompletenessGeneratorInterface
      * @param string                 $productClass
      * @param string                 $productValueClass
      * @param string                 $attributeClass
+     * @param int                    $commitBatchSize
      */
-    public function __construct(EntityManagerInterface $manager, $productClass, $productValueClass, $attributeClass)
-    {
+    public function __construct(
+        EntityManagerInterface $manager,
+        $productClass,
+        $productValueClass,
+        $attributeClass,
+        $commitBatchSize = 1000
+    ) {
         $this->manager = $manager;
         $this->connection = $manager->getConnection();
         $this->productClass = $productClass;
         $this->productValueClass = $productValueClass;
         $this->attributeClass = $attributeClass;
+        $this->commitBatchSize = $commitBatchSize;
     }
 
     /**
@@ -141,22 +151,34 @@ SQL;
         }
         $fetchStmt->execute();
 
-        $this->connection->beginTransaction();
         $insertSql = <<<SQL
     INSERT INTO {$tempTableName} (locale_id, channel_id, value_id) 
     VALUES (?, ?, ?)
 SQL;
         $insertStmt = $this->connection->prepare($insertSql);
+        $count = 0;
 
         try {
             while ($completeness = $fetchStmt->fetch()) {
+                if ($count === 0) {
+                    $this->connection->beginTransaction();
+                }
+
                 $insertStmt->bindValue(1, $completeness['locale_id']);
                 $insertStmt->bindValue(2, $completeness['channel_id']);
                 $insertStmt->bindValue(3, $completeness['value_id']);
                 $insertStmt->execute();
+                $count++;
+
+                if ($count === $this->commitBatchSize) {
+                    $this->connection->commit();
+                    $count = 0;
+                }
             }
 
-            $this->connection->commit();
+            if ($count > 0) {
+                $this->connection->commit();
+            }
         } catch (\Exception $e) {
             $this->connection->rollBack();
             throw $e;
@@ -194,21 +216,33 @@ SQL;
         }
         $fetchStmt->execute();
 
-        $this->connection->beginTransaction();
         $insertSql = <<<SQL
     INSERT INTO {$tempTableName} (locale_id, channel_id, product_id) VALUES (?, ?, ?)
 SQL;
         $insertStmt = $this->connection->prepare($insertSql);
+        $count = 0;
 
         try {
             while ($completeness = $fetchStmt->fetch()) {
+                if ($count === 0) {
+                    $this->connection->beginTransaction();
+                }
+
                 $insertStmt->bindValue(1, $completeness['locale_id']);
                 $insertStmt->bindValue(2, $completeness['channel_id']);
                 $insertStmt->bindValue(3, $completeness['product_id']);
                 $insertStmt->execute();
+                $count++;
+
+                if ($count === $this->commitBatchSize) {
+                    $this->connection->commit();
+                    $count = 0;
+                }
             }
 
-            $this->connection->commit();
+            if ($count > 0) {
+                $this->connection->commit();
+            }
         } catch (\Exception $e) {
             $this->connection->rollBack();
             throw $e;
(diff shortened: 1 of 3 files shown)


src/Pim/Bundle/CatalogBundle/Doctrine/ORM/CompletenessGenerator.php
```diff
@@ -128,22 +121,48 @@ protected function prepareCompletePrices($criteria = [])
         $cleanupStmt = $this->connection->prepare($cleanupSql);
         $cleanupStmt->execute();
 
+        // Create temporary table
+        $createSQL = 'CREATE TEMPORARY TABLE ' .
```
ahocquard: I prefer heredoc for the readability

src/Pim/Bundle/CatalogBundle/Doctrine/ORM/CompletenessGenerator.php
```diff
@@ -128,22 +121,48 @@ protected function prepareCompletePrices($criteria = [])
         $cleanupStmt = $this->connection->prepare($cleanupSql);
         $cleanupStmt->execute();
 
+        // Create temporary table
```
BitOne: Useless comment :p

src/Pim/Bundle/CatalogBundle/Doctrine/ORM/CompletenessGenerator.php
```diff
@@ -128,22 +121,48 @@ protected function prepareCompletePrices($criteria = [])
...
+        $fetchStmt->execute();
 
-        $sql = $this->applyTableNames($sql);
+        // Fill in temporary table
+        $this->connection->beginTransaction();
+        $insertPattern = 'INSERT INTO %s (locale_id, channel_id, value_id) VALUES (%s, %s, %s)';
 
-        $stmt = $this->connection->prepare($sql);
+        try {
+            while ($completeness = $fetchStmt->fetch()) {
```
ahocquard: `false === $completeness = $fetchStmt->fetch()` (in case of empty array, etc)
Should not occur though.

src/Pim/Bundle/CatalogBundle/Doctrine/ORM/CompletenessGenerator.php
```diff
@@ -128,22 +121,48 @@ protected function prepareCompletePrices($criteria = [])
...
+        // Create temporary table
+        $createSQL = 'CREATE TEMPORARY TABLE ' .
+            self::COMPLETE_PRICES_TABLE .
+            ' (locale_id int, channel_id int, value_id int, primary key(locale_id, channel_id, value_id)) ';
+        $createSQL = $this->applyTableNames($createSQL);
+
+        $tempTableStmt = $this->connection->prepare($createSQL);
+        $tempTableStmt->execute();
+
+        // Execute the SELECT
```
BitOne: Idem

src/Pim/Bundle/CatalogBundle/Doctrine/ORM/CompletenessGenerator.php
```diff
@@ -128,22 +121,48 @@ protected function prepareCompletePrices($criteria = [])
...
-            ' (locale_id int, channel_id int, value_id int, primary key(locale_id, channel_id, value_id)) ' .
-            $sql;
+        $fetchStmt = $this->connection->prepare($sql);
+        foreach ($criteria as $placeholder => $value) {
+            $fetchStmt->bindValue($placeholder, $value);
+        }
+        $fetchStmt->execute();
 
-        $sql = $this->applyTableNames($sql);
+        // Fill in temporary table
```
BitOne: Idem
Could be a private method

src/Pim/Bundle/CatalogBundle/Doctrine/ORM/CompletenessGenerator.php
```diff
@@ -270,21 +315,48 @@ protected function applyCriteria($sql, $criteria)
...
+        }
+        $fetchStmt->execute();
+
+        $this->connection->beginTransaction();
+        $insertPattern = 'INSERT INTO pim_catalog_completeness ' .
+            '(locale_id, channel_id, product_id, ratio, missing_count, required_count) VALUES (%s, %s, %s, %s, %s, %s)';
+
+        try {
+            while ($completeness = $fetchStmt->fetch()) {
+                $insertSQL = sprintf(
```
BitOne: too bad to use a prepared statement then use string concatenation to use it ! Better prepare the statement once and use different parameter. That's what it has been done for ;)

src/Pim/Bundle/CatalogBundle/Doctrine/ORM/CompletenessGenerator.php
```diff
@@ -128,22 +121,48 @@ protected function prepareCompletePrices($criteria = [])
...
+                    self::COMPLETE_PRICES_TABLE,
+                    $completeness['locale_id'],
+                    $completeness['channel_id'],
+                    $completeness['value_id']
+                );
 
-        foreach ($criteria as $placeholder => $value) {
-            $stmt->bindValue($placeholder, $value);
+                $insertStmt = $this->connection->prepare($insertSQL);
+                $insertStmt->execute();
```
ahocquard: When inserting data in batch, you should prepare one time the request. Then insert the data. Actually, it avoids Mysql to pre-process the request each time and allow optimisation.
It's much more performant.

BitOne:
What about the performance differences with today code ? As we do far more PHP side stuff than before, it will have an impact.

Performances will certainly be less good, but it can be accepted as we are now far less lock heavy. But depends of the level of performance loss.

src/Pim/Bundle/CatalogBundle/Doctrine/ORM/CompletenessGenerator.php
```diff
@@ -158,22 +177,48 @@ protected function prepareMissingCompletenesses(array $criteria = [])
...
+            while ($completeness = $fetchStmt->fetch()) {
+                $insertSQL = sprintf(
+                    $insertPattern,
+                    $tempTableName,
+                    $completeness['locale_id'],
+                    $completeness['channel_id'],
+                    $completeness['product_id']
+                );
+
+                $insertStmt = $this->connection->prepare($insertSQL);
```
ahocquard: same = prepare statement one time. then insert.

src/Pim/Bundle/CatalogBundle/Doctrine/ORM/CompletenessGenerator.php
```diff
@@ -270,21 +315,48 @@ protected function applyCriteria($sql, $criteria)
...
+                    $insertPattern,
+                    $completeness['locale_id'],
+                    $completeness['channel_id'],
+                    $completeness['product_id'],
+                    $completeness['ratio'],
+                    $completeness['missing_count'],
+                    $completeness['required_count']
+                );
+
+                $insertStmt = $this->connection->prepare($insertSQL);
```
ahocquard: same

ahocquard:
Also, @BitOne , don't you think a batch size to commit could be a good idea?

src/Pim/Bundle/CatalogBundle/Doctrine/ORM/CompletenessGenerator.php
```diff
@@ -128,22 +131,58 @@ protected function prepareCompletePrices($criteria = [])
...
+        $fetchStmt = $this->connection->prepare($sql);
         foreach ($criteria as $placeholder => $value) {
-            $stmt->bindValue($placeholder, $value);
+            $fetchStmt->bindValue($placeholder, $value);
+        }
+        $fetchStmt->execute();
+
+        $insertSql = <<<SQL
+    INSERT INTO {$tempTableName} (locale_id, channel_id, value_id) 
+    VALUES (?, ?, ?)
```
BitOne: Can you use named parameter instead of "?", please ? It would make the reading and debugging easier.

PR #6445, opened 2017-07-20
API-php7-fix: Fix array to string conversion exception

momoss:

Since PHP 7.0 array to string conversion has been improved and now throw E_RECOVERABLE_ERROR instead of a E_NOTICE. [improve-array-to-string-conversion](https://why-cant-we-have-nice-things.mwl.be/requests/improve-array-to-string-conversion)
With the PIM API we can send invalid data like an array for a `pim_catalog_text` attribute instead string that was not an issue with PHP 5.6, before it thrown a E_NOTICE, and as the next PIM version has to support PHP 7.0, we have to fix it.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Pim/Component/Catalog/Comparator/Attribute/ScalarComparator.php
+++ b/src/Pim/Component/Catalog/Comparator/Attribute/ScalarComparator.php
@@ -40,6 +40,10 @@ class ScalarComparator implements ComparatorInterface
         $default = ['locale' => null, 'scope' => null, 'data' => null];
         $originals = array_merge($default, $originals);
 
+        if (is_array($data['data'])) {
+            return $data;
+        }
+
         return (string) $data['data'] !== (string) $originals['data'] ? $data : null;
     }
 }
--- a/src/Pim/Component/Catalog/spec/Comparator/Attribute/ScalarComparatorSpec.php
+++ b/src/Pim/Component/Catalog/spec/Comparator/Attribute/ScalarComparatorSpec.php
@@ -53,4 +53,36 @@ class ScalarComparatorSpec extends ObjectBehavior
 
         $this->compare($changes, $originals)->shouldReturn(null);
     }
+
+    function it_returns_value_when_value_is_array()
+    {
+        $changes = ['data' => ['scalar'], 'locale' => 'en_US', 'scope' => 'ecommerce'];
+        $originals = ['data' => 'scalar', 'locale' => 'en_US', 'scope' => 'ecommerce'];
+
+        $this->compare($changes, $originals)->shouldReturn($changes);
+    }
+
+    function it_returns_value_when_value_is_integer()
+    {
+        $changes = ['data' => 2, 'locale' => 'en_US', 'scope' => 'ecommerce'];
+        $originals = ['data' => 'scalar', 'locale' => 'en_US', 'scope' => 'ecommerce'];
+
+        $this->compare($changes, $originals)->shouldReturn($changes);
+    }
+
+    function it_returns_value_when_value_is_float()
+    {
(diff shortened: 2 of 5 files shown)


src/Pim/Component/Catalog/Comparator/Attribute/ScalarComparator.php
```diff
@@ -40,6 +40,10 @@ public function compare($data, $originals)
         $default = ['locale' => null, 'scope' => null, 'data' => null];
         $originals = array_merge($default, $originals);
 
+        if (is_array($data['data'])) {
```
phaseinducer: If I understand well, if `$originals['data']` is an array too and they are equal you should return `null`, no ?

momoss: Impossible to get an array here, it's a scalar value

phaseinducer: So how you can get an array in `$data['data']` ?

momoss: Look at the integration test ;-), you can send an array via the API

phaseinducer: Can you give a real example ?

ahocquard: Incoming data from user can be array, not the data from the database (if it's correctly validated).

momoss:
Yeah good idea, I'll make it now

src/Pim/Component/Catalog/Comparator/Attribute/ScalarComparator.php
phaseinducer: So isn't it the problem ? Why an array from the db is considered invalid while an array from user input is not ?

ahocquard: - The comparator is launched before the validator
- Data coming from the user can be with an incorrect format, as an array. So the comparator returns this array, as it could not be equals to the database data (that is always a scalar).
- Then the data is validated and the data are rejected because it's an array.

src/Pim/Component/Catalog/spec/Comparator/Attribute/ScalarComparatorSpec.php
```diff
@@ -53,4 +53,28 @@ function it_returns_null_when_values_are_the_same()
...
+        $this->compare($changes, $originals)->shouldReturn($changes);
+    }
+
+    function it_returns_value_when_value_is_float()
+    {
+        $changes = ['data' => 2.44, 'locale' => 'en_US', 'scope' => 'ecommerce'];
+        $originals = ['data' => 'scalar', 'locale' => 'en_US', 'scope' => 'ecommerce'];
+
+        $this->compare($changes, $originals)->shouldReturn($changes);
+    }
```
ahocquard: you can add a spec with null as `$changes` too please

src/Pim/Component/Catalog/spec/Comparator/Field/ScalarComparatorSpec.php
```diff
@@ -43,4 +43,27 @@ function it_returns_null_when_values_are_the_same()
...
+
+        $this->compare($changes, $originals)->shouldReturn($changes);
+    }
+
+    function it_returns_value_when_value_is_float()
+    {
+        $changes = [2.44];
+        $originals = ['scalar'];
+
+        $this->compare($changes, $originals)->shouldReturn($changes);
```
ahocquard: same here

PR #6450, opened 2017-07-21
API-213: Add missing case for many locales

solivier:

Workaround to check the permissions on the locales even if it's not an array, as the check if it's an array is in the pqb so far deeper in the code ....

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Pim/Bundle/ApiBundle/Controller/ProductController.php
+++ b/src/Pim/Bundle/ApiBundle/Controller/ProductController.php
@@ -200,7 +200,7 @@ class ProductController
         $queryParameters = array_merge([
             'limit' => $this->apiConfiguration['pagination']['limit_by_default']
         ], $request->query->all());
-        $pqbOptions = ['limit' => (int)$queryParameters['limit']];
+        $pqbOptions = ['limit' => (int) $queryParameters['limit']];
 
         $searchParameter = null;
         if (isset($queryParameters['search_after'])) {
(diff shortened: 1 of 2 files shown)


src/Pim/Bundle/ApiBundle/Controller/ProductController.php
```diff
@@ -540,6 +540,12 @@ protected function setPQBFilters(
 
                 if (isset($filter['locales'])) {
                     $context['locales'] = $filter['locales'];
```
anaelChardan: what is the meaning of this variable ?

solivier: the $context variable ?

anaelChardan: Ah yes, that is not what I was thinking about, my comment was more about in what the usage of `$localeCodes` is useful to instantiate but it does change the meaning ;).

src/Pim/Bundle/ApiBundle/Controller/ProductController.php
```diff
@@ -540,6 +540,12 @@ protected function setPQBFilters(
...
+                    $localeCodes = $context['locales'];
+                    if (!is_array($localeCodes)) {
+                        $localeCodes = [$context['locales']];
```
LaurentPetard: What if $context['locales'] is null or an empty string ? $localeCodes should not be an empty array ? If it's right so can just cast $context['locales'] as an array.

PR #6457, opened 2017-07-23
Display filepath instead of namespace of the failing scenarios of the CI

ahocquard:

Before the upgrade Behat 2 -> Behat 3, a custom formatter was doing the job to display filepath instead of namespace of the failing behat tests in the CI.

During the migration, we realized with @anaelChardan that is was not so easy to perform the same thing with Behat 3.
After investigation, it has to be done with an extension, and here it is.

screen shot 2017-07-30 at 23 52 55

This PR is heavily inspired from the Junit Formatter. There are some copy/past of classes, because most of the classes are final (not possible to extend), and they don't use interfaces (no DI, no decoration).

Note: what a pain to develop a Behat 3 extension. It is very poorly documented.

Note 2: last commit will not be merged, only for test.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/behat.ci.yml
+++ b/behat.ci.yml
@@ -89,7 +89,8 @@ default:
             factory:
                 page_parameters:
                     base_url: 'http://akeneo/'
+        Pim\Behat\Extension\PimFormatter\PimFormatterExtension: ~
 
     formatters:
-        junit:
+        pim:
             output_path: app/build/logs/behat/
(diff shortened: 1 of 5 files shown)


features/acl/assert_acl.feature
```diff
@@ -92,4 +92,4 @@ Feature: Define user rights
     And I save the role
     Then I should not see the text "There are unsaved changes."
     When I am on the products page
-    Then I should not see the text "2014 Collection"
+    Then I should see the text "2014 Collection"
```
ahocquard: not merged, only for testing purpose

features/acl/enforce_acl_on_history.feature
```diff
@@ -16,7 +16,7 @@ Feature: Enforce ACL on history
     And I save the role
     Then I should not see the text "There are unsaved changes."
     When I am on the "X_SELL" association type page
-    Then I should not see "history"
+    Then I should see "history"
```
ahocquard: same

PR #6461, opened 2017-07-24
TIP-298: use Elasticsearch 5.5

jjanvier:


damien-carcel:
Set also the behats. Look for the line `/var/lib/distributed-ci/dci-master/bin/build`, and replace in it `-e 5` by `-e 5.5`

PR #6464, opened 2017-07-24
[PIM-6503] Update completeness panel

pierallard:
This PR updates:
- the design of the new Completeness blocks
- the "inversion" of completeness blocks. Indeed, it was blocks sorted by locales displaying scopes; now it's blocks sorted by scope displaying locales.
- it removes the attribute list to only display the number of remaining ones.

selection_089


jjanvier:
maybe I don't understand something, but imo, being able to click on one of the missing attributes was the most interesting feature of this panel

do we have another way to perform the same kind of action?

pierallard:
@jjanvier That's an interesting question! Yes, there is another way to perform this, in a a way better way. Indeed, you will have in the PEF firectly in the attributes screen a selector so only see your "need to be filled" attributes (and other special selectors ;) )

jjanvier:
@pierallard ok, so we are sure to do it for the next release? is it a story planned for soon?

src/Pim/Bundle/EnrichBundle/Normalizer/CompletenessCollectionNormalizer.php
```diff
@@ -141,52 +145,66 @@ protected function countComplete(array $completenesses)
...
     protected function normalizeChannelCompletenesses(
         array $completenesses,
         $format,
-        array $context,
-        array $localeCodes
+        array $context
     ) {
         $normalizedCompletenesses = [];
 
         //TODO: workaround in order to handle behat empty completeness
```
phaseinducer: wtf is that ?

src/Pim/Bundle/EnrichBundle/Normalizer/CompletenessCollectionNormalizer.php
```diff
@@ -141,52 +145,66 @@ protected function countComplete(array $completenesses)
...
+     * @return string[]
+     */
+    protected function getChannelLabels(array $channels, array $locales, $channelCode)
+    {
+        $matchingChannels = array_filter($channels, function (ChannelInterface $channel) use ($channelCode) {
+            return $channel->getCode() === $channelCode;
+        });
+        $channel = array_shift($matchingChannels);
+
+        return array_reduce($locales, function ($result, LocaleInterface $locale) use ($channel) {
```
phaseinducer: I don't see the added value of doing a reduce instead of a foreach for example.

src/Pim/Bundle/EnrichBundle/Resources/public/js/product/form/completeness.js
```diff
@@ -70,16 +69,22 @@ define(
                             this.initialFamily = this.getFormData().family;
                         }
 
+                        console.log(this.sortCompleteness(this.getFormData().meta.completenesses));
```
phaseinducer: to remove ?

PR #6467, opened 2017-07-25
PIM-6220: Fix simple select attribute option sorting in products grid

willy-ahva:
When you add a simple select filter and the associated column in the grid, the sorting of the column was in terms of code first and not in terms of label first. 

In ODM, I added an `AttributeOptionSorter`.
Moreover, for the `pim_catalog.doctrine.query.sorter.is_associated` service, the class was the BaseSorter which has `isAttributeSupported() { return true; }`. The compiler pass `RegisterProductQuerySorterPass` added in the `SorterRegistry` the `is_associated` service with the highest priority (30). So when we were searching for a sorter, the first one was `is_associated` with `isAttributeSupported() { return true; }` and it was not possible to find another sorter.
No pb in ORM because the class for the `is_associated` service was a dedicated one.

| Added Behats                      | ok

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Pim/Bundle/CatalogBundle/spec/Doctrine/MongoDBODM/Sorter/AttributeOptionSorterSpec.php
+++ b/src/Pim/Bundle/CatalogBundle/spec/Doctrine/MongoDBODM/Sorter/AttributeOptionSorterSpec.php
@@ -1,0 +1,45 @@
+<?php
+
+namespace spec\Pim\Bundle\CatalogBundle\Doctrine\MongoDBODM\Sorter;
+
+use Doctrine\ODM\MongoDB\Query\Builder;
+use PhpSpec\ObjectBehavior;
+use Pim\Component\Catalog\Model\AttributeInterface;
+use Pim\Component\Catalog\Query\Sorter\AttributeSorterInterface;
+
+class AttributeOptionSorterSpec extends ObjectBehavior
+{
+    function let(Builder $queryBuilder)
+    {
+        $this->setQueryBuilder($queryBuilder);
+    }
+
+    function it_is_an_attribute_sorter()
+    {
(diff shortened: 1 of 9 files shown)


src/Pim/Bundle/CatalogBundle/spec/Doctrine/MongoDBODM/Sorter/IsAssociatedSorterSpec.php
```diff
@@ -0,0 +1,33 @@
...
+class IsAssociatedSorterSpec extends ObjectBehavior
+{
+    function let(Builder $queryBuilder)
+    {
+        $this->setQueryBuilder($queryBuilder);
+    }
+
+    function it_is_a_field_sorter()
+    {
+        $this->shouldImplement('Pim\Component\Catalog\Query\Sorter\FieldSorterInterface');
```
ahocquard: FieldSorterInterface::class

src/Pim/Bundle/CatalogBundle/spec/Doctrine/MongoDBODM/Sorter/AttributeOptionSorterSpec.php
```diff
@@ -0,0 +1,44 @@
...
+class AttributeOptionSorterSpec extends ObjectBehavior
+{
+    function let(Builder $queryBuilder)
+    {
+        $this->setQueryBuilder($queryBuilder);
+    }
+
+    function it_is_an_attribute_sorter()
+    {
+        $this->shouldImplement('Pim\Component\Catalog\Query\Sorter\AttributeSorterInterface');
```
ahocquard: AttributeSorterInterface::class

PR #6474, opened 2017-07-26
Add docker install scripts for 1.7

nidup:
Add simple scripts to install the PIM with docker in 1.7

These scripts are simpler than those on master but at least, makes the doc simpler, as a developer using docker, on any Akeneo version, i run the same scripts (even if they internally run different things, like npm on master). If i want to move to a more advanced use, i can look at the script and manually run internal commands.

In other words, as a dev using docker, I don't want to remind subtilities between 1.7 and 1.8 installations, I'm lazy, I run the same script ;)

**Definition Of Done**

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,6 +25,10 @@ vendor
 composer.phar
 /cov
 bin/*
+!/bin/docker
+/bin/docker/*
+!/bin/docker/pim-initialize.sh
+!/bin/docker/pim-front.sh
 !/bin/merge-coverage
 behat.yml
 /phpspec.yml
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -25,6 +25,10 @@ vendor
 composer.phar
 /cov
 bin/*
```
damien-carcel: /bin/* (very bad practice imo not to put a `/` at the beginning in .gitignore, except if we really meant it like that).

nidup: copy / pasted from master but ok, no pb :)

damien-carcel: If I did that on master, then I am a bad boy!

PR #6487, opened 2017-07-28
PIM-6448: Link product variants to family variants

jjanvier:
We see in our different stories that we have some difficulties to work with products, variants and models. Our design is not clear enough.

After a discussion with @aRn0D and @grena we decided to clearly make the distinction between "regular" products en variant products by:
- keeping the `ProductInterface` as it was before (aka same as 1.7)
- introducing a `VariantProductInterface` (this interface extends `ProductInterface` + `EntityWithFamilyVariantInterface`)
- splitting products and variant products into 2 classes
- changing the `CanHaveFamilyVariantInterface` to `EntityWithFamilyVariantInterface` => `ProductModelInterface` and `VariantProductInterface` implement this interface

The split of the `Product` and `VariantProduct` entities is made via the [Single Table Inheritance](http://docs.doctrine-project.org/projects/doctrine-orm/en/latest/reference/inheritance-mapping.html#single-table-inheritance) principle of Doctrine. That means 
- products and variant products are still stored in the same table `pim_catalog_product`. They are distinguished by the column `product_type`.
- `Pim\Component\Catalog\Model\AbstractProduct` becomes a [Mapped Superclasse](http://docs.doctrine-project.org/projects/doctrine-orm/en/latest/reference/inheritance-mapping.html#mapped-superclasses)


src/Pim/Component/Catalog/FamilyVariant/EntityWithFamilyVariantAttributesProvider.php
```diff
@@ -49,12 +49,11 @@ public function getAxes(EntityWithFamilyVariantInterface $entityWithFamilyVarian
     {
         $familyVariant = $entityWithFamilyVariant->getFamilyVariant();
 
-        if (null === $familyVariant || $entityWithFamilyVariant->isRootVariation()) {
+        $level = $entityWithFamilyVariant->getVariationLevel();
+        if (null === $familyVariant || 0 === $level) {
```
jjanvier: constant for ROOT LEVEL

src/Pim/Component/Catalog/Model/VariantProduct.php
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function setProductModel(ProductModelInterface $productModel): ProductInterface
+    {
+        $this->productModel = $productModel;
+
+        return $this;
```
aRn0D: void?

jjanvier: done

src/Pim/Component/Catalog/Model/VariantProduct.php
```diff
@@ -0,0 +1,60 @@
...
+    /** @var ProductModelInterface $productModel */
+    protected $productModel;
+
+    /** @var FamilyVariantInterface */
+    protected $familyVariant;
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getProductModel(): ?ProductModelInterface
```
grena: When could it be null?

jjanvier: new `VariantProduct`
and you have not set anything :( 
as long as the variant is not validated, it can be null

PR #6488, opened 2017-07-28
[PIM-6670] Optimize attributes normalization

phaseinducer:
**The context**
We need to load the PEF as fast as possible.

**The problem**
Among the most blocking requests : the one getting the list of attributes and the one getting the family. In Enrich, the normalized attributes are embed in the family. It appeared that the normalization of attributes was the common bottleneck.

**The solution**
In the current enrich attribute normalizer we set `created`, `updated`, `structure_version` and `model_type` keys. None of these are used in the PEF, so I split the normalizer in two : `AttributeNormalizer` and `VersionedAttributeNormalizer` (not sure about the naming), the second one decorating the first one. Then use the new one in the family normalizer and in the index route of the attribute rest controller.

For a PEF with 200 attributes :

Request | Before | After
--- | -------- | ------
fetch family | 2,4s | 320ms
fetch attributes | 2,2s | 300ms

### DIFF
--- a/src/Pim/Bundle/EnrichBundle/Normalizer/AttributeNormalizer.php
+++ b/src/Pim/Bundle/EnrichBundle/Normalizer/AttributeNormalizer.php
@@ -6,8 +6,6 @@ use Akeneo\Component\Localization\Localizer\LocalizerInterface;
 use Pim\Bundle\EnrichBundle\Provider\EmptyValue\EmptyValueProviderInterface;
 use Pim\Bundle\EnrichBundle\Provider\Field\FieldProviderInterface;
 use Pim\Bundle\EnrichBundle\Provider\Filter\FilterProviderInterface;
-use Pim\Bundle\EnrichBundle\Provider\StructureVersion\StructureVersionProviderInterface;
-use Pim\Bundle\VersioningBundle\Manager\VersionManager;
 use Pim\Component\Catalog\Model\AttributeInterface;
 use Symfony\Component\Serializer\Normalizer\NormalizerInterface;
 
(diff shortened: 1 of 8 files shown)


src/Pim/Bundle/EnrichBundle/Normalizer/VersionedAttributeNormalizer.php
```diff
@@ -0,0 +1,85 @@
...
+     */
+    public function normalize($attribute, $format = null, array $context = [])
+    {
+        $normalizedAttribute = $this->normalizer->normalize($attribute, 'internal_api', $context);
+
+        $firstVersion = $this->versionManager->getOldestLogEntry($attribute);
+        $lastVersion = $this->versionManager->getNewestLogEntry($attribute);
+
+        $firstVersion = null !== $firstVersion ?
+            $this->versionNormalizer->normalize($firstVersion, 'internal_api') :
```
pierallard: Don't have to pass $context value here ?

src/Pim/Bundle/EnrichBundle/Normalizer/AttributeNormalizer.php
```diff
@@ -52,28 +41,19 @@ class AttributeNormalizer implements NormalizerInterface
      * @param FieldProviderInterface            $fieldProvider
      * @param EmptyValueProviderInterface       $emptyValueProvider
      * @param FilterProviderInterface           $filterProvider
-     * @param VersionManager                    $versionManager
-     * @param NormalizerInterface               $versionNormalizer
-     * @param StructureVersionProviderInterface $structureVersionProvider
```
damien-carcel: Update the bc breaks in changelog please.

src/Pim/Bundle/EnrichBundle/Normalizer/VersionedAttributeNormalizer.php
phaseinducer: Nice catch !

src/Pim/Bundle/EnrichBundle/Normalizer/AttributeNormalizer.php
phaseinducer: done

PR #6520, opened 2017-08-07
Api 332: logs of the import/export should be in a configurable directory for the saas

ahocquard:

The log files created by the export/import are currently written in a directory that is not configurable in the "pim_parameters". 
It prevents to download a log file across several instances of the PIM.

In order to allow horizontal scalability, the directory should be configurable.
This way, the cloud team can configure the path to a shared directory between all the containers/instances of the PIM.

Note: for now, batch_execute.log is written in the configured repository. It will generate problems as the name is not depending of the job execution id, in case of concurrency access to the file across several instances of the PIM.

This case will be handle in another PR, with a "batch execute" file for each job execution. 

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/app/config/pim_parameters.yml
+++ b/app/config/pim_parameters.yml
@@ -16,8 +16,9 @@ parameters:
 
     upload_dir:           '%kernel.root_dir%/uploads/product'
     archive_dir:          '%kernel.root_dir%/archive'
+    logs_dir:             '%kernel.logs_dir%'
     catalog_storage_dir:  '%kernel.root_dir%/file_storage/catalog'
-    tmp_storage_dir:      /tmp/pim/file_storage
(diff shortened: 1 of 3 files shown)


app/config/pim_parameters.yml
```diff
@@ -16,6 +16,7 @@ parameters:
 
     upload_dir:           '%kernel.root_dir%/uploads/product'
     archive_dir:          '%kernel.root_dir%/archive'
+    logs_dir:             '%kernel.logs_dir%'
     catalog_storage_dir:  '%kernel.root_dir%/file_storage/catalog'
     tmp_storage_dir:      /tmp/pim/file_storage
```
solivier: why no quotes here ?

ahocquard: Forgot it, it's legacy, I will fix it

PR #6542, opened 2017-08-10
PIM-6333: rework getValue method

aRn0D:

For product models and variant products, the method `getValues` will return its values and the parent ones. To only get the values at the current level, you need to call `getValuesForVariation` method.

### DIFF
--- a/src/Pim/Bundle/CatalogBundle/EventSubscriber/LoadEntityWithValuesSubscriber.php
+++ b/src/Pim/Bundle/CatalogBundle/EventSubscriber/LoadEntityWithValuesSubscriber.php
@@ -6,7 +6,6 @@ use Doctrine\Common\EventSubscriber;
 use Doctrine\Common\Persistence\Event\LifecycleEventArgs;
 use Doctrine\ORM\Events;
 use Pim\Component\Catalog\Factory\ValueCollectionFactory;
-use Pim\Component\Catalog\Model\EntityWithFamilyVariantInterface;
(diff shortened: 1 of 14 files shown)


src/Pim/Component/Catalog/Model/EntityWithFamilyVariantInterface.php
```diff
@@ -37,4 +39,9 @@ public function getVariationLevel(): int;
      * @return ProductModelInterface|null
      */
     public function getParent(): ?ProductModelInterface;
+
+    /**
+     * @return Collection
```
jjanvier: ValueCollectionInterface

src/Pim/Component/Catalog/Model/EntityWithFamilyVariantInterface.php
```diff
@@ -2,6 +2,8 @@
 
 namespace Pim\Component\Catalog\Model;
 
+use Doctrine\Common\Collections\Collection;
```
jjanvier: to remove

src/Pim/Bundle/CatalogBundle/EventSubscriber/LoadEntityWithValuesSubscriber.php
```diff
@@ -81,31 +77,12 @@ public function postLoad(LifecycleEventArgs $event)
     /**
      * @return ValueCollectionFactory
      */
-    private function getProductValueCollectionFactory(): ValueCollectionFactory
```
jjanvier: why?

PR #6544, opened 2017-08-11
API-324: Convert label option to mandatory argument in  command

momoss:

Label is now a required argument (before it was an option) by the command and mandatory when you create a client with the command line

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Pim/Bundle/ApiBundle/Command/CreateClientCommand.php
+++ b/src/Pim/Bundle/ApiBundle/Command/CreateClientCommand.php
@@ -4,6 +4,7 @@ namespace Pim\Bundle\ApiBundle\Command;
 
 use OAuth2\OAuth2;
 use Symfony\Bundle\FrameworkBundle\Command\ContainerAwareCommand;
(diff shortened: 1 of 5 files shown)


src/Pim/Bundle/ApiBundle/Command/CreateClientCommand.php
```diff
@@ -60,8 +60,8 @@ protected function execute(InputInterface $input, OutputInterface $output)
         $client->setRedirectUris($input->getOption('redirect_uri'));
         $client->setAllowedGrantTypes($input->getOption('grant_type'));
 
-        if ($input->hasOption('label')) {
-            $client->setLabel($input->getOption('label'));
+        if ($input->hasArgument('label')) {
```
ahocquard: useless check btw as it's mandatory now (and moreover `hasArgument` is always true, this function has a bad naming, it would be `null !== getArgument()`).

ahocquard:
It will maybe impact the PHP client of the API.

PR #6554, opened 2017-08-15
Enable web-server bundle

jjanvier:
From Symfony's [upgrade documentation](https://github.com/symfony/symfony/blob/master/UPGRADE-3.3.md#frameworkbundle):

> The server:run, server:start, server:stop and server:status console commands have been moved to a dedicated bundle. Require symfony/web-server-bundle in your composer.json and register Symfony\Bundle\WebServerBundle\WebServerBundle in your AppKernel to use them.

The bundle is already required by composer via `symfony/symfony`.

I use a lot those commands in dev and test environments.


jjanvier:
2 unstable failing on CE..
I launched the build 3 times, the failures are always differrent

PR #6556, opened 2017-08-16
Fix "symlink" and "clean" options on assets and install command

damien-carcel:
## Description

Two new options were recently added to the "pim:install" and "pim:installer:assets" commands:
- `clean` that ensure the removal of all folders that were not present when cloning the PIM from GitHub
- `symlink` that dump assets as relative symlinks

An error has been made in the option detection: the value returned by `InputInterface::getOption` was tested against `null` when it returns a boolean value.

## Definition Of Done

`OK`: Done / Validated
`-`: Not needed


SamirBoulil:
What about EE ?

damien-carcel:
No need on EE, these commands are not overridden.

PR #6560, opened 2017-08-16
PIM-6568: Fix import items invalid data that were no longer saved

LaurentPetard:

When an import fails, the details of the invalid items are no longer displayed on the import job execution details page. (clicking on the link "Display item" does not display anything)

It's because the function `DoctrineJobRepository::insertWarning` expects an array of the invalid data of the item, and not the item itself. No error was raised because the parameter `$item` is not typed. The signature of this method will be fixed on master, because we have to add this method to the interface `JobRepositoryInterface`.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/features/import/view_import_items_invalid_data.feature
+++ b/features/import/view_import_items_invalid_data.feature
@@ -1,0 +1,26 @@
+@javascript
+Feature: Handle import of invalid data
+  In order to correct an import job that failed
+  As a product manager
+  I need to be able to see the content of the items of the import that have invalid data
+
+  Background:
+    Given a "footwear" catalog configuration
+
+  Scenario: Display items of a products import failed
+    Given I am logged in as "Julia"
+    And the following CSV file to import:
+      """
+      sku;family;handmade
+      SKU-001;NO_FAMILY;1
+      SKU-003;sneakers;0
+      """
+    And the following job "csv_footwear_product_import" configuration:
+      | filePath | %file to import% |
+    When I am on the "csv_footwear_product_import" import job page
+    And I launch the import job
+    And I wait for the "csv_footwear_product_import" job to finish
+    And I follow "Display item"
+    Then I should see the text "Warning"
+    And I should see the text "Property \"family\" expects a valid family code. The family does not exist, \"NO_FAMILY\" given."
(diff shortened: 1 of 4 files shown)


src/Akeneo/Component/Batch/spec/Step/ItemStepSpec.php
```diff
@@ -101,7 +101,7 @@ function it_executes_with_an_invalid_item_during_processing(
         );
 
         $repository
-            ->insertWarning(Argument::cetera())
+            ->insertWarning($execution, 'my msg', [], ['r4'])
```
ahocquard: you should mock the exception thrown instead (legacy code of the spec is wrong).

features/import/view_import_items_invalid_data.feature
```diff
@@ -0,0 +1,26 @@
...
+      """
+    And the following job "csv_footwear_product_import" configuration:
+      | filePath | %file to import% |
+    When I am on the "csv_footwear_product_import" import job page
+    And I launch the import job
+    And I wait for the "csv_footwear_product_import" job to finish
+    And I follow "Display item"
+    Then I should see the text "Warning"
+    And I should see the text "Property \"family\" expects a valid family code. The family does not exist, \"NO_FAMILY\" given."
+    And I should see the text "{\"sku\":[{\"locale\":null,\"scope\":null,\"data\":\"SKU-001\"}],\"handmade\":[{\"locale\":null,\"scope\":null,\"data\":true}]}"
```
ahocquard: It should be done with a integration test in theory. As the stack to test the import is not done in 1.7 with phpunit, I'm totally ok with behat.
But when merging on master, we should transform it as an integration test.

LaurentPetard: Ok for an integration test to be sure that the item data are persisted, but the functionality "Display item" (click on a link to display/hide infos) should be tested too no ? And it can't be only done with Behat

ahocquard: You're right. But we could imagine behats that are merged together to test the UI and the different warnings/errors. And a phpunit for each error. This way, the granularity of the tests is on the integration tests side (much more reliable). 

Anyway, no need of phpunit in this PR.

src/Akeneo/Component/Batch/spec/Step/ItemStepSpec.php
LaurentPetard: Unfortunately, the method `getMessage` is called, and it's a "final" method that can't be mocked.

PR #6565, opened 2017-08-16
PIM-6744: Fix field configuration of xlsx product import job edit form

LaurentPetard:

On the edit form of the import jobs based on "xlsx_product_import", the field "Date format" is setted as "readonly" whereas it should not.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Pim/Bundle/EnrichBundle/Resources/config/form_extensions/job_instance/xlsx_product_import_edit.yml
+++ b/src/Pim/Bundle/EnrichBundle/Resources/config/form_extensions/job_instance/xlsx_product_import_edit.yml
@@ -83,7 +83,7 @@ extensions:
         targetZone: global-settings
         config:
             fieldCode: configuration.dateFormat
-            readOnly: true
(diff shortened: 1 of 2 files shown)


src/Pim/Bundle/EnrichBundle/Resources/config/form_extensions/job_instance/xlsx_product_import_edit.yml
```diff
@@ -83,7 +83,7 @@ extensions:
         targetZone: global-settings
         config:
             fieldCode: configuration.dateFormat
-            readOnly: true
+            readOnly: false
```
phaseinducer: You can directly drop the line imo.

LaurentPetard: Yes, it was to be consistent with the other fields (they have all this line). Or should I remove `readOnly: false` for all the fields ?

PR #6570, opened 2017-08-17
Make attribute option field longer

fitn:

The current validation is not overridable and, since 1.7, it breaks the connector with CNET for attribute option labels length.
Discussing with @jjanvier , I propose this PR to make attribute option translations longer.

I will do a second PR on next release to homogeneize translations fields and fix validation issues.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/CHANGELOG-1.7.md
+++ b/CHANGELOG-1.7.md
@@ -4,6 +4,10 @@
 
 - PIM-6568: Fix import items invalid data that were no longer saved
 
+## Improvements
+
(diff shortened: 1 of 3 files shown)


damien-carcel:
The risk is that if they do not update the schema, they can import (for example) data with more than 100 characters, PHP validation will be OK, but MySQL will not. Now, maybe Doctrine handles that very well and it is not a problem at all.

PR #6573, opened 2017-08-17
PIM-6333: Split the import product model step

aRn0D:

Product models are like categories, a tree. We decided to create 2 steps to avoid to:
* use batch size equals to 1 
* some errors related CSV file order. For instance, it avoids error when you try to link a sub product a root product model which is in memory but not flushed.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/features/import/product_model/import_business_rules.feature
+++ b/features/import/product_model/import_business_rules.feature
@@ -117,3 +117,22 @@ Feature: Create product through CSV import
     And I launch the import job
     And I wait for the "csv_catalog_modeling_product_model_import" job to finish
     Then the product model "code-002" should not have the following values "collection, description-en_US-ecommerce, erp_name-en_US, price"
+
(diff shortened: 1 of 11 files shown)


damien-carcel:
Maybe adding a behat on creation through import, with a sub model first in the file, before its parent?

features/import/product_model/import_business_rules.feature
```diff
@@ -119,3 +117,22 @@ Feature: Create product through CSV import
     And I launch the import job
     And I wait for the "csv_catalog_modeling_product_model_import" job to finish
     Then the product model "code-002" should not have the following values "collection, description-en_US-ecommerce, erp_name-en_US, price"
+
+  Scenario: Import a file regardless the product model order, fist the root product model are imported then the sub product model
```
SamirBoulil: `fist` => first

PR #6574, opened 2017-08-17
TIP-803: Replace npm with yarn

tamarasaurus:

This PR switches to [Yarn](https://yarnpkg.com/lang/en/) from npm for our front-end dependencies. The main reasons for this:

### Speed improvements
Installing our dependencies cold (no cache) with npm takes **64.207s** on average, and with the modules cached, this is reduced to **43.043s**. With yarn, installing cold takes **50.56s** and installing with cached modules takes **15.64s** (tested 4 times on my 2015 13-inch Macbook, using office wifi ~35mbps). Yarn has a few advantages over npm here - parallel downloads, automatic retries, offline package install.

### Ongoing issues with npm
We upgraded to npm5 (5.0.3) to introduce a lockfile into the repo. This brought us some recurring issues with packages not being installed after deleting node_modules and re-running `npm install` (https://github.com/npm/npm/issues/16839). This bug is still happening in 5.3.0 (the latest release). Another issue is that the lockfile is not working as expected, for example on every npm install the integrity hashes change in the package-lock.json based on the user's OS. See here for more: https://github.com/npm/npm/issues/17749

### Changes
- [x] Use yarn instead of npm everywhere (including Jenkinsfile and docker scripts)
- [x] Share yarn cache between builds in the CI
- [x] Update the UPGRADE-1.8 docs
- [x] Add a [helper script](https://github.com/akeneo/pim-community-dev/blob/61157d32e242d9f814a59c4a4ad4d763e07dd54b/webpack/check-dependencies.js) to automatically sync outdated node dependencies for EE and standard editions.  
- [x] Update ansible roles for yarn https://github.com/akeneo/ansible/pull/231

When a user of the standard editions (or EE) has a package.json that is outdated with CE, `check-dependencies` will sync them up before building webpack. This can cover the cases where we add/update a dependency in a patch. 

E.g.
yarn

### To do

- [ ] Update the system requirements/installation instructions 

`OK`: Done / Validated
`-`: Not needed


damien-carcel:
You'll need also to:
- update the two scripts in `bin/docker/` (just replace `npm install` with `yarn install`)
- make the same PR on EE, of course 😅

For me, we provide the yarn.lock and recommend its usage, but that doesn't prevent users to use `npm` (with the risk to breack front things if fresher packages have a problem).

Concerning the CI, the cache should be sharable, the same way we did it for composer. We run the PHP container with:
```
docker.image("akeneo/php:${phpVersion}").inside("-v /home/akeneo/.composer:/home/docker/.composer -e COMPOSER_HOME=/home/docker/.composer")
```

The part `-v /home/akeneo/.composer:/home/docker/.composer` allows to share the composer folder from the host with the container. We can do the same for node (I do it on my machine):
```
docker.image('node:8').inside("-v /home/akeneo/.npm:/home/node/.npm)
```

nidup:
Regarding the open question allowing yarn and npm, we have to support only one, the one we choose and that's it, there is no benefits of supporting 2 package managers.

Jenkinsfile
```diff
@@ -107,10 +107,10 @@ stage("Checkout") {
                     stash "pim_enterprise_dev_full"
                 }
 
-                docker.image('node:8').inside {
+                docker.image('node:8').inside("-v /home/akeneo/.yarn-cache:/home/node/.yarn-cache") {
```
damien-carcel: @tamarasaurus Beware, I just test Yarn with this Docker image: cache and config are set "linux style":
```
- /home/akeneo/.cache/yarn:/home/node/.cache/yarn
- /home/akeneo/.config/yarn:/home/node/.config/yarn
```

I think ".yarn-cache" is specific to MacOS X.

jjanvier:
about this point
>  Update the system requirements/installation instructions

we should *NOT* update the [System Requirements](https://docs.akeneo.com/master/install_pim/system_requirements/system_requirements.html) page imo 
but should rather add a "Developer Requirements" section, or something like that

Jenkinsfile
```diff
@@ -62,11 +62,13 @@ stage("Checkout") {
                 stash "pim_community_dev_full"
             }
 
-            docker.image('node:8').inside {
+            sh "mkdir -p /home/akeneo/.yarn-cache"
+
+            docker.image('node:8').inside("-v /home/akeneo/.yarn-cache:/home/node/.yarn-cache") {
```
damien-carcel: Everytime you map the Yarn cache folder, you also need to pass to the container this environment variable: `YARN_CACHE_FOLDER=/home/node/.yarn-cache`
This will force Yarn to use `yarn-cache` as it cache folder, otherwise it will continue to use `.cache/yarn` in the container, and start each build with an empty cache.

damien-carcel:
Could you also update the `docker-compose.yml.dist` file in your PR, please?
Only the `node` service is to update, and it should become this:
```yaml

  node:
    image: node:8
    environment:
      YARN_CACHE_FOLDER: '/home/node/.yarn-cache'
    user: node
    volumes:
      - ./:/srv/pim
      - ~/.npm:/home/node/.npm
      - ~/.cache/yarn:/home/node/.yarn-cache
    working_dir: /srv/pim
    networks:
      - akeneo
      - behat
```

composer.json
```diff
@@ -85,6 +85,7 @@
         "pdepend/pdepend": "2.1.*",
         "phpmd/phpmd": "1.*",
         "phpspec/phpspec": "3.0.*",
+        "phpspec/prophecy": "1.7.0",
```
tamarasaurus: remove

.gitignore
```diff
@@ -52,5 +52,8 @@ npm-debug.log
 !/docker/akeneo-behat.conf
 !/docker/httpd.conf
 docker-compose.yml
+
 yarn.lock
 package-lock.json
+.yarn-metadata.json
+yarn-error.log
```
damien-carcel: missing EOF ;-)

tamarasaurus: fixed

SamirBoulil:
I think we need to update our ansible deployement scenarios to use yarn instead of NPM right ? https://github.com/akeneo/ansible/tree/symony3-upgrade

tamarasaurus:
@SamirBoulil Updated ansible here - https://github.com/akeneo/ansible/pull/231

damien-carcel:
Please update the script `bin/docker/pim-dependencies.sh` too ;-).

tamarasaurus:
@damien-carcel Fixed, thanks!

PR #6584, opened 2017-08-18
API-334: store job parameters into database

ahocquard:

When you launch a job instance, a job execution is created.

The configuration  of this job execution  can be overridden on the fly, and there is no way to find the configuration in database because it's not stored in DB.

To be able to launch job executions in background, we have to:
- make a job execution immutable: not possible to specify a config when launching a job execution, only when we launch a job instance

- store the job parameters in database

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Akeneo/Bundle/BatchBundle/Command/BatchCommand.php
+++ b/src/Akeneo/Bundle/BatchBundle/Command/BatchCommand.php
@@ -1,17 +1,23 @@
 <?php
 
+declare(strict_types=1);
+
 namespace Akeneo\Bundle\BatchBundle\Command;
 
+use Akeneo\Bundle\BatchBundle\Notification\MailNotifier;
 use Akeneo\Component\Batch\Event\BatchCommandEvent;
 use Akeneo\Component\Batch\Event\EventInterface;
 use Akeneo\Component\Batch\Item\ExecutionContext;
 use Akeneo\Component\Batch\Job\ExitStatus;
+use Akeneo\Component\Batch\Job\JobParameters;
 use Akeneo\Component\Batch\Job\JobParametersFactory;
 use Akeneo\Component\Batch\Job\JobParametersValidator;
 use Akeneo\Component\Batch\Job\JobRegistry;
+use Akeneo\Component\Batch\Model\JobInstance;
 use Akeneo\Component\Batch\Model\StepExecution;
-use Doctrine\ORM\EntityManager;
-use Monolog\Handler\StreamHandler;
+use Doctrine\Common\Util\ClassUtils;
+use Doctrine\ORM\EntityManagerInterface;
+use Symfony\Bridge\Monolog\Handler\ConsoleHandler;
 use Symfony\Bundle\FrameworkBundle\Command\ContainerAwareCommand;
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
@@ -19,7 +25,7 @@ use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Validator\Constraints as Assert;
 use Symfony\Component\Validator\ConstraintViolationList;
-use Symfony\Component\Validator\Validator;
+use Symfony\Component\Validator\Validator\ValidatorInterface;
 
 /**
  * Batch command
@@ -80,16 +86,16 @@ class BatchCommand extends ContainerAwareCommand
 
         if (!$noLog) {
             $logger = $this->getContainer()->get('monolog.logger.batch');
-            // Fixme: Use ConsoleHandler available on next Symfony version (2.4 ?)
-            $logger->pushHandler(new StreamHandler('php://stdout'));
+            $logger->pushHandler(new ConsoleHandler($output));
         }
 
         $code = $input->getArgument('code');
         $jobInstanceClass = $this->getContainer()->getParameter('akeneo_batch.entity.job_instance.class');
-        $jobInstance = $this
-            ->getJobManager()
-            ->getRepository($jobInstanceClass)
-            ->findOneByCode($code);
+        $jobInstance = $this->getJobManager()->getRepository($jobInstanceClass)->findOneBy(['code' => $code]);
+
+        if (null === $jobInstance) {
+            throw new \InvalidArgumentException(sprintf('Could not find job instance "%s".', $code));
+        }
 
         $event = new BatchCommandEvent($this, $input, $output, $jobInstance);
         $this->getContainer()->get('event_dispatcher')->dispatch(EventInterface::BATCH_COMMAND_START, $event);
@@ -97,19 +103,7 @@ class BatchCommand extends ContainerAwareCommand
             return self::EXIT_ERROR_CODE;
         }
 
-        if (!$jobInstance) {
-            throw new \InvalidArgumentException(sprintf('Could not find job instance "%s".', $code));
-        }
-
-        $job = $this->getJobRegistry()->get($jobInstance->getJobName());
-        $jobParamsFactory = $this->getJobParametersFactory();
-        $rawParameters = $jobInstance->getRawParameters();
-        if ($config = $input->getOption('config')) {
-            $rawParameters = array_merge($rawParameters, $this->decodeConfiguration($config));
-        }
-        $jobParameters = $jobParamsFactory->create($job, $rawParameters);
         $validator = $this->getValidator();
-
         // Override mail notifier recipient email
         if ($email = $input->getOption('email')) {
             $errors = $validator->validate($email, new Assert\Email());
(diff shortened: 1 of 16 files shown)


src/Akeneo/Bundle/BatchBundle/Resources/config/model/doctrine/JobExecution.orm.yml
```diff
@@ -3,6 +3,8 @@ Akeneo\Component\Batch\Model\JobExecution:
     table: akeneo_batch_job_execution
     # TODO: should be like other entities
     #changeTrackingPolicy: DEFERRED_EXPLICIT
+    entityListeners:
+        \Akeneo\Bundle\BatchBundle\EventListener\LoadJobParametersListener: ~
```
ahocquard: better in yaml (DI) or in entity ?

src/Akeneo/Bundle/BatchBundle/Resources/config/services.yml
```diff
@@ -36,6 +37,13 @@ services:
         tags:
             - { name: kernel.event_subscriber }
 
+    akeneo_batch.load_job_parameters_listener:
+        class: '%akeneo_batch.load_job_parameters_listener.class%'
+        arguments:
+            - '@akeneo_batch.job_parameters_factory'
+        tags:
+            - { name: doctrine.orm.entity_listener, lazy: true }
```
ahocquard: Lazy:

- Short story : pim can't be installed without lazy :(

- Long story : with a normal subscriber, I have to inject the container to avoid circular dependencies. That's dirty and a better thing exists now with Doctrine: entity listener. 
It seems that I don't have this circular dependency to the DBAL anymore.

But when I install the PIM, the loading of the channels fails because the UOW try to insert an existing channel in DB instead of updating it, in this listener: 

https://github.com/akeneo/pim-community-dev/blob/master/src/Pim/Bundle/EnrichBundle/EventListener/Storage/ChannelLocaleSubscriber.php#L111

What is very strange is that it's due to my listener loading the JobParametersFactory. A sub-dependency of this factory should make the UOW completely lost.

When I inject the factory, the channel entity does not exist in the UOW (resultin in `insert` +> PDO error because the foreign key already exists).
When I don't inject the factory, the entity exists in the UOW and the channel is detected as an already existing entity (resulting in `update` => PIM installed without error).

src/Akeneo/Bundle/BatchBundle/Resources/config/model/doctrine/JobExecution.orm.yml
anaelChardan: DI

src/Akeneo/Bundle/BatchBundle/Command/BatchCommand.php
```diff
@@ -256,67 +235,118 @@ protected function writeExceptions(OutputInterface $output, array $exceptions, $
...
+                    'Job instance "%s" running the job "%s" with parameters "%s" is invalid because of "%s"',
+                    $code,
+                    $job->getName(),
+                    print_r($jobParameters->all(), true),
+                    $this->getErrorMessages($errors)
+                )
+            );
+        }
+
+        $this->getDefaultEntityManager()->clear(get_class($jobInstance));
```
skeleton: ClassUtils::getClass

src/Akeneo/Bundle/BatchBundle/Job/DoctrineJobRepository.php
```diff
@@ -202,36 +202,16 @@ public function remove(array $jobsExecutions)
...
+        if (false === $connection->ping()) {
             $connection->close();
             $connection->connect();
         }
     }
 
     /**
-     * Pings the server, returns false if it's not available.
-     * There is a ping() method in Doctrine\DBAL\Connection in the doctrine/dbal package
-     * as of 2.5.0, but  we are currently on 2.4.x
```
skeleton: :+1:

src/Akeneo/Component/Batch/Job/JobParametersFactory.php
```diff
@@ -42,4 +43,16 @@ public function create(JobInterface $job, array $parameters = [])
...
+    /**
+     * Create a JobParameters from the raw parameters of a job execution.
+     *
+     * @param JobExecution $jobExecution the job execution to create the job parameters from
+     *
+     * @return JobParameters
+     */
+    public function createFromRawParameters(JobExecution $jobExecution) : JobParameters
+    {
+        return new $this->jobParametersClass($jobExecution->getRawParameters());
```
skeleton: getRawParameters can be null, it will work if it is?

ahocquard: No: I've changed and raw parameters has a default empty array in JobExecution constructor (and I added spec on it).

src/Akeneo/Bundle/BatchBundle/Resources/config/model/doctrine/JobExecution.orm.yml
jjanvier: same for me

src/Akeneo/Bundle/BatchBundle/EventListener/LoadJobParametersListener.php
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ * Load raw parameters of a job execution in order to inject it as a value object JobParameters,
+ * when loading the object with Doctrine.
+ *
+ * @author    Alexandre Hocquard <alexandre.hocquard@akeneo.com>
+ * @copyright 2017 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+class LoadJobParametersListener
```
jjanvier: normal we don't implement a SF or Doctrine (dunno which one you use) interface?

src/Akeneo/Component/Batch/Model/JobExecution.php
```diff
@@ -572,17 +575,30 @@ public static function formatDate(\DateTime $date = null, $format = \DateTime::A
 
     /**
      * @param JobParameters $jobParameters
+     *
+     * @return JobExecution
      */
-    public function setJobParameters(JobParameters $jobParameters)
+    public function setJobParameters(JobParameters $jobParameters): JobExecution
```
jjanvier: real question

will BatchBundle and component be compatible with > php 7.1 only?

because they provided as standalone libraries

src/Akeneo/Bundle/BatchBundle/tests/integration/Command/BatchCommandIntegration.php
```diff
@@ -0,0 +1,180 @@
...
+
+use Akeneo\Component\Batch\Job\BatchStatus;
+use Akeneo\Test\Integration\Configuration;
+use Akeneo\Test\Integration\TestCase;
+use Pim\Component\Catalog\Model\ProductInterface;
+use Symfony\Bundle\FrameworkBundle\Console\Application;
+use Symfony\Component\Console\Input\ArrayInput;
+use Symfony\Component\Console\Output\BufferedOutput;
+
+class BatchCommandIntegration extends TestCase
```
jjanvier: add phpunit as a dev dependency of the bundle please
(and maybe the component if you added some tests too)

src/Akeneo/Bundle/BatchBundle/EventListener/LoadJobParametersListener.php
ahocquard: I've just follow the documentation: http://doctrine-orm.readthedocs.io/en/latest/reference/events.html#entity-listeners

I thought the same, but I think it uses reflexion behind the hood, so it doesn't matter if you implement an interface or not. And it would force to do one interface per method on Doctrine side.

src/Akeneo/Component/Batch/Model/JobExecution.php
ahocquard: Good question: what do we want ? On the other hand, as you said for the PHP client, it can force to migrate.

ahocquard: Ok, actually it's already forced in PHP 7.1 by @damien-carcel:
https://github.com/akeneo/pim-community-dev/pull/6260/commits/1e939f56c356d89047a15db4634bc1bc47b1acdf

src/Akeneo/Bundle/BatchBundle/tests/integration/Command/BatchCommandIntegration.php
ahocquard: As seen together, I've moved these tests in the PIM directory tests (at the root of the project).

Just to remember:
The tests depend of the PIM and it's a big work to make it independent (need dedicated fixtures, dedicated way to load fixtures, dedicated TestsCase, etc).

As the tests can't be run without the PIM, they should not be put in the bundle. Tests in the bundle should work without the PIM.

src/Akeneo/Bundle/BatchBundle/Resources/config/model/doctrine/JobExecution.orm.yml
ahocquard: I've keep that in the entity because the delcaration works in CE but not in EE! I don't know why but I suppose it's something to do with the initialisation order in the container and Doctrine stuff (maybe a bug because it should work without doing anything).

PR #6569, opened 2017-08-16
PIM-6472: Update product models through imports

damien-carcel:
## Description

This PR ensures the update of the product models through imports.
`First commit` is a cherry pick for this PR #6511, `second commit` is only some alignments automatically performed by PHPStorm, so no real interest here.

The `third commit` adds the basic behat scenarios for update through import.

The PR adds two new business rules (with according tests):
- ensure that a parent cannot be added to a root product model (`fourth commit`),
- ensure that the variant axes values set on a product model cannot be changed once defined (`fifth commit`).

I also add a scenario to ensure that update works with `comparison` parameter enabled (included in `fifth` commit).

## Definition Of Done

| Q                                 | A
| --------------------------------- | ---
| Added Specs                       | OK
| Added Behats                      | OK
| Added integration tests           | OK
| Changelog updated                 | -
| Review and 2 GTM                  | Todo
| Micro Demo to the PO (Story only) | -
| Migration script                  | -
| Tech Doc                          | -

`Todo`: Pending / Work in progress
`OK`: Done / Validated
`-`: Not needed

PR #6598, opened 2017-08-22
PIM-6333: Do not update twice a product model in a batch

damien-carcel:
## Description

**Only the last commit is relevent, the others come from #6569, which needs to be merged first.**

For each family variant, they can only be one entity with family variant (product model or variant product) for a specific combination of variant axes values.

To ensure that, we check that such a entity does not already exists in database (already done in previous PR), but we also need to ensure there is not one in the current batch of import.

This PR introduces a new `Pim\Component\Catalog\Validator\UniqueAxesCombinationSet`, inspired from the `Pim\Component\Catalog\Validator\UniqueValuesSet` used on the products.

It stores all the MySQL ID (or if it still has none, an object hash) of the encountered entities, associated to their axes combination and family variant code (the same combination is authorized for different family variants). A violation will be raised if the same combination for the same family variant is found on a different product model of variant product.

Storing this in a dedicated object (the "unique axes combination set") allows to reset it between each batch.

## Definition Of Done

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Pim/Component/Catalog/Validator/UniqueAxesCombinationSet.php
+++ b/src/Pim/Component/Catalog/Validator/UniqueAxesCombinationSet.php
@@ -1,0 +1,81 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Pim\Component\Catalog\Validator;
+
+use Pim\Component\Catalog\Model\EntityWithFamilyVariantInterface;
+
+/**
+ * Contains the state of the unique axes combination for an entity with family variant.
+ * We use this state to deal with bulk update and validation.
+ *
+ * @author    Damien Carcel <damien.carcel@gmail.com>
+ * @copyright 2017 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+class UniqueAxesCombinationSet
+{
+    /** @var array */
+    private $uniqueAxesCombination;
+
+    /**
+     * Initializes the set.
+     */
+    public function __construct()
+    {
+        $this->uniqueAxesCombination = [];
+    }
+
+    /**
+     * Resets the set.
+     */
+    public function reset(): void
+    {
+        $this->uniqueAxesCombination = [];
+    }
+
+    /**
+     * Returns TRUE if axes combination has been added, FALSE if it already
+     * exists inside the set.
+     *
+     * @param EntityWithFamilyVariantInterface $entity
+     * @param string                           $axesCombination
+     *
+     * @return bool
+     */
+    public function addCombination(EntityWithFamilyVariantInterface $entity, string $axesCombination): bool
+    {
+        $identifier = $this->getEntityId($entity);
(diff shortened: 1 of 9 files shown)


src/Pim/Component/Catalog/Validator/UniqueAxesCombinationSet.php
```diff
@@ -0,0 +1,81 @@
...
+    /**
+     * Returns TRUE if axes combination has been added, FALSE if it already
+     * exists inside the set.
+     *
+     * @param EntityWithFamilyVariantInterface $entity
+     * @param string                           $axesCombination
+     *
+     * @return bool
+     */
+    public function addCombination(EntityWithFamilyVariantInterface $entity, string $axesCombination)
```
jjanvier: string?

damien-carcel: Yes, it is an implosion to create a array key.
For instance, if the axes values are color option "blue" and size option "XS", you'll have `"[blue,XS]"`.

That part was already done by Pipou in the validator.

grena: please add return type

src/Pim/Component/Catalog/Validator/UniqueAxesCombinationSet.php
```diff
@@ -0,0 +1,81 @@
...
+     */
+    public function __construct()
+    {
+        $this->uniqueAxesCombination = [];
+    }
+
+    /**
+     * Resets the set.
+     */
+    public function reset()
```
grena: please add return type

src/Pim/Component/Catalog/Validator/UniqueAxesCombinationSet.php
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    /**
+     * spl_object_hash for new product and id when product exists
+     *
+     * @param EntityWithFamilyVariantInterface $entity
+     *
+     * @return string
+     */
+    private function getEntityId(EntityWithFamilyVariantInterface $entity)
```
grena: return type

src/Pim/Component/Catalog/Validator/UniqueAxesCombinationSet.php
```diff
@@ -0,0 +1,81 @@
...
+     *
+     * @return bool
+     */
+    public function addCombination(EntityWithFamilyVariantInterface $entity, string $axesCombination)
+    {
+        $identifier = $this->getEntityId($entity);
+        $familyVariantCode = $entity->getFamilyVariant()->getCode();
+
+        if (isset($this->uniqueAxesCombination[$familyVariantCode][$axesCombination])) {
+            $storedIdentifier = $this->uniqueAxesCombination[$familyVariantCode][$axesCombination];
```
SamirBoulil: nothing is stored? it is maybe cached, but not stored

PR #6599, opened 2017-08-22
PIM-6353: filter by default products and product models that don't have a parent

jjanvier:
*Easier to read commit by commit*

The goal of this is PR is to be able, with the product and product model query builder, to display by default only entities without a parent. Soon, in the product datagrid, we'll see only the root product models and the non variant products.

For that we:
- indexed the parent code of the product models and variant products
- defined a new filter on this parent field

This PR also fixes a few glitches:
- one about numeric identifiers and our Elasticsearch cursors (basically, a product with identifier 12345 was never returned)
- simplify the "product_type" in the product and product model index

| Added integration tests           | Y

### DIFF
--- a/src/Pim/Bundle/CatalogBundle/Elasticsearch/Filter/Field/ParentFilter.php
+++ b/src/Pim/Bundle/CatalogBundle/Elasticsearch/Filter/Field/ParentFilter.php
@@ -1,0 +1,55 @@
+<?php
+
+namespace Pim\Bundle\CatalogBundle\Elasticsearch\Filter\Field;
+
+use Pim\Component\Catalog\Exception\InvalidOperatorException;
+use Pim\Component\Catalog\Query\Filter\FieldFilterInterface;
+use Pim\Component\Catalog\Query\Filter\Operators;
+
+/**
+ * Parent filter for an Elasticsearch query
+ *
+ * @author    Julien Janvier <j.janvier@gmail.com>
+ * @copyright 2017  Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php Open Software License (OSL 3.0)
+ */
+class ParentFilter extends AbstractFieldFilter implements FieldFilterInterface
+{
+    /**
+     * @param array $supportedFields
+     * @param array $supportedOperators
+     */
+    public function __construct(
+        array $supportedFields = [],
+        array $supportedOperators = []
+    ) {
+        $this->supportedFields = $supportedFields;
+        $this->supportedOperators = $supportedOperators;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function addFieldFilter($field, $operator, $value, $locale = null, $channel = null, $options = [])
+    {
+        if (null === $this->searchQueryBuilder) {
+            throw new \LogicException('The search query builder is not initialized in the filter.');
+        }
+
+        switch ($operator) {
+            case Operators::IS_EMPTY:
+                $clause = [
+                    'exists' => [
+                        'field' => $field,
+                    ],
+                ];
+
+                $this->searchQueryBuilder->addMustNot($clause);
+                break;
+            default:
+                throw InvalidOperatorException::notSupported($operator, static::class);
+        }
+
+        return $this;
+    }
+}
--- a/src/Pim/Bundle/CatalogBundle/tests/integration/Elasticsearch/IndexConfiguration/SearchProductsAndModelsIntegration.php
+++ b/src/Pim/Bundle/CatalogBundle/tests/integration/Elasticsearch/IndexConfiguration/SearchProductsAndModelsIntegration.php
@@ -2,6 +2,9 @@
 
 namespace Pim\Bundle\CatalogBundle\tests\integration\Elasticsearch\IndexConfiguration;
 
+use Pim\Component\Catalog\Model\ProductInterface;
+use Pim\Component\Catalog\Model\ProductModelInterface;
+
 /**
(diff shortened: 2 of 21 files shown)


src/Pim/Bundle/CatalogBundle/tests/integration/Elasticsearch/IndexConfiguration/SearchProductsAndModelsIntegration.php
```diff
@@ -33,9 +36,9 @@ public function testDefaultDisplay()
         $query = [
             'query' => [
                 'bool' => [
-                   'filter' => [
-                       'terms' => [
-                           'product_type' => ['PimCatalogRootProductModel', 'PimCatalogProduct'],
+                   'must_not' => [
```
SamirBoulil: 👍

src/Pim/Component/Catalog/Normalizer/Indexing/ProductAndProductModel/ProductModelNormalizer.php
```diff
@@ -54,26 +54,4 @@ public function supportsNormalization($data, $format = null)
     {
         return $data instanceof ProductModelInterface && self::INDEXING_FORMAT_PRODUCT_AND_MODEL_INDEX === $format;
     }
-
-    /**
```
SamirBoulil: 👍

src/Pim/Bundle/CatalogBundle/Elasticsearch/Filter/Field/ParentFilter.php
```diff
@@ -0,0 +1,52 @@
...
+     */
+    public function __construct(
+        array $supportedFields = [],
+        array $supportedOperators = []
+    ) {
+        $this->supportedFields = $supportedFields;
+        $this->supportedOperators = $supportedOperators;
+    }
+
+    public function addFieldFilter($field, $operator, $value, $locale = null, $channel = null, $options = [])
```
grena: missing phpdoc & return type (php7)

src/Pim/Component/Catalog/Normalizer/Indexing/ProductAndProductModel/ProductModelNormalizer.php
```diff
@@ -41,7 +41,7 @@ public function normalize($productModel, $format = null, array $context = [])
     {
         $data = $this->propertiesNormalizer->normalize($productModel, $format, $context);
 
-        $data[self::FIELD_PRODUCT_TYPE] = $this->getVariationLevelCode($productModel);
+        $data[self::FIELD_PRODUCT_TYPE] = ProductModelInterface::class;
```
grena: I don't understand :(

src/Pim/Component/Catalog/Normalizer/Indexing/ProductAndProductModel/ProductPropertiesNormalizer.php
```diff
@@ -79,6 +80,12 @@ public function normalize($product, $format = null, array $context = [])
         }
         $data[self::FIELD_FAMILY_VARIANT] = $familyVariantCode;
 
+        $parentCode = null;
+        if ($product instanceof VariantProductInterface && null !== $product->getParent()) {
```
grena: Is it possible that a VariantProduct doesn't have parent?!

src/Pim/Component/Catalog/Normalizer/Indexing/ProductAndProductModel/ProductModelNormalizer.php
SamirBoulil: This field is used to differentiate Products from models when they are mixed in the same index.
Before this PR we would use some arbitrary code "PimCatalogProduct" or "PimCatalogRootProductModel" or even "PimCatalogSubProductModel".

This was ineffective as we were mixing two information here: 
- The level in which the documents belongs to
- The type of document it is (products or models).

This PR fixes this issue by having two document types which codes are the FQCN to ProductModelInterface and ProductInterface

We in fact did not need a "level" attribute, for now we only needed to know if a document has a parent. 
The filter "Show me the documents which has no parent" is exactly what we want to display in the datagrid default view.

src/Pim/Component/Catalog/Normalizer/Indexing/ProductAndProductModel/ProductPropertiesNormalizer.php
SamirBoulil: I guess so ? if you remove a product model from a tree ? (even if it's not possible, it is worth a check in my opinion)

src/Pim/Bundle/CatalogBundle/Elasticsearch/Filter/Field/ParentFilter.php
jjanvier: so but it comes from our `FilterInterface`, `FieldFilterInterface` and `AttributeFilterInterface`

and I won't typehint all the PQB filters in this PR :)

jjanvier: php doc fixed

src/Pim/Component/Catalog/Normalizer/Indexing/ProductAndProductModel/ProductPropertiesNormalizer.php
jjanvier: a valid variant product shouldn't
but we should always be defensive in our code

PR #6602, opened 2017-08-22
[PIM-6551] Update association page style

pierallard:
selection_093

Some notes:
- I added code to display images on the associated products panel (as mockups)
- I created a new grid param, to display filters as columns. In fact, it don't display filters as column at all, but it will after @tamarasaurus ' work (product/published product rework)
- I updated some SQL to be able to search on labels.

### DIFF
--- a/src/Pim/Bundle/DataGridBundle/Resources/public/js/datafilter-builder.js
+++ b/src/Pim/Bundle/DataGridBundle/Resources/public/js/datafilter-builder.js
@@ -31,12 +31,10 @@ function($, _, tools,  mediator, FiltersManager) {
             },
 
             build: function () {
-                var displayManageFilters = _.result(this.metadata.options, 'manageFilters', true);
-                var filtersAsColumn = _.result(this.metadata.options, 'filtersAsColumn', false);
                 var options = methods.combineOptions.call(this);
                 options.collection = this.collection;
-                options.displayManageFilters = displayManageFilters;
-                options.filtersAsColumn = filtersAsColumn;
+                options.displayManageFilters = _.result(this.metadata.options, 'manageFilters', true);
+                options.filtersAsColumn = _.result(this.metadata.options, 'filtersAsColumn', false);
                 var filtersList = new FiltersManager(options);
                 this.$el.prepend(filtersList.render().$el);
                 mediator.trigger('datagrid_filters:rendered', this.collection);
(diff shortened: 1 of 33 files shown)


src/Pim/Bundle/DataGridBundle/Resources/public/js/datafilter-builder.js
```diff
@@ -32,9 +32,11 @@ function($, _, tools,  mediator, FiltersManager) {
 
             build: function () {
                 var displayManageFilters = _.result(this.metadata.options, 'manageFilters', true);
+                var filtersAsColumn = _.result(this.metadata.options, 'filtersAsColumn', false);
                 var options = methods.combineOptions.call(this);
                 options.collection = this.collection;
                 options.displayManageFilters = displayManageFilters;
+                options.filtersAsColumn = filtersAsColumn;
```
phaseinducer: Why not assigning it directly ?

src/Pim/Bundle/DataGridBundle/Resources/public/js/datafilter/filters-manager.js
```diff
@@ -39,13 +39,22 @@ function($, _, Backbone, mediator, MultiselectDecorator) {
...
+
+        /**
          * Container classes
          *
          * @property
          */
         className: function () {
             var name = 'AknFilterBox filter-box oro-clearfix-width';
-            if (true !== this.displayManageFilters()) {
+            if (true !== this.filtersAsColumn()) {
```
phaseinducer: I would drop the `true !==`

src/Pim/Bundle/EnrichBundle/Resources/public/js/product/form/associations.js
```diff
@@ -85,6 +85,7 @@ define(
                         getInitialParams: function (associationType) {
                             var params = {};
                             params[this.paramName] = this.getParamValue(associationType);
+                            params.dataLocale = UserContext.get('catalogLocale');
```
phaseinducer: Why is it needed ?

pierallard: SO. This is needed because we currently had a bug on label search.
Before this PR, you don't pass the dataLocale parameter, so it was null. When you searched in the label, it searched with this https://github.com/akeneo/pim-community-dev/pull/6602/files#diff-6158dd114f4d591932c384945247507eR110
(i.e. if you don't have any translation, just fallback with code). As there was never any translation, it always fallback to the code !
So this line fix it.

src/Pim/Bundle/DataGridBundle/Resources/public/js/datafilter/filters-manager.js
pierallard: Personnaly, I prefer be more strict than necessary. Here, you prevent some magic JS typehint

PR #6604, opened 2017-08-23
PIM-6333: Fix "UniqueAxesCombinationSet"

damien-carcel:
## Description

The newly introduced `UniqueAxesCombinationSet` was wrong as it stored axes combinations only according to family variants. However, there can be several root product models for the same family variant, so several time the same axes combination for a family variant.

This PR updates the `UniqueAxesCombinationSet` so axes combination are now stored according to family variant and parent.

## Definition Of Done

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Pim/Component/Catalog/Validator/UniqueAxesCombinationSet.php
+++ b/src/Pim/Component/Catalog/Validator/UniqueAxesCombinationSet.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace Pim\Component\Catalog\Validator;
 
 use Pim\Component\Catalog\Model\EntityWithFamilyVariantInterface;
+use Pim\Component\Catalog\Model\VariantProductInterface;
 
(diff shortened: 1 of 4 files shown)


src/Pim/Component/Catalog/Validator/UniqueAxesCombinationSet.php
```diff
@@ -46,11 +47,16 @@ public function reset(): void
      */
     public function addCombination(EntityWithFamilyVariantInterface $entity, string $axesCombination): bool
     {
-        $identifier = $this->getEntityId($entity);
+        if (null === $entity->getParent()) {
```
jjanvier: you shouldn't be here imo if there is no parent

damien-carcel: Fixed

PR #6605, opened 2017-08-23
API-91: extract product's fields from productFilter

skeleton:

Extract product's fields from ProductFilter to be able to use this new Filter on EE

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Pim/Component/Catalog/Comparator/Filter/ProductFieldFilter.php
+++ b/src/Pim/Component/Catalog/Comparator/Filter/ProductFieldFilter.php
@@ -1,0 +1,70 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Pim\Component\Catalog\Comparator\Filter;
+
+use Akeneo\Component\StorageUtils\Exception\UnknownPropertyException;
+use Pim\Component\Catalog\Comparator\ComparatorRegistry;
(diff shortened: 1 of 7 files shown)


src/Pim/Component/Catalog/Comparator/Filter/ProductFieldFilter.php
```diff
@@ -0,0 +1,68 @@
+<?php
+
```
ahocquard: strict (but useless....)

src/Pim/Component/Catalog/Comparator/Filter/ProductFieldFilter.php
```diff
@@ -0,0 +1,68 @@
...
+     * {@inheritdoc}
+     */
+    public function filter(ProductInterface $product, array $newFields)
+    {
+        $originalProduct = $this->normalizer->normalize($product, 'standard');
+        $result = [];
+
+        foreach ($newFields as $field => $value) {
+            if (!in_array($field, $this->productFields)) {
+                throw new \LogicException(sprintf('Cannot filter value of field "%s"', $field));
```
ahocquard: not a better exception ? Also, what happens if you pass a field that does not exist? This exception is thrown? Because we have already this validation in the updaters as well (with an error 422).
But here, it will be a 500 no ?

PR #6613, opened 2017-08-23
API-348: reactive "offset" pagination type on the API 

skeleton:

At the beginning of the 1.8 when elasticsearch was used to get products, we removed the pagination_type `offset` (or `page`), because elasticsearch forbidden to process more than 10000 items, it sends this exception: 
`Result window is too large, from + size must be less than or equal to: [10000] but was [10050]. See the scroll api for a more efficient way to request large data sets. This limit can be set by changing the [index.max_result_window] index level setting.`

To be able to continue to browse the pages (to be able to access on the next page and the **previous** page), we twisted (a lot) the `search_after` pagination type. 
The code was horrible to code and even more to understand.

After a discussion with PO, it's acceptable to use the [from/size pagination type](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-from-size.html) of elasticsearch and to have an exception. It's not normal to use this type of pagination and want to access to the 10000th products. If we want to access to these products, we have to use the search_after type

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Pim/Bundle/ApiBundle/Controller/ProductController.php
+++ b/src/Pim/Bundle/ApiBundle/Controller/ProductController.php
@@ -10,6 +10,7 @@ use Akeneo\Component\StorageUtils\Remover\RemoverInterface;
 use Akeneo\Component\StorageUtils\Repository\IdentifiableObjectRepositoryInterface;
 use Akeneo\Component\StorageUtils\Saver\SaverInterface;
 use Akeneo\Component\StorageUtils\Updater\ObjectUpdaterInterface;
+use Elasticsearch\Common\Exceptions\ServerErrorResponseException;
 use Pim\Bundle\ApiBundle\Checker\QueryParametersCheckerInterface;
 use Pim\Bundle\ApiBundle\Documentation;
 use Pim\Bundle\ApiBundle\Stream\StreamResourceResponse;
@@ -629,9 +630,8 @@ class ProductController
 
         $queryParameters = array_merge(['page' => 1, 'with_count' => 'false'], $queryParameters);
         $pqb->addSorter('id', Directions::ASCENDING);
-        $products = $pqb->execute();
 
-        $count = 'true' === $queryParameters['with_count'] ? $products->count() : null;
+        $products = $pqb->execute();
 
         $paginationParameters = [
             'query_parameters'    => $queryParameters,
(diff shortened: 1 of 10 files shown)


src/Pim/Bundle/ApiBundle/Controller/ProductController.php
```diff
@@ -616,61 +598,131 @@ protected function validateCodeConsistency(string $code, array $data): void
...
+        array $queryParameters,
+        array $normalizerOptions
+    ): array {
+        $from = isset($queryParameters['page']) ? ($queryParameters['page'] - 1) * $queryParameters['limit'] : 0;
+        $pqb = $this->fromSizePqbFactory->create(['limit' => (int) $queryParameters['limit'], 'from' => (int) $from]);
+
+        try {
+            $this->setPQBFilters($pqb, $request, $channel);
+        } catch (
+            UnsupportedFilterException
```
anaelChardan: :+1:

src/Pim/Bundle/ApiBundle/tests/integration/Controller/Product/ErrorListProductIntegration.php
```diff
@@ -242,6 +242,37 @@ public function testSearchProductAttributeDoesNotExist()
...
+
+        $products = [];
+        for ($i = 0; $i<=10001; $i++) {
+            $products[] = $this->get('pim_catalog.builder.product')->createProduct('sku' . $i);
+        }
+
+        $this->get('pim_catalog.saver.product')->saveAll($products);
+        $this->get('akeneo_elasticsearch.client')->refreshIndex();
+
+        $client->request('GET', 'api/rest/v1/products?page=101&limit=100');
```
anaelChardan: I don't understand, this means that even with a limit of 100 we can't access to the 101th page ?

skeleton: yep. but as explain on the description of the PR, it's not normal to access to the 101th page with this  type of pagination. 
This type is designed to be use by a front (because you want to browse into your pages). 
If a user has to access to the 101th, it can mean two things:
either he has problem with his filters, It does not well configured them or he wants to exports products and `search_after` pagination type is here for that.

src/Pim/Bundle/ApiBundle/Controller/ProductController.php
ahocquard: Anael like your way to you copy/past @skeleton :trollface:

src/Pim/Bundle/ApiBundle/Controller/ProductController.php
```diff
@@ -616,61 +598,131 @@ protected function validateCodeConsistency(string $code, array $data): void
...
+            'item_identifier_key' => 'identifier',
+        ];
+
+        try {
+            $paginatedProducts = $this->offsetPaginator->paginate(
+                $this->normalizer->normalize($products, 'external_api', $normalizerOptions),
+                $paginationParameters,
+                $count
+            );
+        } catch (ServerErrorResponseException $e) {
```
ahocquard: no way to check the code ? Because it's a pretty generic exception I suppose

skeleton:
@ahocquard about tests, I don't think so. We are iso with 1.7 now

src/Pim/Bundle/ApiBundle/Controller/ProductController.php
skeleton: code is always 500 for all errors related to the server.
The only thing I can check is the type in the message. Code is quite ugly, but it's a little bit more precise

PR #6621, opened 2017-08-24
PIM-6335: variant product, business rules about famliy

aRn0D:

* First commit: When we import a variant product without a family it has its parent family
* Second commit: variant product, business rules about family


features/import/product_variant/import_variant_products.feature
```diff
@@ -26,4 +26,16 @@ Feature: Execute a job
       | ean    | EAN           |
       | sku    | SKU-001       |
       | weight | 100.0000 GRAM |
-      | size   | [m]           |
\ No newline at end of file
+      | size   | [m]           |
+
+  Scenario: When we import a variant product without a family it has its parent family
```
SamirBoulil: ...product without a family, then his parent's family is assigned to it.

features/import/product_variant/import_variant_products.feature
```diff
@@ -26,4 +26,16 @@ Feature: Execute a job
...
+    Given the following root product model "code-001" with the variant family clothing_color_size
+    And the following sub product model "code-002" with "code-001" as parent
+    And the following CSV file to import:
+      """
+      parent;categories;ean;sku;weight;weight-unit;size
+      code-001;master_men;EAN;SKU-001;100;GRAM;m
+      """
+    And the following job "csv_default_product_import" configuration:
```
SamirBoulil: The test is not finished.

src/Pim/Component/Catalog/Updater/Setter/ParentSetter.php
```diff
@@ -53,7 +54,12 @@ public function setFieldData($product, $field, $data, array $options = []): void
             );
         }
 
+        $familyVariant = $parent->getFamilyVariant();
+
         $product->setParent($parent);
-        $product->setFamilyVariant($parent->getFamilyVariant());
+        $product->setFamilyVariant($familyVariant);
```
jjanvier: it seems really weird to set the family in something called ParentSetter?

isn't there a FamilySetter for that?

features/import/product_variant/import_business_rules.feature
```diff
@@ -0,0 +1,25 @@
+@javascript
+Feature: Import variant products
```
jjanvier: could we change the name of the folder to `variant_product`?

features/import/product_variant/import_variant_products.feature
```diff
@@ -28,7 +28,7 @@ Feature: Import variant product
       | weight | 100.0000 GRAM |
       | size   | [m]           |
 
-  Scenario: When we import a variant product without a family it has its parent family
+  Scenario: When we import a variant product without a family, then his parent's family is assigned to it.
```
jjanvier: ~his~ _its_ parent's family

src/Pim/Component/Catalog/Validator/Constraints/SameFamilyThanParentValidator.php
```diff
@@ -0,0 +1,41 @@
...
+    {
+        if (!$variantProduct instanceof VariantProductInterface) {
+            throw new UnexpectedTypeException($constraint, VariantProductInterface::class);
+        }
+
+        if (!$constraint instanceof SameFamilyThanParent) {
+            throw new UnexpectedTypeException($constraint, SameFamilyThanParent::class);
+        }
+
+        $parentFamilyVariant = $variantProduct->getParent()->getFamilyVariant()->getFamily();
```
jjanvier: shouldn't we check parent is not null
and family variant is not null
?
and maybe throw exceptions if we consider it should not happen

aRn0D: yes for a parent but for the rest is useless because a product model must have a family variant and a family variant must have a family.

src/Pim/Component/Catalog/Updater/Setter/ParentSetter.php
aRn0D: The family variant is set there because there is no `variant_attribute` column in the CSV file. The family variant of a variant product depends on its parent. For instance, I want to set the family variant, I need to know the parent it is not accessible from a setter that's why I put it there.

For the creation, the family is set in the `ProductBuilder`

features/import/product_variant/import_variant_products.feature
aRn0D: Fixed :)

src/Pim/Component/Catalog/Updater/Setter/ParentSetter.php
```diff
@@ -7,6 +7,7 @@
 use Akeneo\Component\StorageUtils\Exception\InvalidPropertyException;
```
jjanvier: according to the other classes we have, it could be called ParentFieldSetter

src/Pim/Component/Catalog/Updater/Setter/ParentSetter.php
```diff
@@ -53,7 +54,12 @@ public function setFieldData($product, $field, $data, array $options = []): void
             );
         }
 
+        $familyVariant = $parent->getFamilyVariant();
+
         $product->setParent($parent);
-        $product->setFamilyVariant($parent->getFamilyVariant());
+        $product->setFamilyVariant($familyVariant);
+        if (null === $product->getFamily()) {
+            $product->setFamily($familyVariant->getFamily());
```
jjanvier: I don't know if we handle all use cases here

imagine the product was imported as a regular product
then, it's imported as a variant product (with a parent): maybe here we have to handle the "deletion" of values that should not be here

jjanvier: please add this remark in the JIRA ticket linked to this story
Delph should handle this case

damien-carcel: Done

src/Pim/Component/Catalog/Updater/Setter/ParentSetter.php
damien-carcel: Done

PR #6622, opened 2017-08-25
Cloud needs: adding smtp port

mr-prud:

In some case, default smtp parameters must be overloaded like encryption capability and no-standard smtp port.

Default values choose from swiftmailer doc
http://symfony.com/doc/current/reference/configuration/swiftmailer.html#encryption
http://symfony.com/doc/current/reference/configuration/swiftmailer.html#port


jjanvier:
maybe you should do it in standard editions also

PR #6623, opened 2017-08-25
PIM-6338: extract "fill missing attributes" logic

grena:

This PR does 2 things:
- Create a new interface `EntityWithFamilyInterface`. It's now the common interface between **products**, **variant products** and **product models**.
- Extract the logic of filling missing attributes on a product (currently in the product builder) in a dedicated service.

### DIFF
--- a/src/Pim/Component/Catalog/Model/EntityWithFamilyInterface.php
+++ b/src/Pim/Component/Catalog/Model/EntityWithFamilyInterface.php
@@ -3,6 +3,13 @@
 namespace Pim\Component\Catalog\Model;
 
 /**
+ * Entity with a family.
+ *
+ * Having a family means having a structure based on family attributes.
+ * For example, product attributes are all coming from its family.
+ * Product model attributes come from an attribute set, which is a sub-part of family attributes.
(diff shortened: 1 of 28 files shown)


src/Pim/Component/Catalog/Model/EntityWithFamilyInterface.php
```diff
@@ -0,0 +1,16 @@
+<?php
+
+namespace Pim\Component\Catalog\Model;
+
+/**
```
jjanvier: we should explain a little bit things here

maybe take some examples

src/Pim/Component/Catalog/ValuesFiller/EntityWithFamilyValuesFillerInterface.php
```diff
@@ -0,0 +1,23 @@
+<?php
+
+namespace Pim\Component\Catalog\ValuesFiller;
+
+use Pim\Component\Catalog\Model\EntityWithFamilyInterface;
+
+/**
```
jjanvier: we should explain what it does here

PR #6629, opened 2017-08-26
Api 363 : Add an option in the job parameters to authenticate (or not) the user that launched the job

ahocquard:
I've updated my accordingly to the discussion with @BitOne, and here a sum-up of the architecture:

We will have a queue to execute a job in CE/EE/Saas (no mode queue/no queue):
- launched from the UI
- launched from the CLI, with a new command (not implemented in this PR) 

The command "akeneo:batch:job" will not be modified (at least, for now):
- it executes a job synchronously (not pushed into the queue)
- or it executes a job execution synchronously

In the last case, "username" and "config" option are forbidden, because a job execution should be immutable.

- [x] Add an optional "username" in the command "akeneo:batch:job" (to be like the SimpleJobLauncher, except that the SimpleJobLauncher force the use of a user)
- [x] Add an option in the job parameters to authenticate (or not) the user that launched the job.
- [x] Rename key "notification_user" into "user_to_notify" ( @jjanvier comment on already merged PR)

What's left ?
- [ ] Add a table for the queue
- [ ] Push Job Execution messages into the queue with a "QueueJobLauncher"
- [ ] Create a daemon to consume job execution messages into the queue, execute them, and monitor them
- [ ] Create a command to create a job execution and push it into database => should be used by the cron in the future

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Pim/Bundle/ConnectorBundle/Command/AuthenticatedBatchCommand.php
+++ b/src/Pim/Bundle/ConnectorBundle/Command/AuthenticatedBatchCommand.php
@@ -2,7 +2,7 @@
 
 namespace Pim\Bundle\ConnectorBundle\Command;
 
-use Akeneo\Bundle\BatchBundle\Command\BatchCommand as BaseBatchCommand;
+use Akeneo\Bundle\BatchBundle\Command\BatchCommand;
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
@@ -15,7 +15,7 @@ use Symfony\Component\Console\Output\OutputInterface;
  * @copyright 2017 Akeneo SAS (http://www.akeneo.com)
  * @license   http://opensource.org/licenses/MIT MIT
  */
-class BatchCommand extends BaseBatchCommand
+class AuthenticatedBatchCommand extends BatchCommand
 {
     /**
      * {@inheritdoc}
--- a/src/Pim/Bundle/ConnectorBundle/tests/integration/Command/AuthenticatedBatchCommandIntegration.php
+++ b/src/Pim/Bundle/ConnectorBundle/tests/integration/Command/AuthenticatedBatchCommandIntegration.php
@@ -12,7 +12,7 @@ use Symfony\Bundle\FrameworkBundle\Console\Application;
 use Symfony\Component\Console\Input\ArrayInput;
 use Symfony\Component\Console\Output\BufferedOutput;
 
-class BatchCommandIntegration extends TestCase
+class AuthenticatedBatchCommandIntegration extends TestCase
 {
     const EXPORT_DIRECTORY = 'pim-integration-tests-export';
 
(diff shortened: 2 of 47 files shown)


src/Pim/Bundle/ConnectorBundle/Command/AuthenticatedBatchCommand.php
```diff
@@ -0,0 +1,68 @@
...
+use Symfony\Component\Console\Output\OutputInterface;
+
+/**
+ * Batch command to launch an job with an authenticated user.
+ *
+ * @author    Olivier Soulet <olivier.soulet@akeneo.com>
+ * @copyright 2017 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/MIT MIT
+ */
+class AuthenticatedBatchCommand extends BatchCommand
```
ahocquard: Github does not detect it as a rename but it is.

src/Pim/Bundle/ConnectorBundle/tests/integration/Command/AuthenticatedBatchCommandIntegration.php
```diff
@@ -0,0 +1,157 @@
...
+
+        $output = $this->launchJob(['--config' => ['filePath' => $filePath]]);
+        $this->assertEquals('Export csv_product_export has been successfully executed.' . PHP_EOL, $output->fetch());
+        $this->assertTrue(file_exists($filePath));
+    }
+
+    public function testLaunchJobWithNoLog()
+    {
+        $output = $this->launchJob(['--no-log' => true]);
+        $this->assertEquals('Export csv_product_export has been successfully executed.' . PHP_EOL, $output->fetch());
```
momoss: Maybe you can check that there is no log file?

ahocquard: Nop, it is the log of the standard output.

src/Pim/Bundle/ConnectorBundle/Command/AuthenticatedBatchCommand.php
```diff
@@ -0,0 +1,68 @@
...
+{
+    /**
+     * {@inheritdoc}
+     */
+    protected function configure()
+    {
+        $this
+            ->setName('pim:batch:job')
```
jjanvier: sure about the name?

do we still have a command to launch jobs when we don't care about launching it with a user?

ahocquard:
@jjanvier launching a job with a username is important because it's necessary to apply the permissions.
And it's very important that a job execution has the necessary information to apply permissions or not. And, as it is a PIM business rules, it's necessary to apply it in job parameters.

src/Pim/Bundle/ConnectorBundle/Command/AuthenticatedBatchCommand.php
skeleton: BatchCommand I guess

src/Pim/Bundle/ConnectorBundle/Command/AuthenticatedBatchCommand.php
skeleton: you can still run `akeneo:batch:job` without user, it will work.
So job for completeness for instance will work as before

jjanvier: ahhhhhhh
we are in the connector bundle 

ok... I didn't get it at all sorry

ahocquard:
@jjanvier 

Seen IRL, but to sum-up:

If you take a look at the `SimpleJobLauncher`, it always set a user.
At the opposite, before this PR, the command `akeneo:batch:job` never set a username.
It means that depending of your entrypoint (CLI or SimpleJobLauncher), you don't have the same functionality. 

For example, before, it was not possible to notify a username from the CLI (because the notification was based on the username-.

- Second thing:
`akeneo:batch:job` is an entry point. This entry point should be able to launch any kind job.
It's this command that will be called to launch a specific job execution when you will consume it in the queue. ( `akeneo:batch:job job_instance_code execution_code`)

`pim:batch:job` is another entry point. This entry point should not contains any business rule relative to the job execution (that's why I refactor it). 
It would force us to specify the name of the command to apply when launching a job execution in the queue. That's just the same thing that specifying a parameter in the job execution, but in a dirty way.
And it generates a lot of dirty things (one of them is that it's hard to make composition through commands).

The would say that the main rule is:
**A job execution should have all the knowledge to apply its own business rules.**

PR #6631, opened 2017-08-28
PIM-6335: Skip imported variant product with invalid parent

damien-carcel:
## Description

This PR implements two business rules regarding the import of the variant products:
- if the parent product model doesn't exist, the variant product is skipped with a warning message,
- if the parent product model is a root product model and the family variant has 2 levels, the variant product is skipped with a warning message.

The first rule was in fact already implemented and only needed a behat to test the import.

Second rule comes with everything: behat scenario, new validator with its unit tests.

## Definition Of Done

`OK`: Done / Validated
`-`: Not needed


features/import/product_variant/import_variant_products.feature
```diff
@@ -9,9 +9,25 @@ Feature: Execute a job
     And I am logged in as "Julia"
 
   Scenario: Import variant product
-    Given the following root product model "code-001" with the variant family clothing_color_size
-    And the following sub product model "code-002" with "code-001" as parent
-    And the following CSV file to import:
+    Given the following CSV file to import:
```
damien-carcel: Here I am using our catalog instead of product models created without validation.

features/import/product_variant/import_variant_products.feature
```diff
@@ -21,9 +37,31 @@ Feature: Execute a job
...
+      apollon;clothing;master_men;EAN;apollon_blue_medium;100;GRAM;m
+      """
+    And the following job "csv_default_product_import" configuration:
+      | filePath | %file to import% |
+    When I am on the "csv_default_product_import" import job page
+    And I launch the import job
+    And I wait for the "csv_default_product_import" job to finish
+    Then I should see the text "Status: Completed"
+    And I should see the text "skipped 1"
+    And I should see the text "Parent of the variant product \"apollon_blue_medium\" must be at variant level 1, product model \"apollon\" is at level 0"
```
jjanvier: not sure the message is really understandable
to check with Delphine

src/Pim/Bundle/CatalogBundle/Resources/config/validation/variantproduct.yml
```diff
@@ -0,0 +1,18 @@
+Pim\Component\Catalog\Model\VariantProduct:
```
jjanvier: isn't it already in Arnaud's PR?

features/import/product_variant/import_variant_products.feature
SamirBoulil: I'm wondering if those kind of validation tests should be integration tests instead of behat. Maybe we only need the nominal case in the behat, and another case where the import fails (for whatever reasons). And test-out all cases with integration tests.

src/Pim/Component/Catalog/Validator/Constraints/ParentPositionInTheVariantTreeValidator.php
```diff
@@ -0,0 +1,53 @@
...
+        }
+
+        $familyVariant = $variantProduct->getFamilyVariant();
+        $parent = $variantProduct->getParent();
+
+        if (null === $familyVariant || null === $parent) {
+            return;
+        }
+
+        $expectedLevel = $familyVariant->getNumberOfLevel() - 1;
```
jjanvier: imo you should check something like

```
if (!$model->getProductModels()->isEmpty()) {
 // exception
}

// not sure we should handle the following use case as normally, our models tree is supposed to be good
if (!$model->isRootProductModel() && null === $model->getParent()) {
// exception
}
```

features/import/product_variant/import_variant_products.feature
jjanvier: maybe you should simply tell something like

"Parent of the variant product can not have children product models."
"Variant product should be linked to a child product model."

SamirBoulil: Not sure is the we got the validation right :/ In my opinion it does not work not only because of the level mismatch which is a side effect of the root product model can only have a product model has a child (=> type mismatch).
It may be the same but I wonder if it's understandable by the user.  See with Delph if it's ok ;)

src/Pim/Component/Catalog/Model/ProductModelInterface.php
```diff
@@ -51,7 +51,7 @@ public function getProducts(): Collection;
     /**
      * Adds an product to the product model.
      *
-     * @param ProductInterface $product
+     * @param VariantProductInterface $product
```
SamirBoulil: Shouldn't we switch the method name for `addVariantProduct`?

src/Pim/Component/Catalog/Validator/Constraints/ParentPositionInTheVariantTreeValidator.php
```diff
@@ -0,0 +1,53 @@
...
+
+namespace Pim\Component\Catalog\Validator\Constraints;
+
+use Pim\Component\Catalog\Model\VariantProductInterface;
+use Symfony\Component\Validator\Constraint;
+use Symfony\Component\Validator\ConstraintValidator;
+use Symfony\Component\Validator\Exception\UnexpectedTypeException;
+
+/**
+ * @author    Damien Carcel (damien.carcel@akeneo.com)
```
SamirBoulil: A small description is always welcome

src/Pim/Component/Catalog/Validator/Constraints/ParentPositionInTheVariantTreeValidator.php
```diff
@@ -0,0 +1,53 @@
...
+ * @copyright 2017 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php Open Software License (OSL 3.0)
+ */
+class ParentPositionInTheVariantTreeValidator extends ConstraintValidator
```
SamirBoulil: Again, I'm not a fan of this positionning thing. Why can't we validate against the family variant structure and then tell if it's possible or not ?

src/Pim/Bundle/CatalogBundle/Resources/config/validation/variantproduct.yml
damien-carcel: Yes, but I need it to for my new validator.

features/import/product_variant/import_variant_products.feature
damien-carcel: I do not agree. What will happen when we will had more variation levels? With this way of validating, we do not depend on the level number.

features/import/product_variant/import_variant_products.feature
damien-carcel: Agreed, but I think it is part of a bigger plot, to clearly say once and for all what is to be tested in integration tests, and when do we add behats or not. Here I just added the functional scenario described in the ticket.

features/import/product_variant/import_variant_products.feature
damien-carcel: That being said, and as @jjanvier argued, the message is not clear from a user point of view.

src/Pim/Component/Catalog/Model/ProductModelInterface.php
damien-carcel: No opinion about it. But we have then to rename `removeProduct` and `getProducts` too.

features/import/product_variant/import_variant_products.feature
```diff
@@ -21,9 +37,31 @@ Feature: Execute a job
...
+      apollon;clothing;master_men;EAN;apollon_blue_medium;100;GRAM;m
+      """
+    And the following job "csv_default_product_import" configuration:
+      | filePath | %file to import% |
+    When I am on the "csv_default_product_import" import job page
+    And I launch the import job
+    And I wait for the "csv_default_product_import" job to finish
+    Then I should see the text "Status: Completed"
+    And I should see the text "skipped 1"
+    And I should see the text "Parent of the variant product \"apollon_blue_medium\" cannot be a root product model"
```
jjanvier: this message is wrong
it can be a root product model

damien-carcel: Oups. Right.

PR #6651, opened 2017-08-30
Improve Docker performance on Mac

torhoehn:
**Description**

For having a better performance while developing on a Mac it's necessary to use docker-sync with the sync strategy unison.

http://docker-sync.io/
http://brewformulas.org/Unison

So the syncing and the docker container can be started by: docker-sync-stack start

### DIFF
--- a/docker-compose.mac.yml.dist
+++ b/docker-compose.mac.yml.dist
@@ -1,0 +1,10 @@
+version: "2"
+
+services:
+  akeneo:
+    volumes:
+      - akeneo-sync:/srv/pim:rw
+
+volumes:
+  akeneo-sync:
+    external: true
--- a/docker-sync.yml.dist
+++ b/docker-sync.yml.dist
@@ -1,0 +1,12 @@
+version: "2"
+
+options:
+  compose-dev-file-path: 'docker-compose.mac.yml'
+
+syncs:
+  akeneo-sync:
+    src: './'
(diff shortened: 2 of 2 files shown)


aRn0D:
Thank you Torben for this contribution. I am far from a docker expert, @damien-carcel and @ahocquard can you have a look to it?

damien-carcel:
I have mixed feelings about placing it in the PIM. This is already present in `akeneo/Dockerfiles` documentation and could be put in the official documentation when a "Docker" section will be added (I plan to do so soon).

But here in the PIM, I don't know, as it is still very experimental. The pure Docker compose part could eventually be tested on the CI (just running `docker-compose up -d` and testing an install and a curl request, for instance). But those scrips will never be tested as we have no MacOS CI (and will probably never have).

I am not against such a contribution, and anyway thank you @torhoehn to make it, but I'd like other team members opinion.

And if to be merged, it would be nice to have it on 1.6 too (as it will be still supported for 7 months), and we will have to put it in EE as the compose file is there too.

ahocquard:
Thanks @torhoehn.

I share the same vision as @damien-carcel .

As no Mac user is using docker for now at Akeneo, it could be pretty difficult to maintain in the future.

aRn0D:
Ok guys thank you for the reviews. Thank you for your contribution but I will close this PR Torben (@torhoehn)

PR #6652, opened 2017-08-30
PIM-6352: Make the sequential work for products

SamirBoulil:

This PR is a small patch that prevents editing products having the same id that product models selected in the datagrid.

It enables this by filtering the ids of the rows selected in the datagrid (removing the product model ids).

This patch will be reworked in an upcoming PR that makes the sequential edit work for product models as well (https://akeneo.atlassian.net/browse/PIM-6360).

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Pim/Bundle/EnrichBundle/Controller/SequentialEditController.php
+++ b/src/Pim/Bundle/EnrichBundle/Controller/SequentialEditController.php
@@ -91,6 +91,15 @@ class SequentialEditController
 
         $parameters = $this->parameterParser->parse($request);
 
+        # PIM-6360: to refactor for the product models sequential edit to work
+        $filteredIds = [];
+        foreach ($parameters['values'] as $id) {
+            if (1 !== preg_match('/^product_model_/', $id)) {
+                $filteredIds[] = $id;
+            }
+        }
+        $parameters['values'] = $filteredIds;
+
         $sequentialEdit = $this->seqEditManager->createEntity(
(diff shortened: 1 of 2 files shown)


src/Pim/Bundle/EnrichBundle/Controller/SequentialEditController.php
```diff
@@ -91,6 +91,14 @@ public function sequentialEditAction(Request $request)
 
         $parameters = $this->parameterParser->parse($request);
 
+        $filteredIds = [];
+        foreach ($parameters['values'] as $id) {
+            if (preg_match('/^product_model_/', $id) !== 1) {
```
grena: yoda

src/Pim/Bundle/EnrichBundle/Controller/SequentialEditController.php
```diff
@@ -91,6 +91,14 @@ public function sequentialEditAction(Request $request)
 
         $parameters = $this->parameterParser->parse($request);
 
+        $filteredIds = [];
```
grena: Could we put a todo with the ticket number to remove it?

SamirBoulil: Ok :)

src/Pim/Bundle/EnrichBundle/Controller/SequentialEditController.php
SamirBoulil: done

PR #6662, opened 2017-08-31
Fix normalization of a product model with null data for image

grena:

I don't know why I did work before, but if there is null as data for image, the normalization of product model was failing.. now it's fixed, we check for null data too.

### DIFF
--- a/src/Pim/Bundle/EnrichBundle/Normalizer/ProductModelNormalizer.php
+++ b/src/Pim/Bundle/EnrichBundle/Normalizer/ProductModelNormalizer.php
@@ -174,7 +174,7 @@ class ProductModelNormalizer implements NormalizerInterface
      */
     private function normalizeImage(?ValueInterface $data, string $format, array $context = []): ?array
     {
-        if (null === $data) {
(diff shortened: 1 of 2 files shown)


src/Pim/Bundle/EnrichBundle/Normalizer/ProductModelNormalizer.php
```diff
@@ -174,7 +174,7 @@ private function getLabels(ProductModelInterface $productModel): array
      */
     private function normalizeImage(?ValueInterface $data, string $format, array $context = []): ?array
```
damien-carcel: Rather than type-hinting `$data` with `?`, shouldn't we try the `null === $data` before calling this private method?

grena: No opinion, as we do 2 controls now, I think it's better to have the 2 in the same place :p

damien-carcel: No problem ;-).

PR #6675, opened 2017-09-01
PIM-6768: Rename the 'SKU' column in 'Id'

SamirBoulil:

- Add configuration in product and association datagrid
- Remove some dynamic stuff related to the identifier

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/features/product/choose_and_order_products_grid_columns.feature
+++ b/features/product/choose_and_order_products_grid_columns.feature
@@ -14,7 +14,7 @@ Feature: Choose and order product grids columns
     And I am on the products grid
 
   Scenario: Successfully display default columns
-    Then I should see the columns Sku, Image, Label, Family, Status, Complete, Created At, Updated At, Groups, Variant products
+    Then I should see the columns ID, Image, Label, Family, Status, Complete, Created At, Updated At, Groups, Variant products
 
   @skip
   Scenario: Successfully hide some columns
@@ -30,7 +30,7 @@ Feature: Choose and order product grids columns
     Given I display the columns SKU, Family and Name
     When I've removed the "name" attribute
     And I am on the products grid
-    Then I should see the columns Sku and Family
+    Then I should see the columns Family and SKU
 
   @jira https://akeneo.atlassian.net/browse/PIM-4861
   Scenario: Successfully display extra columns content like the name when filter on categories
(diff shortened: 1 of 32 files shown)


src/Pim/Bundle/DataGridBundle/Resources/config/attribute_types.yml
```diff
@@ -1,7 +1,7 @@
 parameters:
     pim_datagrid.product.attribute_type.pim_catalog_identifier:
         column:
-            type:        field
+            type:        product_value_field
```
jjanvier: this seems weird in the attribute types grid...

src/Pim/Bundle/EnrichBundle/Resources/config/datagrid/association_product.yml
```diff
@@ -12,6 +12,10 @@ datagrid:
             entity:            '%pim_catalog.entity.product.class%'
             user_config_alias: product-grid
         columns:
+            identifier:
+                label:     SKU
```
jjanvier: ID

jjanvier: SKU do not exist for all clients

src/Pim/Bundle/EnrichBundle/Resources/config/datagrid/association_product.yml
```diff
@@ -44,6 +48,9 @@ datagrid:
             id: ~
         sorters:
             columns:
+                identifier:
+                    data_name: sku
```
jjanvier: should be `identifier` imo

src/Pim/Bundle/EnrichBundle/Resources/config/datagrid/association_product.yml
SamirBoulil: Humm, so it should be dynamic then I need to change the PR.

jjanvier: the identifier attribute is not always called SKU
that's my point

features/product/choose_and_order_products_grid_columns.feature
```diff
@@ -30,17 +30,17 @@ Feature: Choose and order product grids columns
     Given I display the columns SKU, Family and Name
     When I've removed the "name" attribute
     And I am on the products grid
-    Then I should see the columns Sku and Family
+    Then I should see the columns Family and SKU
```
grena: ID no?

SamirBoulil: ID would also work, it's just that the "Given" step we show the SKU attribute so.

src/Pim/Bundle/DataGridBundle/Resources/config/attribute_types.yml
SamirBoulil: From what I understood, this not part of the attribute types grid. This files specifies how to format a product value based on it's attribute type. Since the ID shows the property 'Identifier' of our normalized array.

Attributes of type pim_catalog_identifier (like sku) should be formatted the same way a another product value is => meaning, it should get the value from the 'values' array.

src/Pim/Bundle/EnrichBundle/Resources/config/datagrid/association_product.yml
SamirBoulil: done

PR #6683, opened 2017-09-04
Fix phpspec error

solivier:
Fix phpspec error 

>  [Symfony\Component\Config\Exception\FileLoaderLoadException]                                                                                                                            
  The file "parameters_test.yml" does not exist (in: /home/olivier/Akeneo/PIM/pim-community-dev/app/config) in parameters_test.yml (which is being imported from "/home/olivier/Akeneo/P  
  IM/pim-community-dev/app/config/config_test.yml").  

We do not have this error on versions > 1.5 because when we moved the specs to Bundle folders we did a mistake where the path is wrong (for the app bundle in the constructor) so it does not call the command and creates a CommandResult with a code status to 1 (which means it is an error) and as we only test the type of the returned class the spec is working and does not show any errors as it does not call the command (and will not look for the test_parameters.yml file)

IMHO we should drop this phpspec because it's not a proper phpspec as it really calls the command in test env.

If we decide to keep the spec, the only way to keep it is to:
- creates a parameters_test.yml by default (not a .dist)
- send the result of the command into logs so it won't be displayed in shell (which is only hiding the error) and it implies to modify directly the code.


solivier:
@damien-carcel yes it will be much better :+1:

PR #6686, opened 2017-09-04
PIM-6333: Fix product model validation

damien-carcel:
## Description

This PR fixes various issues on product model import (creation):
- commit 1: Use the same validator that for product (create a service alias) to ensure product value validation
- commit 2: Count imported line correctly => 1 root product model + 1 sub product model imported = 1 read for each step
- commit 3: Fix two behats regarding product model update, broken by the newly introduced immutability of family variant

## Definition Of Done

| Q                                 | A
| --------------------------------- | ---
| Added Specs                       | OK
| Added Behats                      | OK
| Added integration tests           | OK
| Changelog updated                 | -
| Review and 2 GTM                  | Todo
| Micro Demo to the PO (Story only) | -
| Migration script                  | -
| Tech Doc                          | -

`Todo`: Pending / Work in progress
`OK`: Done / Validated
`-`: Not needed

PR #6694, opened 2017-09-05
PIM-6333: Import product model enhancements

damien-carcel:
## Description

This PR brings a few enhancements to existing product model creation/update.
- **first commit**: during import, non existing fields are handled the same way than for products, with the same error message
- **second commit**: fix the behavior of the previous PR #6686 after talking with @Delphineray => no more decrement of read product model, but a clear message indicating some have been skipped (product models with parent during the first step, product models without parent during the second)
- **third commit**: add some new business rules to fit with the variant product ones:
    - a sub product model can be imported without variant family, and will in this case inherit the one from its parent
    - if a family variant is specified for a sub product model, it cannot be different from the one of the parent

## Definition Of Done

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/features/import/product_model/import_business_rules.feature
+++ b/features/import/product_model/import_business_rules.feature
@@ -217,3 +217,17 @@ Feature: Create product models through CSV import
       code;parent;family_variant;categories;collection;description-en_US-ecommerce;erp_name-en_US;price;color;variation_name-en_US;composition;size;ean;sku;weight
       code-003;code-001;clothing_color_size;master_men_blazers;;;;;blue;Blazers;composition;;;;
       """
+
+  Scenario: Skip import with a unexpected field
+    Given the following CSV file to import:
+      """
+      code;parent;family_variant;comment
+      code-001;;clothing_color_size;"my comment"
+      """
+    And the following job "csv_catalog_modeling_product_model_import" configuration:
+      | filePath | %file to import% |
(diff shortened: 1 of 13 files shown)


features/import/product_model/import_business_rules.feature
```diff
@@ -217,3 +218,17 @@ Feature: Create product models through CSV import
       code;parent;family_variant;categories;collection;description-en_US-ecommerce;erp_name-en_US;price;color;variation_name-en_US;composition;size;ean;sku;weight
       code-003;code-001;clothing_color_size;master_men_blazers;;;;;blue;Blazers;composition;;;;
       """
+
+  Scenario: Skip import with a not expected field
```
SamirBoulil: not expected = unexpected :)

src/Pim/Component/Connector/spec/ArrayConverter/FlatToStandard/ProductModelSpec.php
```diff
@@ -172,4 +163,38 @@ function it_convert_flat_product_model_to_standard_format(
             ]
         ]);
     }
+
+    function it_throws(
```
SamirBoulil: when does it throw?

damien-carcel: 🤦‍

PR #6696, opened 2017-09-06
PIM-6333: Fixing last glitches on product model creation through imports

damien-carcel:
## Description

- Fixing/enhancing messages when validating the parent of product models.
- During import, message "Skipped product (no differences)" becomes "Skipped product model (no differences)".

It also fixes a typo in a colomn name ("EAN" instead of "ean" => attribute codes that make column headers in imported files are case sensible) that was breaking a few behats.

## Definition Of Done

`OK`: Done / Validated
`-`: Not needed


src/Pim/Component/Catalog/Validator/Constraints/ProductModelPositionInTheVariantTreeValidator.php
```diff
@@ -36,15 +39,25 @@ public function validate($productModel, Constraint $constraint): void
...
-        if (false === $productModel->getParent()->isRootProductModel()) {
-            $this->context->buildViolation(ProductModelPositionInTheVariantTree::INVALID_PARENT)->addViolation();
-        }
-
-        $productModelPosition = $productModel->getVariationLevel();
         $numberOfLevel = $productModel->getFamilyVariant()->getNumberOfLevel();
 
-        if ($numberOfLevel === $productModelPosition) {
-            $this->context->buildViolation(ProductModelPositionInTheVariantTree::INVALID_PARENT)->addViolation();
+        if (2 === $numberOfLevel && false === $productModel->getParent()->isRootProductModel()) {
```
jjanvier: what if no parent? 
$productModel->getParent()

damien-carcel: It is tested above: if we have a root product model (so no parent), we early return, no need to validate the parent if there is none.

PR #6706, opened 2017-09-07
PIM-6743: Prevent a regular product to become a variant one

damien-carcel:
## Description

Before this PR, it was already not possible to set a parent to a regular product, but an `InvalidObjectException` was thrown, which is not caught by our import/export stack. So such an attempt to transform a regular product into a variant one resulted in the failure of the import.

We now throw an `InvalidPropertyException`, which allows us to skip the product without the import failing. 

## Definition Of Done

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/features/import/variant_product/import_business_rules.feature
+++ b/features/import/variant_product/import_business_rules.feature
@@ -86,3 +86,26 @@ Feature: Import variant products
       apollon_blue;clothing;master_men;12345;apollon_blue_medium;100;GRAM;
       apollon_blue;clothing;master_men;67890;apollon_blue_large;100;GRAM;
       """
+
+  Scenario: Successfully skip a non variant product when we try to set a parent
+    Given the following product:
+      | sku             |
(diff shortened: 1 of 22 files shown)


features/import/variant_product/import_business_rules.feature
```diff
@@ -77,9 +103,9 @@ Feature: Import variant products
     And I wait for the "csv_default_product_import" job to finish
     Then I should see the text "Status: Completed"
     And I should see the text "skipped 1"
-    And I should see the text "Attribute \"size\" cannot be empty, as it is defined as an axis for this entity: apollon_blue_medium"
+    And I should see the text "Product \"regular_product\" cannot have a parent."
```
SamirBoulil: "parent property" ?

damien-carcel: No, just parent, seen with Delph.

PR #6708, opened 2017-09-08
PIM-6333: Fix sub product model import

damien-carcel:
## Description

During product model import, we filter the flat data to keep only values that should be here, like for products. It means we keep only values corresponding to attribute declared on the variant attribute set.

Problem was the filter didn't handle the case where sub product models were imported without family variant. This PR fixes the problem by using the family variant of the parent when needed.

## Definition Of Done

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/features/import/product_model/import_business_rules.feature
+++ b/features/import/product_model/import_business_rules.feature
@@ -89,3 +89,43 @@ Feature: Create product models through CSV import
     And I should see the text "skipped product model with parent 1"
     And I should see the text "skipped product model without parent 1"
     And I should see the text "read lines 2"
+
+  Scenario: Successfully create product models without specifying the family variant of the sub product models
+    Given the following CSV file to import:
+      """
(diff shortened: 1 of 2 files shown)


features/import/product_model/import_business_rules.feature
```diff
@@ -89,3 +89,43 @@ Feature: Create product models through CSV import
...
+    And the following job "csv_catalog_modeling_product_model_import" configuration:
+      | filePath | %file to import% |
+    When I am on the "csv_catalog_modeling_product_model_import" import job page
+    And I launch the import job
+    And I wait for the "csv_catalog_modeling_product_model_import" job to finish
+    Then I should see the text "Status: Completed"
+    And I should see the text "read lines 3"
+    And I should see the text "processed 1"
+    And I should see the text "processed 2"
+    And I should see the text "skipped product model with parent 2"
```
SamirBoulil: product model(s) ? because we don't know what 2 is and what 1 is ? (maybe the UI is better than that but I don't find the message clear enough.

damien-carcel: in this case, we have to do this for all the other entities, I think.

PR #6711, opened 2017-09-08
PIM-6448: fix setParent API on both variants and models

jjanvier:

### DIFF
--- a/src/Pim/Component/Catalog/Model/EntityWithFamilyVariantInterface.php
+++ b/src/Pim/Component/Catalog/Model/EntityWithFamilyVariantInterface.php
@@ -34,9 +34,14 @@ interface EntityWithFamilyVariantInterface extends EntityWithFamilyInterface
     public function getVariationLevel(): int;
 
     /**
-     * @return ProductModelInterface|null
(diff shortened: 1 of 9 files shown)


src/Pim/Component/Catalog/Model/EntityWithFamilyVariantInterface.php
```diff
@@ -34,9 +34,14 @@ public function setFamilyVariant(FamilyVariantInterface $familyVariant): void;
     public function getVariationLevel(): int;
 
     /**
-     * @return ProductModelInterface|null
+     * @return EntityWithFamilyVariantInterface|null
      */
-    public function getParent(): ?ProductModelInterface;
+    public function getParent(): ?EntityWithFamilyVariantInterface;
```
grena: Hm, it means `VariantProductInteface` can be a parent, I'd prefer to explicitly have `ProductModelInterface`, so, to keep as it

jjanvier: seen IRL

jjanvier:
only `features/attribute/browse/view.feature:13` failing
related to the number of items in the grid

PR #6713, opened 2017-09-08
PIM-6743: Add parent to products from the "catalog_modeling" catalog

damien-carcel:
## Description

This PR updates the "catalog_modeling" catalog used for testing, by adding parents to the products, making them variant products.

This reveals a glitch in variant attribute set: it is possible to import an attribute set without specifying the axis attributes as regular attributes, to avoid duplication in the imported file. But in this case, the axes were not present in the attribute collections. This is fixed in the second commit of this PR.

## Definition Of Done

`OK`: Done / Validated
`-`: Not needed


src/Pim/Component/Catalog/tests/integration/Normalizer/Standard/FamilyVariantIntegration.php
```diff
@@ -26,7 +26,7 @@ public function testFamilyVariant()
                 [
                     'level' => 1,
                     'axes' => ['color'],
-                    'attributes' => ['image', 'variation_name', 'variation_image', 'composition', 'color', 'material'],
```
damien-carcel: Attribute was removed from the family variant.

tests/TestCase.php
```diff
@@ -29,9 +29,6 @@ protected function setUp()
 
         $fixturesLoader = $this->getFixturesLoader($configuration, $databaseSchemaHandler);
         $fixturesLoader->load();
-
-        $this->indexProductModels();
```
grena: 👍

PR #6736, opened 2017-09-13
PIM-6444: Remove variant groups

damien-carcel:
## Description

This PR removes every functional aspects of the variant groups from Akeneo PIM:
- permissions (ACLs)
- bulk actions
- import and exports (including job fixtures)
- every UI elements

This PR does not remove everything however: are still present the models and ORM mappings, most validators… This will be done in another PR, this one being already large enough.

It also does not break the API output: `variant_group` product field is still present, only always null.
It is of course not possible to add a product into a variant group  :-). Another PR will handle that, but behavior is still to discuss (considering the field as an alias for the new `parent` field).

## Definition Of Done

`OK`: Done / Validated
`-`: Not needed


grena:
🔥 🔥 🔥 🔥 🔥 🔥

PR #6746, opened 2017-09-14
Optimize AbstractProduct::getIdentifier

jmleroux:
Optimize AbstractProduct::getIdentifier to not iterate on all values each time we get the identifier.
Noticed after digging for an export memory leak where i seen we were iterating on values thousand times.

### DIFF
--- a/src/Pim/Component/Catalog/Model/AbstractProduct.php
+++ b/src/Pim/Component/Catalog/Model/AbstractProduct.php
@@ -79,7 +79,7 @@ abstract class AbstractProduct implements ProductInterface
     protected $normalizedData;
 
     /** @var string */
-    private $identifierAttributeCode;
+    protected $identifierAttributeCode;
 
     /**
      * Constructor
--- a/src/Pim/Component/Catalog/Model/AbstractProduct.php
+++ b/src/Pim/Component/Catalog/Model/AbstractProduct.php
@@ -39,7 +39,7 @@ abstract class AbstractProduct implements ProductInterface
      */
     protected $scope;
 
-    /** @var ArrayCollection */
+    /** @var ArrayCollection|ProductValueInterface[] */
     protected $values;
 
     /** @var array */
@@ -63,7 +63,7 @@ abstract class AbstractProduct implements ProductInterface
     /** @var bool $enabled */
     protected $enabled = true;
 
-    /** @var ArrayCollection $groups */
+    /** @var ArrayCollection|GroupInterface[] */
     protected $groups;
 
     /** @var array */
@@ -77,6 +77,9 @@ abstract class AbstractProduct implements ProductInterface
 
     /** @var array */
     protected $normalizedData;
+
+    /** @var string */
+    private $identifierAttributeCode;
 
     /**
      * Constructor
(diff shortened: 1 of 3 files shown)


src/Pim/Component/Catalog/Model/AbstractProduct.php
```diff
@@ -394,13 +397,22 @@ public function getFamilyId()
...
+            foreach ($this->values as $value) {
+                $attribute = $value->getAttribute();
+                if (AttributeTypes::IDENTIFIER === $attribute->getAttributeType()) {
+                    $this->identifierAttributeCode = $attribute->getCode();
+                }
             }
         }
 
-        throw new MissingIdentifierException($this);
+        $identifier = $this->getValue($this->identifierAttributeCode);
```
anaelChardan: Is getValue more interesting than the former foreach in terms of performance ?

src/Pim/Component/Catalog/Model/AbstractProduct.php
```diff
@@ -78,6 +78,9 @@
     /** @var array */
     protected $normalizedData;
 
+    /** @var string */
+    private $identifierAttributeCode;
```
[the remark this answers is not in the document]
jmleroux: :+1:

src/Pim/Component/Catalog/Model/AbstractProduct.php
jmleroux: Yes: it uses the `indexedValues`

jmleroux: https://github.com/jmleroux/pim-community-dev/blob/3d7927db2580b2429fe2a01705683b06d08ba52d/src/Pim/Component/Catalog/Model/AbstractProduct.php#L297

jmleroux: I could have used an cached `$identifier` value which would have been even faster, but this way i keep all the other features like `removeValue`, etc. Even if it has no real interest on an identifier... ;)

src/Pim/Component/Catalog/Model/AbstractProduct.php
```diff
@@ -187,7 +190,7 @@ public function addValue(ProductValueInterface $value)
     {
         $this->values[] = $value;
         $this->indexedValues[$value->getAttribute()->getCode()][] = $value;
-        $value->setEntity($this);
+        $value->setProduct($this);
```
damien-carcel: 👍

src/Pim/Component/Catalog/Model/AbstractProduct.php
```diff
@@ -39,7 +39,7 @@
      */
     protected $scope;
 
-    /** @var ArrayCollection */
+    /** @var ArrayCollection|ProductValueInterface[] */
```
damien-carcel: Why is it sometime an array collection, and sometime a regular array of objects?
It seems to me it should always be `Collection` and nothing else.

src/Pim/Component/Catalog/Model/FamilyInterface.php
```diff
@@ -4,6 +4,7 @@
 
 use Akeneo\Component\Localization\Model\TranslatableInterface;
 use Akeneo\Component\Versioning\Model\VersionableInterface;
+use Doctrine\Common\Collections\ArrayCollection;
```
damien-carcel: I imagine you add it because it was in a PHPDoc somewhere in the class. But in this case, it should be `Collection` (the interface of `ArrayCollection`).

damien-carcel:
A few minor comments and a question, but GTM anyway.

jmleroux:
Won't do.

PR #6748, opened 2017-09-15
Merge branch 1.7 into master

solivier:
## Definition Of Done

`OK`: Done / Validated
`-`: Not needed


src/Pim/Bundle/CatalogBundle/Doctrine/ORM/Filter/StringFilter.php
```diff
@@ -132,22 +132,24 @@ protected function prepareCondition($backendField, $operator, $value)
             $value = '';
         }
 
+        $likeValue = str_replace(['%', '_'], ['\\%', '\\_'], $value);
+
         switch ($operator) {
```
damien-carcel: This class (and many other in that folder) are useless since we use ES for the PQB. It will be removed in a dedicated PR.

PR #6768, opened 2017-09-20
PIM-6800: Prevent media thumbnais in PDF to be cropped

damien-carcel:
## Description

This PR:
- update `liip/imagine-bundle` and  `imagine/imagine` to latest versions,
- fix a bug in the way we retrieve values from products in the PdfRender (`ProductInterface::getValue` will return null if we pass a locale and/or a scope code when the corresponding attribute is not localisable and/or scopable), that prevented images to appear in the generated PDF,
- prevent the thumbnail generated in the PDF to be cropped, and reduce its size instead.

## Definition Of Done

`OK`: Done / Validated
`-`: Not needed


src/Pim/Bundle/PdfGeneratorBundle/Renderer/ProductPdfRenderer.php
```diff
@@ -161,16 +160,25 @@ protected function getGroupedAttributes(ProductInterface $product, $locale)
...
         foreach ($this->getAttributes($product, $locale) as $attribute) {
             if (AttributeTypes::IMAGE === $attribute->getType()) {
-                $media = $product->getValue($attribute->getCode(), $locale, $scope)->getData();
+                $mediaValue = $product->getValue(
+                    $attribute->getCode(),
+                    $attribute->isLocalizable() ? $locale : null,
+                    $attribute->isScopable() ? $scope : null
+                );
+
+                if (null === $mediaValue) {
```
solivier: I know it's a detail but :
if (null !== $mediaValue) {
    $media = $mediaValue->getData();
}

is more elegant. And is not in contradiction with the foreach :)

PR #6769, opened 2017-09-20
Use Elasticsearch 5.6.1 with Docker Compose

damien-carcel:
## Description

Update the file `docker-compose.yml.dist` to use the latest Elasticsearch version.

Jenkinsfile cannot be updated for the moment as the current CI cannot handle the official Elasticsearch images. It still uses the Docker library image, which is deprecated and will never be updated to this version (last version available is 5.5). This problem will be solved with our next CI, which will use the official image, in version 5.6.

## Definition Of Done

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/docker-compose.yml.dist
+++ b/docker-compose.yml.dist
@@ -103,7 +103,7 @@ services:
       - behat
 
   elasticsearch:
-    image: docker.elastic.co/elasticsearch/elasticsearch:5.5.2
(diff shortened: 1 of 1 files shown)


damien-carcel:
Will not be merged as Akeneo 2.0 is tested with ES 5.5 on our CI.

PR #6772, opened 2017-09-20
API-246: apply permissions on products import

skeleton:

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Pim/Component/Connector/Processor/Denormalization/ProductProcessor.php
+++ b/src/Pim/Component/Connector/Processor/Denormalization/ProductProcessor.php
@@ -12,6 +12,8 @@ use Pim\Component\Catalog\Builder\ProductBuilderInterface;
 use Pim\Component\Catalog\Comparator\Filter\FilterInterface;
 use Pim\Component\Catalog\Model\ProductInterface;
 use Pim\Component\Connector\Processor\Denormalization\AttributeFilter\AttributeFilterInterface;
+use Symfony\Component\Security\Core\Exception\AccessDeniedException;
+use Symfony\Component\Security\Core\Exception\InvalidArgumentException;
 use Symfony\Component\Validator\ConstraintViolationListInterface;
 use Symfony\Component\Validator\Validator\ValidatorInterface;
 
@@ -103,7 +105,7 @@ class ProductProcessor extends AbstractProcessor implements ItemProcessorInterfa
         $filteredItem = $this->filterItemData($item);
 
         $parent = $item['parent'] ?? '';
-        $product = $this->findOrCreateProduct($identifier, $familyCode, $parent);
+        $product = $this->findOrCreateProduct($identifier, $familyCode, $item, $parent);
 
         if (false === $itemHasStatus && null !== $product->getId()) {
             unset($filteredItem['enabled']);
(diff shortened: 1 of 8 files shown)


tests/JobLauncher.php
```diff
@@ -203,4 +203,58 @@ public function launchImport(
...
+     * Launch an import in a subprocess.
+     *
+     * @param string $command
+     * @param string $jobCode
+     * @param string $content
+     * @param string $username
+     *
+     * @throws \Exception
+     */
+    public function launchSubProcessImport(string $command, string $jobCode, string $content, string $username = null)
```
ahocquard: void

tests/JobLauncher.php
```diff
@@ -203,4 +203,58 @@ public function launchImport(
...
+    {
+        $filePath = sys_get_temp_dir() . DIRECTORY_SEPARATOR. self::EXPORT_DIRECTORY . DIRECTORY_SEPARATOR . 'import_.csv';
```
ahocquard: useless for import

tests/JobLauncher.php
```diff
@@ -203,4 +203,58 @@ public function launchImport(
...
+        );
+
+        $process = new Process($command);
+        $process->run();
+
+        if (!$process->isSuccessful()) {
+            throw new \Exception(sprintf('Import failed, "%s".', $process->getOutput() . PHP_EOL . $process->getErrorOutput()));
+        }
+
+        if (!is_readable($filePath)) {
```
ahocquard: useless for import

tests/JobLauncher.php
```diff
@@ -203,4 +203,58 @@ public function launchImport(
...
+     * @throws \Exception
+     */
+    public function launchSubProcessImport(string $command, string $jobCode, string $content, string $username = null)
+    {
+        $filePath = sys_get_temp_dir() . DIRECTORY_SEPARATOR. self::EXPORT_DIRECTORY . DIRECTORY_SEPARATOR . 'import_.csv';
+        if (file_exists($filePath)) {
+            unlink($filePath);
+        }
+
+        $importDirectoryPath = sys_get_temp_dir() . DIRECTORY_SEPARATOR . self::IMPORT_DIRECTORY;
```
ahocquard: you don't handle the fixtures with this function, check `launchImport`

src/Pim/Component/Connector/Processor/Denormalization/ProductProcessor.php
```diff
@@ -191,25 +196,29 @@ protected function filterItemData(array $item)
...
-        }
+            if (null === $product) {
+                $product = $this->productBuilder->createProduct($identifier, $familyCode);
+            }
 
-        if (null === $product) {
-            $product = $this->productBuilder->createProduct($identifier, $familyCode);
+            return $product;
+        } catch (AccessDeniedException $e) {
+            $this->skipItemWithMessage($item, $e->getMessage(), $e);
```
ahocquard: Does it work with `AccessDeniedException` ? It returns `ProductInterface` (no null accepted), if you are in this catch, it will returns null, and it will throw a PHP exception because the function does not accept a null in output. (that's why I prefer explicit return)

skeleton: `skipItemWithMessage()` returns an exception, no need to return null

tests/JobLauncher.php
skeleton: why do you need fixtures here?

src/Pim/Component/Connector/Processor/Denormalization/ProductProcessor.php
ahocquard: ah ok (then, the naming "skip" is very confusing but not your responsibility ;) )

tests/JobLauncher.php
ahocquard: I'm waiting that `launchImport` &&  `launchImportInSubProcess` to have the same signature, just the way to launch is different.

PR #6773, opened 2017-09-21
Fix configurations of the bundles

jjanvier:
The more configuration we have in the files `pim.yml` and `pimee.yml`, the less we have to update the `config.yml` files of standard editions and EE dev edition.


src/Pim/Bundle/EnrichBundle/Resources/config/pim.yml
```diff
@@ -49,3 +50,22 @@ pim_catalog:
             - { value: 'MM/dd/yyyy', label: 'mm/dd/yyyy' }
             - { value: 'dd/MM/yyyy', label: 'dd/mm/yyyy' }
             - { value: 'dd.MM.yyyy', label: 'dd.mm.yyyy' }
+
+akeneo_storage_utils:
```
damien-carcel: Strange to have that in the `EnrichBundle`…

jjanvier:
Ideally, I would put that elsewhere. But originally, I chose Enrich as it seems the closest thing to "the app configuration, the thing that glues everything together". 

There is no way those things go in Catalog. Catalog is our main domain, not the application itself.

PR #6775, opened 2017-09-21
PIM-6798: Fix select attribute copiers

damien-carcel:
## Description

There is a bug into the simple and multi select attribute copiers. We were directly copying the options from the source value into the destination value.

Problem is, when we copy, inside one product, the value from an attribute to the one of another attribute (where option codes are identical between the two attributes, of course), we end up with the destination attribute containing an option (the relation on the ID in database) from the source attribute, which should never happen!

This PR fixes this behavior by ensuring the copy is made through the option code, not the entire object, validating in the process that this option really exists for the destination attribute.

## Definition Of Done

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Pim/Component/Catalog/Updater/Copier/MultiSelectAttributeCopier.php
+++ b/src/Pim/Component/Catalog/Updater/Copier/MultiSelectAttributeCopier.php
@@ -4,8 +4,10 @@ namespace Pim\Component\Catalog\Updater\Copier;
 
 use Pim\Component\Catalog\Builder\ProductBuilderInterface;
 use Pim\Component\Catalog\Model\AttributeInterface;
+use Pim\Component\Catalog\Model\AttributeOptionInterface;
 use Pim\Component\Catalog\Model\ProductInterface;
 use Pim\Component\Catalog\Model\ProductValueInterface;
+use Pim\Component\Catalog\Repository\AttributeOptionRepositoryInterface;
 use Pim\Component\Catalog\Validator\AttributeValidatorHelper;
 
 /**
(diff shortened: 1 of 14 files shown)


src/Pim/Component/Catalog/Updater/Copier/MultiSelectAttributeCopier.php
```diff
@@ -116,11 +130,54 @@ protected function removeOptions(ProductValueInterface $toValue)
...
+    }
+
+    /**
+     * @param AttributeOptionInterface $fromAttributeOption ,
+     * @param AttributeInterface       $toAttribute
+     *
+     * @throws \InvalidArgumentException
+     * @return AttributeOptionInterface
+     */
+    protected function getOptionForDestinationAttribute(
```
grena: `getMatchingOptionForAttribute()` ? "ForDestinationAttribute" seems a bit weird

src/Pim/Component/Catalog/Updater/Copier/SimpleSelectAttributeCopier.php
```diff
@@ -88,10 +98,60 @@ protected function copySingleValue(
...
-            $toValue->setOption($fromValue->getData());
+            $toValue->setOption($this->getOptionForDestinationAttribute($fromValue, $toAttribute));
+        }
+    }
+
+    /**
+     * Return the destination option corresponding to the one of the original
+     * value.
+     * Return "null" of the original value was empty, so we ensure the
+     * destination value will be empty too.
```
grena: You should put this PHPdoc on the other class too :)

damien-carcel: Yep, I wanted too and forgot :/

PR #6792, opened 2017-09-25
PIM-6820: Validate variant axes during imports

damien-carcel:
## Description

Currently, when creating products and adding them to a variant group, we validate that there is not already a product in this variant group with the same axis values, as each axis values combination are unique for a variant group. For example, you cannot have to red t-shirts in size XL in the same variant group, if color and size are this variant group axes.

However, we did not validate this axis values uniqueness during imports inside one batch. So it was possible to import several products for a same variant group, with the same axis value combination.

This PR introduces a `UniqueAxesCombinationSet` that stores the axis value combinations and is reset after each batch. Its behavior is based on the `UniqueValuesSet`, which does the same for attributes with unique values.

## Definition Of Done

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/CHANGELOG-1.7.md
+++ b/CHANGELOG-1.7.md
@@ -5,6 +5,7 @@
 - PIM-6804: Fix the positioning (z-index) of the datafilter widgets
 - PIM-6491: Fix file extension validation on import job upload
 - PIM-6798: Fix the simple-select and multi-select copiers
+- PIM-6820: During an import, prevent to add in a variant group two products with same variant axes
 
 # 1.7.8 (2017-08-22)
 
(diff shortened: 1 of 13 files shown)


CHANGELOG-1.7.md
```diff
@@ -5,6 +5,7 @@
 - PIM-6804: Fix the positioning (z-index) of the datafilter widgets
 - PIM-6491: Fix file extension validation on import job upload
 - PIM-6798: Fix the simple-select and multi-select copiers
+- PIM-6820: Prevent the adding in the same variant group of two products with same variant axes during an import.
```
grena: > During an import, prevent to add in a variant group two products with same variant axes

PR #6796, opened 2017-09-25
Api 336: create a command to push job into the queue

ahocquard:

`akeneo:batch:job` execute a job directly in synchrone.
It accepts a job execution id as well.

So, in order to push jobs into the queue from the CLI, a new command has been created:
`akeneo:batch:publish-job`

It does not accept a job execution id, but it accepts all the other parameters of the command `akeneo:batch:job`.

`OK`: Done / Validated
`-`: Not needed


src/Akeneo/Bundle/BatchQueueBundle/Command/BatchQueueCommand.php
```diff
@@ -0,0 +1,308 @@
...
+use Symfony\Component\Validator\Validator\ValidatorInterface;
+
+/**
+ * Push a registered job instance to execute into the job execution queue.
+ *
+ * @author    Alexandre Hocquard <alexandre.hocquard@akeneo.com>
+ * @copyright 2017 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+class BatchQueueCommand extends ContainerAwareCommand
```
skeleton: you can't guess the purpose of the command with this naming

src/Akeneo/Bundle/BatchQueueBundle/Command/BatchQueueCommand.php
```diff
@@ -0,0 +1,308 @@
...
+{
+    public const COMMAND_NAME = 'akeneo:batch:publish-job';
```
skeleton: add "queue" in the name of the command would be better imo

skeleton: do you really need to create a constant for the command name?

src/Akeneo/Bundle/BatchQueueBundle/Command/BatchQueueCommand.php
```diff
@@ -0,0 +1,308 @@
...
+    public const EXIT_SUCCESS_CODE = 0;
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function configure()
+    {
+        $this
+            ->setName(self::COMMAND_NAME)
+            ->setDescription('Push a registered job instance to execute into the job execution queue')
```
skeleton: push or publish ?

src/Akeneo/Bundle/BatchQueueBundle/Command/BatchQueueCommand.php
```diff
@@ -0,0 +1,308 @@
...
+     */
+    protected function getValidator(): ValidatorInterface
+    {
+        return $this->getContainer()->get('validator');
+    }
+
+    /**
+     * @return MailNotifier
+     */
+    protected function getMailNotifier(): MailNotifier
```
skeleton: Already said that, but from my point of view, create a dedicated method to use it one or few time is : 
- useless
- make the code less readable (you scroll all the time into your file)
- the better way to have dead code (like here, this method is never called)

ahocquard: And I disagree with that. It is:
- useful because it allows code completion
- no need to scroll as you have code completion

And in a more general, it give it a name (not relevant in this case as it is a setter).

src/Akeneo/Bundle/BatchQueueBundle/Command/BatchQueueCommand.php
ahocquard: I kept the same naming as `BatchCommand` as it is almost the same command.

src/Akeneo/Bundle/BatchQueueBundle/Command/BatchQueueCommand.php
ahocquard: `publish-job-to-queue` ?

ahocquard: I use it in the tests. What does it cost ?

skeleton: almost nothing, I guess there is no impact if a user override the constant?

ahocquard: I don't think so

src/Akeneo/Bundle/BatchQueueBundle/Command/PublishJobToQueueCommand.php
```diff
@@ -0,0 +1,308 @@
...
+        $this
+            ->setName(self::COMMAND_NAME)
+            ->setDescription('Publish a registered job instance to execute into the job execution queue')
+            ->addArgument('code', InputArgument::REQUIRED, 'Job instance code')
+            ->addOption(
+                'config',
+                'c',
+                InputOption::VALUE_REQUIRED,
+                'Override job configuration (formatted as json. ie: ' .
+                'php bin/console akeneo:batch:publish-job -c "{\"filePath\":\"/tmp/foo.csv\"}" acme_product_import)'
```
skeleton: publish-job-to-queue

src/Akeneo/Bundle/BatchQueueBundle/Command/PublishJobToQueueCommand.php
```diff
@@ -0,0 +1,308 @@
...
+     */
+    protected function getValidator(): ValidatorInterface
+    {
+        return $this->getContainer()->get('validator');
+    }
+
+    /**
+     * @return MailNotifier
+     */
+    protected function getMailNotifier(): MailNotifier
```
skeleton: so... you really don't want to remove dead code?!

PR #6809, opened 2017-09-26
Fix integration tests

SamirBoulil:

An attempt at fixing the way we reset the indexes in the integration tests.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/tests/FixturesLoader.php
+++ b/tests/FixturesLoader.php
@@ -74,6 +74,8 @@ class FixturesLoader
     public function load()
     {
         $this->container->get('akeneo_elasticsearch.client.product')->resetIndex();
+        $this->container->get('akeneo_elasticsearch.client.product_model')->resetIndex();
+        $this->container->get('akeneo_elasticsearch.client.product_and_product_model')->resetIndex();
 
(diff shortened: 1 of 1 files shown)


nidup:
I think @damien-carcel & @ahocquard does this on purpose to avoid an issue with the index clean

tests/FixturesLoader.php
```diff
@@ -87,16 +89,13 @@ public function load()
             $this->clearAclCache();
             $this->createUserSystem();
 
-            $this->container->get('akeneo_elasticsearch.client.product_and_product_model')->resetIndex();
```
ahocquard: no, it was done like this on purpose. Dleeting the two indexes `akeneo_elasticsearch.client.product` and `'akeneo_elasticsearch.client.product_and_product_model'` consecutively resulting in an error: the first delete is not totally done.
It's like the second delete prevents the first one to finish.

So, we wait 2 seconds between each delete. And 2 seconds is approximately the time to insert the mysqldump. I t avoids to put a `sleep(2)`.

But it does not work perfectly, we have "random".

SamirBoulil:
Ok. It doesn't fix our issue anyways. @ahocquard explained me this yesterday.

PR #6814, opened 2017-09-26
PIM-6451: enrich product variant

grena:

This PR adds on the PEF modules related to variant structure, for variant products.

**Bonus (PIM-6348 fixes)**
This PR also contains fixes for PIM-6348:
- Error 500 when open the PEF for variant products with metrics
- If multiple axes, not all label axes displayed in the label of the component, only SIZE instead of COLOR, SIZE (ex: ERIS) + if select COLOR, SIZE: Pink, XS.
- Have all the labels clickable in the navigation component to go the previous level (ex: COLOR: Blue, not just Blue)
- No space before ":" in the navigation component labels


src/Pim/Bundle/EnrichBundle/Normalizer/ProductNormalizer.php
```diff
@@ -275,4 +278,46 @@ protected function normalizeImage(?ValueInterface $data, $format, $context = [])
...
+    ): array {
+        $meta = [
+            'variant_navigation'        => [],
+            'attributes_for_this_level' => [],
+            'attributes_axes'           => [],
+            'parent_attributes'         => [],
+            'family_variant'            => null
+        ];
+
+        if (!$product instanceof VariantProductInterface) {
```
jjanvier: shouldn't it be more logical to pass a VariantProductInterface as argument and to do the if in the calling method?

grena: Well, in fact no as I still need empty meta for "normal" products

PR #6816, opened 2017-09-26
Integration tests 1.7

damien-carcel:
## Description

Currently, integration tests are split manually, into small test suites, to be run in parallel on the CI. This is not ideal, as the suites are not equivalent in size, and so in running time, and we can forgot to add new set of tests in the existing suites.

This PR brings back into 1.7 the work that @ahocquard did on master (2.0), but only for CE and ORM. 
The goal is only to ensure all tests are ran, including new ones, whitout having to do anything else than writing the test.

EE tests ran from CE pull requests could be added later, as well as ODM builds. I also had to deactivate a broken test on variant group normalization (which was not ran before anyway...). It will be fixed later in another PR.

This PR also improves build time by performing a `composer update` only during checkout stage.

## Definition Of Done

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Pim/Bundle/VersioningBundle/tests/integration/Normalizer/Flat/VariantGroupIntegration.php
+++ b/src/Pim/Bundle/VersioningBundle/tests/integration/Normalizer/Flat/VariantGroupIntegration.php
@@ -14,6 +14,8 @@ class VariantGroupIntegration extends AbstractFlatNormalizerTestCase
     public function testVariantGroup()
     {
         $variantGroup = $this->get('pim_catalog.repository.group')->findOneByIdentifier('variantA');
+
+        $this->get('pim_serializer');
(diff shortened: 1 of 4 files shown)


src/Pim/Bundle/VersioningBundle/tests/integration/Normalizer/Flat/VariantGroupIntegration.php
```diff
@@ -11,6 +11,7 @@
  */
 class VariantGroupIntegration extends AbstractFlatNormalizerTestCase
 {
+    /** TODO to fix
     public function testVariantGroup()
     {
         $variantGroup = $this->get('pim_catalog.repository.group')->findOneByIdentifier('variantA');
```
ahocquard: don't forget to fix it with my fix

damien-carcel: Done

PR #6817, opened 2017-09-26
API-372: Create variant product though API

solivier:

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Pim/Bundle/ApiBundle/Controller/ProductController.php
+++ b/src/Pim/Bundle/ApiBundle/Controller/ProductController.php
@@ -284,7 +284,7 @@ class ProductController
 
         if (isset($data['parent'])) {
             $data = $this->productAttributeFilter->filter($data);
-            $product = $this->variantProductBuilder->createProduct($data['identifier']);
+            $product = $this->variantProductBuilder->createProduct();
         } else {
             $product = $this->productBuilder->createProduct();
         }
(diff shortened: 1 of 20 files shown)


src/Pim/Bundle/ApiBundle/Controller/ProductController.php
```diff
@@ -271,7 +280,12 @@ public function createAction(Request $request): Response
 
         $data = $this->populateIdentifierProductValue($data);
 
-        $product = $this->productBuilder->createProduct();
+        if (isset($data['parent'])) {
+            $data = $this->productAttributeFilter->filter($data);
+            $product = $this->variantProductBuilder->createProduct($data['identifier']);
```
LaurentPetard: Is this condition really mandatory to create the variant product ? It seems that it could works with  `$this->productBuilder->createProduct();`

solivier: I tried to do this, but it fails for some tests. so I will investigate more. But for the moment i think we will leave it like this. And it has been done this way for the import. Thank for pointing this out.

src/Pim/Bundle/ApiBundle/Controller/ProductController.php
```diff
@@ -271,7 +280,12 @@ public function createAction(Request $request): Response
 
         $data = $this->populateIdentifierProductValue($data);
 
-        $product = $this->productBuilder->createProduct();
+        if (isset($data['parent'])) {
+            $data = $this->productAttributeFilter->filter($data);
```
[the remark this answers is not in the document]
solivier: This filter remove the optional attribute and unsure that the data in the common attributes and axis from the product model remain the same.

PR #6839, opened 2017-09-27
TIP-749: upgrade total fields limit in ES

jjanvier:
With Leo we imported a catalog they use at presales.
This catalog contains 1000 attributes.

We already reach the limit of 3000 fields in ES with this catalog.
We decided with Benoit to upgrade this limit to 10000 fields.


jjanvier:
FYI, once the limit is reached, we are not able to index any other document.

LeoBenoist:
👍

PR #6844, opened 2017-09-27
Prevents from exporting the unit of an empty metric value

SamirBoulil:
On a product export, the unit of empty metrics would be exported. This throws warning during the import of the exported file.

In this PR, if the metric amount is null, the file exported will not contain the amount nor the unit.

Before:
```
sku;weight;weight-unit
13871461;;KILOGRAM
```

After fix (what we have 1.7):
```
sku;weight;weight-unit
13871461;;
```

`OK`: Done / Validated
`-`: Not needed


features/attribute/browse/view.feature
```diff
@@ -11,7 +11,7 @@ Feature: View attributes
 
   @ce
   Scenario: Successfully view attributes
```
SamirBoulil: Seen with @pierallard

PR #6845, opened 2017-09-27
TIP-613: take ES errors into accound during indexation

jjanvier:
During indexation, we didn't take into account ES errors that can occur.
There are 2 types of errors:
- sometimes an exception is thrown by the official ES PHP client 
- sometimes there is just a field `error` set in the response

Errors can occur for instance when we have reach the `index.mapping.total_fields.limit` parameter of Elasticsearch.

Before this PR, the indexation would silently fail. That means, you think that everything's working, but actually, your products or product models are not indexed (and thus are not available in the datagrid)

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Akeneo/Bundle/ElasticsearchBundle/spec/ClientSpec.php
+++ b/src/Akeneo/Bundle/ElasticsearchBundle/spec/ClientSpec.php
@@ -3,6 +3,7 @@
 namespace spec\Akeneo\Bundle\ElasticsearchBundle;
 
 use Akeneo\Bundle\ElasticsearchBundle\Client;
+use Akeneo\Bundle\ElasticsearchBundle\Exception\IndexationException;
 use Akeneo\Bundle\ElasticsearchBundle\Exception\MissingIdentifierException;
 use Akeneo\Bundle\ElasticsearchBundle\IndexConfiguration\IndexConfiguration;
 use Akeneo\Bundle\ElasticsearchBundle\IndexConfiguration\Loader;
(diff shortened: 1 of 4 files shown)


src/Akeneo/Bundle/ElasticsearchBundle/spec/ClientSpec.php
```diff
@@ -181,6 +209,43 @@ function it_indexes_with_bulk_several_documents($client)
         $this->bulkIndexes('an_index_type', $documents, 'identifier', Refresh::waitFor());
     }
 
+    public function it_triggers_an_exception_during_the_indexation_of_a_several_documents($client)
```
SamirBoulil: it_throws_an_exception_during_the_indexation_of_several_documents

src/Pim/Bundle/InstallerBundle/Command/DatabaseCommand.php
```diff
@@ -127,6 +127,7 @@ protected function resetElasticsearchIndex(OutputInterface $output)
         $output->writeln('<info>Reset elasticsearch indexes</info>');
 
         $this->getContainer()->get('akeneo_elasticsearch.client.product')->resetIndex();
+        $this->getContainer()->get('akeneo_elasticsearch.client.product_model')->resetIndex();
```
SamirBoulil: 👍

nidup: :+1:

ahocquard: to do in fixture loader too no ? (seems not done)

jjanvier: @ahocquard yep, I got a post it for that
but I don't want to struggle more than needed with the CI tonight ^^

jjanvier:
only `navigate_among_variant_entities.feature`* fail

PR #6871, opened 2017-09-29
PIM-6356: display ascendant or descendant image for product models

grena:

Main goal of this PR is to **display image or product models in the grid and in the PEF in a smart way**.
If the attribute as image is on an entity below (product model or variant product), then it returns the first item it finds.

**Bonus fixes from other cards:**
- PIM-6451: Hide the button to edit a family on a variant product edit form
- Change label from "Complete product" to "Variant products"
- PIM-6451: Now display variant axes coming from parent as "Variant Axis" on the product edit form
- PIM-6348: Display a red label in the variant navigation if no product is complete

### DIFF
--- a/src/Pim/Component/Catalog/ProductModel/ProductModelImageAsLabel.php
+++ b/src/Pim/Component/Catalog/ProductModel/ProductModelImageAsLabel.php
@@ -1,0 +1,98 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Pim\Component\Catalog\ProductModel;
+
+use Pim\Component\Catalog\Model\ProductModelInterface;
+use Pim\Component\Catalog\Model\ValueInterface;
+use Pim\Component\Catalog\Repository\ProductModelRepositoryInterface;
+use Pim\Component\Catalog\Repository\ProductRepositoryInterface;
+
+/**
+ * For a given ProductModel, this class retrieves the ValueInterface of its attribute as image,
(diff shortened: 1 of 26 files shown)


src/Pim/Component/Catalog/ProductModel/ProductModelImageAsLabel.php
```diff
@@ -0,0 +1,98 @@
...
+/**
+ * For a given ProductModel, this class retrieves the ValueInterface of its attribute as image,
+ * defined in its family. This value can come from ascendant or descendant entities (from product models above
+ * or variant product below).
+ *
+ * @author    Adrien Pétremann <adrien.petremann@akeneo.com>
+ * @copyright 2017 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php Open Software License (OSL 3.0)
+ */
+class ProductModelImageAsLabel
```
aRn0D: `Pim\Component\Catalog\ProductModel\ProductModelImageAsLabel::getImage()` I guess we can rename this class to `ImageAsLabel` because this class is in the namespace `ProductModel`. So now we have ImageAsLabel::getImage(), it seems obvious that this class will give the image use as a label. I think we should rename the method `getImage`. We should remove the `get` because objects are living organisms, with their own behavior, properties and a life cycle, we don't manipulate data structure like procedural programming. We should rename it to value() because the method returns a `ProductValue`.

PR #6875, opened 2017-09-29
Create a kernel + DI for the integration tests

ahocquard:
This PR is the very beginning to improve our current integration tests stack.

Here is the problems of our stack:
- Currently, fixtures are installed with the same kernel as the kernel to do the test.
So, our kernel has a state before doing anything in the test (yes, mainly UOW...).

- We are doing `extends`to add a behavior. `Composition over inheritance` was not respected.

- We were doing `new` instancation in the `setUp` class (or in the class used by the `setUp`like `FixtureLoader`. It prevents us to use another class in EE, if we want to use a different behavior in EE and CE. 
For example, `ReferenceDataLoader` was used by `FixtureLoader` in CE because it was containing reference to `Fabric`and `Color` from a CE bundle.
As we were using `new`to create this class, it was impossible to adapt it for EE.

- We are using static to get the different paths of the catalogs: impossible to add a path in EE to execute a test of the CE from the EE, the catalog is not installable.

All of these problems are mainly resolved by a kernel specific to the tests, with a new bundle `IntegrationTestsBunle`. IN EE and CE.
This kernel should not be used to execute the test itself, just help to install the fixtures and make the assertion once the test is done.

What has been done in this PR (CE/EE):
- [x] Create the kernel and the bundle
- [x] Use it to load the fixtures: UOW is clean when you start a test (except if you create fixtures on the fly in the test or the `setUp` of the test)
- [x] Use the DI to load reference data in EE and CE (was not done before in EE)
- [x] Delete the tests case in EE that extends CE: use composition with the DI

What's left:
- [ ] Do not inject the kernel/container in classes like `FixturesLoader` (so many dependencies for now, meaning that it should be split).
- [ ] Execute CE tests in EE: find a new way to laod catalog, should be a lot easier with a DI
- [ ] Use the test kernel when creating the fixtures on the fly in a test (not mandatory though, we talk together the pros and cons)

`OK`: Done / Validated
`-`: Not needed


tests/AppKernelTest.php
```diff
@@ -0,0 +1,58 @@
...
+    /**
+     * Registers your custom bundles
+     *
+     * @return array
+     */
+    protected function registerProjectBundles()
+    {
+        return [
+            new Acme\Bundle\AppBundle\AcmeAppBundle(),
+            new Akeneo\Test\IntegrationTestsBundle\AkeneoIntegrationTestsBundle()
```
aRn0D: I am not a fan of creating a bundle for everything.

tests/IntegrationTestsBundle/Security/SystemUserAuthenticator.php
```diff
@@ -0,0 +1,48 @@
...
+use Psr\Container\ContainerInterface;
+use Symfony\Component\DependencyInjection\Container;
+use Symfony\Component\Security\Core\Authentication\Token\UsernamePasswordToken;
+
+/**
+ * @author    Alexandre Hocquard <alexandre.hocquard@akeneo.com>
+ * @copyright 2017 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+class SystemUserAuthenticator
```
aRn0D: Your class is a factory that depends on the container to create a token, that sounds weird.

tests/TestCase.php
```diff
@@ -46,6 +58,16 @@ protected function get($service)
      *
      * @return mixed
      */
+    protected function getFromTestKernel($service)
```
aRn0D: you don't get a service from the kernel but the container ;)

aRn0D: missing type hint

tests/IntegrationTestsBundle/Security/SystemUserAuthenticator.php
```diff
@@ -0,0 +1,48 @@
...
+{
+    /** @var ContainerInterface */
+    protected $container;
```
aRn0D: private

tests/IntegrationTestsBundle/Security/SystemUserAuthenticator.php
aRn0D: I think there is no need to extend this class, it should be final.

tests/IntegrationTestsBundle/Loader/ReferenceDataLoader.php
```diff
@@ -0,0 +1,89 @@
...
+ * Integration test loader for reference data
+ *
+ * @author    Alexandre Hocquard <alexandre.hocquard@akeneo.com>
+ * @copyright 2016 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+class ReferenceDataLoader
+{
+    /** @var EntityManagerInterface */
+    protected $entityManager;
```
aRn0D: private

tests/IntegrationTestsBundle/Loader/ReferenceDataLoader.php
```diff
@@ -0,0 +1,89 @@
...
+use Doctrine\ORM\EntityManagerInterface;
+
+/**
+ * Integration test loader for reference data
+ *
+ * @author    Alexandre Hocquard <alexandre.hocquard@akeneo.com>
+ * @copyright 2016 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+class ReferenceDataLoader
```
aRn0D: Same comment that `SystemUserAuthenticator` about visibility

tests/AppKernelTest.php
```diff
@@ -0,0 +1,58 @@
...
+     */
+    public function getLogDir(): string
+    {
+        return dirname(__DIR__) . DIRECTORY_SEPARATOR . 'var' . DIRECTORY_SEPARATOR . 'logs';
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getName()
```
aRn0D: :string

tests/AppKernelTest.php
```diff
@@ -0,0 +1,58 @@
...
+            . DIRECTORY_SEPARATOR
+            . 'var'
+            . DIRECTORY_SEPARATOR
+            . 'cache'
+            . DIRECTORY_SEPARATOR
+            . 'test_kernel';
+    }
+
+    /**
+     * @return string
```
aRn0D: inheritdoc

tests/AppKernelTest.php
```diff
@@ -0,0 +1,58 @@
...
+    protected function registerProjectBundles()
+    {
+        return [
+            new Acme\Bundle\AppBundle\AcmeAppBundle(),
+            new Akeneo\Test\IntegrationTestsBundle\AkeneoIntegrationTestsBundle()
+        ];
+    }
+
+    /**
+     * @return string
```
aRn0D: inheritdoc

tests/AppKernelTest.php
aRn0D: You create a bundle because you want to create load services with a container extension, right? This is a lot of code lines for nothing. It should be done by:

```php
public function registerContainerConfiguration(LoaderInterface $loader)
{
    parent::registerContainerConfiguration($loader);

    $loader->load('path/to/file.yml');
}
```

composer.json
```diff
@@ -16,9 +16,10 @@
         },
         "psr-4": {
             "Pim\\Upgrade\\": "upgrades/",
+            "Akeneo\\Test\\IntegrationTestsBundle\\": "tests/IntegrationTestsBundle/",
```
aRn0D: Why don't we load `'Akeneo\\Test':"tests"` ?

tests/AppKernelTest.php
```diff
@@ -0,0 +1,58 @@
...
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+class AppKernelTest extends AppKernel
+{
+    /**
+     * Registers your custom bundles
+     *
+     * @return array
+     */
+    protected function registerProjectBundles()
```
aRn0D: type hint

ahocquard: no, part of the signature when you extends something

tests/IntegrationTestsBundle/Security/SystemUserAuthenticator.php
ahocquard: Used by both kernel so it's normal.

tests/AppKernelTest.php
ahocquard: I don't like it.
Impossible to use compiler pass (for example, to create a new catalog).

tests/AppKernelTest.php
ahocquard: same: change signature

tests/IntegrationTestsBundle/Security/SystemUserAuthenticator.php
ahocquard: And it's more than a factory, as it create the user AN athenticates it.

tests/AppKernelTest.php
aRn0D: no you're wrong, you can add compiler pass.

tests/AppKernelTest.php
aRn0D: I disagree the parent return a string, yes Symfony code is not "typehinted" but they follow semver this method will always return a string.

tests/IntegrationTestsBundle/Security/SystemUserAuthenticator.php
aRn0D: yes, that is the problem (you break SRP). You don't understand "Used by both kernel so it's normal"

tests/AppKernelTest.php
ahocquard: I mean in a clean and documented way. I don't want to do what symfony is doing under the hood in my code.

tests/AppKernelTest.php
ahocquard: Do it for an interface and the rule is not the same. So, I don't do it at all when I extends or implements something that is not "typehinted".

tests/IntegrationTestsBundle/Security/SystemUserAuthenticator.php
ahocquard: This code is copied from our code, in a listener: I think it's more important to not break SRP in our codebase than in our test at first. We break 10x SRP in the fixture loader, and this is a far bigger problem.

It's used by the "test kernel" to load the fixtures. And by the other kernel to execute the tests (a lot of tests need a user).

composer.json
ahocquard: Already used by Akeneo\\Test\\Integration\\

tests/AppKernelTest.php
ahocquard: finally done (it's a kind of covariance after all)

PR #6890, opened 2017-10-02
PIM-6797: Automatically set identifier and unique attributes at the variant product level

damien-carcel:
## Description

When creating a family variant, family attributes that are not assigned to an attribute set are automatically put in the common attribute.

However, this should not be the case for identifier attribute and attributes with unique value, which make sens only at the variant product level.

This PR adds a subscriber that automatically set those attributes in the variant attribute set corresponding to the variant product. This is done on two occasions:
- when creating a family variant (automatically handles the non assigned attributes),
- when adding a unique attribute to a family that already have family variants.

## Definition Of Done

`OK`: Done / Validated
`-`: Not needed


src/Pim/Bundle/CatalogBundle/EventSubscriber/AddUniqueAttributesToVariantProductAttributeSet.php
```diff
@@ -0,0 +1,97 @@
...
+            foreach ($familyVariants as $familyVariant) {
+                $this->addUniqueAttributesToFamilyVariant($familyVariant);
+            }
+        }
+    }
+
+    /**
+     * @param FamilyVariantInterface $familyVariant
+     */
+    private function addUniqueAttributesToFamilyVariant(FamilyVariantInterface $familyVariant)
```
jjanvier: I would extract this logic in a service, in the `FamilyVariant` folder

src/Pim/Component/Catalog/FamilyVariant/AddUniqueAttributesToVariantProductAttributeSet.php
```diff
@@ -0,0 +1,60 @@
...
+ * @author    Damien Carcel <damien.carcel@akeneo.com>
+ * @copyright 2017 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+class AddUniqueAttributesToVariantProductAttributeSet
+{
+    /**
+     * @param FamilyVariantInterface $familyVariant
+     */
+    public function addUniqueAttributesToFamilyVariant(FamilyVariantInterface $familyVariant)
```
aRn0D: `AddUniqueAttributesToVariantProductAttributeSet::addUniqueAttributesToFamilyVariant()` the class name and the method have almost the same name. It  could be `AddUniqueAttributes::addToFamilyVariant`

damien-carcel: As CI was full green, I'll fix it in https://github.com/akeneo/pim-community-dev/pull/6905

PR #6902, opened 2017-10-04
PIM-6861: Correctly display a product model if it has no product children

grena:

Correctly display a product model if it has no product children.

### DIFF
--- a/features/product-model/edit_product_model.feature
+++ b/features/product-model/edit_product_model.feature
@@ -48,3 +48,22 @@ Feature: Edit a product model
     And I visit the "Product" group
     Then the field Color (Variant axis) should be read only
     And I should see the text "Color (Variant axis)"
+
+  @jira https://akeneo.atlassian.net/browse/PIM-6861
+  Scenario: Display a product model without any children
+    Given I am logged in as "Mary"
+    When I am on the "1111111113" product page
+    And I press the secondary action "Delete"
+    Then I should see the text "Confirm deletion"
+    And I confirm the removal
+    When I am on the "1111111112" product page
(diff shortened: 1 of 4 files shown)


src/Pim/Bundle/EnrichBundle/Resources/public/js/product/form/variant-navigation.js
```diff
@@ -239,10 +239,11 @@ define([
              * @param entity
              */
             redirectToEntity: function (entity) {
-                const params = {
-                    id: entity.id
-                };
+                if (!entity) {
```
jjanvier: ! or null !== ?

grena: `null` or `undefined` or `''` or `[]`

features/product-model/edit_product_model.feature
```diff
@@ -48,3 +48,21 @@ Feature: Edit a product model
...
+    When I am on the "1111111112" product page
+    And I press the secondary action "Delete"
+    Then I should see the text "Confirm deletion"
+    And I confirm the removal
+    When I am on the "1111111111" product page
+    And I press the secondary action "Delete"
+    Then I should see the text "Confirm deletion"
+    And I confirm the removal
+    When I edit the "amor" product model
+    And I visit the "Product" group
```
SamirBoulil: I don't understand the tests also there is no "Then"

grena: Yup, I'll add a fake check. 
It's just to know if the product page simply show normally instead of having a JS error.

PR #6903, opened 2017-10-04
PIM-6446: fix variant family code uniqueness

jjanvier:
Even if it's a change to a MySQL table, it's better imo to fix that very early in the 2.0 life rather getting some really painful bugs in the future.

### DIFF
--- a/CHANGELOG-2.0.md
+++ b/CHANGELOG-2.0.md
@@ -2,6 +2,7 @@
 
 ## Bug Fixes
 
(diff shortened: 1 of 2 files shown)


CHANGELOG-2.0.md
```diff
@@ -2,6 +2,7 @@
 
 ## Bug Fixes
 
+- [BC BREAK] PIM-6446: fix variant family code uniqueness. Beware, this changes the MySQL table.
```
pierallard: Maybe add this into a specific subcategory ## BC BREAK

PR #6905, opened 2017-10-04
PIM-6847: Fix variant product history

damien-carcel:
### The problem

Currently, there is no version visible in the variant products history panel.

This is because of the way our versioning work: every entity versions are in the same table `pim_versioning_version`. To know to what entity a version belong, there is a field `resource_name` in that table that contains the namespace of the versioned entity.

Problem is, the variant product and regular product share the same interface, but are different entities, with different namespaces. But on the front side, they are both treated as products. So versions are fetched using the "product-history-fetcher", which looks only for the regular product namespace in the `resource_name` field.

### Solving on the front part

A way to solve that would be to create two product forms, with two different routings: one for `product` and one for `variant_product`. JS code would be shared for both except for the history fetcher: there would still be one `product-history-fetcher` for regular products, and a `variant-product-history-fetcher` for variant products.

Problem of this solution is that we create kind of a big BC break in the PEF, introducing new routings, new forms.

### Solution proposed in this PR

The solution is clearly a hack, but allows no BC break at all. The idea is to always use the product namespace, for the regular products and for variant products (whenever the versionable entity implements a `ProductInterface`).

This way, the product versions are still working without any changes, and the new variant product versions are visible too. Only the previous versions (for an already in production system) will remain not visible in PEF.

### Bonus

There is also a renaming of `AddUniqueAttributesToVariantProductAttributeSet::addUniqueAttributesToFamilyVariant` into `AddUniqueAttributes::addToFamilyVariant` (and also the renaming of the corresponding service) as requested by @aRn0D in #6903 (didn't do it in the original PR because the CI was green, sorry for that :trollface: ).

## Definition Of Done

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Pim/Bundle/CatalogBundle/Resolver/FQCNResolver.php
+++ b/src/Pim/Bundle/CatalogBundle/Resolver/FQCNResolver.php
@@ -17,6 +17,9 @@ class FQCNResolver
     /** @var array */
     protected $classNames = [];
 
+    /** @var ContainerInterface */
+    protected $container;
+
     /**
      * @param ContainerInterface $container
      */
@@ -36,8 +39,7 @@ class FQCNResolver
     {
         try {
             $className = $this->container->getParameter(
-                sprintf('pim_catalog.entity.%s.class', $entityType),
-                ContainerInterface::NULL_ON_INVALID_REFERENCE
+                sprintf('pim_catalog.entity.%s.class', $entityType)
             );
         } catch (InvalidArgumentException $e) {
             $className = null;
--- a/src/Pim/Bundle/VersioningBundle/Builder/VersionBuilder.php
+++ b/src/Pim/Bundle/VersioningBundle/Builder/VersionBuilder.php
@@ -5,6 +5,8 @@ namespace Pim\Bundle\VersioningBundle\Builder;
 use Akeneo\Component\Versioning\Model\Version;
 use Doctrine\Common\Util\ClassUtils;
 use Pim\Bundle\VersioningBundle\Factory\VersionFactory;
+use Pim\Component\Catalog\Model\Product;
+use Pim\Component\Catalog\Model\ProductInterface;
 use Symfony\Component\Serializer\Normalizer\NormalizerInterface;
 
 /**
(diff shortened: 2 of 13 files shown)


src/Pim/Bundle/CatalogBundle/Resolver/FQCNResolver.php
```diff
@@ -36,8 +39,7 @@ public function getFQCN($entityType)
     {
         try {
             $className = $this->container->getParameter(
-                sprintf('pim_catalog.entity.%s.class', $entityType),
-                ContainerInterface::NULL_ON_INVALID_REFERENCE
```
damien-carcel: useless

src/Pim/Bundle/VersioningBundle/Builder/VersionBuilder.php
```diff
@@ -176,4 +187,18 @@ function ($item) {
             }
         );
     }
+
+    /**
+     * @param object $versionable
+     *
+     * @return string
+     */
+    protected function getResourceName($versionable)
```
aRn0D: : string

aRn0D: private

src/Pim/Bundle/VersioningBundle/EventSubscriber/AddRemoveVersionSubscriber.php
```diff
@@ -105,4 +107,18 @@ public function addRemoveVersion(RemoveEvent $event)
         $options = $event->getArguments();
         $this->versionSaver->save($version, $options);
     }
+
+    /**
+     * @param object $versionable
+     *
+     * @return string
+     */
+    private function getRessourceName($versionable)
```
aRn0D: idem

src/Pim/Bundle/VersioningBundle/Builder/VersionBuilder.php
damien-carcel: Those are very old stuff, not typehinted at all. I prefer not add typehinting on such stuff in a patch.

PR #6908, opened 2017-10-04
PIM-6862: save products on "pim:completeness:calculate" command

skeleton:
With the rework on the completeness, products are not saved anymore when we call the `generateMissing` method.
Now to be sure ES is synchro with the db, we refresh the product index, get all products without completeness and save products in the db & ES (when we save a product, the completeness is automatically calculated) 

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Pim/Bundle/CatalogBundle/Command/CalculateCompletenessCommand.php
+++ b/src/Pim/Bundle/CatalogBundle/Command/CalculateCompletenessCommand.php
@@ -2,6 +2,7 @@
 
 namespace Pim\Bundle\CatalogBundle\Command;
 
+use Pim\Component\Catalog\Query\Filter\Operators;
 use Symfony\Bundle\FrameworkBundle\Command\ContainerAwareCommand;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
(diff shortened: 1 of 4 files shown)


src/Pim/Bundle/CatalogBundle/Command/CalculateCompletenessCommand.php
```diff
@@ -31,19 +32,34 @@ protected function configure()
...
+        ];
+
+        $container = $this->getContainer();
+        $container->get('akeneo_elasticsearch.client.product')->refreshIndex();
+
+        $pqb = $container->get('pim_catalog.query.product_query_builder_factory')->create($options);
+        $products = $pqb->execute();
+
+        $productsToSave = [];
+        foreach ($products as $product) {
```
phaseinducer: You can use http://php.net/manual/fr/function.array-chunk.php instead of doing that if you want.

tests/CommandLauncher.php
```diff
@@ -0,0 +1,63 @@
...
+use Akeneo\Bundle\BatchBundle\Command\BatchCommand;
+use Symfony\Bundle\FrameworkBundle\Console\Application;
+use Symfony\Component\Console\Input\ArrayInput;
+use Symfony\Component\Console\Output\BufferedOutput;
+use Symfony\Component\HttpKernel\KernelInterface;
+
+/**
+ * Launch commands for the integration tests.
+ */
+class CommandLauncher
```
ahocquard: Should be in  IntegrationTestBundle/launcher

src/Pim/Bundle/CatalogBundle/Command/CalculateCompletenessCommand.php
ahocquard: I suppose it's cursor s I'm not sure.

PR #6912, opened 2017-10-04
PIM-6866: fix PQB sorter when attribute is not localizable and/or scopable

skeleton:
PQB throws an exception when you define a `default_locale` and/or a `default_scope` and when you want to sort on an attribute localizable and/or scopable 

`OK`: Done / Validated
`-`: Not needed


src/Pim/Component/Catalog/Query/ProductQueryBuilder.php
```diff
@@ -278,6 +278,15 @@ protected function addAttributeSorter(
         array $context
     ) {
         $sorter->setQueryBuilder($this->getQueryBuilder());
+
```
jjanvier:
```
$localeCode = $context['locale'];
if (!$attribute->isLocalizable() && !$attribute->isLocaleSpecific()) {
           $localeCode = null;
}
```
no? it's seems weird to me this way

gtm anyway

PR #6913, opened 2017-10-04
PIM-6835: Fix project view in column

tamarasaurus:
This PR fixes the overlapping project details css glitch

Before
capture

After
screen shot 2017-10-04 at 17 44 54

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Pim/Bundle/UIBundle/Resources/public/less/components/Column.less
+++ b/src/Pim/Bundle/UIBundle/Resources/public/less/components/Column.less
@@ -116,7 +116,7 @@
 
   &-part:not(:empty) {
     border-bottom: 1px solid @AknBorderColor;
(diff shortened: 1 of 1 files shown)


src/Pim/Bundle/UIBundle/Resources/public/less/components/Column.less
```diff
@@ -116,13 +116,13 @@
 
   &-part:not(:empty) {
     border-bottom: 1px solid @AknBorderColor;
-    margin-top: -10px;
+    margin-top: 0px;
```
grena: I dunno if we put the unit when 0

tamarasaurus: Fixed 👍

PR #6915, opened 2017-10-04
Apply integration tests from CE in EE

ahocquard:

This PR aims to execute integrations tests from CE in EE.
The catalog from CE was not installable in EE (due to missing files in CE that were mandatory in EE). 
The configuration of the files to load was done in static function in CE.
Therefore, it was not possible to override in EE.

To achieve that, I use the DI for the tests that I implemented in a previous PR: https://github.com/akeneo/pim-community-dev/pull/6875

To inject the Configuration object into the DI of the container, I use synthetic services. https://symfony.com/doc/current/service_container/synthetic_services.html
It allows to inject a service after having booting the kernel.

Before doing any fix, there were 575 fails in EE.
After fix, 140 tests failing test in EE left, due to additional data in EE (permissions, metadata, is_read_only property).
I added a group "ce" for these tests, in order to not be executed in EE.

I've some ideas about the technical strategy to use for these tests to be both executed in CE and EE, but we need to discuss about it together.

**Problems detected**

- API

   - Test on products are failing, due the field `metadata` in EE: not executed in EE

   - No user associated to the the token when using services in the container in EE(resulting in fatal error).
This is due to the function `$client = $this->createAuthenticatedClient();` executed before a test. 

      When creating a client, the `tokenStorage` does not have the system user. In EE, a user has to be authenticated to use some services, for example to create a product before executing a test.

      Solution: use the container created for the tests `$this->getFromTestContainer(..)` to use services (a system user is loaded in the token storage in this container). 

    - Missing a `return` in EE for the ACL on the API: does not work correctly in EE (it fixes 90 tests)

- PQB

A lot of tests don't create properly fixtures in integration tests, because we save the entity without validating it.
So, we create a lot of attributes without the property `group`. Therefore, in EE, the attribute is considered as not existing (even with system user having all roles).

The solution is just to add a group when creating an attribute.
I've fixed a lot of attributes with missing data in the standard format as well. 

And a bug in the category filter fixed in this PR: https://github.com/akeneo/pim-community-dev/pull/6942

- Property "is_read_only"

In EE, on some entities, you have an additional property "is_read_only". So, the data is not the same between EE an CE. I added the group "ce" on these tests to not being executed in EE.

    

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Akeneo/Bundle/BatchQueueBundle/tests/integration/Command/JobQueueConsumerCommandIntegration.php
+++ b/src/Akeneo/Bundle/BatchQueueBundle/tests/integration/Command/JobQueueConsumerCommandIntegration.php
@@ -252,6 +252,6 @@ class JobQueueConsumerCommandIntegration extends TestCase
      */
     protected function getConfiguration()
     {
-        return new Configuration([Configuration::getTechnicalCatalogPath()]);
+        return $this->catalog->useTechnicalCatalog();
     }
 }
--- a/src/Akeneo/Bundle/BatchQueueBundle/tests/integration/Command/PublishJobToQueueCommandIntegration.php
+++ b/src/Akeneo/Bundle/BatchQueueBundle/tests/integration/Command/PublishJobToQueueCommandIntegration.php
@@ -203,8 +203,6 @@ class PublishJobToQueueCommandIntegration extends TestCase
      */
     protected function getConfiguration()
     {
-        return new Configuration(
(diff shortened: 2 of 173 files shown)


solivier:
What is the getRootDirectory file ?

ahocquard:
@solivier good question!I did a lot of rebase, I suppose I did something wrong. Deleted.

PR #6951, opened 2017-10-09
PIM-6872: fix exception when attribute sorted is not in mapping

skeleton:
In ES, the search request will fail if there is no mapping associated with a field

Example:
You want to sort on the attribute `description` with the channel `ecommerce`. 
If in all your data, you have no result for attribute `description` and the channel `ecommerce`, ES will throw:

`{"error":{"root_cause":[{"type":"query_shard_exception","reason":"No mapping found for [values.description-textarea.mobile.en_US.preprocessed] in order to sort on","index_uuid":"xt_aOy_hT5m2Y-rNHJkSUg","index":"akeneo_pim_product_and_product_model"}],"type":"search_phase_execution_exception","reason":"all shards failed","phase":"query","grouped":true,"failed_shards":[{"shard":0,"index":"akeneo_pim_product_and_product_model","node":"OQV8VTpeTHijWclZJW0RKA","reason":{"type":"query_shard_exception","reason":"No mapping found for [values.description-textarea.mobile.en_US.preprocessed] in order to sort on","index_uuid":"xt_aOy_hT5m2Y-rNHJkSUg","index":"akeneo_pim_product_and_product_model"}}]},"status":400}`

To prevent that, I added `unmapped_type` on sorter (see https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-sort.html#_ignoring_unmapped_fields). 
I hardcoded with the value `string` but I don't know what I'm doing ^^
Initially, I tought we were obligate to put the same value defined in https://github.com/akeneo/pim-community-dev/blob/master/src/Pim/Bundle/CatalogBundle/Resources/elasticsearch/index_configuration.yml#L40, but it's not the case because integration tests are OK on boolean sorter, even with the `string` hardcoded.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Pim/Bundle/CatalogBundle/Elasticsearch/Sorter/Attribute/AbstractAttributeSorter.php
+++ b/src/Pim/Bundle/CatalogBundle/Elasticsearch/Sorter/Attribute/AbstractAttributeSorter.php
@@ -42,6 +42,8 @@ abstract class AbstractAttributeSorter implements AttributeSorterInterface
 
     /**
      * {@inheritdoc}
+     *
+     * About "unmapped_type", see https://www.elastic.co/guide/en/elasticsearch/reference/5.x/search-request-sort.html#_ignoring_unmapped_fields
      */
     public function addAttributeSorter(AttributeInterface $attribute, $direction, $locale = null, $channel = null)
     {
@@ -64,6 +66,7 @@ abstract class AbstractAttributeSorter implements AttributeSorterInterface
                     $attributePath => [
                         'order'   => 'ASC',
                         'missing' => '_last',
+                        'unmapped_type' => 'long',
                     ],
                 ];
                 $this->searchQueryBuilder->addSort($sortClause);
@@ -74,6 +77,7 @@ abstract class AbstractAttributeSorter implements AttributeSorterInterface
                     $attributePath => [
                         'order'   => 'DESC',
                         'missing' => '_last',
+                        'unmapped_type' => 'long',
                     ],
                 ];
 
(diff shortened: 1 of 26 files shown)


src/Pim/Bundle/CatalogBundle/Elasticsearch/Sorter/Attribute/AbstractAttributeSorter.php
```diff
@@ -64,6 +64,7 @@ public function addAttributeSorter(AttributeInterface $attribute, $direction, $l
                     $attributePath => [
                         'order'   => 'ASC',
                         'missing' => '_last',
+                        'unmapped_type' => 'string',
```
BitOne: string is deprecated in elasticsearch 5: https://www.elastic.co/guide/en/elasticsearch/reference/current/string.html
`text` or `token` is what you want to use.

BitOne: by the way, why not using long, as in the example in the documentation ?

skeleton: on my first tests, `long` did not work but maybe I had added other stuff which broke the grid.
It's works now :)

src/Pim/Bundle/CatalogBundle/Elasticsearch/Sorter/Attribute/AbstractAttributeSorter.php
```diff
@@ -64,6 +64,7 @@ public function addAttributeSorter(AttributeInterface $attribute, $direction, $l
                     $attributePath => [
                         'order'   => 'ASC',
                         'missing' => '_last',
+                        'unmapped_type' => 'long',
```
jjanvier: could you add a link to the doc please?

PR #6955, opened 2017-10-09
PIM-6893: Improve axis value uniqueness validation

damien-carcel:
### The problem

When creating a new variant product, we validate that there isn't already another product with the same variant axes values: for example, another **black** t-shirt in size **xl**.

There is two validation:
- the first check if we already validated the same axis value combination in a batch (using the `UniqueAxisCombinationSet`): this one works fine
- the second check if there is already a product in database: this one is broken.

### Why

This bug happen because the integration tests were wrong. Two products were created. The first was supposed to be OK (when in fact it wasn't have…), and the second not to pass the validation.

When a variant product is validated, its axis values combination is kept in a `UniqueAxisCombinationSet`, so future variant product can be compared with it during an import. The Set is reset by an event listener triggered by `StorageEvents::POST_SAVE_ALL`, at the end of every batch.

Except that in the integration tests, we used `ProductSaver::save` which does not trigger the reset. So the test was always because the axis value combination was kept in memory when it shouldn't have been.

*This PR fixes the integration tests, not only for variant products but also for product models.*

### Fixing the feature

To know if there already is a variant product with similar values in database, we use `EntityWithFamilyVariant::findSiblings` in the `UniqueVariantAxisValidator`. In the case of the variant products, this call did nothing but always returning an empty array (no siblings, so no duplicated values possible).

We know call `VariantProductRepository::findSiblingsProducts` to find those siblings and ensure there isn't already one with the same axis values.

### Bonus

The `UniqueVariantAxisValidator` were renamed during variant group removal, but only the validator. This PR also renames the associated constraint and service.

## Definition Of Done

`OK`: Done / Validated
`-`: Not needed


src/Pim/Bundle/CatalogBundle/spec/Doctrine/ORM/Repository/EntityWithFamilyVariantRepositorySpec.php
```diff
@@ -0,0 +1,101 @@
...
+    {
+        $this->shouldHaveType(EntityWithFamilyVariantRepository::class);
+    }
+
+    function it_is_an_entity_with_family_variant_repository()
+    {
+        $this->shouldImplement(EntityWithFamilyVariantRepositoryInterface::class);
+    }
+
+    function it_find_no_siblings_if_entity_has_no_family_variant(
```
aRn0D: finds or does not find?

PR #6956, opened 2017-10-10
Fix reference data relation import

LeoBenoist:
Fix:

08:33:42 ERROR     [console] Error thrown while running command "--env=prod akeneo:batch:job -c '{"filePath":"/import\/product_model_clothing_colorsize.csv"}' csv_product_model_import". Message: "Call to undefined method MyProject\Bundle\CatalogBundle\Entity\Supplier::getReference()" ["error" => Error { …},"command" => "--env=prod akeneo:batch:job -c '{"filePath":"/import\/product_model_clothing_colorsize.csv"}' csv_product_model_import","message" => "Call to undefined method MyProject\Bundle\CatalogBundle\Entity\Supplier::getReference()"] []

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Pim/Component/ReferenceData/Model/AbstractReferenceData.php
+++ b/src/Pim/Component/ReferenceData/Model/AbstractReferenceData.php
@@ -63,6 +63,11 @@ abstract class AbstractReferenceData implements ReferenceDataInterface
 
         return $this;
     }
+    
+    public function getReference()
+    {
+        return $this->getId();
+    }
 
     /**
      * {@inheritdoc}


LeoBenoist:
It seem getReference is something necessary in order to have RefData working during import of product model with ref data. You may want to implement an interface

damien-carcel:
Hi Leo. Thank you for the PR, but your branch destination is wrong. Can you change it for the `2.0` please (and of course rebase your branch on it)?

src/Pim/Component/ReferenceData/Model/AbstractReferenceData.php
```diff
@@ -63,6 +63,11 @@ public function setSortOrder($sortOrder)
 
         return $this;
     }
+    
+    public function getReference()
```
solivier: Don't forget the doc, and maybe you should add it into the interface.

damien-carcel: Not maybe imo :-)

damien-carcel:
I will not put a GTM on this PR yet (nor block it), because I think you should not need this method in any cases: we have product imports with reference data in our behat test suites (`features/reference-data/import/import_products.feature`) which are using the entities defined in the AcmeAppBundle (in `src/Acme/Bundle/AppBundle/Entity/Color.php` and `src/Acme/Bundle/AppBundle/Entity/Fabric.php`).

These two entities do not implement the `getReference` method, and yet the imports are working, so I think they are not mandatory.

I also see we use this method for the Assets in EE, but I honnestly don't know why we need it in this case. @jjanvier wrote this entity, so maybe he knows.

PR #6960, opened 2017-10-10
New continuous integration system (2.0)

sebglon:
In this pull request, in addition to the new Jenkinsfile and related files for the new continuous integration system to work, is 2 little notable changes:

### phpunit / phpspec upgrade
In order to list all the behat scenarios we decided to use a well maintained behat extension available for both behat2 and behat3 but this library required a newer version of phpunit and phpspec so we updated it. While doing so some tests failed so we had to rewrite them, basically shouldHaveCount changed its internal implementation and now calls more function on Iterable classes. we used `$this->count()->shouldReturn(xxx);` instead.

### Behat history instability fix
The step `I should see history` was poorly written and worked only with very slow firefox. It tried to look at a row that was just removed by the click on the "expand" button just before.

### Concerning the new files
Well it's a brand new Jenkinsfile, make sure to read it with caution. The Kubernetes job definition is also very important. The behat.*.yml are here to not have to do uggly "sed" in the Jenkinsfile. The enterprise one will be removed when the EE have been migrated on the new CI with the old behat.ci.yml at the root level as it will become useless.

### DIFF
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -1,0 +1,419 @@
+#!groovy
+
+import org.csanchez.jenkins.plugins.kubernetes.pipeline.PodTemplateAction
+
+String[] editions = ["ce"]
+String[] features = ["features"]
+String launchUnitTests = "yes"
+String launchIntegrationTests = "yes"
+String launchBehatTests = "yes"
+
+stage("Build") {
+
+    milestone 1
+    if (env.BRANCH_NAME =~ /^PR-/) {
+        timeout(time:5, unit:'DAYS') {
+            userInput = input(message: 'Launch tests?', parameters: [
+                choice(choices: 'yes\nno', description: 'Run unit tests and code style checks', name: 'launchUnitTests'),
+                choice(choices: 'yes\nno', description: 'Run integration tests', name: 'launchIntegrationTests'),
+                choice(choices: 'yes\nno', description: 'Run behat tests', name: 'launchBehatTests'),
+                string(defaultValue: 'ee,ce', description: 'PIM edition the behat tests should run on (comma separated values)', name: 'editions'),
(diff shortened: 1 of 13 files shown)


composer.json
```diff
@@ -86,10 +86,12 @@
...
         "phpmd/phpmd": "1.*",
-        "phpspec/phpspec": "3.0.*",
-        "phpunit/phpunit": "5.4.*",
+        "phpspec/phpspec": "~3.4.2",
+        "phpunit/phpunit": "~5.7.22",
+        "sebastian/exporter": "~2.0.0",
         "sensiolabs/behat-page-object-extension": "2.1.0",
-        "squizlabs/php_codesniffer": "2.*"
+        "squizlabs/php_codesniffer": "2.*",
+        "liuggio/fastest": "1.4.4"
```
nidup: with these changes, the CI v1 will not work anymore?

.ci/Jenkinsfile
```diff
@@ -0,0 +1,439 @@
...
+
+def withDockerScm(body) {
+    clearTemplateNames()
+    def uuid = UUID.randomUUID().toString()
+    podTemplate(label: "dockerscm-" + uuid, containers: [
+        containerTemplate(name: "docker", image: "paulwoelfel/docker-gcloud", ttyEnabled: true, command: 'cat', resourceRequestCpu: '100m', resourceRequestMemory: '200Mi', envVars: [envVar(key: "DOCKER_API_VERSION", value: "1.23")])
+    ], volumes: [
+        hostPathVolume(hostPath: "/var/run/docker.sock", mountPath: "/var/run/docker.sock")
+    ]) {
+        node("dockerscm") {
```
LudovicTOURMAN: Change line with `node("dockerscm-" + uuid) {` or at least delete the function if not used anymore

composer.json
ClementGautier: Should still work

PR #6967, opened 2017-10-11
PIM-6894: Allow any special characters in password field

skeleton:
Remove constraints `Valid` on the `plainPassword` field. It's useless and cause a bug with special characters when the form is validated

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/features/Pim/Behat/Context/NavigationContext.php
+++ b/features/Pim/Behat/Context/NavigationContext.php
@@ -95,11 +95,12 @@ class NavigationContext extends PimContext implements PageObjectAware
     }
 
     /**
-     * @param string $username
+     * @param string      $username
+     * @param string|null $password
      *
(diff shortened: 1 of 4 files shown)


features/Pim/Behat/Context/NavigationContext.php
```diff
@@ -109,8 +110,9 @@ public function iAmLoggedInAs($username)
             return $this->getSession()->getPage()->find('css', '.AknLogin-title');
         }, 'Cannot open the login page');
 
+        $password = null !== $password ? str_replace(' with password ', '', $password) : $username;
```
phaseinducer: You can use named groups in your regexp if you want to avoid doing that :hankey:

grena: Tricky part! Can't we capture the parenthesis to just have the password? :(

PR #6970, opened 2017-10-11
PIM-6346: Fix EntityWithFamilyVariant versioning

damien-carcel:
## Description

Versioning is broken on sub product models and variant products: the generated versions can contain the data of their parents, if they have been previously updated.

This PR ensure only the data from the saved entity are put in the version change set.

## Definition Of Done

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Pim/Bundle/VersioningBundle/Manager/VersionManager.php
+++ b/src/Pim/Bundle/VersioningBundle/Manager/VersionManager.php
@@ -2,7 +2,6 @@
 
 namespace Pim\Bundle\VersioningBundle\Manager;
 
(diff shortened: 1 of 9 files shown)


src/Pim/Bundle/VersioningBundle/Manager/VersionManager.php
```diff
@@ -51,6 +50,9 @@ class VersionManager
     /** @var VersionContext */
     protected $versionContext;
 
+    /** @var EventDispatcherInterface */
+    protected $eventDispatcher;
```
damien-carcel: Is already in the constructor.

PR #6972, opened 2017-10-11
TIP-813: Make attribute form fields generic

phaseinducer:

Attribute form fields were already quite generic but not at the right place.
Also, a new textarea field is introduced here.

`OK`: Done / Validated
`-`: Not needed


src/Pim/Bundle/EnrichBundle/Resources/public/js/form/common/fields/field.js
```diff
@@ -168,6 +169,15 @@ function (
...
             throw new Error('Please implement the getFieldValue() method in your concrete field class.');
+        },
+
+        /**
+         * Returns a pseudo-unique code used as reference inside templates (as "for" attributes values for example).
+         *
+         * @returns {String}
+         */
+        getFieldId: function () {
+            return Math.random().toString(10).substring(2);
```
pierallard: Not a huge fan of this!!!

phaseinducer: Me neither but I took what we did for the PEF : https://github.com/akeneo/pim-community-dev/blob/2.0/src/Pim/Bundle/EnrichBundle/Resources/public/js/product/field/field.js#L157

pierallard: Maybe remove the substring, you will have less conflicts... The best is to be sure you will never have, NEVER, the same number. Random is not ok for this. Maybe a stateful service... Will see.

PR #6931, opened 2017-10-06
TIP-808: Use version strategy for frontend assets

tamarasaurus:
**Description (for Contributor and Core Developer)**

This PR adds a version strategy for our JS and CSS assets so that we can bust the cache when we release a patch version. 

E.g. 
`<script type="text/javascript" src="/dist/manifest.min.js"></script>` becomes
`<script type="text/javascript" src="/dist/manifest.min.js?2.0.1 "></script>`

**Definition Of Done (for Core Developer only)**

| Q                                 | A
| --------------------------------- | ---
| Added Specs                       | OK
| Added Behats                      | -
| Added integration tests           | -
| Changelog updated                 | OK
| Review and 2 GTM                  | OK
| Micro Demo to the PO (Story only) | -
| Migration script                  | -
| Tech Doc                          | -

`Todo`: Pending / Work in progress
`OK`: Done / Validated
`-`: Not needed

PR #6973, opened 2017-10-12
css should be included by the asset twig helper

BenoitLeveque:
The css file was in 404 when using the symfony dev environment.

using the asset helper fix that.

capture d ecran 2017-10-12 a 11 51 16


BenoitLeveque:
has been fixed by another PR #6931

PR #6976, opened 2017-10-12
Fix filtering in product imports

damien-carcel:
## Description

This PR fixes a EE bug (revealed by a broken integration test).

When importing a product without having the rights to even view this product, the import stop with an error. This is a wrong behavior introduced very recently (it is OK in 2.0.1), as the import should only skip the product with a warning.

The bug was caused by a filter that thrown an exception it didn't previously. This exception was only catch during product retrieval from database (after the data filtering).

The fix consists in catching the two possible exceptions at once in the same `try/catch`.

## Definition Of Done

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Pim/Component/Connector/Processor/Denormalization/ProductProcessor.php
+++ b/src/Pim/Component/Connector/Processor/Denormalization/ProductProcessor.php
@@ -11,7 +11,6 @@ use Akeneo\Component\StorageUtils\Updater\ObjectUpdaterInterface;
 use Pim\Component\Catalog\Builder\ProductBuilderInterface;
 use Pim\Component\Catalog\Comparator\Filter\FilterInterface;
 use Pim\Component\Catalog\Model\ProductInterface;
-use Pim\Component\Catalog\Model\VariantProductInterface;
 use Pim\Component\Catalog\ProductModel\Filter\AttributeFilterInterface;
 use Symfony\Component\Security\Core\Exception\AccessDeniedException;
 use Symfony\Component\Security\Core\Exception\InvalidArgumentException;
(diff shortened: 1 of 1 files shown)


src/Pim/Component/Connector/Processor/Denormalization/ProductProcessor.php
```diff
@@ -101,12 +102,17 @@ public function process($item)
...
-        $item = $this->productAttributeFilter->filter($item);
 
-        $familyCode = $this->getFamilyCode($item);
-        $filteredItem = $this->filterItemData($item);
+        try {
+            $item = $this->productAttributeFilter->filter($item);
 
-        $product = $this->findOrCreateProduct($identifier, $familyCode, $item, $parent);
+            $familyCode = $this->getFamilyCode($item);
+            $filteredItem = $this->filterItemData($item);
```
damien-carcel: The exception can be thrown here

src/Pim/Component/Connector/Processor/Denormalization/ProductProcessor.php
```diff
@@ -101,12 +102,17 @@ public function process($item)
...
+
+            $product = $this->findOrCreateProduct($identifier, $familyCode, $parent);
```
damien-carcel: but also here.

src/Pim/Component/Connector/Processor/Denormalization/ProductProcessor.php
```diff
@@ -13,6 +13,7 @@
 use Pim\Component\Catalog\Model\ProductInterface;
 use Pim\Component\Catalog\Model\VariantProductInterface;
 use Pim\Component\Catalog\ProductModel\Filter\AttributeFilterInterface;
+use PimEnterprise\Component\Security\Exception\ResourceAccessDeniedException;
```
skeleton: nope, it's a EE exception. Use Symfony\Component\Security\Core\Exception\AccessDeniedException please

phaseinducer: Nope, you don't know that as it depends on the implementation of your dependencies. And you're not allowed to add a `use` from CE to EE.

PR #6987, opened 2017-10-13
API-399: Create a product model

solivier:

Add new api endpoint to create a new product model.


src/Pim/Component/Catalog/Updater/ProductModelUpdater.php
```diff
@@ -75,7 +75,7 @@ public function update($productModel, array $data, array $options = [])
             );
         }
 
-        if (isset($data['parent'])) {
+        if (array_key_exists('parent', $data)) {
```
solivier: if parent is explicitly set to null, the isset method will return false and the updater won't set the product model parent to null and no exception will be raised during the validation if it's a sub product model. For the api it's possible to explicitly set the parent model to null, in this case we want the validation to raise an error.

src/Pim/Component/Catalog/Updater/ProductModelUpdater.php
```diff
@@ -126,12 +126,12 @@ private function updateParentAndFamily(ProductModelInterface $productModel, arra
...
      *
      * @param ProductModelInterface $productModel
-     * @param string                $parentCode
+     * @param string|null                $parentCode
      *
      * @throws ImmutablePropertyException
      * @throws InvalidPropertyException
      */
-    private function updateParent(ProductModelInterface $productModel, string $parentCode): void
+    private function updateParent(ProductModelInterface $productModel, ?string $parentCode): void
```
solivier: Same here when parent is set to null.

src/Pim/Bundle/ApiBundle/Controller/ProductModelController.php
```diff
@@ -113,6 +162,32 @@ public function getAction($code): JsonResponse
...
+     *
+     * @return Response
+     */
+    public function createAction(Request $request): Response
+    {
+        $data = $this->getDecodedContent($request->getContent());
+        $data = $this->productModelAttributeFilter->filter($data);
+        $productModel = $this->factory->create();
+
+        if (!isset($data['code'])) {
```
solivier: if the code is missing or set to null, the updater will raise an exception as it waits for a string (because of the strong typing we are using). To prevent this and to have the error from the validation saying we can't have an empty code i have to do this trick.

CHANGELOG-2.0.md
```diff
@@ -1,5 +1,8 @@
 # 2.0.x
 
+## Tech improvements
```
skeleton: "Better manage products with variants!" to be consistent

src/Pim/Bundle/ApiBundle/Controller/ProductModelController.php
```diff
@@ -75,17 +112,29 @@ public function __construct(
...
+        $this->offsetPaginator             = $offsetPaginator;
+        $this->searchAfterPaginator        = $searchAfterPaginator;
+        $this->primaryKeyEncrypter         = $primaryKeyEncrypter;
+        $this->apiConfiguration            = $apiConfiguration;
+        $this->updater                     = $updater;
+        $this->factory                     = $factory;
+        $this->saver                       = $saver;
+        $this->router                      = $router;
+        $this->productValidator            = $productValidator;
+        $this->productModelAttributeFilter = $productModelAttributeFilter;
```
skeleton: we don't align usually

src/Pim/Bundle/ApiBundle/tests/integration/Controller/ProductModel/CreateProductModelIntegration.php
```diff
@@ -0,0 +1,523 @@
...
+            '_links'  => [
+                'documentation' => [
+                    'href' => 'http://api.akeneo.com/api-reference.html#post_product_model'
+                ],
+            ],
+        ];
+
+        $response = $client->getResponse();
+
+        $this->assertSame($expectedContent, json_decode($response->getContent(), true));
```
skeleton: avoid to use `json_decode` when you test your response, test directly the json response with `assertJsonStringEqualsJsonString`

solivier: Yes but expectedContent is not json but and array.

src/Pim/Bundle/ApiBundle/Controller/ProductModelController.php
solivier: :trollface:

CHANGELOG-2.0.md
solivier: Done

PR #6994, opened 2017-10-14
Fix error "index_already_exists_exception" on the CI

ahocquard:
**Goal of this PR**

As you have probably experienced during the last months, the integration tests are not stable anymore.
Currently, two errors exists.

1. Error when resetting index

```
Elasticsearch\Common\Exceptions\BadRequest400Exception: {"error":{"root_cause":[{"type":"index_already_exists_exception","reason":"index [akeneo_pim_product_model/sDejpmt-RACUeyl5e3SCUw] already exists","index_uuid":"sDejpmt-RACUeyl5e3SCUw","index":"akeneo_pim_product_model"}],"type":"index_already_exists_exception","reason":"index [akeneo_pim_product_model/sDejpmt-RACUeyl5e3SCUw] already exists","index_uuid":"sDejpmt-RACUeyl5e3SCUw","index":"akeneo_pim_product_model"},"status":400}
```

2. A new one: error on ACL of the API

```
Pim\Bundle\ApiBundle\tests\integration\Security\AttributeAuthorizationIntegration::testAccessDeniedForListingAttributes
Failed asserting that 200 is identical to 403.
```

This PR aims to resolve the first problem.
The second one would be fixed in another PR later on.
 
**Cause of the error on the creation of the index**

First, we thought that was Elasticsearch that was not reliable when resetting the index with the code:
```
    public function resetIndex()
    {
        if ($this->hasIndex()) {
            $this->deleteIndex();
        }

        $this->createIndex();
    }
```

Guess what? We were guilty. This piece of code works.
Actually, the index was already re-created **between** the delete and the create methods. But it was "random". 
And that's why the create method was throwing an exception "I already exist".

The creation of the index was actually done by processes in background: the jobs calculating the completeness for product models. 

In these processes, if the index does not exist, it creates it again (I don't know if it's ES API itself, ES PHP client, or our code that does it).
And these processes were still existing in background from the previous test when loading the fixtures of the current test. 

The main responsible is the EntityBuilder that was creating product models without waiting the end of the processes. That's why only the tests using this class (without any `sleep`) were failing.

Therefore, the next test was in conflict with theses processes when resetting the index.

Solution: Actually, these jobs should be pushed in the queue. It was not the case because we were using the SimpleJobLauncher. 
Now, it's easier to test: just launch synchronously the daemon until no job to execute left into the queue.

`OK`: Done / Validated
`-`: Not needed


src/Pim/Bundle/CatalogBundle/tests/integration/Doctrine/Common/Saver/SavingProductModelDescendantsIntegration.php
```diff
@@ -27,6 +30,8 @@ protected function setUp()
         $this->esProductAndProductModelClient = $this->get('akeneo_elasticsearch.client.product_and_product_model');
 
         $this->authenticateUserAdmin();
+
+        $this->launcher = new JobLauncher(static::$kernel);
```
SamirBoulil: static::$kernel => $this->testKernel ?

src/Pim/Bundle/CatalogBundle/tests/integration/Doctrine/Common/Saver/SavingProductModelDescendantsIntegration.php
```diff
@@ -48,7 +53,9 @@ public function testIndexingProductModelDescendantsOnUnitarySave()
 
         $this->get('pim_catalog.saver.product_model')->save($rootProductModel);
 
-        sleep(10);
```
SamirBoulil: 👏

tests/IntegrationTestsBundle/Launcher/JobLauncher.php
```diff
@@ -270,6 +270,18 @@ public function waitCompleteJobExecution(JobExecution $jobExecution): void
     }
 
     /**
+     * Indicates whether the queue has a job still not consumed in the queue.
+     *
+     * @return bool
+     */
+    public function hasJobInQueue(): bool
+    {
+        $jobExecutionMessage = $this->kernel->getContainer()->get('akeneo_batch_queue.queue.job_execution_message_repository')->getAvailableJobExecutionMessage();
```
SamirBoulil: line too long right ?

src/Pim/Bundle/CatalogBundle/tests/integration/Classification/ClassifyProductModelIntegration.php
```diff
@@ -118,21 +118,21 @@ public function setUp()
...
         parent::setUp();
 
-        $builder = new EntityBuilder(static::$kernel->getContainer());
+        $builder = new EntityBuilder($this->testKernel->getContainer());
 
         $productModel = $builder->createProductModel('model-tee', 'clothing_color_size', null, []);
-        $this->get('pim_catalog.updater.product_model')->update($productModel, ['categories' => ['supplier_zaro']]);
-        $this->get('pim_catalog.validator.product_model')->validate($productModel);
-        $this->get('pim_catalog.saver.product_model')->save($productModel);
+        $this->getFromTestContainer('pim_catalog.updater.product_model')->update($productModel, ['categories' => ['supplier_zaro']]);
```
SamirBoulil: Simple question: If we'd get the service via `$this->get` we would have the same object right ? Those lines are identical right ?

src/Akeneo/Bundle/BatchQueueBundle/Resources/config/services.yml
```diff
@@ -43,7 +43,7 @@ services:
             - '@akeneo_batch_queue.queue.database_job_execution_queue'
             - '%kernel.environment%'
 
-    akeneo_batch_queue.launcher.simple_job_launcher:
```
SamirBoulil: xD

src/Pim/Bundle/CatalogBundle/tests/integration/Classification/ClassifyProductModelIntegration.php
```diff
@@ -118,21 +118,21 @@ public function setUp()
...
+        $this->getFromTestContainer('pim_catalog.validator.product_model')->validate($productModel);
+        $this->getFromTestContainer('pim_catalog.saver.product_model')->save($productModel);
```
grena: Here you take the service from the test container, AFAIK, it's not mandatory but it's kind of a good practice to take in Integration tests from now, right?

src/Pim/Bundle/CatalogBundle/tests/integration/Doctrine/Common/Saver/SavingProductModelDescendantsIntegration.php
grena: Don't we have to take the launcher from the test container?

jjanvier:
Great news, but it's GTM for me only if everybody (aka the POs) clearly understands the consequences of that. It means, if an import/export/whatever is running, the product models descendance updates will be performed after.

Another solution is to replace the service `akeneo_batch.launcher.simple_job_launcher` by the `akeneo_batch_queue.launcher.queue_job_launcher` ONLY in the integrations tests.
And I'm more and more sure this is the kind of things we should do more in our tests (aka replacing services by others - like mocks, but not only - more suitable for our tests).

src/Pim/Bundle/CatalogBundle/tests/integration/Classification/ClassifyProductModelIntegration.php
ahocquard: Not the same UOW. So, I strongly supposed Doctrine will cry.

src/Pim/Bundle/CatalogBundle/tests/integration/Doctrine/Common/Saver/SavingProductModelDescendantsIntegration.php
ahocquard: We should talk together IRL about it, because it can be complex (to use and to maintain).

For now:
If it's a fixture to load, or an assertion to make with a clean UOW, use the test kernel. This kernel has some more services in the container to help you. Maybe one time, we will have entity builders in it to help you also (and avoid copy/past).

If it's really the container of production you want to test, go with the static::kernel.
Because this is the same container that the production (not extra services for the integration tests).

As you want to test the indexation itself here, I choose to use the `static::$kernel`. If it was just to prepare your tests, I would choosed $this->testKernel.

That's quite a good thing to have two kernel, because when loading the fixtures, you have a different UOW.
But I don't know if we should use the test container fo everything. 

Another drawback of this kernel: an entity created with one UOW is not managed by the other one without `merge`.

And to finish: my idea was originally to not modify the container of production with services to load fixtures. 
But I'm wondering if it's a good idea. Because I think we should mock a lot of things.
So, we should accept that the container that we are testing does not have the same state as the container in production. For example, mocked repositories.

So, for now: use it it, it will not be hard to change, but we must take a decision quickly to avoid a legacy if it's a bad thing.

ahocquard: It was on purpose: I explain why in my response to Samir

src/Pim/Bundle/CatalogBundle/tests/integration/Classification/ClassifyProductModelIntegration.php
ahocquard: Yes, but see my comment to Samir as well. It allows to not pollute your UOW.

ahocquard:
@jjanvier I saw that with Delphine, she is to push completeness jobs into the queue and she is well aware of the pros and cons.

src/Akeneo/Bundle/BatchQueueBundle/Resources/config/services.yml
SamirBoulil: it is the misspell

src/Pim/Bundle/ApiBundle/tests/integration/Controller/Product/AbstractProductTestCase.php
```diff
@@ -96,12 +96,6 @@ protected function createProductModel(array $data = []) : ProductModelInterface
         }
         $this->get('pim_catalog.saver.product_model')->save($productModel);
 
-        $launcher = $this->getFromTestContainer('akeneo_integration_tests.launcher.job_launcher');
```
SamirBoulil: Ok but this means, that if we need to test another behavior which is executed via jobExecutions, the consumer will need to first execute all those product model updates before executing the behavior I want to test right ? (because they are still pushed in the queue even tough they are not executed).

ahocquard: Yes but I think it's normal and we should do this more and more. Here I want to create fixtures, I don't care about completeness of the children in my test.

And btw, tell me if I'm wrong, but it's useless to compute completeness when creating a product model. Am I right?

SamirBoulil: Ask PO's about it, but it make sense for me. OK then I hope we won't have surprises.

PR #6999, opened 2017-10-16
API-404: Update product model via API

solivier:
This PR allow updates of Product model via the API

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Pim/Bundle/ApiBundle/Controller/ProductModelController.php
+++ b/src/Pim/Bundle/ApiBundle/Controller/ProductModelController.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace Pim\Bundle\ApiBundle\Controller;
 
 use Akeneo\Component\StorageUtils\Exception\PropertyException;
+use Akeneo\Component\StorageUtils\Exception\UnknownPropertyException;
 use Akeneo\Component\StorageUtils\Factory\SimpleFactoryInterface;
 use Akeneo\Component\StorageUtils\Saver\SaverInterface;
 use Akeneo\Component\StorageUtils\Updater\ObjectUpdaterInterface;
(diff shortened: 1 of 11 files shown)


src/Pim/Bundle/ApiBundle/Controller/ProductModelController.php
```diff
@@ -135,6 +143,7 @@ public function __construct(
         $this->router                      = $router;
         $this->productValidator            = $productValidator;
         $this->productModelAttributeFilter = $productModelAttributeFilter;
+        $this->productModelRepository = $productModelRepository;
```
willy-ahva: align

solivier: We don't align anymore, if it passes the CI it's ok. This way we do not mess the diff and if shows exactly what I added.

src/Pim/Bundle/ApiBundle/Controller/ProductModelController.php
```diff
@@ -187,6 +197,43 @@ public function createAction(Request $request): Response
...
+        $data = $this->productModelAttributeFilter->filter($data);
+
+        $productModel = $this->productModelRepository->findOneByIdentifier($code);
+        $isCreation = null === $productModel;
+
+        if ($isCreation) {
+            $this->validateCodeConsistency($code, $data);
+            $productModel = $this->factory->create();
+        } else {
+            if (null !== $productModel->getParent() && array_key_exists('parent', $data) && null === $data['parent']) {
```
solivier: In the case someone is trying to update a sub product model with a parent set to null, we should raise an exception here as it won't be raised later because else the product model will be updated with the parent set to null, but the validator won't raise an error as it will think it's a root product model (because it has no parent) and at this point there are no way to know if it was a sub product model or not.

src/Pim/Bundle/ApiBundle/Controller/ProductModelController.php
```diff
@@ -187,6 +197,41 @@ public function createAction(Request $request): Response
...
+        $data = $this->getDecodedContent($request->getContent());
+        $data = $this->productModelAttributeFilter->filter($data);
+
+        $productModel = $this->productModelRepository->findOneByIdentifier($code);
+        $isCreation = null === $productModel;
+
+        if ($isCreation) {
+            $this->validateCodeConsistency($code, $data);
+            $productModel = $this->factory->create();
+        } elseif (null !== $productModel->getParent() && array_key_exists('parent', $data) && null === $data['parent']) {
```
solivier: In the case someone is trying to update a sub product model with a parent set to null, we should raise an exception here as it won't be raised later because else the product model will be updated with the parent set to null, but the validator won't raise an error as it will think it's a root product model (because it has no parent) and at this point there are no way to know if it was a sub product model or not.

ahocquard: You should put it in a dedicated updater of the API.

src/Pim/Bundle/ApiBundle/tests/integration/Controller/ProductModel/PartialUpdateProductModelIntegration.php
```diff
@@ -0,0 +1,1051 @@
...
+            $response->headers->get('location')
+        );
+        $this->assertSame('', $response->getContent());
+
+        $product = $this->get('pim_catalog.repository.product')->findOneByIdentifier('apollon_optionb_false');
+        $standardizedProduct = $this->get('pim_serializer')->normalize($product, 'standard');
+        $this->assertSame($standardizedProduct['values']['a_text'][0]['data'], 'My awesome text');
+    }
+
+    public function testUpdateAxeSubProductModel()
```
ahocquard: `Axis`

You are finally testing the updater. This test should be already covered in the tests on the updater.

PR #7011, opened 2017-10-17
Add use class alias for proper form parent indication

mkilmanas:
**Description**
Fixed `LocaleType` form extension with a `use` alias so that indicates it's parent correctly.


mkilmanas:
Hello? Anyone?
I don't know how it goes for others, but on our project this has recently completely broken container building process and the software is completely unusable without this manual change. (So CI builds are now all failing because they pull vanilla code via composer)
Could be related to php 7.1 in terms of stricter validation, I guess.
Anyway, I think this is a valid fix and would appreciate merging it ASAP

nidup:
Hi @mkilmanas

Thank you for the pull request :tada: 

Could you do it on 2.0 branch? (because master = upcoming 2.1 release)

mkilmanas:
Sure thing, will be there in a few minutes

PR #7018, opened 2017-10-17
Add a "webpack-dev" command

damien-carcel:
## Description

Currently, we have two commands to build our assets with webpack:
- `webpack` which run in production environment, and so minify the code
- `webpack-watch` which build in development environment and then keep running, watching the code and building it again at each modification (which is cool!).

This PR introduces a third command: `webpack-dev`, which does a simple build like `webpack`, but in development environment, so the code will not be minified.

This would be usefull to build the assets when developping with Docker, as there is some stability issues with `webpack-watch`, and provide a development alternative anyway.

## Definition Of Done

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,6 +5,7 @@
   "homepage": "http://www.akeneo.com",
   "scripts": {
     "webpack": "webpack --config webpack.config.js --env=prod",
+    "webpack-dev": "webpack --config webpack.config.js",
     "webpack-watch": "webpack --progress --config webpack.config.js --watch",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -5,6 +5,7 @@
   "homepage": "http://www.akeneo.com",
   "scripts": {
     "webpack": "webpack --config webpack.config.js --env=prod",
+    "webpack-dev": "webpack --config webpack.config.js --env=dev",
```
tamarasaurus: Is the --env=dev necessary? I think it's the default. Anyway GTM

PR #7019, opened 2017-10-17
[PIM-6892] Add locked categories for product models

pierallard:
This PR disable jstree items if they belongs to parent of grandparent product model.
If you find the front crappy, ask me IRL because I didn't found any solution and can list all the tested ones.

| Micro Demo to the PO (Story only) | y

### DIFF
--- a/src/Pim/Bundle/EnrichBundle/Doctrine/ORM/Query/AscendantCategories.php
+++ b/src/Pim/Bundle/EnrichBundle/Doctrine/ORM/Query/AscendantCategories.php
@@ -50,8 +50,7 @@ class AscendantCategories implements AscendantCategoriesInterface
             $result = array_map(function ($id) {
                 return intval($id['id']);
             }, $queryBuilder->getQuery()->getResult());
-        }
-        else if ($entity instanceof VariantProductInterface) {
+        } elseif ($entity instanceof VariantProductInterface) {
             $queryBuilder
                 ->select('category.id AS id, parent_category.id AS parent_id')
                 ->from(VariantProductInterface::class, 'variant_product')
--- a/src/Pim/Bundle/EnrichBundle/Doctrine/ORM/Query/AscendantCategories.php
+++ b/src/Pim/Bundle/EnrichBundle/Doctrine/ORM/Query/AscendantCategories.php
@@ -62,7 +62,7 @@ class AscendantCategories implements AscendantCategoriesInterface
                 ->setParameter(':id', $entity->getId());
 
             foreach ($queryBuilder->getQuery()->getResult() as $resultItem) {
-                if (null !== $resultItem['id'] && !in_array(intval($resultItem['id']), $result)) {
+                if (!in_array(intval($resultItem['id']), $result)) {
                     $result[] = intval($resultItem['id']);
                 }
                 if (null !== $resultItem['parent_id'] && !in_array(intval($resultItem['parent_id']), $result)) {
--- a/src/Pim/Bundle/EnrichBundle/Doctrine/ORM/Query/AscendantCategories.php
+++ b/src/Pim/Bundle/EnrichBundle/Doctrine/ORM/Query/AscendantCategories.php
@@ -1,0 +1,77 @@
+<?php
+declare(strict_types=1);
+
+namespace Pim\Bundle\EnrichBundle\Doctrine\ORM\Query;
+
+use Doctrine\ORM\EntityManagerInterface;
+use Pim\Component\Catalog\Model\CategoryInterface;
+use Pim\Component\Catalog\Model\EntityWithFamilyVariantInterface;
+use Pim\Component\Catalog\Model\ProductModelInterface;
+use Pim\Component\Catalog\Model\VariantProductInterface;
+use Pim\Component\Enrich\Query\AscendantCategoriesInterface;
+
+/**
+ * Query data to get the ascendant categories of entities with family variant
+ *
+ * @author    Pierre Allard <pierre.allard@akeneo.com>
+ * @copyright 2017 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+class AscendantCategories implements AscendantCategoriesInterface
+{
+    /** @var EntityManagerInterface */
+    private $entityManager;
+
+    /**
+     * @param EntityManagerInterface $entityManager
+     */
+    public function __construct(EntityManagerInterface $entityManager)
+    {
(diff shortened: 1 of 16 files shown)


src/Pim/Bundle/EnrichBundle/Doctrine/ORM/Query/AscendantCategories.php
```diff
@@ -0,0 +1,77 @@
...
+                ->innerJoin('product_model.parent', 'parent_product_model')
+                ->innerJoin('parent_product_model.categories', 'category')
+                ->where('product_model.id = :id')
+                ->setParameter(':id', $entity->getId());
+
+            $result = array_map(function ($id) {
+                return intval($id['id']);
+            }, $queryBuilder->getQuery()->getResult());
+        }
+        else if ($entity instanceof VariantProductInterface) {
```
willy-ahva: } else if () {

src/Pim/Bundle/EnrichBundle/Resources/public/js/product/form/categories.js
```diff
@@ -48,6 +51,7 @@ define(
             treeAssociate: null,
             cache: {},
             trees: [],
+            loadedEvent: null,
```
grena: I'd say `onLoadedEventCallback`, as here we can think its the event itself, what do you think?

src/Pim/Bundle/EnrichBundle/Doctrine/ORM/Query/AscendantCategories.php
```diff
@@ -0,0 +1,76 @@
...
+
+            $result = array_map(function ($id) {
+                return intval($id['id']);
+            }, $queryBuilder->getQuery()->getResult());
+        } elseif ($entity instanceof VariantProductInterface) {
+            $queryBuilder
+                ->select('category.id AS id, parent_category.id AS parent_id')
+                ->from(VariantProductInterface::class, 'variant_product')
+                ->innerJoin('variant_product.parent', 'product_model')
+                ->leftJoin('product_model.parent', 'parent_product_model')
```
grena: This `left join` is to prevent the failure of the request is there is only 1 level, right?

src/Pim/Bundle/EnrichBundle/Doctrine/ORM/Query/AscendantCategories.php
```diff
@@ -0,0 +1,76 @@
...
+                ->leftJoin('product_model.categories', 'category')
+                ->leftJoin('parent_product_model.categories', 'parent_category')
+                ->where('variant_product.id = :id')
+                ->setParameter(':id', $entity->getId());
+
+            foreach ($queryBuilder->getQuery()->getResult() as $resultItem) {
+                if (null !== $resultItem['id'] && !in_array(intval($resultItem['id']), $result)) {
```
grena: Not possible to avoid all this stuff with a DISTINCT as the request above?

pierallard: An other solution should be to create a subquery like
SELECT * from categories WHERE id IN ("product model" + "parent product model")
but I don't like subqueries.

src/Pim/Bundle/EnrichBundle/Doctrine/ORM/Query/AscendantCategories.php
pierallard: Exact

src/Pim/Bundle/EnrichBundle/Resources/public/js/product/form/categories.js
pierallard: It's not a callback, but I will find a more relevant name

src/Pim/Bundle/EnrichBundle/Doctrine/ORM/Query/AscendantCategories.php
grena: Ok and why do you check the existence of the id before adding it? and not running a array_unique on it? It would simplify this imho

pierallard: I can remove the existence of id, you're right, but I have to keep the second one because sometimes this column can be null.

src/Pim/Bundle/EnrichBundle/Normalizer/ProductModelNormalizer.php
```diff
@@ -163,6 +170,11 @@ public function normalize($productModel, $format = null, array $context = []): a
                 'variant_navigation'        => $this->navigationNormalizer->normalize($productModel, $format, $context),
             ] + $this->getLabels($productModel);
 
+        // TODO Refactor this condition in 2.1 to remove default null parameter.
```
SamirBoulil: This normalizer was released with the 2.0, you could easily do a BC here. it's acceptable imo.

src/Pim/Bundle/EnrichBundle/Doctrine/ORM/Query/AscendantCategories.php
```diff
@@ -0,0 +1,76 @@
...
+     */
+    public function __construct(EntityManagerInterface $entityManager)
+    {
+        $this->entityManager = $entityManager;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getCategoryIds(EntityWithFamilyVariantInterface $entity): array
```
SamirBoulil: Does it returns his category ids + it's ascendants or only his ascendants ?

grena: Only ascendants yes

src/Pim/Bundle/EnrichBundle/Doctrine/ORM/Query/AscendantCategories.php
```diff
@@ -0,0 +1,76 @@
...
+        if ($entity instanceof ProductModelInterface) {
+            $queryBuilder
+                ->select('DISTINCT(category.id) AS id')
+                ->from(ProductModelInterface::class, 'product_model')
+                ->innerJoin('product_model.parent', 'parent_product_model')
+                ->innerJoin('parent_product_model.categories', 'category')
+                ->where('product_model.id = :id')
+                ->setParameter(':id', $entity->getId());
+
+            $result = array_map(function ($id) {
```
anaelChardan: You could return directly :)

pierallard: I prefer my notation sorry, if you don't have more arguments I will let it as it.

PR #7034, opened 2017-10-19
API-407: Fix too many error messages

solivier:

This fix prevent the empty variant axes to return a wrong error message when we try to create a sub product model that extends another sub product model. Else the validator thinks it's a variant product (as it will be on the 3 level sub_product_model_2 -> sub_product_model_1 -> root_product_model) and will return the axes on the 3 level.

### DIFF
--- a/src/Pim/Bundle/ApiBundle/tests/integration/Controller/ProductModel/CreateProductModelIntegration.php
+++ b/src/Pim/Bundle/ApiBundle/tests/integration/Controller/ProductModel/CreateProductModelIntegration.php
@@ -323,67 +323,66 @@ JSON;
         $this->assertSameProductModels($expectedProductModel, 'root_product_model');
     }
 
-    //TODO: to be fixed in API-407
-//    public function testSubProductModelCreationThatSetsASubProductModelAsParent()
-//    {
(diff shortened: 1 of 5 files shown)


grena:
I don't understand, this use case is impossible :( You should discuss with @Delphineray :)
We have maximum 2 levels of variation, which means:
**Root Product Model** (common attributes) > **Product Model** (axis lvl 1) > **Variant Product** (axis lvl 2)

src/Pim/Bundle/ApiBundle/tests/integration/Controller/ProductModel/CreateProductModelIntegration.php
```diff
@@ -323,67 +323,66 @@ public function testRootProductModelCreationWithParentToNull()
...
-//    }
-//  ]
-//}
-//JSON;
-//
-//        $response = $client->getResponse();
-//
-//        $this->assertJsonStringEqualsJsonString($expectedContent, $response->getContent());
-//        $this->assertSame(Response::HTTP_UNPROCESSABLE_ENTITY, $response->getStatusCode());
-//    }
```
grena: you can drop this comment

PR #7035, opened 2017-10-19
[Bugfix 2.0] Add use class alias for proper form parent indication

mkilmanas:
This is a backport of fix #7011 to 2.0 branch


nidup:
thank you @mkilmanas !

PR #7055, opened 2017-10-23
API-411: Fix missing case when no values are provided

solivier:

When no values are provided via the API it raises a 500 in the ProductModelAttributeFilter as the variantAttribtueSet variable will be null. This PR aims to fix it.

`OK`: Done / Validated
`-`: Not needed


src/Pim/Bundle/ApiBundle/tests/integration/Controller/ProductModel/CreateProductModelIntegration.php
```diff
@@ -488,6 +488,57 @@ public function testSubProductModelCreationWithAParentThatIsNotARootProductModel
         $this->assertSame(Response::HTTP_UNPROCESSABLE_ENTITY, $response->getStatusCode());
     }
 
+    public function testTOTO()
```
solivier: to change :)

damien-carcel:
Imho, you should also add a phpspec test corresponding to the absence of value.

PR #7056, opened 2017-10-24
Fix assets install executed twice

amansilla:
It seems versions from 1.6 to 2.0 are affected by this bug. During assets install the `assets:install` command is going to be executed twice if the `symlink` option is given.

### DIFF
--- a/src/Pim/Bundle/InstallerBundle/Command/AssetsCommand.php
+++ b/src/Pim/Bundle/InstallerBundle/Command/AssetsCommand.php
@@ -68,19 +68,21 @@ class AssetsCommand extends ContainerAwareCommand
             }
         }
 
+        $assetInstallOptions = [];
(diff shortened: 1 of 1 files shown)


damien-carcel:
Hello @amansilla and thank you for this contribution.

I am OK with this PR on principle. But from what I tested on your branch (and already tested when I introduced the `symlink` option), this cannot work at first install.

The `AssetsCommand` run several commands, including `assets:install` and `oro:assetic:dump`. Problem is, `oro:assetic:dump` fails if assets are installed as symlinks, and it is mandatory that they are installed first. Note that this fails only happen at first install, or if you deletes the assets from `web` folder.

I didn't had time to investigate why `oro:assetic:dump` failed with assets as symlink, but this cannot be merged as it is. However, if you can find why it is failing with symlink, I will be more than happy to give my PR approval :-).

amansilla:
Hi @damien-carcel first thanks for your quick feedback.

I understood the problem and hopefully will check why the `oro:assetic:dump` fails when symlinks are used and update this PR with more info.

PR #7064, opened 2017-10-24
TIP-818: Add integration tests for index commands

SamirBoulil:

Munch needed integration tests for the index product command as well as for the index product model command.

My bad :)

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/CHANGELOG-2.0.md
+++ b/CHANGELOG-2.0.md
@@ -29,7 +29,7 @@ IMPORTANT: In order to use the new mass edit, please execute `bin/console akeneo
 
 ## BC breaks
 
-Changes the constructor of `Pim\Bundle\EnrichBundle\Controller\Rest\ProductModelController` to add `Pim\Component\Catalog\Repository\FamilyVariantRepositoryInterface` 
+Changes the constructor of `Pim\Bundle\EnrichBundle\Controller\Rest\ProductModelController` to add `Pim\Component\Catalog\Repository\FamilyVariantRepositoryInterface`
 
 
 # 2.0.7 (2017-11-23)
--- a/src/Akeneo/Bundle/ElasticsearchBundle/tests/integration/AbstractIndexCommandIntegration.php
+++ b/src/Akeneo/Bundle/ElasticsearchBundle/tests/integration/AbstractIndexCommandIntegration.php
@@ -1,0 +1,106 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Akeneo\Bundle\ElasticsearchBundle\tests\integration;
+
+use Akeneo\Test\Integration\TestCase;
+use Akeneo\Test\IntegrationTestsBundle\Launcher\CommandLauncher;
+
+/**
+ * @author    Samir Boulil <samir.boulil@akeneo.com>
+ * @copyright 2017 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+abstract class AbstractIndexCommandIntegration extends TestCase
+{
+    /**
+     * Resets all ES indexes registered in the PIM.
+     */
+    protected function runResetIndexesCommand(): int
+    {
+        $commandLauncher = new CommandLauncher($this->testKernel);
+        $exitCode = $commandLauncher->execute('akeneo:elasticsearch:reset-indexes', null, ['inputs' => ['yes']]);
+
+        return $exitCode;
+    }
+
+    /**
+     * @param int $exitCode
+     */
+    protected function assertExitCodeSuccess(int $exitCode)
+    {
+        $this->assertSame(0, $exitCode);
+    }
+
+    /**
+     * Checks that all registered indexes in the PIM are not empty.
+     */
+    protected function assertAllRegisteredIndexesNotEmpty(): void
+    {
+        $esClients = $this->get('akeneo_elasticsearch.registry.clients')->getClients();
+
+        foreach ($esClients as $esClient) {
+            $allDocuments = $esClient->search('pim_catalog_product', [
+                '_source' => 'identifier',
+                'query'   => [
+                    'match_all' => new \StdClass(),
+                ],
+            ]);
+            $this->assertGreaterThan(0, count($allDocuments['hits']['hits']));
+        }
+    }
+
+    /**
+     * Checks whether all registered indexes in the PIM are empty.
+     */
+    protected function assertIndexesEmpty(): void
+    {
+        $esClients = $this->get('akeneo_elasticsearch.registry.clients')->getClients();
+
+        foreach ($esClients as $esClient) {
+            $allDocuments = $esClient->search(
+                'pim_catalog_product',
+                [
+                    'query' => [
+                        'match_all' => new \StdClass(),
+                    ],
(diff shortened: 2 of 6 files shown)


CHANGELOG-2.0.md
```diff
@@ -6,6 +6,7 @@
 - PIM-6567: Fix attributes filter to not remove axes
 - PIM-6933: Fix menu display in case of acl restriction
 - PIM-6922: fix sort order on attribute groups
+- TIP-818: Add missing integration tests for index commands
```
phaseinducer: No need to add it to the changelog imho

src/Akeneo/Bundle/ElasticsearchBundle/tests/integration/AbstractIndexCommandIntegration.php
```diff
@@ -0,0 +1,101 @@
...
+                'query'   => [
+                    'match_all' => new \StdClass(),
+                ],
+            ]);
+            $this->assertGreaterThan(0, count($allDocuments['hits']['hits']));
+        }
+    }
+
+    /**
+     * Checks wether the given indexes name are empty, if none given it will check that all registered indexes in the
```
phaseinducer: whether (same in all the PR)

src/Akeneo/Bundle/ElasticsearchBundle/tests/integration/AbstractIndexCommandIntegration.php
```diff
@@ -0,0 +1,101 @@
...
+     * PIM are empty.
+     *
+     * @param array $indexesName
+     */
+    protected function assertIndexesEmpty(array $indexesName = []): void
+    {
+        $esClients = $this->get('akeneo_elasticsearch.registry.clients')->getClients();
+
+        foreach ($esClients as $esClient) {
+            if (!in_array($esClient->getIndexName(), $indexesName) && !empty($indexesName)) {
```
phaseinducer: Why this logic is not in the regristry ?

src/Pim/Bundle/CatalogBundle/tests/integration/Command/IndexProductModelsCommandIntegration.php
```diff
@@ -0,0 +1,79 @@
...
+
+    /**
+     * Runs the index product command.
+     */
+    private function runIndexProductModelsCommand(array $productModelIdentifiers = []): void
+    {
+        $options = $this->getIndexCommandOptions($productModelIdentifiers);
+        $commandLauncher = new CommandLauncher($this->testKernel);
+        $exitCode = $commandLauncher->execute('pim:product-model:index', null, $options);
+        $this->assertSame(0, $exitCode);
```
phaseinducer: It feels weird to do an assertion in something that's named run*, maybe just return the exit code and test it outside ?

src/Akeneo/Bundle/ElasticsearchBundle/tests/integration/AbstractIndexCommandIntegration.php
SamirBoulil: Damn It !

src/Akeneo/Bundle/ElasticsearchBundle/tests/integration/AbstractIndexCommandIntegration.php
SamirBoulil: Humm, why ? This if allows me to assert certain things on a list of indexes passed as arguments OR run those assertions on every clients.
What is possible is make the getClients function return a subset of indexes, but I don't like it either especially that it will be only used for tests purposes.
So, that's why it's not in the registry I guess.

src/Pim/Bundle/CatalogBundle/tests/integration/Command/IndexProductModelsCommandIntegration.php
SamirBoulil: I agree! Thanks! I'll return the exitCode and assert on it instead.

CHANGELOG-2.0.md
SamirBoulil: ok

src/Akeneo/Bundle/ElasticsearchBundle/tests/integration/AbstractIndexCommandIntegration.php
```diff
@@ -0,0 +1,101 @@
...
+
+use Akeneo\Test\Integration\TestCase;
+use Akeneo\Test\IntegrationTestsBundle\Launcher\CommandLauncher;
+
+/**
+ * @author    Samir Boulil <samir.boulil@akeneo.com>
+ * @copyright 2017 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+abstract class AbstractIndexCommandIntegration extends TestCase
```
aRn0D: I think you should `AbstractTestCase` now you have a DI and we need to use it. You need to mutualize all these assert function with https://phpunit.de/manual/current/en/extending-phpunit.html

src/Akeneo/Bundle/ElasticsearchBundle/tests/integration/AbstractIndexCommandIntegration.php
```diff
@@ -0,0 +1,101 @@
...
+{
+    /**
+     * Resets all ES indexes registered in the PIM.
+     */
+    protected function runResetIndexesCommand(): void
```
aRn0D: It could be interesting to have several objects which wraps that logic
```php
class ResetIndex
{
    public function execute(array $argument = [])
}

class IndexProduct
{
     public function execute(array $argument = [])
}

class IndexProductModel
{
     public function execute(array $argument = [])
}
```

`CommandLauncher` is define as a service in the pim, we could do the same thing in the test DI. We could define theses classes as a service as well and use them in every test case.

src/Akeneo/Bundle/ElasticsearchBundle/tests/integration/AbstractIndexCommandIntegration.php
SamirBoulil: Seen irl, I'll keep this method without the argument and add dedicated methods to check for specific indexes emptiness

src/Akeneo/Bundle/ElasticsearchBundle/tests/integration/AbstractIndexCommandIntegration.php
SamirBoulil: `I think you should AbstractTestCase now you have a DI` <- I don't understand. Do you have an example in the PIM I can rely on to extend the assertions ?

SamirBoulil:
@aRn0D I'm willing to try to refactor using the ideas you suggested, but I want to merge this version. Then we can refactor. 
So if this versions is good enough for you could you add a GTM please ?

src/Akeneo/Bundle/ElasticsearchBundle/tests/integration/AbstractIndexCommandIntegration.php
SamirBoulil: Seen IRL won't do

PR #7066, opened 2017-10-24
Fix form error's 'message' index

QwertyZW:
The expected message array key index is actually 'code' and not 'message'. The 'message' is not used for any purposes


QwertyZW:
I confused 2 of the controllers, it's good as it is.

PR #7072, opened 2017-10-25
TIP-819 Remove wait refresh for indexing commands

BitOne:

Remove the waiting time between each batch of 100 items when indexing products and products model from the command lines.
The refresh strategy is now at disable, relying on the standard refreshing feature of Elasticsearch without waiting for it to take effect.
More details here:
https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-refresh.html

Reindexing icecat_demo_dev results:
 - before: 28s
 - after: 9s

| Added Specs                       |  OK

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Akeneo/Bundle/ElasticsearchBundle/Refresh.php
+++ b/src/Akeneo/Bundle/ElasticsearchBundle/Refresh.php
@@ -39,7 +39,7 @@ final class Refresh
     /**
      * @return Refresh
      */
-    public static function disabled()
+    public static function disable()
     {
         return new self(Refresh::DISABLE);
     }
--- a/src/Pim/Bundle/CatalogBundle/Elasticsearch/Indexer/ProductIndexer.php
+++ b/src/Pim/Bundle/CatalogBundle/Elasticsearch/Indexer/ProductIndexer.php
@@ -77,6 +77,9 @@ class ProductIndexer implements IndexerInterface, BulkIndexerInterface, RemoverI
     /**
      * Indexes a product in both the product index and the product and product model index.
      *
+     * If the index_refresh is provided, it uses the refresh strategy defined.
+     * Otherwise the waitFor strategy is by default.
+     *
      * {@inheritdoc}
      */
     public function indexAll(array $objects, array $options = []) : void
@@ -84,6 +87,8 @@ class ProductIndexer implements IndexerInterface, BulkIndexerInterface, RemoverI
         if (empty($objects)) {
             return;
         }
+
+        $indexRefresh = $options["index_refresh"] ?? Refresh::waitFor();
 
         $normalizedProducts = [];
         $normalizedProductModels = [];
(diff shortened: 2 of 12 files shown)


src/Pim/Bundle/CatalogBundle/Elasticsearch/Indexer/ProductIndexer.php
```diff
@@ -85,6 +88,8 @@ public function indexAll(array $objects, array $options = []) : void
             return;
         }
 
+        $indexRefresh = $options["index_refresh"] ?? Refresh::waitFor();
```
phaseinducer: simple quotes please (don't we have a cs rule for that ?)

src/Akeneo/Bundle/ElasticsearchBundle/Refresh.php
```diff
@@ -39,7 +39,7 @@ public static function enable()
     /**
      * @return Refresh
      */
-    public static function disabled()
+    public static function disable()
```
SamirBoulil: isn'it a BC ?

jmleroux: It is a BC break. Is it allowed between minor releases?

src/Pim/Bundle/CatalogBundle/Elasticsearch/Indexer/ProductIndexer.php
BitOne: Nop, it seems we don't have a CS for that ;)

src/Akeneo/Bundle/ElasticsearchBundle/Refresh.php
BitOne: Yes, but this class is really shitty named: enable(), waitFor() and disabled()... So I can let the shit here, of be a good boy scout and fix it.

BitOne:
@SamirBoulil : I would like to document the BC break, but there's no CHANGELOG file present for 2.1. Where should I put it ? Thanks !

SamirBoulil:
Here: https://github.com/akeneo/pim-community-dev/blob/2.0/CHANGELOG-2.0.md you might want to merge this PR in the 2.0 branch as well

jmleroux:
I would add the `disable` method and deprecate the `disabled` method to remove it in 3.0

BitOne:
Thanks @jmleroux , will do that.

src/Akeneo/Bundle/ElasticsearchBundle/Refresh.php
aRn0D: What do you mean by "but this class is really shitty named"?

SamirBoulil: I think he ment the "functions" because we have enable/disabled instead of enable/disable

PR #7075, opened 2017-10-25
Add prometheus exporter to the Analytics bundle

ClementGautier:
# Description

Prometheus is a time-series data store. We use it for the cloud for monitoring dashboard and also for clients dashboard (so they can know if they are reaching their quota).

# Definition Of Done

| Added Behats                      | :ok: 
| Changelog updated                 | :ok: 
| Micro Demo to the PO (Story only) | :ok: 

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/app/AppKernel.php
+++ b/app/AppKernel.php
@@ -116,6 +116,7 @@ class AppKernel extends Kernel
             new Doctrine\Bundle\MigrationsBundle\DoctrineMigrationsBundle(),
             new FOS\OAuthServerBundle\FOSOAuthServerBundle(),
             new Oneup\FlysystemBundle\OneupFlysystemBundle(),
+            new TweedeGolf\PrometheusBundle\TweedeGolfPrometheusBundle(),
         ];
     }
 
--- a/composer.json
+++ b/composer.json
@@ -70,7 +70,8 @@
         "symfony/symfony": "3.3.6",
         "symfony/polyfill-apcu": "1.4.0",
         "twig/extensions": "1.2.0",
-        "box/spout": "2.7.2"
+        "box/spout": "2.7.2",
+        "tweedegolf/prometheus-bundle": "^0.2.2"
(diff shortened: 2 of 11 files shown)


app/AppKernel.php
```diff
@@ -116,6 +116,7 @@ protected function getPimDependenciesBundles()
             new Doctrine\Bundle\MigrationsBundle\DoctrineMigrationsBundle(),
             new FOS\OAuthServerBundle\FOSOAuthServerBundle(),
             new Oneup\FlysystemBundle\OneupFlysystemBundle(),
+            new TweedeGolf\PrometheusBundle\TweedeGolfPrometheusBundle(),
```
nidup: not possible in a patch

composer.json
```diff
@@ -70,7 +70,8 @@
         "symfony/symfony": "3.3.6",
         "symfony/polyfill-apcu": "1.4.0",
         "twig/extensions": "1.2.0",
-        "box/spout": "2.7.2"
+        "box/spout": "2.7.2",
+        "tweedegolf/prometheus-bundle": "^0.2.2"
```
nidup: not possible in a patch

app/AppKernel.php
ClementGautier: base changed to master

composer.json
ClementGautier: base changed to master

ClementGautier:
@skeleton base changed to master

app/AppKernel.php
BitOne: I don't think it's a good idea to add this kind of very specific bundle to the standard configuration of the PIM.

composer.json
BitOne: Why not adding this bundle only for the cloud installations ? Shouldn't be in the standard base.

PR #7081, opened 2017-10-27
Add Continuous Integration v2 support

ClementGautier:
CF https://github.com/akeneo/pim-community-dev/pull/6960 for more details

### About the removed Behat
This test was a duplicate of "features/family/display_family_history.feature" and was crashing for some reason because selenium is now slower than firefox (it was already random on the V1) so I decided to drop it.


features/versioning/ensure_versioning_on_family.feature
```diff
@@ -1,22 +0,0 @@
-Feature: Ensure versioning on family
```
sebglon: Why removed?

ClementGautier: read the pull request description

PR #7096, opened 2017-10-31
PIM-6948 Uses search after instead of offset limit

BitOne:
The offset-limit method cannot handle consequent volume of data because it has an exponential response time. The search after method provides a linear response time.

See https://www.eversql.com/faster-pagination-in-mysql-why-order-by-with-limit-and-offset-is-slow/ for some detals.

### DIFF
--- a/CHANGELOG-2.0.md
+++ b/CHANGELOG-2.0.md
@@ -4,11 +4,16 @@
 
 - PIM-6354: Adds product models during quick exports
 
+## Bug fixes
+- PIM-6948: Use search after method for products and product models indexing instead of offset limit
+
 ## BC breaks
 
 - Rename `Pim\Bundle\EnrichBundle\Connector\Job\JobParameters\ConstraintCollectionProvider\ProductQuickExport` to `ProductAndProductModelQuickExport`
 - Rename `Pim\Bundle\EnrichBundle\Connector\Processor\QuickExport\ProductProcessor` to `ProductAndProductModelProcessor`
 - Updates quick export configurations to remove `filePath` and add `filePathProduct` and `filePathProductModel`.
+- Adds `Pim\Component\Catalog\Repository\ProductRepositoryInterface.php::searchAfter()` and `Pim\Component\Catalog\Repository\ProductModelRepositoryInterface::searchAfter()` methods
+- Deletes `Pim\Component\Catalog\Repository\ProductRepositoryInterface.php::findAllWithOffsetAndSize()` and `Pim\Component\Catalog\Repository\ProductModelRepositoryInterface::findRootProductModelsWithOffsetAndSize()` methods.
 
 ## Update jobs
 
--- a/src/Pim/Bundle/CatalogBundle/Doctrine/ORM/Repository/ProductModelRepository.php
+++ b/src/Pim/Bundle/CatalogBundle/Doctrine/ORM/Repository/ProductModelRepository.php
@@ -79,19 +79,6 @@ class ProductModelRepository extends EntityRepository implements ProductModelRep
     /**
      * {@inheritdoc}
      */
-    public function findRootProductModelsWithOffsetAndSize($offset = 0, $size = 100): array
-    {
-        $queryBuilder = $this->createQueryBuilder('pm')
-            ->andWhere('pm.parent IS NULL')
-            ->setFirstResult($offset)
-            ->setMaxResults($size);
-
-        return $queryBuilder->getQuery()->getResult();
-    }
-
-    /**
-     * {@inheritdoc}
-     */
     public function findChildrenProductModels(ProductModelInterface $productModel): array
     {
         $qb = $this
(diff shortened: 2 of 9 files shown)


CHANGELOG-2.0.md
```diff
@@ -4,11 +4,16 @@
...
+## Bug fixes
+- PIM-6948: Use search after method for products and product models indexing instead of offset limit
+
 ## BC breaks
 
 - Rename `Pim\Bundle\EnrichBundle\Connector\Job\JobParameters\ConstraintCollectionProvider\ProductQuickExport` to `ProductAndProductModelQuickExport`
 - Rename `Pim\Bundle\EnrichBundle\Connector\Processor\QuickExport\ProductProcessor` to `ProductAndProductModelProcessor`
 - Updates quick export configurations to remove `filePath` and add `filePathProduct` and `filePathProductModel`.
+- Adds `Pim\Component\Catalog\Repository\ProductRepositoryInterface.php::searchAfter()` and `Pim\Component\Catalog\Repository\ProductModelRepositoryInterface::searchAfter()` methods
+- Deletes `Pim\Component\Catalog\Repository\ProductRepositoryInterface.php::findAllWithOffsetAndSize()` and `Pim\Component\Catalog\Repository\ProductModelRepositoryInterface::findRootProductModelsWithOffsetAndSize()` methods
```
pierallard: Let's discuss with @nidup if these method have to be @deprecated or deleted please

src/Pim/Component/Catalog/Repository/ProductModelRepositoryInterface.php
```diff
@@ -78,4 +70,12 @@ public function findByIdentifiers(array $codes): array;
...
      */
     public function findChildrenProducts(ProductModelInterface $productModel): array;
+
+    /**
+     * Get root products models after the one provided. Mainly used to iterate
+     * through a large collecion.
+     *
+     * The limit parameter defines the number of products to return.
+     */
+    public function searchRootProductModelsAfter(ProductModelInterface $product = null, int $limit): array;
```
pierallard: I think we forbid usage of (parameter with default, parameter without default) (Not sure at all)

CHANGELOG-2.0.md
BitOne: Yes, indeed. And it was deprecated in my first version of the PR. But as I anyway add new method to the interface, things are broken.
So I have the feeling that we shouldn't do half of the job. And by the way these 2 methods have been added for 2.0, so it's not a problem for migration.

src/Pim/Component/Catalog/Repository/ProductModelRepositoryInterface.php
BitOne: It's very useful to have a simple way to use the searchAfter method even for the first time when you don't have yet the product to search after.

ahocquard: `?ProductModelInterface` seems better (as the limit is mandatory, default value is useless).

src/Pim/Bundle/CatalogBundle/Doctrine/ORM/Repository/ProductRepository.php
```diff
@@ -173,4 +161,22 @@ public function getAssociatedProductIds(ProductInterface $product)
...
+     * {@inheritdoc}
+     */
+    public function searchAfter(ProductInterface $product = null, int $limit): array
+    {
+        $qb = $this->createQueryBuilder('p')
+            ->orderBy('p.id', 'ASC')
+            ->setMaxResults($limit);
+        ;
+
+        if ($product != null) {
```
ahocquard: may the force be with you

CHANGELOG-2.0.md
BitOne: So for me, two solutions:
 - switch the limit/offset function to deprecated and use the query builder directly inside the command (so no BC break at all)
 -  adds the new function to the interface and remove the limit/offset function (so full BC break).

But I don't think people already started overriding the ProductRepository, and if they did, they certainly extends the current version. So no real BC break here.

src/Pim/Bundle/CatalogBundle/Doctrine/ORM/Repository/ProductModelRepository.php
```diff
@@ -143,4 +130,23 @@ public function findChildrenProducts(ProductModelInterface $productModel): array
...
+     */
+    public function searchRootProductModelsAfter(ProductModelInterface $productModel = null, int $limit): array
+    {
+        $qb = $this->createQueryBuilder('pm')
+            ->andWhere('pm.parent IS NULL')
+            ->orderBy('pm.id', 'ASC')
+            ->setMaxResults($limit);
+        ;
+
+        if ($productModel != null) {
```
ahocquard: yoda

PR #7099, opened 2017-11-01
TIP-824 Reduce indexing time by around 20%

BitOne:

This is the last obvious batch of perf optimization for products indexing. The next step could be to totally short circuit Doctrine and object hydration by working with arrays, but that would only provides a performance increase about to 2x, maybe 3x top.

This is not bad, but that would the cost in maintainability would be quite high (no objects, very different way of manipulating data compared to the rest of the program, etc...). 

### DIFF
--- a/src/Pim/Bundle/CatalogBundle/Command/IndexProductModelCommand.php
+++ b/src/Pim/Bundle/CatalogBundle/Command/IndexProductModelCommand.php
@@ -5,8 +5,8 @@ declare(strict_types=1);
 namespace Pim\Bundle\CatalogBundle\Command;
 
 use Akeneo\Bundle\ElasticsearchBundle\Refresh;
-use Akeneo\Component\StorageUtils\Detacher\BulkObjectDetacherInterface;
 use Akeneo\Component\StorageUtils\Indexer\BulkIndexerInterface;
+use Doctrine\Common\Persistence\ObjectManager;
 use Pim\Component\Catalog\Repository\ProductModelRepositoryInterface;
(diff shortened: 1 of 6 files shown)


src/Pim/Component/Catalog/Normalizer/Indexing/FamilyNormalizer.php
```diff
@@ -24,6 +24,9 @@ class FamilyNormalizer implements NormalizerInterface
     /** @var LocaleRepositoryInterface */
     protected $localeRepository;
 
+    /** @var Locale[] */
```
ahocquard: `LocaleInterface[]`

src/Pim/Bundle/CatalogBundle/Command/IndexProductModelCommand.php
```diff
@@ -119,7 +119,7 @@ private function indexAll(OutputInterface $output): int
 
             $this->bulkProductModelIndexer->indexAll($rootProductModels, ['index_refresh' => Refresh::disable()]);
             $this->bulkProductModelDescendantsIndexer->indexAll($rootProductModels, ['index_refresh' => Refresh::disable()]);
-            $this->bulkProductModelDetacher->detachAll($rootProductModels);
+            $this->objectManager->clear();
```
ahocquard: Just to know, why detach is not enough?

BitOne: It's enough, just too slow ;)

PR #7101, opened 2017-11-02
Delete cache on acl between integration tests of the API

ahocquard:

Fix the CI failing since 9 days.
A cache on ACL is responsible of this error. This cache is kept between the tests.
When you are loading different catalogs, it can fail, because the ACL are not the same for the users.

This cache stores user and associated role. So, it's a cache on data (probably a problem for horizontal scalability..).

`OK`: Done / Validated
`-`: Not needed


ahocquard:
@phaseinducer cache is not in memory. 
And during API cleint creation, you create a new kernel, loading the cache from the disk.

PR #7104, opened 2017-11-02
PIM-6567: Fix VariantAttributeSet setters (axis, attributes)

SamirBoulil:

    Before:
    Whenever you would set an attribute as an axis within a
    VariantAttributeSet you could also put it as attribute within this same
    VariantAttributeSet.

    The appearance of the axis within this list was implicit and optionnal.

    Problem:
    This was a in regarding the import/export formats regarding the
    variant-attributes column as sometimes the axis would figure in the list
    of attributes and sometimes it won't depending on how it would have been
    saved.

    Fix:
    When setting an axis, we check if it exists in the attributes list if
    it's case we remove it.
    When setting an attribute, we check if it is not in the axes list, if it
    is we do not add it to the attribute list.

`OK`: Done / Validated
`-`: Not needed


SamirBoulil:
@ahocquard fixed it.

PR #7118, opened 2017-11-06
PIM-6891 Count only when object have been loaded, not before

BitOne:

When data have not yet been loaded on the cursor, a count() call executes the ES query and load the objects.

But the foreach calls rewind() on the same cursor, triggering another load of the exact same data, with the exact same query.
By putting the count() after the foreach() the data are loaded in the cursor, and the count() really counts loaded data and don't trigger any data loading.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/CHANGELOG-2.0.md
+++ b/CHANGELOG-2.0.md
@@ -4,6 +4,7 @@
 
 - IM-825: allow concurrent AJAX requests by closing the session in a listener
 - PIM-6838: Display completeness panel after Attributes in the PEF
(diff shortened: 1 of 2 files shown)


BitOne:
@jjanvier : Thanks, CHANGELOG updated. We could test with phpSpec I think, but not easy for the dev to foresee this kind of problem. In fact, this should be hidden inside the cursor itself. But with the Iterator interface, it's pretty difficult to do it without making the code very complex.

PR #7119, opened 2017-11-07
Api 400/401: update a family variant/list of familyvariants with the API

ahocquard:

Add the endpoint "update a family variant" and "update a list of family variants" in the API.

Discussed and confirmed with PO:
- axes are automatically added in the attributes, even if we don't specify axes in attributes when doing the request/importing. Samir'PR was breaking this behavior. I fixed it.

- no more magical stuff in the family variant updater to detect if we are moving an attribute to a sub-level: if we want to move an attribute to a sub-level, you have to delete it in the parent level and add it the sub-level. Meaning that you can't just update one column in the import.

Note: As the standard format is not very well designed for the API, I have to put some validation in the API updater. 

Note: I wait the next PATCH to add it to the changelog.

| Added Behats                      | Done
| Added integration tests           | Done
| Changelog updated                 | Todo !!!

`OK`: Done / Validated
`-`: Not needed


src/Pim/Bundle/ApiBundle/tests/integration/Controller/FamilyVariant/PartialUpdateFamilyVariantIntegration.php
```diff
@@ -0,0 +1,310 @@
...
+            "attributes": ["a_yes_no"]
+        }
+    ],
+    "labels": {
+        "en_US": "English label"
+    }
+}
+JSON;
+
+        $client->request('PATCH', 'api/rest/v1/families/familyA/variants/new_family_variant', [], [], [], $data);
```
SamirBoulil: If it's created shouldn't this be a POST ?

src/Pim/Bundle/ApiBundle/tests/integration/Controller/FamilyVariant/PartialUpdateFamilyVariantIntegration.php
```diff
@@ -0,0 +1,310 @@
...
+        $client = $this->createAuthenticatedClient();
+
+        $data = <<<JSON
+{
+    "code": "familyVariantA1",
+    "variant_attribute_sets": [
+        {
+            "level": 1,
+            "axes": ["a_simple_select"],
+            "attributes": []
```
SamirBoulil: you need to add this: `+                    'attributes' => ['a_metric', 'a_simple_select'],` or you are testing out two things:
- removing a level
- emptying an attribute sets' attributes

src/Pim/Component/Api/spec/Updater/FamilyVariantUpdaterSpec.php
```diff
@@ -58,14 +63,66 @@ function it_updates_a_family_variant($familyUpdater, FamilyVariantInterface $fam
...
+        ];
+        $options = ['familyCode' => 'boots'];
+        $familyVariant->getId()->willReturn(1);
+        $familyUpdater->update(Argument::cetera())->shouldNotBeCalled();
+
+        $this
+            ->shouldThrow(MandatoryPropertyException::class)
+            ->during('update', [$familyVariant, $dataToUpdate, $options]);
+    }
+
```
SamirBoulil: Shouldn't it also throw if the axes field is missing ?

src/Pim/Component/Catalog/Model/VariantAttributeSet.php
```diff
@@ -70,6 +70,12 @@ public function addAttribute(AttributeInterface $attribute): void
     public function setAttributes(array $attributes): void
     {
         $this->attributes = new ArrayCollection($attributes);
+
```
SamirBoulil: 👍

src/Pim/Component/Catalog/Model/VariantAttributeSet.php
```diff
@@ -70,6 +70,12 @@ public function addAttribute(AttributeInterface $attribute): void
...
+        foreach ($this->axes as $axis) {
+            if (!$this->attributes->contains($axis)) {
```
SamirBoulil: Could you not use the `contains` function of the collection and add a private function which compares the attributes via code ?
There is an issue with the doctrine collections as they use an in_array(blablabl, strict=true) which compares object references. Sometimes two attributes that have the same code do not have the same reference and then the `contains` function can't tell they are the same. 
It is a small improvement, but it could be cool if you'd add it to this PR (btw, I can do it if you allow me to push on your branch)

src/Pim/Bundle/ApiBundle/tests/integration/Controller/FamilyVariant/PartialUpdateListFamilyVariantIntegration.php
```diff
@@ -0,0 +1,294 @@
...
+    {"code": "familyVariantA1","labels": {"en_US": "US label"}}
+    {"code": "familyVariantA1","labels": {"fr_FR": "FR label"}}
+JSON;
+
+        $expectedContent = <<<JSON
+{"line":1,"code":"newFamilyVariant","status_code":201}
+{"line":2,"code":"familyVariantA1","status_code":204}
+{"line":3,"code":"familyVariantA1","status_code":204}
+JSON;
+
```
SamirBoulil: extra line

src/Pim/Bundle/ApiBundle/tests/integration/Controller/FamilyVariant/PartialUpdateFamilyVariantIntegration.php
```diff
@@ -0,0 +1,310 @@
...
+        $familyVariant = $this->get('pim_catalog.repository.family_variant')->findOneByIdentifier('familyVariantA1');
+        $familyVariantStandard = [
+            'code'                   => 'familyVariantA1',
+            'labels'                 => [],
+            'family'                 => 'familyA',
+            'variant_attribute_sets' => [
+                [
+                    'level'      => 1,
+                    'axes'       => ['a_simple_select'],
+                    'attributes' => ['a_metric', 'a_simple_select'],
```
SamirBoulil: What happens if this field is not present ?

src/Pim/Bundle/ApiBundle/tests/integration/Controller/FamilyVariant/PartialUpdateFamilyVariantIntegration.php
ahocquard: Nop, our PATCH in the API are upsert.

src/Pim/Component/Api/spec/Updater/FamilyVariantUpdaterSpec.php
ahocquard: No, because in the standard format, it is not mandatory to specify "key" in a JSON object. So, it's not mandatory.
For the level it's different. The standard format is problematic, because the level should be a key => it's indexed by key in the updater.

So, we are force to do it in the updater of the API. Pretty ugly but necessary :/

src/Pim/Bundle/ApiBundle/Controller/FamilyVariantController.php
```diff
@@ -224,7 +231,7 @@ public function createAction(Request $request, string $familyCode): Response
     /**
      * @param Request $request
      * @param string  $familyCode
-     * @param string  $familyVariantCode
```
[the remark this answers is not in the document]
ahocquard: I did it on purpose, we need a convention it to forward requests. Or it adds another configuration:
https://github.com/akeneo/pim-community-dev/blob/2.0/src/Pim/Bundle/ApiBundle/Stream/StreamResourceResponse.php#L109

I prefer convention over configuration for that. And it avoid BC break. 
And it's still boring to handle the case "identifier" for product and "code" for other entities inthe standard. It creates a lot of exception, and then configuration.

src/Pim/Bundle/EnrichBundle/Resources/public/templates/family/tab/variants/add-variant-form-fields-container.html
```diff
@@ -9,10 +9,12 @@
         <% if (fieldErrors.code) { %>
         <div class="AknFieldContainer-footer footer-elements-container error-code">
             <div class="AknFieldContainer-validationErrors validation-errors">
+                <% fieldErrors.code.forEach((error) => { %>
```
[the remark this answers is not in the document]
ahocquard: I changed nothing, I fixed bug ;)

src/Pim/Component/Catalog/Updater/FamilyVariantUpdater.php
```diff
@@ -209,45 +205,6 @@ private function getNumberOfLevel(array $attributeSets): int
     }
 
     /**
-     * We consider that if an attribute added in a given attribute set is also
```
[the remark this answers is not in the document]
ahocquard: I've seen with Delph and she is ok.

src/Pim/Bundle/ApiBundle/tests/integration/Controller/FamilyVariant/PartialUpdateFamilyVariantIntegration.php
ahocquard: Ignored, as it's a key.

src/Pim/Bundle/ApiBundle/tests/integration/Controller/FamilyVariant/PartialUpdateFamilyVariantIntegration.php
SamirBoulil: Then when do you test that the 'POST' call works in your integration tests ?

src/Pim/Bundle/ApiBundle/Controller/FamilyVariantController.php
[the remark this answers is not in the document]
SamirBoulil: :ah: does that mean that if you change the naming here, it breaks the StreamResourceResponse ? if that's the case I'm *really* not a fan of this.

src/Pim/Component/Catalog/Updater/FamilyVariantUpdater.php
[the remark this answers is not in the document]
SamirBoulil: 👍

src/Pim/Bundle/ApiBundle/tests/integration/Controller/FamilyVariant/PartialUpdateFamilyVariantIntegration.php
ahocquard: In another test file `CreateFamilyVariantIntegration`. But it was another card already merged.

src/Pim/Bundle/ApiBundle/tests/integration/Controller/FamilyVariant/PartialUpdateFamilyVariantIntegration.php
ahocquard: As I get an error, it changes nothing. And we test a lot of time two things (or more) in one test, and I think it's acceptable with integration tests for now, as it is slow.

src/Pim/Bundle/ApiBundle/Controller/FamilyVariantController.php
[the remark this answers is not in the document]
ahocquard: I'm really not a fan of adding another complexity in `StreamResourceResponse`.
I will maybe change it in 2.1, but for sure not in a patch because it changes definition of a lot of services.

src/Pim/Bundle/ApiBundle/Resources/config/controllers.yml
```diff
@@ -60,6 +60,7 @@ services:
             - '@pim_api.updater.family_variant'
             - '@pim_catalog.saver.family_variant'
             - '@router'
+            - '@pim_api.stream.family_variant_partial_update_stream'
```
damien-carcel: BC break to add in the changelog?

ahocquard: 👍

PR #7136, opened 2017-11-10
[SLA] PIM-6995: sweets in the object detacher

jjanvier:
We have a customer that imports 200K products with a CSV file in MongoDB. And the step "association import" leaks badly. The import crashes telling all memory has been consumed.

Imports have not changed that much since 1.6. So how is it possible we have a memory leak on the product import? We should have seen it before.
- It happens only with MongoDB
- When there is no associations, the association step is very fast. And even if it leaks, we usually don't have time to observe it.
- The leak increases slowly, which means, unless you don't import many many products (like 50K+), you won't pay attention to it.

The fix here bypasses a Doctrine bug about the detach. It removes unnecessary objects from the private variables of the unit of work. So such kind of dirty tricks? Because in a 1.7 patch, we don't do BC breaks ;)

| Added Specs                       | arghhh, this shit is not specable...
| Review and 2 GTM                  | I'll need 2 people that sell their soul to the devil

### DIFF
--- a/src/Akeneo/Bundle/StorageUtilsBundle/Doctrine/Common/Detacher/ObjectDetacher.php
+++ b/src/Akeneo/Bundle/StorageUtilsBundle/Doctrine/Common/Detacher/ObjectDetacher.php
@@ -46,7 +46,10 @@ class ObjectDetacher implements ObjectDetacherInterface, BulkObjectDetacherInter
         $visited = [];
 
         if ($objectManager instanceof DocumentManager) {
-            $this->doDetach($object, $visited);
+            $this->doDetach($object);
+            if ($object instanceof ProductInterface) {
+                $this->hardcoreDetachForOdmUoW($object);
+            }
         } else {
             $objectManager->detach($object);
             $this->doDetachScheduled($object, $visited);
@@ -165,21 +168,9 @@ class ObjectDetacher implements ObjectDetacherInterface, BulkObjectDetacherInter
      * Do detach objects on DocumentManager
      *
      * @param mixed $document
-     * @param array $visited  Prevent infinite recursion
      */
-    protected function doDetach($document, array &$visited)
+    protected function doDetach($document)
     {
-        $oid = spl_object_hash($document);
-        if (isset($visited[$oid])) {
-            return;
-        }
-
-        $documentManager = $this->getObjectManager($document);
-
-        $visited[$oid] = $document;
-
-        $documentManager->detach($document);
-
         if ($document instanceof ProductInterface) {
             foreach ($document->getValues() as $value) {
                 if (null !== $value->getMedia()) {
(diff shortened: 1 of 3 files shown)


src/Akeneo/Bundle/StorageUtilsBundle/Doctrine/Common/Detacher/ObjectDetacher.php
```diff
@@ -189,4 +190,109 @@ protected function doDetach($document, array &$visited)
...
+
+    /**
+     * There is bug in the Doctrine ODM detach method. Even if it's properly called,
+     * some objects are still not detached and remain in the unit of work.
+     *
+     * This subtle and artistic piece of code aims to remove an object from the private
+     * variables of the unit of work (originalDocumentData, parentAssociations, embeddedDocumentsRegistry and
+     * identityMap).
+     *
+     * This is the worst piece of code I've coded in my whole life. And I know I loose a lot of karma with that...
```
solivier: Is that your autobiography ? :trollface:

jjanvier: to sugar the pill...

fitn:
What about a wall of shame?! :)

src/Akeneo/Bundle/StorageUtilsBundle/Doctrine/Common/Detacher/ObjectDetacher.php
```diff
@@ -47,6 +47,7 @@ public function detach($object)
 
         if ($objectManager instanceof DocumentManager) {
             $this->doDetach($object, $visited);
```
docteurklein: pardon the intrusion, but it seems to me the only reason why ODM's UoW wouldn't unset its internal arrays is because the object isn't [managed](https://github.com/doctrine/mongodb-odm/blob/1.1.x/lib/Doctrine/ODM/MongoDB/UnitOfWork.php#L1997) **or** if we [return too early](https://github.com/akeneo/pim-community-dev/blob/d12997acd6e85e54db0337733796de5bdb5d1217/src/Akeneo/Bundle/StorageUtilsBundle/Doctrine/Common/Detacher/ObjectDetacher.php#L175). 
This early return has the consequence to never call the actual `UoW::detach` method.
Do you think trying to not return early would avoid doing this ourselves ?

jjanvier: @docteurklein you don't have to pardon your intrusion :)

we experience the leak via this class https://github.com/akeneo/pim-community-dev/blob/1.7/src/Pim/Component/Connector/Processor/Denormalization/ProductAssociationProcessor.php#L80
afaik, as we find the product from the repository, it means the product is managed by the UoW, no?

which early return do you talk about?
I'm available IRL, it'll be easier

docteurklein: I was talking about this early return:

https://github.com/akeneo/pim-community-dev/blob/d12997acd6e85e54db0337733796de5bdb5d1217/src/Akeneo/Bundle/StorageUtilsBundle/Doctrine/Common/Detacher/ObjectDetacher.php#L175

jjanvier:
You were right @docteurklein, we did some weird stuff here. Unfortunately, even I removed the weird code we had, it's not enough. We still had a memory leak, even with calling properly the detach() method. Unfortunately, I have no time to dig further. Also, this problem is fixed in 2.0. Which we'll have to "be satisfied with" this dirty patch :(

PR #7155, opened 2017-11-14
PIM-6959: scopable text attributes cannot be used as Label

LaurentPetard:

Since 2.0 if a "scopable" attribute is setted as label in a family, the product label displayed is still the identifier.

The goal of this PR is to fix this bug to have the same behavior that in 1.7. In a future patch the behavior of a "scopable" "attribute as label" will be changed to be identical to the locales.

Technically, the product normalizer of the EnrichBundle gets the label for each locale, but the scope is not used. 

Before 2.0 the scope was used in the method `getValue` in`Pim/Component/Catalog/Model/AbstractProduct.php`, but it was rewritten and there's no longer a default value (the property `scope` of this class)

The scope must be passed to `getValue` from `getLabel`, and to it from the normalizer. We could used again the property `scope`, but it doesn't exists in ProductModel and in the absolute it should not belong to any product entity.

It's why I added a second optional parameter to the method `getLabel` to set the scope, and I added the service `UserContext` to `ProductModelNormalizer` to retrieve the user default scope. To modify the signature of the method `getLabel` in the interfaces `ProductInterface` and `ProductModelInterface` would be a bc-break more annoying so it will be done in the master.

For the datagrid, it's not the user default scope that is used, but the scope of the request context.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Pim/Bundle/EnrichBundle/Normalizer/ProductModelNormalizer.php
+++ b/src/Pim/Bundle/EnrichBundle/Normalizer/ProductModelNormalizer.php
@@ -162,6 +162,8 @@ class ProductModelNormalizer implements NormalizerInterface
         $variantProductCompletenesses = $this->variantProductRatioQuery->findComplete($productModel);
         $closestImage = $this->imageAsLabel->value($productModel);
 
+        $scopeCode = $context['channel'] ?? null;
+
         $normalizedProductModel['meta'] = [
                 'variant_product_completenesses' => $variantProductCompletenesses->values(),
                 'family_variant'            => $normalizedFamilyVariant,
@@ -177,7 +179,7 @@ class ProductModelNormalizer implements NormalizerInterface
                 'ascendant_category_ids'    => $this->ascendantCategoriesQuery->getCategoryIds($productModel),
                 'completenesses'            => $this->incompleteValuesNormalizer->normalize($productModel, $format, $context),
                 'level'                     => $productModel->getVariationLevel(),
-            ] + $this->getLabels($productModel);
+            ] + $this->getLabels($productModel, $scopeCode);
 
         return $normalizedProductModel;
     }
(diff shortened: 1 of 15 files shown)


src/Pim/Bundle/EnrichBundle/Normalizer/ProductModelNormalizer.php
```diff
@@ -105,7 +110,8 @@ public function __construct(
         VariantProductRatioInterface $variantProductRatioQuery,
         ImageAsLabel $imageAsLabel,
         AscendantCategoriesInterface $ascendantCategoriesQuery,
-        NormalizerInterface $incompleteValuesNormalizer
+        NormalizerInterface $incompleteValuesNormalizer,
+        UserContext $userContext = null
```
phaseinducer: I really don't like adding this dependency. The UserContext is a facade for a bunch of crappy stuff and we should try to avoid depending on it when possible. What about using the context like for the datagrid normalizer ? (not perfect neither but better than this imho)

src/Pim/Component/Catalog/spec/Model/ProductModelSpec.php
```diff
@@ -179,6 +219,43 @@ function it_gets_the_code_as_label_if_the_label_value_data_is_empty(
...
+
+        $mobileNameValue->getAttribute()->willReturn($attributeAsLabel);
+        $mobileNameValue->getScope()->willReturn('mobile');
+        $mobileNameValue->getLocale()->willReturn('fr_FR');
+        $mobileNameValue->getData()->willReturn('Petite pelle');
+
+        $ecommerceNameValue->getAttribute()->willReturn($attributeAsLabel);
+        $ecommerceNameValue->getScope()->willReturn('ecommerce');
+        $ecommerceNameValue->getLocale()->willReturn('fr_FR');
+        $ecommerceNameValue->getData()->willReturn('Petit outil agricole authentique');
```
phaseinducer: :heart:

src/Pim/Bundle/EnrichBundle/Normalizer/ProductModelNormalizer.php
LaurentPetard: Because I need the user default scope and its not in the variable `$context` . I wanted to avoid changing it (it's defined in `UserContext::toArray()` . But yes it's maybe a "less worse" solution, especially since it will be redone soon.

src/Pim/Bundle/EnrichBundle/Resources/config/normalizers.yml
```diff
@@ -93,6 +93,7 @@ services:
             - '@pim_catalog.product_models.image_as_label'
             - '@pim_enrich.doctrine.query.ascendant_categories'
             - '@pim_enrich.normalizer.product_model_incomplete_values'
+            - '@pim_user.context.user'
```
phaseinducer: It's not needed anymore.

PR #7157, opened 2017-11-14
POC: transform system test into acceptance test [DO NOT MERGE]

phaseinducer:
Here's an attempt to transform what we can call a system test (end-to-end behat scenario) into an acceptance test (not sure if it's the correct term).

**What's the advantage of doing that ?**
- It's much, much faster : on my local machine, ~17 seconds for the system test, ~300 milliseconds for the acceptance test.
- It should be more stable on the CI.
- The test is focused on the interesting part, we don't test all the stack.

**What it tests**
The test calls our validator service, so it calls Symfony's validator component, and on our side it calls the `DelegatingClassMetadataFactory`, `ProductValueMetadataFactory`, `ChainedAttributeConstraintGuesser`, the specific constraint guessers and the constraint validators.

**What it doesn't test**
The UI is not needed here. We don't care about the controller neither : it mainly normalizes the constraint violations and it can be tested elsewhere.
I also tried to avoid calling the database since it shouldn't be needed when you validate an object. In this case it's only needed here by the ProductBuilder : https://github.com/akeneo/pim-community-dev/blob/2.0/src/Pim/Component/Catalog/Builder/ProductBuilder.php#L86
So I wrote a quick & dirty implementation of an in-memory repository for the attributes. And it works \o/

The fixtures loading and the redefinition of the attribute repository service are also "à l'arrache" (it's a POC remember ?)

**Remarks**
Just writing this test shows several interesting things about the code :
- We can't create a product object without calling the database.
- We have currently no way to validate a `ValueCollection`, it has to be part of a product or a product model.
- It's possible to create an attribute without code, it shouldn't be.

### DIFF
--- a/features/product/validation/validate_text_attributes.feature
+++ b/features/product/validation/validate_text_attributes.feature
@@ -47,19 +47,6 @@ Feature: Validate text attributes of a product
     Then I should see validation tooltip "This value is too long. It should have 8 characters or less."
     And there should be 1 error in the "Other" tab
 
-  Scenario: Validate the email validation rule constraint of text attribute
-    Given I change the Email to "foo"
-    And I save the product
-    Then I should see validation tooltip "This value is not a valid email address."
-    And there should be 1 error in the "Other" tab
-
-  Scenario: Validate the email validation rule constraint of scopable text attribute
-    Given I switch the scope to "ecommerce"
-    And I change the Recipient to "foo"
-    And I save the product
-    Then I should see validation tooltip "This value is not a valid email address."
-    And there should be 1 error in the "Other" tab
-
   Scenario: Validate the url validation rule constraint of text attribute
     Given I change the Link to "bar"
     And I save the product
--- a/src/Pim/Component/Catalog/tests/acceptance/Value/Validation/ValidateEmailValuesIntegration.php
+++ b/src/Pim/Component/Catalog/tests/acceptance/Value/Validation/ValidateEmailValuesIntegration.php
@@ -1,0 +1,68 @@
+<?php
+
+declare(strict_types = 1);
+
+namespace Pim\Component\Catalog\tests\acceptance\Value\Validation;
+
+use Akeneo\Test\Integration\Catalog\InMemoryAttributeRepository;
+use Pim\Component\Catalog\AttributeTypes;
+use Symfony\Bundle\FrameworkBundle\Test\KernelTestCase;
+
+/**
+ * @author    Yohan Blain <yohan.blain@akeneo.com>
+ * @copyright 2017 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php Open Software License (OSL 3.0)
+ */
+class ValidateEmailValuesIntegration extends KernelTestCase
+{
+    /**
+     * {@inheritdoc}
+     */
+    protected function setUp()
+    {
+        static::bootKernel(['debug' => false]);
+
+        $attributeRepository = new InMemoryAttributeRepository();
+
+        static::$kernel->getContainer()->set('pim_catalog.saver.attribute', $attributeRepository);
+        static::$kernel->getContainer()->set('pim_catalog.repository.attribute', $attributeRepository);
+
+        $attributeFactory = static::$kernel->getContainer()->get('pim_catalog.factory.attribute');
+        $identifierAttribute = $attributeFactory->createAttribute(AttributeTypes::IDENTIFIER);
+        $identifierAttribute->setCode('sku');
+        $attributeRepository->save($identifierAttribute);
+    }
+
+    public function testAcceptCorrectEmailTextValue()
+    {
+        $productBuilder = static::$kernel->getContainer()->get('pim_catalog.builder.product');
+        $attributeFactory = static::$kernel->getContainer()->get('pim_catalog.factory.attribute');
+
+        $emailAttribute = $attributeFactory->createAttribute(AttributeTypes::TEXT);
+        $emailAttribute->setCode('email');
+        $emailAttribute->setValidationRule('email');
+
+        $product = $productBuilder->createProduct('new_product');
+        $productBuilder->addOrReplaceValue($product, $emailAttribute, null, null, 'michel@akeneo.com');
+
+        $errors = static::$kernel->getContainer()->get('pim_catalog.validator.product')->validate($product);
+        $this->assertEquals(0, $errors->count());
+    }
+
+    public function testRejectIncorrectEmailTextValue()
+    {
+        $productBuilder = static::$kernel->getContainer()->get('pim_catalog.builder.product');
+        $attributeFactory = static::$kernel->getContainer()->get('pim_catalog.factory.attribute');
+
+        $emailAttribute = $attributeFactory->createAttribute(AttributeTypes::TEXT);
+        $emailAttribute->setCode('email');
+        $emailAttribute->setValidationRule('email');
+
+        $product = $productBuilder->createProduct('new_product');
+        $productBuilder->addOrReplaceValue($product, $emailAttribute, null, null, 'not an email address');
+
+        $errors = static::$kernel->getContainer()->get('pim_catalog.validator.product')->validate($product);
+        $this->assertEquals(1, $errors->count());
+        $this->assertEquals('This value is not a valid email address.', $errors[0]->getMessage());
+    }
+}
--- a/tests/Catalog/InMemoryAttributeRepository.php
+++ b/tests/Catalog/InMemoryAttributeRepository.php
@@ -1,0 +1,158 @@
+<?php
+
+declare(strict_types = 1);
+
+namespace Akeneo\Test\Integration\Catalog;
+
+use Akeneo\Component\StorageUtils\Saver\SaverInterface;
+use Doctrine\ORM\QueryBuilder;
+use Pim\Component\Catalog\AttributeTypes;
+use Pim\Component\Catalog\Model\AttributeGroupInterface;
+use Pim\Component\Catalog\Model\FamilyInterface;
+use Pim\Component\Catalog\Repository\AttributeRepositoryInterface;
+
+/**
+ *
+ *
+ * @author    Yohan Blain <yohan.blain@akeneo.com>
+ * @copyright 2017 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php Open Software License (OSL 3.0)
+ */
+final class InMemoryAttributeRepository implements
+    AttributeRepositoryInterface,
+    SaverInterface
+{
+    /** @var \ArrayObject */
+    private $attributes;
+
+    public function __construct()
+    {
+        $this->attributes = new \ArrayObject();
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function save($attribute, array $options = [])
+    {
+        $this->attributes->append($attribute);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getIdentifier()
+    {
+        foreach($this->attributes as $attribute) {
+            if (AttributeTypes::IDENTIFIER === $attribute->getType()) {
+                return $attribute;
+            }
+        }
+
+        return null;
+    }
+
+    /*********************************
+     * NOT IMPLEMENTED YET
+     ********************************/
+
+    public function findAllInDefaultGroup()
+    {
+        throw new \RuntimeException(__METHOD__.' not implemented.');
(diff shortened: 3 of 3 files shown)


tests/Catalog/InMemoryAttributeRepository.php
```diff
@@ -0,0 +1,158 @@
...
+use Pim\Component\Catalog\Repository\AttributeRepositoryInterface;
+
+/**
+ *
+ *
+ * @author    Yohan Blain <yohan.blain@akeneo.com>
+ * @copyright 2017 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php Open Software License (OSL 3.0)
+ */
+final class InMemoryAttributeRepository implements
```
aRn0D: I think our in-memory repository should only implement `SaverInterface` and `IdentifiableObjectRepositoryInterface`

tests/Catalog/InMemoryAttributeRepository.php
```diff
@@ -0,0 +1,158 @@
...
+    {
+        $this->attributes = new \ArrayObject();
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function save($attribute, array $options = [])
+    {
+        $this->attributes->append($attribute);
```
aRn0D: You need to check if the object is already in the collection because a repository is a collection of unique object.

tests/Catalog/InMemoryAttributeRepository.php
```diff
@@ -0,0 +1,158 @@
...
+        }
+
+        return null;
+    }
+
+    /*********************************
+     * NOT IMPLEMENTED YET
+     ********************************/
+
+    public function findAllInDefaultGroup()
```
aRn0D: I am not sure that we need those methods but we should add an integration test for each of them.

tests/Catalog/InMemoryAttributeRepository.php
```diff
@@ -0,0 +1,158 @@
+<?php
+
+declare(strict_types = 1);
+
+namespace Akeneo\Test\Integration\Catalog;
```
aRn0D: In the perfect world without breaking everything, it should go in `Pim\Bundle\CatalogBundle\Persistence\InMemory`. `Persistence` will be a port and `InMemory` will an adapter. At the same level, you could have `Doctrine` or `ORM` adapter too.

src/Pim/Component/Catalog/tests/acceptance/Value/Validation/ValidateEmailValuesIntegration.php
```diff
@@ -0,0 +1,68 @@
...
+use Akeneo\Test\Integration\Catalog\InMemoryAttributeRepository;
+use Pim\Component\Catalog\AttributeTypes;
+use Symfony\Bundle\FrameworkBundle\Test\KernelTestCase;
+
+/**
+ * @author    Yohan Blain <yohan.blain@akeneo.com>
+ * @copyright 2017 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php Open Software License (OSL 3.0)
+ */
+class ValidateEmailValuesIntegration extends KernelTestCase
```
aRn0D: Weird to have integration for an acceptance test :)

tests/Catalog/InMemoryAttributeRepository.php
aRn0D: Where do you add that repository in the DI? Do you plan to create a dedicated environment?

features/product/validation/validate_text_attributes.feature
```diff
@@ -47,19 +47,6 @@ Feature: Validate text attributes of a product
     Then I should see validation tooltip "This value is too long. It should have 8 characters or less."
     And there should be 1 error in the "Other" tab
 
-  Scenario: Validate the email validation rule constraint of text attribute
```
aRn0D: I think it a mistake to remove those Gherkins. We should rewrite them 
```
Given the following product "toto"
When I change its email to "titi"
Then I should get an error "titi is not a valid email"
```
This is readable, understandable by everyone (especially by newcomers) but we need to run them without @javascript

Don't forget that Behat is more extensible than PHPUnit.

src/Pim/Component/Catalog/tests/acceptance/Value/Validation/ValidateEmailValuesIntegration.php
```diff
@@ -0,0 +1,68 @@
...
+{
+    /**
+     * {@inheritdoc}
+     */
+    protected function setUp()
+    {
+        static::bootKernel(['debug' => false]);
+
+        $attributeRepository = new InMemoryAttributeRepository();
```
aRn0D: ah ok you instantiate it here (see my need comment)

src/Pim/Component/Catalog/tests/acceptance/Value/Validation/ValidateEmailValuesIntegration.php
```diff
@@ -0,0 +1,68 @@
...
+
+        static::$kernel->getContainer()->set('pim_catalog.saver.attribute', $attributeRepository);
+        static::$kernel->getContainer()->set('pim_catalog.repository.attribute', $attributeRepository);
```
aRn0D: Why don't you use your in-memory repo?

src/Pim/Component/Catalog/tests/acceptance/Value/Validation/ValidateEmailValuesIntegration.php
```diff
@@ -0,0 +1,68 @@
...
+
+        $attributeFactory = static::$kernel->getContainer()->get('pim_catalog.factory.attribute');
+        $identifierAttribute = $attributeFactory->createAttribute(AttributeTypes::IDENTIFIER);
+        $identifierAttribute->setCode('sku');
+        $attributeRepository->save($identifierAttribute);
+    }
+
+    public function testAcceptCorrectEmailTextValue()
```
aRn0D: You should not camel case to describe our use case because it is not readable.

src/Pim/Component/Catalog/tests/acceptance/Value/Validation/ValidateEmailValuesIntegration.php
phaseinducer: It was because it's the suffix rule we chose in the phpunit config, but I'll remove that anyway if we use behat.

src/Pim/Component/Catalog/tests/acceptance/Value/Validation/ValidateEmailValuesIntegration.php
phaseinducer: That's what I do. I don't understand.

tests/Catalog/InMemoryAttributeRepository.php
phaseinducer: So you think that the code used only by the tests should be mixed with the rest of the code ?

tests/Catalog/InMemoryAttributeRepository.php
phaseinducer: Why integration ? More unit test, no ?

jjanvier: integration
because you don't to test that the database will effectively return what you expect

phaseinducer: But we're talking about an in-memory repo here.

jjanvier:
Your POC is great, it clearly demonstrates some lacks/architectures "glitches" we currently have (your "remarks" section especially).

I didn't follow Mathias worshop, so maybe what I'm saying in the following is nonsense.

AFAIK, acceptance tests are meant to ensure business rules are correctly coded. In her stories, Delph writes the acceptance criterias she expects. And she defined by herself this section "acceptance criterias". They are one of the keypoint between the pos/devs communication. Therefore, IMO, acceptance tests should use Gherkin.

I really like the way you replaced the services `pim_catalog.repository.attribute` and `pim_catalog.saver.attribute` directly in your test. It's very pragmatic. It allows you to fake/mock/stub (I don't know which term) without breaking the whole stack. Even if it's surely not ideal, and what to target, this is IMO a very good way to start and to decouple our acceptance tests from the infrastructure. 

## Small other remarks, not necessarily linked to your POC

About 
> We have currently no way to validate a ValueCollection, it has to be part of a product or a product model.
This is something we'll have to take care in a real future. We'll need that for mass edit attributes of a product model.

I don't like the term system. It's imo too vague. In system, I'm tempted to put performance/load/end-to-end etc... I have also seen the term "smoke testing", but I'm not sure this is what we want. 

I can't stand anymore the "features" folder we have (even if I really understand why Behat does that by default). I'd vote for something like
```
   src/
   test/
      end-to-end/
      acceptance/
      integration/
      specification/
```

jjanvier:
And by and large, I agree with @aRn0D's remarks.

tests/Catalog/InMemoryAttributeRepository.php
jjanvier: this is something I am personally still not sure about :/

aRn0D: I am for but perhaps I am wrong :)

tests/Catalog/InMemoryAttributeRepository.php
jjanvier: we should test the method of the real attribute repo by an integration test
that's what Arnaud meant ;)

aRn0D: I spoke about integration test for a repository, I spoke about testing Doctrine repository with integration test but I think that those methods are useless for an in-memory repository.

aRn0D:
IMHO, you should share feature between system/end-to-end tests and acceptance because the business rules are the same. You tag all scenarios and write difference context depending on the "layer" you want to test.

aRn0D:
I agree with @jjanvier for test

tests/Catalog/InMemoryAttributeRepository.php
phaseinducer: Ok !
And about the methods, I don't like the idea of implementing them in the in-memory repo, but if what you want to test needs them you'll have no choice.

phaseinducer:
@aRn0D It would be nice. Where would you put the common Gherkin then ?

tests/Catalog/InMemoryAttributeRepository.php
phaseinducer: For the moment I can leave it outside of src/ and we can move it later if the team agrees on that.

jjanvier:
Except in end-to-end to test, I don't want to test the business rules anymore. They already work as they have been tested by the acceptance tests.
In end-to-end test, I want to check that the critical features of the app are usable. For instance I want to go on a product, fill in values, click on the "Save" button and see there was no error. 

And I really don't understand why and even how the Gherkin should be written the same. This is IMO more confusing than helping. Maybe @aRn0D you should open a new POC for that precise topic? So that we can it discuss it somewhere and not pollute this one? As there is already a lot to say here :)

tests/Catalog/InMemoryAttributeRepository.php
skeleton: Personally, I would prefer to keep dedicated classes for tests outside of src

skeleton:
:+1: for a POC about Gherkin and a share feature between system & acceptance tests. As jj, I really don't see how can it work :(

Agree too for the structure of folder proposed by jj (ok too for `end-to-end` instead of `system` even if `system` acceptable for me too)

Big :-1: to have dedicated classes for tests directly in our src folder. 
- some integrators will be maybe attempt to use them for business code 
- it will be complicated to see what classes have been faked/mocked
- I personally prefer to have a clear separation between what is needed for tests

ahocquard:
One question about en-to-end tests:
- should we test controller used by the UI with end-to-end tests as well? Or do we consider end-to-end tests as UI +back in these cases ?

tests/Catalog/InMemoryAttributeRepository.php
aRn0D: yes but for now, we don't need them so I guess we will implement them when we need them :).

docteurklein:
Maybe the "why and even how the Gherkin should be written the same" is explained by http://stakeholderwhisperer.com/posts/2014/10/introducing-modelling-by-example ? I played with this idea on a few projects now for more than 2 years and I can see the benefits but also the limitations:
There were a lot of duplication in the unit and acceptance tests (same Arrangements, same Acts, same Assertions).

jjanvier:
Thanks for the article. It's really interesting :)

PR #7198, opened 2017-11-23
PIM-6958: ValueCollectionFactory loads values gracefully.

SamirBoulil:

Prior to this PR, whenever an attribute option was removed from the
PIM (or a file, or a reference data value was removed), the PIM would
load the product values by completely ignoring the product value as it
was not able to properly load the data for the corresponding attribute
(normal because it has been removed).

In this case, some of the factories implementing
Pim\Component\Catalog\Factory\Value\ValueFactoryInterface would throw an
exception stating they weren't able to load the product value's data
resulting in the ingoring of the whole product value for this attribute.

Let's say My product "P1" has an product value linked to the attribute
"multi_select" with data ["valueA", "valueB"].
Now a user removes the value "valueB" from the PIM.
If a user whould load the product edit form of "P1", the "multi_select"
attribute would not show up at all.

In this PR, we changed a bit the inner workings of those factories so
that they don't throw exceptions but instead try to load what they can.

So in the previous example, when the user would load the product edit
form of "P1", it would show an attribute "multi_select" with value
["valueA"].

For other attribute types such as Media or files, if the file has been
removed from the PIM, the product edit form will still show the media
attribute but without any data (ie, creates an empty product value).

`OK`: Done / Validated
`-`: Not needed


SamirBoulil:
This version broke too many import behavior as the factories are used:
1. to load products from DB to render in PEF
2. to load products from DB to update in import export.

SamirBoulil:
https://github.com/akeneo/pim-community-dev/pull/7214

PR #7209, opened 2017-11-24
Transform end-to-end tests into acceptance tests using Behat with in-memory repos

phaseinducer:
This POC is a follow-up to https://github.com/akeneo/pim-community-dev/pull/7157.

The idea is the same : find a way to transform end-to-end tests into other more relevant and efficient types of tests.
The first POC was very naive and pragmatic, this time I wanted to go further and try to use Behat with Symfony integration to be able to use the DI.

### About the tests themselves
I chose validation scenarii as a simple (but not so much at the end) work base.
The old tests were end-to-end, they used :
- UI + controllers + normalizers
- business code
- database

The goal was to find a way to focus on the business logic.
First step : rework the Gherkin to remove all notions of UI components and call directly business services from the new Behat context. The UI behavior can still be tested by other end-to-end tests.

About how I reworded the tests : I discussed a lot with @jjanvier and @Delphineray about the fact that it should be possible to validate a value collection without using a product (technically we're really close to be able to do it). Indeed, a value collection could be seen as a standalone entity and can be contained by something else, a product model for instance. So it could make sense even from a functional point of view.
Anyway we didn't find any good solution, it looks like it's still not a well defined concept, or at least we couldn't find a common (technical/functional) way to talk about it. 
So for the moment i chose to mention products in the Gherkin, it can change later.

### Technical solution
Then, to avoid calling the database I used an in-memory repository when needed (for these use cases only the attribute repository was called).
To do that, I needed :
- A **new Symfony environment**, as light as possible, mainly used to replace ORM repos by in-memory repos in the DI.
- A **new Behat profile** with a new "acceptance" suite, using this new environment (unfortunately for the moment it's not possible to set a different environment per suite, see https://github.com/Behat/Symfony2Extension/issues/131)

I also created a brand new Behat context. Still not sure about the naming, the files organisation, the steps design, the stateful behavior, how to split it later... Feel free to propose ideas and vision.

### Possible next steps
- Continue to transform other scenarii and use it to improve the new context(s).
- Find how to launch it on the CI.
- Validate it with the POs.

### Try it
Launch `bin/behat --profile=acceptance --suite=acceptance` on your machine.

### DIFF
--- a/app/config/config_in_memory.yml
+++ b/app/config/config_in_memory.yml
@@ -1,0 +1,27 @@
+imports:
+    - { resource: config.yml }
+
+framework:
+    test: ~
+    session:
+        storage_id: session.storage.mock_array
+
+swiftmailer:
+    disable_delivery: true
+
+assetic:
+    use_controller: false
+
+pim_enrich:
+    record_mails: true
+
+services:
+    session.storage.mock_array:
+        class: Symfony\Component\HttpFoundation\Session\Storage\MockArraySessionStorage
+        public: false
+
+    pim_catalog.saver.attribute:
+        class: Pim\Test\Common\Persistence\InMemoryAttributeRepository
+
+    pim_catalog.repository.attribute:
+        alias: pim_catalog.saver.attribute
--- a/composer.json
+++ b/composer.json
@@ -22,6 +22,9 @@
         "classmap": [ "app/AppKernel.php", "app/AppCache.php", "tests/AppKernelTest.php" ]
     },
     "autoload-dev": {
+        "psr-4": {
+            "Pim\\Test\\": "tests/"
+        },
         "files": [
             "vendor/phpunit/phpunit/src/Framework/Assert/Functions.php"
         ]
--- a/behat.yml.dist
+++ b/behat.yml.dist
@@ -103,4 +103,20 @@ default:
                 page: [Context\Page]
             factory:
                 page_parameters:
-                    base_url: http://akeneo-pim-behat.local/
+                    base_url: http://ce-20-behat.pim.local
+
+acceptance:
+    suites:
+        acceptance:
+            paths: ['%paths.base%/tests/Acceptance/']
+            contexts:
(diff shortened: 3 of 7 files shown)


app/config/config_in_memory.yml
```diff
@@ -0,0 +1,27 @@
...
+services:
+    session.storage.mock_array:
+        class: Symfony\Component\HttpFoundation\Session\Storage\MockArraySessionStorage
+        public: false
+
+    pim_catalog.saver.attribute:
+        class: Pim\Test\Common\Persistence\InMemoryAttributeRepository
+
+    pim_catalog.repository.attribute:
+        alias: pim_catalog.saver.attribute
```
phaseinducer: Yeah it looks weird but I didn't find another way to use the same instance for two different services :/

docteurklein: looks ok to me

composer.json
```diff
@@ -22,6 +22,9 @@
         "classmap": [ "app/AppKernel.php", "app/AppCache.php", "tests/AppKernelTest.php" ]
     },
     "autoload-dev": {
+        "psr-4": {
+            "Pim\\Test\\": "tests/"
```
docteurklein: if that was meant only for behat, you could move it in the `autoload` part of behat.yml

app/config/config_in_memory.yml
```diff
@@ -0,0 +1,27 @@
+imports:
```
docteurklein: we don't see how you decide which kernel environment you'll use (I guess in your behat.yml ?)

composer.json
phaseinducer: It can be used by something else than Behat imho.

app/config/config_in_memory.yml
phaseinducer: You're right ! I forgot to push the Behat conf. Here it is : https://github.com/akeneo/pim-community-dev/pull/7209/files#diff-7bde54db60a6e933518d8b61b929edceR121

PR #7246, opened 2017-12-04
API-442: make the content type validation of the API extensible for EE

ahocquard:

Fos Rest bundle allows to have configurate about validate `Accept` type.
This bundle does not allow configuration about content type.
So, I did a PR to add our own configuration, based on the same behavior as Fos Rest.
Unfortunately, this behavior was not extensible for the EE.
So, it was impossible to add configuration without repeating everything in EE.

It would imply to modify EE as soon as you modify the configuration in CE.
It would be dirty and hard to maintain.

So, I had a priority order for the rules about the content-type. Because the order of the rules is important. It's exactly the same behavior as tagged services, but for parameters.

It allows to add rules in EE without repeating everything.

Do note that Symfony DI consider an array as an associative array `[0=> x, ...]`.
So, it's possible to extend an array in EE.

`OK`: Done / Validated
`-`: Not needed


src/Pim/Bundle/EnrichBundle/Resources/config/bundles/fos_rest.yml
```diff
@@ -12,9 +12,6 @@ fos_rest:
             akeneo_multi_json: ['application/vnd.akeneo.collection+json']
     format_listener:
         rules:
-            - { path: '^/api/rest/v\d+/media-files', methods: ['POST'], priorities: ['form_data'], fallback_format: json, prefer_extension: true }
```
ahocquard: legacy, not used anymore since I added the configuration for the content type.
PUT/PATCH/POST only validate content-type header, not accept type.

PR #7250, opened 2017-12-05
PIM-7037: Product Model Import - Error the field "code" should contain a string

aRn0D:

The product model array convert only accept string for the following field `code`, `categories`, `parent` and `family_variant` but these fields should be scalars. For instance, if I want to create a product model with a integer as a code, it should be possible because the code validation rules is `\/^[^,;]+$/`. The regex used to validate the code of other entities is `/^[a-zA-Z0-9_]+$/` for you information. That means that you can create a family with a code 1234 for instance.

### DIFF
--- a/src/Pim/Component/Connector/ArrayConverter/FlatToStandard/ProductModel/FieldConverter.php
+++ b/src/Pim/Component/Connector/ArrayConverter/FlatToStandard/ProductModel/FieldConverter.php
@@ -35,7 +35,7 @@ class FieldConverter implements FieldConverterInterface
     /**
      * {@inheritdoc}
      */
-    public function convert(string $fieldName, string $value): ConvertedField
+    public function convert(string $fieldName, $value): ConvertedField
     {
         $associationFields = $this->assocFieldResolver->resolveAssociationColumns();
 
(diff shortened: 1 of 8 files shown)


aRn0D:
Close in favor of https://github.com/akeneo/pim-community-dev/pull/7252

src/Pim/Component/Connector/ArrayConverter/FlatToStandard/ProductModel/FieldConverter.php
```diff
@@ -52,7 +52,7 @@ public function convert(string $fieldName, string $value): ConvertedField
             return new ConvertedField($fieldName, $categories);
         }
 
-        return new ConvertedField($fieldName, $value);
+        return new ConvertedField($fieldName, (string) $value);
```
aRn0D: I may remove that and add a condition on the code.

jjanvier:
I'm not sure I understand the fix.
When we import a file, every column is read as string no?
So I don't understand why we have to convert something else than strings

But maybe I'm totally wrong

aRn0D:
Yes, for A CSV file we only work with string but not with XSLX Spout return int, datetime and string... At the begin I wanted to solve this problem with https://github.com/akeneo/pim-community-dev/pull/7252 but it break some scenarios about float and date presentation. We have specific rules for XLSX :(

I don't like this solution but I did not find a better idea. Any idea are welcome by the way!

PR #7252, opened 2017-12-05
PIM-7037: Product Model Import - Error the field "code" should contain a string

aRn0D:

The product model array convert only accept string for the following field `code`, `categories`, `parent` and `family_variant` but these fields should be scalars. For instance, if I want to create a product model with a integer as a code, it should be possible because the code validation rules is \/^[^,;]+$/`. The regex used to validate the code of other entities is `/^[a-zA-Z0-9_]+$/` for you information. That means that you can create a family with a code 1234 for instance.


CHANGELOG-2.0.md
```diff
@@ -15,6 +15,7 @@ Changes the constructor of `Pim\Bundle\EnrichBundle\Controller\Rest\ProductModel
 ## Bug fixes
 
 - PIM-6958: fix loading a product with a reference data that is not available (simpleselect or multiselect)
+- PIM-7037: Product Model Import - Error the field "code" should contain a string
```
pierallard: The sentence is not so explicit, try something like "Allows code to be an integer on product model import"

PR #7255, opened 2017-12-05
Impove update of the healthcheck in the daemon (scalability optimisation)

ahocquard:

When consuming the queue, a job can run few seconds (less than 5). For example, a calculation fo the completeness when modifying a product model (and impacting children). 

In that case, most of the time was spent at waiting the next update of the healthcheck (due to the sleep in the loop). 

So, the next job to consume was not done before 5 seconds, which is wasted time. 

When there are a lot of small jobs to run, it could lead to scalability problems. For example, a lot of modification of product models. 

This commit improves it by checking more often that the process is still running far more often, but does not update the healthcheck with the same frequency (to not overload the database).

Without this optimization, it generates fails on the CI because:
- some scenarios load fixture data relative to product models (catalog modeling in general)
- it triggers the calculation of the completeness on products/productmodels/etc. So, some jobs are in the queue.
- you run a job in your test, it is not consumed by the daemon, because it is still busy on the jobs triggered when loading fixture data.
- wait time exception 🎉 

Note: the problem could in theory still occur, but the daemon is optimized the best I can without overloading the database with query.

Second commit: add a wait time when catching an exception. It avoids to loop when database is not accessible for example, consuming all the CPU resources.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Akeneo/Bundle/BatchQueueBundle/Command/JobQueueConsumerCommand.php
+++ b/src/Akeneo/Bundle/BatchQueueBundle/Command/JobQueueConsumerCommand.php
@@ -40,6 +40,9 @@ class JobQueueConsumerCommand extends ContainerAwareCommand
 
     public const COMMAND_NAME = 'akeneo:batch:job-queue-consumer-daemon';
 
+    /** Interval in seconds to wait after an exception occurred.*/
+    private const EXCEPTION_WAIT_INTERVAL = 5;
+
     /** Interval in microseconds before checking if the process is still running. */
     private const RUNNING_PROCESS_CHECK_INTERVAL = 200000;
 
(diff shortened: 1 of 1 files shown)


src/Akeneo/Bundle/BatchQueueBundle/Command/JobQueueConsumerCommand.php
```diff
@@ -77,13 +83,25 @@ protected function execute(InputInterface $input, OutputInterface $output)
 
                 $process->start();
 
-                do {
-                    $this->getJobExecutionManager()->updateHealthCheck($jobExecutionMessage);
-                    // TODO: standard and error output in a dedicated file for each job execution
+                $this->getJobExecutionManager()->updateHealthCheck($jobExecutionMessage);
+
+                $nbIterationBeforeUpdatingHealthcheck = self::HEALTH_CHECK_INTERVAL * 1000000 / self::RUNNING_PROCESS_CHECK_INTERVAL;
```
ahocquard: I didn't find better, but it's explicit, any suggestion welcome :)

Doodoune: typo nbIterationBeforeUpdatingHealth**c**heck - capitalize C

src/Akeneo/Bundle/BatchQueueBundle/Command/JobQueueConsumerCommand.php
```diff
@@ -35,11 +35,17 @@
...
+    /** Interval in seconds before updating health check if job is still running. */
     public const HEALTH_CHECK_INTERVAL = 5;
 
     public const COMMAND_NAME = 'akeneo:batch:job-queue-consumer-daemon';
 
+    /** Interval in microseconds before checking if the process is still running. */
+    private const RUNNING_PROCESS_CHECK_INTERVAL = 200000;
+
+    /** Interval in seconds to wait after an exception occurred.*/
+    private const EXCEPTION_WAIT_INTERVAL = 5;
```
pierallard: really weird to have Seconds and Microseconds in same level of const... Can you unify plz ? in second, it's more human-readable.

ahocquard: The goal is to check every X milliseconds with `usleep` so it's necessary and not doable with `sleep`. So it has to be microseconds.
Do you want me to switch the order of the declaration ?

PR #7266, opened 2017-12-07
[PIM-7029] Better move of search filters

pierallard:
See EE PR
https://github.com/akeneo/pim-enterprise-dev/pull/3352

### DIFF
--- a/src/Pim/Bundle/DataGridBundle/Resources/public/js/datafilter/filters-list.js
+++ b/src/Pim/Bundle/DataGridBundle/Resources/public/js/datafilter/filters-list.js
@@ -99,7 +99,7 @@ define(['underscore', 'pim/form', 'oro/mediator', 'oro/tools'],
                         filter.render();
                     }
                     if (filter.$el.length > 0) {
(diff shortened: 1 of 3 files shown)


src/Pim/Bundle/DataGridBundle/Resources/public/js/datafilter/filters-list.js
```diff
@@ -99,13 +99,22 @@ define(['underscore', 'pim/form', 'oro/mediator', 'oro/tools'],
...
                 }, this);
+            },
+
+            /**
+             * Returns true if the filters list is displayed in the column
+             *
+             * @returns {boolean}
+             */
+            isDisplayedInColumn() {
+                return this.getParent().config.template === 'pim/template/product/grid/column-inner'
```
tamarasaurus: instead of doing this check can you just pass a config flag that says the search filter should be displayed in a column ?

pierallard: That is what I want to avoid, but still better than this shit.

PR #7285, opened 2017-12-11
PIM-6980: Missing labels for attribute prevent you from creating a variant family

aRn0D:

The serialized attribute labels look like 
```
[
    'en_US': 'label',
    'fr_FR': 'label'
]
```
but when there is no translation configured for an attribute, the normalizer return an empty array. So trying to get something that does not exist raise an error and break the select render.

| Javascript test                      | I did not find them, weird :trollface: 

### DIFF
--- a/src/Pim/Bundle/EnrichBundle/Resources/public/js/family/form/variants/add-family-variant-fields-container.js
+++ b/src/Pim/Bundle/EnrichBundle/Resources/public/js/family/form/variants/add-family-variant-fields-container.js
@@ -73,7 +73,7 @@ define(
             render() {
                 const catalogLocale = UserContext.get('catalogLocale');
                 const familyVariant = _.defaults(this.getFormData(), this.getNewFamilyVariant());
-                const label = familyVariant.labels[catalogLocale];
+                const label = this.getEntityLabel(familyVariant, catalogLocale);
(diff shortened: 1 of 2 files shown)


src/Pim/Bundle/EnrichBundle/Resources/public/js/family/form/variants/add-family-variant-fields-container.js
```diff
@@ -264,6 +268,21 @@ define(
...
+            /**
+             * Return the label/code of a serialized entity.
+             *
+             * @param {string} entity
+             * @param {string} locale
+             * @returns {string}
+             */
+            getEntityLabel(entity, locale) {
+                if (0 === entity.labels.length) {
+                    return entity.code;
```
damien-carcel: Shouldn't it be something like 
```javascript
return '[' + entity.code + ']';
```

aRn0D: ok I will update that :)

aRn0D:
I am against adding a Behat for that. This is not the right tool/way to do that, JS tests must introduce in the PIM to convert those kind of things. We lose more time to maintain these kind of tests than dealing with a ticket support.

PR #7292, opened 2017-12-11
PIM-6357: Rename "Edit common attributes" to "Edit attributes"

SamirBoulil:

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/features/Context/Page/Batch/Operation.php
+++ b/features/Context/Page/Batch/Operation.php
@@ -14,9 +14,9 @@ use Context\Page\Base\Wizard;
  */
 class Operation extends Wizard
 {
-    protected $steps = array(
+    protected $steps = [
         'Change status'                    => 'Batch ChangeStatus',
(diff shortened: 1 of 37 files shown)


features/Context/Page/Batch/Operation.php
```diff
@@ -14,18 +14,18 @@
  */
 class Operation extends Wizard
 {
-    protected $steps = array(
+    protected $steps = [
         'Change status'                    => 'Batch ChangeStatus',
-        'Edit common attributes'           => 'Batch EditCommonAttributes',
+        'Edit attributes'                  => 'Batch EditCommonAttributes',
         'Modifier les attributs communs'   => 'Batch EditCommonAttributes',
```
Delphineray: Can we remove it also in french? :)

SamirBoulil: This is used for the behats, so I'm not sure we should remove it for now. but I'll change it in crowndin so that it does not brake the behats.

PR #7299, opened 2017-12-12
PIM-7039: Association grid, scopable attributes used as labels do not appear

aRn0D:

The normalizer called the method `Product::getLabel()` with a locale but without the channel. The identifier was returned all the time because we could not find the value without the channel.

### DIFF
--- a/CHANGELOG-2.0.md
+++ b/CHANGELOG-2.0.md
@@ -4,6 +4,7 @@
 
 - PIM-7037: Allows code to be an integer on product model import
 - PIM-7011: XLS Options Import - Simple select attribute option cannot be updated for options with numeric codes
+- PIM-7039: Association grid, scopable attributes used as labels do not appear
 
 ## Better manage products with variants!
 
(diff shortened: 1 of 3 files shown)


aRn0D:
`shouldBeCalled` won't change anything. For now, if you remove the score and you keep the spec like that I got: 

```
 - it normalizes a product with label
      method call:
        - getLabel("en_US")
      on Double\ProductInterface\P7 was not expected, expected calls were:
        - getFamily()
        - getIdentifier()
        - isEnabled()
        - getCreated()
        - getUpdated()
        - getLabel(exact("en_US"), exact("ecommerce"))
        - getCompletenesses()
```
It fails, it is far from perfect it covers the case... `willReturn` is more than a stub, it is a mock too.

Don't forget that PhpSpec is not testing tool but a tool which help you to design your code. That's why it is useless to write spec before code. The right solution is that we must not mock our model. For now, it is possible because model are not final and they implement an interface. So what do you want?

1) keep it like that?
2) use `shouldBeCalled`?
3) Write unit/integration test with PhpUnit without stub but we are not used to do that! Should we ask the point of view of the team?

jjanvier:
I think you made a mistake here oO
> That's why it is useless to write spec before code.

ok thanks for the explanation, you can keep it like that

but imo, we have a huge education problem in the team
nobody except you imo knows how to use phpspec properly

because today, a `willReturn` is not considered as something important, something that is part of the design
and it can be removed very easily, without asking ourself any question :/

PR #7312, opened 2017-12-13
PIM-6652: Add parent filter for the datagrid

SamirBoulil:

Simply add a new parent filter in the datagrid with the operators "in list" and "is empty".

One or two things I'm not proud of:
1. PQB ParentFilter:
I removed the fact that the filter would check if the code was an actual parent code as it would throw an exception if the user would type a product model that does not exist (there is no autocomplete on this filter unlike the family filter).

3. Front-end ORO filter:
Is simply an extended version of the existing ChoiceFilter, mainly used to inject operators and instanciate the select2 element on show instead of the onOperatorChange.

2. Backend ORO filter:
I didn't re-developped a whole Symfony form type for this filter (yes grid filters are based on that), instead I added the "is_empty" operator instead I used the existing src/Oro/Bundle/FilterBundle/Form/Type/Filter/TextFilterType.php and added the is_empty operator to it.

4. Extend the curent datagrid configuration with a new property `other_columns` dedicated to declare fields that are not directly displayed but can be added using the column configurator. 

`OK`: Done / Validated
`-`: Not needed


src/Pim/Bundle/CatalogBundle/Elasticsearch/Filter/Field/ParentFilter.php
```diff
@@ -88,11 +88,6 @@ protected function checkValue($field, $values)
         FieldFilterHelper::checkArray($field, $values, static::class);
         foreach ($values as $value) {
             FieldFilterHelper::checkIdentifier($field, $value, static::class);
-            if (null === $this->productModelRepository->findOneByIdentifier($value)) {
```
jjanvier: kind of weird of we do it for all other filters

why don't we have autocomplete on the UI of this filter?
is it a need or a shortcut because we don't have enough time?

SamirBoulil: I'm afraid it wouldn't scale ? we don't have an autocomplete on the identifiers so I figured we wouldn't do it for the product models as well.

src/Pim/Bundle/DataGridBundle/Resources/public/js/datafilter/filter/parent-filter.js
```diff
@@ -0,0 +1,75 @@
...
+            },
+
+            /**
+             * @inheritDoc
+             */
+            _readDOMValue: function() {
+                const operator = this.emptyChoice ? this.$('.active .operator_choice').data('value') : 'in';
+
+                return {
+                    value: _.contains(['empty', 'not empty'], operator) ? '' : this._getInputValue(this.criteriaValueSelectors.value),
```
tamarasaurus: you can use `['empty', 'not empty'].includes(operator)`

features/product/filtering/filter_products_and_product_models.feature
```diff
@@ -43,3 +43,15 @@ Feature: Filter product and product models
...
+
+  Scenario: Successfully filters on the parent field with codes
+    Given I am on the products grid
+    When I filter by "parent" with operator "in list" and value "model-braided-hat,diana"
+    Then I should see products braided-hat-m, braided-hat-xxxl, diana_pink, diana_red
+
+  Scenario: Successfully filters on the parent field with empty operator
+    Given I am on the products grid
+    And I show the filter "weight"
+    And I filter by "weight" with operator "is not empty" and value ""
```
jjanvier: why 2 filters on the same test?

src/Pim/Bundle/CatalogBundle/Elasticsearch/Filter/Field/ParentFilter.php
jjanvier: seen IRL, ok

features/product/filtering/filter_products_and_product_models.feature
SamirBoulil: Because I wanted to test the 'is empty' operator with a subset of the available products and product models. the filter on weight is really about giving me this subset.

jjanvier: ah ok yes
I thought it was on the same field

PR #7340, opened 2017-12-18
PIM-7087: Fix completeness normalizer on numeric channel code

anaelChardan:

During generation of data using data generator, we encountered a problem because it generates numeric channel code on the completeness normalizer. This PR aims to allow it.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/CHANGELOG-2.0.md
+++ b/CHANGELOG-2.0.md
@@ -11,6 +11,8 @@
 - PIM-7049: Fix random order of attribute options 
 - PIM-7080: Fix memory leak on product export
 - PIM-6955: Fix delete user
+- PIM-7065: Fix versioning when attribute codes are numerics.
+- PIM-7087: Fix completeness normalization when channel code is numeric.
 
 ## Improvements
 
@@ -28,7 +30,6 @@
 
 - PIM-7032: Fix close button when clicking on "Compare/Translate" option on the product edit form
 - PIM-7063: Add validation for AttributeGroup - cannot remove AttributeGroup containing attributes and cannot remove AttributGroup "other"
-- PIM-7065: Fix versioning when attribute codes are numerics.
 
 # 2.0.9 (2017-12-15)
 
--- a/src/Pim/Bundle/EnrichBundle/Normalizer/CompletenessCollectionNormalizer.php
+++ b/src/Pim/Bundle/EnrichBundle/Normalizer/CompletenessCollectionNormalizer.php
@@ -96,6 +96,7 @@ class CompletenessCollectionNormalizer implements NormalizerInterface
         }
 
         foreach ($sortedCompletenesses as $channelCode => $channelCompletenesses) {
+            $channelCode = (string) $channelCode;
             $normalizedCompletenesses[] = [
                 'channel'   => $channelCode,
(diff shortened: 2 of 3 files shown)


src/Pim/Bundle/EnrichBundle/Normalizer/CompletenessCollectionNormalizer.php
```diff
@@ -223,15 +224,15 @@ protected function normalizeAttributeLabels(AttributeInterface $attribute, $loca
      *
      * @return string[]
      */
-    protected function getChannelLabels(array $channels, array $locales, $channelCode)
+    protected function getChannelLabels(array $channels, array $locales, string $channelCode)
     {
         $matchingChannels = array_filter($channels, function (ChannelInterface $channel) use ($channelCode) {
-            return $channel->getCode() === $channelCode;
+            return (string) $channel->getCode() === $channelCode;
```
aRn0D: I think you should not cast here but in the model. I think that you cannot update the setter because it will break lot of thing (`Channel::setCode(string code)`) but we can cast it in the getter to be sure that the code will a string all the time.

CHANGELOG-2.0.md
```diff
@@ -11,6 +11,8 @@
 - PIM-7049: Fix random order of attribute options 
 - PIM-7080: Fix memory leak on product export
 - PIM-6955: Fix delete user
+- PIM-7065: Fix versioning when attribute codes are numerics.
+- Fix completeness normalization when channel code is numeric.
```
SamirBoulil: PIM code ?

src/Pim/Bundle/EnrichBundle/Normalizer/CompletenessCollectionNormalizer.php
```diff
@@ -96,6 +96,7 @@ public function normalize($completenesses, $format = null, array $context = [])
         }
 
         foreach ($sortedCompletenesses as $channelCode => $channelCompletenesses) {
+            $channelCode = (string) $channelCode;
```
SamirBoulil: do you really need to cast here ? even tough you've done the cast in the getCode() function ?

anaelChardan: yes, because when you get a key from an array if it is castable into an integer, it will do it :/ :viande_php:

CHANGELOG-2.0.md
anaelChardan: I've just got it ;)

src/Pim/Bundle/EnrichBundle/Normalizer/CompletenessCollectionNormalizer.php
SamirBoulil: ok

src/Pim/Bundle/EnrichBundle/Normalizer/CompletenessCollectionNormalizer.php
ahocquard: It's useless as it dies not resolve the problem. It is coming from php array: see explication of @anaelChardan in the response to Samir.

But I agree that the fix should be to typehint of the setter.

PR #7347, opened 2017-12-18
PIM-7061: Fix the edit form of multi-select attributes with a lot of options

damien-carcel:
## Description

In Akeneo 1.x, a lot of forms are still using Twig. This is the case for the attribute edit form, and it can be a problem when we need to load a lot of data, for example a select attribute with several hundred or thousand of options.

To solve this issue, options are loaded asynchronously thanks to JavaScript, so the edit form can be displayed even of the options are still not loaded.

This works fine for simple-select attributes, but not for multi-select ones, as the old FormType is still present and used instead of the new JavaScript form.

This PR removes this old deprecated code so only JavaScript is used to load multi-select options. No tests were added as it does not change the current  behavior of the application.

## Definition Of Done

`OK`: Done / Validated
`-`: Not needed


src/Pim/Bundle/EnrichBundle/Form/Type/AttributeProperty/OptionsType.php
```diff
@@ -1,47 +0,0 @@
...
-use Symfony\Component\OptionsResolver\OptionsResolver;
-
-/**
- * Form type related to options property of AttributeInterface
- *
- * @author    Filips Alpe <filips@akeneo.com>
- * @copyright 2013 Akeneo SAS (http://www.akeneo.com)
- * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
- */
-class OptionsType extends AbstractType
```
solivier: it is not a BC ?

damien-carcel: Possibly yes. This is a piece of code that should not exists anymore and should not be used, but still...

CHANGELOG-1.7.md
```diff
@@ -1,3 +1,7 @@
+# 1.7.x
+
+- PIM-7347: Fix the edit form of multi-select attributes with a lot of options
```
phaseinducer: No `## Bug Fixes` section ?

damien-carcel: oups

src/Pim/Bundle/EnrichBundle/Form/Type/AttributeProperty/OptionsType.php
damien-carcel: I added it to the changelog. Sad to do BC in 1.7, but there is no choice here...

PR #7356, opened 2017-12-20
PIM-7065 Fix versioning when attribute code is numeric

anaelChardan:

This aims to make the history useable when some attributes have a numeric code. Before, using array_merge, the code was normalized from '12345678' to '1' because array_merge reindexes the array.
Use array_replace instead fix this bug by reusing the key. Furthermore, array_replace conserve the ensemblist union from the right to the left.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/CHANGELOG-2.0.md
+++ b/CHANGELOG-2.0.md
@@ -27,6 +27,7 @@
 
 - PIM-7032: Fix close button when clicking on "Compare/Translate" option on the product edit form
 - PIM-7063: Add validation for AttributeGroup - cannot remove AttributeGroup containing attributes and cannot remove AttributGroup "other"
+- PIM-7065: Fix versioning when attribute codes are numerics.
 
 # 2.0.9 (2017-12-15)
 
--- a/src/Pim/Bundle/VersioningBundle/spec/Builder/VersionBuilderSpec.php
+++ b/src/Pim/Bundle/VersioningBundle/spec/Builder/VersionBuilderSpec.php
@@ -48,7 +48,7 @@ class VersionBuilderSpec extends ObjectBehavior
         $pending->setSnapshot(['foo' => 'bar'])->willReturn($pending);
         $pending->getChangeset()->willReturn(['foo' => 'bar']);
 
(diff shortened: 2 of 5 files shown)


CHANGELOG-2.0.md
```diff
@@ -1,5 +1,9 @@
 # 2.0.x
 
+## Bug fixes
+
+- PIM-7065: Fix versioning when attribute code is numeric.
```
phaseinducer: "attribute codes are numeric" ?

src/Pim/Bundle/VersioningBundle/spec/Builder/VersionBuilderSpec.php
```diff
@@ -52,4 +52,15 @@ function it_builds_pending_versions(Version $pending)
...
         $this->buildPendingVersion($pending);
     }
+
+    function it_builds_pending_versions_with_attribute_with_numeric_code(Version $pending)
+    {
+        $pending->setVersion(1)->willReturn($pending);
+        $pending->setSnapshot([12345678 => 'bar'])->willReturn($pending);
+        $pending->getChangeset()->willReturn([12345678 => 'bar']);
+
+        $pending->setChangeset([12345678 => ['old' => '', 'new' => 'bar']])->willReturn($pending);
```
phaseinducer: You don't test anything, I think you should use a shouldBeCalled() here, no ?

CHANGELOG-2.0.md
anaelChardan: it works too.

src/Pim/Bundle/VersioningBundle/spec/Builder/VersionBuilderSpec.php
anaelChardan: Effectively, I copy paste the code from the previous one which does not fix anything neither. I'll fix both ;)

anaelChardan: Actually, the willReturn asserts that the call is well done, but I agree that this is not clearly explicit. ;)

PR #7368, opened 2017-12-21
PIM-7048: Family imports return a cascade persist issue when families have variant family

aRn0D:

During a family import, family variants are loaded in the unit of work but they are not detached after the family saving. At the end of the job we send notification to user, notification are saved in database. So the `flush` method is called here that means that the previous loaded family variant are synchronize with the DB but cascade persist is not configured between Family and Family Variant.

We got this error:
```
A new entity was found through the relationship 'Pim\Component\Catalog\Model\FamilyVariant#family' that was not configured to cascade persist operations for entity: Accessories. To solve this issue: Either explicitly call EntityManager#persist() on this unknown entity or configure cascade persist this association in the mapping for example @ManyToOne(..,cascade={"persist"}). 
```

If we detach family cariant we avoid that error.

### DIFF
--- a/src/Pim/Bundle/CatalogBundle/EventSubscriber/SaveFamilyVariantOnFamilyUpdateSubscriber.php
+++ b/src/Pim/Bundle/CatalogBundle/EventSubscriber/SaveFamilyVariantOnFamilyUpdateSubscriber.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Pim\Bundle\CatalogBundle\EventSubscriber;
 
+use Akeneo\Component\StorageUtils\Detacher\BulkObjectDetacherInterface;
 use Akeneo\Component\StorageUtils\Saver\BulkSaverInterface;
 use Akeneo\Component\StorageUtils\StorageEvents;
 use Pim\Component\Catalog\Model\FamilyInterface;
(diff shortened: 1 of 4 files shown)


src/Pim/Bundle/CatalogBundle/EventSubscriber/SaveFamilyVariantOnFamilyUpdateSubscriber.php
```diff
@@ -26,14 +27,22 @@ class SaveFamilyVariantOnFamilyUpdateSubscriber implements EventSubscriberInterf
...
+     * @param ValidatorInterface          $validator
+     * @param BulkSaverInterface          $familyVariantSaver
+     * @param BulkObjectDetacherInterface $objectDetacher
      */
-    public function __construct(ValidatorInterface $validator, BulkSaverInterface $familyVariantSaver)
-    {
+    public function __construct(
+        ValidatorInterface $validator,
+        BulkSaverInterface $familyVariantSaver,
+        BulkObjectDetacherInterface $objectDetacher
```
damien-carcel: = null

src/Pim/Bundle/CatalogBundle/EventSubscriber/SaveFamilyVariantOnFamilyUpdateSubscriber.php
```diff
@@ -74,6 +83,7 @@ public function validateAndSaveFamilyVariants(GenericEvent $event): void
         }
 
         $this->familyVariantSaver->saveAll($validFamilyVariants);
+        $this->objectDetacher->detachAll($validFamilyVariants);
```
damien-carcel: if (null !=== $this->objectDetacher)

damien-carcel:
If you choose to add `= null` it will need to be removed in master (not sure we'll have the time to merge it in 2.1 for the release...)

aRn0D:
BC added to the changelog

src/Pim/Bundle/CatalogBundle/EventSubscriber/SaveFamilyVariantOnFamilyUpdateSubscriber.php
jjanvier: fixed

src/Pim/Bundle/CatalogBundle/EventSubscriber/SaveFamilyVariantOnFamilyUpdateSubscriber.php
jjanvier: fixed

jjanvier:
one Behat failing in the CI, but it's not related to this CI and it works locally

PR #7380, opened 2017-12-22
PIM-7041: Fix reference data sorter

damien-carcel:
## Description

Let's say you have a simple-select reference data named `color`, and you create an attribute to use it, its code being `color` too, and make this attribute usable in grid filter.
You'll then be able to add a column for this attribute in the product grid and you'll see the values in the column.

But the moment you sort on this attribute, the values will disappear (and be back if you sort on an other attribute or reinitialize the grid). This bug is caused by a wrong datagrid query in the simple-select reference data sorter, fixed by this PR. The problem is present only in ORM.

## Definition Of Done

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/CHANGELOG-1.6.md
+++ b/CHANGELOG-1.6.md
@@ -1,6 +1,13 @@
+# 1.6.x
+
+## Bug fixes
+
+- PIM-7041: Fix a bug that prevents to sort on a reference data attribute on product grid
+
 # 1.6.20 (2017-10-25) 
 
 ## Bug fixes
+
 - GITHUB-6984: Fix wrong path when generating js routes, cheers @amansilla!
 
 # 1.6.19 (2017-09-05)
--- a/features/reference-data/product/sorting/sort_products_per_reference_data.feature
+++ b/features/reference-data/product/sorting/sort_products_per_reference_data.feature
@@ -2,9 +2,9 @@
 Feature: Sort products
   In order to enrich my catalog
   As a regular user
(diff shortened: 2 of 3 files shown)


CHANGELOG-1.6.md
```diff
@@ -1,6 +1,13 @@
+# 1.6.x
+
+## Bug fixes
+
+- PIM-7041: Fix a bug that prevent to sort on a reference data attribute on product grid
```
anaelChardan: that prevents

features/reference-data/product/sorting/sort_products_per_reference_data.feature
```diff
@@ -19,9 +19,31 @@ Feature: Sort products
       | postit  | heel_color  | Pink              |
       | postit  | sole_fabric | Cashmerewool,Silk |
     And I am logged in as "Mary"
+    When I am on the products page
+    Then the grid should contain 2 elements
+    When I display the columns SKU, Sole color, Heel color and Sole fabric
+    Then I sort by "Sole color" value ascending
```
anaelChardan: Where is the assertion ?

damien-carcel: Don't care, it is not my test :D.
More seriously, yes this test is crappy. But there is no assertion already existing. The one for regular attributes doesn't work for reference data attributes. I will not fix that in this PR.

PR #7391, opened 2017-12-22
PIM-7074: Fix Completeness and label filter for OR clauses

SamirBoulil:

Fixes an issue when filtering on both the `label_or_identifier` and `completeness` at the same time.

This issue lays out a more general issue regarding the `SearchQueryBuilder::addShould` function explained in the phpdoc here: https://github.com/akeneo/pim-community-dev/blob/88e69e6ac5763816cf4fdb96c42e988dc8c43d9c/src/Pim/Bundle/CatalogBundle/Elasticsearch/SearchQueryBuilder.php#L81

`OK`: Done / Validated
`-`: Not needed


features/Context/CatalogConfigurationContext.php
```diff
@@ -56,7 +56,7 @@ public function aCatalogConfiguration($catalog)
     {
         $this->initializeReferenceRepository();
 
-        $this->loadCatalog($this->getConfigurationFiles($catalog));
+        $this->loadCatalog($this-getConfigurationFiles($catalog));
```
Doodoune: Why ? - mistake I think

aRn0D: missing `>`

src/Pim/Bundle/CatalogBundle/Elasticsearch/SearchQueryBuilder.php
```diff
@@ -78,6 +78,39 @@ public function addFilter(array $clause)
...
+     *      'A' => 1
+     *    ]
+     *  ],
+     *  [
+     *   'terms' => [
+     *      'A' => 1
+     *    ]
+     *  ]);`
+     *
+     * The resulting logical request looks like this: A = 1 || A = 2 || B = 1 || B = 2 where in fact what the user meant
```
jjanvier: is it on purpose?
do we really use the `should` clause somewhere? should be use it?

SamirBoulil: I don't think we should prevent people from adding should clauses in the API, but yes in the search context using the addShould clauses can damage the search results.

PR #7414, opened 2017-12-28
[PIM-7080] Fix memory leak on product export

pierallard:
# What was wrong ?

Doing a product export on our test environments raised a `Can not allocate memory` after 20-30 minutes of computing.
After running a PHP MemInfo, we saw that some `VariantProduct` entities were still loaded in memory, though Doctrine services.

# What does the fix ?

The fix simply changes the `detach` method to a direct call to `clear()` of UnitOfWork.

# And the impacts on perfs ?

The performances are better with the UoW clear:
- Before the fix: 1227 products / minute
- After the fix: 1842 products / minutes

### DIFF
--- a/src/Akeneo/Bundle/StorageUtilsBundle/Doctrine/Common/Detacher/ObjectDetacher.php
+++ b/src/Akeneo/Bundle/StorageUtilsBundle/Doctrine/Common/Detacher/ObjectDetacher.php
@@ -40,19 +40,9 @@ class ObjectDetacher implements ObjectDetacherInterface, BulkObjectDetacherInter
      */
     public function detach($object)
     {
-        $objectManager = $this->getObjectManager($object);
         $visited = [];
-        $objectManager->detach($object);
+        $this->objectManager->detach($object);
         $this->doDetachScheduled($object, $visited);
-    }
-
-    /**
-     * {@inheritdoc}
-     */
-    public function cleanEntityManager($object): void
-    {
-        $objectManager = $this->getObjectManager($object);
-        $objectManager->clear();
     }
 
     /**
@@ -78,7 +68,7 @@ class ObjectDetacher implements ObjectDetacherInterface, BulkObjectDetacherInter
             return;
         }
 
-        $objectManager = $this->getObjectManager($entity);
+        $objectManager = $this->objectManager;
         $uow = $objectManager->getUnitOfWork();
         $class = $objectManager->getClassMetadata(ClassUtils::getClass($entity));
         $rootClassName = $class->rootEntityName;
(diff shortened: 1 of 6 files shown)


src/Akeneo/Component/StorageUtils/Detacher/ObjectDetacherInterface.php
```diff
@@ -15,4 +15,11 @@
      * @param object $object
      */
     public function detach($object);
+
+    /**
+     * Clean the object using the entity manager.
+     *
+     * @param $object
+     */
+    public function cleanEntityManager($object): void;
```
jjanvier: instead of adding this method in the ObjectDetacher (that ideally, we should not use), wouldn't it be better to use the `UnitOfWorkAndRepositoriesClearer` in the `src/Pim/Component/Connector/Processor/Normalization/ProductProcessor.php` ?

src/Akeneo/Bundle/StorageUtilsBundle/Doctrine/Common/Detacher/ObjectDetacher.php
```diff
@@ -42,13 +42,17 @@ public function detach($object)
...
-        } else {
-            $objectManager->detach($object);
-            $this->doDetachScheduled($object, $visited);
-        }
+    /**
+     * {@inheritdoc}
+     */
+    public function cleanEntityManager($object): void
+    {
+        $objectManager = $this->getObjectManager($object);
```
momoss: Is it needed to pass `$object` to `getObjectManager` because the parameter is not use inside the method?

src/Akeneo/Bundle/StorageUtilsBundle/Doctrine/Common/Detacher/ObjectDetacher.php
```diff
@@ -159,33 +163,4 @@ protected function getObjectManager($object)
...
         return $this->objectManager;
     }
-
-    /**
-     * Do detach objects on DocumentManager
-     *
-     * @param mixed $document
-     * @param array $visited  Prevent infinite recursion
-     */
-    protected function doDetach($document, array &$visited)
```
momoss: ❤️

BitOne:
@pierallard : Where is the service definition changes for the processor ?

WIthout it, this PR won't do anything....

src/Pim/Component/Connector/Processor/Normalization/ProductProcessor.php
```diff
@@ -110,7 +116,12 @@ function ($attributeCode) use ($mediaAttributes) {
             );
         }
 
-        $this->detacher->detach($product);
+        if (null !== $this->cacheClearer) {
+            $this->cacheClearer->clear();
```
BitOne: The UnitOfWork is not a cache at all. What class are you expected here ?

pierallard: Discussed IRL ; to avoid BC, I let it as it, and this has to be changed to master.

pierallard:
This PR was merged with 3 failing behats
features/export/product-export-builder/export_variant_products_by_specific_date.feature:7
features/export/variant_product/csv/export_variant_products.feature:11
features/export/variant_product/xlsx/export_variant_products.feature:11

Explain IRL with the team, @ahocquard is on it.

PR #7432, opened 2018-01-03
API-568: Forbid the use of an non existing attribute in product and variant product creation and update

momoss:

Forbid the use of a non-existing attribute in product and variant product creation and update,
Throw exception when trying to create or update a product or variant product with a non-existing attribute,
it will throw an exception like that:
Property "a_description" does not exist. Check the expected format on the API documentation.

Delete ProductAttributeFilterIntegration.php, don't need to access to db data, because ProductAttributeFilter take array and return array, the tests already exists in phpspec, just add a missing one 

`OK`: Done / Validated
`-`: Not needed


src/Pim/Component/Catalog/Updater/ProductUpdater.php
```diff
@@ -126,20 +127,113 @@ public function update($product, array $data, array $options = [])
...
+                }
+        }
+    }
+
+    /**
+     * Validate association data
+     *
+     * @param $data
+     */
+    protected function validateAssociations($data)
```
LaurentPetard: $data type itself is not validated ?

src/Pim/Component/Catalog/Updater/ProductUpdater.php
```diff
@@ -126,20 +127,113 @@ public function update($product, array $data, array $options = [])
...
         return $this;
     }
 
+    /**
+     * @param ProductInterface $product
+     * @param                  $field
+     * @param                  $data
+     * @param array            $options
+     */
+    protected function setData(ProductInterface $product, $field, $data, array $options = [])
```
LaurentPetard: No strict types ?

momoss: No strict... because setData can handle string or array or whatever you want

src/Pim/Component/Catalog/Updater/ProductUpdater.php
momoss: It could be more readable if we called the method validateAssociationsDataType

src/Pim/Component/Catalog/Updater/ProductUpdater.php
```diff
@@ -126,20 +127,113 @@ public function update($product, array $data, array $options = [])
...
+    }
+
+    /**
+     * Validate association data
+     *
+     * @param $data
+     */
+    protected function validateAssociations($data)
+    {
+        foreach ($data as $associationTypeCode => $associationTypeValues) {
```
LaurentPetard: Check if $data is an array

src/Pim/Component/Catalog/Updater/ProductUpdater.php
```diff
@@ -126,20 +127,113 @@ public function update($product, array $data, array $options = [])
...
+     * Validate that it is an array with scalar values.
+     *
+     * @param string $field
+     * @param mixed  $data
+     *
+     * @throws InvalidPropertyTypeException
+     */
+    protected function validateScalarArray($field, $data)
+    {
+        if (!is_array($data)) {
```
LaurentPetard: There's no test in specs for this case

momoss:
[Green Ci ](http://core-contrib-ci.akeneo.com/blue/organizations/jenkins/akeneo%2Fpim-community-dev/detail/PR-7432/5/pipeline)

PR #7433, opened 2018-01-03
PIM-6355: Fix count by categories in the grid to count products

grena:

This PRs changes the way we count the number of products in category tree.
Now we use ES index to retrieve the number of product per category.

**This results in a small performance improvement**.
For **103k products**:

- Before (MySQL COUNT): `200~300ms`
- After (ES): `50~150ms`

_Tested on my machine (Intel Core i7, 8Gb RAM, SSD)_

screenshot at 2018-01-09 16-29-21

| Added integration tests           | Y

### DIFF
--- a/src/Pim/Bundle/EnrichBundle/Doctrine/Counter/CategoryProductsCounter.php
+++ b/src/Pim/Bundle/EnrichBundle/Doctrine/Counter/CategoryProductsCounter.php
@@ -1,0 +1,61 @@
+<?php
+
+namespace Pim\Bundle\EnrichBundle\Doctrine\Counter;
+
+use Akeneo\Component\Classification\Model\CategoryInterface;
+use Akeneo\Component\Classification\Repository\CategoryRepositoryInterface;
+use Pim\Component\Catalog\Query\Filter\Operators;
+use Pim\Component\Catalog\Query\ProductQueryBuilderFactoryInterface;
+
+/**
+ * Category product counter, using a PQB.
+ *
+ * @author    Adrien Pétremann <adrien.petremann@akeneo.com>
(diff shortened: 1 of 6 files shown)


jjanvier:
I think there is a performance drop because we hydrate products for nothing
we can directly have the number of impacted products if we want, without hydrating them

jjanvier:
We should check with @ahocquard if it's OK or not to have a performance regression here.
Regarding their recent work, I'm not sure they'll be happy with that.

ahocquard:
It will not scale at all for sure. 
The problem is more about the number of product values than the number of products (but the number of products is a factor too). Hydrating all products for the datagrid was already very slow (just for 25 products with 500 product values), imagine about counting all the in the categories!

We should find a way to count with ES.

Note: The whole architecture of this "count" is not scalable. It does a lot of requests. But you will not fix it in a patch for sure.

grena:
Yep @ahocquard , seen with @jjanvier & @SamirBoulil , we changed the way we count for ES Cursor (we don't hydrate), so now we have a small improvement for the request 👍

src/Pim/Bundle/EnrichBundle/Doctrine/Counter/CategoryProductsCounter.php
```diff
@@ -0,0 +1,63 @@
...
+        $this->pqbFactory = $pqbFactory;
+        $this->categoryRepository = $categoryRepository;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getItemsCountInCategory(CategoryInterface $category, $inChildren = false, $inProvided = true)
+    {
+        $categoryCodes = $inChildren
```
SamirBoulil: I'd put this logic in a small private 'getCategoryCodes' for instance

PR #7451, opened 2018-01-05
PIM-7064: [SLA] Keep family attribute panel state on edit

tamarasaurus:

This PR fixes an issue in the family edit attributes screen where all panels were being reopened after any change was made (e.g add/remove attribute or group). Now the panel state is restored on re-render 

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Pim/Bundle/UIBundle/Resources/public/less/components/grid/Grid.less
+++ b/src/Pim/Bundle/UIBundle/Resources/public/less/components/grid/Grid.less
@@ -27,6 +27,10 @@
     background: @AknDefaultHoverBackground;
   }
 
+  &-bodyContainer--collapsed &-bodyRow:not(.group) {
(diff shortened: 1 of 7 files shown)


src/Pim/Bundle/UIBundle/Resources/public/less/components/grid/Grid.less
```diff
@@ -27,6 +27,10 @@
     background: @AknDefaultHoverBackground;
   }
 
+  &-bodyContainer--collapsed &-bodyRow:not(.group) {
```
pierallard: The `.group` is not "BEM" !

tamarasaurus: Fixed

PR #7463, opened 2018-01-08
PIM-7102: do not hydrate product variant with m2m associations to avoid unit of work messing up

docteurklein:
UoW::registerManaged seems to not like the fact the m2m associations are already hydrated.
This fix resolves the problem by not generating an extra set of `DELETE` queries on the related `groups` and `categories` m2m associations.

### DIFF
--- a/src/Pim/Component/Catalog/EntityWithFamily/CreateVariantProduct.php
+++ b/src/Pim/Component/Catalog/EntityWithFamily/CreateVariantProduct.php
@@ -85,16 +85,16 @@ class CreateVariantProduct
 
         $variantProduct->setId($product->getId());
         $variantProduct->setIdentifier($identifierValue);
-        $variantProduct->setGroups($product->getGroups());
         $variantProduct->setAssociations($product->getAssociations());
         $variantProduct->setEnabled($product->isEnabled());
         $variantProduct->setCompletenesses($product->getCompletenesses());
         $variantProduct->setFamily($product->getFamily());
-        $variantProduct->setCategories($product->getCategories());
         $variantProduct->setCreated($product->getCreated());
         $variantProduct->setUpdated($product->getUpdated());
         $variantProduct->setUniqueData($product->getUniqueData());
         $variantProduct->setRawValues($product->getRawValues());
+        // IMPORTANT: we're not assigning categories and groups, because doing so would delete the association!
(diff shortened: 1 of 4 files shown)


jjanvier:
Does that mean you loose the categories and the groups when you turn a product into a variant product?
I don't understand

ahocquard:
I don't understand the fix as well, could you elaborate?

docteurklein:
yes indeed, currently we loose groups and categories (the 2 only m2m assocs on product and variant).
As strange as it sounds, by not assigning them to the variant in `CreateVariantProduct` fixes this wrong behavior.
Note that the "real" assignement is made later by `UoW::registerManaged`. I imagine that doctrine is lost by this double assignement.

jjanvier:
I'm sorry but I don't understand.
Why propose a fix if it introduces a business regression? I think I'm missing something..

jjanvier:
OK seen IRL. I "understand" now.
Not copying groups and categories will actually make them available to the variant product. 
#blackmagic

docteurklein:
exactly

src/Pim/Component/Catalog/tests/integration/EntityWithFamily/CreateVariantProductFromProductIntegration.php
```diff
@@ -89,17 +89,13 @@ public function test_that_the_variant_product_have_the_same_properties_than_the_
 
         $this->assertSame($this->productModel, $variantProduct->getParent());
         $this->assertSame('clothing_color_size', $variantProduct->getFamilyVariant()->getCode());
-        $this->assertSame($this->product->getGroups(), $variantProduct->getGroups());
```
ahocquard: let assert null to remember that it's a dirty fix

src/Pim/Component/Catalog/EntityWithFamily/CreateVariantProduct.php
```diff
@@ -85,12 +85,10 @@ function (ValueInterface $value) {
 
         $variantProduct->setId($product->getId());
         $variantProduct->setIdentifier($identifierValue);
-        $variantProduct->setGroups($product->getGroups());
```
ahocquard: comment with TIP coming from @jjanvier

ahocquard:
Also, it should break spec

PR #7477, opened 2018-01-11
API-565 & API-445: Associated product and variant product to product model

damien-carcel:
## Description

This PR updates and adds some tests about product association throught the API, for both creation and partial updates:
- associate  a non variant product  to a product model
- associate a variant product to a product model (this model being the variant product ancestor)
- test the response when associating to a product model that does not exists

Bad formating of the `associations.*.product_models` field is already unit tested.

## Definition Of Done

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Pim/Bundle/ApiBundle/tests/integration/Controller/Product/CreateProductIntegration.php
+++ b/src/Pim/Bundle/ApiBundle/tests/integration/Controller/Product/CreateProductIntegration.php
@@ -2,7 +2,6 @@
 
 namespace Pim\Bundle\ApiBundle\tests\integration\Controller\Product;
 
-use Akeneo\Test\Integration\Configuration;
 use Pim\Component\Catalog\tests\integration\Normalizer\NormalizedProductCleaner;
 use Symfony\Component\HttpFoundation\Response;
 
@@ -158,20 +157,27 @@ JSON;
 
     public function testProductCreationWithAssociations()
     {
+        $this->createProductModel([
+            'code' => 'a_product_model',
+            'family_variant' => 'familyVariantA1',
+            'values'  => [],
+        ]);
+
         $client = $this->createAuthenticatedClient();
 
-        $data =
-<<<JSON
-    {
-        "identifier": "product_creation_associations",
-        "associations": {
-            "X_SELL": {
-                "groups": ["groupA"],
-                "products": ["simple"],
-                "product_models": []
-            }
+        $data = <<<JSON
+{
+    "identifier": "product_creation_associations",
+    "associations": {
+        "UPSELL": {
+            "product_models": ["a_product_model"]
+        },
+        "X_SELL": {
+            "groups": ["groupA"],
+            "products": ["simple"]
         }
     }
+}
 JSON;
 
         $client->request('POST', 'api/rest/v1/products', [], [], [], $data);
@@ -204,7 +210,7 @@ JSON;
                 "UPSELL"       => [
                     "groups"   => [],
                     "products" => [],
-                    "product_models" => [],
+                    "product_models" => ["a_product_model"],
                 ],
                 "X_SELL"       => [
                     "groups"   => ["groupA"],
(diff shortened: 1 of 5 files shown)


src/Pim/Bundle/ApiBundle/tests/integration/Controller/Product/CreateProductIntegration.php
```diff
@@ -930,6 +936,41 @@ public function testSuccessfullyToCreateProductWithControlCaracter()
...
+    public function testResponseWhenAssociatingToNonExistingProductModel()
+    {
+        $client = $this->createAuthenticatedClient();
+
+        $data = <<<JSON
+{
+    "identifier": "a_product",
+    "associations": {
+        "X_SELL": {
+            "product_models": ["a_non_exiting_product_model"]
```
ahocquard: existing (or it's on purpose as it does not exist ;) )

src/Pim/Bundle/ApiBundle/tests/integration/Controller/Product/PartialUpdateProductIntegration.php
```diff
@@ -1410,6 +1435,41 @@ public function testProductPartialUpdateWithInvalidAttributeData()
...
+    public function testResponseWhenAssociatingToNonExistingProductModel()
+    {
+        $client = $this->createAuthenticatedClient();
+
+        $data = <<<JSON
+{
+    "identifier": "big_boot",
+    "associations": {
+        "X_SELL": {
+            "product_models": ["a_non_exiting_product_model"]
```
ahocquard: same

src/Pim/Bundle/ApiBundle/tests/integration/Controller/Product/CreateProductIntegration.php
```diff
@@ -158,20 +157,27 @@ public function testProductCreationWithCategories()
 
     public function testProductCreationWithAssociations()
     {
+        $this->createProductModel([
```
ahocquard: Maybe you could add an association on a product. It will test that your variant product does inherit of the association parent (opposite behavior of the product values).

src/Pim/Bundle/ApiBundle/tests/integration/Controller/Product/CreateProductIntegration.php
```diff
@@ -2,7 +2,6 @@
 
 namespace Pim\Bundle\ApiBundle\tests\integration\Controller\Product;
 
-use Akeneo\Test\Integration\Configuration;
```
willy-ahva: Why don't add `protected function getConfiguration(): Configuration` with the strict type instead ?

damien-carcel: I though about it, but if I want to be logical, I'll need to typehint the entire class, making quite a bit diff that will polute the PR.

src/Pim/Bundle/ApiBundle/tests/integration/Controller/Product/CreateProductIntegration.php
damien-carcel: On purpose. I want it to fail.

src/Pim/Bundle/ApiBundle/tests/integration/Controller/Product/PartialUpdateProductIntegration.php
damien-carcel: Same ^^.

src/Pim/Bundle/ApiBundle/tests/integration/Controller/Product/CreateProductIntegration.php
damien-carcel: I can't, there is no association on product models :)

ahocquard: oh right (but there will be)

PR #7480, opened 2018-01-11
PIM-7105: Fix un-index variant product on deletion

Doodoune:
**Description**

When we delete a product model, its product variants are not un-indexed from the "product" index.
It's simply because the deletion query is based on "ancestors.ids" but we don't put the "ancestors.ids" in this index.

**Definition Of Done**

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Pim/Component/Catalog/Normalizer/Indexing/Product/PropertiesNormalizer.php
+++ b/src/Pim/Component/Catalog/Normalizer/Indexing/Product/PropertiesNormalizer.php
@@ -121,7 +121,7 @@ class PropertiesNormalizer implements NormalizerInterface, SerializerAwareInterf
     {
         $ancestorsIds = [];
         $ancestorsCodes = [];
-        if ($product instanceof VariantProductInterface) {
+        if ($product instanceof EntityWithFamilyVariantInterface) {
             $ancestorsIds = $this->getAncestorsIds($product);
             $ancestorsCodes = $this->getAncestorsCodes($product);
         }
@@ -155,7 +155,7 @@ class PropertiesNormalizer implements NormalizerInterface, SerializerAwareInterf
      *
      * @return array
      */
-    private function getAncestorsCodes(EntityWithFamilyVariantInterface $entityWithFamilyVariant)
+    private function getAncestorsCodes(EntityWithFamilyVariantInterface $entityWithFamilyVariant): array
     {
         $ancestorsCodes = [];
         while (null !== $parent = $entityWithFamilyVariant->getParent()) {
--- a/src/Pim/Component/Catalog/Normalizer/Indexing/Product/PropertiesNormalizer.php
+++ b/src/Pim/Component/Catalog/Normalizer/Indexing/Product/PropertiesNormalizer.php
@@ -2,6 +2,7 @@
 
 namespace Pim\Component\Catalog\Normalizer\Indexing\Product;
 
+use Pim\Component\Catalog\Model\EntityWithFamilyVariantInterface;
 use Pim\Component\Catalog\Model\ProductInterface;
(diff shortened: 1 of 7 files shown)


src/Pim/Component/Catalog/Normalizer/Indexing/Product/PropertiesNormalizer.php
```diff
@@ -106,4 +110,59 @@ private function getUpdatedAt(ProductInterface $product): \Datetime
...
+
+        return $ancestors;
+    }
+
+    /**
+     * @param EntityWithFamilyVariantInterface $entityWithFamilyVariant
+     *
+     * @return array
+     */
+    private function getAncestorsIds(EntityWithFamilyVariantInterface $entityWithFamilyVariant): array
```
pierallard: I don't understand the signature of this méthod... You only call this method if $product instanceof VariantProductInterface, but sign it with EntityWithFamilyVariantInterface... Can you be coherent between these 2 signatures please ?

src/Pim/Component/Catalog/Normalizer/Indexing/Product/PropertiesNormalizer.php
```diff
@@ -24,6 +25,7 @@ class PropertiesNormalizer implements NormalizerInterface, SerializerAwareInterf
     const FIELD_COMPLETENESS = 'completeness';
     const FIELD_IN_GROUP = 'in_group';
     const FIELD_ID = 'id';
+    private const FIELD_ANCESTORS = 'ancestors';
```
pierallard: Seen with PS, we try to avoid a mix of public/protected/private const in the classes. cc @MarieMinasyan

src/Pim/Component/Catalog/Normalizer/Indexing/Product/PropertiesNormalizer.php
pierallard: There is a typo, it's "getAncestorIds" (without s)
... And I'm not sure it returns "ids", but something with 'product_model_' before... Check if you have no better name.

docteurklein: we only access `getParent`, so `EntityWithFamilyVariantInterface` makes sense here

docteurklein: `product_model_*` actually are ids, from the elasticsearch POV. Since this method iterates over parents of parents, it could be OK to name it "ancestor**s**"?

src/Pim/Component/Catalog/Normalizer/Indexing/Product/PropertiesNormalizer.php
docteurklein: It only makes sense to make it public if it's accessed from the outside, which is not the case here. "Private by default", public if needed" is a good mantra

src/Pim/Component/Catalog/Normalizer/Indexing/Product/PropertiesNormalizer.php
```diff
@@ -106,4 +110,59 @@ private function getUpdatedAt(ProductInterface $product): \Datetime
...
+
+        return $ancestorsIds;
+    }
+
+    /**
+     * @param EntityWithFamilyVariantInterface $entityWithFamilyVariant
+     *
+     * @return array
+     */
+    private function getAncestorsCodes(EntityWithFamilyVariantInterface $entityWithFamilyVariant)
```
grena: return type

docteurklein:
OK, so we already test that `deleteByQuery` (for children of ancestors) is called on **each** index, and we verify each index is correctly indexed, so we indirectly test it by associativity :) no?

PR #7494, opened 2018-01-15
Merge 2.0 -> 2.1 (migration scripts)

SamirBoulil:

`OK`: Done / Validated
`-`: Not needed


upgrades/schema/Version_2_1_171129131942_product_model_associations.php
```diff
@@ -8,7 +8,7 @@
 /**
  * Auto-generated Migration: Please modify to your needs!
  */
-class Version20171129131942 extends AbstractMigration
+class Version_2_1_171129131942_product_model_associations extends AbstractMigration
```
SamirBoulil: Renaming this migration can be harmful if it has already been executed :/

perhaps we should add a simple check ?

PR #7533, opened 2018-01-24
PIM-7116: allow only xhr calls for some actions

grena:

As detected by @BitOne , prior to this PR, it was possible - **as a user logged in the PIM** - to request several endpoints with direct curl call (or other methods, such as using a `src` of an image or a `<form>` element in a page). Example:
```js
productIdsParam = "";

for (i=0; i < 20000; i++) {
    //productIdsParam += "product_" + i + "%2C"; // For 2.0
    productIdsParam += i + "%2C"; // For 1.x

    if (i % 500 == 0) {
        document.write('<img src="'+ targetDomain + '/datagrid/product-grid/massAction/delete?inset=1&values=' + productIdsParam + '" >');
        productIdsParam = "";
    }
}
```

Or 
```html
<html>
<head>
</head>
<body onload='document.getElementById("csrf").submit();'>
<form id="csrf" method="POST" action="http://ref-pim-ee-1-7-orm.akeneo.com/enrich/product/rest/1124">
    <input type="hidden" name="_method" value="DELETE" />
</form>
</body>
</html>
```

According to : https://www.owasp.org/index.php/Cross-Site_Request_Forgery_(CSRF)_Prevention_Cheat_Sheet#Protecting_REST_Services:_Use_of_Custom_Request_Headers, one way to fix is that `X-Requested-With: XMLHttpRequest` would need to be checked on the back-end side.

### DIFF
--- a/features/security/csrf_attacks/allow_only_xhr_requests_for_association_types.feature
+++ b/features/security/csrf_attacks/allow_only_xhr_requests_for_association_types.feature
@@ -1,0 +1,11 @@
+Feature: Allow only XHR requests for some association types actions
+  In order to protect association types from CSRF attacks
+  As a developer
+  I need to only do XHR calls for some association types actions
+
+  Background:
+    Given a "footwear" catalog configuration
+
+  Scenario: Authorize only XHR calls for association types deletion
+    When I make a direct authenticated DELETE call on the "PACK" association type
+    Then there should be a "PACK" association type
--- a/features/security/csrf_attacks/allow_only_xhr_requests_for_association_types.feature
+++ b/features/security/csrf_attacks/allow_only_xhr_requests_for_association_types.feature
@@ -4,7 +4,7 @@ Feature: Allow only XHR requests for some association types actions
   I need to only do XHR calls for some association types actions
 
   Background:
-    Given a "footwear" catalog configuration
+    Given a "default" catalog configuration
 
   Scenario: Authorize only XHR calls for association types deletion
     When I make a direct authenticated DELETE call on the "PACK" association type
--- a/src/Pim/Bundle/EnrichBundle/Controller/AttributeGroupController.php
+++ b/src/Pim/Bundle/EnrichBundle/Controller/AttributeGroupController.php
@@ -253,10 +253,14 @@ class AttributeGroupController
      *
      * @AclAncestor("pim_enrich_attributegroup_remove")
      *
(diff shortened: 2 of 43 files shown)


docteurklein:
Interesting, and surely a good first step, but I wouldn't only rely on only request headers to handle CSRF.

The fact that unsafe operations are doable via GET is the main problem in the first example,
and the second example lacks a CSRF token (even if these tokens could potentially be sniffed by a XSS script (or even [CSS!](https://github.com/dxa4481/cssInjection)).

Concerning redirect, wouldn't it be better to return a [406 not acceptable](https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html#sec10.4.7) or [412 precondition failed](https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html#sec10.4.13) ?

src/Pim/Bundle/EnrichBundle/Controller/AttributeGroupController.php
```diff
@@ -253,10 +253,14 @@ public function sortAction(Request $request)
      *
      * @AclAncestor("pim_enrich_attributegroup_remove")
      *
-     * @return \Symfony\Component\HttpFoundation\RedirectResponse
+     * @return Response
```
damien-carcel: 👍

features/security/csrf_attacks/allow_only_xhr_requests_for_association_types.feature
```diff
@@ -0,0 +1,11 @@
+Feature: Allow only XHR requests for some association types actions
+  In order to protect association types from CSRF attacks
+  As a developer
+  I need to only do XHR calls for some association types actions
+
+  Background:
+    Given a "footwear" catalog configuration
```
skeleton: do you really need the footwear catalog? the default catalog is not enough (it takes less time to be loaded)?

grena: I can check for every one of my test, I don't know if it's really faster though, btw most of the behats I wrote are headless (without `@javascript`) so pretty quick already :)

skeleton: 5 seconds differences. not a big diff I agree, but it's always better to load only what we need

grena: Yup, I'm looking :) Thanks for the idea

PR #5373, opened 2016-12-19
[ON HOLD] PIM-6037: Prevent association code to be an integer

damien-carcel:
**Description**

**Definition Of Done**

| Q                                 | A
| --------------------------------- | ---
| Added Specs                       | :negative_squared_cross_mark:
| Added Behats                      | :white_check_mark:
| Changelog updated                 | :white_check_mark:
| Review and 2 GTM                  | :clock1:
| Micro Demo to the PO (Story only) | :clock1:
| Migration script                  | :negative_squared_cross_mark:
| Tech Doc                          | :negative_squared_cross_mark:

PR #7535, opened 2018-01-24
PIM-7130: Fix product update when association type code is an integer

SamirBoulil:

Related to https://github.com/akeneo/pim-community-dev/pull/5373

The solution provided is to support the association type having a code which is a string representing an integer (ie, "666").

The standard format representation of the associations goes like:
```
{
   "XSELL": {
       "groups": ["group1", "group2"],
       "products": ["AKN_TS1", "AKN_TSH2"]
   },
   "UPSELL": {
       "groups": ["group3", "group4"],
       "products": ["AKN_TS3", "AKN_TSH4"]
   }
}
```
So in our usecase, when updating a product, the front sends to the back the following representation:
```
{
   "666": {
       "groups": ["group1", "group2"],
       "products": ["AKN_TS1", "AKN_TSH2"]
   },
}
```
When json_decoding it, the representation becomes in PHP:
```
[
   666 => [
       "groups": ["group1", "group2"],
       "products": ["AKN_TS1", "AKN_TSH2"]
   ],
]
```
_(the association code is a `"666"`, json_decode will output it as `666`)_

As the association type `666` is not known, the PIM fails to update the product.

__Solution__:

When iterating over the list of associations, explicitly cast the code to a string. _(implemented)_

__Other solution__:

Add a validation rule to prevent the creation of associations having an integer as a code.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/CHANGELOG-1.7.md
+++ b/CHANGELOG-1.7.md
@@ -1,3 +1,9 @@
+# 1.7.X (2018-01-11)
+
+## Bug Fixes
+
+- PIM-7130: Fix product update when association type code is an integer
+
 # 1.7.16 (2018-01-11)
 
 ## Bug Fixes
--- a/src/Pim/Component/Catalog/Updater/Adder/AssociationFieldAdder.php
+++ b/src/Pim/Component/Catalog/Updater/Adder/AssociationFieldAdder.php
@@ -168,6 +168,7 @@ class AssociationFieldAdder extends AbstractFieldAdder
         }
 
         foreach ($data as $assocTypeCode => $items) {
+            $assocTypeCode = (string) $assocTypeCode;
             $this->checkAssociationData($field, $data, $assocTypeCode, $items);
         }
     }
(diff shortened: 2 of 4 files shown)


features/association-type/create_association_type.feature
```diff
@@ -34,3 +34,24 @@ Feature: Association type creation
       | Code | cross_sell |
     And I press the "Save" button
     Then I should see validation error "This value is already used."
+
+  @jira @PIM-7130 @PIM-6037
```
damien-carcel: As it is the 1.7, couldn't you perform an integration test (no UI) rather that a behat?

SamirBoulil: I think I could, even though I don't think we have integration tests for association types yet and I probably won't have the time to write them.

damien-carcel: OK then.

phaseinducer: Even unit tests would be enough from what I saw. Please don't add yet another unjustified end-to-end test.

SamirBoulil: I've already added the specs, If that's enough for you I can remove the behat (even if that's not what we are used to do for the maintenance).

phaseinducer: If we add a test for the (previously) failing case, i.e. testing that passing an integer as code won't raise an exception, then the unit test is enough for me.
What's important is not to add a behat or not like we're used to, it's that you guarantee the non-regression, whatever the tool.

damien-carcel: Agree with Yohan. If the unit test is enough, no need to add a heavy behat using the UI.

PR #7565, opened 2018-01-30
PIM-7131: Fix mass add product values when number of products selected is greater than batch size

willy-ahva:
When mass editing a number of product greater than the batch size an error was throw
`A new entity was found through the relationship 'Pim\Component\Catalog\Model\Completeness#locale' that was not configured to cascade persist operations for entity: en_US. To solve this issue: Either explicitly call EntityManager#persist() on this unknown entity or configure cascade persist this association in the mapping for example @ManyToOne(..,cascade={"persist"}).`.

With @jjanvier we found that the bug was due to a problem of architecture in our Reader/Writer:
    Info:
- The Cursor used in the Reader loads a page of 50 products during `->next()` if there is no more product left in its current page.
- When the batch size (100) is reached, the Writer clears all repositories and doctrine UOW just after writing in DB.
    Steps
- When initializing the reader, a first page of 50 products is loaded by the Cursor.
- At the 50th products, the Cursor has no more product left and during the call to `->next()` it reloads a new page of 50 products.
- At the 100th products, the Cursor has no more product left and during the call to `->next()` it reloads a new page of 50 products.
    -> So at this moment, we have 100 products in the ItemStep and the 50 next products in the UOW.
- The batch size is reached. The Writer writes and ... TADADA ! The UOW is cleared with our 100 old products that have been write and our 50 next products.

We found a dirty fix: call `->next()` before `->current()` and don't `->next()` the first time we should need it. This results in calling `->next()` AFTER the clear of the UOW and fix the problem.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/features/Context/FixturesContext.php
+++ b/features/Context/FixturesContext.php
@@ -139,6 +139,18 @@ class FixturesContext extends BaseFixturesContext
     }
 
     /**
+     * @param int $numberOfProducts
+     *
+     * @Given /^([0-9]+) empty products$/
+     */
+    public function createEmptyProducts(int $numberOfProducts)
+    {
+        for (;$numberOfProducts > 0; $numberOfProducts--) {
+            $this->createProduct(sprintf('product_%s', $numberOfProducts));
+        }
+    }
+
+    /**
      * @param TableNode $table
      *
      * @Given /^the following products?:$/
--- a/CHANGELOG-2.0.md
+++ b/CHANGELOG-2.0.md
@@ -1,5 +1,8 @@
-# 2.0.X
+# 2.0.x
 
+## Bug fixes
+
+- PIM-7131: Fix mass add product values when number of products selected is greater than batch size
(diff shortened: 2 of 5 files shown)


features/Context/FixturesContext.php
```diff
@@ -138,6 +138,18 @@ public function createProduct($data)
         return $product;
     }
 
+    /**
+     * @param int $numberOfProducts
+     *
+     * @Given /^([0-9]+) empty products$/
+     */
+    public function createManyEmptyProducts(int $numberOfProducts)
```
pierallard: IMHO You can crop "Many"

features/Context/FixturesContext.php
```diff
@@ -138,6 +138,18 @@ public function createProduct($data)
...
+    {
+        for (;$numberOfProducts > 0; $numberOfProducts--) {
```
pierallard: useless line

PR #7566, opened 2018-01-30
PIM-7144: Fix translated label of attribute groups in the PEF

SamirBoulil:

Prior to this fix, the label translated would use the uiLocale, now it
uses the catalogLocale.

`OK`: Done / Validated
`-`: Not needed


pierallard:
There is a contribution from an external user for this error

PR #7569, opened 2018-01-30
PIM-7142: Enable the creating of ref-data pv with full numerical codes

SamirBoulil:
**Context:**
Today, when creating a reference data product value using the
`Pim\Component\ReferenceData\Factory\Value\ReferenceDataValueFactory`,
it is only possible to create the product value by passing-in ref-data
codes that are strings (this is ensured by the factory).

In fact, we also want full numerical codes to be supported.

**Solution:**
In the back-end, allow the ReferecenDataValueFactory and
ReferenceDataCollectionValueFactory to use scalar codes to fetch the
data in order to create the product value.

In the front-end, a quick fix that let the front-end input display the
right reference data label when its code is an integer.

`OK`: Done / Validated
`-`: Not needed


SamirBoulil:
The problem comes from the XLSX reader.

PR #7588, opened 2018-02-01
[POC] Improve hydration of the product

ahocquard:

Basically it improves performances of the hydration by 25% for the API.
It should improve the performance of the PIM in general, as it concerns the hydration of a product. 

From 6 seconds to 4,5 seconds for a call in local with 1000 products with ~260 product values.

Note: it reduces the number of request, which is, imho, one of the biggest problem about the scalability of the PIM.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Akeneo/Component/StorageUtils/Repository/IdentifiableManyObjectsRepositoryInterface.php
+++ b/src/Akeneo/Component/StorageUtils/Repository/IdentifiableManyObjectsRepositoryInterface.php
@@ -1,0 +1,29 @@
+<?php
+
+namespace Akeneo\Component\StorageUtils\Repository;
+
+/**
+ * Interface IdentifiableManyObjectRepositoryInterface
+ *
+ * @author    Alexandre Hocquard <alexandre.hocquard@akeneo.com>
+ * @copyright 2017 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+interface IdentifiableManyObjectsRepositoryInterface
+{
+    /**
+     * Returns an array containing the name of the unique identifier properties
+     *
+     * @return array
+     */
+    public function getIdentifierProperties();
+
+    /**
+     * Find an object by its identifier
+     *
+     * @param array $identifiers
+     *
+     * @return mixed
+     */
+    public function findManyByIdentifier(array $identifiers);
+}
--- a/src/Pim/Bundle/CatalogBundle/Doctrine/ORM/Repository/AttributeOptionRepository.php
+++ b/src/Pim/Bundle/CatalogBundle/Doctrine/ORM/Repository/AttributeOptionRepository.php
@@ -140,6 +140,33 @@ class AttributeOptionRepository extends EntityRepository implements
     /**
      * {@inheritdoc}
      */
+    public function findManyByIdentifier(array $codes)
+    {
+        $attributeCodes = array_map(function($code) {
+            [$attributeCode, $optionCode] = explode('.', $code);
+
+            return $attributeCode;
+        }, $codes);
+
+        $optionCodes = array_map(function($code) {
+            [$attributeCode, $optionCode] = explode('.', $code);
+
+            return $optionCode;
+        }, $codes);
+
+        return $this->createQueryBuilder('o')
+            ->innerJoin('o.attribute', 'a')
+            ->where('a.code IN (:attribute_codes)')
+            ->andWhere('o.code IN (:option_codes)')
+            ->setParameter('attribute_codes', $attributeCodes)
+            ->setParameter('option_codes', $optionCodes)
+            ->getQuery()
+            ->getResult();
+    }
+
+    /**
+     * {@inheritdoc}
+     */
     public function findCodesByIdentifiers($code, array $optionCodes)
     {
         return $this->createQueryBuilder('o')
--- a/src/Pim/Bundle/CatalogBundle/Doctrine/ORM/Repository/CachedAttributeOptionRepository.php
+++ b/src/Pim/Bundle/CatalogBundle/Doctrine/ORM/Repository/CachedAttributeOptionRepository.php
@@ -1,0 +1,89 @@
+<?php
+
+namespace Pim\Bundle\CatalogBundle\Doctrine\ORM\Repository;
+
+use Akeneo\Component\StorageUtils\Repository\CachedObjectRepositoryInterface;
+use Akeneo\Component\StorageUtils\Repository\IdentifiableManyObjectsRepositoryInterface;
+use Pim\Component\Catalog\Repository\AttributeOptionRepositoryInterface;
+
+/**
+ * Class CachedAttributeOptionRepository
+ *
+ * @author    Alexandre Hocquard <alexandre.hocquard@akeneo.com>
+ * @copyright 2017 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+class CachedAttributeOptionRepository implements CachedObjectRepositoryInterface, IdentifiableManyObjectsRepositoryInterface
+{
+    /** @var AttributeOptionRepository*/
+    protected $repository;
+
+    /** @var array */
+    protected $objectsCache;
+
+    /**
+     * @param AttributeOptionRepositoryInterface $repository
+     */
+    public function __construct(AttributeOptionRepositoryInterface $repository)
+    {
+        $this->repository = $repository;
+        $this->objectsCache = [];
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function findOneByIdentifier($identifier)
+    {
+        if (!array_key_exists($identifier, $this->objectsCache)) {
+            $this->objectsCache[$identifier] = $this->repository->findOneByIdentifier($identifier);
+        }
+
+        return $this->objectsCache[$identifier];
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function findManyByIdentifier(array $codes)
+    {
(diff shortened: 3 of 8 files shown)


src/Akeneo/Component/StorageUtils/Repository/IdentifiableManyObjectsRepositoryInterface.php
```diff
@@ -0,0 +1,29 @@
...
+namespace Akeneo\Component\StorageUtils\Repository;
+
+/**
+ * Interface IdentifiableManyObjectRepositoryInterface
+ *
+ * @author    Alexandre Hocquard <alexandre.hocquard@akeneo.com>
+ * @copyright 2017 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+interface IdentifiableManyObjectsRepositoryInterface
```
aRn0D: I am not fan of this kind of generic interface, you should introduce interface related to the business now!

src/Akeneo/Component/StorageUtils/Repository/IdentifiableManyObjectsRepositoryInterface.php
```diff
@@ -0,0 +1,29 @@
...
+{
+    /**
+     * Returns an array containing the name of the unique identifier properties
+     *
+     * @return array
+     */
+    public function getIdentifierProperties();
```
aRn0D: Why do you need to method? do you call it in this PR, I don't think.

src/Akeneo/Component/StorageUtils/Repository/IdentifiableManyObjectsRepositoryInterface.php
```diff
@@ -0,0 +1,29 @@
...
+
+    /**
+     * Find an object by its identifier
+     *
+     * @param array $identifiers
+     *
+     * @return mixed
+     */
+    public function findManyByIdentifier(array $identifiers);
```
aRn0D: I don't like this naming `findOne` or `findMany`, this is so generic!

src/Pim/Bundle/CatalogBundle/Doctrine/ORM/Repository/AttributeOptionRepository.php
```diff
@@ -137,6 +137,33 @@ public function findOneByIdentifier($code)
             ->getOneOrNullResult();
     }
 
+    /**
+     * {@inheritdoc}
+     */
+    public function findManyByIdentifier(array $codes)
+    {
+        $attributeCodes = array_map(function($code) {
+            [$attributeCode, $optionCode] = explode('.', $code);
```
aRn0D: This is not the first time saw that, this logic should be encapsulate into an object.

src/Pim/Component/Catalog/Factory/ValueCollectionFactory.php
```diff
@@ -23,25 +27,30 @@ class ValueCollectionFactory implements ValueCollectionFactoryInterface
...
      * @param ValueFactory                    $valueFactory
-     * @param CachedObjectRepositoryInterface $attributeRepository
+     * @param AttributeRepository $attributeRepository
      * @param LoggerInterface                 $logger
      */
     public function __construct(
         ValueFactory $valueFactory,
-        CachedObjectRepositoryInterface $attributeRepository,
+        AttributeRepository $attributeRepository,
+        IdentifiableManyObjectsRepositoryInterface $attributeOptionRepository,
```
jjanvier: missing in the phpdoc

src/Pim/Bundle/CatalogBundle/Doctrine/ORM/Repository/CachedAttributeOptionRepository.php
```diff
@@ -0,0 +1,89 @@
...
+    /** @var array */
+    protected $objectsCache;
+
+    /**
+     * @param AttributeOptionRepositoryInterface $repository
+     */
+    public function __construct(AttributeOptionRepositoryInterface $repository)
+    {
+        $this->repository = $repository;
+        $this->objectsCache = [];
```
aRn0D: I am not fan of caching objects into a repository, yes this is already exist in the PIM but I am still not fan of that. This is a doctrine repository and doctrine use IdentityMap pattern, we could not use it because of our design...

src/Pim/Component/Catalog/Factory/ValueCollectionFactory.php
```diff
@@ -60,8 +69,43 @@ public function createFromStorageFormat(array $rawValues)
...
+                    foreach ($localeRawValue as $localeCode => $data) {
+                        foreach ($data as $optionCode) {
+                            $attributeOptionsCodes[] = $attribute->getCode() . '.' . $optionCode;
+                        }
+                    }
+                }
+            }
+        }
+
+        $this->attributeOptionRepository->findManyByIdentifier($attributeOptionsCodes);
```
jjanvier: hmm, is it normal you don't assign it to a variable?
kind of weird no?

src/Pim/Component/Catalog/Factory/ValueCollectionFactory.php
```diff
@@ -60,8 +69,43 @@ public function createFromStorageFormat(array $rawValues)
     {
         $values = [];
 
+        $attributes = $this->attributeRepository->findManyByIdentifier(array_keys($rawValues));
```
jjanvier: shouldn't / couldn't wrap all this in a private method?

src/Pim/Bundle/CatalogBundle/Doctrine/ORM/Repository/CachedAttributeOptionRepository.php
ahocquard: I disagree on that point. UOW is not a cache. 
It keeps data only because it needs it to track the change of your entity. We should not rely on this behavior for our cache. 
There is absolutely not cache strategy in Doctrine. If it was the case, it would allow to keep only in memory the most used entity (meaning that could detach when you want an entity without failing).

src/Akeneo/Component/StorageUtils/Repository/IdentifiableManyObjectsRepositoryInterface.php
anaelChardan: As it is tied to the Class I don't think this is really generic, but as the Repository is not a business one yes you loose the intention. It it was `MySuperbEntityRepository->findOne($toto)` you have the meaning. The problem with this kind of interfaces is that you also can't type (you can only type mixed), this shouln't exist and can't exists in strong typed languagues.

src/Pim/Bundle/CatalogBundle/Doctrine/ORM/Repository/CachedAttributeOptionRepository.php
anaelChardan: Here this is just a matter of mnemoization, I do agree with @ahocquard  ;)

src/Pim/Bundle/CatalogBundle/Doctrine/ORM/Repository/CachedAttributeOptionRepository.php
```diff
@@ -0,0 +1,89 @@
...
+            $this->objectsCache[$identifier] = $this->repository->findOneByIdentifier($identifier);
+        }
+
+        return $this->objectsCache[$identifier];
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function findManyByIdentifier(array $codes)
```
anaelChardan: why not $identifiers?

src/Pim/Bundle/CatalogBundle/Doctrine/ORM/Repository/AttributeOptionRepository.php
ahocquard: This logic should not even exist. We are forbidding use of "." in the codes with this way of doing it. Meaning that we are coupling validation (forbid to use ".") and indexation.

src/Pim/Bundle/CatalogBundle/Doctrine/ORM/Repository/CachedAttributeOptionRepository.php
ahocquard: Generic interface... for concrete type.

ahocquard:
@aRn0D I understand your comments.
But my principal concern is: we need to improve the performance. Generally, it's an architecture problem. If I need to rewrite the PIM, it would be impossible to determine the limits (my goal), because of the time it can takes to improve performance.

Any advice of the best strategy to adopt is welcome. Also, it's is strongly related to your work.

src/Pim/Component/Catalog/Factory/ValueCollectionFactory.php
ahocquard: oh just a poc you know ^^

ahocquard: oh just a poc you know ^^

src/Pim/Component/Catalog/Factory/ValueCollectionFactory.php
ahocquard: It's like a quick and dirty "cache warmup" of the cache repo.
Otherwise, I assign it to a variable and pass it with an argument function to the generic `$this->valueFactory->create($attribute, $channelCode, $localeCode, $data);`

resulting in:

`$this->valueFactory->create($attribute, $channelCode, $localeCode, $data, $attributeOptions);

jjanvier: yeah ok
so at least could we try to wrap those in private methods like `warmupAttributeOptions` and `fetchAttributes`?
even if that means 2 loops over the `$rawValues`

PR #7596, opened 2018-02-02
PIM-6367: Make property updaters more generic

damien-carcel:
## Description

Property field and attribute updaters should work with any entity with values. Today, we check against `ProductInterface` and `ProductModelInterface`, and the property copier still work only on products.

This PRs update all the property updaters: setter, copier, adder and remover, so they check against `EntityWithValuesInterface`. This is mandatory for the rule engine of the enterprise edition to work properly on variant products and product models.

There is also some renaming: those property updaters were all (except for the setter) prefixed with `Product`, both classes and services. They are not anymore.

Some new unit tests were also added.

## Definition Of Done

`OK`: Done / Validated
`-`: Not needed


src/Pim/Component/Catalog/Updater/Copier/FieldCopierInterface.php
```diff
@@ -17,17 +17,17 @@
...
+     * @param string                    $fromField
+     * @param string                    $toField
+     * @param array                     $options
      *
      * @throws PropertyException
      */
     public function copyFieldData(
-        ProductInterface $fromProduct,
-        ProductInterface $toProduct,
+        EntityWithValuesInterface $fromProduct,
```
phaseinducer: You don't rename the var too ?

PR #7609, opened 2018-02-05
Use correct variant labels when using reference data.

yunosh:

When using a simple-select reference data attribute for a variant axis, the attribute's code is used for the attribute value drop down in the product model navigation bar. Use the select labels instead of the select values by casting the reference data options to strings which eventually uses getLabelProperty().


yunosh:
This is still valid for the current 2.3 branch.

PR #7615, opened 2018-02-06
Fix #7614 wrong base_url for js routes when using dev environment

### LINKED ISSUE #7614: Issue with JS URLs when using dev environment
I guess I found a bug affecting all 2.X versions.

I found the bug while testing some import/export customization where I was using the `dev` front controller `app_dev.php` and see who all Ajax request were performed to the `prod` front controller  `app.php`. In fact all or at least all routes genereated in JS by the FOSJsRoutingBundle are wrong when using the `environment`.

After debugging I found that the `base_url` under `web/js/routes.js` is empty what is ok if you're using `prod` but it should be `app_dev.php` when using `dev`.

amansilla:
I guess I found a bug affecting all 2.X versions.

I found the bug while testing some import/export customization where I was using the dev front controller app_dev.php and see who all Ajax request were performed to the prod front controller app.php. In fact all or at least all routes genereated in JS by the FOSJsRoutingBundle are wrong when using the environment.

After debugging I found that the base_url under web/js/routes.js is empty what is ok if you're using prod but it should be app_dev.php when using dev.

This PR fix the problem.

### DIFF
--- a/app/config/config_dev.yml
+++ b/app/config/config_dev.yml
@@ -29,3 +29,6 @@ oro_assetic:
 
 swiftmailer:
     disable_delivery: true
+
+parameters:
+    router.request_context.base_url: "/app_dev.php"


damien-carcel:
I will perform a few tests on my side, and then merge your PR.
I'll also port your commit to the standard edition and enterprise edition.

amansilla:
@jjanvier I'm not sure as well if it's the final solution but at least fix the problem in the meantime. I'm not sure if using environment variables would allow to concurrently execute requests for dev and prod environments at the same time. I guess that feature of Symfony shouldn't be changed since it's pretty usable if for some reason you want to debug a production server for example.

@damien-carcel Thx to you for the quick feedback and merge 👍 .

damien-carcel:
Hello @amansilla.

I am so sorry, but we had to revert your pull request. This my fault, as it was on `app/config/config_dev.yml`, I assumed it would only affect the `dev` environment (and everything worked fine on my machine using `app_dev.php`).

And so I didn't run a CI on it, as our tests do not run on `dev` environment. Except that it broke completely our nightly builds...

As a workaround, I suggest that you add `router.request_context.base_url: "/app_dev.php"` to your `app/config/parameters.yml` only when you need it, and keep the default configuration the rest of the time.

Regards,
Damien.

amansilla:
Hi @damien-carcel,

sorry for for that. Do you mean this additional configuration on `app/config/config_dev.yml` is affecting the `prod` environment?

If you give me info about how to access to the nightly build results I can try to fix it. Anyway the issue #7614 is there and we should find a way to solve it, since I don't think running always in `prod` environment or even add the parameter to the `app/config/parameters.yml` is the right way to go, at least as final solution.

PR #7618, opened 2018-02-07
PoC: Remove UI from import tests

jjanvier:
## Description

All our import and export Behat currently uses the UI to work.
We log in, configure our job, go to the job page, and click on the "Launch" button of the UI.

This is very time consuming because we boot a browser and Selenium.
This is also useless, as jobs can be launched and tested directly via a CLI.

So, that's the idea of this PR. To test imports/exports via the CLI instead of the UI.
This is not perfect. The ideal would to turn most of those system tests into real acceptance tests. But it's more work. 

**With that simple way of doing, each refactored Behat is twice faster.**

This PR works as is and can be considered as an example to rewrite all other import and export tests.

## About the code 
- we created a new job context, to not pollute the original one. The new one is dedicated to a job you've launched and what you check once the job is finished. It reuses as many code as possible.
- we have a new Gherkin syntax, more natural, to express the fact we import things from a file

### DIFF
--- a/.ci/behat.community.yml
+++ b/.ci/behat.community.yml
@@ -70,6 +70,8 @@ legacy:
                     - 1024
                 - Pim\Behat\Context\JobContext:
                     - 'Context\FeatureContext'
+                - Pim\Behat\Context\ImportFileContext:
+                    - 'Context\FeatureContext'
                 - Pim\Behat\Context\Storage\FileInfoStorage:
                     - 'Context\FeatureContext'
                 - Pim\Behat\Context\Storage\ProductStorage:
--- a/.ci/behat.enterprise.yml
+++ b/.ci/behat.enterprise.yml
@@ -81,6 +81,8 @@ default:
                     - 1024
                 - PimEnterprise\Behat\Context\JobContext:
                     - 'Context\EnterpriseFeatureContext'
+                - Pim\Behat\Context\ImportFileContext:
+                    - 'Context\FeatureContext'
                 - PimEnterprise\Behat\Context\NavigationContext:
                     - 'Context\EnterpriseFeatureContext'
                     - 'http://127.0.0.1/'
--- a/behat.yml.dist
+++ b/behat.yml.dist
@@ -67,6 +67,8 @@ legacy:
                     - 1024
                 - Pim\Behat\Context\JobContext:
                     - 'Context\FeatureContext'
+                - Pim\Behat\Context\ImportFileContext:
+                    - 'Context\FeatureContext'
                 - Pim\Behat\Context\Storage\FileInfoStorage:
                     - 'Context\FeatureContext'
                 - Pim\Behat\Context\Storage\ProductStorage:
--- a/features/Context/FeatureContext.php
+++ b/features/Context/FeatureContext.php
@@ -24,6 +24,7 @@ use Pim\Behat\Context\Domain\Spread\XlsxFileContext;
 use Pim\Behat\Context\Domain\System\PermissionsContext;
 use Pim\Behat\Context\Domain\TreeContext;
 use Pim\Behat\Context\HookContext;
+use Pim\Behat\Context\ImportFileContext;
 use Pim\Behat\Context\JobContext;
 use Pim\Behat\Context\PimContext;
 use Pim\Behat\Context\Storage\FileInfoStorage;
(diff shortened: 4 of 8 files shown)


features/Pim/Behat/Context/LaunchJobContext.php
```diff
@@ -0,0 +1,82 @@
...
+/**
+ * This context aims to contain all methods to the launch of a job.
+ *
+ * When we launch a job, we keep in this context the job instance and execution so that we can
+ * check what happens in a next step. For instance, we want to check to 1 item has been skipped.
+ *
+ * This also allows to not use several time the same name of a job in the different steps of a use case.
+ * It gives a more natural language.
+ */
+class LaunchJobContext extends PimContext implements SnippetAcceptingContext
```
skeleton: do you really need to extends PimContext? because you expose your context to some elements related to the UI (Page, Mink, etc)

features/Pim/Behat/Context/LaunchJobContext.php
```diff
@@ -0,0 +1,82 @@
...
+            $this->getFixturesContext()->refresh($this->jobInstance);
+            $this->jobExecution = $this->jobInstance->getJobExecutions()->last();
+            $this->getFixturesContext()->refresh($this->jobExecution);
+
+            return $this->jobExecution && !$this->jobExecution->isRunning();
+        }, sprintf('The job execution of "%s" was too long', $this->jobInstance->getJobName()));
+    }
+
+    /**
+     * @Then there should be :number product(s) skipped because there is no difference
```
skeleton: replace `product(s)` by `products?`?

features/Pim/Behat/Context/LaunchJobContext.php
```diff
@@ -0,0 +1,82 @@
...
+    use SpinCapableTrait;
+
+    /** @var JobInstance */
+    private $jobInstance;
+
+    /** @var JobExecution */
+    private $jobExecution;
+
+    /**
+     * @When I import it via the job :jobName as :username
```
skeleton: protect your regexp with anchors

docteurklein: it's not a regex, it's "turnip"

features/Pim/Behat/Context/LaunchJobContext.php
docteurklein: turnip syntax, again. this allow to write `1 product` or `2 products`

features/Pim/Behat/Context/LaunchJobContext.php
jjanvier: this is the way we access to `$placeholderValues`
so except if we rewrite that part, yes, we need to

features/Pim/Behat/Context/LaunchJobContext.php
```diff
@@ -0,0 +1,82 @@
...
+        $this->iImportItViaTheJobAsWithOptions($jobName, $username, new TableNode([]));
+    }
+
+    /**
+     * @When I import it via the job :jobName as :username with options:
+     */
+    public function iImportItViaTheJobAsWithOptions($jobName, $username, TableNode $jobOptions)
+    {
+        $newJobOptions = new TableNode(
+            array_merge($jobOptions->getTable(), [['filePath', self::$placeholderValues['%file to import%']]])
```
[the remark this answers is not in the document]
jjanvier: this holds the "variables" we use in our Gherkin
like `%file to import%` for instance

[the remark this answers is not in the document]
aRn0D: weird, that means you created a statefull context?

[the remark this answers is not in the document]
aRn0D: it could be replace by :
```
the following file is imported via the job "csv_footwear_currency_import" as "Julia":
"""
file content
"""
```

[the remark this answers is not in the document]
jjanvier: @aRn0D yes, you're right
we asked this exact same question with @docteurklein 

but we agreed that, during an import, "waiting" is important
so we tried to express that in our Gherkin
which is why we end up with a statefull context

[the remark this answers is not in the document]
jjanvier: @ahocquard because this is something our users actually do
it's an import
they understand it can take a long time
this state of affairs should be expressed imo

jjanvier:
@skeleton @ahocquard @aRn0D 

I updated the PR:
- the step "I wait for the job to finish" is removed
- the Gherkin is now more natural
- the Gherkin doesn't care about who launched the import. And that's what we want in those use cases (setup imports, or products coming each night from the ERP).
- the context has been renamed to avoid confusion

PR #7620, opened 2018-02-08
PIM-7151: Remove family limit used on export profile

Doodoune:
**Description**

When you select more than 20 families to export in the Family Field, only 20 are displayed whereas they are all well exported.
Expected behaviour: all selected families should be displayed in the family field.

### DIFF
--- a/CHANGELOG-1.6.md
+++ b/CHANGELOG-1.6.md
@@ -6,6 +6,7 @@
 - PIM-7169: Fix a memory leak on product export when having many variant groups
 - PIM-7170: Fix media files unnecessarily generated during quick export
 - PIM-7182: Fix the clean mongodb command for reference data
+- PIM-7151: Fix the family filter display on import/export profile
 
 ## Security fixes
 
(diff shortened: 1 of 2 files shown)


willy-ahva:
Do you know why the limit were here ? Seems GTM but this limit be there for some reason.

Doodoune:
The limit appear after : PIM-5542 & it was for optimise time (that is cool - but limit is just a small optimisation in all the ticket), and it show incorrect informations on UI => if I pick 23 families, it only show me 20 families

LaurentPetard:
Maybe the limit should be removed only if the there's a filter on identifiers ?

Doodoune:
Yeah It can work ! I change that.

CHANGELOG-1.6.md
```diff
@@ -6,6 +6,7 @@
 - PIM-7169: Fix a memory leak on product export when having many variant groups
 - PIM-7170: Fix media files unnecessarily generated during quick export
 - PIM-7182: Fix the clean mongodb command for reference data
+- PIM-7151: Remove family limit used on export profile
```
phaseinducer: This is the technical solution, I think describing which bug it fixes would be more useful.

PR #7635, opened 2018-02-09
PIM-7155: Fix remove a locale from a channel

willy-ahva:

The process was synchronous and if a lot of products were impacted the processus was too long to finished. Moreover, there was a memory leak which does not allow to make the process asynchronous.
Each bug are fixed.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Pim/Bundle/EnrichBundle/EventListener/Storage/ChannelLocaleSubscriber.php
+++ b/src/Pim/Bundle/EnrichBundle/EventListener/Storage/ChannelLocaleSubscriber.php
@@ -2,14 +2,16 @@
 
 namespace Pim\Bundle\EnrichBundle\EventListener\Storage;
 
+use Akeneo\Component\Console\CommandLauncher;
 use Akeneo\Component\StorageUtils\Saver\BulkSaverInterface;
 use Akeneo\Component\StorageUtils\StorageEvents;
-use Pim\Component\Catalog\Completeness\CompletenessGeneratorInterface;
 use Pim\Component\Catalog\Completeness\CompletenessRemoverInterface;
 use Pim\Component\Catalog\Model\ChannelInterface;
+use Pim\Component\Catalog\Model\LocaleInterface;
 use Pim\Component\Catalog\Repository\LocaleRepositoryInterface;
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
 use Symfony\Component\EventDispatcher\GenericEvent;
+use Symfony\Component\Security\Core\Authentication\Token\Storage\TokenStorageInterface;
 
 /**
  * Storage event subscriber that update channel locales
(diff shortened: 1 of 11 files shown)


pierallard:
Don't see CHANGELOG

src/Pim/Bundle/EnrichBundle/EventListener/Storage/ChannelLocaleSubscriber.php
```diff
@@ -90,25 +106,75 @@ public function updateChannel(GenericEvent $event)
             return;
         }
 
-        $oldLocales = $this->repository->getDeletedLocalesForChannel($channel);
-        $newLocales = $channel->getLocales();
-        $updatedLocales = [];
+        /**
+         * TODO: Pull-up day. During merge in master remove condition and `else` part.
+         */
+        if (null !== $this->commandLauncher && null !== $this->tokenStorage) {
```
pierallard: replace && by || (you need both to work)

willy-ahva: nope it's &&, I need both.

pierallard: Maybe you should extract all this { } into another private function in this file, to be clearer.

src/Pim/Bundle/EnrichBundle/spec/EventListener/Storage/ChannelLocaleSubscriberSpec.php
```diff
@@ -61,31 +67,75 @@ function it_does_not_upadte_channel_on_non_channels($saver, GenericEvent $event,
         $this->updateChannel($event);
     }
 
-    function it_updates_channel(
+    function it_updates_channel_when_a_locale_has_been_remove(
```
pierallard: removed*

src/Pim/Bundle/EnrichBundle/spec/EventListener/Storage/ChannelLocaleSubscriberSpec.php
```diff
@@ -61,31 +67,75 @@ function it_does_not_upadte_channel_on_non_channels($saver, GenericEvent $event,
...
+
+        $commandLauncher
+            ->executeBackground(
+                'pim:catalog:remove-completeness-for-channel-and-locale print en_US julia'
+            )->shouldBeCalled();
+
+        $this->updateChannel($event);
+    }
+
+    function it_updates_channel(
```
pierallard: _without_removed_locale

src/Pim/Bundle/EnrichBundle/EventListener/Storage/ChannelLocaleSubscriber.php
pierallard: Sorry I read it in the opposite way... I prefer early return.
`if (null === $this->commandLauncher || null === $this->tokenStorage) { return; }`

momoss:
What do you think about use `pim_connector.doctrine.cache_clearer`, it's JJ's one, and it make a great job

src/Pim/Bundle/EnrichBundle/EventListener/Storage/ChannelLocaleSubscriber.php
willy-ahva: nope because if one of them is not here I need to launch the old way to do.

willy-ahva:
@momoss It was the first thing I tried, but it can't work because it is in a PRE_SAVE event, so it clears also what I need to save just after.

src/Pim/Bundle/EnrichBundle/EventListener/Storage/ChannelLocaleSubscriber.php
willy-ahva: done

PR #7648, opened 2018-02-12
PIM-7165: Add "FamilyVariantInterface::getLevelForAttributeCode"

damien-carcel:
## Description

Add a new method in the family variant to know to which variant level an attribute belongs too (if it belongs to the family of the family variant at all).

## Definition Of Done

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Pim/Component/Catalog/Model/FamilyVariant.php
+++ b/src/Pim/Component/Catalog/Model/FamilyVariant.php
@@ -90,7 +90,7 @@ class FamilyVariant implements FamilyVariantInterface
      */
     public function getVariantAttributeSet(int $level): ?VariantAttributeSetInterface
     {
-        if ($level <= 0) {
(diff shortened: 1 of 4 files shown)


src/Pim/Component/Catalog/Model/FamilyVariant.php
```diff
@@ -90,7 +90,7 @@ public function getCommonAttributes(): CommonAttributeCollection
      */
     public function getVariantAttributeSet(int $level): ?VariantAttributeSetInterface
     {
-        if ($level <= 0) {
+        if (0 >= $level) {
```
grena: I don't see the point of doing yoda when using `<`, but ok ;)

pierallard: For more unreadability @grena

PR #7649, opened 2018-02-13
PIM-7170 Fix media files unnecessarily generated during quick export

LaurentPetard:

When a products grid context quick export is performed, the products media files are generated even if no attribute of type media are selected in the grid context.

To fix this bug without doing BC breaks, I added a filter on the product values that are used to fetch the media, to keep only those that are selected from the context.

I didn't add specs because there isn't any for this class, and it no longer exists in 2.0

I did another PR to fix this bug for version 2.0 and beyond.

CI green except failing tests due to CI instability (issue raised). Tests are ok in local: merge.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Pim/Bundle/EnrichBundle/Connector/Processor/QuickExport/ProductProcessor.php
+++ b/src/Pim/Bundle/EnrichBundle/Connector/Processor/QuickExport/ProductProcessor.php
@@ -7,6 +7,7 @@ use Akeneo\Component\Batch\Job\JobInterface;
 use Akeneo\Component\Batch\Job\JobParameters;
 use Akeneo\Component\Batch\Model\StepExecution;
 use Akeneo\Component\StorageUtils\Detacher\ObjectDetacherInterface;
+use Doctrine\Common\Collections\Collection;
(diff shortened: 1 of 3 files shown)


src/Pim/Bundle/EnrichBundle/Connector/Processor/QuickExport/ProductProcessor.php
```diff
@@ -152,15 +157,16 @@ protected function areAttributesToFilter(JobParameters $parameters)
...
     /**
-     * Fetch medias on the local filesystem
+     * Fetch medias on the local filesystem for given product values.
      *
      * @param ProductInterface $product
      * @param string           $directory
+     * @param ArrayCollection  $productValues
      */
-    protected function fetchMedia(ProductInterface $product, $directory)
+    protected function fetchMedia(ProductInterface $product, $directory, ArrayCollection $productValues)
```
damien-carcel: `Collection` (interface of ArrayCollection)

LaurentPetard:
The build has errors but they're related to an issue on the CI : https://github.com/akeneo/continuous-integration/issues/306
All the tests are green and I launched locally those related to the quick export.

PR #7671, opened 2018-02-14
[Proposal] Introduce makefile

arnolanglade:
What do you think about using makefile instead of several small shell script? For now, the makefile I created is very simple but it can be improved.

5 simples commands have been introduced
* Clear caches
* Install front stuff
* Install the the pim

Run a command:
```bash
$ make pim-front
$ make install ENV=dev # default env is prod
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,13 @@
+ENV=prod
+
+clear-cache: 
+	rm -rf ./var/cache
+	
+pim-front: clear-cache
+	rm -rf ./web/js ./web/css
+	bin/console pim:installer:assets --env=$(ENV) --symlink --clean
+	yarn run webpack-dev
+
+install: clear-cache
+	bin/console console pim:install --env=$(ENV) --force
+


ronanguilloux:
Love the concept!

Maybe you could check how to make it clear that certain targets here are actually _phony_, in the sense they do not generate file as an outcome – Read more here: https://www.gnu.org/software/make/manual/html_node/Phony-Targets.html

anaelChardan:
Hi, very good idea, as @ronanguilloux said please use Phony ;). And also purpose a way to display all available commands through a menu as it is there for example: https://github.com/akeneo/transporteo/blob/master/Makefile ;).

Makefile
```diff
@@ -0,0 +1,13 @@
+ENV=prod
+
+clear-cache: 
+	rm -rf ./var/cache
```
ahocquard: Is the command executed relative to your actual directory or relative to the makefile (`./var/cache`)?
In shell, it's relative to your actual directory. So, it's a good practice to execute it relative to the shell script.
Probably the same thing here.

aRn0D:
I agree with you @ahocquard. For now I just want to get feedback!

PR #7674, opened 2018-02-15
Small Docker Compose enhancements

damien-carcel:
## Description

This PR adds small improvements to the Docker Compose file and helper scripts:
- use `rm` command to delete cache, but only in containers (so it prevent deleting anything on the user machine outside of the PIM project)
- fix some minor default in YAML syntax
- remove `PHP_XDEBUG_REMOTE_HOST` as it is not needed most of the time (and is documented anyway)
- put more folder/files in read-only
- change the mapping of the port to avoid using already used ports on the host machine

## Definition Of Done

| Changelog updated                 | --

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/docker-compose.yml.dist
+++ b/docker-compose.yml.dist
@@ -7,8 +7,7 @@ services:
       COMPOSER_HOME: '/home/docker/.composer'
       PHP_IDE_CONFIG: 'serverName=pim-ce'
       PHP_XDEBUG_ENABLED: 0
-      PHP_XDEBUG_IDE_KEY: XDEBUG_IDE_KEY
-      PHP_XDEBUG_REMOTE_HOST: xxx.xxx.xxx.xxx
+      PHP_XDEBUG_IDE_KEY: 'XDEBUG_IDE_KEY'
       XDEBUG_CONFIG: 'remote_host=xxx.xxx.xxx.xxx'
     ports:
       - '8080:80'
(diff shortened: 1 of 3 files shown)


docker-compose.yml.dist
```diff
@@ -7,8 +7,7 @@ services:
       COMPOSER_HOME: '/home/docker/.composer'
       PHP_IDE_CONFIG: 'serverName=pim-ce'
       PHP_XDEBUG_ENABLED: 0
-      PHP_XDEBUG_IDE_KEY: XDEBUG_IDE_KEY
-      PHP_XDEBUG_REMOTE_HOST: xxx.xxx.xxx.xxx
+      PHP_XDEBUG_IDE_KEY: 'XDEBUG_IDE_KEY'
       XDEBUG_CONFIG: 'remote_host=xxx.xxx.xxx.xxx'
```
pierallard: Just thinking... Should be better to run `docker-compose up XDEBUG_CONFIG=$(hostname -I)` (or something like this) to avoid this weird configuration ?

damien-carcel: Dunno what's the best. Personnaly, I don't like to add config into the command. I prefer set it in my docker-compose file once and for all.
That being said, it is also possible to use environment variables in the compose file. It is something I want to investigate (but not in this PR).

PR #7675, opened 2018-02-15
reset-es-index: Reset Es indexes between tests (behat, integration)

momoss:

Reset Es indexes between tests (behat, integration), the main goal is to not worry about potential new indexes added in the future 

`OK`: Done / Validated
`-`: Not needed


fitn:
Just a probably silly question but why do we have many clients?

solivier:
@fitn because we have different indexes for different entities. So each client has a different configuration.

momoss:
@fitn actually we have one client by index

momoss:
[Green EE](https://ci.akeneo.com/blue/organizations/jenkins/akeneo%2Fpim-community-dev/detail/PR-7675/2/pipeline/)

PR #7686, opened 2018-02-19
PIM-7188 only detach products if not seen as modified by a previous line

docteurklein:
The diff detection allows one product to be listed as "to write", but if the same identifier appears later in the read stream and no data changed, it is marked as detached.

This results in a detached entity waiting to be written, resulting in an INSERT instead of an UPDATE.
Consequence: we get duplicated products.

The only solution without breaking the public API is to only detach the product if it has not been appended to the "to write" list.

`OK`: Done / Validated
`-`: Not needed


src/Pim/Component/Connector/Processor/Denormalization/ProductProcessor.php
```diff
@@ -93,16 +98,20 @@ public function process($item)
...
-                $this->detachProduct($product);
+                if (!isset($this->seenWithDiff[$identifier])) {
+                    // only detach products when no previous line has modified the same identifier (to avoid duplicates)
+                    $this->detachProduct($product);
+                }
                 $this->stepExecution->incrementSummaryInfo('product_skipped_no_diff');
 
                 return null;
+            } else {
+                $this->seenWithDiff[$identifier] = true;
```
ahocquard: this list should be reset for each batch no ?

docteurklein: the writer already `detachAll` after each write, so the "real stuff" (entity) is already freed from RAM.
This array will continue to grow, but it only contains boolean values, resulting in ~30MB for a million entries.
We could try to add a hook to reset the array, but it would force us to hack around the public API of the processors and steps, which are not made for that.

src/Pim/Component/Connector/Processor/Denormalization/ProductProcessor.php
```diff
@@ -43,6 +43,9 @@ class ProductProcessor extends AbstractProcessor implements ItemProcessorInterfa
     /** @var ProductFilterInterface */
     protected $productFilter;
 
+    /** @var array list of identifiers previously seen with at least one diff in this process */
+    private $seenWithDiff = [];
```
pierallard: I don't like the naming at all ! Try to be more explicit please.

docteurklein: `seenWithModifiedData` ? I have no better idea, as this is litterally what it is about, but I'm open to any alternative :)

docteurklein:
@willy-ahva what should I test in addition to the fact that the detacher is called only when it worths it?

src/Pim/Component/Connector/Processor/Denormalization/ProductProcessor.php
ahocquard: Ok so if a product already handled exist in the next batch, it will not be detached if there is not modification, as it's it's already in the list `seenWithDiff`.
So, it's a kind of leak. 
But I think it does not occur, it would mean that we only have duplicates in the file to produce a noticeable leak.

src/Pim/Component/Connector/Processor/Denormalization/ProductProcessor.php
pierallard: First, it's an identifier list, so I think you should at least have identifier on the name.

docteurklein: Don't forget it's not part of the public API (doesn't mean it shouldn't be named correctly, but still)

rybus: `modifiedProductIdentifiers`

PR #7698, opened 2018-02-21
PIM-7164: Detach associated products during export

damien-carcel:
## Description

There is currently a memory leak on product export, caused by associated products not being detached.
The idea to fix this is to get the associations of an exported product, detach all the products they contain, then finally detach the exported product.

However, there is a problem here: some of the detached associated products could be in the batch too, needing to be exported later. In this case, we would need not to detach them as associated products, but wait they are normalized and passed to the writer, except we cannot predict if an associated product will be in the batch or not.

The solution proposed here is to stop detaching the exported products one by one in the processor, as they are currently, and to store them in a new detacher dedicated to product export. Then the detach itself will be perform from the writer, all the products of the batch at once. This way, we ensure we are not loosing data during the export.

It is hackish and dirty, but it is working. On a data set from a client, with about 4000 products, each with between 400 and 1200 values, the export memory was reaching more than 1 GiB in 20 minutes. Client reported exports of 700 to 900 products in half an hour.

With this fix, memory is stable at 286 MiB, and the export of 7000 products takes 15 minutes.

## Definition Of Done

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Pim/Bundle/ConnectorBundle/Doctrine/Common/Detacher/StoredProductDetacherInterface.php
+++ b/src/Pim/Bundle/ConnectorBundle/Doctrine/Common/Detacher/StoredProductDetacherInterface.php
@@ -1,0 +1,27 @@
+<?php
+
+namespace Pim\Bundle\ConnectorBundle\Doctrine\Common\Detacher;
+
+use Pim\Component\Catalog\Model\ProductInterface;
+
+/**
+ * Detach at once products that have been stored for.
+ *
+ * @author    Damien Carcel <damien.carcel@akeneo.com>
+ * @copyright 2018 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+interface StoredProductDetacherInterface
+{
+    /**
+     * Stores the products to be bulk detached.
+     *
+     * @param ProductInterface $product
+     */
+    public function storeProductToDetach(ProductInterface $product);
+
+    /**
+     * Detaches the products that have been stored.
+     */
+    public function detachStoredProducts();
+}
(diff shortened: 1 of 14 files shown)


src/Pim/Bundle/ConnectorBundle/Doctrine/Common/Detacher/StoredProductDetacherInterface.php
```diff
@@ -0,0 +1,27 @@
...
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+interface StoredProductDetacherInterface
+{
+    /**
+     * Stores the products to be bulk detached.
+     *
+     * @param ProductInterface $product
+     */
+    public function storeProduct(ProductInterface $product);
```
grena: I don't know, maybe not needed, but maybe we could name this method `storeProductToDetach` to avoid any misunderstood of the "store" action here. But again, maybe too much.

src/Pim/Component/Connector/Processor/Normalization/ProductProcessor.php
```diff
@@ -182,7 +185,7 @@ protected function getAttributesToFilter(JobParameters $parameters)
     }
 
     /**
-     * Are there attributes to filters ?
+     * Are there attributes to filters?
```
grena: (...) to filter?

damien-carcel:
@grena Regarding your concern, I don't think it is a problem. If the `StoredDetacher` was not injected, then the old way of detaching product one by one is still kept.

Plus, this is only for 1.7. This fix is not needed in 2.x.

grena:
@damien-carcel You're right! I forgot your fallback on the classic way.

jjanvier:
please remember to revert that in 2.0
we should completely clear the UoW to avoid those problems

PR #7707, opened 2018-02-22
IM-824: Fix security issue when requesting password

SamirBoulil:

**Context:**
When a user asks for a new password the pim will show a message to the user indicating the email address the email has been set or tell the user if the username/email is invalid.

**Problem:**
This is a security issue as an attacker can then guess if a username/elam exists or not.

**Solution:**
Print out a standard message, in all cases, saying that an email has been set to the user.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/CHANGELOG-1.7.md
+++ b/CHANGELOG-1.7.md
@@ -1,6 +1,10 @@
 # 1.7.x
 
-## Big Fixes
+# Improvements
+
+- IM-824: Change message when the user or email is not valid to a more generic message
+
+## Bug Fixes
 
 - PIM-7164: Fix a memory leak on product export caused by associated products not being detached
 
--- a/src/Pim/Bundle/UserBundle/Resources/views/Reset/request.html.twig
+++ b/src/Pim/Bundle/UserBundle/Resources/views/Reset/request.html.twig
@@ -10,13 +10,6 @@
 
     <form action="{{ path('oro_user_reset_send_email') }}" method="post" class="form-reset">
         <h2 class="AknLogin-subTitle">{{ 'Forgot Password'|trans }}</h2>
-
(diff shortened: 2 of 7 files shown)


CHANGELOG-1.7.md
```diff
@@ -1,5 +1,9 @@
 # 1.7.x
 
+# Improvements
+
+- IM-824: Fix security issue when requesting password
```
BitOne: A little less panic inducing message would be welcome. Like "change message when the user or email is not valid to a more generic message" ;)

src/Pim/Bundle/UserBundle/Resources/views/Reset/request.html.twig
```diff
@@ -13,7 +13,7 @@
 
         {% if invalid_username is defined %}
             <div class="AknMessageBox AknMessageBox--error">
-                {{ 'The username or email address "%username%" does not exist.'|trans({ '%username%': invalid_username }) }}
+                {{ 'An email has been sent to the user account.'|trans }}
```
pierallard: You have to update the translation please.

CHANGELOG-1.7.md
SamirBoulil: xD

SamirBoulil: Done

src/Pim/Bundle/UserBundle/Resources/views/Reset/request.html.twig
SamirBoulil: Changed the fix

BitOne:
I agree with @ahocquard  on the same message whatever happens.

PR #7736, opened 2018-02-26
PIM-7134: Clear repository during version purge

damien-carcel:
## Description

We currently have a memory leak on the version purger in MongoDB. This is probably caused by the same bug in the Doctrine ODM detach method that was already affecting the product, that was fixed in the PR [PIM-6995](https://github.com/akeneo/pim-community-dev/pull/7136).

The same fix was tried on versions [in this PR](https://github.com/akeneo/pim-community-dev/pull/7545), but saddly did not worked: the memory didn't increased, but the versions were no longer removed...

This new PR seems to fix the issue for good: the memory is stable around 110 MiB (tested with the data of the client that raised the bug), and versions are indeed deleted.

## Definition Of Done

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Pim/Bundle/VersioningBundle/Purger/VersionPurger.php
+++ b/src/Pim/Bundle/VersioningBundle/Purger/VersionPurger.php
@@ -89,6 +89,7 @@ class VersionPurger implements VersionPurgerInterface
                     $this->versionRemover->removeAll($versionsToPurge);
                     $this->objectDetacher->detachAll($versionsToPurge);
                     $versionsToPurge = [];
+                    $this->versionRepository->clear();
                 }
             } else {
                 $this->objectDetacher->detach($version);
@@ -146,8 +147,6 @@ class VersionPurger implements VersionPurgerInterface
      * Configure an option resolver with default option values
      *
      * @param OptionsResolver $optionResolver
-     *
-     * @return OptionsResolver
      */
     protected function configureOptions(OptionsResolver $optionResolver)
     {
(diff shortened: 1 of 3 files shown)


src/Pim/Bundle/VersioningBundle/Purger/VersionPurger.php
```diff
@@ -89,6 +89,7 @@ public function purge(array $options = [])
                     $this->versionRemover->removeAll($versionsToPurge);
                     $this->objectDetacher->detachAll($versionsToPurge);
                     $versionsToPurge = [];
+                    $this->versionRepository->clear();
```
willy-ahva: No phpspec to update ?

damien-carcel: Current spec is green. But you are right, I should add some unit tests (and if it is still green, that probably means that current ones are not good...).

damien-carcel: OK, so in fact, no, I cannot add a unit test for this case. The `clear` will be called only when the bulk limit of 1000 is reached. This is a little hard to achieve through a unit test, and the limit is set through a class constant, no DI, so no way to reconfigure it in the unit test.

SamirBoulil: Can't you foreach in the unit test  and ass dummy versions to purge so that you check you hit the clear ? I've seen this done before I think.

PR #7737, opened 2018-02-27
CLOUD-405/483/495/404: Move build step into a Dockerfile

ClementGautier:
# Description

This pull request move all the Build logic from the Jenkinsfile to the Dockerfile. We've merged every containers into one so it's easier to test and reproduce. Also it allows us to get rid of init-containers. We also added a shared library for DRY purpose.

# Benchmarks

| | Before | After | Gain |
| --- | --- | --- | --- |
| Enterprise | 58 min | 38 mins | 35% |
| Community | 41 mins | 33 mins | 20% |

# Definition Of Done (for Core Developer only)


.ci/Jenkinsfile
```diff
@@ -22,7 +20,7 @@ stage("Build") {
                 choice(choices: 'no\nyes', description: 'Enable Verbose mode', name: 'verboseOutputs'),
                 string(defaultValue: '', description: 'Channel or user to notify (example : "#channel,@user")', name: 'slackChannel'),
                 string(defaultValue: 'ee,ce', description: 'PIM edition the behat tests should run on (comma separated values)', name: 'editions'),
-                string(defaultValue: 'features,vendor/akeneo/pim-community-dev/features', description: 'Behat scenarios to build', name: 'features'),
+                string(defaultValue: 'features,tests/features,packages/pim-community-dev/features,packages/pim-community-dev/tests/features', description: 'Behat scenarios to build', name: 'features'),
```
damien-carcel: packages? Why `packages` and not `vendor`?

docker/vhost.conf
```diff
@@ -0,0 +1,27 @@
...
+    </Directory>
+
+    <Directory /var/www/pim/web/bundles>
+        RewriteEngine Off
+    </Directory>
+
+    ErrorLog /var/log/apache2/pim_error.log
+    CustomLog /var/log/apache2/pim_access.log combined
+
+</VirtualHost>
```
damien-carcel: Missing EOF

docker/php.ini
```diff
@@ -0,0 +1,6 @@
+date.timezone = Etc/UTC
+memory_limit = 2G
+upload_max_filesize = 20M
+post_max_size = 20M
+apc.enable_cli = On
+short_open_tag = Off
```
damien-carcel: Missing EOF.

bin/docker/start-servers
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
```
damien-carcel: The `bin/docker` folder is supposed to be for helper scripts for `docker-compose`, and `bin` in general is for the PIM itself. Imo, this kind of CI script should be in `.ci` (maybe a `.ci/bin`).

The same comment applies to the files `php.ini` and `vhost.conf` in the `docker` folder. I am not confortable mixing "CI stuff" with "PIM stuff". I'm afraid it could be confusing for developpers that see that for the first time.

.gcloudignore
```diff
@@ -0,0 +1,11 @@
...
+.git
+.gitignore
+.github
+.ci
+
+node_modules
+vendor
+upgrades
+
+*.md
```
damien-carcel: Missing EOF, please configure your editor ;-).

.gitignore
```diff
@@ -33,6 +33,7 @@ composer.phar
 !/bin/docker/pim-dependencies.sh
 !/bin/docker/pim-front.sh
 !/bin/docker/pim-initialize.sh
+!/bin/docker/start-servers
```
damien-carcel: Imo, to remove as those scripts should be in `.ci`.

.ci/Jenkinsfile
ClementGautier: We use a "path" vcs repository in order to fetch the right CE version through the EE without fetching a useless version of it beforehand. As it creates a symlink by default the pat was now packages instead of vendor as I was not able to put the repository inside the vendor directory itself. But, as it was misleading I guess I changed the behavior so it doesn't create a symlink but hard copy the files inside the vendor directory. This way it doesn't change this line ;)

ClementGautier: cf https://getcomposer.org/doc/05-repositories.md#path

bin/docker/start-servers
ClementGautier: I kind of agree but not entirely. This is not purely CI as the dockerfile could be used for dev purpose. But I guess you are right, having everything inside .ci is better for readability. I moved them

PR #7757, opened 2018-03-01
Fix #7752 wrong position of warning icon

### LINKED ISSUE #7752: Warning Icon not properly displayed if not code is given.
**I'm reporting a Bug**
In some edit attributes views, like "multi select" or "single select" if any code is given, a warning icon is displayed. A tooltip is displayed with the message "This value should not be blank". This warning is displayed in an other input where the value should be blank.  

image

AngelVazquezArroyo:
Changes on Less rules, to display properly the Warning Icon after a field is required.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Pim/Bundle/UIBundle/Resources/public/less/components/product-edit-form/FieldContainer.less
+++ b/src/Pim/Bundle/UIBundle/Resources/public/less/components/product-edit-form/FieldContainer.less
@@ -37,10 +37,7 @@
   }
 
   &-iconsContainer {
-    min-width: @AknIconSize + 4px;
     padding-top: 4px;
-    position: absolute;
-    right: -40px;
   }
 
   &-contextContainer {


AngelVazquezArroyo:
After Changes on Less rules, the Icon will look like on the example.
image

pierallard:
Hi @AngelVazquezArroyo !
Thanks for your contribution, we didn't see this bug! There is a little issue with your fix : the text in the input field says under the icon and is not visible.
Maybe a better solution should be to disable this absolute position when there is a warning icon on it, like this :
```
.AknFieldContainer-iconsContainer:not(:empty) {
    position: initial;
}
```
(From what I saw, this solution works, at least for this screen).

Best regards
Pierre

AngelVazquezArroyo:
Hi @pierallard  
If your solution is used, the position of .AknFieldContainer-iconsContainer will be position:initial, because The :empty pseudo-class represents an element that has no children at all. In this case under .AknFieldContainer-iconsContainer you have the following 
image

And you will habe allways an unnecessary space between code field and your next locale field. 
image

AngelVazquezArroyo:
Hi @pierallard 
Another solution could be, insert a new class on the div .AknFieldContainer-iconsContainer like .hasError and then in FieldContainer.less something like 
.hasError {
display: initial;
}

Best Regards
Angel

AngelVazquezArroyo:
Hi :) @pierallard @damien-carcel 

I've updated the PR and did other changes. After the changes on the less rules, the result is like on the following example.
image

pierallard:
Perfect !

phaseinducer:
Thanks Angel ; )

AngelVazquezArroyo:
Thx you all for the fast review!

amansilla:
Thanks to @flagbit as well for let us invest work time in fixing it.

PR #7761, opened 2018-03-02
Fix #7760 wrong HTTP status on response

### LINKED ISSUE #7760: Invalid HTTP status on case a exception is thrown
This is an issue which is affecting a controller for all Akeneo PIM versions from 1.6 to 2.2.

When deleting a option from a attribute of type "Single Choice" or "Multiple Choice" a invalid HTTP status will be sent due the exception code is used as HTTP response status.

amansilla:
Use the exception code as HTTP status is wrong since
will produce invalid HTTP responses e.g. a response where
the HTTP status is 0.

The HTTP status should be one of the defined in the following
class Symfony\Component\HttpFoundation\Response as constants.


damien-carcel:
Relates to #7760

amansilla:
Thanks to @flagbit as well for let us invest work time in fixing it.

PR #7768, opened 2018-03-05
Merge 1.6 into 1.7

jjanvier:


jjanvier:
I merge with a red CI.
Last build fail on EE because no Behat have been launched on those editions (which is normal).
Otherwise, Behats are green.

PR #7772, opened 2018-03-05
Fixtures builder instances should not be shared

phaseinducer:
This fixes a mistake introduced in https://github.com/akeneo/pim-community-dev/pull/7677.
Since builders are stateful and are declared as services, the instances are shared by default. This can lead to weird behaviors like values being shared between two calls.

A "better" solution would be to not declare them as services, but dependencies injection would be another problem at this point.
Relying on the DI is a not perfect but acceptable answer right now.


ahocquard:
Why the build function does not reset the state of a builder?

phaseinducer:
@ahocquard Could be a good solution too. It's the first iteration, we can improve that later.

PR #7774, opened 2018-03-05
PIM-7069: Fix Channel Export conversion units

anaelChardan:

The Channel Export was wrong regarding conversion_units, it converted like that

| conversion_units |
| KILOGRAM, DECIMETER |

instead of:

| conversion_units |
| weight:KILOGRAM, size:DECIMETER |

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Pim/Component/Connector/ArrayConverter/StandardToFlat/Channel.php
+++ b/src/Pim/Component/Connector/ArrayConverter/StandardToFlat/Channel.php
@@ -21,8 +21,14 @@ class Channel extends AbstractSimpleArrayConverter implements ArrayConverterInte
         switch ($property) {
             case 'locales':
             case 'currencies':
-            case 'conversion_units':
                 $convertedItem[$property] = implode(',', array_filter($data));
+                break;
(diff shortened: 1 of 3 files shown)


src/Pim/Component/Connector/ArrayConverter/StandardToFlat/Channel.php
```diff
@@ -21,9 +21,21 @@ protected function convertProperty($property, $data, array $convertedItem, array
...
+                    return array_map($function, array_keys($data), $data);
+                };
+
+                $itemToKeyValueString = function($key, $value) {
+                    return sprintf('%s:%s', trim($key), trim($value));
+                };
+
+                $formattedConvertedUnits = $array_mapk($itemToKeyValueString, array_filter($data));
+
+                $convertedItem[$property] = implode(',', $formattedConvertedUnits);
```
docteurklein:
```php

array_map(function($key) use ($data) {
    return sprintf('%s:%s', trim($key), trim($data[$key]));
}, array_keys(array_filter($data)));

PR #7781, opened 2018-03-06
[PIM-7203] Fix : bulk action mass edit attibute to edit all the product selection.

christophe-chausseray:

While we tried to edit an attribute on all the product, it was editing only some of the products selected. The method called to check if the family had the attribute was doing a contains on the doctrine collection. So it was checking the reference of the attribute instead of the attribute code.

I changed this condition to check with the attribute code. Also, I had to update the ProductAndProductModelReader to avoid an error thrown when mass editing a number of product greater than the batch size (ref: https://github.com/akeneo/pim-community-dev/pull/7565).

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/features/mass-action/mass_edit_many_products_at_once.feature
+++ b/features/mass-action/mass_edit_many_products_at_once.feature
@@ -59,3 +59,30 @@ Feature: Mass edit many products at once via a form
     And I should see the text "processed 345"
     And I should see the text "skipped 83"
     And the category of the product "product_1" should be "custom_category"
+
+  @jira https://akeneo.atlassian.net/browse/PIM-7203
+  Scenario: Edit an attribute on several products
+    Given the "default" catalog configuration
+    And the following attributes:
+      | code    | label-en_US | type              | group |
+      | name    | Name        | pim_catalog_text  | other |
+    And the following family:
+      | code     | attributes  | attribute_as_label |
+      | flipflop | name        | name               |
+    And 103 empty products for the "flipflop" family
+    And I am logged in as "Julia"
+    And I am on the products grid
+    When I sort by "ID" value ascending
+    And I select rows "product_1"
+    And I select all entities
+    And I press the "Bulk actions" button
+    And I choose the "Edit attributes" operation
+    And I move on to the choose step
+    And I choose the "Edit attributes" operation
+    And I display the Name attribute
+    And I change the "Name" to "Same product"
+    And I confirm mass edit
+    And I wait for the "edit_common_attributes" job to finish
+    Then I should see the text "COMPLETED"
+    And I should see the text "read 103"
+    And I should see the text "processed 103"
--- a/src/Pim/Bundle/EnrichBundle/Connector/Reader/MassEdit/ProductAndProductModelReader.php
+++ b/src/Pim/Bundle/EnrichBundle/Connector/Reader/MassEdit/ProductAndProductModelReader.php
@@ -44,6 +44,9 @@ class ProductAndProductModelReader implements
     /** @var CursorInterface */
     private $productsAndProductModels;
 
+    /** @var bool  */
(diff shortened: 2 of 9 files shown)


features/mass-action/mass_edit_many_products_at_once.feature
```diff
@@ -59,3 +59,31 @@ Feature: Mass edit many products at once via a form
...
+    And 103 empty products
+    And the following attributes:
+      | code    | label-en_US | type              | group |
+      | name    | Name        | pim_catalog_text  | other |
+    And the following family:
+      | code     | attributes  | attribute_as_label |
+      | flipflop | name        | name               |
+    And I am logged in as "Julia"
+    And I am on the products grid
+    Then I should see the text "103 results"
```
damien-carcel: useless check imo.

features/mass-action/mass_edit_many_products_at_once.feature
```diff
@@ -59,3 +59,31 @@ Feature: Mass edit many products at once via a form
...
+    And I sort by "ID" value ascending
```
damien-carcel: When

features/mass-action/mass_edit_many_products_at_once.feature
```diff
@@ -59,3 +59,31 @@ Feature: Mass edit many products at once via a form
...
     And the category of the product "product_1" should be "custom_category"
+
+  @jira https://akeneo.atlassian.net/browse/PIM-7203
+  Scenario: Edit an attribute on several products
+    Given the "default" catalog configuration
+    And 103 empty products
+    And the following attributes:
+      | code    | label-en_US | type              | group |
+      | name    | Name        | pim_catalog_text  | other |
+    And the following family:
```
willy-ahva: why do you add a family if you don"t put your products in ?

src/Pim/Bundle/EnrichBundle/Connector/Reader/MassEdit/ProductAndProductModelReader.php
```diff
@@ -44,6 +44,9 @@ class ProductAndProductModelReader implements
     /** @var CursorInterface */
     private $productsAndProductModels;
 
+    /** @var bool  */
+    private $firstRead = true;
```
willy-ahva: You didn't update specs about this new behaviour.

christophe-chausseray: done

src/Pim/Component/Catalog/EntityWithFamilyVariant/CheckAttributeEditable.php
```diff
@@ -33,7 +33,7 @@ public function isEditable(EntityWithFamilyInterface $entity, AttributeInterface
             return true;
         }
 
-        if (!$family->hasAttribute($attribute)) {
+        if (!$family->hasAttributeCode($attribute->getCode())) {
```
SamirBoulil: This is not enough as $family->hasAttribute is still harmful in the context of import export. You can do this fix, but mostly you need make the `hasAttribute` work :) we can see this together.

damien-carcel: As a quick, non breaking fix, I would try in a dedicated PR to make `hasAttribute` use `hasAttributeCode` internally, and deprecate `hasAttribute`.

SamirBoulil: 👍 yep, this is what I am expecting ^^

omolto:
+1

This is urgent to be solved. Bulk Edit is an important feature.

PR #5453, opened 2017-01-04
WIP Fix identifiers import duplication

jjanvier:
**Description (for Contributor and Core Developer)**

Introduces a new BulkIdentifierBagInterface responsible for keeping track of the entities identifiers processed on import (products, product associations and simple entities).

This object is hold by the Step execution context. When a processor wants to check if the entity it is processing has already been processed during the current batch, it just asks the BulkIdentifier if it knows it. 
if not, the entity identifier is added to the BulkIdentifier class and the processing process continues.

Depending on the type of identifier (simple string code, or composite key (array of string like Attribute options), the StepExecution might instanciate a BulkSimpleIdentifierBag class or a BulkCompositeIdentifierBag class.

The bulk is reset every time the step reach the batch size set (default value in the constructor to 100).
We will reset this list of identifiers once a full batch of entities is processed, starting from a clean state to the next batch.

**Definition Of Done (for Core Developer only)**

| Q                                 | A
| --------------------------------- | ---
| Added Specs                       | :white_check_mark:
| Added Behats                      | :white_check_mark:
| Changelog updated                 | :clock1:
| Review and 2 GTM                  | :clock1:
| Micro Demo to the PO (Story only) | :negative_squared_cross_mark:
| Migration script                  | :negative_squared_cross_mark:
| Tech Doc                          | :negative_squared_cross_mark::

:white_check_mark: Done and pass

:red_circle: Done but fail

:clock1: Done but pending

:negative_squared_cross_mark: Not needed

PR #7810, opened 2018-03-13
PIM-7237: fix integrity constraint violation during import

LaurentPetard:
**What's the problem?**

When there are two items with the same identifier(s) in an import file that don't exist yet in the PIM, and they're processed in the same batch, then an MySQL error is thrown. 

It doesn't concern products import. The process is different and the duplication problem already handled.
It's about all the import jobs processed in [this generic Processor](https://github.com/akeneo/pim-community-dev/blob/2.0/src/Pim/Component/Connector/Processor/Denormalization/Processor.php)

Indeed, the items are read and processed one by one, but they're written by batch (100 items by default). There's no problem if the duplicate items already exists, they are simply updated twice; but if they don't exist, two distinct entities are created and persisted at the same time with the same identifier(s).

**How to fix that?**

Functionally, the goal is to act as if the duplicated items already exist, and so the last values overwrite the previous ones. (seen with the POs)

The difficulty is to do it without doing any BC-break. 

The idea (thanks to @jjanvier who already worked on that problem, but with another functional goal in the PR #5453) is to keep in memory the created items in the execution context.

To avoid memory issues, the processed items kept in memory are removed between each batch. It's done with a new event and a subscriber. It's not perfect, so if you have another idea I'm a buyer!

It might seem disproportionate to add an event for that, but it will be used to fix the SLA PIM-7240 too (a memory link during import)

**Why did I change the batch behavior in `ItemStep`?**

Now the batch size affects reading and processing items, not only writing. It allows to avoid that the number of processed items kept in memory inflates too much if there are no items to write during a long time (because validation errors per instance). Moreover, it should solve the memory leak when there are too many warnings (for the same reasons)

The counterpart is that it could be more calls to the writer if there are a lot of invalid items in the import file, but it seems acceptable, because it should be uncommon.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/features/Pim/Behat/Context/Storage/AttributeOptionStorage.php
+++ b/features/Pim/Behat/Context/Storage/AttributeOptionStorage.php
@@ -1,0 +1,119 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Pim\Behat\Context\Storage;
+
+use Akeneo\Component\StorageUtils\Repository\IdentifiableObjectRepositoryInterface;
+use Behat\Behat\Context\Context;
+use Behat\Gherkin\Node\TableNode;
+use Doctrine\ORM\EntityManagerInterface;
+use PHPUnit\Framework\Assert;
+use Pim\Bundle\CatalogBundle\Entity\AttributeOption;
+use Pim\Component\Catalog\Model\AttributeOptionInterface;
+
+class AttributeOptionStorage implements Context
+{
+    /** @var IdentifiableObjectRepositoryInterface */
+    private $attributeOptionRepository;
+
+    /** @var EntityManagerInterface */
+    private $entityManager;
+
+    /**
+     * @param IdentifiableObjectRepositoryInterface $attributeOptionRepository
+     * @param EntityManagerInterface $entityManager
+     */
+    public function __construct(IdentifiableObjectRepositoryInterface $attributeOptionRepository, EntityManagerInterface $entityManager)
+    {
+        $this->attributeOptionRepository = $attributeOptionRepository;
+        $this->entityManager = $entityManager;
+    }
+
+    /**
+     * @param TableNode $table Must contains columns "attribute" and "code".
+     *
+     * @throws \Exception
+     *
+     * @Then /^there should be the following attribute options:$/
+     */
+    public function thereShouldBeTheFollowingAttributeOptions($table): void
+    {
+        $this->entityManager->clear();
+
+        foreach ($table->getHash() as $attributeOptionData) {
+            $attributeOption = $this->getAttribute($attributeOptionData);
+            unset($attributeOptionData['attribute']);
+            unset($attributeOptionData['code']);
+
+            foreach ($attributeOptionData as $property => $expectedValue) {
+                if (preg_match('/^label-(?<locale>.*)$/', $property, $matches)) {
+                    $this->assertAttributeOptionLabelEqual($expectedValue, $matches['locale'], $attributeOption);
+                } else {
+                    // Implement other properties if needed.
+                    throw new \Exception(sprintf('The property "%s" is not supported', $property));
+                }
+            }
+        }
+    }
+
+    /**
+     * @param array $attributeOptionData
+     *
+     * @throws \Exception If the attribute option does not exist
+     *
+     * @return AttributeOption
+     */
+    private function getAttribute(array $attributeOptionData): AttributeOption
+    {
+        if (!isset($attributeOptionData['attribute'])) {
+            throw new \Exception('You must give the attribute code in the column "attribute"');
+        }
+
+        if (!isset($attributeOptionData['code'])) {
+            throw new \Exception('You must give the option code in the column "code"');
+        }
+
+        $attributeOptionIdentifier = sprintf('%s.%s', $attributeOptionData['attribute'], $attributeOptionData['code']);
+        $attributeOption = $this->attributeOptionRepository->findOneByIdentifier($attributeOptionIdentifier);
+
+        if (null === $attributeOption) {
+            throw new \Exception(sprintf(
+                'There is no option "%s" for the attribute "%s"',
+                $attributeOptionData['code'],
+                $attributeOptionData['attribute']
+            ));
+        }
+
+        return $attributeOption;
+    }
+
+    /**
+     * @param string $expectedLabel
+     * @param string $locale
+     * @param AttributeOptionInterface $attributeOption
+     *
+     * @throws \Exception
+     */
+    private function assertAttributeOptionLabelEqual(string $expectedLabel, string $locale, AttributeOptionInterface $attributeOption): void
+    {
+        $attributeOption->setLocale($locale);
+        $optionValue = $attributeOption->getOptionValue();
+
+        Assert::assertNotNull($optionValue, sprintf(
+            'The option "%s" of the attribute "%s" has no label for the locale "%s"',
+            $attributeOption->getCode(),
+            $attributeOption->getAttribute()->getCode(),
+            $locale
+        ));
+
+        $label = $optionValue->getLabel();
+
+        Assert::assertEquals($expectedLabel, $label, sprintf(
+            'The label "%s" of the option "%s" of the attribute "%s" is not equal to the expected one',
+            $locale,
+            $attributeOption->getCode(),
+            $attributeOption->getAttribute()->getCode()
+        ));
+    }
+}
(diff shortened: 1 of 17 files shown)


src/Akeneo/Bundle/BatchBundle/EventListener/ResetProcessedItemsBatchSubscriber.php
```diff
@@ -0,0 +1,32 @@
...
+class ResetProcessedItemsBatchSubscriber implements EventSubscriberInterface
+{
+    public static function getSubscribedEvents()
+    {
+        return [
+            EventInterface::ITEM_STEP_AFTER_BATCH => 'reset',
+        ];
+    }
+
+    public function reset(StepExecutionEvent $event)
```
anaelChardan: Could you be a little bit more explicit, reset what? :) And type :void too please :)

src/Akeneo/Component/Batch/Step/ItemStep.php
```diff
@@ -114,21 +115,33 @@ public function doExecute(StepExecution $stepExecution)
                 continue;
             }
 
+            $batchCount++;
```
anaelChardan: $batchCount is for what, for read as it is updated after reading?

src/Akeneo/Component/Batch/Step/ItemStep.php
```diff
@@ -114,21 +115,33 @@ public function doExecute(StepExecution $stepExecution)
...
             }
         }
 
-        if (count($itemsToWrite) > 0) {
+        if (!empty($itemsToWrite)) {
             $this->write($itemsToWrite);
         }
+
+        if ($batchCount > 0) {
+            $this->dispatchStepExecutionEvent(EventInterface::ITEM_STEP_AFTER_BATCH, $stepExecution);
```
anaelChardan: Could it be line 139 ? (just a question)

src/Pim/Component/Connector/Processor/Denormalization/AbstractProcessor.php
```diff
@@ -57,21 +57,38 @@ public function setStepExecution(StepExecution $stepExecution)
...
+     * Get the identifier of a processed item
+     *
+     * @param IdentifiableObjectRepositoryInterface $repository
+     * @param array                                 $item
+     *
+     * @throws MissingIdentifierException if the processed item doesn't contain the identifier properties
+     *
+     * @return string
+     */
+    protected function getItemIdentifier(IdentifiableObjectRepositoryInterface $repository, array $item)
```
anaelChardan: It misses `:string` ;)

src/Akeneo/Component/Batch/Step/ItemStep.php
LaurentPetard: To count the number of items in the batch. these items are read, processed, but not necessarily written. Hard to find a more specific name, and it's related to the `batchSize` parameter

src/Akeneo/Component/Batch/Step/ItemStep.php
LaurentPetard: No, there may be items left in the batch but none of them have to be written (because they're invalid)

src/Akeneo/Component/Batch/Step/ItemStep.php
anaelChardan: Ok ;)

src/Akeneo/Component/Batch/Step/ItemStep.php
anaelChardan: Ok ;)

features/attribute/option/import_attribute_options.feature
```diff
@@ -33,3 +33,21 @@ Feature: Import attribute options
...
+      code;attribute;sort_order;label-en_US
+      new_red;color;0;New Red
+      new_red;color;0;Red is dead
+      """
+    And the following job "csv_footwear_option_import" configuration:
+      | filePath | %file to import% |
+    When I am on the "csv_footwear_option_import" import job page
+    And I launch the import job
+    And I wait for the "csv_footwear_option_import" job to finish
+    And I am on the "color" attribute page
```
damien-carcel: This verification part should not be done through the UI but directly with the database.

src/Akeneo/Bundle/BatchBundle/EventListener/ResetProcessedItemsBatchSubscriber.php
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * Reset the processed items saved into the job execution context after each batch executed during a step.
+ *
+ * @author    Laurent Petard <laurent.petard@akeneo.com>
+ * @copyright 2018 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+class ResetProcessedItemsBatchSubscriber implements EventSubscriberInterface
+{
+    public static function getSubscribedEvents()
```
damien-carcel: : array

src/Pim/Component/Connector/Processor/Denormalization/AbstractProcessor.php
```diff
@@ -1,5 +1,7 @@
 <?php
 
+declare(strict_types=1);
```
damien-carcel: If you add this, please typehint the methods too.

src/Pim/Component/Connector/Processor/Denormalization/Processor.php
```diff
@@ -74,24 +75,48 @@ public function process($item)
...
     }
 
     /**
-     * @param array $convertedItem
+     * @param string $itemIdentifier
      *
      * @return mixed
      */
-    protected function findOrCreateObject(array $convertedItem)
+    protected function findOrCreateObject(string $itemIdentifier)
```
damien-carcel: As you typehinted the string, please add a `declare(strict_type=1);` at the begining of the class and typhint the reste of the methods, or do not add typehint at all if you fear to break stuff.

LaurentPetard:
The changes on `ResetUniqueValidationSubscriber` will be done in #7815

features/Pim/Behat/Context/Storage/AttributeOptionStorage.php
```diff
@@ -0,0 +1,124 @@
...
+     * @param AttributeOptionInterface $attributeOption
+     *
+     * @throws \Exception
+     */
+    private function assertAttributeOptionLabelEqual(string $expectedLabel, string $locale, AttributeOptionInterface $attributeOption): void
+    {
+        $attributeOption->setLocale($locale);
+        $optionValue = $attributeOption->getOptionValue();
+
+        if (null === $optionValue) {
```
ahocquard: Personally, instead of big long exception, I prefer Assert::notNull, Assert::equals, etc.

features/attribute/option/import_attribute_options.feature
```diff
@@ -33,3 +33,20 @@ Feature: Import attribute options
     When I visit the "Product information" group
     And I change the "Manufacturer" to "[Converse]"
     Then I should see the text "[Converse]"
+
+  @jira https://akeneo.atlassian.net/browse/PIM-7237
+  Scenario: Successfully create new attribute options even if there are duplicates
```
ahocquard: `Successfully create new attribute options with duplicate options codes` (I don't like the "even")

src/Akeneo/Bundle/BatchBundle/EventListener/ResetProcessedItemsBatchSubscriber.php
```diff
@@ -0,0 +1,32 @@
...
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+
+/**
+ * Reset the processed items saved into the job execution context after each batch executed during a step.
+ *
+ * @author    Laurent Petard <laurent.petard@akeneo.com>
+ * @copyright 2018 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+class ResetProcessedItemsBatchSubscriber implements EventSubscriberInterface
```
ahocquard: Not in BatchBundle as it's our own PIM logic

src/Akeneo/Component/Batch/Item/ItemProcessorInterface.php
```diff
@@ -17,6 +17,8 @@
  */
 interface ItemProcessorInterface
 {
+    const PROCESSED_ITEMS_BATCH_CONTEXT_KEY = 'processed_items_batch';
```
ahocquard: It's a key dedicated to our own logic (putting object in the context), it should not be put in Batch bundle but in the Connector bundle.

LaurentPetard: Agreed. There's no interface that can bring this because it's too related to the implementation itself, so I removed the use of a constant.

PR #7812, opened 2018-03-14
PIM-7129: Fix the yaml writer

anaelChardan:

There were two problems:
- First one: 
  The file was never deleted as it was in FILE_APPEND mode, now the Writer is stateful (shame on me) because in case of a batch we need, first to overwrite the file and then append in it.

- Second one:
  During a batch, the header was written each time, but we had this kind of file:
  ```YAML
  rules:
      rule_1: coucou
      rule_100: plop
  rules:
     rule_101: ahahah
  ```
  So when we imported the file only a part of this one was imported, here is where the ugly stuff comes.
  The Yaml is a structured format so we do have to respect it, in order to avoid to repeat the header we had 3 solutions:
     - First one: Write only one time the file when all item would have been processed
           - Pro: The file is written one time only and it respects the structure
           - Cons: The RAM would have been overloaded for sure, the writer is stateful
     - Second one: Load the file, merge with the current batch chunk and overwrite the file
           - Pro: The structure is kept and we remove doubles, the writer is stateless
           - Cons: The RAM and access to the files systems and processing, we load 100, we write 200, we load 200, we write 300 an so one
     - Third one (the chosen one): We write into the file chunk by chunk but we have to remove the header and preserve the indentation to keep a valid YAML.
          - Pro: We do not overload the RAM and we keep the structure
          - Cons: We have to change the string  we write to add the offset indentation and the writer is stateful during a batch 

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Pim/Component/Connector/Writer/File/Yaml/Writer.php
+++ b/src/Pim/Component/Connector/Writer/File/Yaml/Writer.php
@@ -2,6 +2,7 @@
 
 namespace Pim\Component\Connector\Writer\File\Yaml;
 
+use Akeneo\Component\Batch\Item\FlushableInterface;
 use Akeneo\Component\Batch\Item\ItemWriterInterface;
 use Akeneo\Component\Batch\Job\RuntimeErrorException;
 use Akeneo\Component\Batch\Step\StepExecutionAwareInterface;
@@ -18,15 +19,20 @@ use Symfony\Component\Yaml\Yaml;
  */
 class Writer extends AbstractFileWriter implements
     ItemWriterInterface,
-    StepExecutionAwareInterface
+    StepExecutionAwareInterface,
+    FlushableInterface
 {
     const INLINE_ARRAY_LEVEL = 8;
+    const INDENT_SPACES = 4;
 
     /** @var ArrayConverterInterface */
     protected $arrayConverter;
 
     /** @var string */
     protected $header;
+
+    /** @var bool */
+    protected $isFirstWriting;
 
     /**
      * @param ArrayConverterInterface $arrayConverter
@@ -38,6 +44,7 @@ class Writer extends AbstractFileWriter implements
 
         $this->arrayConverter = $arrayConverter;
         $this->header = $header;
+        $this->isFirstWriting = true;
     }
 
     /**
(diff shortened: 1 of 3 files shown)


src/Pim/Component/Connector/Writer/File/Yaml/Writer.php
```diff
@@ -51,23 +58,75 @@ public function write(array $items)
...
-            $data[$this->header] = $flatItems;
-        }
 
         $path = $this->getPath();
         if (!is_dir(dirname($path))) {
             $this->localFs->mkdir(dirname($path));
         }
 
-        $yaml = Yaml::dump($data, self::INLINE_ARRAY_LEVEL);
+        if (!$this->shouldAppend) {
```
pierallard: I don't get at all what do this variable. Please take another naming!

src/Pim/Component/Connector/Writer/File/Yaml/Writer.php
```diff
@@ -51,23 +58,75 @@ public function write(array $items)
...
+     * @param string $path
+     *
+     * @return array
+     */
+    protected function appendWriting(array $items, string $path): array
+    {
+        $yaml = Yaml::dump($items, self::INLINE_ARRAY_LEVEL, self::INDENT_SPACES);
+
+        if (null !== $this->header) {
+            $yaml = preg_replace('/^(.*)/m', '    $1', $yaml);
```
docteurklein: maybe `preg_replace('/^/m', '    ', $yaml)` could work too (didn't try)

anaelChardan: Hum yeah, maybe, I'll try ;)

anaelChardan: it works, I update it ;)

src/Pim/Component/Connector/Writer/File/Yaml/Writer.php
LaurentPetard: Maybe rename and inverse value of `$this->shouldAppend` to something like `if ($this->isFirstWriting) {` ? Seems clearer and it avoids to use a negative condition

anaelChardan: done

src/Pim/Component/Connector/Writer/File/Yaml/Writer.php
```diff
@@ -51,23 +58,75 @@ public function write(array $items)
...
+        $this->isFirstWriting = true;
+    }
+
+    /**
+     * @param array $items
+     * @param string $path
+     *
+     * @return array
+     */
+    protected function overwriting(array $items, string $path): array
```
damien-carcel: `private function overwrite`

src/Pim/Component/Connector/Writer/File/Yaml/Writer.php
```diff
@@ -51,23 +58,75 @@ public function write(array $items)
...
+        return $data;
+    }
+
+    /**
+     * @param array $items
+     * @param string $path
+     *
+     * @return array
+     */
+    protected function appendWriting(array $items, string $path): array
```
damien-carcel: `private function append`

PR #7815, opened 2018-03-15
PIM-7240: Add subscriber to clear the cache between each job step batch

LaurentPetard:

See the SLA 7237 for more details. In a nutshell, when there are no writing during a product import (and when "Compare values" is enabled) then the cache of the repositories and the UnitOfWork are not cleared.

The goal of this PR is to delegating the responsibility of clearing the UOW and the repositories caches in a dedicated subscriber. 

I had to change `FilteredFamilyReader` because it loaded all the families at once. When the UOW is cleared, all the remaining families (in the next batch loop) were considered as new entities. 

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Pim/Bundle/CatalogBundle/EventSubscriber/ResetUniqueValidationSubscriber.php
+++ b/src/Pim/Bundle/CatalogBundle/EventSubscriber/ResetUniqueValidationSubscriber.php
@@ -8,7 +8,7 @@ use Pim\Component\Catalog\Validator\UniqueValuesSet;
 /**
  * The UniqueValueSet class is stateful, and used when you import several product, to check if in the product batch
  * there is no unique identifier issues or unique axis combination issues.
- * This listener listen the StorageEvents::POST_SAVE_ALL to reset the UniqueValueSet information, to be able to
+ * This listener listen the EventInterface::ITEM_STEP_AFTER_BATCH to reset the UniqueValueSet information, to be able to
  * work in another product batch without uniqueness issues.
  *
  * @author    Pierre Allard <pierre.allard@akeneo.com>
@@ -35,9 +35,9 @@ class ResetUniqueValidationSubscriber
 
     /**
      * Reset the Unique Value Set.
-     * Called on StorageEvents::POST_SAVE_ALL
+     * Called on EventInterface::ITEM_STEP_AFTER_BATCH
      */
-    public function onAkeneoStoragePostsaveall()
+    public function onAkeneoBatchItemStepAfterBatch()
     {
         $this->uniqueValueSet->reset();
         $this->uniqueAxesCombinationSet->reset();
--- a/src/Pim/Bundle/ConnectorBundle/EventListener/ClearBatchCacheSubscriber.php
+++ b/src/Pim/Bundle/ConnectorBundle/EventListener/ClearBatchCacheSubscriber.php
@@ -1,0 +1,45 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Pim\Bundle\ConnectorBundle\EventListener;
+
+use Akeneo\Component\Batch\Event\EventInterface;
+use Akeneo\Component\StorageUtils\Cache\CacheClearerInterface;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+
+/**
+ * Clear the internal cache between each batch during the execution of an item step.
+ *
+ * @author    Laurent Petard <laurent.petard@akeneo.com>
+ * @copyright 2018 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+class ClearBatchCacheSubscriber implements EventSubscriberInterface
+{
+    /** @var CacheClearerInterface */
+    private $cacheClearer;
+
+    /**
+     * @param CacheClearerInterface $cacheClearer
+     */
+    public function __construct(CacheClearerInterface $cacheClearer)
+    {
+        $this->cacheClearer = $cacheClearer;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public static function getSubscribedEvents(): array
+    {
+        return [
+            EventInterface::ITEM_STEP_AFTER_BATCH => 'clearCache',
+        ];
+    }
+
+    public function clearCache(): void
+    {
+        $this->cacheClearer->clear();
+    }
+}
--- a/src/Pim/Bundle/EnrichBundle/Connector/Reader/MassEdit/FilteredFamilyReader.php
+++ b/src/Pim/Bundle/EnrichBundle/Connector/Reader/MassEdit/FilteredFamilyReader.php
@@ -19,9 +19,6 @@ class FilteredFamilyReader implements ItemReaderInterface, StepExecutionAwareInt
     /** @var StepExecution */
     protected $stepExecution;
 
-    /** @var bool */
-    protected $isExecuted;
-
     /** @var ArrayCollection */
     protected $families;
 
@@ -34,7 +31,6 @@ class FilteredFamilyReader implements ItemReaderInterface, StepExecutionAwareInt
     public function __construct(FamilyRepositoryInterface $familyRepository)
     {
         $this->familyRepository = $familyRepository;
-        $this->isExecuted = false;
     }
 
     /**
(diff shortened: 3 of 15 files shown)


oliverde8:
I am quite happy seeing this PR, but not having this issue in the 2.0 is in my point of view a blocker.

We are working on a migration from 1.7 to 2.0 for one of our clients and during our tests we stumbled upon the issue(we actually reported it for the first time in december). We put in place a lot of custom code so migrating every 3 month is not an option. We will therefore use the 2.0 version of Akeneo which is a Long term support.

I can understand that this proper fix can't be merged in the 2.0 due to the BC, but a fix is needed in the 2.0 version. 

Turning around the pot, akeneo 2.0 was far from ready, each new release we can see it getting closer of what should have been there from start....

src/Pim/Bundle/CatalogBundle/EventSubscriber/ResetUniqueValidationSubscriber.php
```diff
@@ -35,9 +35,9 @@ public function __construct(UniqueValuesSet $uniqueValueSet, UniqueAxesCombinati
 
     /**
      * Reset the Unique Value Set.
-     * Called on StorageEvents::POST_SAVE_ALL
+     * Called from ItemStep::doExecute
      */
-    public function onAkeneoStoragePostsaveall()
+    public function onAkeneoBatchItemStepAfterBatch()
```
ahocquard: 👍

ahocquard:
@damien-carcel what is a big BC break ? the fact that we clear the UOW for every jobs ?

src/Akeneo/Bundle/BatchBundle/EventListener/ClearBatchCacheSubscriber.php
```diff
@@ -0,0 +1,42 @@
...
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+
+/**
+ * Clear the internal cache between each batch during the execution of an item step.
+ *
+ * @author    Laurent Petard <laurent.petard@akeneo.com>
+ * @copyright 2018 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+class ClearBatchCacheSubscriber implements EventSubscriberInterface
```
ahocquard: It should not be in the bath bundle but in the connector bundle

damien-carcel:
@ahocquard The fact that we completely redefine the arguments of some services.
But in fact, I realize the BC is easily avoidable: just don't use the cache clearer in the processors/writers, only in the new subscriber, but don't remove them from the services in this patch, only on master.

src/Pim/Bundle/ConnectorBundle/EventListener/ClearBatchCacheSubscriber.php
```diff
@@ -0,0 +1,42 @@
...
+ * Clear the internal cache between each batch during the execution of an item step.
+ *
+ * @author    Laurent Petard <laurent.petard@akeneo.com>
+ * @copyright 2018 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+class ClearBatchCacheSubscriber implements EventSubscriberInterface
+{
+    /** @var CacheClearerInterface */
+    protected $cacheClearer;
```
damien-carcel: private

src/Pim/Bundle/ConnectorBundle/EventListener/ClearBatchCacheSubscriber.php
```diff
@@ -0,0 +1,42 @@
...
+
+    /**
+     * @param CacheClearerInterface $cacheClearer
+     */
+    public function __construct(CacheClearerInterface $cacheClearer)
```
damien-carcel: : void

src/Pim/Bundle/ConnectorBundle/EventListener/ClearBatchCacheSubscriber.php
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    public static function getSubscribedEvents()
+    {
+        return [
+            EventInterface::ITEM_STEP_AFTER_BATCH => 'clearCache',
+        ];
+    }
+
+    public function clearCache(): void
```
damien-carcel: : array

Please add PHPDoc too (yes I know it is arguable since PHP 7.1, but as we still do it everywhere else...)

src/Pim/Component/Connector/Writer/Database/ProductModelDescendantsWriter.php
```diff
@@ -30,6 +30,8 @@ class ProductModelDescendantsWriter implements ItemWriterInterface, StepExecutio
     /**
      * @param SaverInterface        $descendantsSaver
      * @param CacheClearerInterface $cacheClearer
+     *
+     * @merge Remove $cacheClearer. It's not used anymore.
```
damien-carcel: :+1:

damien-carcel: Maybe "@todo [MERGE]" or something like that, because the @todo are detected by PHPStorm, so it can ease the process.

src/Pim/Bundle/ConnectorBundle/EventListener/ClearBatchCacheSubscriber.php
LaurentPetard: Not on the constructor

src/Pim/Bundle/ConnectorBundle/EventListener/ClearBatchCacheSubscriber.php
LaurentPetard: Nothing more to write on the PHPDoc of `clearCache` that is already wrote in the PHPDoc of the class (this subscriber will not have another methods)

damien-carcel: Arf, yes you're right ><

src/Pim/Component/Connector/Writer/Database/ProductModelDescendantsWriter.php
ahocquard: We definitely have to find a convention for it but people seems not agree on that point 😠

src/Pim/Bundle/EnrichBundle/Connector/Reader/MassEdit/FilteredFamilyReader.php
```diff
@@ -87,7 +81,13 @@ protected function getFamilies(array $filters)
 
         $familiesIds = $filter['value'];
 
-        return new ArrayCollection($this->familyRepository->findByIds($familiesIds));
+        foreach ($familiesIds as $familyId) {
+            $family = $this->familyRepository->find($familyId);
```
ahocquard: so, one request per family ? we should do it per batch (you easily do that with an array_chunk)

LaurentPetard: I don't think that it will have a noticeable impact on the performances. I would be agree for the products, but It's only for the families and for a mass edit from the UI. Do it per batch requires to known the batch size in `FilteredFamilyReader` and it's maybe not easy without doing a BC-break

ahocquard: Imho, you have already a batch size of 1 implicitly hard coded by requesting with the `find` function.

So, I would just do a private const with a batch size. No BC, costs nothing, works exactly as before, faster.

src/Pim/Bundle/ConnectorBundle/EventListener/ClearBatchCacheSubscriber.php
damien-carcel: Right, sorry.

src/Pim/Bundle/EnrichBundle/Connector/Reader/MassEdit/FilteredFamilyReader.php
LaurentPetard: No, it has to be the same batch size that in `ItemStep`, else there will be detached entities when the UOW will be cleared.

PR #7824, opened 2018-03-16
CLOUD-577: Split integration tests

ClementGautier:
# Description

This pull request make use of the new version of the Jenkins shared library that allows a split of phpunit tests per test and not per file. This should fix an issue on continuous-integration : https://github.com/akeneo/continuous-integration/issues/309


ahocquard:
By doing it, you will increase the average run time of a single test from 1,5 sec to 13 sec because we have a mysldump cache and it's very efficient.

I think it will destroy the performance and it's really overkill to have a dedicated environment for each test.
It would be far less costly to improve the test stack instead of doing this kind of workaround.

PR #7837, opened 2018-03-20
Migration: Add migration to remove the product type from product table

SamirBoulil:

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/upgrades/schema/Version_2_2_20180320093100_remove_product_type.php
+++ b/upgrades/schema/Version_2_2_20180320093100_remove_product_type.php
@@ -1,0 +1,28 @@
+<?php
+
+namespace Pim\Upgrade\Schema;
+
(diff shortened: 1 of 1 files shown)


upgrades/schema/Version_2_2_20180320093100_remove_product_type.php
```diff
@@ -0,0 +1,28 @@
...
+    {
+        $this->addSql('ALTER TABLE pim_catalog_product DROP product_type');
+    }
+
+    /**
+     * @param Schema $schema
+     */
+    public function down(Schema $schema)
+    {
+        $this->throwIrreversibleMigrationException();
```
docteurklein: is it really irreversible, or could we imagine putting all the product to "simple" in case of a rollback?

PR #7844, opened 2018-03-20
Reduce slowness due to ES pre-checks

anaelChardan:

This PR removes a lot of checks following this one https://github.com/akeneo/pim-community-dev/pull/7821 which removed 
- Check category existence before using them in ES

This PR removes as well:
- Price: Check that currency exists and is active before filtering on them
- Ancestor: A ProductModel Id exists before filtering on them
- Family : Check that families exists before filtering on them
- Group: Check that groups exists before filtering on them
- Parent: Check that a ProductModel id exists before filtering on them
- SelfAndAncestorFilter: Check that a ProductModel id and a Product exist before filtering on them
- WithAttributes: Check that the attributes exists before filtering on them

We did that in order to have feedback on the API in case of querying a non-existent things.
As it is not documented we remove it to  have better performance and to be consistent with the previous PR I made.

The consequencies it is that it will return nothing in case of a wrong queryied value (only in API use case as the other use cases didn't used it)

This impacts the PQB so PO may be aware of that, Nolwenn is Ok for the API.
But import / **rules** etc may be impacted.

In CE the possible points of failure are (I tried to remove the API impacts/ datagrid impacts / and filter where we check before trying to apply a filter) :
- RemoveProductCommand in Catalog
- GroupUpdater
- FilterDataValidator.

| Micro Demo to the PO (Story only) | OK for Nono

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/features/family/remove_attribute_from_a_family.feature
+++ b/features/family/remove_attribute_from_a_family.feature
@@ -8,29 +8,19 @@ Feature: Remove attribute from a family
     Given the "catalog_modeling" catalog configuration
     And I am logged in as "Peter"
 
-  Scenario: Successfully remove an attribute from a family and display it as removable from product
-    Given I am on the "accessories" family page
-    And I visit the "Attributes" tab
-    Then I should see attributes "EAN" in group "ERP"
-    When I remove the "material" attribute
-    And I save the family
-    And I should not see the text "There are unsaved changes."
-    Then I should see the flash message "Attribute successfully removed from the family"
-    When I am on the "1111111292" product page
-    Then I should not see the text "Material"
-
-  Scenario: Successfully remove an attribute from a family and it does not appear in the variant product product model edit pages
+  Scenario: Successfully remove an attribute from a family and assert its removal from product, product-model, variant-product
     Given I am on the "accessories" family page
     And I visit the "Attributes" tab
     When I remove the "material" attribute
     And I save the family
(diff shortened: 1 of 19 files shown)


features/family/remove_attribute_from_a_family.feature
```diff
@@ -8,29 +8,19 @@ Feature: Remove attribute from a family
     Given the "catalog_modeling" catalog configuration
     And I am logged in as "Peter"
 
-  Scenario: Successfully remove an attribute from a family and display it as removable from product
+  Scenario: Successfully remove an attribute from a family and assert its removal from product, product-model, variant-product
```
anaelChardan: It uses the UI only to delete
The check is now done through repository

jjanvier:
I know I'll be really annoying, but the title of your PR shouldn't be "Reduce slowness due to ES pre-checks".
Why? Simply because with this title I don't know if it's something the product owners are aware of or if if it's just a the wish of a developer.

Also, please explain the consequences of this PR (are there consequences on the API?, on the imports?).

anaelChardan:
It was explained(at least some parts) in the PR referenced in the beginning ;)

jjanvier:
In the PR referenced in the beginning, there is no explanation of what functionalities are impacted by this change. So please, I'll ask it again, list all the functional consequences of this change. And make sure the POs are aware of those changes. I really doubt only the API is impacted.

PR #7853, opened 2018-03-21
Less UI in Behat imports

jjanvier:
To follow https://github.com/akeneo/pim-community-dev/pull/7618.

~65 import Behats were I removed the useless usage of the UI.
It's doable. It took me only a few hours.
If each people took 1 hour to do the same, it would be finished in a few weeks. 
I'm available for help if somebody wants to do the same.

### DIFF
--- a/features/Pim/Behat/Context/ImportFileContext.php
+++ b/features/Pim/Behat/Context/ImportFileContext.php
@@ -36,7 +36,7 @@ final class ImportFileContext extends PimContext implements SnippetAcceptingCont
     private const USERNAME_FOR_JOB_LAUNCH = 'admin';
 
     /**
-     * @When the :entities are imported via the job :jobName
+     * @When /^the (.*) are imported via the job ([\w\_]+)$/
      */
     public function entitiesAreImportedViaTheJob($entities, $jobName)
     {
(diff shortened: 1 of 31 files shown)


features/Pim/Behat/Context/ImportFileContext.php
```diff
@@ -36,15 +36,15 @@
     private const USERNAME_FOR_JOB_LAUNCH = 'admin';
 
     /**
-     * @When the :entities are imported via the job :jobName
+     * @When /^the (.*) are imported via the job ([\w\_]+)$/
```
damien-carcel: What is the reason to prefer regular expressions to the `:entities` notation.

aRn0D: yes, this is weird!

anaelChardan: And `.*` is very open, if you want to replace the mask pattern by a regex you should find a more restrictive one ;)

jjanvier: @damien-carcel @aRn0D because some of our entities contains two word like `attribute group`. I was not able to catch them with the turninp syntax. If you have a solution I change.

@anaelChardan yeah... why?

anaelChardan: Just to avoid some bad surprises ;), But if it works and you are aware that it works perfectly well GTM, it was just a guard ;)

ahocquard: You can surrounding it with double quotes otherwise. I think it works (https://github.com/jnicklas/turnip#placeholders)

jjanvier: yes it works, it's just that quotes mean something
when you write, you don't use quotes randomly in the middle of your sentences

PR #7875, opened 2018-03-26
PIM-7241: Fix memory consumption issue when archiving an import file

LaurentPetard:

To archive an import file, its content is loaded in memory using `file_get_contents`. Guess what happens when the file is big...

To fix that I used `putStream` instead of `put` and simply open the import file instead of loading its content. It's not perfect because we shouldn't use any PHP function on a file like that, but it would require too much changements for a simple patch.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Pim/Component/Connector/Archiver/FileReaderArchiver.php
+++ b/src/Pim/Component/Connector/Archiver/FileReaderArchiver.php
@@ -48,13 +48,19 @@ class FileReaderArchiver extends AbstractFilesystemArchiver
             if ($this->isReaderUsable($reader)) {
                 $jobParameters = $jobExecution->getJobParameters();
                 $filePath = $jobParameters->get('filePath');
-                $key = strtr(
+                $archivePath = strtr(
                     $this->getRelativeArchivePath($jobExecution),
                     [
                         '%filename%' => basename($filePath),
                     ]
                 );
-                $this->filesystem->put($key, file_get_contents($filePath));
+
+                $fileResource = fopen($filePath, 'r');
+                $this->filesystem->putStream($archivePath, $fileResource);
+
+                if (is_resource($fileResource)) {
+                    fclose($fileResource);
+                }
             }
         }
     }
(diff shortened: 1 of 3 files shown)


src/Pim/Component/Connector/Archiver/FileReaderArchiver.php
```diff
@@ -48,13 +48,13 @@ public function archive(JobExecution $jobExecution)
...
                 $filePath = $jobParameters->get('filePath');
-                $key = strtr(
+                $archivePath = strtr(
                     $this->getRelativeArchivePath($jobExecution),
                     [
                         '%filename%' => basename($filePath),
                     ]
                 );
-                $this->filesystem->put($key, file_get_contents($filePath));
+                $this->filesystem->putStream($archivePath, fopen($filePath, 'r'));
```
anaelChardan: Does it need a fclose afterward?

src/Pim/Component/Connector/Archiver/FileReaderArchiver.php
```diff
@@ -48,13 +48,19 @@ public function archive(JobExecution $jobExecution)
...
+                $archivePath = strtr(
                     $this->getRelativeArchivePath($jobExecution),
                     [
                         '%filename%' => basename($filePath),
                     ]
                 );
-                $this->filesystem->put($key, file_get_contents($filePath));
+
+                $fileResource = fopen($filePath, 'r');
+                $this->filesystem->putStream($archivePath, $fileResource);
```
anaelChardan: I know that, it ws note here before but it can returns false in case of failure, do we have to take care of it?

LaurentPetard: Yes, I hesitated to add a check. In the absolute it can return false, but if the file is not readable an error will be thrown before. Moreover the second argument of `putStream` is not typed so it's its responsibility to check that (It's done in the implementation of League/FlySystem)

PR #7878, opened 2018-03-26
PIM-6825: fix cancel button redirection on user profile edit

momoss:

Cancel button were always redirecting on Admin user profile, now if you press cancel button when editing a user, it redirect you on the users grid

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/CHANGELOG-1.7.md
+++ b/CHANGELOG-1.7.md
@@ -2,6 +2,7 @@
 
 ## Bug fixes
 
+- PIM-6825: Fix cancel button redirection when editing a user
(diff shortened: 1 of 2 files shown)


anaelChardan:
No test?

jjanvier:
a test for that is imo useless
if we could unit test it in JS, OK
but if we need to add a Behat for that :-1:

momoss:
I don't want to add a behat for that kind of fix

anaelChardan:
Yes, I totally agree that a behat for that is totally overkill ;), but that was a question ;)

PR #7879, opened 2018-03-26
PIM-7259: Remove only attribute's option removed for multi select attribute

skeleton:
When an attribute option was removed and if this option was linked to a value inside a product, instead of just remove this option from values, we were throwing an exception without create values with good options. 

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Pim/Component/Catalog/Exception/InvalidOptionsException.php
+++ b/src/Pim/Component/Catalog/Exception/InvalidOptionsException.php
@@ -1,0 +1,88 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Pim\Component\Catalog\Exception;
+
+use Akeneo\Component\StorageUtils\Exception\InvalidPropertyException;
+
+/**
+ * Exception thrown when performing an action on a property with invalid options.
+ *
+ * @copyright 2018 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+class InvalidOptionsException extends InvalidPropertyException
+{
+    const VALID_ENTITY_CODE_EXPECTED_CODES = 306;
+
+    /** @var array */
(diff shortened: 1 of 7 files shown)


src/Pim/Component/Catalog/Exception/InvalidOptionsException.php
```diff
@@ -0,0 +1,73 @@
...
+     *
+     * @param string $propertyName
+     * @param string $key
+     * @param string $because
+     * @param string $className
+     * @param array  $values
+     *
+     * @return InvalidOptionsException
+     */
+    public static function validEntityListCodesExpected(string $propertyName, string $key, string $because, string $className, array $values)
```
anaelChardan: : InvalidOptionsException

src/Pim/Component/Catalog/Exception/InvalidOptionsException.php
```diff
@@ -0,0 +1,73 @@
+<?php
+
```
anaelChardan: declare(strict-types=1);

src/Pim/Component/Catalog/Exception/InvalidOptionsException.php
```diff
@@ -0,0 +1,80 @@
...
+            $className,
+            sprintf($message, $propertyName, $key, $because, $flatValues),
+            self::VALID_ENTITY_CODE_EXPECTED_CODES
+        );
+    }
+
+    /**
+     * @return array
+     */
+    public function getPropertyValues(): array
```
aRn0D: This getter break some principle like "Tell don't ask!". If you want to print these codes, you should add a `toString()` method. If you want to know if the value of a data is valid you should add a method `isValid($values)`.

skeleton: it was more to respect the same logic of other exceptions, but ok

PR #7884, opened 2018-03-26
PIM-7252: add option to do not apply permissions on attribute settings page

skeleton:
Today, we use the same route to view an attribute in the PEF (which is in enrich part) and to view an attribute in the settings part. It's a mistake because the behavior is not the same. We need to apply permissions on the enrich part, but not in the settings part. 
As this subject is quite complex and should be rethink/rework, I made the same logic than all other routes (because the same mistake is done on family for instance), that's means: add an option in the URL to define if permissions have to be apply 
 

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/CHANGELOG-2.0.md
+++ b/CHANGELOG-2.0.md
@@ -9,6 +9,7 @@
 - PIM-6945: Fix missing header elements on identifier attribute edit form
 - PIM-7259: Remove only attribute's option removed for multi select attribute
 - PIM-7270: Fix [object Object] empty filter value
+- PIM-7252: Fix permissions on attributes settings page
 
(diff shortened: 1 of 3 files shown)


CHANGELOG-2.0.md
```diff
@@ -4,6 +4,7 @@
 
 - PIM-7240: Add subscriber to clear the cache between each job step batch
 - PIM-7241: Fix memory consumption issue when archiving an import file
+- PIM-7252: Do not call filters on attributes settings page
```
phaseinducer: The visible effect for the end user is more fixing the attribute edit page.

PR #7889, opened 2018-03-27
PIM-7253: do not schedule queries to update normalized data for newly created entities

docteurklein:
…  as they cannot appear in the normalizedData anyway.

see https://github.com/akeneo/ElasticSearchBundle/pull/163


jmleroux:
@docteurklein ,
What about updated not wen entities but updated ones?
If an entity (like attribute option) is updated, are you sure we don't need to schedule an update?

PR #7895, opened 2018-03-27
Use native docker .env and override feature instead of docker-compose.yml.dist

jmleroux:
A docker lovers PR.
When using `docker-compose.yml.dist` , devs must keep their own `docker-compose.yml` synchronized.
If they miss a modification, they could not be aware of it.

Adding `docker-compose.yml` to the git history ensure that devs will always use the "official" docker configuration. And they will still be able to customize their installation with the native `override` feature.

The corresponding [documentation PR is here](https://github.com/akeneo/pim-docs/pull/894)

| Tech Doc                          | Todo (link to the pim-docs PR)

`OK`: Done / Validated
`-`: Not needed


anaelChardan:
Good Job JM ;-) In my opinion I don’t know if I would prefer an Override file or to use environment variables for this kind of thing that are only configuration ;-). Maybe we should use both ;-)

damien-carcel:
I'm still not sure what is the best for this use case between using docker override or just environment variables in a `.env` file (which is also a native feature of docker-compose). Btw, both solutions are not exclusive and could be mixed.

In any case, whatever the final choice, imo a documentation PR linked to this PR is mandatory to explain the usage of override.

damien-carcel:
And I just saw @anaelChardan did almost the exact same comment :-D.

docker-compose.yml
```diff
@@ -46,10 +40,6 @@ services:
     image: httpd:2.4
     depends_on:
       - fpm
-    environment:
-      PHP_IDE_CONFIG: 'serverName=pim-ce'
-    ports:
-      - '8080:80'
     volumes:
```
docteurklein: volumes could be moved to override as well.

jmleroux:
> both solutions are not exclusive and could be mixed

Yes!

damien-carcel:
I agree with you @docteurklein, it could be merged as a first step.

BUT I insist, it must be merged with a documentation PR.

Also, I would remove the `docker-compose.override.yml.dist` file. Just add the non dist one in `.gitignore` and document its usage.

docker-compose.yml
jmleroux: > volumes could be moved to override as well.

Not agree. On my point of view, the `docker-compose.` should be sufficient for a working PIM installation, and the override would be there only for custom values.

jmleroux:
> In my opinion I don’t know if I would prefer an Override file or to use environment variables

@anaelChardan , speking of ports exposed, I prefer the override. This way i can tune my ports on multiple running applications without having to play with env var naming.
With env vars, how would you set the exposed ports if you want to PIMs in the same time? ;)

jmleroux:
@damien-carcel , could you give me the link to the doc repo section?

anaelChardan:
@jmleroux your environments variables should be terminal closed  or if you use tmux ;) so with that you can run multiple applications ;)

damien-carcel:
@jmleroux Doc about Docker install is here: https://github.com/akeneo/pim-docs/blob/master/install_pim/docker/installation_docker.rst

docker-compose.yml
damien-carcel: Nope indeed, without those volumes, you have not PIM anymore.

docker-compose.yml
```diff
@@ -46,10 +40,6 @@ services:
     image: httpd:2.4
     depends_on:
       - fpm
-    environment:
-      PHP_IDE_CONFIG: 'serverName=pim-ce'
```
damien-carcel: I would not remove this environment variable

docker-compose.yml
```diff
@@ -6,7 +6,7 @@ services:
     environment:
       BEHAT_TMPDIR: '/srv/pim/var/cache/tmp'
       COMPOSER_HOME: '/home/docker/.composer'
-    user: docker
```
damien-carcel: Nope, if your remove that, the CLI will generate cache as root. This one is mandatory.

docker-compose.yml
```diff
@@ -7,10 +7,7 @@ services:
       BEHAT_TMPDIR: '/srv/pim/var/cache/tmp'
       COMPOSER_HOME: '/home/docker/.composer'
```
damien-carcel: This one should be in the override too, as it can be different on some systems.
For example, on my machine (and most recent linux distro), I have 2 folders: `.cache/composer` and `.config/composer`.
Plus, an install can work without it.

Same should be done for Yarn cache.

damien-carcel:
This PR will have to be replicated in enterprise-dev, and in standard editions.

docker-compose.yml
jmleroux: Yep, but in our context, we know exactly what will be the image used and what is the COMPOSER_HOME in the container. That's why i choose to keep them in the docker-compose file.

jmleroux:
I added the `.env` file wich can be mixed with the override file.
The two values defined in the `.env` are mandatory to access the PIM UI (well, maybe not the behat one but it's a detail). And it can be easily overriden in the shell.
The other will be in the override file because they are totally optional.

What do you think of it?

damien-carcel:
Indeed using an .env file allows more flexibility, but the problem is you must define the environment variables used in the compose file. You cannot just use an override if you want to.
If you don't declare the variable, you'll have an error: `services.httpd.ports is invalid: Port ranges don't match in length`.

So in the end, we are still stuck with a .dist that need to be filled. I think there is no real perfect solution here anyway. This is already way better than previously.

PR #7902, opened 2018-03-29
PIM-7267: Fix boolean attributes not added to variant product

LaurentPetard:

To avoid to handle "three-values" boolean, it had been decided to add default values for boolean attributes when a product is created.

The problem is that for a variant product, all the boolean attributes of the family are added, while only the attributes of the related level of the variant family should be added.

To fix that I choose to extract the boolean adding in a dedicated subscriber listening the event `StorageEvents::PRE_SAVE` . It couldn't be done anymore in `ProductBuilder` because when `ProductBuilder::create` is called for a variant product, the parent is not yet known (it's done in the updater).

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Pim/Bundle/CatalogBundle/EventSubscriber/AddBooleanValuesToNewProductSubscriber.php
+++ b/src/Pim/Bundle/CatalogBundle/EventSubscriber/AddBooleanValuesToNewProductSubscriber.php
@@ -1,0 +1,136 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Pim\Bundle\CatalogBundle\EventSubscriber;
+
+use Akeneo\Component\StorageUtils\StorageEvents;
+use Pim\Component\Catalog\AttributeTypes;
+use Pim\Component\Catalog\Factory\ValueFactory;
+use Pim\Component\Catalog\Manager\AttributeValuesResolverInterface;
+use Pim\Component\Catalog\Model\FamilyInterface;
+use Pim\Component\Catalog\Model\FamilyVariantInterface;
+use Pim\Component\Catalog\Model\ProductInterface;
+use Pim\Component\Catalog\Model\ProductModelInterface;
+use Pim\Component\Catalog\Model\VariantAttributeSetInterface;
+use Pim\Component\Catalog\Model\VariantProductInterface;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\EventDispatcher\GenericEvent;
+
+/**
+ * When a new product is created, and before it is saved.
+ * Add a product value "false" for every boolean attributes of the product's family.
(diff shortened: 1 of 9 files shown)


grena:
Once this PR will be merged, we'll have to merge the command to clean values : https://github.com/akeneo/pim-community-dev/pull/7891

src/Pim/Bundle/CatalogBundle/EventSubscriber/AddBooleanValuesToNewProductSubscriber.php
```diff
@@ -0,0 +1,135 @@
...
+        $variationLevel = $product->getVariationLevel();
+        $variantAttributeSet = $familyVariant->getVariantAttributeSet($variationLevel);
+        if (!$variantAttributeSet instanceof VariantAttributeSetInterface) {
+            return [];
+        }
+
+        $attributes = $variantAttributeSet->getAttributes()->toArray();
+        $axes = $familyVariant->getAxes();
+
+        return array_filter($attributes, function ($attribute) use ($axes) {
```
anaelChardan: You can type $attribute in the callback ;)

anaelChardan: You could have use yield to to have the same way than the other method ;)

src/Pim/Bundle/CatalogBundle/EventSubscriber/AddBooleanValuesToNewProductSubscriber.php
```diff
@@ -0,0 +1,135 @@
...
+     */
+    public function addBooleansDefaultValues(GenericEvent $event): void
+    {
+        $product = $event->getSubject();
+
+        if (!$product instanceof ProductInterface || null !== $product->getId()) {
+            return;
+        }
+
+        $booleanAttributes = $product instanceof VariantProductInterface
```
damien-carcel: To be merged carefully in 2.2, need to be replaced by `$product->isVariant()`.

src/Pim/Bundle/CatalogBundle/EventSubscriber/AddBooleanValuesToNewProductSubscriber.php
LaurentPetard: Indeed... I don't remember at what moment I remove the generator...

src/Pim/Bundle/CatalogBundle/EventSubscriber/AddBooleanValuesToNewProductSubscriber.php
LaurentPetard: Yes, I moved the comment closer

PR #7911, opened 2018-03-30
PIM-7260: Add validation in the 'configure' step of mass edits

SamirBoulil:
Prior to this PR, there was missing validation when configuring a mass-edit.

For instance, in a "change family" mass-edit we didn't check if the user actually choose a family before running the mass-edit job.

This PR adds those checks for the mass-edits of categories (remove, move, add), groups, edit attributes, family and add to existing product model.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Pim/Bundle/UIBundle/Resources/public/less/pages/Default.less
+++ b/src/Pim/Bundle/UIBundle/Resources/public/less/pages/Default.less
@@ -25,7 +25,7 @@
 
   &-flashContainer {
     position: fixed;
-    z-index: 15;
+    z-index: 910;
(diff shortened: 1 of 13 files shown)


src/Pim/Bundle/UIBundle/Resources/public/less/pages/Default.less
```diff
@@ -25,7 +25,7 @@
 
   &-flashContainer {
     position: fixed;
-    z-index: 15;
+    z-index: 910;
```
SamirBoulil: @pierallard I've set this to appear above the mass-edit page( which z-index is 900)

PR #7928, opened 2018-04-03
PIM-7263: extract command's logic into a dedicated service

grena:

This is a rework to extract logic of the command into a dedicated service. This is the same work than the backport on 2.0 : https://github.com/akeneo/pim-community-dev/pull/7891

| Added legacy Behats               | N
| Added acceptance tests            | N


src/Pim/Bundle/CatalogBundle/Command/Cleaner/WrongBooleanValuesOnVariantProductCleaner.php
```diff
@@ -0,0 +1,73 @@
...
+     * Return true if the product has been modified, false otherwise
+     *
+     * @param ProductInterface $variantProduct
+     *
+     * @return bool
+     */
+    public function cleanProduct(ProductInterface $variantProduct): bool
+    {
+        $isModified = false;
+
```
damien-carcel: I would add an early return here if `$variantProduct->isVariant()` is false. In 2.0, the typehint of the method prevented this case to happen. Not the case here anymore, so anyone could missuse this service.

src/Pim/Component/Catalog/spec/Updater/WrongBooleanValuesOnVariantProductUpdaterSpec.php
```diff
@@ -0,0 +1,115 @@
+<?php
```
christophe-chausseray: This one could be removed as now we have the WrongBooleanValuesOnVariantProductCleanerSpec.php

src/Pim/Bundle/CatalogBundle/Command/Cleaner/WrongBooleanValuesOnVariantProductCleaner.php
christophe-chausseray: The check is already done there : https://github.com/akeneo/pim-community-dev/pull/7928/commits/84faa7c9c39719124977dbbba638a3d05beb5893#diff-7e00d45aa78c74e5d86d6db40f4d5316R62

grena: Nice catch, I add it.

PR #7929, opened 2018-04-04
PIM-7246 Get volume for every entities with fake queries

LaurentPetard:

This PR add the "in memory" count query for the tests, on it complete the acceptance tests for each monitored entity.

The case of the volumes without limit is handled by setting the limit at a negative value. 

`OK`: Done / Validated
`-`: Not needed


src/Pim/Bundle/CatalogVolumeMonitoringBundle/tests/Acceptance/Context/CategoryContext.php
```diff
@@ -0,0 +1,80 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Pim\Bundle\CatalogVolumeMonitoringBundle\tests\Acceptance\Context;
+
+use Behat\Behat\Context\Context;
+use Pim\Bundle\CatalogVolumeMonitoringBundle\tests\Acceptance\Persistence\Query\InMemory\InMemoryCountQuery;
```
aRn0D: Why don't you create a port and an adapter ?

src/Pim/Bundle/CatalogVolumeMonitoringBundle/tests/Acceptance/Context/ReportContext.php
```diff
@@ -0,0 +1,41 @@
...
+     */
+    public function theAdministratorUserAsksForTheCatalogVolumeMonitoringReport(): void
+    {
+        $this->volumes = $this->volumesNormalizer->volumes();
+    }
+
+    /**
+     * @return array
+     */
+    public function getVolumes(): array
```
aRn0D: Why this context has a getter ?

LaurentPetard: We need to share  the normalisation of the volumes for each context, triggered by the same Gherkin sentence. This is not ideal, but instead of doing a different sentence for each type of volume, or create an artificial service only used in the contexts, I preferred to add a `ReportContext` that is a dependency in the other context.

src/Pim/Bundle/CatalogVolumeMonitoringBundle/tests/Acceptance/Context/CategoryContext.php
LaurentPetard: Because the "InMemory" queries have methods that are not in the interface (It has been already discussed in this PR : https://github.com/akeneo/pim-community-dev/pull/7860 ). It's not ideal, but I think this is acceptable for acceptance tests

PR #7943, opened 2018-04-06
End up with this shitty day (and fix master branch...)

jjanvier:
First I tried to merge 2.3 into master. 
Problem is that I merged that into 2.3 (and not into master). Let's call this unfortunate operation `commit A`.
Then I revert commit A on the branch 2.3. Let's call this correcting operation `commit B`.

OK so now 2.3 is back to normal.
But hey, I still want to merge 2.3 into master.

So I checkout master and 2.3 locally. They're up to date. 2.3 contains commits `A` and `B`.
I prepare the branch able to merge, and finally merge it proudly on master. Let's call this operation `commit C`.

Master now contains commits `A`, `B` and `C`. Problem is that I don't want `B` in master. `B` reverts all the changes I previously wanted! 

This PR is about to remove this commit `B` from master (and master only).


jjanvier:
Doesn't work `UserEvents` is missing :(

PR #7960, opened 2018-04-10
Merge 2.3 into master

jjanvier:
The `JenkinsFile` (and CI way to launch builds) has completely changed between 2.3 and master.
@tamarasaurus does the JenkinsFile here seems good to you? Did you add only acceptance tests?


docker-compose.yml.dist
```diff
@@ -21,7 +21,7 @@ services:
       - behat
 
   node:
-    image: node:8
+    image: juliensnz/node
```
damien-carcel: ?

tamarasaurus: @julien made a custom node image that includes the install for puppeteer and cucumber

damien-carcel: If maintained in time, it could be interesting to add it in the akeneo Dockerfiles.

PR #7969, opened 2018-04-12
Merge 2.2 into 2.3

jmleroux:


src/Pim/Bundle/EnrichBundle/Resources/public/js/controller/attribute/edit.js
```diff
@@ -36,38 +36,41 @@ function (
...
+                        attribute.code
+                    )
+                );
 
-                    PageTitle.set({'attribute.label': label});
+                PageTitle.set({'attribute.label': label});
 
-                    return FormBuilder.getFormMeta('pim-attribute-edit-form')
+                    return FormBuilder.getFormMeta(
+                        'pim-attribute-edit-form' )
```
tamarasaurus: Here you can make this into one line 
`FormBuilder.getFormMeta('pim-attribute-edit-form')` and remove the space before the last parenthesis. 
Then run `yarn run lint` to check that it was fixed.

PR #7976, opened 2018-04-12
PIM-7043 : Fix to add the last page into the pagination only if the total records is under 10000.

christophe-chausseray:

When we have more than 10 000 products on the product grid, we couldn't access to the last page.  We got an Elasticsearch error because it has a limitation of 10000 records set by the "max_rescore_window" parameter.

So to solve this problem, we decided to remove the button to access to the last page in the pagination. We know that it doesn't solve completely the problem. The client will get the problem if he want to display the 10000th product and more. However,  he has to go at least on the 400th page to see this error. So we admit that he will probably make a search or filter these products instead to use the pagination.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -73,10 +73,6 @@ stage("Build") {
                     sh "mkdir -m 777 vendor"
 
                     container("php") {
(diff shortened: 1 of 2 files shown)


LaureBro:
changelog please.. :(

PR #6204, opened 2017-05-22
PIM 6429 : Improve the loading of the completeness widget on dashboard

ahocquard:
[//]: <> (<3 Thanks for taking the time to contribute! You're awesome! <3)

[//]: <> (If you've never contributed to this repository before, please read https://github.com/akeneo/pim-community-dev/blob/master/.github/CONTRIBUTING.md)

**Description (for Contributor and Core Developer)**

The widget of the completeness could take a lot of time to be displayed (200 seconds for a client).
It is due to an unoptimised SQL request.
It's a big problem, because during this time, the session is locked. So, the PIM is totally unavailable for the user (even with a refresh).

This problem has already been raised in this ticket: https://github.com/akeneo/pim-community-dev/issues/3771

**1. First point that this PR intends to resolve: [ABORTED]**

_IMPORTANT : This commit will be done on master due to race conditions on the CI.
It needs more investigation and it probably implies more stuff to do, so I prefer to not take the risk to apply it in a patch._

Failing scenarios:

- features/proposal/notification_preferences.feature:40
- features/permissions/category/enforce_no_permissions_for_category.feature:15
- vendor/akeneo/pim-community-dev/features/acl/enforce_acl_on_history.feature:7
- features/permissions/category/define_permissions_for_category.feature:22
- vendor/akeneo/pim-community-dev/features/category/display_category_history.feature:7
- features/localization/proposal/display_localized_numbers_in_proposals.feature:29  

A request should not block all the other concurrent requests by locking the session, except if that session is used in a controller.

For that, a listener is in charge to close the session before calling a controller. It applies for every routes, and so, it fixes the problem for every potential long running request.
If the controller is using the session (flashbag for example), the session handler will re-open it automatically, so don't worry :)

Impact : a completeness SQL request that is taking 200 sec to be calculated will not prevent the UI to work (tested and validated) by locking the session. 
It can have a real benefit in term of performance for the whole UI of the PIM.

Note: this solution has been discussed with @BitOne and @jjanvier and it is a first step to improve our session management.

**2. Second point that this PR intends to resolve:**

The original request to calculate the completeness is slow due to the DISTINCT in the count.

```
    SELECT ch.label, lo.code as locale, COUNT(DISTINCT co.product_id) as total FROM pim_catalog_channel ch
    JOIN %category_table% ca ON ca.root = ch.category_id
    JOIN %category_join_table% cp ON cp.category_id = ca.id
    JOIN %product_table% p ON p.id = cp.product_id
[...]

PR #7987, opened 2018-04-16
Improve the performance of the loading of the completeness widget

ahocquard:
I already did a PR almost one year ago:
https://github.com/akeneo/pim-community-dev/pull/6204

But I found a better optimisation.

**Before**
7sec on the generated catalog "80%".

**After**
70ms on the generated catalog "80%".

**For fun: before the PIM-6429 (original query)**
1m27 on the generated catalog "80%"

**How**
Delete the `LEFT JOIN` on the completeness table. It was useless and only here in case there is 0 product completed. An INNER JOIN would delete the row and not display the couple channel/locale if we don't change anything. 

So, the trick is to do all the stuff in a subrequest with a JOIN and then doing the LEFT JOIN on the aggregated SUM (which is almost free).

Performance depends of the distributivity of the data. The more you have complete products, the longer this it is to calculate. But it is normal and was the case before.

**Bonus**
It fixes a potential SQL injection as well (I didn't find a way to exploit, so don't worry).

`OK`: Done / Validated
`-`: Not needed


ahocquard:
Closed in favor of a far better solution, using ES.

PR #7990, opened 2018-04-16
PIM-7304 : Fix the SKU filter to be able to be scrollable

christophe-chausseray:

When we add a long list of values in the SKU filters, we can't see the update button anymore. This is because the popup is too big compare to the window.

So to solved that, I had to fix the max height and add the scroll on the popup. Also, I have to update the position of the popup when we add a value in the select2.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Pim/Bundle/UIBundle/Resources/public/less/lib/select2.less
+++ b/src/Pim/Bundle/UIBundle/Resources/public/less/lib/select2.less
@@ -491,6 +491,14 @@
   }
 }
 
+//override for select2 choice filters
+.AknFilterChoice {
+  .select2-choices {
+    overflow: auto;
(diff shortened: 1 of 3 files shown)


src/Pim/Bundle/UIBundle/Resources/public/less/lib/select2.less
```diff
@@ -491,6 +491,14 @@
   }
 }
 
+//override for select2 choice filters
+.AknFilterChoice {
+  .select2-choices {
+    overflow: auto;
+    max-height: 250px;
```
pierallard: You can increase a little bit this value IMHO (like 500px)... Ask for Stéphane, he will give you the right number.

PR #7998, opened 2018-04-17
[SLA] PIM-7313: prevent the same job to be executed by two different daemons due to race condition

ahocquard:

A same job can executed by two different daemons, with obvious consequences:
- concurrent INSERT on primary key
- job failing
- etc

In theory, it's impossible to execute the same job by two different daemons, due to a locking strategy that I implemented with the database. 

But it occurs when you have a lot of daemons (for example 10, probability to have the bug increases with the number of daemons). 
Also, it has more chance to occur when you have a lot of "fast" job in the queue, such as "akeneo_batch_job_execution_queue".

The problem is due to the locking strategy that was not robust enough, allowing a race condition to consume a job two times.

**Explication of the race condition:**

- Expected behavior:

| Daemon 1                                 | Daemon 2
| Ask for unconsumed job                       | Ask for unconsumed job
| Answer job 1              | Answer job 1
| Ask lock for job 1           | Ask lock for job 1
| Locked           | Not locked, already lock by process 1
| update job with consumer UUID : job 1 is not considered as unconsumed anymore               | -
| Release lock                  | Ask for another unconsumed job
| execute job 1                 | Answer job 2

But this scenario can occur as well:

| Daemon 1                                 | Daemon 2
| Ask for unconsumed job   | Ask for unconsumed job
| Answer job 1              | Answer job 1
| Ask lock for job 1           | -
| Locked           | -
| update job with consumer UUID : job 1 is not considered as unconsumed anymore               | -
| Release lock                  | -
| **execute job 1**                 | -
| -           | Ask lock for job 1
| -           | Not locked anymore by daemon 1: let's lock it!  
| -           | update job with consumer UUID (and override daemon UUID of process 1)
| -                  | Release lock
| -                 | **execute job 1**

And 🎉, you have two times the same job running.

**How to prevent this**

There are multiple strategies when we talk about locking and queue with Mysql.

1. `SELECT FOR UPDATE`

First one is to use `SELECT FOR UPDATE`. 
But it's dangerous, because you can have what we call `gaps lock` when isolation level is "REPEATBLE READ" (by default on mysql). The consequence is potential dead lock on the table.

https://www.percona.com/blog/2012/03/27/innodbs-gap-locks/
https://www.percona.com/blog/2013/12/12/one-more-innodb-gap-lock-to-avoid/

And that's why Symfony change the isolation level for the session.

https://github.com/symfony/symfony/blob/master/src/Symfony/Component/HttpFoundation/Session/Storage/Handler/PdoSessionHandler.php#L551

2. `GET_LOCK`

You can get a lock on mysql, shared between all the session/connection with a simple `SELECT GET_LOCK("your_lock_name")`.
It's powerful. It works as expected. No gaps lock.

The drawback is that you don't really lock any row (= message of the queue) as with `SELECT FOR UPDATE`.

So, the logic of the lock (the name of the lock, how you lock it, etc) is in your application.
Meaning that if you have multiple applications sharing the same database (ERP is a good example), they have to implement the same logic in the code base to get the lock on a message.

As sharing the database across multiple application is a kind of anti-pattern, and not the case of the PIM, it was the solution I choosed.

It means as well that if you don't lock it correctly, you can have bugs.
And it is the reason of this bug.

It's possible to do a working queue with this lock, and it would be possible to fix the bug with this method. But I found a far better and cleaner solution.

3. `UPDATE` and count row

It's the easier solution to fix this bug and to implement a queue.
You lock a row by updating the message. You can get the number of updated rows with PDO/DBAL. If no row is updated, it means a concurrent consumer has already consumed the message.
So the daemon poll for the next message.

It works as an UPDATE is atomic by using an exclusive lock on the updated row. 
Less code. Less complex. It's a win-win.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Akeneo/Bundle/BatchQueueBundle/Queue/JobExecutionMessageRepository.php
+++ b/src/Akeneo/Bundle/BatchQueueBundle/Queue/JobExecutionMessageRepository.php
@@ -60,9 +60,14 @@ SQL;
     }
 
     /**
+     * Update a job execution message.
+     * Only an unconsumed job execution message can be updated.
+     *
      * @param JobExecutionMessage $jobExecutionMessage
+     *
+     * @return bool return whether the job has been updated or not
      */
-    public function updateJobExecutionMessage(JobExecutionMessage $jobExecutionMessage)
+    public function updateJobExecutionMessage(JobExecutionMessage $jobExecutionMessage): bool
     {
         $sql = <<<SQL
 UPDATE 
@@ -74,7 +79,8 @@ SET
     q.create_time = :create_time,
     q.updated_time = :updated_time
 WHERE
-    q.id = :id;
+    q.id = :id
+    AND q.consumer IS NULL;
 SQL;
 
         $stmt = $this->entityManager->getConnection()->prepare($sql);
@@ -85,6 +91,8 @@ SQL;
         $stmt->bindValue('updated_time', new \DateTime('now', new \DateTimeZone('UTC')), Type::DATETIME);
         $stmt->bindValue('id', $jobExecutionMessage->getId());
         $stmt->execute();
+
+        return $stmt->rowCount() > 0;
     }
 
     /**
--- a/CHANGELOG-2.0.md
+++ b/CHANGELOG-2.0.md
@@ -6,6 +6,7 @@
 - PIM-7282: Add validation on the code value during the attribute creation to prohibit the "entity_type" value.
 - PIM-7043: Remove the last pagination button if it has more than 10 000 products on the product grid.
 - PIM-7299: Add pagination for family variants on several screens
+- PIM-7313: prevent the same job to be executed by two different daemons due to race condition
 
 ## BC Breaks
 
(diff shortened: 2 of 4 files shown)


src/Akeneo/Bundle/BatchQueueBundle/Queue/JobExecutionMessageRepository.php
```diff
@@ -60,9 +60,14 @@ public function createJobExecutionMessage(JobExecutionMessage $jobExecutionMessa
...
     /**
+     * Update a job execution message.
+     * Only an unconsumed job execution message can be updated.
+     *
      * @param JobExecutionMessage $jobExecutionMessage
+     *
+     * @return bool return whether the job has been updated or not
      */
-    public function updateJobExecutionMessage(JobExecutionMessage $jobExecutionMessage)
+    public function updateJobExecutionMessage(JobExecutionMessage $jobExecutionMessage): bool
```
ahocquard: not a BC in my opinion, as it's an internal class without interface and it should not be extended (and properties are private).

BitOne:
Did you check that the transaction isolation is not undoing what you are trying to do?
On two different transactions, you will be able to update the same rows until the commit happens.

Curious about that; One of the main interest of the LOCK is that its scope is global, whereas an update, the scope is in the transaction.

ahocquard:
@BitOne you try to get an exclusive lock when you are trying to update a row.
Exclusive lock, by definition, can not be got by two different rows.

https://dev.mysql.com/doc/refman/5.7/en/innodb-locking.html

> If a transaction T1 holds an exclusive (X) lock on row r, a request from some distinct transaction T2 for a lock of either type on r cannot be granted immediately. Instead, transaction T2 has to wait for transaction T1 to release its lock on row r.

ahocquard:
Note: after talking IRL, GTM by Benoit

PR #8002, opened 2018-04-18
PIM-7312: Fix attribute requirements update for a newly created channel

mmetayer:

This PR fixes an issue when trying to add an attribute requirement for a family, and the attribute requirement already exists but its `required` property is set to `false` (which is the case for newly created channels)

`OK`: Done / Validated
`-`: Not needed


mmetayer:
@fitn as seen together, we still have to perform that check in the FamilyUpdater, in case an requirement is set to `false` another way (e.g custom dev)

PR #8003, opened 2018-04-18
Add an in memory attribute related repository

damien-carcel:
## Description

- Attribute repository
- Attribute option repository
- Attribute group repository

For each are implemented `getIdentifierProperties`, `findOneByIdentifier`, `findBy`, and `save` methods.
Every implemented methods are unit tested.

## Definition Of Done

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/tests/back/Acceptance/Resources/config/pim/repositories.yml
+++ b/tests/back/Acceptance/Resources/config/pim/repositories.yml
@@ -7,4 +7,6 @@ parameters:
     pim_catalog.repository.family_variant.class: Akeneo\Test\Acceptance\FamilyVariant\InMemoryFamilyVariantRepository
     pim_catalog.repository.product.class: Akeneo\Test\Acceptance\Product\InMemoryProductRepository
     pim_catalog.repository.product_model.class: Akeneo\Test\Acceptance\ProductModel\InMemoryProductModelRepository
+    pim_catalog.repository.attribute.class: Akeneo\Test\Acceptance\Attribute\InMemoryAttributeRepository
     pim_catalog.repository.attribute_option.class: Akeneo\Test\Acceptance\AttributeOption\InMemoryAttributeOptionRepository
+    pim_catalog.repository.attribute_group.class: Akeneo\Test\Acceptance\AttributeGroup\InMemoryAttributeGroupRepository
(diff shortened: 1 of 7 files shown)


tests/back/Acceptance/Resources/config/pim/repositories.yml
```diff
@@ -3,3 +3,4 @@ parameters:
     pim_catalog.repository.channel.class: Akeneo\Test\Acceptance\Channel\InMemoryChannelRepository
     pim_catalog.repository.category.class: Akeneo\Test\Acceptance\Category\InMemoryCategoryRepository
     pim_catalog.repository.currency.class: Akeneo\Test\Acceptance\Currency\InMemoryCurrencyRepository
+    pim_catalog.repository.attribute_option.class: Akeneo\Test\Acceptance\AttributeOption\InMemoryAttributeOptionRepository
```
skeleton: missing InMemoryAttributeRepository

tests/back/Acceptance/spec/Attribute/InMemoryAttributeRepositorySpec.php
```diff
@@ -0,0 +1,49 @@
...
+
+namespace spec\Akeneo\Test\Acceptance\Attribute;
+
+use PhpSpec\ObjectBehavior;
+use Pim\Bundle\CatalogBundle\Entity\Attribute;
+use Pim\Component\Catalog\Model\AttributeInterface;
+
+class InMemoryAttributeRepositorySpec extends ObjectBehavior
+{
+    function it_returns_an_identifier_property()
```
SamirBoulil: We added the "it_is_a_attribute_repository_interface" as well as "it_is_a_saver_interface".

tests/back/Acceptance/spec/Attribute/InMemoryAttributeRepositorySpec.php
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    function it_finds_attributes_by_criteria()
+    {
+        $attribute = $this->createAttribute('attribute_1');
+        $this->beConstructedWith([$attribute->getCode() => $attribute]);
+
+        $this->findBy(['code' => 'attribute_1'])->shouldReturn([$attribute]);
+    }
+
```
SamirBoulil: What about the case where it does not find ?

tests/back/Acceptance/spec/Attribute/InMemoryAttributeRepositorySpec.php
jjanvier: we did too

damien-carcel: I can add it if you really want, but Matthias aggrees it adds no value to the unit test. And it is what we are supposed to write here: unit tests, even if we are using a spec tool do to it.

We tests the behavior of the method `save`, so from the test point of view, we don't care it implements the interface or not.

tests/back/Acceptance/spec/Attribute/InMemoryAttributeRepositorySpec.php
damien-carcel: Added.

PR #8006, opened 2018-04-18
Clean Gherkin/features directories

jjanvier:
The goal of this PR is to move Gherkin files into proper folders.
Now we have:
- `tests/features` for all new Gherkin that could lead to a frontend acceptance test, backend acceptance test and end to end test
- `tests/legacy/features` for all our old Gherkin, which are end to end tests (with or without UI)


.ci/Jenkinsfile
```diff
@@ -22,7 +22,7 @@ stage("Build") {
                 choice(choices: 'no\nyes', description: 'Enable Verbose mode', name: 'verboseOutputs'),
                 string(defaultValue: '', description: 'Channel or user to notify (example : "#channel,@user")', name: 'slackChannel'),
                 string(defaultValue: 'ee,ce', description: 'PIM edition the behat tests should run on (comma separated values)', name: 'editions'),
-                string(defaultValue: 'features,vendor/akeneo/pim-community-dev/features', description: 'Behat scenarios to build', name: 'features'),
+                string(defaultValue: 'tests/legacy/features,vendor/akeneo/pim-community-dev/tests/legacy/features', description: 'Behat scenarios to build', name: 'features'),
```
skeleton: you don't add `tests/features`?

anaelChardan: You're right, we need both ;)

jjanvier: @skeleton actually no, we don't need it
because here "features" means "end to end" tests
the acceptance tests don't use this parameter

.ci/behat.community.yml
```diff
@@ -2,6 +2,7 @@ legacy:
     suites:
         all:
             paths:
+                - '%paths.base%/tests/legacy/features/'
```
jjanvier: to remove

.ci/behat.enterprise.yml
```diff
@@ -2,6 +2,7 @@ default:
     suites:
         all:
             paths:
+                - '%paths.base%/tests/legacy/features/'
```
jjanvier: to remove

.ci/Jenkinsfile
```diff
@@ -52,8 +52,10 @@ stage("Build") {
             }
             container("docker") {
                 sh "cp .ci/behat.community.yml behat.yml"
+                String featureEscaped = ""
                 for (feature in features) {
-                    sh "sed -i \"/paths/a\\ \\ \\ \\ \\ \\ \\ \\ \\ \\ \\ \\ \\ \\ \\ \\ -\\ ${feature}\" behat.yml"
+                    featureEscaped = feature.replaceAll("/", "\\\\/");
```
jjanvier: This fixes a bug.
Before the PR, all occurrences of `path` were replace by the features file provided by the input of the build.
And we want to use this input only for the `legacy` profile, which is the first profile.

.ci/behat.community.yml
```diff
@@ -143,6 +144,7 @@ acceptance:
         FriendsOfBehat\ContextServiceExtension:
            imports:
                - "tests/back/Acceptance/Resources/config/behat/services.yml"
+               - "src/Pim/Bundle/CatalogVolumeMonitoringBundle/tests/Acceptance/Resources/config/behat/services.yml"
```
jjanvier: Missing acceptance services.
I don't understand how it could work before...

.ci/behat.enterprise.yml
```diff
@@ -142,18 +142,20 @@ acceptance:
...
         acceptance:
             paths:
-                - '%paths.base%/tests/back/features/'
-                - '%paths.base%/vendor/akeneo/pim-community-dev/tests/back/features/'
+                - '%paths.base%/tests/features/'
+                - '%paths.base%/vendor/akeneo/pim-community-dev/tests/features/'
             contexts_services:
                 - test.locale.context
                 - test.channel.context
+                - test.catalog_volume_limits.attribute_per_family_context
```
jjanvier: Missing acceptance services.
I don't understand how it could work before...

.ci/behat.enterprise.yml
```diff
@@ -142,18 +142,20 @@ acceptance:
...
             filters:
                 tags: '@acceptance-back'
     extensions:
         FriendsOfBehat\ContextServiceExtension:
            imports:
                - "tests/back/Acceptance/Resources/config/behat/services.yml"
                - "vendor/akeneo/pim-community-dev/tests/back/Acceptance/Resources/config/behat/services.yml"
+               - "vendor/akeneo/pim-community-dev/src/Pim/Bundle/CatalogVolumeMonitoringBundle/tests/Acceptance/Resources/config/behat/services.yml"
```
jjanvier: Missing acceptance services.
I don't understand how it could work before...

behat.yml.dist
```diff
@@ -115,12 +117,14 @@ legacy:
             factory:
                 page_parameters:
                     base_url: http://akeneo-pim-behat.local/
+        Pim\Behat\Extension\PimFormatter\PimFormatterExtension: ~
```
jjanvier: I added those extensions so that we have the same extensions running on the CI and locally.

behat.yml.dist
```diff
@@ -136,3 +140,5 @@ acceptance:
         FriendsOfBehat\SymfonyExtension:
             kernel:
                 env: acceptance
+        Pim\Behat\Extension\PimFormatter\PimFormatterExtension: ~
```
jjanvier: I added those extensions so that we have the same extensions running on the CI and locally.

composer.json
```diff
@@ -10,12 +10,18 @@
...
-            "Context": "features/"
+            "": "src/"
         },
         "psr-4": {
-            "Pim\\Upgrade\\": "upgrades/",
-            "Pim\\Behat\\": "features/Pim/Behat",
+            "Pim\\Upgrade\\": "upgrades/"
+        }
+    },
+    "autoload-dev": {
```
jjanvier: Fixing the autoload to require only what's useful for production.

behat.yml.dist
skeleton: what do we need this extension?

behat.yml.dist
```diff
@@ -136,3 +140,5 @@ acceptance:
...
+        Liuggio\Fastest\Behat\ListFeaturesExtension\Extension: ~
```
LaurentPetard: Added twice. Could be fixed in .ci/behat.community.yml too

.ci/behat.enterprise.yml
LaurentPetard: It didn't. It should be fix in https://github.com/akeneo/pim-community-dev/pull/7929 but it has been delayed too much...

behat.yml.dist
jjanvier: @LaurentPetard it has been added to two different profiles
extensions are registered per profile

tests/legacy/features/Context/Page/Dashboard/Index.php
```diff
@@ -18,6 +18,24 @@ class Index extends Base
      */
     protected $path = '#/';
 
+    /**
+     * {@inheritdoc}
+     */
+    public function __construct($session, $pageFactory, $parameters = [])
```
jjanvier: dirty fix for the EE Behat that are failing (related to proposals on the dashboard)
the problem is that the class `Context\Page\Dashboard\Index` exists in both CE and EE
and now, with the proper autoloading of the dev class, only the CE class is loaded (which is logical, we can't have 2 different classes with the same FQCN...)
the proper way to fix it is to change all the namespaces of the EE edition for the Behat classes

anaelChardan:
GG JJ ;)

PR #8007, opened 2018-04-18
Run legacy & acceptance tests on CI

skeleton:

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -286,28 +286,16 @@ stage("Test") {
             "behat-ee": {
                 if (launchBehatTests.equals("yes") && editions.contains("ee")) {
                     queue({
-                        def scenarios = sh (returnStdout: true, script: "cd /home/jenkins/pim && php vendor/bin/behat --list-scenarios").tokenize('\n')
                         def messages = new net.sf.json.JSONArray()
-                        for (scenario in scenarios) {
-                            def commands = [
-                                [container: "php", script: "bin/console --env=behat --quiet pim:install --force"],
-                                [container: "php", script: "touch var/logs/behat.log"],
-                                [container: "php", script: "chmod 777 -R var/cache var/logs app/archive /tmp/pim app/file_storage app/uploads"],
-                                [container: "php", script: "mkdir -m 777 -p app/build/logs/behat web/media"],
-                                [container: "php", script: "sed -i '2 a umask(0000);' vendor/behat/behat/bin/behat"],
-                                [container: "php", script: "ln -s ../bin/console app/console"],
-                                [
-                                    container: "php",
-                                    junit: [in: "/home/jenkins/pim/app/build/logs/behat/", name: "*.xml"],
-                                    artifacts: [in: "/tmp/behat/screenshots", name: "*.png"],
-                                    script: "php vendor/bin/behat --strict -vv " + scenario
-                                ]
-                            ]
-                            def message = new net.sf.json.JSONObject()
-                            message.put("name",scenario)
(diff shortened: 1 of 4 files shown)


.ci/behat.community.yml
```diff
@@ -136,7 +136,7 @@ acceptance:
                 - test.locale.context
                 - test.channel.context
             filters:
-                tags: '@acceptance'
+                tags: '@acceptance-back'
```
ahocquard: but acceptance-front is with cucumber, so is it really useful to tag it with back anyway?

.ci/behat.enterprise.yml
```diff
@@ -136,3 +136,27 @@ default:
         pretty: true
         pim:
             output_path: app/build/logs/behat/
+
+
+acceptance:
+    suites:
+        acceptance:
```
ahocquard: Ok for now, but it has no meaning to call a suite "acceptance".
It should be a kind of context.

.ci/Jenkinsfile
```diff
@@ -286,28 +286,16 @@ stage("Test") {
...
-                        }
+
+                        def legacyScenarios = sh (returnStdout: true, script: "cd /home/jenkins/pim && php vendor/bin/behat -p legacy --list-scenarios").tokenize('\n')
+                        def acceptanceScenarios = sh (returnStdout: true, script: "cd /home/jenkins/pim && php vendor/bin/behat -p acceptance --list-scenarios").tokenize('\n')
+
+                        def legacyMessages = runBehatScenarios(legacyScenarios, 'legacy')
+                        messages.addAll(legacyMessages)
+
+                        def acceptanceMessages = runBehatScenarios(acceptanceScenarios, 'acceptance')
+                        messages.addAll(acceptanceMessages)
```
ahocquard: ok for now, but we should execute all acceptance tests in the same container to really have the benefit of it.

skeleton: yes, it's planned, but not in my PR. In fact we did not know acceptance tests were run during legacy tests. We will make an other PR to change that

.ci/behat.community.yml
skeleton: yes as feature files are the same for front and back (see https://github.com/akeneo/pim-community-dev/blob/2.3/tests/features/volume-monitoring/monitor_attribute_per_family_volume.feature#L6), it's better to have an explicit tag to avoid confusion

.ci/behat.enterprise.yml
skeleton: for the moment, I have no idea for an other naming, we will see when the suite will grow

.ci/behat.community.yml
ahocquard: Yep I know that we use the same, it's what we talk with Tamara.
But there is no confusion: the runner of the behat on the backend is behat, and on frontend it's cucumber.

So, you know that a context declared in PHP is for the back.

.ci/behat.enterprise.yml
ahocquard: for the volume, it's called "volume-monitoring" for example. for now, it could be called "everything_else" :trollface: 

It will be easier later on when the PIM will be splitted.

.ci/behat.community.yml
anaelChardan: Some time you will need to run only back acceptance tests so it eases the reading to know it trough tag ;-)

damien-carcel: Aggreed with Anael. Tags are mandatory to prevent choose if you want to run Gherkin on front, back or both.

PR #7641, opened 2018-02-12
Fix doctrine cursor bug

jjanvier:
**Description (for Contributor and Core Developer)**

**Definition Of Done (for Core Developer only)**

| Q                                 | A
| --------------------------------- | ---
| Added Specs                       | Todo
| Added Behats                      | Todo
| Added integration tests           | Todo
| Changelog updated                 | Todo
| Review and 2 GTM                  | Todo
| Micro Demo to the PO (Story only) | Todo
| Migration script                  | -
| Tech Doc                          | -

`Todo`: Pending / Work in progress
`OK`: Done / Validated
`-`: Not needed

PR #8018, opened 2018-04-20
PIM-7317: Fix iterator misuse in the database product model reader

LaurentPetard:

Since #7815 The UnitOfWork is cleared between each job execution batch. The problem is that there was an error in the database `ProductModelReader` so the items of the next batch were loaded from the database before the last current batch item is processed.

This problem has been fixed for the `ProductReader` in #7641 on 2.2 and brought back to 2.0, but not for the `ProductModelReader` because it doesn't exist on 2.2. So I just report the same fix but on `ProductModelReader`

`OK`: Done / Validated
`-`: Not needed


LaurentPetard:
@juliensnz There's no specific spec to add for this case, but yes I forgot to change the existing spec to take account of this problem. It's fixed

PR #8019, opened 2018-04-20
PIM-7322: Fix cast of a product model as string to format correctly violations

ahocquard:

When you have a violation on a product model when applying rules, the product model can not be cast as string:

https://github.com/akeneo/pim-enterprise-dev/blob/2.0/src/PimEnterprise/Component/CatalogRule/Engine/ProductRuleApplier/ProductsValidator.php#L69

It is the case of the product.

So, an exception is thrown:

"CATCHABLE FATAL ERROR: OBJECT OF CLASS PIM\COMPONENT\CATALOG\MODEL\PRODUCTMODEL COULD NOT BE CONVERTED TO STRING"

screen shot 2018-04-16 at 16 25 35

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Pim/Component/Catalog/spec/Model/ProductModelSpec.php
+++ b/src/Pim/Component/Catalog/spec/Model/ProductModelSpec.php
@@ -387,4 +387,25 @@ class ProductModelSpec extends ObjectBehavior
             'otherValue-<all_channels>-<all_locales>' => $otherValue
         ]);
     }
+
+    function it_gets_label_when_casting_object_as_string(
+        FamilyVariantInterface $familyVariant,
+        FamilyInterface $family,
+        AttributeInterface $attributeAsLabel,
+        ValueCollectionInterface $values
+    ) {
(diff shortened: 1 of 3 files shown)


src/Pim/Component/Catalog/spec/Model/ProductModelSpec.php
```diff
@@ -387,4 +387,25 @@ function it_has_values(
             'otherValue-<all_channels>-<all_locales>' => $otherValue
         ]);
     }
+
+    function it_gets_label_when_casting_object_as_string(
```
anaelChardan: I would remove casting from the method name ;)

anaelChardan: I would put instead `it_gets_the_label_from_to_string`

ahocquard: cosmetic

PR #8070, opened 2018-04-30
PIM-7334: Fix identifier filter with special characters

### LINKED ISSUE #7806: [2.x] Internal Server Error when filtering on SKU with forward slash "/"
Steps to reproduce:

1. Product listing
2. Filter on SKU contains SKU/1234 (full SKU is SKU/1234/234)
3. Akeneo present a popup with title 'Server Error' and text 'Error! Incorrect server response.'  and an elastic search error appears in the logs 

What should happen:

Akeneo should be escaping any reserved characters (ie forward slashes) before sending a query to elastic search.  This is explained here: https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-query-string-query.html#_reserved_characters 

screen shot 2018-03-13 at 10 05 00

[akeneo-pim-system-info_2018-03-13_10_37.txt](https://github.com/akeneo/pim-community-dev/files/1806545/akeneo-pim-system-info_2018-03-13_10_37.txt)

mmetayer:

`Pim\Bundle\CatalogBundle\Elasticsearch\Filter\IdentifierFilter` uses ES query_string clauses for `CONTAINS`, `DOES NOT CONTAIN` and `STARTS WITH` operators, so the values have to be escaped (see [elasticsearch documentation](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-query-string-query.html#_reserved_characters))
Fixes #7806 

| Added legacy Behats               | Ok

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Pim/Component/Catalog/Query/Escaper/QueryString.php
+++ b/src/Pim/Component/Catalog/Query/Escaper/QueryString.php
@@ -1,0 +1,29 @@
+<?php
+
+namespace Pim\Component\Catalog\Query\Escaper;
+
+/**
+ * @author    Mathias METAYER <mathias.metayer@akeneo.com>
+ * @copyright 2018 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+class QueryString
+{
+    /**
+     * Escapes particular values prior to doing a search query escaping whitespace, newlines or reserved characters.
+     *
+     * This is useful when using ES 'query_string' clauses in a search query.
+     *
+     * @see https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-query-string-query.html#_reserved_characters
+     *
+     * @param string $value
+     *
+     * @return string
+     */
+    public static function escapeValue(?string $value): string
+    {
+        $regex = '#[-+=|! &(){}\[\]^"~*<>?:/\\\]#';
+
+        return preg_replace($regex, '\\\$0', $value);
+    }
+}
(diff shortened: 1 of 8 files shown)


src/Pim/Component/Catalog/Query/Escaper/QueryString.php
```diff
@@ -0,0 +1,29 @@
...
+     *
+     * This is useful when using ES 'query_string' clauses in a search query.
+     *
+     * @see https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-query-string-query.html#_reserved_characters
+     *
+     * @param string $value
+     *
+     * @return string
+     */
+    public static function escapeValue(string $value): string
```
rouliane: Couldn't it be a service (instead of a static method) ? For example an Escaper service with an ElasticSearch implementation.

mmetayer: I'm OK with that (I actually considered doing it), but I wonder what would be the advantage of declaring it as a service: the method is fully stateless (it's acctually just a  call to `preg_replace` with a hard-coded pattern, which is not likely to change soon), we won't ever need to mock or extend it. Plus it avoids us from changing the constructors of the affected filters :wink: .

rouliane: Yep to be honest the only real advantage of using it as a service is that the "dependency" with the escaper is clearly visible in the constructor of the class that use it. It would make the dependency more obvious.

But it's true it's stateless and it will probably never change, so in this case it's debatable...

PR #8080, opened 2018-05-02
PIM-7330: Better validation message on sibling duplicate

damien-carcel:
## Description

When changing the parent of a variant product or a product model, or when creating a new one, we need to ensure there is not already an other variant product or product model with the same axis values.

The previous validation message was not explicit enough, as it didn't mention which entities were impacted (updated/created one, nor already existing one).

This PR updates this message, and also makes the distinction between variant products and product models.

## Definition Of Done

| Added legacy Behats               | Updated
| Added integration tests           | Updated

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Pim/Component/Catalog/Exception/AlreadyExistingAxisValueCombinationException.php
+++ b/src/Pim/Component/Catalog/Exception/AlreadyExistingAxisValueCombinationException.php
@@ -1,0 +1,57 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Pim\Component\Catalog\Exception;
+
+/**
+ * @author    Damien Carcel <damien.carcel@akeneo.com>
+ * @copyright 2018 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+class AlreadyExistingAxisValueCombinationException extends \Exception
+{
+    /** @var string */
+    private $identifier;
+
+    /** @var string */
+    private $entityClass;
+
+    /** @var string */
+    private $axisCombination;
+
+    /**
+     * @param string $identifier
+     * @param string $entityClass
+     * @param string $axisCombination
+     */
+    public function __construct(string $identifier, string $entityClass, string $axisCombination)
+    {
+        $this->identifier = $identifier;
+        $this->entityClass = $entityClass;
+        $this->axisCombination = $axisCombination;
+
+        parent::__construct($this->createExceptionMessage());
+    }
+
+    /**
+     * @return string
+     */
+    public function getEntityIdentifier(): string
+    {
+        return $this->identifier;
+    }
+
+    /**
+     * @return string
+     */
+    private function createExceptionMessage(): string
+    {
+        return sprintf(
+            'The %s "%s" already have a value for the "%s" axis combination.',
+            $this->entityClass,
+            $this->identifier,
+            $this->axisCombination
+        );
+    }
+}
--- a/src/Pim/Component/Catalog/Exception/AlreadyExistingAxisValueCombinationException.php
+++ b/src/Pim/Component/Catalog/Exception/AlreadyExistingAxisValueCombinationException.php
@@ -5,6 +5,9 @@ declare(strict_types=1);
 namespace Pim\Component\Catalog\Exception;
 
 /**
+ * This exception is thrown when an entity with family variant uses a combination
+ * of variant axis values that already exists.
+ *
  * @author    Damien Carcel <damien.carcel@akeneo.com>
  * @copyright 2018 Akeneo SAS (http://www.akeneo.com)
  * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
@@ -14,44 +17,24 @@ class AlreadyExistingAxisValueCombinationException extends \Exception
     /** @var string */
     private $identifier;
 
-    /** @var string */
-    private $entityClass;
-
-    /** @var string */
-    private $axisCombination;
-
     /**
      * @param string $identifier
-     * @param string $entityClass
-     * @param string $axisCombination
+     * @param string $message
      */
-    public function __construct(string $identifier, string $entityClass, string $axisCombination)
+    public function __construct(string $identifier, string $message)
     {
         $this->identifier = $identifier;
-        $this->entityClass = $entityClass;
-        $this->axisCombination = $axisCombination;
 
-        parent::__construct($this->createExceptionMessage());
+        parent::__construct($message);
     }
 
     /**
+     * Returns the identifier of the entity having an already existing axis value combination.
+     *
      * @return string
      */
     public function getEntityIdentifier(): string
     {
         return $this->identifier;
-    }
-
-    /**
-     * @return string
-     */
-    private function createExceptionMessage(): string
-    {
-        return sprintf(
-            'The %s "%s" already have a value for the "%s" axis combination.',
-            $this->entityClass,
-            $this->identifier,
-            $this->axisCombination
-        );
     }
 }
--- a/src/Pim/Component/Catalog/Model/ProductModel.php
+++ b/src/Pim/Component/Catalog/Model/ProductModel.php
@@ -549,91 +549,11 @@ class ProductModel implements ProductModelInterface
     }
 
     /**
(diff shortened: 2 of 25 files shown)


src/Pim/Component/Catalog/Validator/UniqueAxesCombinationSet.php
```diff
@@ -37,24 +38,29 @@ public function reset(): void
...
-        $identifier = $this->getEntityCode($entity);
+        $identifier = $this->getEntityIdentifier($entity);
 
         if (isset($this->uniqueAxesCombination[$familyVariantCode][$parentCode][$axesCombination])) {
             $cachedIdentifier = $this->uniqueAxesCombination[$familyVariantCode][$parentCode][$axesCombination];
             if ($cachedIdentifier !== $identifier) {
-                return false;
+                throw new AlreadyExistingAxisValueCombinationException(
+                    $cachedIdentifier,
+                    get_class($entity),
```
fitn: `ClassUtils::getClass()`
`get_class` could return you Proxy made by Doctrine so could be a bit confusing

src/Pim/Component/Catalog/EntityWithFamilyVariant/BuildAxisCombination.php
```diff
@@ -0,0 +1,44 @@
...
+
+use Pim\Component\Catalog\Model\AttributeInterface;
+use Pim\Component\Catalog\Model\EntityWithFamilyVariantInterface;
+
+/**
+ * @author    Damien Carcel <damien.carcel@akeneo.com>
+ * @copyright 2018 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+class BuildAxisCombination
```
arnolanglade: This should be a part of your model

src/Pim/Component/Catalog/spec/EntityWithFamilyVariant/BuildAxisCombinationSpec.php
```diff
@@ -0,0 +1,58 @@
...
+    {
+        $this->shouldHaveType(BuildAxisCombination::class);
+    }
+
+    function it_returns_a_combination_of_axis_as_a_list_of_string_for_one_attribute(
+        EntityWithFamilyVariantInterface $entityWithFamilyVariant,
+        AttributeInterface $color,
+        OptionValue $red
+    ) {
+        $color->getCode()->willReturn('color');
```
arnolanglade: you should not use collaborator but real model instance.

src/Pim/Component/Catalog/Exception/AlreadyExistingAxisValueCombinationException.php
```diff
@@ -0,0 +1,63 @@
...
+
+        parent::__construct($this->createExceptionMessage());
+    }
+
+    /**
+     * Returns the identifier of the entity having an already existing axis value combination.
+     *
+     * @return string
+     */
+    public function getEntityWithExistingCombinationIdentifier(): string
```
arnolanglade: `entityIndentifier` ?

src/Pim/Component/Catalog/Validator/UniqueAxesCombinationSet.php
```diff
@@ -37,24 +39,29 @@ public function reset(): void
...
-        $identifier = $this->getEntityCode($entity);
+        $identifier = $this->getEntityIdentifier($entity);
 
         if (isset($this->uniqueAxesCombination[$familyVariantCode][$parentCode][$axesCombination])) {
             $cachedIdentifier = $this->uniqueAxesCombination[$familyVariantCode][$parentCode][$axesCombination];
             if ($cachedIdentifier !== $identifier) {
-                return false;
+                throw new AlreadyExistingAxisValueCombinationException(
+                    $cachedIdentifier,
+                    ClassUtils::getClass($entity),
```
arnolanglade: you should give the entity to the exception constructor, the exception should be able to extract the class name.

src/Pim/Component/Catalog/EntityWithFamilyVariant/BuildAxisCombination.php
```diff
@@ -0,0 +1,44 @@
...
+     * The combination of this entity will be "[blue],[xl]".
+     *
+     * This allows use to compare multiple combinations, to look for a potential duplicate.
+     *
+     * @param EntityWithFamilyVariantInterface $entityWithFamilyVariant
+     * @param AttributeInterface[]             $axes
+     *
+     * @return string
+     */
+    public function asListOfString(EntityWithFamilyVariantInterface $entityWithFamilyVariant, array $axes): string
```
arnolanglade: Why do you build something to give it to something else. This break the ["tell don't ask"](https://martinfowler.com/bliki/TellDontAsk.html) principle. Tell to an object to do something don't ask data to do something.

src/Pim/Component/Catalog/Exception/AlreadyExistingAxisValueCombinationException.php
```diff
@@ -0,0 +1,70 @@
...
+     * @param EntityWithFamilyVariantInterface $entityClass
+     * @param string                           $axisCombination
+     */
+    public function __construct(
+        string $identifier,
+        EntityWithFamilyVariantInterface $entityClass,
+        string $axisCombination
+    ) {
+        $this->identifier = $identifier;
+        $this->entityClass = $entityClass;
```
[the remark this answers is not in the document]
fitn: To me, $identifier is `$entity->getIdentifier()`. I let you rename this in a better way :)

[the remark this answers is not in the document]
damien-carcel: Done, there is no more need for the entity, only the identifier and a message.

src/Pim/Component/Catalog/Model/ProductModel.php
```diff
@@ -700,4 +628,92 @@ public function setAssociations(Collection $associations): AssociationAwareInter
 
         return $this;
     }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getCombinationOfAxisValues(array $axes): string
```
jjanvier: same comment

src/Pim/Component/Catalog/Model/EntityWithFamilyVariantInterface.php
```diff
@@ -49,4 +49,24 @@ public function setParent(ProductModelInterface $parent = null): void;
...
+     *
+     * @todo TIP-857: "$axes" should not be provided as an argument, but directly by the entity. This implies to move
+     *       the code of the "Pim\Component\Catalog\FamilyVariant\EntityWithFamilyVariantAttributesProvider" directly
+     *       in the models.
+     *
+     * @param AttributeInterface[] $axes
+     *
+     * @return string
+     */
+    public function getCombinationOfAxisValues(array $axes): string;
```
jjanvier: you don't return a value here
what is a value is defined here https://github.com/akeneo/pim-community-dev/blob/2.3/STANDARD_FORMAT.md#values

imo it's better to return the real values object and to format them where you need 
it you use those axis values at a lot of place, you can consider create your own collection, with a formater method inside

naming is important
encapsulation is important

src/Pim/Component/Catalog/Model/ProductModel.php
```diff
@@ -700,4 +628,92 @@ public function setAssociations(Collection $associations): AssociationAwareInter
...
+        }
+
+        return implode(',', $combination);
+    }
+
+    /**
+     * @param EntityWithFamilyVariantInterface $entity
+     * @param ValueCollectionInterface         $valueCollection
+     *
+     * @return ValueCollectionInterface
```
jjanvier: you see the difference?
here we really have some values

damien-carcel: OK, so sorry, but this PR is only about changing a message. So even if I completely agree with what you are saying, I'm just reverting the refactoring, and I will add it to the TIP card I already created for the axis provider.
This PR was functional in half a day, and I just loose 3 days of refactoring to achieve something obviously bad. I don't want to loose anymore time on this for now.

jjanvier: ok :(

PR #8093, opened 2018-05-04
TIP-856: Move Pim\MeasureBundle to Akeneo\Tool

arnolanglade:

`OK`: Done / Validated
`-`: Not needed


src/Akeneo/Bundle/MeasureBundle/phpspec.yml
```diff
@@ -1,5 +0,0 @@
-suites:
```
jjanvier: isn't the MeasureBundle provided as a standalone library?
same for the BatchBundle by the way

if it's the case, this file should stay here

arnolanglade: I don't understand, this file still exist locally...

PR #8139, opened 2018-05-16
PIM-7347: Fix completeness for locale specific attributes

SamirBoulil:
**Context:**

The calculation of the completeness of products is a two step process.

It basically starts in this class: `Pim\Component\Catalog\Completeness\CompletenessCalculator::Calculate`

**1. Build the list of required values based on the attribute requirements.**

Based on the attribute requirements of the family, we generate a
`Pim\Component\Catalog\EntityWithFamily\RequiredValueCollection`, which holds a list of `\Pim\Component\Catalog\EntityWithFamily\RequiredValue`.

Basically, this object holds a reference for each required value for:
- An attribute
- A channel
- A locale

Important note:

Conviniently, this RequiredValue extends `Pim\Component\Catalog\Model\ValueInterface` which
gives us the opportunity to get the value of a product based on this
object with
```
$productValueCollection->getSame($requiredValue); // the actual value
```

**2. Use the required values and the values of the product to generate**
completeness.

In
`Pim\Component\Catalog\EntityWithFamily\IncompleteValueCollectionFactory`, we use
this list of RequiredValues to find the value for a given product using
the *ValueCollection::getSame method* and test if it's complete or not.

And that's it!

**Problem: Let there be non-localizable locale specific attributes**

The coupling between a required value (the information it holds) and the
fact that we can directly use this object to retrieve the data from the
product value collection becomes an issue when the locale specific values comes
in.

Let's say we have:
- An attribute "description" which is localeSpecific on
'en_US' and 'fr_FR'.
- A family "camcorders" with two attributes: "sku", "description"
required on all channels
- A product "webcam" with the following values:
```
| sku | description |
| 001 | ""          |
```

Now, when we generate the required values in the first step for this family, we get
something like this:
```
[
  'requiredValue1' => [
      'attribute' => 'description',
      'locale' => 'fr_FR',
      'scope' => null
  ],
  'requiredValue2' => [
      'attribute' => 'description',
      'locale' => 'en_US',
      'scope' => null
  ]
]
```

In the second step, we try to check if the values of the products are
complete or not, we try to get the value using:

```
$myWebcapProduct->getSame($requiredValue1); // => returns null
$myWebcapProduct->getSame($requiredValue2); // => returns null
```

indeed, the value collection cannot return it's data for the attribute
'description' on the locale 'fr_FR' or 'en_US' because it is not localizable!

Internally, in the value collection of the product they are indexed
using the keys:
'description-<all_channels>-<all_locales>' and not 'description-<all_channels>-fr_FR' nor
'description-<all_channels>-en_US'.

That's why we need to decouple the required value structure and the way
the value of the products are retrieved.

**Solution:**

The solution I suggest involves 2 things:
- Redesign of the RequiredValue object which should contain the information about the attribute requirements (attribute, channel, locale) as well as the information needed to retrieve the value from the product value collection (depending on the products).
- Take into account locale specific attributes when building the incompleteRequiredValueCollection.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Pim/Component/Catalog/EntityWithFamily/IncompleteValueCollectionFactory.php
+++ b/src/Pim/Component/Catalog/EntityWithFamily/IncompleteValueCollectionFactory.php
@@ -70,13 +70,12 @@ class IncompleteValueCollectionFactory
      * @return bool
      */
     private function isValueMissingOrEmpty(
-        ValueInterface $value,
+        RequiredValue $value,
         ChannelInterface $channel,
         LocaleInterface $locale,
         EntityWithValuesInterface $entityWithValues
     ) {
-        $values = $entityWithValues->getValues();
-        $actualValue = $values->getSame($value);
+        $actualValue = $entityWithValues->getValues()->getByKey($value->getValueKey());
 
         if (null === $actualValue) {
             return true;
--- a/src/Pim/Component/Catalog/EntityWithFamily/IncompleteValueCollectionFactory.php
+++ b/src/Pim/Component/Catalog/EntityWithFamily/IncompleteValueCollectionFactory.php
@@ -75,7 +75,12 @@ class IncompleteValueCollectionFactory
         LocaleInterface $locale,
         EntityWithValuesInterface $entityWithValues
     ) {
-        $actualValue = $entityWithValues->getValues()->getByKey($value->getValueKey());
+        $actualValue = $value->valueFromEntity()(
+            $entityWithValues,
+            $value->getAttribute(),
+            $value->getScope(),
+            $value->getLocale()
+        );
 
         if (null === $actualValue) {
             return true;
--- a/src/Pim/Component/Catalog/EntityWithFamily/IncompleteValueCollectionFactory.php
+++ b/src/Pim/Component/Catalog/EntityWithFamily/IncompleteValueCollectionFactory.php
@@ -62,7 +62,7 @@ class IncompleteValueCollectionFactory
     }
 
     /**
-     * @param ValueInterface            $value
+     * @param ValueInterface            $requiredValue
      * @param ChannelInterface          $channel
      * @param LocaleInterface           $locale
      * @param EntityWithValuesInterface $entityWithValues
@@ -70,16 +70,15 @@ class IncompleteValueCollectionFactory
      * @return bool
      */
     private function isValueMissingOrEmpty(
-        RequiredValue $value,
+        RequiredValue $requiredValue,
         ChannelInterface $channel,
         LocaleInterface $locale,
         EntityWithValuesInterface $entityWithValues
     ) {
-        $actualValue = $value->valueFromEntity()(
-            $entityWithValues,
-            $value->getAttribute(),
-            $value->getScope(),
-            $value->getLocale()
+        $actualValue = $entityWithValues->getValues()->getByCodes(
+            $requiredValue->forAttributeCode(),
+            $requiredValue->forScope(),
+            $requiredValue->forLocale($locale)
         );
 
         if (null === $actualValue) {
--- a/src/Pim/Component/Catalog/EntityWithFamily/IncompleteValueCollectionFactory.php
+++ b/src/Pim/Component/Catalog/EntityWithFamily/IncompleteValueCollectionFactory.php
@@ -62,7 +62,7 @@ class IncompleteValueCollectionFactory
     }
 
     /**
-     * @param ValueInterface            $requiredValue
+     * @param RequiredValue             $requiredValue
      * @param ChannelInterface          $channel
      * @param LocaleInterface           $locale
      * @param EntityWithValuesInterface $entityWithValues
@@ -76,9 +76,9 @@ class IncompleteValueCollectionFactory
         EntityWithValuesInterface $entityWithValues
     ) {
         $actualValue = $entityWithValues->getValues()->getByCodes(
-            $requiredValue->forAttributeCode(),
-            $requiredValue->forScope(),
-            $requiredValue->forLocale($locale)
+            $requiredValue->attribute(),
+            $requiredValue->scope(),
+            $requiredValue->locale()
         );
 
         if (null === $actualValue) {
--- a/src/Pim/Component/Catalog/EntityWithFamily/IncompleteValueCollectionFactory.php
+++ b/src/Pim/Component/Catalog/EntityWithFamily/IncompleteValueCollectionFactory.php
@@ -77,7 +77,7 @@ class IncompleteValueCollectionFactory
     ) {
         $actualValue = $entityWithValues->getValues()->getByCodes(
             $requiredValue->attribute(),
-            $requiredValue->scope(),
+            $requiredValue->channel(),
             $requiredValue->locale()
         );
 
--- a/src/Pim/Component/Catalog/EntityWithFamily/RequiredValue.php
+++ b/src/Pim/Component/Catalog/EntityWithFamily/RequiredValue.php
@@ -25,16 +25,30 @@ use Pim\Component\Catalog\Model\ValueInterface;
  */
 class RequiredValue extends AbstractValue implements ValueInterface
 {
+    /** @var string */
+    private $valueKey;
+
     /**
      * @param AttributeInterface $attribute
      * @param string             $channel
      * @param string             $locale
+     * @param string             $valueKey
      */
-    public function __construct(AttributeInterface $attribute, string $channel = null, string $locale = null)
-    {
+    public function __construct(
+        AttributeInterface $attribute,
+        string $channel = null,
+        string $locale = null,
+        string $valueKey = null
+    ) {
         $this->setAttribute($attribute);
         $this->setScope($channel);
         $this->setLocale($locale);
+        $this->valueKey = $valueKey;
+    }
+
+    public function getValueKey(): string
+    {
+        return $this->valueKey;
     }
 
     /**
--- a/src/Pim/Component/Catalog/EntityWithFamily/RequiredValue.php
+++ b/src/Pim/Component/Catalog/EntityWithFamily/RequiredValue.php
@@ -25,30 +25,30 @@ use Pim\Component\Catalog\Model\ValueInterface;
  */
 class RequiredValue extends AbstractValue implements ValueInterface
 {
-    /** @var string */
-    private $valueKey;
+    /** @var \Closure  */
+    private $valueGetter;
 
     /**
      * @param AttributeInterface $attribute
      * @param string             $channel
      * @param string             $locale
-     * @param string             $valueKey
+     * @param string             $valueGetter
      */
     public function __construct(
         AttributeInterface $attribute,
         string $channel = null,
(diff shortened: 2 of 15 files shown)


src/Pim/Component/Catalog/EntityWithFamily/IncompleteValueCollectionFactory.php
```diff
@@ -70,13 +70,12 @@ public function forChannelAndLocale(
      * @return bool
      */
     private function isValueMissingOrEmpty(
-        ValueInterface $value,
```
SamirBoulil: I understand this could be an issue, since we typehint with the actual object. But IMO it's ok.

src/Pim/Component/Catalog/Model/AbstractValue.php
```diff
@@ -116,4 +116,17 @@ protected function setLocale($locale)
 
         $this->locale = $locale;
     }
+
+    /**
+     * @param string $locale
+     *
+     * @return bool
+     */
+    protected function isProductValueNotLocalizableNeitherLocaleSpecific($locale): bool
```
SamirBoulil: Pass in locale and attribute

SamirBoulil: typehint

src/Pim/Component/Catalog/Model/ValueCollection.php
```diff
@@ -319,4 +314,20 @@ public function filter(\Closure $filterBy)
...
     }
+
+    /**
+     * @param string $attributeCode
+     * @param string $channelCode
+     * @param string $localeCode
+     *
+     * @return string
+     */
+    public static function generateKey(string $attributeCode, ?string $channelCode, ?string $localeCode): string
```
arnolanglade: You should create an object instead of using static method. I think it will be better to encapsulate those data into an object which will have its own behavior. Moreover it will be easier to add extra behavior if you create an object. You should have a look to https://www.yegor256.com/2017/02/07/private-method-is-new-class.html or https://www.yegor256.com/2014/05/05/oop-alternative-to-utility-classes.html

src/Pim/Component/Catalog/EntityWithFamily/RequiredValue.php
```diff
@@ -25,16 +25,30 @@
...
      * @param string             $locale
+     * @param string             $valueKey
      */
-    public function __construct(AttributeInterface $attribute, string $channel = null, string $locale = null)
-    {
+    public function __construct(
+        AttributeInterface $attribute,
+        string $channel = null,
+        string $locale = null,
+        string $valueKey = null
```
arnolanglade: I don't know if it is a good idea to do that because you expose an implementation detail of this object. You should only to tell an object to do something depending on its state (the encapsulated data ). Here, you change the behavior of your object. I think it is more a hack than a solution.

src/Pim/Component/Catalog/EntityWithFamily/RequiredValue.php
```diff
@@ -25,16 +25,30 @@
...
+    ) {
         $this->setAttribute($attribute);
         $this->setScope($channel);
         $this->setLocale($locale);
+        $this->valueKey = $valueKey;
+    }
+
+    public function getValueKey(): string
```
arnolanglade: Getter are evil, it break the tell don't ask principle. You should have a look to https://www.yegor256.com/2014/09/16/getters-and-setters-are-evil.html for instance.

src/Pim/Component/Catalog/Model/ValueCollection.php
SamirBoulil: Ok, I agree I hesitated to create a dedicated object anyways

src/Pim/Component/Catalog/EntityWithFamily/RequiredValue.php
SamirBoulil: Yep, I understand and agree but do you have another solution in mind ?

src/Pim/Component/Catalog/EntityWithFamily/RequiredValue.php
SamirBoulil: Depending on the final solution, this getter might disappear. so ok.

jjanvier:
I don't like the solution. Exposing the key of the values was the worst thing I've done about single storage. It leaks some implementation details everywhere. It complexifies the code, and it's clearly a weak design.

Can we pair today or tomorrow to try to find a better solution @SamirBoulil?

SamirBoulil:
@jjanvier I'm just coming back on this ticket for the fish'N ship and saw your comment now.

Yep I don't like it either, I implemented it this way to trigger discussions.So Yeah, I'll definitely come back to you once you are back to find another solution.

src/Pim/Component/Catalog/EntityWithFamily/RequiredValue.php
```diff
@@ -19,37 +19,98 @@
  *  - a price is required on the channel "ecommerce" and all locales
  *  - ...
  *
+ * This object gives you the information of the requirement in the matrix: attribute, channel, locale.
+ * that's why the functions forAttribute, forScope and ForLocale always return an object.
```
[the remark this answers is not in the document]
SamirBoulil: I agree.

src/Pim/Component/Catalog/EntityWithFamily/RequiredValue.php
```diff
@@ -19,37 +19,98 @@
...
+ *
+ * It also helps you how to retrieve the corresponding value in a value collection.
+ * For instance, if the attribute is not scopable and not localizable. if we try to compute the completeness on the
+ * channel 'ecommerce' and locale 'fr_FR' for this attribute.
+ * - forScope will return the channel ecommerde
```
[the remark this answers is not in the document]
SamirBoulil: xD holy shit it happened more than once. thanks

src/Pim/Component/Catalog/EntityWithFamily/RequiredValue.php
```diff
@@ -19,37 +19,98 @@
...
+ * - forLocale will return the locale fr_FR
```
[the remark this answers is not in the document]
SamirBoulil: It's actually the object here, so I'll clarify

src/Pim/Component/Catalog/EntityWithFamily/RequiredValue.php
```diff
@@ -19,37 +19,98 @@
...
+ * the information of the matrix, and
```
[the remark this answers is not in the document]
SamirBoulil: Nope, it's just the whole block which is not not clear enough. I'll rewrite this documentation

src/Pim/Component/Catalog/EntityWithFamily/RequiredValue.php
```diff
@@ -19,37 +19,98 @@
...
+ * - scope() will return null
+ * - locale('fr_FR') will return null
```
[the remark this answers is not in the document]
SamirBoulil: I agree.

src/Pim/Component/Catalog/EntityWithFamily/RequiredValue.php
```diff
@@ -19,37 +19,98 @@
...
-    public function getData()
+    public function scope(): ?string
     {
-        return null;
+        return $this->attribute->isScopable() ? $this->channel->getCode() : null;
     }
 
     /**
-     * {@inheritdoc}
+     * @param LocaleInterface $locale
```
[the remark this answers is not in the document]
SamirBoulil: thanks

src/Pim/Component/Catalog/EntityWithFamily/RequiredValueCollectionFactory.php
```diff
@@ -35,17 +38,13 @@ public function forChannel(FamilyInterface $family, ChannelInterface $channel):
...
+
                     if ($attribute->isLocaleSpecific() && !$attribute->hasLocaleSpecific($locale)) {
                         continue;
                     }
 
-                    $channelCode = $attribute->isScopable() ? $channel->getCode() : null;
-                    $localeCode = $attribute->isLocalizable() ? $locale->getCode() : null;
-
-                    $requiredValues[] = new RequiredValue($attribute, $channelCode, $localeCode);
+                    $requiredValues[] = new RequiredValue($attribute, $attributeRequirement->getChannel(), $locale);
```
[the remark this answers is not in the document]
SamirBoulil: I don't understand your comment. This line initializes the requiredValue for the attribute, channel and locale object (the parameter of the matrix), nothing more.

src/Pim/Component/Catalog/EntityWithFamily/RequiredValueCollection.php
```diff
@@ -70,14 +69,20 @@ public function filterByChannelAndLocale(
     ): RequiredValueCollection {
         $filteredValues = array_filter(
             $this->values,
-            function (ValueInterface $value) use ($channel, $locale) {
-                $attribute = $value->getAttribute();
+            function (RequiredValue $requiredValue) use ($channel, $locale) {
+                $attribute = $requiredValue->forAttribute();
 
-                if ($attribute->isScopable() && $channel->getCode() !== $value->getScope()) {
+                if ($attribute->isScopable() && $requiredValue->forScope()->getCode() !== $channel->getCode()) {
```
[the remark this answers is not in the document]
SamirBoulil: If you look closely, the code didn't change. It's just a matter of rewering correctly the functions since forChannel now returns an object.

src/Pim/Component/Catalog/EntityWithFamily/RequiredValue.php
```diff
@@ -19,37 +19,103 @@
...
+ *
  * @author    Julien Janvier <j.janvier@gmail.com>
  * @copyright 2017 Akeneo SAS (http://www.akeneo.com)
  * @license   http://opensource.org/licenses/osl-3.0.php Open Software License (OSL 3.0)
  */
-class RequiredValue extends AbstractValue implements ValueInterface
+class RequiredValue
 {
+    /** @var AttributeInterface */
+    private $attribute;
```
SamirBoulil: rename with forAttribute

src/Pim/Component/Catalog/EntityWithFamily/RequiredValue.php
```diff
@@ -19,37 +19,103 @@
  *  - a price is required on the channel "ecommerce" and all locales
  *  - ...
  *
+ * This object gives you the information of the requirement in the matrix: attribute, channel, locale.
+ * that's why the functions forAttribute, forScope and ForLocale always return an object representing the value in those
```
jjanvier: forChannel everywhere

src/Pim/Component/Catalog/EntityWithFamily/RequiredValueCollectionFactory.php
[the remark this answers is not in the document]
jjanvier: ok explained IRL, this is indeed what's done

PR #8140, opened 2018-05-16
Spring cleaning - Burn "I should see the flash message" from all behat steps

momoss:
**Time to clean our own shit**

Remove Old Behat Legacy step "I should see the flash message"  where code of function was a `return;` since many time...

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/features/Context/AssertionContext.php
+++ b/features/Context/AssertionContext.php
@@ -248,27 +248,11 @@ class AssertionContext extends PimContext
     }
 
     /**
-     * @deprecated This function was disabled because it generates too many failing tests. Warning, some tests are
-     * always using it, and it checks nothing at all.
-     *
-     * @param $text
-     *
      * @Then /^I should see the flash message "(.*)"$/
      */
     public function iShouldSeeTheFlashMessage($text)
     {
-        return;
-
-        $this->spin(function () use ($text) {
-            $flashes = $this->getCurrentPage()->findAll('css', '.flash-messages-holder > div');
-            foreach ($flashes as $flash) {
-                if (false !== strpos($flash->getText(), $text)) {
-                    return true;
-                }
-            }
-
-            return null;
-        }, sprintf('Can not find flash message with text "%s"', $text));
+        Throw new \Exception('Do not use this legacy function.');
     }
 
     /**
--- a/features/attribute-group/delete_attribute_group.feature
+++ b/features/attribute-group/delete_attribute_group.feature
@@ -18,17 +18,14 @@ Feature: Attribute group creation
   Scenario: Successfully delete an attribute group
     Given I am on the "sizes" attribute group page
     When I press the secondary action "Delete"
-    And I confirm the deletion
-    Then I should see the flash message "Attribute group successfully removed"
+    Then I confirm the deletion
 
   Scenario: Fail to delete an attribute group that contains attributes
     Given I am on the "colors" attribute group page
     When I press the secondary action "Delete"
-    And I confirm the deletion
-    Then I should see the flash message "Attribute group containing attributes cannot be removed. Please remove its attributes prior to delete it."
+    Then I confirm the deletion
 
   Scenario: Fail to delete the attribute "other"
     Given I am on the "other" attribute group page
     When I press the secondary action "Delete"
-    And I confirm the deletion
-    Then I should see the flash message "Attribute group \"other\" cannot be removed."
+    Then I confirm the deletion
--- a/features/category/remove_a_category.feature
+++ b/features/category/remove_a_category.feature
@@ -17,7 +17,6 @@ Feature: Remove a category
     When I press the secondary action "Delete"
     And I confirm the deletion
     Then I should be on the category "summer_collection" edit page
-    And I should see the flash message "Category successfully removed"
     And I should not see the "Sandals" category under the "summer_collection" category
 
   @skip
@@ -26,7 +25,6 @@ Feature: Remove a category
     When I press the secondary action "Delete"
     And I confirm the deletion
     Then I should be on the category "2014_collection" edit page
-    And I should see the flash message "Category successfully removed"
     And I should not see "Winter collection"
 
   @unstable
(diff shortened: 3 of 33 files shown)


ahocquard:
@pierallard I don't agree with you:

- having died code is useless. 

How many new behats are using this sentence, thinking that it works as expected?
_but everyone in the code team should be aware of this_: hardly doable to follow every exceptions.

I already saw some new PRs using it, and when I raised the problem, the developer was "omg but why?". That's the best way to have new random actually.

- we should test differently and stopping to use the UI for everything. This is a big rework. Meaning that all the tests should be *entirely* rewritten, from the gherkin to the code executed behind.

So, we will never "find a way to check". We have far better to do on the test stack.

- a test should not use this sentence "and I wait..". This is purely technical stuff to make assertion. It should be in the code, not in the gherkin.

ahocquard:
@momoss for this kind of PR, merging on the next future version is maybe better

momoss:
@pierallard Ok but when the future will happens ?
it's more than **2 years** that we have those useless steps in our code.
- Put "I should not see the text" in scenarii is propose when you save something, maybe not the right solution.
- But keep scenarii as they are is not good to, and when I see last year features tests with some "I should see the flash message"... Sorry but it is not acceptable too.
- If you want to help to find a workaround it could be very nice

features/attribute-group/delete_attribute_group.feature
```diff
@@ -18,17 +18,14 @@ Feature: Attribute group creation
...
-    And I confirm the deletion
-    Then I should see the flash message "Attribute group successfully removed"
+    Then I confirm the deletion
 
   Scenario: Fail to delete an attribute group that contains attributes
     Given I am on the "colors" attribute group page
     When I press the secondary action "Delete"
-    And I confirm the deletion
-    Then I should see the flash message "Attribute group containing attributes cannot be removed. Please remove its attributes prior to delete it."
+    Then I confirm the deletion
```
anaelChardan: What does confirm that you don't have the right to do that?

features/attribute-group/delete_attribute_group.feature
```diff
@@ -18,17 +18,14 @@ Feature: Attribute group creation
   Scenario: Successfully delete an attribute group
     Given I am on the "sizes" attribute group page
     When I press the secondary action "Delete"
-    And I confirm the deletion
-    Then I should see the flash message "Attribute group successfully removed"
```
anaelChardan: What does confirm that the call was successful?

features/category/remove_a_category.feature
```diff
@@ -51,9 +48,8 @@ Feature: Remove a category
     When I press the secondary action "Delete"
     And I confirm the deletion
     Then I should be on the category "2014_collection" edit page
-    And I should see the flash message "Category successfully removed"
-    Then I should not see "Winter collection"
-    And I should not see "Winter boots"
+    And I should not see "Winter collection"
+    Then I should not see "Winter boots"
```
anaelChardan: And

features/category/remove_a_category.feature
```diff
@@ -51,9 +48,8 @@ Feature: Remove a category
     When I press the secondary action "Delete"
     And I confirm the deletion
     Then I should be on the category "2014_collection" edit page
-    And I should see the flash message "Category successfully removed"
-    Then I should not see "Winter collection"
-    And I should not see "Winter boots"
+    And I should not see "Winter collection"
```
anaelChardan: Maybe see the text is better

anaelChardan:
IMHO we shouldn't replace all of those occurrences without adding a "security net"

damien-carcel:
Those flash messages are useless and dangerous because non reliable. And anyway, we should stop writing this kind of behat step. It has no business value, it is not how we should write gherkins.

Checking that the flash message work is a front test, not back, so no place in behats.

anaelChardan:
I do agree with you @damien-carcel but just removing them only is not a solution, maybe this PR should come with another that adds this value?

damien-carcel:
It already adds value by removing useless code that other dev are using thinking it does something when in fact it does nothing.

features/attribute-group/delete_attribute_group.feature
momoss: @anaelChardan if you want to purpose something better, you can, but a `Then I should see the ....` that does a `return;`is better you think?

features/category/remove_a_category.feature
momoss: It comes from the default `MinkContext` that we use in many behat, and not sure that this kind of edit is valuable

momoss:
And @anaelChardan about your question:
Why you didn't replace all occurences of I should (not) see the flash message) by I should (not) see the text` ? 
- because `I should (not) see the text` add spin and it could increase significantly the execution time of behat tests

features/category/remove_a_category.feature
anaelChardan: Cc @pierallard already tell me that it was for more better?

features/attribute-group/delete_attribute_group.feature
anaelChardan: The null value was actually throwing an exception so you are removing a thing here :/

anaelChardan:
Actually the null value was throwing an exception: https://github.com/akeneo/pim-community-dev/blob/2.0/features/Context/Spin/SpinCapableTrait.php#L57

features/Context/AssertionContext.php
```diff
@@ -248,27 +248,11 @@ public function theFieldsShouldBeDisabled($fields)
...
-     * @deprecated This function was disabled because it generates too many failing tests. Warning, some tests are
-     * always using it, and it checks nothing at all.
-     *
-     * @param $text
-     *
      * @Then /^I should see the flash message "(.*)"$/
      */
     public function iShouldSeeTheFlashMessage($text)
     {
-        return;
```
anaelChardan: HERE the thing that is useless ;)

PR #8156, opened 2018-05-21
PimConnectorBundle: optimize product read

gplanchat:

Channel fetch was executed on each product read. As the exports are limited to one channel at a time, the case where it is required should not appear and as the fetch is already done in the `initialize()` stage. We now store this info at this moment and use the stored channel instance from this point on each read.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Pim/Component/Connector/Reader/Database/ProductReader.php
+++ b/src/Pim/Component/Connector/Reader/Database/ProductReader.php
@@ -86,6 +86,11 @@ class ProductReader implements ItemReaderInterface, InitializableInterface, Step
 
         $this->firstRead = true;
     }
+    
+    protected function getConfiguredChannel()
+    {
+        return $this->configuredChannel;
+    }
 
     /**
      * {@inheritdoc}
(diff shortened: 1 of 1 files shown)


src/Pim/Component/Connector/Reader/Database/ProductReader.php
```diff
@@ -46,6 +46,9 @@ class ProductReader implements ItemReaderInterface, InitializableInterface, Step
 
     /** @var bool */
     private $firstRead = true;
+    
+    /** @var ChannelInterface|null */
+    private $configuredChannel;
```
fitn: As it is probably already overridden in some projects. Could be better to put as protected or add a getter on it ;)

src/Pim/Component/Connector/Reader/Database/ProductReader.php
```diff
@@ -73,16 +76,21 @@ public function __construct(
...
         }
 
         $filters = $this->getConfiguredFilters();
-        $this->products = $this->getProductsCursor($filters, $channel);
+        $this->products = $this->getProductsCursor($filters, $this->configuredChannel);
 
         $this->firstRead = true;
     }
+    
+    protected function getConfiguredChannel()
```
gplanchat: @fitn the protected accessor has been added

PR #8164, opened 2018-05-22
[Continuous Integration] Fix firefox random behavior

ClementGautier:


.ci/vhost.conf
```diff
@@ -1,12 +1,8 @@
 <VirtualHost *:80>
     ServerName 127.0.0.1
+    DocumentRoot /var/www/html/web
 
-    <FilesMatch \.php$>
```
momoss: Is it normal to remove the use of php-fpm?

Dockerfile
```diff
@@ -1,140 +1,65 @@
-FROM php:7.1-fpm
```
momoss: Is it normal to remove the use of php-fpm?

Dockerfile
```diff
@@ -1,140 +1,65 @@
...
+FROM php:7.1-apache
```
damien-carcel: What? You're using Apache with mod-php instead of FPM? That's not good as it is not the PIM 2.0 requirements.
What don't you keep FPM and configure an httpd container with the previous vhost?

app/config/config_behat.yml
```diff
@@ -1,6 +1,5 @@
 imports:
     - { resource: config.yml }
-    - { resource: parameters_test.yml }
```
damien-carcel: Why do you remove this parameter file? I'm not saying I am against using the same config for all environments, but I'd like an explanation. This environment separation between prod/dev and test/behat is quite usefull in our everyday work.
Plus, you did it in CE, but not in EE...

ClementGautier: We use environement variable to generate a parameters.yml through the composer handler. There is 2 way of dealing with parameters: a yml file (parameters.yml) or environement variables. The first one is kind of deprecated since the .env revolution so it doesn't really make sense to me to keep the both ways. Anyway I didn't meant to do it in this PR, it will got reverted don't worry

Dockerfile
ClementGautier: cf https://github.com/akeneo/pim-community-dev/pull/8164#discussion_r190206058

Dockerfile
ClementGautier: It brings complexity in the Dockerfile for no good reason from my point of view. Also it's another things to maintains instead of relying on the official image.

damien-carcel: Sorry, but I really don't see how complex it is to add `apt install apache2` then `apache2enmod fcgi` (OK, the commands are not exactly that, but still, in 2 or 3 commands you have apache in FCGI mod, able to run with FPM) in `php:7.1-fpm`.
And I repeat myself: we ask our user to use FPM, not mod_php. So maybe I am wrong, but from my point of view it is not acceptable to test only mod_php when we ask people not to use it.

damien-carcel: Perfect.

PR #8165, opened 2018-05-22
TIP-858: Enable php coupling detector for user management BC

arnolanglade:
The following drawing present the different bounded contexts we will introduce the next major version. 

dessin sans titre

A bounded context should not be coupled to another one. That's why we use PHP coupling detector to check automatically that. For now, it is now possible to decouple all BC because of our legacy code, we don't to rewrite the PIM/have time. We added some exceptions to the PHP coupling detector configuration but we added comment to explain how fix those coupling issues.


composer.json
```diff
@@ -78,7 +78,7 @@
     },
     "require-dev": {
         "akeneo/phpspec-skip-example-extension": "2.0.*",
-        "akeneo/php-coupling-detector": "0.2.0",
+        "akeneo/php-coupling-detector": "dev-master",
```
damien-carcel: or we could tag a 0.3? :-)

src/Akeneo/UserManagement/Component/Model/UserInterface.php
```diff
@@ -46,7 +45,7 @@ public function getEmail();
     /**
      * Return birthday
      *
-     * @return DateTime
+     * @return \DateTime
```
anaelChardan: Why is it part of this PR?

arnolanglade: I fixed some stuff with more strict rules.

composer.json
arnolanglade: It is not plan for now.

damien-carcel: OK then.

jjanvier: I will do it soon :)

PR #8171, opened 2018-05-23
PIM-7358: Cascade remove variant attribute sets when removing a family variant

mmetayer:

As of version 2.2, it is possible to remove family variants. Problem was, related entities (variant attribute sets) were not removed, hence preventing from e.g removing an attribute, even if it was not used as an axis anymore.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/upgrades/schema/Version_2_2_20180523125419.php
+++ b/upgrades/schema/Version_2_2_20180523125419.php
@@ -1,0 +1,35 @@
+<?php
+
+namespace Pim\Upgrade\Schema;
+
+use Doctrine\DBAL\Migrations\AbstractMigration;
+use Doctrine\DBAL\Schema\Schema;
+
+/**
+ * Add CASCADE DELETE on pim_catalog_family_variant_attribute_set related foreign keys
+ */
+class Version_2_2_20180523125419 extends AbstractMigration
+{
+    /**
+     * @param Schema $schema
+     */
+    public function up(Schema $schema)
+    {
+        $this->addSql('ALTER TABLE pim_catalog_family_variant_has_variant_attribute_sets DROP FOREIGN KEY FK_1F4DC7028A37AD0');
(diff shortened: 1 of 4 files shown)


upgrades/schema/Version_2_2_20180523125419.php
```diff
@@ -0,0 +1,35 @@
...
+ * Add CASCADE DELETE on pim_catalog_family_variant_attribute_set related foreign keys
+ */
+class Version_2_2_20180523125419 extends AbstractMigration
+{
+    /**
+     * @param Schema $schema
+     */
+    public function up(Schema $schema)
+    {
+        $this->addSql('ALTER TABLE pim_catalog_family_variant_has_variant_attribute_sets DROP FOREIGN KEY FK_1F4DC7028A37AD0');
```
mmetayer: Should I also add a script to cleanup orphan variant attribute sets?

upgrades/schema/Version_2_2_20180523125419.php
```diff
@@ -0,0 +1,45 @@
...
+        $this->addSql('ALTER TABLE pim_catalog_family_variant_has_variant_attribute_sets DROP FOREIGN KEY FK_1F4DC702D8404D');
+        $this->addSql('ALTER TABLE pim_catalog_family_variant_has_variant_attribute_sets ADD CONSTRAINT FK_1F4DC7028A37AD0 FOREIGN KEY (family_variant_id) REFERENCES pim_catalog_family_variant (id) ON DELETE CASCADE');
```
rouliane: As discussed together, i don't know if the index id is only specific to your mysql instance and if it will work on another mysql instance. I will test it on my computer.

rouliane: The key ids are the same on my computer, i don't understand why, but it works :)

PR #8178, opened 2018-05-24
PIM-7373 Avoid deleting and reinsert all attributes relation at family save

BitOne:
This PR avoid generating lots of DELETE and INSERT queries by removing or adding only attributes that have been removed or added to the family at save time.

| Changelog updated                 | Done

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/CHANGELOG-1.7.md
+++ b/CHANGELOG-1.7.md
@@ -1,3 +1,8 @@
+# 1.7.x
+
+## Bug fixes
+ - PIM-7373: Fix deletion and reinsertion of all attributes relations at family save time
+
 # 1.7.23 (2018-06-25)
 
 ## Bug fixes
--- a/src/Pim/Component/Catalog/Updater/FamilyUpdater.php
+++ b/src/Pim/Component/Catalog/Updater/FamilyUpdater.php
@@ -336,12 +336,25 @@ class FamilyUpdater implements ObjectUpdaterInterface
      */
     protected function addAttributes(FamilyInterface $family, array $data)
     {
+        $currentAttributeCodes = [];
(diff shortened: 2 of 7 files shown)


src/Pim/Component/Catalog/Updater/FamilyUpdater.php
```diff
@@ -336,12 +336,29 @@ protected function createAttributeRequirement(
      */
     protected function addAttributes(FamilyInterface $family, array $data)
     {
+        $currentAttributeCodes = [];
+        $wantedAttributeCodes = [];
+
+        foreach ($data as $attributeCode) {
```
ahocquard: is it necessary? can't we just name the parameter "data" as "wantedAttributeCodes" directly?

CHANGELOG-1.7.md
```diff
@@ -1,3 +1,8 @@
+# 1.7.x
+
+## Bug fixes
+ - SDS-3606: Fix deletion and reinsertion of all attributes relations at family save time
```
ahocquard: PIM-7373

src/Pim/Component/Catalog/Updater/FamilyUpdater.php
jmleroux: or `$wantedAttributeCodes = array_values($data);` if it's just a matter of key.

LaurentPetard:
There's a side-effect to the fix : there's no versioning on the family if only attribute requirements are changed. 

In this case the Doctrine unit of work doesn't detect any change on the family. So I added a "Guesser" to mark the family as updated when an attribute requirement is added or removed.

PR #8182, opened 2018-05-24
PIM-7364: use catalog locale to display categories tree

jmleroux:
Currently, the user ui (interface) locale is used to display the category tree. We must use the catalog locale for this tree.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Pim/Bundle/EnrichBundle/Resources/public/js/mass-edit/form/product/category.js
+++ b/src/Pim/Bundle/EnrichBundle/Resources/public/js/mass-edit/form/product/category.js
@@ -84,7 +84,7 @@ define(
 
                         this.$el.html(this.template({
                             i18n: i18n,
-                            locale: UserContext.get('uiLocale'),
+                            locale: UserContext.get('catalogLocale'),
                             trees: trees,
                             currentTree: _.findWhere(trees, {code: this.currentTree}),
                             selectedCategories: this.selectedCategories


jmleroux:
> Don't we have test for this?

I ope we have. I'm running the CI to check. If it's green out of the box, then i'll probable have to add tests.

jmleroux:
> Need a test I think

Yep, and it will be easy easy thanks to you : #8078

:+1:

features/localization/mass-action/classify_product.feature
```diff
@@ -0,0 +1,48 @@
...
+    And the following products:
+      | sku       | categories |
+      | bigfoot   | foocat     |
+      | horseshoe | foocat     |
+    And I am logged in as "Julia"
+    And I am on the products grid
+
+  @info https://akeneo.atlassian.net/browse/PIM-7364
+  Scenario: Mass edit categories in the catalog locale
+    Given I select rows bigfoot and horseshoe
```
ahocquard: imho, this kind of test should not exist. It's a whole and long end-to-end test, hardly readable, without any big value.

jmleroux: @ahocquard 
I didn't go the tests workshop...
What is the right way to test such a feature now?

jmleroux:
> GTM. Same question than Alex for the test, no opinion, you can merge with or without this test.

Should be a front test, but we haven't them in 2.0

features/localization/mass-action/classify_product.feature
jmleroux: > I don't agree with the other to add tests for such things.

:+1: 
I remove it.

ahocquard: Ideally, a front-end test is enough. Or no test as you did (👍 ). Meaning maybe more bugs, but very fast to fix.
We can't guarantee 100% quality, even with end-to end tests.

PR #8184, opened 2018-05-24
PIM-7362: Fix error to save and detach at batch size

fitn:
Fix memory leak on completeness.
Batch size is not taken into account as the assertion is wrong.

`OK`: Done / Validated
`-`: Not needed


CHANGELOG-2.0.md
```diff
@@ -1,3 +1,7 @@
+# 2.0.*
```
damien-carcel: .x or MichelTag will not understand.

PR #8201, opened 2018-05-28
PIM-7375: Fix metric unit values on export/import with empty values

Doodoune:
**Description**

When we export a product with a metric value like that : 
{ 'amout' => null, 'unit' => 'an-unit'}

& then we imported the export we have some errors & the product is skipped : 
values[maximum_scan_size].unit: Please specify a valid metric unit: 0.0000
values[maximum_scan_size].metric.data: This value should be a valid number.: 0.0000
values[maximum_scan_size].metric.data: This value should be a valid number.: 0.0000

For the export, if an amount of a metric is empty, we don't export the unit. 

`OK`: Done / Validated
`-`: Not needed


Doodoune:
I hesitate to make the change on the saver of the metric value - like when we have an empty amount, we save "null" for the metric value instead of { 'amout' => null, 'unit' => 'an-unit'}
But I think if the user select an unit & delete his amount, we have to keep the selected unit...

So I'm not sure if it what the best to do. Can I have your opinion on that ?

src/Pim/Component/Catalog/Normalizer/Standard/Product/MetricNormalizer.php
```diff
@@ -31,9 +32,13 @@ public function normalize($metric, $format = null, array $context = [])
                 ? number_format($amount, static::DECIMAL_PRECISION, '.', '') : (int) $amount;
         }
 
+        if (null == $amount) {
```
phaseinducer: Why not just an early return at the very beginning like this `return ['amount' => null, 'unit'   => null];` ?
Much more understandable imho.

PR #8206, opened 2018-05-29
[PIM-7367] Fix associations of product and product model with same ids

pierallard:
There is an issue when you associate a product and a product model with the same id.
The Datagrid do not display duplicates with the same "id" field (not used in the association grid).
We just added a specific label for splitting product and product models.

### DIFF
--- a/src/Pim/Bundle/DataGridBundle/Datasource/AssociatedProductDatasource.php
+++ b/src/Pim/Bundle/DataGridBundle/Datasource/AssociatedProductDatasource.php
@@ -176,6 +176,22 @@ class AssociatedProductDatasource extends ProductDatasource
     }
 
     /**
+     * Generates a unique identifier for a product or a product model.
+     *
+     * @param $productOrProductModel ProductInterface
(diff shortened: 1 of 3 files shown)


src/Pim/Bundle/DataGridBundle/Datasource/AssociatedProductDatasource.php
```diff
@@ -175,6 +175,22 @@ protected function getAssociatedProductModels(
...
     }
 
+    /**
+     * Generates a unique identifier for a product or a product model.
+     *
+     * @param $productOrProductModel ProductInterface
+     *
+     * @return string
+     */
+    protected function getProductOrProductModelIdentifier($productOrProductModel)
```
grena: typehint

grena:
For the reference: I just saw that the fix was incomplete as we need to update the `\Pim\Bundle\DataGridBundle\Datasource\AssociatedProductModelDatasource` too. It's done in #8286

PR #8207, opened 2018-05-29
Fix UserContext::toArray() method

mmetayer:
**Description (for Contributor and Core Developer)**

This PR fixes `Pim\Bundle\UserBundle\Context\UserContext::toArray()` method, which wrongly provides channels' labels instead of codes since #6374 (in [here](https://github.com/akeneo/pim-community-dev/pull/6374/commits/b8df6efac6d3fd2ed4a8476cd4ad6256894753fd#diff-02a42b2a6d7255a5300ae9d4b57f6a49R31))

**Definition Of Done (for Core Developer only)**

| Q                                 | A
| --------------------------------- | ---
| Added Specs                       | OK
| Added legacy Behats               | -
| Added acceptance tests            | -
| Added integration tests           | -
| Changelog updated                 | -
| Review and 2 GTM                  | Todo
| Micro Demo to the PO (Story only) | Todo
| Migration script                  | -
| Tech Doc                          | -

`Todo`: Pending / Work in progress
`OK`: Done / Validated
`-`: Not needed

PR #8208, opened 2018-05-29
PIM-7382: Fix scopable attributes disappearing from edit form after editing a product model

mmetayer:
In `Pim\Bundle\EnrichBundle\Controller\Rest\ProductModelController::postAction()`, the normalization context was different from the other actions (no `filter_types` key provided, which resulted in applying the default filters: https://github.com/akeneo/pim-community-dev/blob/2f4f43f99f864d725b3515d59b6c0243fe9ddf5e/src/Pim/Component/Catalog/Normalizer/Standard/ProductModelNormalizer.php#L49

One of these filters is `pim_enrich.filter.product_value.channel`, which filters values for scopable attributes when the value's channel is not in the user context. 

Problem is: the context's channels are provided by the `Pim\Bundle\UserBundle\Context::toArray()` method which is wrong (it returns chanlles' labels instead of codes, see #8207)
This results in every scopable value being filtered unless channel code and label are exactly the same, and the corresponding fields disappearing from the edit form after saving.

This PR fixes several small issues in `Pim\Bundle\EnrichBundle\Controller\Rest\ProductModelController`:
- use the same normalization context (by using `normalizeProductModel()` method) for each action
- remove useless filters from the normalization context (the correct filters are applied in the EE's ProductModelRepository)
- reuse `findProductModelOr404()` protected method to avoid code duplication

| Added legacy Behats               | Yes

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/features/product-model/edit_product_model.feature
+++ b/features/product-model/edit_product_model.feature
@@ -91,3 +91,13 @@ Feature: Edit a product model
     And I press the "Save" button
     And I visit the "All" group
     Then the Care instructions, Collection, Model picture fields should be highlighted
+
+  @jira https://akeneo.atlassian.net/browse/PIM-7382
+  Scenario: Successfully display a product model's scopable value after editing it 
+    Given I am logged in as "Mary"
+    And I am on the "bacchus" product model page
+    And I fill in the following information:
+      | Model description | Another great description |
+    And I press the "Save" button
+    Then I should not see the text "There are unsaved changes."
+    And the field Model description should contain "Another great description"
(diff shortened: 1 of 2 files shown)


features/product-model/edit_product_model.feature
```diff
@@ -91,3 +91,13 @@ Feature: Edit a product model
     And I press the "Save" button
     And I visit the "All" group
     Then the Care instructions, Collection, Model picture fields should be highlighted
+
+  @jira https://akeneo.atlassian.net/browse/PIM-7382
+  Scenario: Successfully edit and display a product model's scopable value
```
arnolanglade: This title does not describe what you try to fix.

PR #8215, opened 2018-05-30
PIM-7388: [SLA] Completeness filter on product grid for models does not work as expected

arnolanglade:

When you save a product model:

- The product model is saved in database
- The product model is indexed in ES (post save event)
- The completeness is computed for all product model children (post save event)
- The product model indexation is done
- The job which computes product completeness is finished

Problem is that when the indexation is done, the product completeness computation job is not finished. That means the ratio of complete product models may change during this task so we need to index again the product model. If we don't do that it may break the complete filter on the grid because wrong data was indexed in ES.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Pim/Bundle/CatalogBundle/Doctrine/Common/Saver/ProductModelDescendantsSaver.php
+++ b/src/Pim/Bundle/CatalogBundle/Doctrine/Common/Saver/ProductModelDescendantsSaver.php
@@ -66,7 +66,7 @@ class ProductModelDescendantsSaver implements SaverInterface
         CompletenessManager $completenessManager,
         BulkIndexerInterface $productIndexer,
         BulkIndexerInterface $bulkProductModelIndexer,
-        IndexerInterface $productModelIndexer
+        IndexerInterface $productModelIndexer = null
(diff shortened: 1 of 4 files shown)


src/Pim/Bundle/CatalogBundle/Doctrine/Common/Saver/ProductModelDescendantsSaver.php
```diff
@@ -46,28 +47,34 @@ class ProductModelDescendantsSaver implements SaverInterface
...
      */
     public function __construct(
         ObjectManager $entityManager,
         ProductModelRepositoryInterface $productModelRepository,
         ProductQueryBuilderFactoryInterface $pqbFactory,
         CompletenessManager $completenessManager,
         BulkIndexerInterface $productIndexer,
-        BulkIndexerInterface $productModelIndexer
+        BulkIndexerInterface $bulkProductModelIndexer,
+        IndexerInterface $productModelIndexer
```
skeleton: add `= null` to avoid a BC

PR #8217, opened 2018-05-30
CI for 1.7 : be Back

AymericPlanche:

Since the last cluster update, some cluster features has been broken.
This PR fix the broken feature and use the auto-scale pools for behat and integration test.
 
`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -460,7 +460,7 @@ stage("Test") {
         clearTemplateNames()
         def uuid = UUID.randomUUID().toString()
         podTemplate(label: "cleanup-" + uuid, containers: [
-            containerTemplate(name: "docker", image: "paulwoelfel/docker-gcloud:latest", ttyEnabled: true, command: 'cat', envVars: [envVar(key: "DOCKER_API_VERSION", value: "1.23")], resourceRequestCpu: '100m', resourceRequestMemory: '200Mi')
+            containerTemplate(name: "docker", image: "google/cloud-sdk:198.0.0", ttyEnabled: true, command: 'cat', envVars: [envVar(key: "DOCKER_API_VERSION", value: "1.23")], resourceRequestCpu: '100m', resourceRequestMemory: '200Mi')
         ]) {
             node("cleanup-" + uuid) {
                 container("docker") {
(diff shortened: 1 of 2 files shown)


.ci/Jenkinsfile
```diff
@@ -457,7 +460,7 @@ stage("Test") {
         clearTemplateNames()
         def uuid = UUID.randomUUID().toString()
         podTemplate(label: "cleanup-" + uuid, containers: [
-            containerTemplate(name: "docker", image: "paulwoelfel/docker-gcloud:v1.13", ttyEnabled: true, command: 'cat', envVars: [envVar(key: "DOCKER_API_VERSION", value: "1.23")], resourceRequestCpu: '100m', resourceRequestMemory: '200Mi')
+            containerTemplate(name: "docker", image: "paulwoelfel/docker-gcloud:latest", ttyEnabled: true, command: 'cat', envVars: [envVar(key: "DOCKER_API_VERSION", value: "1.23")], resourceRequestCpu: '100m', resourceRequestMemory: '200Mi')
```
rouliane: noob question : using the latest tag isn't "dangerous" ? (the image can be updated one day and have errors)

BitOne: Good remark, and it happened in the past (Apache image without MPM-prefork enabled for example).

AymericPlanche: the maintainer do not tag frequently https://hub.docker.com/r/paulwoelfel/docker-gcloud/tags/, i think that you need latests updates @mr-prud  ?

rouliane: @AymericPlanche indeed, the maintener seems to almost never tag its image, so we may have no choice. The latest tag is really required or the 1.13 could be enough ? I fear that what happened last week with the PHP docker images happens again...

mr-prud: Why i choose latest, because in the same time kubernetes tool changes frequently, and BC break is not a problem for them and a kubernetes version could be deprecated very soon, in few week/month, if you want i can define to the latest tag, but i could have issue in other way.
IMYO it's more confortable like this, but i change the latest fix tag

mr-prud: I use an other image, coming from google with a fixed version, the change works the ci still running

PR #8228, opened 2018-05-31
PIM-7386: Fix 'NOT IN' operator not taking empty values into account for select fields

mmetayer:
When using 'NOT IN' operator with `Pim\Bundle\CatalogBundle\Elasticsearch\Filter\Attribute\OptionFilter`, an 'exists' filter clause for the field is appended to the query. This results in products with an empty value for the attribute (either with a null/empty array value, or not indexed yet in ES) not being returned by the query.
This PR removes this filter clause.

Downside: Now the query returns: 
- products with this attribute but with a different value as the one(s) provided
- products with an empty value for this attribute
- BUT also products **without this attribute** (that's also the current behaviour of other filter/operator couples - e.g EMPTY operator with most of the attribute filters)

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/CHANGELOG-2.2.md
+++ b/CHANGELOG-2.2.md
@@ -1,7 +1,11 @@
 # 2.2.x
 
+## Bug fixes
+
 - PIM-7316: Fix overlap of boolean fields on product edit form
 - PIM-7319: Fix association display on product edit form when managing the association type permissions
+- PIM-7382: Fix scopable attributes disappearing from edit form after editing a product model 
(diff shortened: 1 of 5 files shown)


jmleroux:
Indeed, it's clearly stated in the doc: https://www.elastic.co/guide/en/elasticsearch/reference/5.4/query-dsl-exists-query.html

> Returns documents that have at least one non-null value in the original field:

GTM then.

PR #8231, opened 2018-05-31
PIM-7311: Fix the product grid filters list with a sort order is a huge number

pierallard:
Previously, when you set a sort order like 100000000 on an attribute group, the grid was unable to load.
Why ? Because on the function sorting the attribute groups, it generated an array of 1000000000 elements... And it crashed.
Second thing is that if 2 attribute groups had the same sort order, it puts the second encountered one at the end of the list.

Now, the sort is a little bit complex, but it works as designed and does not fail.
IMHO, this is not "easily" testable (and have no interest)

### DIFF
--- a/src/Pim/Bundle/DataGridBundle/Resources/public/js/datafilter/filters-manager.js
+++ b/src/Pim/Bundle/DataGridBundle/Resources/public/js/datafilter/filters-manager.js
@@ -300,12 +300,13 @@ define(
             if (_.isEmpty(this.filters)) {
                 this.$el.hide();
             } else {
+                this._fillEmptyGroupFilters();
                 this.$el.append(fragment);
                 if (this.displayManageFilters()) {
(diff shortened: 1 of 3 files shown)


src/Pim/Bundle/DataGridBundle/Resources/public/js/datafilter/filters-manager.js
```diff
@@ -315,6 +316,53 @@ define(
...
+            })).concat(unsortedGroups);
+        },
+
+        /**
+         * Fills the filters having no group to put it into the "System" one.
+         * This method is just for display, it's not registered in database, as the "system" group does not exist.
+         */
+        _fillEmptyGroupFilters: function () {
+            Object.keys(this.filters).forEach((filterKey) => {
+                console.log(this.filters[filterKey].group);
```
nidup: to be dropped

PR #8235, opened 2018-06-01
PIM-7393: Improve error message when importing fields without locale or scope specification

pierallard:
The previous error message was really confusing.
Now, it is more explicit, explaining why the field is "unknown".

| Added Specs                       | y
| Added legacy Behats               | y
| Micro Demo to the PO (Story only) | y

### DIFF
--- a/src/Pim/Component/Connector/ArrayConverter/FlatToStandard/Product.php
+++ b/src/Pim/Component/Connector/ArrayConverter/FlatToStandard/Product.php
@@ -314,25 +314,8 @@ class Product implements ArrayConverterInterface
             }
         }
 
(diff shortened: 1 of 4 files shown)


src/Pim/Component/Connector/ArrayConverter/FlatToStandard/Product.php
```diff
@@ -314,10 +314,43 @@ protected function validateOptionalFields(array $item): void
             }
         }
 
+        $nonLocalizableOrScopableFields = [];
```
jjanvier: possible to extract this part somehow in a private method?

PR #8263, opened 2018-06-05
PIM-7308: bulk associate to product models

jmleroux:
Currently, when we try to bulk associate products to product model, it will in fact make associations with all the variant products.
We now can associate directly to product models so we don't need to associate to it's children.
This is done very simply just by using the 'grouped' reader which will apply associations on the selected items without fetching the children.

| Added legacy Behats               | Updated

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/tests/legacy/features/Context/Page/Batch/Operation.php
+++ b/tests/legacy/features/Context/Page/Batch/Operation.php
@@ -26,7 +26,7 @@ class Operation extends Wizard
         'Move between categories'          => 'Batch Classify',
         'Remove from categories'           => 'Batch Classify',
         'Add to an existing product model' => 'Batch AddToExistingProductModel',
-        'Associate to products'            => 'Batch AssociateToProduct'
+        'Associate'                        => 'Batch AssociateToProduct',
     ];
 
     /**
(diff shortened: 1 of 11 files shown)


jjanvier:
I'm totally against such developer shortcuts. 

Basically, you introduce a business concept that does not exist, the `ProductOrProductModel`, just in order to be able reuse some code somewhere else (your `RawProductReader`).

[DRY is about functional knowledge, not about technical code](http://verraes.net/2014/08/dry-is-about-knowledge/).
Copying/pasting a few lines (for 2 different readers for instance) is not a big deal. Creating a bad abstraction can become a nightmare.

We should not create business concepts that do not exist. Product and product models are different. Even if they a few similar behaviors.

tests/legacy/features/Context/Page/Batch/Operation.php
```diff
@@ -26,7 +26,8 @@ class Operation extends Wizard
         'Move between categories'          => 'Batch Classify',
         'Remove from categories'           => 'Batch Classify',
         'Add to an existing product model' => 'Batch AddToExistingProductModel',
-        'Associate to products'            => 'Batch AssociateToProduct'
+        'Associate to products'            => 'Batch AssociateToProduct',
```
grena: Should we still keep this one?

PR #8266, opened 2018-06-05
Reduce catalogs fixtures

skeleton:
I know Juju is working on a PR to avoid to run catalogs on each tests on the CI, that's why I did not create this one to improve the time of the CI but to reduce the time to debug our tests. 
It's really a small improvement but it's clean a little bit useless fixtures (and maybe the next step is to remove some catalogs, always use the `default` and create what we need on feature files).

Before:
default catalog: ~5.6 seconds
apparel catalog: ~7.1 seconds
footwear catalog: ~6.48 seconds
catalog_modeling catalog: ~83 seconds

After:
default catalog: ~3.7 seconds
apparel catalog: ~5.2 seconds
footwear catalog: ~4.6 seconds
catalog_modeling catalog: ~61 seconds

The diff for all catalogs except catalog_modeling is due to the deletion of all documents in ES instead of delete & recreate the index.

`OK`: Done / Validated
`-`: Not needed


tests/legacy/features/Pim/Behat/Context/HookContext.php
```diff
@@ -292,15 +283,19 @@ private function getDoctrine()
...
-    private function resetElasticsearchIndex()
+    private function deleteElasticsearchDocuments()
     {
         $clientRegistry = $this->getService('akeneo_elasticsearch.registry.clients');
         $clients = $clientRegistry->getClients();
 
         foreach ($clients as $client) {
-            $client->resetIndex();
+            $client->deleteByQuery([
+                'query' => [
```
ahocquard: is it asynchronous?

skeleton: how can be sure the index is clean before continue the installation? with refresh? cc @SamirBoulil

SamirBoulil: Reading the documentation it seems that refresh will ensure this but I never tried it myself.

skeleton: I know we do that on our integration tests https://github.com/akeneo/pim-community-dev/blob/2.3/src/Pim/Bundle/EnrichBundle/tests/integration/PQB/Filter/CategoryFilterIntegration.php#L287 to avoid random

ahocquard: Refresh should be ok, but I'm wondering if it worths to do it in term of performance gain. Resetting index seems to be more appropriate but I'm maybe wrong.

skeleton: I will rebench all catalogs to check if it's better/worth/same.
For the moment, I'm creating 1000000 products to check if the refresh fix the problem of asynchronous (btw it's so slow :/ )

ahocquard:
Note: we already deleteByquery in a very efficient way on multiple indexes at once.
And we use a dump for integration and tests now.

That's why this PR is not relevant anymore.

PR #8284, opened 2018-06-07
Pim 7384

docteurklein:
Fix quick export memory leak

### DIFF
--- a/CHANGELOG-2.2.md
+++ b/CHANGELOG-2.2.md
@@ -3,6 +3,7 @@
 ##Bug fixes
 
 - PIM-7399: Fix attributes order on Product model export
+- PIM-7384: Fix Memory leak on Quick export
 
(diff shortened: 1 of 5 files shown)


jjanvier:
what are the results of those changes? 
do you have number to share?

please add the ticket number in your commit

CHANGELOG-2.2.md
```diff
@@ -1,5 +1,7 @@
 # 2.2.x
 
+- PIM-7384: Fix Memory leak on Quick export
```
Doodoune: Add it to "bug fixes"

docteurklein:
the product reader used to iterate over product and product models and then filter out all the unwanted results. We now filter them out at the query level, so we don't have to hydrate them all.

We went from 10MB/sec memory leak to 0.

PR #8304, opened 2018-06-11
PIM-7420: Keep item ordering in item picker

damien-carcel:
## Description

Items in the item picker (only used today for the assets in EE) should be organized in the order they are selected.

However, this is not working if the codes of the entities are integers and not strings. As the fetcher creates an object using the item codes as key, when those codes are integers, the items are reorganized, automatically sorted by code ascending, even if the items were already sorted by the back-end.

So the only way is to reorder the items according to the original list of codes provided to the fetcher.

## Definition Of Done

`OK`: Done / Validated
`-`: Not needed


src/Pim/Bundle/EnrichBundle/Resources/public/js/common/item-picker.js
```diff
@@ -310,12 +310,16 @@ define(
...
-                FetcherRegistry.getFetcher(this.config.fetcher).fetchByIdentifiers(this.getItems())
-                    .then(function (items) {
-                        this.renderBasket(items);
-                        this.delegateEvents();
-                    }.bind(this));
+            updateBasket: function (itemCodes) {
+                FetcherRegistry.getFetcher(this.config.fetcher).fetchByIdentifiers(this.getItems()).then(items => {
+                    let orderedItems = [];
+                    itemCodes.forEach(itemCode => {
+                        orderedItems = orderedItems.concat(items.filter(item => item.code === itemCode));
```
rouliane: i'm not good in JS, but i'm not sure what the line does... It erase orderedItems in each iteration ?

damien-carcel: it is the same than an array_merge in PHP

fitn: Could you explain to me what this code is used for?
In the doc, it's written fetch the new items and render the basket. To me, if it is about adding new items, it should be a push and not a merge.

damien-carcel: Indeed, a `push` would be way better. I forgot it exists.

damien-carcel: A push cannot work, as I use an `array.filter` which will return an array.
Or I need to perform an `array.push(resultOfFilter[0])`.

damien-carcel: And for the explanation: it does what I told in the description. The `orderedItems` are a new list of the elements of `items`, but ordered by their codes (meaning `item.code`) as those are in `itemCodes`.

anaelChardan: Maybe you could do ```orderedItems.flatMap(itemCode => items.filter(item => item.code === idemCode))``` which is a little bit more functional and is immutable.

damien-carcel: mh. MDN lists it as an experimental technology. And for now, only Chrome supports it, and Firefox only on nightly builds, so not an option.

damien-carcel: What underscore method are you thinking of? If there is one that can shorten this writing, I am in for, but if it is just syntax sugar for the exact same operations, I prefer keep native javascript.

damien-carcel: Then it will come, one day :-)

fitn: Please, as discussed yesterday together, add a few comments to explain why we are doing that.

damien-carcel: Done

PR #8340, opened 2018-06-18
PIM-7345 : Remove "is_empty" operator on the sku filter.

christophe-chausseray:

Following the behavior error we have on the parent filter, we can't remove the 'is_empty' from the operator choice in the TextFilterType. So, we had to add a specific js filter for the identifier.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Pim/Bundle/DataGridBundle/Resources/public/js/datafilter/filter/identifier-filter.js
+++ b/src/Pim/Bundle/DataGridBundle/Resources/public/js/datafilter/filter/identifier-filter.js
@@ -1,0 +1,38 @@
+/* global define */
+define(['underscore', 'oro/translator', 'oro/datafilter/choice-filter'],
+  function (_, __, ChoiceFilter) {
+    'use strict';
+
+    /**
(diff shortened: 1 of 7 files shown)


src/Pim/Bundle/DataGridBundle/Resources/public/js/datafilter/filter/identifier-filter.js
```diff
@@ -0,0 +1,38 @@
...
+      /**
+       * {@inheritdoc}
+       */
+      _getOperatorChoices() {
+        return {
+          '1': __('pim.grid.choice_filter.label_contains'),
+          '2': __('pim.grid.choice_filter.label_does_not_contain'),
+          '3': __('pim.grid.choice_filter.label_equal'),
+          '4': __('pim.grid.choice_filter.label_start_with'),
+          'in': __('pim.grid.choice_filter.label_in_list'),
```
jmleroux: It would be way much easier to read with some kind of constants like in the PHP code.

PR #8344, opened 2018-06-18
PIM-7442: Bulk actions ALL - count not taking into account variants

jmleroux:
When selecting "all" for bulk actions AND a filter on a parent identifier to have only the variants of a certain product model, the product count displayed in the bulk action confirmation equals 0.
This is because we don't have the `parent` field indexed in the products index and we must use the product and product model index where the field `parent` exists.

Note: we could also fix the product index, but it's less doable in a patch.

This PR uses the product and product model index and filter on product entity types to achieve the same filter result.

| Added Specs                       | Fixed

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/CHANGELOG-2.2.md
+++ b/CHANGELOG-2.2.md
@@ -1,3 +1,9 @@
+# 2.2.x
+
+## Bug fixes
+
+- PIM-7442: Bulk actions ALL - count not taking variant products into account
+
 # 2.2.9 (2018-06-14)
 
 ## Bug fixes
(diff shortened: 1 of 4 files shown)


CHANGELOG-2.2.md
```diff
@@ -1,3 +1,9 @@
+# 2.2.x
+
+## Bug fixes
+
+- PIM-7442: Bulk actions ALL - count not taking into account variant products
```
mmetayer: not taking variant products into account

src/Pim/Bundle/EnrichBundle/Doctrine/ORM/Query/CountImpactedProducts.php
```diff
@@ -135,7 +135,7 @@ private function countUnSelectedProducts(array $filters): int
     }
```
SamirBoulil: if the productQueryBuilderFactory is not used anymore, maybe put it optional so that we remove it on master ?

PR #8368, opened 2018-06-21
Release 2.3: Fix migrations and add missing

SamirBoulil:

- Make sure the indexing migration is done the last.
- Add missing migrations.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/upgrades/schema/Version_2_3_20180620102817_add_table_product_model_associations.php
+++ b/upgrades/schema/Version_2_3_20180620102817_add_table_product_model_associations.php
@@ -1,0 +1,42 @@
+<?php
+
+namespace Pim\Upgrade\Schema;
+
+use Doctrine\DBAL\Migrations\AbstractMigration;
+use Doctrine\DBAL\Schema\Schema;
+
+/**
+ * Add tables needed to support the product model association feature
+ */
+class Version_2_3_20180620102817_add_table_product_model_associations extends AbstractMigration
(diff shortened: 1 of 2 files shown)


upgrades/schema/Version_2_3_20180620102817_add_table_product_model_associations.php
```diff
@@ -0,0 +1,51 @@
...
+        $this->addSql('ALTER TABLE pim_catalog_association_product_model_to_product_model ADD CONSTRAINT FK_12D4D59CB2C5DD70 FOREIGN KEY (product_model_id) REFERENCES pim_catalog_product_model (id) ON DELETE CASCADE');
+    }
+
+    /**
+     * @param Schema $schema
+     */
+    public function down(Schema $schema)
+    {
+        // this down() migration is auto-generated, please modify it to your needs
+        $this->abortIf($this->connection->getDatabasePlatform()->getName() !== 'mysql', 'Migration can only be executed safely on \'mysql\'.');
```
grena: I don't think we write the `down` method of migrations to avoid data loss, but to confirm :point_up:

SamirBoulil: There is no dataloss because those are new tables, but you are right we don't do it in other migrations so. I'll just remove this piece of code.

PR #8392, opened 2018-06-25
PIM-7459 Fix perf issue by removing combinatory explosion

BitOne:
This PR does two things:
 - avoid generating a high volume of data from a combinatory explosion from the family repository
 - avoid reloading all families information for each product

These two fixes can reduce import time by x10 on a situation where there's a high number of locales (around 30 in the examples).

### DIFF
--- a/src/Pim/Bundle/CatalogBundle/Doctrine/MongoDBODM/CompletenessGenerator.php
+++ b/src/Pim/Bundle/CatalogBundle/Doctrine/MongoDBODM/CompletenessGenerator.php
@@ -40,6 +40,18 @@ class CompletenessGenerator implements CompletenessGeneratorInterface
     /** @var FamilyRepositoryInterface */
     protected $familyRepository;
 
+    /** @var array All whole Families requirements indexed by ids */
+    protected $allFamilyRequirements;
+
+    /** @var array Families requirements by family ids */
+    protected $familyRequirements;
+
+    /** @var array Families requirements by channels and family ids */
+    protected $allPartialFamilyRequirements;
+
+    /** @var array Families requirements indexed by channels and ids */
+    protected $partialFamilyRequirements;
+
     /**
      * Constructor
      *
@@ -58,6 +70,10 @@ class CompletenessGenerator implements CompletenessGeneratorInterface
         $this->productClass = $productClass;
         $this->channelRepository = $channelRepository;
         $this->familyRepository = $familyRepository;
+
+        $this->familyRequirements = [];
+        $this->allPartialFamilyRequirements = [];
+        $this->partialFamilyRequirements = [];
     }
 
     /**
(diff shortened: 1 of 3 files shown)


src/Pim/Bundle/CatalogBundle/Doctrine/MongoDBODM/CompletenessGenerator.php
```diff
@@ -40,6 +40,18 @@ class CompletenessGenerator implements CompletenessGeneratorInterface
     /** @var FamilyRepositoryInterface */
     protected $familyRepository;
 
+    /** @var array All whole Families requirements indexed by ids */
+    protected $allFamilyRequirements;
+
+    /** @var array Families requirements by family ids */
+    protected $familyRequirements;
```
anaelChardan: Why isn't it named like the comment? ;) same for other variables ;)

src/Pim/Bundle/CatalogBundle/Doctrine/MongoDBODM/CompletenessGenerator.php
```diff
@@ -253,6 +269,91 @@ protected function saveCompletenesses($productId, array $compData)
...
+     *
+     * @param ProductInterface $product
+     *
+     * @return array
+     */
+     protected function getWholeFamilyRequirements(ProductInterface $product)
+     {
+        if (null === $product || null === $product->getFamily()) {
+            if (null === $this->allFamilyRequirements) {
+                $this->allFamilyRequirements = $this->computeFamilyRequirements();
```
anaelChardan: compute could do this job like "mnemoization" no?

BitOne: What ? Didn't understand this one ;)

anaelChardan: Why do we have to check the nullness if it is always fulfilled by computeFamilyRequirements, you could always ```return $this->computeFamilyRequirements``` ;) it's just to remove some checks here to help the readability. ;)

src/Pim/Bundle/CatalogBundle/Doctrine/MongoDBODM/CompletenessGenerator.php
BitOne: Yes, missing the variable name. Only the explanation is there. no need for the variable name.

anaelChardan: I find it difficult because I need to read the comment to know how the variable works, maybe I would have find ```$familyRequirementsById``` easier, but maybe it is only me ;)

PR #8393, opened 2018-06-25
PIM-7465: Set front-end entity into field context

damien-carcel:
## Description

Until now, product and product model fields were not aware of the entity they belonged too.

With this PR, we send to the form data (meaning, the edited entity) into the field context..

## Definition Of Done

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Pim/Bundle/EnrichBundle/Resources/public/js/form/common/attributes.js
+++ b/src/Pim/Bundle/EnrichBundle/Resources/public/js/form/common/attributes.js
@@ -263,10 +263,11 @@ define(
                         AttributeManager.isOptional(field.attribute, object)
                     );
                 }).then(function (field, channels, isOptional) {
-                    var scope = _.findWhere(channels, { code: UserContext.get('catalogScope') });
-                    var locale = UserContext.get('catalogLocale');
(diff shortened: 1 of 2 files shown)


src/Pim/Bundle/EnrichBundle/Resources/public/js/form/common/attributes.js
```diff
@@ -263,10 +263,16 @@ define(
...
-                    var locale = UserContext.get('catalogLocale');
+                    const scope = _.findWhere(channels, { code: UserContext.get('catalogScope') });
+                    const locale = UserContext.get('catalogLocale');
+                    const entityType = this.getFormData().meta.model_type;
+                    const entityIdentifier = 'product_model' === entityType
+                        ? this.getFormData().code
+                        : this.getFormData().identifier;
 
                     field.setContext({
+                        entityType: entityType,
```
phaseinducer: I'd rather pass the whole entity as we could need anything. For example in the onboarder we'll need the family.
What do you think ?

damien-carcel: You mean the whole getFormData() result? Honnestly no opinion on that. @juliensnz what do you think?

PR #8395, opened 2018-06-25
Improve usage of docker-compose

damien-carcel:
## Description

This PR aims to finish the clean up the `docker-compose.yml` file on `master` branch, and keep in it only what is mandatory to have a working PIM. As the port mapping of the Apache container is mandatory, it was placed in a previous PR in a `.env.yml.dist` file that the developer needs to copy and configure first (this way, the Apache output ports can still be configured).

Every other configuration is to be placed in `docker-compose.override.yml`, at the developer's convenience. A `.dist` example is added.

If this PR is accepted, it will be ported to EE and standard editions, and documentation will be updated.

## Definition Of Done

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -2,96 +2,86 @@ version: '2'
 
 services:
   fpm:
-    image: akeneo/fpm:php-7.1
-    environment:
-      COMPOSER_HOME: '/home/docker/.composer'
-      BEHAT_TMPDIR: '/srv/pim/var/cache/tmp'
-    user: docker
+    image: 'akeneo/fpm:php-7.1'
+    user: 'docker'
     volumes:
-      - ./:/srv/pim
-      - ~/.composer:/home/docker/.composer
-      - /tmp/behat/screenshots:/tmp/behat/screenshots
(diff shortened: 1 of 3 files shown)


docker-compose.yml
```diff
@@ -2,96 +2,86 @@ version: '2'
 
 services:
   fpm:
-    image: akeneo/fpm:php-7.1
-    environment:
-      COMPOSER_HOME: '/home/docker/.composer'
```
jmleroux: You could keep this one in this file as it depends on our internal docker image.

docker-compose.yml
```diff
@@ -2,96 +2,86 @@ version: '2'
...
     networks:
-      - akeneo
-      - behat
+      - 'akeneo'
+      - 'behat'
 
   node:
-    image: juliensnz/node
-    environment:
-      YARN_CACHE_FOLDER: '/home/node/.yarn-cache'
```
jmleroux: same, you could keep it here

damien-carcel: Seen IRL.

PR #8407, opened 2018-06-26
add acceptance tests path in CS fixer configuration

rouliane:

`OK`: Done / Validated
`-`: Not needed


.php_cs.php
```diff
@@ -16,5 +16,6 @@
             ->notName('*Integration.php')
             ->in(__DIR__ . '/tests/legacy/features')
             ->in(__DIR__ . '/tests/features')
+            ->in(__DIR__ . '/tests/back/Acceptance')
```
grena: Don't we have to update some files in this directory? Or they already respect the CS fixer?

damien-carcel: If it is the case, the CI will tell and be red. But running it locally could help, for sure :-).

rouliane: there was indeed a few files with wrong CS. I did a commit to fix them.

PR #8408, opened 2018-06-27
PIM-7455: Fix product model indexation issue when computing descendants completeness

mmetayer:
Ensures that the product model being passed to `Pim\Bundle\CatalogBundle\Doctrine\Common\Saver\ProductModelDescendantsSaver::save()` is rehydrated before being indexed in order to index the full data. See full description in https://github.com/akeneo/pim-enterprise-dev/pull/4149

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Pim/Bundle/CatalogBundle/Doctrine/Common/Saver/ProductModelDescendantsSaver.php
+++ b/src/Pim/Bundle/CatalogBundle/Doctrine/Common/Saver/ProductModelDescendantsSaver.php
@@ -109,7 +109,8 @@ class ProductModelDescendantsSaver implements SaverInterface
          * @TODO We need to remove the if condition during the pull day
          */
         if (null !== $this->productModelIndexer) {
-            $this->productModelIndexer->index($productModel);
+            $fullProductModel = $this->productModelRepository->findOneByIdentifier($productModel->getIdentifier());
+            $this->productModelIndexer->index($fullProductModel);
         }
     }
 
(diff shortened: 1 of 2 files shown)


src/Pim/Bundle/CatalogBundle/Doctrine/Common/Saver/ProductModelDescendantsSaver.php
```diff
@@ -109,7 +109,8 @@ public function save($productModel, array $options = []): void
          * @TODO We need to remove the if condition during the pull day
          */
         if (null !== $this->productModelIndexer) {
-            $this->productModelIndexer->index($productModel);
+            $fullProductModel = $this->productModelRepository->findOneByIdentifier($productModel->getIdentifier());
```
BitOne: Where are the detach?

BitOne:
Clearly, we are putting problematic over problematic code to fix this bug...

I know it's a SLA, so I don't think we can redo everything, but it's a good occasion to discuss about the indexing process.

By the way

BitOne:
@fitn For the hydration niside the indexer, I would say that's the way to go. But with read model, in order to avoid Unit of Work memory management problem and by speeding things up by using a properly crafted SQL query to get all the data we need for the indexing process with a minimum of process time.

src/Pim/Bundle/CatalogBundle/Doctrine/Common/Saver/ProductModelDescendantsSaver.php
mmetayer: It's in the callers services: `Pim\Component\Connector\Writer\Database\ProductModelDescendantsWriter` and `Pim\Component\Catalog\Job\ComputeProductModelsDescendantsTasklet` (but only in 2.0, the fixes made by @fitn have not been merged in 2.2 yet)

PR #8446, opened 2018-07-04
Bounded context definitions

jjanvier:
Here the definition of each bounded context provider by our dear Product Owners.

I hope it will help people to understand better "what should go where".

(I don't know if we should those kind of files, like `STANDARD_FORMAT.md`, in a particular folder or not. Something like `reference_doc`?)

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -1,0 +1,23 @@
+# Architecture
+
+## Bounded contexts
+
+According to Eric Evans in [Domain-Driven Design Reference](https://domainlanguage.com/ddd/reference/), a bounded context is: 
+
+*"A description of a boundary (typically a subsystem, or the work of a particular team) within which a particular model is defined and applicable."*
+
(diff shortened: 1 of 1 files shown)


ARCHITECTURE.md
```diff
@@ -0,0 +1,23 @@
+# Architecture
+
+## Bounded contexts
+
+According to Eric Evans in [Domain-Driven Design Reference](https://domainlanguage.com/ddd/reference/), a bounded context is:
```
anaelChardan: Maybe add the fact that it belongs to the solution space

anaelChardan: Add also the we have to be clear in relationships between bounded context and it is called "context mapping"

anaelChardan:
IMHO it should be in a dedicated folder as you suggested ;)

ARCHITECTURE.md
jjanvier: > Maybe add the fact that it belongs to the solution space

what would it bring here?

> Add also the we have to be clear in relationships between bounded context and it is called "context mapping"

yes! you're right! I'll do that in another PR with a schema

PR #8452, opened 2018-07-04
PIM-7488: use catalog locale for attributes list in attribute groups

jmleroux:
In Attribute group Edition, attributs labels translation use ui locale when we should use catlaog locale.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/CHANGELOG-2.3.md
+++ b/CHANGELOG-2.3.md
@@ -1,3 +1,7 @@
+# 2.3.x
+
+- PIM-7488: use catalog locale for attributes list in attribute groups
+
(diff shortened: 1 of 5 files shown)


CHANGELOG-2.3.md
```diff
@@ -1,3 +1,7 @@
+# 2.3.x
+
```
mmetayer: Missing "Bug fixes" subtitle

willy-ahva:
In properties Tab label translations should be in UI locale, maybe you could fix it in the same PR

PR #8500, opened 2018-07-19
Fix the CI "java.io.IOException"

LaurentPetard:
## Description

This PR makes use of the new tag `v2.0.0` from `akeneo/jenkins-k8s-utils` in the Jenkinsfile. This new version gets read of the usage of "templates" in the pods, and replaces it by a YAML syntax.

This should fix the `java.io.IOException` we currently have on the CI.

This PR also moves the CI Dockerfile into the `.ci` directory, where it should be.

## Definition Of Done

`OK`: Done / Validated
`-`: Not needed


.ci/Jenkinsfile
```diff
@@ -1,4 +1,6 @@
 #!groovy
+@Library("k8s-utils@yaml")
```
damien-carcel: Should merge the utilities PR and use "k8s-utils@master" here, then remove the line once it is tagged and configured on the CI?

PR #8528, opened 2018-07-24
WIP - PIM-7509: Fix completeness calculation on Product Models bulk events

Doodoune:

To fix an SLA -> the compute product model descendant is not called by the execution of rules in EE 
-> we pass by the SAVE_ALL event and the subscriber was only for the SAVE (don't take in account bulk events)

=> TODO : I've to add tests on EE part for rules

`OK`: Done / Validated
`-`: Not needed


src/Pim/Bundle/CatalogBundle/EventSubscriber/Bulk/BulkComputeProductModelDescendantsSubscriber.php
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    /**
+     * @param GenericEvent $event
+     */
+    public function bulkComputeProductModelDescendantsCompleteness(GenericEvent $event): void
+    {
+        $productModels = $event->getSubject();
+
+        if (count($productModels) === 0){
```
damien-carcel: The case will probably never happen.
You should do this test on `$productModelCodes`, after your foreach.

fitn:
Could you ensure it is not launched for mass actions and import?

Doodoune:
It's launch on Mass action & import I think 
=> it's lauch every time you call a SAVE_ALL (so for bulk actions)

 it's the ticket purpose

src/Pim/Bundle/CatalogBundle/EventSubscriber/Bulk/BulkComputeProductModelDescendantsSubscriber.php
```diff
@@ -0,0 +1,91 @@
...
+
+    /**
+     * @param GenericEvent $event
+     */
+    public function bulkComputeProductModelDescendantsCompleteness(GenericEvent $event): void
+    {
+        $productModels = $event->getSubject();
+
+        $productModelsCodes = [];
+        foreach ($productModels as $productModel){
```
rouliane: You should check if $productModels is at least iterable (via is_iterable() or is_array()) before the foreach.

rouliane: @Doodoune i did the fix and launch a build to help :)

Doodoune: Tks !!

src/Pim/Bundle/CatalogBundle/EventSubscriber/Bulk/BulkComputeProductModelDescendantsSubscriber.php
```diff
@@ -0,0 +1,91 @@
...
+            if (!$productModel instanceof ProductModelInterface) {
+                return;
```
grena: Hm, so it means that if you mass edited products & product models, the job won't be launch for the product models? We should just skip the object if it's not a product model, instead of leaving directly, otherwise descendants of product models won't be re computed.

PR #8555, opened 2018-07-26
PIM-7546: fix the batch size of the cursor for the datagrid

ahocquard:

Tested on a catalog with 1 000 000 products and 20 000 categories.
This fix allows to have a gain of ~800 ms (from 2.5 seconds to 1.7 seconds to get data from the grid).

The PQB requests Elasticsearch 3 times for the datagrid. It's due to the batch size at 10. 
Each ES request takes more than 400 ms. 
For a total of 1.2 seconds, which is a lot. 

Batch size at 10 was done for memory leak purpose due to problems with the garbage collector and circular references during a long running batch.
It's useless for HTTP requests. 

Note: it's the same thing for the API. It's an HTTP call, so we don't care about the garbage collector.

The batch size slows down the whole API for nothing, as we don't handle memory leaks in the API.
We are currently doing 10 requests to ES, for a total of 4 seconds of ES requests, for 100 products.
I think it could be reduced to 400 ms if we adjust the batch size to 100 for the API.

`OK`: Done / Validated
`-`: Not needed


ahocquard:
CI red on asset tests, as for the nightly builds... : I merge it.

PR #8563, opened 2018-07-27
PIM-7453: forbid to remove a role if user won't have role after deletion

skeleton:
Forbid to remove a role if a user won't have role after deletion. 
A user without role has no sense on the PIM. 
Moreover, there is a little thing "funny" on the PIM: if a user without role want to connect to the PIM, the application will try to find the role "ROLE_USER" which is a role defined as a default role (don't ask me why...) and then update the user to associated him to "ROLE_USER" role. That's mean a user without role can have much more access than before to the PIM after an authentification \o/

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Oro/Bundle/UserBundle/Controller/RoleController.php
+++ b/src/Oro/Bundle/UserBundle/Controller/RoleController.php
@@ -58,14 +58,12 @@ class RoleController extends Controller
             throw $this->createNotFoundException(sprintf('Role with id %d could not be found.', $id));
         }
 
-        $em->remove($role);
+        $this->container->get('pim_user.remover.role')->remove($role);
 
         $aclSidManager = $this->get('oro_security.acl.sid_manager');
         if ($aclSidManager->isAclEnabled()) {
             $aclSidManager->deleteSid($aclSidManager->getSid($role));
         }
-
-        $em->flush();
 
         return new JsonResponse('', 204);
     }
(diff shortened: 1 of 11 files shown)


src/Oro/Bundle/UserBundle/Controller/RoleController.php
```diff
@@ -58,15 +59,17 @@ public function deleteAction(Request $request, $id)
...
 
-        $aclSidManager = $this->get('oro_security.acl.sid_manager');
-        if ($aclSidManager->isAclEnabled()) {
-            $aclSidManager->deleteSid($aclSidManager->getSid($role));
+            $aclSidManager = $this->get('oro_security.acl.sid_manager');
+            if ($aclSidManager->isAclEnabled()) {
+                $aclSidManager->deleteSid($aclSidManager->getSid($role));
+            }
+        } catch (\InvalidArgumentException $e) {
+            return new JsonResponse(['message' => $e->getMessage()], 422);
```
rouliane: You should use a constant instead of 422. There should be one in the Symfony Response class.

src/Pim/Bundle/UserBundle/Resources/config/queries.yml
```diff
@@ -0,0 +1,8 @@
+parameters:
+    pim_user.query.is_there_user_without_role.class: Pim\Bundle\UserBundle\Doctrine\ORM\Query\IsThereUserWithoutRole
+
+services:
+    pim_user.query.is_there_user_without_role:
+        class: '%pim_user.query.is_there_user_without_role.class%'
```
rouliane: Shouldn't we directly use the FQDN instead of using a parameter for the class ?

PR #8564, opened 2018-07-27
PIM-7548: split normalizers by format to improve performances

ahocquard:
Before:
3.6 seconds in CE to return 100 products with the API, 380 PV/product.

After:
2.7 seconds in CE to return 100 products with the API, 380 PV/product.

It should probably improve:
- indexation
- persistence into the storage
- export with API
- export with jobs

I will benchmark it later more in details, once every improvement are done.

**Lazy service declaration**

If you read the PR, you will see I use sometimes the "lazy" loading of Symfony services instead of the `setSerializer`.
https://symfony.com/doc/current/service_container/lazy_services.html

**Why?**

If you inject directly the normalizer service (without calling the serializer), the function `setSerializer` will not be called. So, the service will not be correctly initialized. And then, it fails.

Actually, we are doing it quite often.
It was working before only because the serializer `pim_serializer` was called before calling directly the normalizer (so, the initialization was done). But it was not done in purpose: we are just kind of lucky.

The solution is to inject the desired normalizer with the DI, instead of relying on `setSerializer`.  
Then you can use directly the normalizer.
Also, the service is immutable, which is a good thing. And you are sure to have the serializer injected.

But then, you have circular dependencies.
That's why I use the "lazy" option.

Note: using `setSerializer` is actually the same trick as declaring a service as "lazy".

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/CHANGELOG-3.0.md
+++ b/CHANGELOG-3.0.md
@@ -12,6 +12,12 @@
 
 ## BC breaks
 
+- Register standard format normalizers into  `pim_standard_format_serializer` serializer and untagged them from `pim_serializer` serializer 
+- Register indexing normalizers into  `pim_indexing_serializer` serializer and untagged them from `pim_serializer` serializer
+- Register datagrid normalizers into  `pim_datagrid_serializer` serializer and untagged them from `pim_serializer` serializer
+- Register storage normalizers into  `pim_storage_serializer` serializer and untagged them from `pim_serializer` serializer
+- Register external API normalizers into  `pim_external_api_serializer` serializer and untagged them from `pim_serializer` serializer
+
 - Change constructor of `Pim\Component\Catalog\Updater\ProductUpdater`, remove `$supportedFields` argument 
 
 - Move `Pim\Bundle\CatalogBundle\DependencyInjection\Compiler\Localization\RegisterLocalizersPass` to `Akeneo\Pim\Enrichment\Bundle\DependencyInjection\Compiler\Localization\RegisterLocalizersPass`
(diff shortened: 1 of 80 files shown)


ahocquard:
@jjanvier 

**`pim_serializer` serializer**

For now, I know there is still some "flat" normalizers.

For integrators, maybe it's better to let them with this service as well. And it can maybe useful for us for all normalizers without any dedicated and well identified format.

**`pim_catalog.normalizer.storage_with_standard_format.product.product_value`**

The service `pim_catalog.normalizer.storage.product.product_value` was injecting the service 
 `pim_catalog.normalizer.standard.product.product_value` for the storage. 

But it was calling it with the `storage` format, not the standard format...

So, it was working well when normalizing a datetime to the storage format:
- `pim_catalog.normalizer.storage.product.product_value` called with `storage`format
- `pim_catalog.normalizer.standard.product.product_value` called with `storage`format
- storage datetime normalizer found in `pim_serializer` injected in `pim_catalog.normalizer.standard.product.product_value`

After split:
- `pim_catalog.normalizer.storage.product.product_value` called with `storage`format
- `pim_catalog.normalizer.standard.product.product_value` called with `storage`format
- storage datetime normalizer *not found* in `pim_standard_format_serializer` injected in `pim_catalog.normalizer.standard.product.product_value`

I had to declare another service.
- `pim_catalog.normalizer.storage.product.product_value` called with `storage`format
- `pim_catalog.normalizer.storage_with_standard_format.product.product_value` called with `storage`format
- storage datetime normalizer *found* in `pim_storage_serializer` injected in `pim_catalog.normalizer.storage_with_standard_format.product.product_value`

PR #8576, opened 2018-07-30
[SLA] PIM-7538: Reset datagrid state correctly

tamarasaurus:

This PR fixes an issue for the reset collection action in the datagrid. Currently, we store the initial state of the filters, and when we click reset, we revert to those filters. 

If we set the family to e.g. 'Accessories' and then change to another page, if we come back to the grid and click reset, it doesn't remove 'Accessories' because the initial state now includes this filter. This behaviour is a bug and it also happens when the user has a view selected.  

To fix this issue - first for the default view I updated the reset-collection-action to correctly reset the state using `initialState` instead of `_initState`. For a selected view, we now store the initial view filters so that when the user clicks reset we can revert to them. 

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/features/product/filtering/filter_products.feature
+++ b/features/product/filtering/filter_products.feature
@@ -69,6 +69,29 @@ Feature: Filter products
     When I reset the grid
     Then the grid should contain 6 elements
 
+  Scenario: Successfully reset the filters after switching pages
+    Given I am on the products page
+    Then I filter by "enabled" with operator "" and value "Enabled"
+    And the grid should contain 5 elements
+    Then I am on the families page
+    Then I am on the products page
+    When I reset the grid
+    Then the grid should contain 6 elements
+
+  Scenario: Successfully reset the filters for a selected view
+    Given I am on the products page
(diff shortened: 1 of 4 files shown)


features/product/filtering/filter_products.feature
```diff
@@ -69,6 +69,29 @@ Feature: Filter products
...
+
+  Scenario: Successfully reset the filters for a selected view
+    Given I am on the products page
+    And I filter by "family" with operator "in list" and value "furniture"
+    And I create the view:
+      | new-view-label | Furniture only |
+    Then the grid should contain 2 elements
+    When I refresh the grid
+    Then I filter by "enabled" with operator "" and value "Disabled"
+    Then the grid should contain 0 elements
```
damien-carcel: And

features/product/filtering/filter_products.feature
```diff
@@ -69,6 +69,29 @@ Feature: Filter products
...
+    Then I am on the families page
+    Then I am on the products page
```
damien-carcel: And

PR #8587, opened 2018-07-31
PIM-7551: Fix issue on product model import with custom column headers

mmetayer:
This PR adds new `ProductModelAssociationReader`s (for both CSV and XLSX) in order to take the job instance parameters  into account (e.g `familyVariantColumn` and `categoriesColumn`) during the last step of the product model import jobs (import_product_model_associations).

`OK`: Done / Validated
`-`: Not needed


src/Pim/Bundle/ConnectorBundle/Resources/config/readers.yml
```diff
@@ -12,11 +12,13 @@ parameters:
     pim_connector.reader.file.csv_association.class: Pim\Component\Connector\Reader\File\Csv\ProductAssociationReader
     pim_connector.reader.file.csv_product.class: Pim\Component\Connector\Reader\File\Csv\ProductReader
     pim_connector.reader.file.csv_product_model.class: Pim\Component\Connector\Reader\File\Csv\ProductModelReader
+    pim_connector.reader.file.csv_product_model_association.class: Pim\Component\Connector\Reader\File\Csv\ProductModelAssociationReader
```
damien-carcel: Do not create class parameters. Use the class directly in the service.

src/Pim/Component/Connector/Reader/File/Csv/ProductModelAssociationReader.php
```diff
@@ -0,0 +1,36 @@
...
+{
+    /**
+     * {@inheritdoc}
+     */
+    protected function getArrayConverterOptions()
+    {
+        $jobParameters = $this->stepExecution->getJobParameters();
+
+        return [
+            // for the array converters
```
damien-carcel: I would prefer an explanation in the PHPDoc rather that comment in code :slightly_smiling_face: .

PR #8592, opened 2018-08-01
PIM-7433: Rename 'acceptance' env into 'test_fake'

SamirBoulil:

- Renames the symfony environment 'acceptance' into 'test_fake'.
- Updates the behat.yml.dist accordingly

Motivation for this change:

**Situational**:
In the enriched entities feature, we have two different kinds of integration tests:
- Integration tests of the DB: which uses the Sql implementation of repositories and query functions
- Integration tests of the Controller: which uses the InMemory implementation of repositories and query functions.

Both of thoses types of tests currently use the 'test' environment. 

Up until today, what did to use the inMemory implems instead of the Sql implems (wired in the 'test' environment) is the following: 
1. On phpUnit::setUp(), create & boot a 'test' kernel.
2. Take this list of services and inMemory* classes and set it on the container.

**Pitfall**:

Nothing replaces a good DI and eventually we fell into the case where we needed to override a service which was tagged (in our case a custom constraint validator). After the kernel was booted, there was no way we could override the service with it's inMemory implementation sibling.

| Changelog updated                 |  should I ?

`OK`: Done / Validated
`-`: Not needed


jjanvier:
OK, we discussed with Romain.
The problem is the naming. 
Because for instance, he would like to create a fake service a an HTTP client. So it's not really a "in memory" data like what we have for our repositories.

Maybe we could propose a new name?

jjanvier:
I propose to change `in_memory` by `test_fake`
and to change `test` by `test_real`

SamirBoulil:
Yep I get it, this way it's more generic. Indeed using InMemory implementation is Faking things but faking things can be so much more than inMemory fakes.

So 'test_fake' seems good to me. we could also use a less technical term refering as to some unclosed world or a simulation environment. What do you think about "the_matrix" ?
xD

PR #8604, opened 2018-08-02
PIM-7559: Fix versionning query too slow on the PEF

skeleton:
With too many versions in `pim_versioning_version` table, the query was too slow and broke the PEF (`getNewestLogEntryForRessources()` is only called in it).

FYI, we could only reproduce the bug with a custo on category (the client overidden the table)

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Pim/Bundle/VersioningBundle/Repository/VersionRepositoryInterface.php
+++ b/src/Pim/Bundle/VersioningBundle/Repository/VersionRepositoryInterface.php
@@ -51,7 +51,7 @@ interface VersionRepositoryInterface
      *
      * @param array $resourceNames
      *
-     * @return Version|null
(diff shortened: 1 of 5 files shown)


src/Pim/Bundle/VersioningBundle/Repository/VersionRepositoryInterface.php
```diff
@@ -51,7 +51,7 @@ public function getNewestLogEntry($resourceName, $resourceId, $pending = false);
      *
      * @param array $resourceNames
      *
-     * @return Version|null
+     * @return array|null
```
mmetayer: Maybe a `BC breaks` entry in the CHANGELOG?

PR #8606, opened 2018-08-02
PIM-7537: Use pure JS for reset password trigger form change

tamarasaurus:

Since we don't use jQuery on the old login pages anymore when the reset password form was validated it threw an error. This PR updates the old twig to use pure js to trigger the form change instead. 

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Pim/Bundle/UIBundle/Resources/views/Form/pim-fields.html.twig
+++ b/src/Pim/Bundle/UIBundle/Resources/views/Form/pim-fields.html.twig
@@ -169,7 +169,7 @@
 {% block form_errors %}{% spaceless %}
     {% if errors|length > 0 %}
         <script type="text/javascript">
-            var el = document.getElementById($('{{ _self.getRootId(form) }}');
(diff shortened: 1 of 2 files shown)


src/Pim/Bundle/UIBundle/Resources/views/Form/pim-fields.html.twig
```diff
@@ -169,9 +169,12 @@
 {% block form_errors %}{% spaceless %}
     {% if errors|length > 0 %}
         <script type="text/javascript">
-            $(function() {
-                $('#{{ _self.getRootId(form) }}').trigger('change');
-            });
+            var el = document.getElementById($('{{ _self.getRootId(form) }}');
```
phaseinducer: Are you sure about the `$(...)` ?

grena: Same question

tamarasaurus: oops, fixed!

PR #8615, opened 2018-08-06
PIM-7543: Forbid usage of 'label' code for attributes to prevent UI bugs

pierallard:
When you set 'label' as code, it raises some bugs on the datagrid because the 'label' attribute is generated automatically.
You can not, for example, use the filters on the "label" field.

### DIFF
--- a/CHANGELOG-2.3.md
+++ b/CHANGELOG-2.3.md
@@ -4,6 +4,7 @@
 
 - PIM-7537: Fix console error in password reset form
 - PIM-7552: Fix SKU filter display bug on group grid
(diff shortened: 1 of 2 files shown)


grena:
Any test?

pierallard:
@grena Already tested for some of the other regexp words... I think it's useless.

grena:
What happens if someone already has a "label" attribute?

pierallard:
@grena The 'search' on attributes index page will not work. I checked, the product grid works, and attribute groups too.

oliverde8:
Just want to point out that this is a BC break for us. 

As I understand, It means we can't reinstall a PIM with our current fixtures.

PR #8626, opened 2018-08-07
 PIM-7569: Fix limit of 20 results for view selector

pierallard:
When the user create more than 20 views, he can not access to the "load more" select2 feature.
Why ? Because there was a test saying "if there is 20 results from the back, please show the "load more"". But there is a function it automatically adds the default view to this list... So the results are 21, and it always fail.
Now, it did the test of 20 === 20 before adding the default view.
I use this PR to better format the file, and to remove a useless route (not used in EE too)
I can not add a test because we don't have any methods to scroll down into a select2. If you want me to add this one, I can try to make something.

### DIFF
--- a/src/Pim/Bundle/EnrichBundle/Resources/public/js/grid/view-selector.js
+++ b/src/Pim/Bundle/EnrichBundle/Resources/public/js/grid/view-selector.js
@@ -111,8 +111,8 @@ define(
                         displaySwitcher: (this.config.viewTypes.length > 1),
                         viewTypes: this.config.viewTypes
                     }));
-
                     this.initializeSelectWidget();
+
                     this.renderExtensions();
                 }.bind(this));
             },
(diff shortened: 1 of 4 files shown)


src/Pim/Bundle/EnrichBundle/Resources/public/js/grid/view-selector.js
```diff
@@ -221,6 +157,77 @@ define(
...
+                    if (this.currentLoadingPage === page && this.currentLoadingTerm === options.term) {
+                        return;
+                    }
+
+                    this.currentLoadingPage = page;
+                    this.currentLoadingTerm = options.term;
+
+                    FetcherRegistry.getFetcher(fetcher).search(searchParameters).then(function (views) {
+                        let choices = this.toSelect2Format(views);
+                        const more = choices.length === this.getResultsPerPage();
```
pierallard: The real fix is here :) Before, it was done in the options.callback a few lines after.

PR #8713, opened 2018-08-20
TIP-880: move dashboard

LaurentPetard:


jjanvier:
I would have put `display_completeness_widget.feature` in the enrichment.

PR #8730, opened 2018-08-20
PIM-7414 : Use catalog locale to normalize the image when we save the product

christophe-chausseray:

We got an SLA because the UI locale was used instead of the Catalog locale to normalize the image from a localizable asset when we save the product.

`OK`: Done / Validated
`-`: Not needed


src/Pim/Bundle/EnrichBundle/Normalizer/ProductNormalizer.php
```diff
@@ -145,6 +150,7 @@ public function __construct(
         CollectionFilterInterface $collectionFilter,
         NormalizerInterface $completenessCollectionNormalizer,
         UserContext $userContext,
+        CatalogContext $catalogContext,
```
momoss: Put parameter to last with null default value to avoid BC break

PR #8747, opened 2018-08-22
PIM-7614 Improve performance and reduces SQL queries on PEF

BitOne:
In order to improve PEF performances, this PR reduces the number of queries and loaded rows from DB.
On the main route (/enrich/product/rest/{product_id}),this PR allows:
 - to reduce SQL queries count from 474 to 275
 - to reduce PHP function calls from 260k to 210k

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Pim/Bundle/EnrichBundle/Normalizer/VersionNormalizer.php
+++ b/src/Pim/Bundle/EnrichBundle/Normalizer/VersionNormalizer.php
@@ -6,6 +6,7 @@ use Akeneo\Component\Localization\Presenter\PresenterInterface;
 use Akeneo\Component\Versioning\Model\Version;
 use Oro\Bundle\UserBundle\Entity\UserManager;
 use Pim\Component\Catalog\Localization\Presenter\PresenterRegistryInterface;
+use Pim\Component\Catalog\Repository\AttributeRepositoryInterface;
 use Symfony\Component\Serializer\Normalizer\NormalizerInterface;
 use Symfony\Component\Translation\TranslatorInterface;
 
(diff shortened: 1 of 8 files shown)


src/Pim/Bundle/EnrichBundle/Normalizer/VersionNormalizer.php
```diff
@@ -114,21 +123,43 @@ protected function normalizeAuthor($author)
...
-                $attributeName = $matches['attribute'];
-            }
+        $attributeCodes = [];
+        foreach (array_keys($changeset) as $valueHeader) {
+            $attributeCode = $this->extractAttributeCode($valueHeader);
+
+            $attributeCodes[$attributeCode] = true;
+        }
+
+        $attributeTypes = $this->attributeRepository->getAttributeTypeByCodes(array_keys($attributeCodes));
```
mmetayer: You should test if `$this->attributeRepository` is not null before doing that

BitOne: Fixed, cannot be null ;)

src/Pim/Bundle/EnrichBundle/Normalizer/VersionNormalizer.php
```diff
@@ -36,22 +37,30 @@ class VersionNormalizer implements NormalizerInterface
...
+     * @param PresenterRegistryInterface   $presenterRegistry
+     * @param AttributeRepositoryInterface $attributeRepository
      */
     public function __construct(
         UserManager $userManager,
         TranslatorInterface $translator,
         PresenterInterface $datetimePresenter,
-        PresenterRegistryInterface $presenterRegistry
+        PresenterRegistryInterface $presenterRegistry,
+        AttributeRepositoryInterface $attributeRepository
```
pierallard: please add = null, then use the previous behavior if this parameter is null
Else, it's a BC break. Try to avoid them.

BitOne: Without AttributeRepository, the code doesn't work anymore. So putting a "= null" would be a little bit hypocrite 

If you have some practice that could be applied to avoid that, let me know. Still struggling on that one. Thanks.

fitn: To avoid BC break, you can keep the old code and execute it only if attribute repository is injected (so not null).

I agree it's dirty and we don't know if it is necessary or not (and I feel it's not). It raises again that we should define what is extensible or not and how to plug to Akeneo for some cases :P

src/Pim/Bundle/EnrichBundle/Normalizer/VersionNormalizer.php
```diff
@@ -114,21 +123,64 @@ protected function normalizeAuthor($author)
...
+                    $attributeName = $matches['attribute'];
+                }
+                $presenter = $this->presenterRegistry->getPresenterByAttributeCode($attributeName);
+                if (null !== $presenter) {
+                    foreach ($changes as $key => $value) {
+                        $changeset[$attribute][$key] = $presenter->present($value, $context);
+                    }
+                }
             }
+            return $changeset;
```
pierallard: add line before return; remove line after.

src/Pim/Component/Catalog/Localization/Presenter/PresenterRegistryInterface.php
```diff
@@ -32,6 +32,15 @@ public function register(PresenterInterface $presenter, $type);
...
     public function getPresenterByAttributeCode($code);
 
+    /**
+     * Get the first presenter supporting an attribute type
+     *
+     * @param string $attributeType
+     *
+     * @return PresenterInterface|null
+     */
+    public function getPresenterByAttributeType($attributeType);
```
pierallard: Here it's a BC break. You have to add it to the CHANGELOG, for the developers who developed their own PresenterRegistryInterface.

PR #8750, opened 2018-08-23
Use a dedicated system user token instead of UsernamePasswordToken for commands

phaseinducer:

**Why ?**
To run system commands (especially installer commands) and still be able to use permissions, we create a system user on-the-fly and set it in a `UsernamePasswordToken`.
The problem is that the `UsernamePasswordToken` is only supported by the `DaoAuthenticationProvider`. This provider is loaded only when a `login_form` is declared in a security firewall, so we coupled two things completely unrelated. And when we want to use a different firewall configuration, the installer is broken.

**Solution**
Use a dedicated token for the system user, and since it's authenticated by default there is no need for an authentication provider (nor a firewall listener).

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Pim/Bundle/UserBundle/EventListener/CreateSystemUserForCommandsListener.php
+++ b/src/Pim/Bundle/UserBundle/EventListener/CreateSystemUserForCommandsListener.php
@@ -1,24 +1,26 @@
 <?php
 
-namespace Pim\Bundle\UserBundle\EventSubscriber;
+declare(strict_types=1);
+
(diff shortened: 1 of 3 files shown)


src/Pim/Bundle/UserBundle/EventListener/CreateSystemUserForCommandsListener.php
```diff
@@ -1,24 +1,24 @@
 <?php
```
damien-carcel: `declare(strict_types=1);`

PR #8755, opened 2018-08-24
Increase PHP memory limit for composer during docker build

phaseinducer:


.ci/Dockerfile
```diff
@@ -1,6 +1,7 @@
 FROM php:7.1.16-fpm
 
 ENV COMPOSER_CACHE_DIR=/tmp/composer/cache
+ENV COMPOSER_MEMORY_LIMIT=-1
```
BitOne: Please use a limit, even high, but no -1

PR #8756, opened 2018-08-24
Import PHP class instead of using their FQCN

jjanvier:


jjanvier:
One Behat failing, but it passes locally. `tests/legacy/features/pim/structure/product-group/remove_products.feature:20`

merging

PR #8777, opened 2018-08-29
APAI-217 - Cleaning style

pierallard:
Removes useless JS + template
Extract DescriptionHeader


src/Akeneo/Platform/Bundle/UIBundle/Resources/public/less/components/CatalogVolume.less
```diff
@@ -1,221 +1,197 @@
 @import '../base/variables';
 
 .AknCatalogVolume {
```
damien-carcel: shouldn't this `less` file be renamed too? And the main class, in something like `AknDescriptionHeader`?

PR #8780, opened 2018-08-30
$this->catalogContext may be null.

yunosh:

b8dcc498c7ef9c242791ebc341144ac138543aff introduced an (optional) $catalogContext parameter, but it's used unconditionally within normalize(). This is a real problem, because when having a non-localizable, non-scopable image in a product, opening that product now crashes with:
"Call to a member function getLocaleCode() on null" at /srv/pim/vendor/akeneo/pim-community-dev/src/Pim/Bundle/EnrichBundle/Normalizer/ProductNormalizer.php line 237
Maybe the locally provided $context should also be considered?

`OK`: Done / Validated
`-`: Not needed


yunosh:
To be more exact, b8dcc498c7ef9c242791ebc341144ac138543aff introduced the parameter but df61c6ff2f4bea3216f8c8e22a5b010db5a2142b made it optional technically.

grabelle:
Real problem, yes ...

PR #8799, opened 2018-09-04
PIM-7612 allow scalable frontend architecture for media content delivery

AymericPlanche:

**Problem**

On a scalable frontend architecture some image were not delivered properly to the browser that requested the media.

You may have the choice to share the `web/media/cache` folder across several frontend (with whatever shared storage technology (nfs, glusterfs, ...)) but in our case we want each frontend server to have it's own cache.

**Actual behavior**

The true image file is generated on demand with a specific route that handle the image resizing, the storage on `web/media/cache` folder and the response is a http redirect to the real resource.

The following request: `media/show/xxx.jpg/thumbnail_small` (R1) will redirect to `media/cache/thumbnail_small/xxx.jpg` (R2)

In a scalable architecture with :
- a load balancing mechanism with non sticky strategy
- 3 frontend (F1, F2, F3), 

The first request (R1) could be handled by one frontend (F1), but the redirected request (R2) can be handled by another frontend (F3).
The image is generated in F1 filesystem, so the R2 that target directly the cached resource will respond a 404 because the file is not generated on the F3 server.

**Solution**

I've implemented a fallback strategy on the `media/cache/thumbnail_small/xxx.jpg` route.
If the file does not exists on the filesystem, i force the cache to be generated on the frontend that has been requested.

If the file exists on the filesystem, the vhost will deliver directly the resource as usual.

**For the future**

The best solution is to manage a complete decentralized cached storage like google cloud storage and deliver the content through a CDN but it's a project on its own ;)

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/CHANGELOG-2.3.md
+++ b/CHANGELOG-2.3.md
@@ -4,6 +4,10 @@
 
 - PIM-7628: Fix the initialization of the product datagrid identifier filter.
 
+## Enhancements
+
+- PIM-7612: Add the media/cache/{filter}/{path} route support in order to handle scalable frontend architecture for media content delivering
+
 # 2.3.6 (2018-09-06)
 
 ## Enhancements
--- a/src/Pim/Bundle/EnrichBundle/Controller/FileController.php
+++ b/src/Pim/Bundle/EnrichBundle/Controller/FileController.php
@@ -12,6 +12,8 @@ use Pim\Bundle\EnrichBundle\File\FileTypes;
 use Symfony\Component\HttpFoundation\File\MimeType\MimeTypeGuesser;
 use Symfony\Component\HttpFoundation\RedirectResponse;
 use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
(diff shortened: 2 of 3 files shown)


AymericPlanche:
> @AymericPlanche I'm not a very big fan of the solution because it adds additional computation where it should not. Why is R1 not returning directly a binary response with the media ?

We use the liip imagine bundle that actually return the `RedirectResponse`, we could change the behavior but it's more about an improvement because i can't handle this and the side impacts in this ticket.

> In my opinion, if the solution is to not share the cache then this action / redirection have to be removed completely so every route end on the showAction. If we want to keep the cache mechanism we have to do it right: using a Google Storage bucket as CDN could be the real solution here.

The solution to not share the cache between frontends is one of the cloud team input (with basic shared storage like nfs or so).
As i said in the description we agree on the best solution (GCS + CDN) but it's a project on its own.

We had several options here to allow the scalability of the PIM around this "media" subject (sticky session on the load balancer for this context, or rewrite rules for some routes) but this solution was discussed with @mr-prud and we agreed that it's the best solution we could provide in the time we have.

PR #8803, opened 2018-09-05
Override on JenkinsFile

Doodoune:


.ci/Jenkinsfile
```diff
@@ -1,5 +1,5 @@
 #!groovy
-@Library("k8s-utils@v2.0.0")
+@Library("k8s-utils@override")
```
jmleroux: You can use `@Library("k8s-utils@2.x")`

see here https://github.com/akeneo/jenkins-k8s-utils

PR #8805, opened 2018-09-05
Pim 7637 Completeness not calculated on variant products after bulk action edition on attributes

AymericPlanche:

**Problem**
When edit an attribute that is at product model level on a bulk action, the completeness of the descendants is not computed.

There is a subscriber that handle the completeness computation asynchronously (add a job in the queue) but if its not a "unitary" saved entity its skip the computation. As we are in a bulk action, it's not unitary. 

This check is certainly due to performance needs and i don't think that we have to remove it.

**Solution**

On the bulk action product and product model writer step, if there is product models, i add a job in the queue with all the product models to compute.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Pim/Bundle/EnrichBundle/Connector/Writer/MassEdit/ProductAndProductModelWriter.php
+++ b/src/Pim/Bundle/EnrichBundle/Connector/Writer/MassEdit/ProductAndProductModelWriter.php
@@ -2,16 +2,19 @@
 
 namespace Pim\Bundle\EnrichBundle\Connector\Writer\MassEdit;
 
+use Akeneo\Bundle\BatchBundle\Launcher\JobLauncherInterface;
 use Akeneo\Component\Batch\Item\InitializableInterface;
 use Akeneo\Component\Batch\Item\ItemWriterInterface;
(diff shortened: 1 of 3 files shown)


src/Pim/Bundle/EnrichBundle/Connector/Writer/MassEdit/ProductAndProductModelWriter.php
```diff
@@ -37,24 +40,48 @@ class ProductAndProductModelWriter implements ItemWriterInterface, StepExecution
...
+     * @param IdentifiableObjectRepositoryInterface $jobInstanceRepository
+     * @param string $jobName
      */
     public function __construct(
         VersionManager $versionManager,
         BulkSaverInterface $productSaver,
         BulkSaverInterface $productModelSaver,
-        CacheClearerInterface $cacheClearer
+        CacheClearerInterface $cacheClearer,
+        TokenStorageInterface $tokenStorage,
```
damien-carcel: Add some `= null`, or it is a BC break. Add also a `@todo merge` for master.

PR #8807, opened 2018-09-05
PIM-7635: fix elasticsearch config override

jmleroux:
**The problem**
ElasticSearch config is not currently overridable. As contrary to what is said in https://docs.akeneo.com/2.3/maintain_pim/common_issues/index.html#the-limit-of-total-fields-is-reached-in-elasticsearch, it's not currently possible to change the total_fields.limit value.

The problem comes from the use of `array_merge_recursive` with associative arrays. Instead of overriding the first array entry with the second, instead it will create an array containing both values.

Per the [doc](http://php.net/manual/fr/function.array-merge-recursive.php)

> If the input arrays have the same string keys, then the values for these keys are merged together into an array, and this is done recursively

**The solution**

By using `array_replace_recursive`, the config keys are replace and the override is then possible.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -1,5 +1,5 @@
 #!groovy
-@Library("k8s-utils@v2.0.0")
+@Library("k8s-utils@2.x")
 
 String[] editions = ["ce"]
 String[] legacyFeatures = ["tests/legacy/features"]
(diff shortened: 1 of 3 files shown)


docteurklein:
I approved cause I like the simplicity of the fix, but maybe now it means we can't "replace" an array entry with another, only modify the existing keys or add ones. I guess it's OK tho.

PR #8808, opened 2018-09-05
PIM-7628: Fix identifier filter initialization

damien-carcel:

The identifier filter "SKU" on the product grid is not initialized properly. Its default operator "in list" should present a multi-input , but it is initialized as a simple string filter. So it is impossible to input several product identifiers.

This PR extracts the initialization into dedicated private methods, so they can be called when switching operator (which already work) and at first rendering too (which didn't work before).

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -1,10 +1,10 @@
 {
     "_readme": [
         "This file locks the dependencies of your project to a known state",
-        "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
+        "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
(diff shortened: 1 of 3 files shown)


composer.lock
```diff
@@ -1,10 +1,10 @@
 {
     "_readme": [
         "This file locks the dependencies of your project to a known state",
-        "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
```
Doodoune: Why do you commit your composer.lock ?

damien-carcel: It is committed in CE since 2.3 to improve CI install times.
I just updated it (as we should do on a regular basis).

PR #8826, opened 2018-09-10
TIP-883 Removing attribute object from the product value

BitOne:

This PR aims to do a first step toward defining a clean Product aggregate, by removing the Attribute object from inside the ProductValue, where it does not belong. With this PR, the ProductValue now has a reference to the Attribute, via the Attribute's code.

This has several benefits:
 - clear Product aggregate, as define in https://github.com/akeneo/pim-community-dev/blob/TIP-883/internal_doc/ARCHITECTURE.md#the-prooduct-aggregate
 - better dependency awareness: to get an Attribute object referenced by a ProductValue, most of the time, an attribute repository would need to be injected, highlighting the dependency on attribute repository of the service
 - better performance by better management of attribute loading

The performance subject needs to be work on once this PR is merged. As for now, to keep things simple, the PR only contains adaptations to the separation of Value and Attribute, and uses an Attribute cached repository where previously the code accessed the attribute through the value.

The installation time is similar to current time, at 4 minutes for master and this branch on my computer. So nothing revolutionary for now on the performance side.

But it opens wider perspectives for optimizations than the current state of the code, as it provides a centralized usage of the AttributeRepository, whereas it was kind of hidden previously.

| Changelog updated                 | Done
| Review and 2 GTM                  | Done

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/CHANGELOG-3.0.md
+++ b/CHANGELOG-3.0.md
@@ -7,14 +7,50 @@
 - GITHUB-8451: Add basic compatibility for PHP 7.2  (Thanks [janmyszkier](https://github.com/janmyszkier)!)
 - PIM-7371: Improve the performance to display the category tree in the product grid
 - PIM-7506: Cache default views and columns on the product grid
-- TIP-879: Uses utf8mb4 as encoding for MySQL instead of the less efficient utf8
+- TIP-879: Uses utf8mb4 as encoding for MySQL instead of the less complete utf8
 - Centralizes technical requirements checks to reuse them on standard edition
+- TIP-883: In order to have a clean and independant product aggregate, ProductValue only provides attribute code and no more direct attribute access.
 
 ## Enhancements
 
 - TIP-832: Enable regional languages for UI
 
 ## BC breaks
+- `AbstractValue->getAttribute()` has been replaced by `AbstractValue->getAttributeCode()`. You will need to inject the AttributeRepository in your service if you need to access the full Attribute object related to the provided attribute code.
+- `AbstractValue->getLocale()` has been renamed to `AbstractValue->getLocaleCode()` to better represent its behaviour
+- `AbstractValue->getScope()` has been renamed to `AbstractValue->getScopeCode()` to better represent its behaviour
+- MySQL charset for Akeneo is now utf8mb4, instead of the flawed utf8. If you have custom table, you can convert them with `ALTER TABLE my_custom_table CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci`. For Akeneo native tables, the migration scripts apply the conversion.
+- ProductValue objects must now be instantiated through the named constructor must be called to instantiate a new ProductValue. See `Akeneo\Pim\Enrichment\Component\Product\Model\AbstractValue`, methods `value()`, `scopableValue()`, `localizablevalue()` and `scopableAndLocalizableValue()`
+- The service `pim_catalog.repository.cached_attribute`, of type `Akeneo\Tool\Component\StorageUtils\Repository\IdentifiableObjectRepositoryInterface`, has been added to the construtor of the following classes:
+  - `Akeneo\Pim\Enrichment\Bundle\Doctrine\Common\Saver\ProductUniqueDataSynchronizer`
+  - `Akeneo\Pim\Enrichment\Bundle\Form\Subscriber\FilterLocaleSpecificValueSubscriber`
+  - `Akeneo\Pim\Enrichment\Bundle\PdfGeneration\Renderer\ProductPdfRenderer`
+  - `Akeneo\Pim\Enrichment\Component\Product\Completeness\Checker\MediaCompleteChecker`
+  - `Akeneo\Pim\Enrichment\Component\Product\Completeness\Checker\MetricCompleteChecker`
+  - `Akeneo\Pim\Enrichment\Component\Product\Completeness\Checker\PriceCompleteChecker`
+  - `Akeneo\Pim\Enrichment\Component\Product\Completeness\Checker\ValueCompleteChecker`
+  - `Akeneo\Pim\Enrichment\Component\Product\Converter\MetricConverter`
+  - `Akeneo\Pim\Enrichment\Component\Product\Normalizer\Indexing\Value\AbstractProductValueNormalizer`
+  - `Akeneo\Pim\Enrichment\Component\Product\Normalizer\InternalApi\EntityWithFamilyVariantNormalizer`
+  - `Akeneo\Pim\Enrichment\Component\Product\Normalizer\Standard\Product\ProductValueNormalizer`
+  - `Akeneo\Pim\Enrichment\Component\Product\Normalizer\Versioning\Product\ValueNormalizer`
+  - `Akeneo\Pim\Enrichment\Component\Product\Validator\Constraints\Product\UniqueProductEntityValidator`
+  - `Akeneo\Pim\Enrichment\Component\Product\Validator\Mapping\ProductValueMetadataFactory`
+  - `Akeneo\Pim\Enrichment\Component\Product\ValuesFiller\AbstractEntityWithFamilyValuesFiller`
+  - `Akeneo\Tool\Component\Api\Normalizer\Exception\ViolationNormalizer`
+  - `Oro\Bundle\PimDataGridBundle\Normalizer\Product\ReferenceDataCollectionNormalizer`
+  - `Oro\Bundle\PimDataGridBundle\Normalizer\Product\ReferenceDataNormalizer`
+
+- The service `pim_catalog.repository.cached_attribute_option`, of type `Akeneo\Tool\Component\StorageUtils\Repository\IdentifiableObjectRepositoryInterface`, has been added to the construtor of the following classes:
+  - `Akeneo\Pim\Enrichment\Component\Product\Normalizer\Versioning\Product\ValueNormalizer`
+  - `Oro\Bundle\PimDataGridBundle\Normalizer\Product\OptionNormalizer`
+  - `Oro\Bundle\PimDataGridBundle\Normalizer\Product\OptionsNormalizer`
+
+- The service `pim_reference_data.repository_resolver`, of type `Akeneo\Pim\Enrichment\Component\Product\Repository\ReferenceDataRepositoryResolverInterface`, has been added to the constructor of the following classes:
+  - `Oro\Bundle\PimDataGridBundle\Normalizer\Product\ReferenceDataCollectionNormalizer`
+  - `Oro\Bundle\PimDataGridBundle\Normalizer\Product\ReferenceDataNormalizer`
+
+- In `Akeneo\Pim\Enrichment\Component\Product\Factory\ValueCollectionFactory`, the attribute repository parameter is now defined as `Akeneo\Tool\Component\StorageUtils\Repository\IdentifiableObjectRepositoryInterface` instead of ``Akeneo\Tool\Component\StorageUtils\Repository\CachedObjectRepositoryInterface`
 
 
 - Change constructor of `Akeneo\UserManagement\Component\Updater\UserUpdater` to add `Akeneo\Tool\Component\FileStorage\File\FileStorerInterface`
@@ -91,7 +127,6 @@
 - Move `Pim\Component\VersioningBundle\Normalizer\Flat\FileNormalizer` to `Akeneo\Pim\Enrichment\Component\Product\Normalizer\Versioning\Product\FileNormalizer`
 - Move `Pim\Component\VersioningBundle\Normalizer\Flat\MetricNormalizer` to `Akeneo\Pim\Enrichment\Component\Product\Normalizer\Versioning\Product\MetricNormalizer`
 - Move `Pim\Component\VersioningBundle\Normalizer\Flat\PriceNormalizer` to `Akeneo\Pim\Enrichment\Component\Product\Normalizer\Versioning\Product\PriceNormalizer`
-- MySQL charset for Akeneo is now utf8mb4, instead of the flawed utf8. If you have custom table, you can convert them with `ALTER TABLE my_custom_table CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci`. For Akeneo native tables, the migration scripts apply the conversion.
 - Move `Pim\Bundle\LocalizationBundle\Twig\AttributeExtension` to `Akeneo\Platform\Bundle\UIBundle\Twig\AttributeExtension`
 - Move `Pim\Bundle\LocalizationBundle\Twig\LocaleExtension` to `Akeneo\Platform\Bundle\UIBundle\Twig\LocaleExtension`
 - Move `Pim\Bundle\ReferenceDataBundle\DataGrid\Extension\Sorter\ReferenceDataSorter` to `Oro\Bundle\PimDataGridBundle\Extension\Sorter\Produc\ReferenceDataSorter`
--- a/internal_doc/ARCHITECTURE.md
+++ b/internal_doc/ARCHITECTURE.md
@@ -2,7 +2,7 @@
 
 ## You said "Bounded contexts"?
 
-According to Eric Evans in [Domain-Driven Design Reference](https://domainlanguage.com/ddd/reference/), a bounded context is: 
+According to Eric Evans in [Domain-Driven Design Reference](https://domainlanguage.com/ddd/reference/), a bounded context is:
 
 *"A description of a boundary (typically a subsystem, or the work of a particular team) within which a particular model is defined and applicable."*
 
@@ -23,3 +23,133 @@ To go towards this direction, we began to split the code into big set of feature
 - *Tool*: Technical libraries that could be used outside Akeneo.
 - *Platform*: Everything that glues all the other contexts together to make it a consistent application.
 
+## The Product Aggregate
+
+### Overview
+
+According to Martin Fowler, "A DDD aggregate is a cluster of domain objects that can be treated as a single unit."
+(https://martinfowler.com/bliki/DDD_Aggregate.html).
+
+For most objects, the aggregate definition is pretty simple: a User is a single unit that has its own properties, like
+the user name, the login, the password.
+
+But for our main object, i.e. the product, things are a little bit more complex, as the product itself has different parts
+with pretty strong domain meaning. But still, we could define the product aggregate as the product object itself, with
+all the other objects that belong to it.
+
+As an example, a `MetricValue` object, that is part of the Product `ValueCollection`, has a `Metric` object. This `Metric`
+object exclusively belongs to the value, which belongs to product itself. It's not shared and its life cycle is the one of
+the product.
+
+To the contrary, an `OptionValue` provides a reference to an `AttributeOption`. But the `AttributeOption` does not belong to the
+`OptionValue` object: the option can be shared between different products and has its own life cycle.
+
+Same thing for the `Attribute`. Each product value has a reference to an attribute, but the attribute doesn't belong to any
+product value.
+
+See http://www.cqrs.nu/Faq/aggregates for even more details.
+
+### Aggregate Root
+
+The Product itself is the root of the aggregate. "Any references from outside the aggregate should only go to the aggregate root. The root can thus ensure the integrity of the aggregate as a whole.", Martin Fowler.
+
+### Definition
+
+Here is the product aggregate, with the objects that belong to it (scalars are not shown):
+
+#### Objects belonging to the aggregate
+
+```
+Product
+    - associations: Collection
+        - Association
+    - values: ValueCollection
+        - DateValue
+            - data: DateTime
+        - MediaValue
+            - data: FileInfo
+        - MetricValue
+            - data: Metric
+        - OptionValue
+        - OptionsValue
+        - PriceCollectionValue
+            - data: PriceCollection
+                - Price
+        - ScalarValue
+        - ReferenceDataValue
+        - ReferenceDataCollectionValue
+    - updated: DateTime
+    - created: DateTime
+```
+
+#### References to object outside of the aggregate
+
+```
+Product:
+    - categories: references to many Categories
+    - associations: Collection
+        - Association
+            - type: reference to one AssociationType
+            - product: reference to a Product
+            - group: reference to a Group
+    - groups: references to many Groups
+    - values: ValueCollection
+        - *Value:
+            - attribute: reference to an Attribute
+            - locale: reference to a Locale
+            - scope: reference to a channel
+        - OptionValue
+            - data: reference to an AttributeOption
+        - OptionsValue
+            - data: references to many AttributeOption
+        - PriceCollection
+            - Price
+                - currency: reference to a Currency
+        - ReferenceDataValue
+            - data: reference to a ReferenceData
+        - ReferenceDataCollectionValue
+            - data: reference to many ReferenceData
+    - family: reference to a Family
+    - familyVariant: reference to a FamilyVariant
+    - parent: reference to a ProductModel
+```
+
+### Development impacts
+
+Defining the product aggregate gives a us a clearer view of responsibilities and references between objects.
+And this has some technical impacts:
+
+ -  Transaction
+
+When persisting all the parts of the aggregate, this must be done in one transaction. Indeed, if one object
+of the aggregate cannot be persisted, then the whole aggregate must be rejected, as it's not valid anymore.
+
+ - Accessing referenced objects outside of the aggregate
+
+For objects that are not part of the aggregate, we reference them by their identifiers. For example,
+`$myValue->getAttributeCode()`.
+If a service needs to get the object behind the reference, it can use the associated repository.
+
+### What about completenesses?
+
+Completenesses represent statistics of the product data state, according to conditions defined at
+family level.
+
+If we had an ideal storage system, completeness would not need to be pre-computed and would be used
+directly in filters and computed from the persistence system in real time.
+
+But as our technical stack doesn't allow this, we need to compute them when products are saved, so
+they can be used for filtering.
+
+They are basically a projection, so they don't belong to the product aggregate, and will be persisted
+during a different transaction.
+
+### The special case of MediaValue and FileInfo.
+
+Technically speaking the `FileInfo` object can be shared between different products, as each time we upload
+a file, a checksum is computed and if another file already exists with this content, a reference is created between
+the existing file and the new product value.
+
+This behaviour only exists for technical reasons, so it doesn't have an impact on the aggregate itself.
+
+In the future, we may remove this behavior, to simplify the code as well as avoid confusing functional behavior.
--- a/src/Akeneo/Pim/Enrichment/Bundle/Doctrine/Common/Saver/ProductUniqueDataSynchronizer.php
+++ b/src/Akeneo/Pim/Enrichment/Bundle/Doctrine/Common/Saver/ProductUniqueDataSynchronizer.php
@@ -7,6 +7,7 @@ use Akeneo\Pim\Enrichment\Component\Product\Model\ProductInterface;
 use Akeneo\Pim\Enrichment\Component\Product\Model\ProductUniqueDataInterface;
 use Akeneo\Pim\Enrichment\Component\Product\Model\ValueInterface;
 use Akeneo\Pim\Structure\Component\Model\AttributeInterface;
+use Akeneo\Tool\Component\StorageUtils\Repository\IdentifiableObjectRepositoryInterface;
 use Doctrine\Common\Collections\ArrayCollection;
 use Doctrine\Common\Collections\Collection;
 
@@ -26,12 +27,13 @@ class ProductUniqueDataSynchronizer
     /** @var ProductUniqueDataFactory */
     protected $factory;
 
-    /**
-     * @param ProductUniqueDataFactory $factory
-     */
-    public function __construct(ProductUniqueDataFactory $factory)
+    /** @var IdentifiableObjectRepositoryInterface */
+    protected $attributeRepository;
+
+    public function __construct(ProductUniqueDataFactory $factory, IdentifiableObjectRepositoryInterface $attributeRepository)
     {
         $this->factory = $factory;
+        $this->attributeRepository = $attributeRepository;
     }
 
     /**
@@ -49,9 +51,7 @@ class ProductUniqueDataSynchronizer
         // We get the actual unique data from the product that just got updated
         $actualUniqueDataCollection = $this->createUniqueDataFromProduct($product);
         // We also extract the attribute code list from this collection
-        $actualAttributeCodes = $this->getAttributeCodes(
-            $product->getValues()->getUniqueValues()
-        );
+        $actualAttributeCodes = $this->getAttributeCodes($actualUniqueDataCollection);
 
         // We substract from the original collection the new attributes to have the attribute codes to remove
         $attributeCodeToRemoveFromUniqueCollection = array_diff(
@@ -130,7 +130,9 @@ class ProductUniqueDataSynchronizer
 
         // We now iterate over the collection to update the UniqueData of the original collection
         foreach ($uniqueDataCollectionToUpdateValue as $uniqueData) {
-            $uniqueData->setProductValue($product->getValue($uniqueData->getAttribute()->getCode()));
+            $uniqueData->setAttribute($uniqueData->getAttribute());
+            $value = $product->getValue($uniqueData->getAttribute()->getCode());
+            $uniqueData->setRawData($value->__toString());
         }
     }
 
@@ -156,11 +158,15 @@ class ProductUniqueDataSynchronizer
 
     private function createUniqueDataFromProduct(ProductInterface $product)
     {
-        return array_map(
-            function (ValueInterface $value) use ($product) {
-                return $this->factory->create($product, $value);
-            },
-            $product->getValues()->getUniqueValues()
-        );
+        $uniqueData = [];
+
+        foreach ($product->getValues() as $value) {
+            $attribute = $this->attributeRepository->findOneByIdentifier($value->getAttributeCode());
+            if (null !== $attribute && $attribute->isUnique()) {
+                $uniqueData[] = $this->factory->create($product, $attribute, $value->__toString());
+            }
+        }
+
+        return $uniqueData;
     }
 }
--- a/src/Akeneo/Pim/Enrichment/Bundle/Form/Subscriber/FilterLocaleSpecificValueSubscriber.php
+++ b/src/Akeneo/Pim/Enrichment/Bundle/Form/Subscriber/FilterLocaleSpecificValueSubscriber.php
@@ -2,6 +2,7 @@
 
 namespace Akeneo\Pim\Enrichment\Bundle\Form\Subscriber;
 
+use Akeneo\Tool\Component\StorageUtils\Repository\IdentifiableObjectRepositoryInterface;
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
 use Symfony\Component\Form\FormEvent;
 use Symfony\Component\Form\FormEvents;
@@ -15,15 +16,16 @@ use Symfony\Component\Form\FormEvents;
  */
 class FilterLocaleSpecificValueSubscriber implements EventSubscriberInterface
 {
+    /** @var IdentifiableObjectRepositoryInterface */
+    protected $attributeRepository;
+
     /** @var string $currentLocale */
     protected $currentLocale;
 
-    /**
-     * @param string $currentLocale
-     */
-    public function __construct($currentLocale)
+    public function __construct(?string $currentLocale, IdentifiableObjectRepositoryInterface $attributeRepository)
     {
         $this->currentLocale = $currentLocale;
+        $this->attributeRepository = $attributeRepository;
     }
 
     /**
@@ -44,13 +46,15 @@ class FilterLocaleSpecificValueSubscriber implements EventSubscriberInterface
         $data = $event->getData();
         $form = $event->getForm();
 
-        if (null === $data) {
+        if (null === $data || null === $this->currentLocale) {
             return;
         }
 
         foreach ($data as $name => $value) {
-            if ($this->currentLocale && $value->getAttribute()->isLocaleSpecific()) {
-                $availableCodes = $value->getAttribute()->getLocaleSpecificCodes();
+            $attribute = $this->attributeRepository->findOneByIdentifier($value->getAttributeCode());
+
+            if ($attribute->isLocaleSpecific()) {
+                $availableCodes = $attribute->getLocaleSpecificCodes();
                 if (!in_array($this->currentLocale, $availableCodes)) {
                     $form->remove($name);
                 }
--- a/src/Akeneo/Pim/Enrichment/Bundle/PdfGeneration/Renderer/ProductPdfRenderer.php
+++ b/src/Akeneo/Pim/Enrichment/Bundle/PdfGeneration/Renderer/ProductPdfRenderer.php
@@ -6,6 +6,7 @@ use Akeneo\Pim\Enrichment\Bundle\PdfGeneration\Builder\PdfBuilderInterface;
 use Akeneo\Pim\Enrichment\Component\Product\Model\ProductInterface;
 use Akeneo\Pim\Structure\Component\AttributeTypes;
 use Akeneo\Pim\Structure\Component\Model\AttributeInterface;
+use Akeneo\Tool\Component\StorageUtils\Repository\IdentifiableObjectRepositoryInterface;
 use Liip\ImagineBundle\Imagine\Cache\CacheManager;
 use Liip\ImagineBundle\Imagine\Data\DataManager;
 use Liip\ImagineBundle\Imagine\Filter\FilterManager;
@@ -41,6 +42,9 @@ class ProductPdfRenderer implements RendererInterface
     /** @var FilterManager */
     protected $filterManager;
 
+    /** @var AttributeRepository */
+    protected $attributeRepository;
+
     /** @var string */
     protected $template;
 
@@ -50,31 +54,23 @@ class ProductPdfRenderer implements RendererInterface
     /** @var string */
     protected $customFont;
 
-    /**
-     * @param EngineInterface     $templating
-     * @param PdfBuilderInterface $pdfBuilder
-     * @param DataManager         $dataManager
-     * @param CacheManager        $cacheManager
-     * @param FilterManager       $filterManager
-     * @param string              $template
-     * @param string              $uploadDirectory
-     * @param string|null         $customFont
-     */
     public function __construct(
         EngineInterface $templating,
         PdfBuilderInterface $pdfBuilder,
         DataManager $dataManager,
         CacheManager $cacheManager,
         FilterManager $filterManager,
-        $template,
-        $uploadDirectory,
-        $customFont = null
+        IdentifiableObjectRepositoryInterface $attributeRepository,
+        string $template,
+        string $uploadDirectory,
+        ?string $customFont = null
     ) {
         $this->templating = $templating;
         $this->pdfBuilder = $pdfBuilder;
         $this->dataManager = $dataManager;
         $this->cacheManager = $cacheManager;
         $this->filterManager = $filterManager;
+        $this->attributeRepository = $attributeRepository;
         $this->template = $template;
         $this->uploadDirectory = $uploadDirectory;
         $this->customFont = $customFont;
@@ -117,16 +113,16 @@ class ProductPdfRenderer implements RendererInterface
     }
 
     /**
-     * Get attributes to display
+     * Get attributes codes to display
      *
      * @param ProductInterface $product
      * @param string           $locale
      *
-     * @return AttributeInterface[]
+     * @return
      */
-    protected function getAttributes(ProductInterface $product, $locale)
+    protected function getAttributeCodes(ProductInterface $product, $locale)
     {
-        return $product->getAttributes();
+        return $product->getUsedAttributeCodes();
     }
 
     /**
@@ -141,13 +137,17 @@ class ProductPdfRenderer implements RendererInterface
     {
         $groups = [];
 
-        foreach ($this->getAttributes($product, $locale) as $attribute) {
-            $groupLabel = $attribute->getGroup()->getLabel();
-            if (!isset($groups[$groupLabel])) {
-                $groups[$groupLabel] = [];
-            }
+        foreach ($this->getAttributeCodes($product, $locale) as $attributeCode) {
+            $attribute = $this->attributeRepository->findOneByIdentifier($attributeCode);
 
-            $groups[$groupLabel][$attribute->getCode()] = $attribute;
+            if (null !== $attribute) {
+                $groupLabel = $attribute->getGroup()->getLabel();
+                if (!isset($groups[$groupLabel])) {
+                    $groups[$groupLabel] = [];
+                }
+
+                $groups[$groupLabel][$attribute->getCode()] = $attribute;
+            }
         }
 
         return $groups;
@@ -166,8 +166,10 @@ class ProductPdfRenderer implements RendererInterface
     {
         $imagePaths = [];
 
-        foreach ($this->getAttributes($product, $locale) as $attribute) {
-            if (AttributeTypes::IMAGE === $attribute->getType()) {
+        foreach ($this->getAttributeCodes($product, $locale) as $attributeCode) {
+            $attribute = $this->attributeRepository->findOneByIdentifier($attributeCode);
+
+            if (null !== $attribute && AttributeTypes::IMAGE === $attribute->getType()) {
                 $mediaValue = $product->getValue(
                     $attribute->getCode(),
                     $attribute->isLocalizable() ? $locale : null,
--- a/src/Akeneo/Pim/Enrichment/Bundle/Resources/config/completeness_checkers.yml
+++ b/src/Akeneo/Pim/Enrichment/Bundle/Resources/config/completeness_checkers.yml
@@ -8,22 +8,30 @@ parameters:
 services:
     pim_catalog.completeness.checker:
         class: '%pim_catalog.completeness.checker.class%'
+        arguments:
+            - '@pim_catalog.repository.cached_attribute'
 
     pim_catalog.completeness.checker.media:
         class: '%pim_catalog.completeness.checker.media.class%'
         public: false
+        arguments:
+            - '@pim_catalog.repository.cached_attribute'
         tags:
             - { name: pim_catalog.completeness.checker.product_value, priority: 100 }
 
     pim_catalog.completeness.checker.metric:
         class: '%pim_catalog.completeness.checker.metric.class%'
         public: false
+        arguments:
+            - '@pim_catalog.repository.cached_attribute'
         tags:
             - { name: pim_catalog.completeness.checker.product_value, priority: 100 }
 
     pim_catalog.completeness.checker.price:
         class: '%pim_catalog.completeness.checker.price.class%'
         public: false
+        arguments:
+            - '@pim_catalog.repository.cached_attribute'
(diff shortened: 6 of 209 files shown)


internal_doc/ARCHITECTURE.md
```diff
@@ -23,3 +23,127 @@ To go towards this direction, we began to split the code into big set of feature
...
+(https://martinfowler.com/bliki/DDD_Aggregate.html).
+
+For most object, the aggregate definition is pretty simple: a User is a single unit that has is own properties, like
+the username, the login, the password.
+
+But for our main object, ie the product, things are a little bit more complex, as the product itsef has different parts
+with pretty strong domain meaning. But still, we could define the product aggregate as the product object itself, with
+all the other objects that belongs to it.
+
+As an example, a `MetricValue` object, that is part of the Product `ValueCollection`, has a `Metric` object. This `Metric`
```
ahocquard: But I'm wondering if `getBaseUnit` and `getBaseData` should be part of this object `Metric`.

I mean, currency value is a reference to a currency, why not for the same behavior for the `metric`?

internal_doc/ARCHITECTURE.md
```diff
@@ -23,3 +23,127 @@ To go towards this direction, we began to split the code into big set of feature
...
+
+But as our technical stack don't allow this, we need to compute them when products are saved, so
+they can be used for filtering.
+
+They are basically a projection, so they don't belong to the product aggregate, and will be persisted
+during a different transaction.
+
+### The special case of MediaValuee and FileInfo.
+
+Technically speaking the `FileInfo` object can be shared between different products, as each time we upload
```
ahocquard: You don't upload a media (image/file) in the same transaction as for the product (at least, in API/flat import).
So, is it really part of the product? Is it normal to have access to the original name in the product value? 

Note: maybe we *should* upload it at the same time (import/API) and we are not doing it correctly for now. 


Currently, our current implementation is more a 1-1 association.

Also, I think an asset can be shared between several products.
It has its own lifecycle (transformation, etc). So, it's an association and not part of the aggregate.

internal_doc/ARCHITECTURE.md
```diff
@@ -23,3 +23,127 @@ To go towards this direction, we began to split the code into big set of feature
...
+    - values: ValueCollection
+        - *Value:
+            - attribute: reference to an Attribute
+            - locale: reference to a Locale
+            - scope: reference to a channel
+        - OptionValue
+            - data: reference to an AttributeOption
+        - OptionsValue
+            - data: references to many AttributeOption
+        - PriceCollection
```
ahocquard: and what about reference data?

internal_doc/ARCHITECTURE.md
```diff
@@ -23,3 +23,127 @@ To go towards this direction, we began to split the code into big set of feature
...
+        - ScalarValue
+    - updated: DateTime
+    - created: DateTime
+```
+
+#### References to object outside of the aggregate
+
+```
+Product:
+    - categories: references to many Categories
```
ahocquard: Two points:

- the main one is:
Is the categories from the parent part of the aggregate?

Personally, I don't think so. 
Inheritance from the product model is just for the projection of a product (export).

Same for the association, etc.

The second one, more technical:
- reference by surrogate key (id, uuid) or by natural key (code)?

CHANGELOG-3.0.md
```diff
@@ -6,13 +6,18 @@
 - GITHUB-8451: Add basic compatibility for PHP 7.2  (Thanks [janmyszkier](https://github.com/janmyszkier)!)
 - PIM-7371: Improve the performance to display the category tree in the product grid
 - PIM-7506: Cache default views and columns on the product grid
+- TIP-883: In order to have a clean and independant product aggregate, ProdiuctValue only provide attribute code and no direct attribute access.
```
ahocquard: `ProductValue`

src/Akeneo/Pim/Enrichment/Bundle/Resources/config/completeness_checkers.yml
```diff
@@ -8,22 +8,30 @@ parameters:
 services:
     pim_catalog.completeness.checker:
         class: '%pim_catalog.completeness.checker.class%'
+        arguments:
```
ahocquard: BC break?

internal_doc/ARCHITECTURE.md
```diff
@@ -23,3 +23,127 @@ To go towards this direction, we began to split the code into big set of feature
 - *Tool*: Technical libraries that could be used outside Akeneo.
 - *Platform*: Everything that glues all the other contexts together to make it a consistent application.
 
+## The Prooduct Aggregate
```
ahocquard: `Product`

internal_doc/ARCHITECTURE.md
```diff
@@ -23,3 +23,127 @@ To go towards this direction, we began to split the code into big set of feature
...
+```
+
+### Development impacts
+
+Defining the product aggregate gives a us a clearer view of responsabilities and references between objects.
+And this has some technical impacts:
+
+ -  Transaction
+
+When peristing all the parts of the aggregate, this must be done in one transaction. Indeed, if one object
```
ahocquard: persisting

internal_doc/ARCHITECTURE.md
```diff
@@ -23,3 +23,127 @@ To go towards this direction, we began to split the code into big set of feature
...
+
+### What about completenesses?
+
+Completenesses represent statistics of the product data state, according to conditions defined at
+family level.
+
+If we had an ideal storage system, completeness would not need to be pre-computed and would be used
+directly in filters and read from the persistence system in real time.
+
+But as our technical stack don't allow this, we need to compute them when products are saved, so
```
ahocquard: doesn't

src/Akeneo/Pim/Enrichment/Bundle/Doctrine/Common/Saver/ProductUniqueDataSynchronizer.php
```diff
@@ -156,11 +158,15 @@ function (ProductUniqueDataInterface $uniqueData) use ($attributeCodes) {
...
-                return $this->factory->create($product, $value);
-            },
-            $product->getValues()->getUniqueValues()
-        );
+        $uniqueData = [];
+
+        foreach ($product->getValues() as $value) {
+            $attribute = $this->attributeRepository->findOneByIdentifier($value->getAttributeCode());
+            if ($attribute->isUnique()) {
+                $uniqueData[$value->getAttributeCode()] = $this->factory->create($product, $attribute, $value->__toString());
```
ahocquard: I didn't check in all part of the code, but take care about casting of array keys "string -> int" when a code is a numeric string such as "123". 
As PHP casts it as integer, we had some unexpected behavior (maybe in the diff here?).

BitOne:
@ahocquard : First of all, please enjoy your holidays, we will have time to discuss when you come back ;)

On the rich model, that's a good question. But about the factory which would be required to build valid objects, with this PR this is not 100% true: the new named constructors forces you to provide at least a proper type for the data object (except from the Scalar whose type is too wide) and locale or/and scope if your want your value to be localizable or/and scopable.

But yes, it's true the Factory adds validation on the existence of the attributes and the options

I would even say this validation could be removed as we have proper superficial validation on those topic. Of course, the deleted attribute and deleted option are still real cases which needs to be addressed ;)

internal_doc/ARCHITECTURE.md
BitOne: A currency is really something that doesn't belong to the product value. Firstly, it can be shared between a lot of different entities, and secondly, its lifecycle is not part of the product lifecycle: if I delete the product, it doesn't make sense to delete the currency.

But in case of a metric, the metric is really a part of the ProductValue, at the same level than a string would be for a ScalarValue: if I remove the ProductValue, the metric is removed as well, it cannot live outside of the product value.
So for me, it's part of the aggregate cluster of objects.

internal_doc/ARCHITECTURE.md
BitOne: That's a very good question. And in fact, in the first version of this PR that I presented to Arnaud, the Fileinfo was not part anymore of the ProductValue, and the ProductValue had only the FileInfo "key", which is its identifier.

But while talking with Arnaud, we realized it was pretty close to a MetricValue: the FileInfo has the lifecycle of the ProductValue. It cannot live outside of it, and is not supposed to be shared.
And in fact, your right, it should be part of the transaction.

If our technical solution for file storage was a BLOB field in the DB, that would be pretty natural to have everything in the DB. But here, we have a "hybrid" storage, with most parts of the data in the DB, and some other parts in the filesystem.

And yes, technically, we have a strange behavior, where the FileInfo can be shared between ProductValue if the content is the same. It has been done for reasons, I suppose, but it's pretty weird to me,

And for the transformation, you are talking about assets (simple image don't have transformations in CE), and indeed they have a different lifecycle. And this PR on EE would put assets as reference, and not has embedded objects.

internal_doc/ARCHITECTURE.md
BitOne: Thanks missing from the doc (but managed by the PR).

internal_doc/ARCHITECTURE.md
BitOne: Agree with the first point.
For the second point, good question, and we definitely need to discuss about that with Arnaud and JJ. I don't have a definite answer right now ;)

src/Akeneo/Pim/Enrichment/Bundle/Resources/config/completeness_checkers.yml
BitOne: Oh, yeah a lot of them ;) Will document them latter on.

src/Akeneo/Pim/Enrichment/Bundle/Doctrine/Common/Saver/ProductUniqueDataSynchronizer.php
BitOne: :+1: You're right. It's even worse if two attributes have codes like "00123" and "0123". They are both going to be replaced by "123" :(

gplanchat: Wouldn't `SplObjectStorage` or php-ds be some alternative ?

BitOne: SplObjectStorage only allows numerical keys so not really a solution. php-ds could be interesting (the docs are down, I cannot check how they manage keys), but it would need to install a new dependency on the server, so definitely overkill ;)

gplanchat: Are you sure about this for `SplObjectStorage` ?

Isn't it more a Map ? (object-keyed, mixed value)

BitOne: I never used it, but from the documentation, you don't provide a key when adding objects, and the `key()` function gives you a position. So it's pretty similar to an array where you don't use key.

BitOne: Fixed: the attribute code are now collected directly in an array as values.

src/Akeneo/Pim/Enrichment/Component/Product/Model/ValueCollectionInterface.php
```diff
@@ -125,16 +121,9 @@ public function getByCodes($attributeCode, $channelCode = null, $localeCode = nu
...
-    public function getAttributes();
-
-    /**
-     * Get the attribute codes used in the collection.
-     *
-     * @return array
+     * @return string[]
      */
-    public function getAttributesKeys();
+    public function getAttributeCodes();
```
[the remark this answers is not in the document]
BitOne: Nop, finally. getKeys() stays, and it's the value keys.
And getAttributeKeys() used to store the attribute codes as key of the array of attributes. So it was flawed ("01234" and "001234 attribute codes would considered the same).

getAttributes() has been removed and getAttributeCodes() will provide similar features (lisiting the attributes used by the values).

src/Akeneo/Pim/Enrichment/Bundle/Doctrine/Common/Saver/ProductUniqueDataSynchronizer.php
gplanchat: Actually, it can be easily used as a map : https://3v4l.org/CDcMY

```php
<?php

class Attribute
{
    public $code;
    
    public function __construct($code)
    {
        $this->code = $code;
    }
}

class ProductValue
{
    public $value;
    
    public function __construct($value)
    {
        $this->value = $value;
    }
}

$map = new SplObjectStorage();

$map[new Attribute('name')] = new ProductValue('Long T-Shirt');
$map[new Attribute('description')] = new ProductValue('Lorem ipsum dolor sit amet');

echo 'Using Foreach with native behavior' . PHP_EOL . PHP_EOL . PHP_EOL;
foreach ($map as $attribute => $value) {
    var_dump($attribute, $value);
}

$coroutine = function(SplObjectStorage $iteratee): \Generator {
    $iteratee->rewind();
    
    while ($iteratee->valid()) {
        yield $iteratee->current() => $iteratee->getInfo();
        $iteratee->next();
    }
};

echo 'Using Foreach with a coroutine' . PHP_EOL . PHP_EOL . PHP_EOL;
foreach ($coroutine($map) as $attribute => $value) {
    var_dump($attribute, $value);
}
```

BitOne: Cool, thanks for the example and the tip!
We would need value object to contains the code, as you cannot use a string as a key with this structure.

BitOne: By the way @ahocquard , I don't have keys collision, even with seemingly same integer. See https://3v4l.org/0TBSD
The main (and only?) problem is the type of the index that is indeed "integer" when it's a real integer. But even the order is correct.

src/Akeneo/Pim/Enrichment/Bundle/Doctrine/Common/Saver/ProductUniqueDataSynchronizer.php
```diff
@@ -156,11 +158,15 @@ function (ProductUniqueDataInterface $uniqueData) use ($attributeCodes) {
...
-            function (ValueInterface $value) use ($product) {
-                return $this->factory->create($product, $value);
-            },
-            $product->getValues()->getUniqueValues()
-        );
+        $uniqueData = [];
+
+        foreach ($product->getValues() as $value) {
+            $attribute = $this->attributeRepository->findOneByIdentifier($value->getAttributeCode());
+            if ($attribute->isUnique()) {
```
BitOne: Check if attribute is null as well

src/Akeneo/Pim/Enrichment/Bundle/PdfGeneration/Renderer/ProductPdfRenderer.php
```diff
@@ -141,7 +137,9 @@ protected function getGroupedAttributes(ProductInterface $product, $locale)
     {
         $groups = [];
 
-        foreach ($this->getAttributes($product, $locale) as $attribute) {
+        foreach ($this->getAttributeCodes($product, $locale) as $attributeCode) {
+            $attribute = $this->attributeRepository->findOneByIdentifier($attributeCode);
```
BitOne: Check if attribute is null

src/Akeneo/Pim/Enrichment/Bundle/PdfGeneration/Renderer/ProductPdfRenderer.php
```diff
@@ -166,7 +164,9 @@ protected function getImagePaths(ProductInterface $product, $locale, $scope)
     {
         $imagePaths = [];
 
-        foreach ($this->getAttributes($product, $locale) as $attribute) {
+        foreach ($this->getAttributeCodes($product, $locale) as $attributeCode) {
+            $attribute = $this->attributeRepository->findOneByIdentifier($attributeCode);
```
BitOne: Check if attribute is null

src/Akeneo/Pim/Enrichment/Component/Product/Completeness/Checker/MediaCompleteChecker.php
```diff
@@ -49,6 +58,8 @@ public function supportsValue(
         ChannelInterface $channel,
         LocaleInterface $locale
     ) {
-        return AttributeTypes::BACKEND_TYPE_MEDIA === $value->getAttribute()->getBackendType();
+        $attribute = $this->attributeRepository->findOneByIdentifier($value->getAttributeCode());
```
BitOne: Check if attribute is null

src/Akeneo/Pim/Enrichment/Component/Product/Completeness/Checker/MetricCompleteChecker.php
```diff
@@ -57,6 +66,8 @@ public function supportsValue(
         ChannelInterface $channel,
         LocaleInterface $locale
     ) {
-        return AttributeTypes::METRIC === $value->getAttribute()->getType();
+        $attribute = $this->attributeRepository->findOneByIdentifier($value->getAttributeCode());
```
BitOne: check if attribute is null

src/Akeneo/Pim/Enrichment/Component/Product/Completeness/Checker/PriceCompleteChecker.php
```diff
@@ -57,6 +66,8 @@ public function supportsValue(
         ChannelInterface $channel,
         LocaleInterface $locale
     ) {
-        return AttributeTypes::PRICE_COLLECTION === $value->getAttribute()->getType();
+        $attribute = $this->attributeRepository->findOneByIdentifier($value->getAttributeCode());
```
BitOne: Check if attribute is null

src/Akeneo/Pim/Enrichment/Component/Product/Completeness/Checker/ValueCompleteChecker.php
```diff
@@ -50,16 +59,18 @@ public function supportsValue(
...
         }
 
-        if (null !== $value->getLocale() && $locale !== $value->getLocale()) {
+        if (null !== $value->getLocaleCode() && $locale->getCode() !== $value->getLocaleCode()) {
             return false;
         }
 
-        if ($value->getAttribute()->isLocaleSpecific() &&
-            !$value->getAttribute()->hasLocaleSpecific($locale)
+        $attribute = $this->attributeRepository->findOneByIdentifier($value->getAttributeCode());
```
BitOne: Check if attribute is null

src/Akeneo/Pim/Enrichment/Component/Product/Converter/MetricConverter.php
```diff
@@ -44,23 +52,24 @@ public function convert(EntityWithValuesInterface $entityWithValues, ChannelInte
...
                 }
 
                 $measureFamily = $data->getFamily();
-                $channelUnit = $channelUnits[$attribute->getCode()];
+                $channelUnit = $channelUnits[$value->getAttributeCode()];
                 $amount = $this->converter
                     ->setFamily($measureFamily)
                     ->convert($data->getUnit(), $channelUnit, $data->getData());
 
+                $attribute = $this->attributeRepository->findOneByIdentifier($value->getAttributeCode());
```
BitOne: Check if attribute is null and if really needed here

src/Akeneo/Pim/Enrichment/Component/Product/Normalizer/Indexing/Value/AbstractProductValueNormalizer.php
```diff
@@ -20,15 +21,26 @@
...
      * {@inheritdoc}
      */
     public function normalize($productValue, $format = null, array $context = [])
     {
-        $locale = (null === $productValue->getLocale()) ? '<all_locales>' : $productValue->getLocale();
-        $channel = (null === $productValue->getScope()) ? '<all_channels>' : $productValue->getScope();
+        $locale = (null === $productValue->getLocaleCode()) ? '<all_locales>' : $productValue->getLocaleCode();
+        $channel = (null === $productValue->getScopeCode()) ? '<all_channels>' : $productValue->getScopeCode();
+
+        $attribute = $this->attributeRepository->findOneByIdentifier($productValue->getAttributeCode());
```
BitOne: Check if attribute is null

src/Akeneo/Pim/Enrichment/Component/Product/Normalizer/Indexing/Value/BooleanNormalizer.php
```diff
@@ -21,8 +21,13 @@ class BooleanNormalizer extends AbstractProductValueNormalizer implements Normal
      */
     public function supportsNormalization($data, $format = null)
     {
-        return $data instanceof ValueInterface &&
-            AttributeTypes::BACKEND_TYPE_BOOLEAN === $data->getAttribute()->getBackendType() && (
+        if (! $data instanceof ValueInterface) {
+            return false;
+        }
+
+        $attribute = $this->attributeRepository->findOneByIdentifier($data->getAttributeCode());
```
BitOne: Check if attribute is null

src/Akeneo/Pim/Enrichment/Component/Product/Normalizer/Indexing/Value/NumberNormalizer.php
```diff
@@ -25,8 +25,13 @@ class NumberNormalizer extends AbstractProductValueNormalizer implements Normali
      */
     public function supportsNormalization($data, $format = null)
     {
-        return $data instanceof ValueInterface &&
-            AttributeTypes::BACKEND_TYPE_DECIMAL === $data->getAttribute()->getBackendType() && (
+        if (! $data instanceof ValueInterface) {
+            return false;
+        }
+
+        $attribute = $this->attributeRepository->findOneByIdentifier($data->getAttributeCode());
```
BitOne: Check if attribute is null

src/Akeneo/Pim/Enrichment/Component/Product/Normalizer/Indexing/Value/TextAreaNormalizer.php
```diff
@@ -23,8 +23,13 @@ class TextAreaNormalizer extends AbstractProductValueNormalizer implements Norma
      */
     public function supportsNormalization($data, $format = null)
     {
-        return $data instanceof ValueInterface &&
-            AttributeTypes::BACKEND_TYPE_TEXTAREA === $data->getAttribute()->getBackendType() && (
+        if (! $data instanceof ValueInterface) {
+            return false;
+        }
+
+        $attribute = $this->attributeRepository->findOneByIdentifier($data->getAttributeCode());
```
BitOne: Check if attribute is null

src/Akeneo/Pim/Enrichment/Component/Product/Normalizer/Indexing/Value/TextNormalizer.php
```diff
@@ -23,8 +23,13 @@ class TextNormalizer extends AbstractProductValueNormalizer implements Normalize
      */
     public function supportsNormalization($data, $format = null)
     {
-        return $data instanceof ValueInterface &&
-            AttributeTypes::BACKEND_TYPE_TEXT === $data->getAttribute()->getBackendType() && (
+        if (! $data instanceof ValueInterface) {
+            return false;
+        }
+
+        $attribute = $this->attributeRepository->findOneByIdentifier($data->getAttributeCode());
```
BitOne: Check if attribute is null

src/Akeneo/Pim/Enrichment/Component/Product/Normalizer/Standard/Product/ProductValueNormalizer.php
```diff
@@ -68,14 +70,16 @@ public function supportsNormalization($data, $format = null): bool
      */
     protected function getCollectionValue(ValueInterface $value, ?string $format = null, array $context = []): array
     {
-        $attributeType = $value->getAttribute()->getType();
-        $context['is_decimals_allowed'] = $value->getAttribute()->isDecimalsAllowed();
+        $attribute = $this->attributeRepository->findOneByIdentifier($value->getAttributeCode());
```
BitOne: Check if attribute is null

src/Akeneo/Pim/Enrichment/Component/Product/Normalizer/Standard/Product/ProductValueNormalizer.php
```diff
@@ -97,12 +101,14 @@ protected function getSimpleValue(ValueInterface $value, ?string $format = null,
             return null;
         }
 
-        $attributeType = $value->getAttribute()->getType();
+        $attribute = $this->attributeRepository->findOneByIdentifier($value->getAttributeCode());
```
BitOne: Check if attribute is null

src/Akeneo/Pim/Enrichment/Component/Product/Validator/Constraints/UniqueValueValidator.php
```diff
@@ -53,11 +57,13 @@ public function __construct(ProductUniqueDataRepositoryInterface $repository, Un
      */
     public function validate($value, Constraint $constraint)
     {
-        if (empty($value)) {
+        if (empty($value) || !$value instanceof ValueInterface) {
             return;
         }
 
-        if ($value instanceof ValueInterface && $value->getAttribute()->isUnique()) {
+        $attribute = $this->attributeRepository->findOneByIdentifier($value->getAttributeCode());
```
BitOne: Check if attribute is null

src/Akeneo/Pim/Enrichment/Component/Product/Validator/Mapping/ProductValueMetadataFactory.php
```diff
@@ -77,7 +79,7 @@ public function hasMetadataFor($value)
     protected function createMetadata(ValueInterface $value)
     {
         $class = ClassUtils::getClass($value);
-        $attribute = $value->getAttribute();
+        $attribute = $this->attributeRepository->findOneByIdentifier($value->getAttributeCode());
```
BitOne: Check if attribute is null

src/Akeneo/Pim/Enrichment/Component/Product/ValuesFiller/AbstractEntityWithFamilyValuesFiller.php
```diff
@@ -91,11 +92,13 @@ protected function getExistingValues(EntityWithValuesInterface $entity): array
         $existingValues = [];
         $values = $entity->getValues();
         foreach ($values as $value) {
+            $attribute = $this->attributeRepository->findOneByIdentifier($value->getAttributeCode());
```
BitOne: Check if attribute is null

src/Akeneo/Pim/Enrichment/Component/Product/ValuesFiller/AbstractEntityWithFamilyValuesFiller.php
```diff
@@ -112,7 +115,7 @@ protected function addMissingPricesToProduct(EntityWithValuesInterface $entity):
         $activeCurrencyCodes = $this->currencyRepository->getActivatedCurrencyCodes();
 
         foreach ($entity->getValues() as $value) {
-            $attribute = $value->getAttribute();
+            $attribute = $this->attributeRepository->findOneByIdentifier($value->getAttributeCode());
```
BitOne: Check if attribute is null

src/Akeneo/Pim/Enrichment/Component/Product/ValuesFiller/ProductValuesFiller.php
```diff
@@ -36,9 +36,10 @@ protected function getExpectedAttributes(EntityWithFamilyInterface $entity): arr
         $attributes = [];
 
         // TODO: remove this when optional attributes are gone
-        $productAttributes = $entity->getAttributes();
-        foreach ($productAttributes as $attribute) {
-            $attributes[$attribute->getCode()] = $attribute;
+        $productAttributeCodes = $entity->getUsedAttributeCodes();
+
+        foreach ($entity->getUsedAttributeCodes() as $productAttributeCode) {
+            $attributes[$productAttributeCode] = $this->attributeRepository->findOneByIdentifier($productAttributeCode);
```
BitOne: Check if attribute is null

src/Akeneo/Pim/Enrichment/Bundle/Doctrine/Common/Saver/ProductUniqueDataSynchronizer.php
```diff
@@ -156,11 +158,15 @@ function (ProductUniqueDataInterface $uniqueData) use ($attributeCodes) {
...
-        return array_map(
-            function (ValueInterface $value) use ($product) {
-                return $this->factory->create($product, $value);
-            },
-            $product->getValues()->getUniqueValues()
-        );
+        $uniqueData = [];
+
+        foreach ($product->getValues() as $value) {
+            $attribute = $this->attributeRepository->findOneByIdentifier($value->getAttributeCode());
```
anaelChardan: isn't it possible to load all needed attributes in one query ?

src/Akeneo/Pim/Enrichment/Bundle/Form/Subscriber/FilterLocaleSpecificValueSubscriber.php
```diff
@@ -44,13 +46,15 @@ public function preSetData(FormEvent $event)
         $data = $event->getData();
         $form = $event->getForm();
 
-        if (null === $data) {
+        if (null === $data || null === $this->currentLocale) {
```
anaelChardan: :+1:

src/Akeneo/Pim/Enrichment/Bundle/Doctrine/Common/Saver/ProductUniqueDataSynchronizer.php
BitOne: Yes, but that will come after this PR ;) This one is already quite huge, and I didn't want to mix refactoring and optimizations. But it will definitely help provide further optimizations once merged.

src/Akeneo/Pim/Enrichment/Bundle/Form/Subscriber/FilterLocaleSpecificValueSubscriber.php
```diff
@@ -44,13 +46,15 @@ public function preSetData(FormEvent $event)
...
             return;
         }
 
         foreach ($data as $name => $value) {
-            if ($this->currentLocale && $value->getAttribute()->isLocaleSpecific()) {
-                $availableCodes = $value->getAttribute()->getLocaleSpecificCodes();
+            $attribute = $this->attributeRepository->findOneByIdentifier($value->getAttributeCode());
```
anaelChardan: Same here?

BitOne: Sam reply ;)

src/Akeneo/Pim/Enrichment/Bundle/Doctrine/Common/Saver/ProductUniqueDataSynchronizer.php
anaelChardan: Ah yeah, maybe it's because you are using the CachedRepo? ;)

src/Akeneo/Pim/Enrichment/Component/Product/Factory/Value/AbstractValueFactory.php
```diff
@@ -0,0 +1,70 @@
...
+    {
+        return $attributeType === $this->supportedAttributeType;
+    }
+
+    /**
+     * Prepare the data and check if everything is correct
+     *
+     * @throws Exception
+     */
+    abstract protected function prepareData(AttributeInterface $attribute, $data, bool $ignoreUnknownData);
```
anaelChardan: prepareValueData maybe ?

internal_doc/ARCHITECTURE.md
```diff
@@ -23,3 +23,132 @@ To go towards this direction, we began to split the code into big set of feature
...
 - *Platform*: Everything that glues all the other contexts together to make it a consistent application.
 
+## The Product Aggregate
+
+### Overview
+
+According to Martin Fowler, "A DDD aggregate is a cluster of domain objects that can be treated as a single unit."
+(https://martinfowler.com/bliki/DDD_Aggregate.html).
+
+For most object, the aggregate definition is pretty simple: a User is a single unit that has is own properties, like
```
jjanvier:
```suggestion
For most objects, the aggregate definition is pretty simple: a User is a single unit that has is own properties, like
```

internal_doc/ARCHITECTURE.md
```diff
@@ -23,3 +23,132 @@ To go towards this direction, we began to split the code into big set of feature
...
+the user name, the login, the password.
+
+But for our main object, i.e. the product, things are a little bit more complex, as the product itself has different parts
+with pretty strong domain meaning. But still, we could define the product aggregate as the product object itself, with
+all the other objects that belongs to it.
```
jjanvier:
```suggestion
all the other objects that belong to it.
```

src/Akeneo/Pim/Enrichment/Component/Product/Factory/Value/AbstractValueFactory.php
BitOne: No it's the really the data only. And we are already inside a ValueFactory. So no real need to repeat that?

src/Akeneo/Pim/Enrichment/Bundle/Doctrine/Common/Saver/ProductUniqueDataSynchronizer.php
BitOne: Don't hesitate the read the PR description by the way :p

internal_doc/ARCHITECTURE.md
```diff
@@ -23,3 +23,132 @@ To go towards this direction, we began to split the code into big set of feature
...
+Completenesses represent statistics of the product data state, according to conditions defined at
+family level.
+
+If we had an ideal storage system, completeness would not need to be pre-computed and would be used
+directly in filters and read from the persistence system in real time.
+
+But as our technical stack doesn't allow this, we need to compute them when products are saved, so
+they can be used for filtering.
+
+They are basically a projection, so they don't belong to the product aggregate, and will be persisted
```
jjanvier: Still sure about that?

I mean, in an ideal world, I completely agree.

But today:
- This has consequences on exports (you save a product, the projection fails, maybe you can export "bad" data).
- Same consequences on the PEF. 
- POs are not aware of the consequences of the eventual consistency such a change would have

To reformulate: do we want the completeness to be eventual consistent? IMO, this is not a decision that should be taken alone by a dev. POs must understand what it really means.

internal_doc/ARCHITECTURE.md
```diff
@@ -23,3 +23,132 @@ To go towards this direction, we began to split the code into big set of feature
...
+during a different transaction.
+
+### The special case of MediaValue and FileInfo.
+
+Technically speaking the `FileInfo` object can be shared between different products, as each time we upload
+a file, a checksum is computed and if another file already exists with this content, a reference is created between
+the existing file and the new product value.
+
+This behaviour only exists for technical reasons, so it doesn't have an impact on the aggregate itself.
```
jjanvier: This behavior was something I deliberately introduced because I thought it would be better this way.
This was an error. This was a premature technical optimization. And of course, it was not based on the opinion of a PO.

internal_doc/ARCHITECTURE.md
BitOne: At least it would be ISO bug with today's eventual consistency ;) The PR doesn't change anything on this subject, but I agree this document will define what the next PRs would do. So we need to make sure about that indeed.

internal_doc/ARCHITECTURE.md
BitOne: Yeap, agree with that. So what should be in the documentation then? That maybe we should remove it?

internal_doc/ARCHITECTURE.md
```diff
@@ -23,3 +23,132 @@ To go towards this direction, we began to split the code into big set of feature
...
+
+Defining the product aggregate gives a us a clearer view of responsibilities and references between objects.
+And this has some technical impacts:
+
+ -  Transaction
+
+When persisting all the parts of the aggregate, this must be done in one transaction. Indeed, if one object
+of the aggregate cannot be persisted, then the whole aggregate must be rejected, as it's not valid anymore.
+
+ -  Accessing objects of the aggregate
```
arnolanglade: You should remove that from the doc because this is a bad practice. This breaks tell don't ask principle. Moreover it breaks a basic principle of OOP that is `encapsulation`. We must not ask data to an object but tell it to do something. What happens if we want to rework the value collection? We will need to rework every piece of code which use the method `getValues`.

src/Akeneo/Pim/Enrichment/Bundle/PdfGeneration/Renderer/ProductPdfRenderer.php
arnolanglade: or our repository should throw exceptions :)

src/Akeneo/Pim/Enrichment/Bundle/PdfGeneration/Renderer/ProductPdfRenderer.php
```diff
@@ -141,13 +137,17 @@ protected function getGroupedAttributes(ProductInterface $product, $locale)
...
-        foreach ($this->getAttributes($product, $locale) as $attribute) {
-            $groupLabel = $attribute->getGroup()->getLabel();
-            if (!isset($groups[$groupLabel])) {
-                $groups[$groupLabel] = [];
-            }
+        foreach ($this->getAttributeCodes($product, $locale) as $attributeCode) {
+            $attribute = $this->attributeRepository->findOneByIdentifier($attributeCode);
 
-            $groups[$groupLabel][$attribute->getCode()] = $attribute;
+            if ($attribute !== null) {
```
arnolanglade: yoda style ?

src/Akeneo/Pim/Enrichment/Component/Product/Factory/Value/AbstractValueFactory.php
```diff
@@ -0,0 +1,70 @@
+<?php
+
+namespace Akeneo\Pim\Enrichment\Component\Product\Factory\Value;
+
+use Akeneo\Pim\Enrichment\Component\Product\Model\ValueInterface;
+use Akeneo\Pim\Structure\Component\Model\AttributeInterface;
+
+/**
+ * Abstract value factory
```
arnolanglade: I am not sure this comment is really useful.

src/Akeneo/Pim/Enrichment/Component/Product/Factory/Value/AbstractValueFactory.php
```diff
@@ -0,0 +1,70 @@
...
+ *
+ * @author    Benoit Jacquemont (benoit@akeneo.com)
+ * @copyright 2018 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php Open Software License (OSL 3.0)
+ */
+abstract class AbstractValueFactory implements ValueFactoryInterface
```
arnolanglade: I don't know if this class should implement `ValueFactoryInterface` or each `Value`

src/Akeneo/Pim/Enrichment/Component/Product/Factory/Value/AbstractValueFactory.php
```diff
@@ -0,0 +1,70 @@
...
+{
+    /** @var string */
+    protected $productValueClass;
+
+    /** @var string */
+    protected $supportedAttributeType;
```
arnolanglade: It should be hardcoded in the child class. It relies on business it should not be configurable.

src/Akeneo/Pim/Enrichment/Component/Product/Factory/Value/AbstractValueFactory.php
```diff
@@ -0,0 +1,70 @@
...
+ * Abstract value factory
+ *
+ * @author    Benoit Jacquemont (benoit@akeneo.com)
+ * @copyright 2018 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php Open Software License (OSL 3.0)
+ */
+abstract class AbstractValueFactory implements ValueFactoryInterface
+{
+    /** @var string */
+    protected $productValueClass;
```
arnolanglade: it should be private, we will only use it in this class. If someone need to override the `create` method that means he does need to extends this class.

src/Akeneo/Pim/Enrichment/Component/Product/Factory/Value/OptionValueFactory.php
```diff
@@ -17,68 +17,28 @@
  * @copyright 2017 Akeneo SAS (http://www.akeneo.com)
  * @license   http://opensource.org/licenses/osl-3.0.php Open Software License (OSL 3.0)
  */
-class OptionValueFactory implements ValueFactoryInterface
+class OptionValueFactory extends AbstractValueFactory
 {
     /** @var IdentifiableObjectRepositoryInterface */
     protected $attrOptionRepository;
```
arnolanglade: Now every `Value` properties should be private.

src/Akeneo/Pim/Enrichment/Component/Product/Model/AbstractValue.php
```diff
@@ -13,109 +13,111 @@
...
+        $this->scopeCode = $scopeCode;
+        $this->localeCode = $localeCode;
+    }
+
+    /**
+     * Named constructor for non scopable, non localizable value
+     */
+    public static function value(string $attributeCode, $data): ValueInterface
+    {
+        return new static($attributeCode, $data, null, null);
```
arnolanglade: In another PR we should introduce value object for locale and channel. Those objects will manage that those codes can be null.

src/Akeneo/Pim/Enrichment/Component/Product/Model/ProductUniqueDataInterface.php
```diff
@@ -18,35 +18,17 @@
...
+
+    public function getProduct(): ProductInterface;
+
+    public function getAttribute(): AttributeInterface;
+
+    public function getRawData(): string;
+
+    public function setAttribute(AttributeInterface $attribute);
+
+    public function setRawData(string $rawData);
```
arnolanglade: : void

src/Akeneo/Pim/Enrichment/Component/Product/Model/ValueInterface.php
```diff
@@ -22,43 +22,39 @@ public function getData();
...
 
     /**
-     * @return string
+     * Checks that the product value is equal to another.
      */
-    public function __toString();
+    public function isEqual(ValueInterface $value): bool;
+
+
+    public function __toString(): string;
```
arnolanglade: Should we put magic method in the interface? this is an open question!

internal_doc/ARCHITECTURE.md
```diff
@@ -23,3 +23,132 @@ To go towards this direction, we began to split the code into big set of feature
 - *Tool*: Technical libraries that could be used outside Akeneo.
 - *Platform*: Everything that glues all the other contexts together to make it a consistent application.
 
+## The Product Aggregate
```
anaelChardan: Maybe have a sentence that a the Product is also the aggregate root of the aggregate and that everything should done through it.

src/Akeneo/Pim/Enrichment/Bundle/PdfGeneration/Renderer/ProductPdfRenderer.php
anaelChardan: useless in PHP

internal_doc/ARCHITECTURE.md
ahocquard: what do you mean by "that everything should done through it"?

src/Akeneo/Pim/Enrichment/Component/Product/Factory/Value/AbstractValueFactory.php
anaelChardan: For me, Data is to much abstract, even if it is inside the ValueFactory it could have been shaped to do another thing, no?

src/Akeneo/Pim/Enrichment/Component/Product/Model/ValueInterface.php
anaelChardan: It's magic only in PHP, I would say yes ;)

src/Akeneo/Pim/Enrichment/Component/Product/Normalizer/Indexing/Value/AbstractProductValueNormalizer.php
```diff
@@ -20,19 +21,34 @@
...
+    }
+
     /**
      * {@inheritdoc}
      */
     public function normalize($productValue, $format = null, array $context = [])
     {
-        $locale = (null === $productValue->getLocale()) ? '<all_locales>' : $productValue->getLocale();
-        $channel = (null === $productValue->getScope()) ? '<all_channels>' : $productValue->getScope();
+        $locale = (null === $productValue->getLocaleCode()) ? '<all_locales>' : $productValue->getLocaleCode();
```
anaelChardan: Can be done via coalesce, easier to read ;)

src/Akeneo/Pim/Enrichment/Bundle/PdfGeneration/Renderer/ProductPdfRenderer.php
BitOne: That would be a very bad practice to do only `if ($attribute)`. First of all, it doesn't have any meaning. Is attribute a boolean, an object, etc....

src/Akeneo/Pim/Enrichment/Component/Product/Factory/Value/AbstractValueFactory.php
BitOne: It isn't indeed ;)

src/Akeneo/Pim/Enrichment/Bundle/PdfGeneration/Renderer/ProductPdfRenderer.php
anaelChardan: No I was speaking about the yoda style comment ;), of course i'm for the null checking ;)

src/Akeneo/Pim/Enrichment/Component/Product/Normalizer/Indexing/Value/AbstractProductValueNormalizer.php
BitOne: I'm trying to avoid changing too much existing code in this PR, which is already quite big.

internal_doc/ARCHITECTURE.md
anaelChardan: "An aggregate will have one of its component objects be the aggregate root. Any references from outside the aggregate should only go to the aggregate root. The root can thus ensure the integrity of the aggregate as a whole."

That to access anything from the ProductAggregate, you have to do it from the root.

src/Akeneo/Pim/Enrichment/Bundle/PdfGeneration/Renderer/ProductPdfRenderer.php
BitOne: You mean, the yoda still is useless in PHP ?

src/Akeneo/Pim/Enrichment/Component/Product/Factory/Value/OptionValueFactory.php
BitOne: That could be done in another PR. I'm trying to reduce the coverage of this one as much as possible ;)

src/Akeneo/Pim/Enrichment/Component/Product/Factory/Value/AbstractValueFactory.php
BitOne: I agree, but it's maybe not in the scope of this PR as this was the way it was implemented before the PR.

src/Akeneo/Pim/Enrichment/Component/Product/Factory/Value/AbstractValueFactory.php
BitOne: We can discuss about it, but the way PHP enforce private (hint: it does it only silently) makes me unconfortable...

src/Akeneo/Pim/Enrichment/Component/Product/Factory/Value/AbstractValueFactory.php
BitOne: "data" is the centerpiece of the value (see getData()). So it's pretty clear in this context.

internal_doc/ARCHITECTURE.md
BitOne: Okay, I will change it to remove the completeness during the product transaction.

internal_doc/ARCHITECTURE.md
BitOne: Add a remark to say we could remove this behavior latter.

src/Akeneo/Pim/Enrichment/Bundle/Doctrine/Common/Saver/ProductUniqueDataSynchronizer.php
anaelChardan: I read it twice :P

src/Akeneo/Pim/Enrichment/Bundle/PdfGeneration/Renderer/ProductPdfRenderer.php
anaelChardan: In my humble opinion yes ;)

internal_doc/ARCHITECTURE.md
anaelChardan: Thanks ;)

src/Akeneo/Pim/Enrichment/Component/Product/Normalizer/Indexing/Value/AbstractProductValueNormalizer.php
anaelChardan: Yes, of course ;)

src/Akeneo/Pim/Enrichment/Component/Product/ValuesFiller/ProductValuesFiller.php
```diff
@@ -36,9 +36,13 @@ protected function getExpectedAttributes(EntityWithFamilyInterface $entity): arr
         $attributes = [];
 
         // TODO: remove this when optional attributes are gone
-        $productAttributes = $entity->getAttributes();
-        foreach ($productAttributes as $attribute) {
-            $attributes[$attribute->getCode()] = $attribute;
+        $productAttributeCodes = $entity->getUsedAttributeCodes();
+
+        foreach ($entity->getUsedAttributeCodes() as $productAttributeCode) {
+            $attribute = $this->attributeRepository->findOneByIdentifier($productAttributeCode);
```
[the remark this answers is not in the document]
BitOne: It's not really about checking the attribute existence, it's more about getting the Attribute object in the array, for future use.

PR #8835, opened 2018-09-12
PIM-7647: Export products various errors

pierallard:
This PR fixes 2 issues:
- First issue is that the Product exports (CSV/XLSX) were not initialized with the right values (the values it can be set by UI)
- Second issue is that the history adds lines even if the user does nothing because the `configuration` is a JSON and the order of its components changes.

So this PR fix it with:
- addition of the missing locales in the completeness. If the completeness is set to "Complete on at least one selected locale", it should set the context-locales to the current locales (fr, en, de).
- update the code it updates the `configuration` field to keep the order of the fields.

### DIFF
--- a/src/Pim/Bundle/EnrichBundle/Resources/public/js/filter/product/completeness.js
+++ b/src/Pim/Bundle/EnrichBundle/Resources/public/js/filter/product/completeness.js
@@ -26,7 +26,7 @@ define([
          * {@inheritdoc}
          */
         configure: function () {
-            this.on('locales:update:after', this.updateState.bind(this));
+            this.listenTo(this.parentForm.getRoot(), 'locales:update:after', this.updateState.bind(this));
             this.listenTo(this.getRoot(), 'pim_enrich:form:entity:pre_update', function (data) {
                 _.defaults(data, {field: this.getCode(), operator: _.first(this.config.operators), value: 100});
             }.bind(this));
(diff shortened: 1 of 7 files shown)


src/Pim/Bundle/EnrichBundle/Resources/public/js/filter/product/completeness.js
```diff
@@ -26,7 +26,7 @@ define([
          * {@inheritdoc}
          */
         configure: function () {
-            this.on('locales:update:after', this.updateState.bind(this));
+            this.listenTo(this.parentForm.getRoot(), 'locales:update:after', this.updateState.bind(this));
```
jmleroux: Why do we need `parentForm`?

jmleroux: Because our JS stack is modular, and there may not be a parent form, or the parent form could be something completely different

pierallard: Really tricky question, I should have put this in the description.
`parentForm` is not the `parent` field of the BaseForm.
`parentForm` is something specific for the Product Export Builder filters, and is always related to the `data` module containing the data filters.
So ; the `completeness` module don't have a "parent" in term of the modules hierarchy of our front infra, and `data` have a full hierarchy.
As the `locales` module sent the event in to `getRoot` of itself, and `data` have exactly the same `getRoot`, calling in this file `this.parentForm.getRoot` will link to the `data` then the global parent of the form, where the event is sent.

src/Pim/Bundle/InstallerBundle/Resources/fixtures/icecat_demo_dev/jobs.yml
```diff
@@ -56,6 +56,10 @@ jobs:
                         field: completeness
                         operator: '>='
                         value: 100
+                        locales:
```
jjanvier: if something like fixes a bug, it means something is missing to fix the bug for custom imports/exports
an explanation in the CHANGELOG maybe?

PR #8849, opened 2018-09-14
Fix integration tests

jjanvier:

### DIFF
--- a/.ci/Dockerfile
+++ b/.ci/Dockerfile
@@ -72,7 +72,7 @@ COPY . .
 COPY .ci/php.ini /usr/local/etc/php/
 COPY .ci/vhost.conf /etc/apache2/sites-available/000-default.conf
 
-RUN composer $COMPOSER_COMMAND --no-ansi --optimize-autoloader --no-interaction --no-progress --prefer-dist --no-suggest \
+RUN composer $COMPOSER_COMMAND --no-ansi --no-interaction --no-progress --prefer-dist --no-suggest \
     && bin/console pim:installer:assets --env=prod \
     && bin/console pim:installer:dump-extensions --env=prod \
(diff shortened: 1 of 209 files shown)


.ci/Dockerfile
```diff
@@ -72,7 +72,7 @@ COPY . .
 COPY .ci/php.ini /usr/local/etc/php/
 COPY .ci/vhost.conf /etc/apache2/sites-available/000-default.conf
 
-RUN composer $COMPOSER_COMMAND --no-ansi --optimize-autoloader --no-interaction --no-progress --prefer-dist --no-suggest \
```
skeleton: why do you remove --optimize-autoloader ?

skeleton: ok, I just saw the comment of Damien

jjanvier: read Slack @skeleton ;)

jjanvier:
From @damien-carcel 

> First, you may have heard we had some issues with the Integration tests on master. When running them locally, Benoit discovered that almost half of them had a wrong namespace after having been moved, resulting in fatal errors.
We were quite afraid, as the CI was green nonetheless… Hopefully, the CI was right and not skipping any tests. The explanation is that we perform `composer install` on the CI with the option `--optimize-autoloader`. As a result, PSR-4/PSR-0 rules are converted into classmap rules. As our PSR-0/PSR-4 definitions for the tests is shitty, the tests were still miraculously working. However, if you install or dump the PSR rules without `--optimize-autoloader`, they do not.
@julien.janvier is currently fixing the tests namespaces, and we will remove the `--optimize-autoloader` from the CI for now. It will be back (as it is a good practice for production, so it is for the CI) once our namespaces are clean, once the split of the tests by bounded context is done.

PR #8854, opened 2018-09-17
Improve storage and indexation normalization performance by using variadic function

ahocquard:

### Problem

The normalization of PV is time consuming for the PHP process and the complexity is not linear.

As stated it in this article [No array_merge In Loops](https://github.com/dseguy/clearPHP/blob/master/rules/no-array_merge-in-loop.md), array_merge is inefficient in a loop.

```
$result = [];
foreach ($source as $list) {
  $result = array_merge($result, $list);
}
```

At first loop, array_merge copy the variable `$result` having a size of 1.
At second loop, array_merge copy the variable `$result` having a size of 2.
At third loop, array_merge copy the variable `$result` having a size of 3.... etc

The complexity is `*O(n^2)* (where n is the number of PV).  
Explanation: 1 + 2 + 3 + ....  is a suite. Formula: O(n*(n+1)/2) => O(n^2/2) => O(n^2)

### Solution

Use variadic function to not copy the `$result` array in each loop, but only one time.
The complexity is *O(n)* as seen in the graph.

And proof:
https://gist.github.com/ahocquard/e24346317511c359213bb79ba42652c2

results_array_merge_10000

### How I saw that?

When I did a very simple POC to test:
- the limit of the DB
- by normalizing product values
- 1 transaction/product (not the real product of the PIM)
- 5000 PV

Before fix:
- 8 seconds for the saving part of the values in the DB
- 193 seconds for the normalization part (x24!)

After fix:
- 8 seconds for the saving part of the values in the DB
- 8 seconds for the normalization part

### What does it change in import?

I tested it quickly.
For now, almost nothing, as doing other tasks is far most costly due to the number of IO we are doing.
But it costs nothing and it will be a problem after optimisation of the PIM.
Normalisation of the product values is very important and have to be optimised the best we can (for reading and writing).

`OK`: Done / Validated
`-`: Not needed

Note: green CI, tests already failing on master

### DIFF
--- a/src/Akeneo/Pim/Enrichment/Component/spec/Product/Normalizer/Indexing/Value/ValueCollectionNormalizerSpec.php
+++ b/src/Akeneo/Pim/Enrichment/Component/spec/Product/Normalizer/Indexing/Value/ValueCollectionNormalizerSpec.php
@@ -2,6 +2,7 @@
 
 namespace spec\Akeneo\Pim\Enrichment\Component\Product\Normalizer\Indexing\Value;
 
+use Akeneo\Pim\Enrichment\Component\Product\Value\ScalarValue;
 use PhpSpec\ObjectBehavior;
 use Akeneo\Pim\Structure\Component\Model\AttributeInterface;
 use Akeneo\Pim\Enrichment\Component\Product\Model\ValueCollection;
@@ -40,6 +41,10 @@ class ValueCollectionNormalizerSpec extends ObjectBehavior
             ->shouldReturn(false);
         $this->supportsNormalization($valueCollection, ProductModelNormalizer::INDEXING_FORMAT_PRODUCT_AND_MODEL_INDEX)
             ->shouldReturn(true);
+    }
+
+    function it_normalizes_an_empty_value_collection() {
+        $this->normalize(new ValueCollection(), ProductNormalizer::INDEXING_FORMAT_PRODUCT_INDEX,[])->shouldReturn([]);
     }
 
     function it_normalizes_product_value_collection(
--- a/src/Akeneo/Pim/Enrichment/Component/spec/Product/Normalizer/Storage/Product/ProductValuesNormalizerSpec.php
+++ b/src/Akeneo/Pim/Enrichment/Component/spec/Product/Normalizer/Storage/Product/ProductValuesNormalizerSpec.php
@@ -17,10 +17,9 @@ use Symfony\Component\Serializer\SerializerInterface;
 
 class ProductValuesNormalizerSpec extends ObjectBehavior
 {
-    function let(SerializerInterface $serializer)
+    function let(NormalizerInterface $normalizer)
     {
-        $serializer->implement(NormalizerInterface::class);
-        $this->setSerializer($serializer);
+        $this->beConstructedWith($normalizer);
     }
 
     function it_is_initializable()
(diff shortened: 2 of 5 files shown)


src/Akeneo/Pim/Enrichment/Component/spec/Product/Normalizer/Storage/Product/ProductValuesNormalizerSpec.php
```diff
@@ -57,8 +55,21 @@ function it_supports_storage_format_and_collection_values()
         $this->supportsNormalization(new \stdClass(), 'other_format')->shouldReturn(false);
     }
 
+    function it_normalizes_an_empty_collection_of_product_values(
+        ValueCollectionInterface $values,
+        \ArrayIterator $valuesIterator
+    ) {
+        $values->getIterator()->willReturn($valuesIterator);
```
arnolanglade: idem here

src/Akeneo/Pim/Enrichment/Component/spec/Product/Normalizer/Storage/Product/ProductValuesNormalizerSpec.php
```diff
@@ -140,4 +151,41 @@ function it_normalizes_collection_of_product_values_in_storage_format(
                 ]
             );
     }
+
+    function it_normalizes_collection_of_product_values_with_numeric_attribute_code(
+        $normalizer,
+        AttributeInterface $textAttribute
+    ) {
+        $textAttribute->getCode()->willReturn('123');
```
arnolanglade: idem

src/Akeneo/Pim/Enrichment/Component/spec/Product/Normalizer/Indexing/Value/ValueCollectionNormalizerSpec.php
```diff
@@ -42,6 +43,17 @@ function it_support_product_value_collection($valueCollection)
             ->shouldReturn(true);
     }
 
+    function it_normalizes_an_empty_value_collection(
+        $valueCollection,
+        \ArrayIterator $valueCollectionIterator
+    ) {
+        $valueCollection->getIterator()->willReturn($valueCollectionIterator);
```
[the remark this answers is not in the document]
ahocquard: Because ValueCollection is an interface and should not be.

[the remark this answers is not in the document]
ahocquard: fixed

src/Akeneo/Pim/Enrichment/Component/spec/Product/Normalizer/Storage/Product/ProductValuesNormalizerSpec.php
ahocquard: fixed

src/Akeneo/Pim/Enrichment/Component/spec/Product/Normalizer/Storage/Product/ProductValuesNormalizerSpec.php
ahocquard: fixed

PR #8869, opened 2018-09-19
 TIP-860: Move form provider to the right BC 

arnolanglade:

`OK`: Done / Validated
`-`: Not needed


src/Akeneo/Platform/Bundle/ImportExportBundle/Resources/config/providers.yml
```diff
@@ -0,0 +1,62 @@
+services:
+    pim_enrich.provider.form.job_instance:
+        class: 'Akeneo\Platform\Bundle\ImportExportBundle\Provider\Form\JobInstanceFormProvider'
+        arguments:
+            -   csv_association_type_export: pim-job-instance-csv-base-export
```
jjanvier: Shouldn't we have a form provider service per context? (If possible)
And it registers only its own imports/exports

arnolanglade: What do you want? copying this class and the service definition in each context or you only spoke about the service definition?

jjanvier: hmmm
I talkled only about service definition

PR #8871, opened 2018-09-20
TIP-873: remove EE datagrid bundle

jjanvier:
> Isn't it dirty (or at least weird) to add some Enterprise Edition related parameters in the datagrid services?

### TL;DR: 

Yes it is. But decoupling those services per context isn't possible because of the grid architecture. Overriding those services (like it was before this PR) is not acceptable : we want to get rid of the Datagrid Bundle override and enforce context decoupling. So here is the last solution. Please note that the _PimDataGridBundle_ is considered as technical debt.

### Long explanation:

Let's take example of the `pim_datagrid.extension.pager.pager_resolver` service. The logic is more or less the same for the other services. This service is overridden in Enterprise Edition for adding the grids `published-product-grid` and `proposal-grid`.

My first wish was to create a pager service per context:
    - pager A : for the classical grid using the ORM pagination (channels, attributes...)
    - pager B : for the Community products' grids using Elasticsearch (products, associations...)
    - pager C : for the Enterprise workflow's grids (published and proposals)
We end up with 3 pager services.

The main problem is that a grid does not define which pager service it can use (it's called an _extension_ in the grids realm). Instead, a pager defines which grid it can be plugged to (it's called _applicable_ in the grids realm). _All pagers are added to all applicable grids._
Ok so it's easy:
    - pager B is applicable to the grids products, associations and a few others
    - pager C is applicable to the grids published and proposals
    - pager A should be applicable to all other grids <= here is the problem

We have 2 possibilities for the pager A:
    - we give it the grids it should not apply for (ie: products, associations, published, proposals...). Which leads us to the same problem that the current PR.
    - we give it all grids it should apply for (channels, attributes...). We (and integrators/customers) don't have a default pager anymore when we create a new grid. 

That's why I preferred to pollute the Community _PimDataGridBundle_ with those parameters.

### DIFF
--- a/bin/check-pullup
+++ b/bin/check-pullup
@@ -195,6 +195,7 @@ $pullUpOk = $check(
         ),
         new ForbiddenDirectory('src/PimEnterprise/Bundle/WorkflowBundle', new Moved('src/Akeneo/Pim/WorkOrganization')),
         new ForbiddenDirectory('src/PimEnterprise/Bundle/InstallerBundle', new Moved('src/Akeneo/Platform')),
+        new ForbiddenDirectory('src/PimEnterprise/Bundle/DataGridBundle', Dispatched::mainEnterpriseBusinessTopics()),
 
         // Enterprise Components
         new ForbiddenDirectory('src/PimEnterprise/Component/Api', Dispatched::mainEnterpriseBusinessTopics()),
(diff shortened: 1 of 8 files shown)


jjanvier:
2 behats failing but they pass locally
merging

PR #8874, opened 2018-09-20
PIM-7652: Fix concurrent edition with the parent of a product or a product model in the UI

ahocquard:

01.	Logged in as Admin, select a product model in the grid 
02. In another browser, logged in as Mary, select a product variant from the product model selected by Admin. 
03. With Admin, edit any product model attributes, SAVE. 
04. With Mary, any product variant attribute, SAVE. 
05. It fails.

The error is due to the fact that we don't filter product values coming from the parent.
As in the import or in the API, we have to filter it.

I reproduced it also with a product model and its parent.

Note: As it's about concurrency, it is not testable (and I think it would be overkill to do that).

Note 2: 
I check in the backend if it's a creation of a product or an update by checking if the product has an id or not.
I do it because when creating a variant product, the frontend does not send the identifier if the identifier is empty. And I need it to filter the values from the parent during an update.
I tried (very hard) to fix the frontend to send the property "identifier" instead of passing it as a product value.
But:
- it does not catch properly errors from the backend because the sku does not match on the the value (`attribute: sku, message:..`.)
- I tried to use the module used for the creation of a simple product: it does not work

I know it's not the best, but I passed a lot of time to find the good solution without understanding how it properly works. So, I stick with this solution for this SLA.

| Added legacy Behats               | No
| Added acceptance tests            | No

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/CHANGELOG-2.3.md
+++ b/CHANGELOG-2.3.md
@@ -8,6 +8,7 @@
   - Please, for this fix, if you implemented `Pim\Component\Catalog\Model\AbstractValue` in specific code be warned that
     the `isEqual(ValueInterface $value)` method does not work due to a bug. Please, implement it in your own code for
     your specific business.
+- PIM-7652: Fix concurrent edition with the parent of a product or a product model in the UI
 
 # 2.3.8 (2018-09-14)
 
--- a/src/Pim/Bundle/EnrichBundle/Controller/Rest/ProductController.php
+++ b/src/Pim/Bundle/EnrichBundle/Controller/Rest/ProductController.php
@@ -16,6 +16,8 @@ use Pim\Component\Catalog\Exception\ObjectNotFoundException;
 use Pim\Component\Catalog\Localization\Localizer\AttributeConverterInterface;
 use Pim\Component\Catalog\Model\AttributeInterface;
 use Pim\Component\Catalog\Model\ProductInterface;
+use Pim\Component\Catalog\Normalizer\Storage\Product\ProductValueNormalizer;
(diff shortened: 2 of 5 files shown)


CHANGELOG-2.3.md
```diff
@@ -1,5 +1,7 @@
 # 2.3.x
 
+- PIM-7652: Fix concurrent edition with the parent of a product or a product model in the UI
```
skeleton: add ## Bug fixes

src/Pim/Bundle/EnrichBundle/Controller/Rest/ProductController.php
```diff
@@ -422,6 +432,11 @@ protected function updateProduct(ProductInterface $product, array $data)
             $data['values'] = [];
         }
 
+        // TODO: @merge remove null
+        if (isset($data['parent']) || $product->isVariant() && null !== $this->productAttributeFilter) {
```
skeleton: I think the condition `$product->isVariant() && null !== $this->productAttributeFilter` is enough

src/Pim/Bundle/EnrichBundle/Resources/config/controllers.yml
```diff
@@ -359,7 +359,7 @@ services:
             - '@pim_enrich.converter.enrich_to_standard.product_value'
             - '@pim_enrich.normalizer.product_violation'
             - '@pim_catalog.builder.product'
-            - '@pim_catalog.association.filter.parent_associations'
+            - '@pim_enrich.filter.product_attribute_filter'
```
skeleton: if I remember well, this service is override in EE

PR #8885, opened 2018-09-24
PIM-7663: fix API endpoint to list products updated since N days

LaurentPetard:

Regression from a previous SLA. There must not be no formatting on date filter for the operators SINCE_LAST_N_DAYS and SINCE_LAST_JOB

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/CHANGELOG-2.3.md
+++ b/CHANGELOG-2.3.md
@@ -2,6 +2,7 @@
 
 ## Bug fixes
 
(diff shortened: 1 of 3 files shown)


LaurentPetard:
@ahocquard Totally agree, but I think there would too much work to do in the context of this SLA. We should entirely rework this controller.

PR #8905, opened 2018-09-28
PIM-7656: Fix link insertion in wysiwyg mass edit field

damien-carcel:
**Description**

A bug prevents from inserting a link when a user mass-edit a text area with the WYSIWYG enabled.
The pop-in window appears, but the "modal-backdrop" mask stays on it.

modal

The pop-in has a higher z-index to stay atop the "modal-backdrop", and this works fine on the PEF.

The problem is that the mass-edit form is in a bootstrap modal, and the pop-in is declared in this modal, but the mask is not (t is not even in the HTML page body, but in a `div` element placed directly at the very end of the `html` tag, after the `body`).

As a result, the mask z-index always win, because even if it is lower than the pop-in z-index, it is higher than the bootstrap modal z-index, so the pop-in one is ignored.

The (very dirty) fix consists in moving the "modal-backdrop" in the bootstrap modal, so its z-index can be applied.

fixed

Please note that the ugly pop-in styling has nothing to do with this PR, and is already like that on released tags of the 2.3.

**Definition Of Done**

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/CHANGELOG-2.3.md
+++ b/CHANGELOG-2.3.md
@@ -7,6 +7,7 @@
 - PIM-7619: Fix search on groups for the variant products.
 - PIM-7671: Fix associations tab cannot display more than 24 associated products/product models or 25 groups.
 - PIM-7668: Fix issues with timezone in various screen, to always use current user timezone.
+- PIM-7656: Fix a bug preventing a link insertion in WYSIWYG mass edit field.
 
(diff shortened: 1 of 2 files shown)


damien-carcel:
2 broken behats. Known unstable ones on nightly, and pass locally. Merging.

PR #8922, opened 2018-10-02
PIM-7681: Harmonize entity exports for locales

docteurklein:
This is a simple implementation of a feature request to normalize the way active locales are present in all exports. 
We want all the active locales to be in the export, whether there is a value for it or not. Some entities already behaved like that, but not all (attributes, family, …). 
In order to do that, I always initialize the translations array with the active locales, and if the object being normalized has a value for it, I'll populate the entry with it. otherwise, it stays `null`.
Note that I still respect the behavior introduced by https://github.com/akeneo/pim-community-dev/pull/8877.

### DIFF
--- a/CHANGELOG-2.3.md
+++ b/CHANGELOG-2.3.md
@@ -1,5 +1,9 @@
 # 2.3.x
 
+## Bug fixes
+
+- PIM-7681: Harmonize entity exports for locales
+
 # 2.3.10 (2018-10-01)
 
 ## Bug fixes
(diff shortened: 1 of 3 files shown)


docteurklein:
@ahocquard sure, I edited the description

ahocquard:
@docteurklein because the API is designed on purpose like it.
We don't return null values, we don't return non existing label, etc.

Moreover, as I said it to PO, I think it's has a big impact to do it.
We are doing it just for a customer that need it for the UI export to  be in the same order with null values, to be enriched with Excel.

It not the use-case of the flat export and the API at all.

PR #8948, opened 2018-10-07
Clearer separation between EE and CE on jenkins builds

anaelChardan:

The goal of this PR is to be able to launch what we want on the CI because today the "Check" of "launch EE behat" was for everything which could led to a longer build time. Yes the form is longer to fill but the build may be shorter.

It also adds the ability to choose the EE branch.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -1,12 +1,19 @@
 #!groovy
 @Library("k8s-utils@2.x")
 
-String[] editions = ["ce"]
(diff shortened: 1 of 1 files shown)


ahocquard:
I don't understand the problem you try to solve.
You launch in EE exactly the same stack as for the CE and it seems logic.
I don't have a use-case in mind where you want to execute different things between CE and EE.

ahocquard:
Then, launch it for CE and EE? What is the problem? This is an edge case.

PR #8959, opened 2018-10-10
PIM-7655: Fix issue with Attribute Groups order

grena:

There were 2 problems:

1) First was **due to a previous fix** (https://github.com/akeneo/pim-community-dev/pull/8813), because the following case could happened due to the `_.sortBy()` method:
```
0_marketing
10_other
1_erp
```
See https://jsfiddle.net/jkc4xowh/2/ for proof.

2) Second was when importing attribute groups, you could have **blank and colliding sort orders**. As we can't make sort order mandatory (_this would be a functional breaking change_), we add a step at the import that ensures the consistent sort orders for attribute groups.


test.php
```diff
@@ -0,0 +1,29 @@
+<?php
```
AymericPlanche: wrongly commited ?

AymericPlanche:
plz update de Jenkins file also like this change https://github.com/akeneo/pim-enterprise-dev/pull/4765/files#diff-8db37e8fcea0f1a8f2f39667e94ebcc4L71 the `gcloud container builds ...` is deprecated and will not work at the end of the month.

thanks

PR #8985, opened 2018-10-15
TIP-860: move general configuration outside of EnrichBundle

jjanvier:
The goal of this PR is to move the configuration of the PIM outside of the EnrichBundle (that we'll remove). So, question is, where to put this configuration?

In a regular Symfony 3 project, we should put it in `app`. This is the best practice. The problem with that solution is that is forces us to change the Pim EE dev, the PIM CE std and the PIM EE std.

That's why we chose Platform. As a reminder, Platform is "Everything that glues all the other contexts together to make it a consistent application." Which is exactly what configuration is about.
I didn't put it in the UIBundle. It's not related to the UI at all. I didn't create a new bundle just for that. It's really useless imo.

I just put it in the `src/Platform/config` folder. That's simple and works well.

(In this PR I also moved some general configuration that was put by "mistake" in `app/config/config.yml`).


app/config/config.yml
```diff
@@ -1,5 +1,5 @@
 imports:
-    - { resource: '@PimEnrichBundle/Resources/config/pim.yml' }
+    - { resource: '../../src/Akeneo/Platform/config/pim.yml' }
```
arnolanglade: :+1:

jjanvier:
same behats broken that on master...
merging

PR #9004, opened 2018-10-19
Fix forgotten bc-break to manage

LaurentPetard:
A "@todo" has been forgotten in the last pull-up  from 2.3

See the original PR : https://github.com/akeneo/pim-community-dev/pull/8918


LaurentPetard:
@jmleroux This parameter is optional by nature. It's not to avoid a BC-break

PR #9005, opened 2018-10-19
Set explicit white background for light inputs

grena:
**Because they are grey by default in Firefox.**

## After


grena:
@pierallard Good question, I just checked and yes, for the disabled ones, the background color is explicitly set (line 99)
```
background: @AknDisabledInputColor;
```

PR #9022, opened 2018-10-24
TIP-860: Add a class parameter for file controller storages

anaelChardan:

`OK`: Done / Validated
`-`: Not needed


anaelChardan:
CI is good, ready to merge @jjanvier

jjanvier:
You introduce this parameter to override it in EE?
I'm not sure this is the right solution.

Can't we instead duplicate the service in EE with the catalog storage you need?

What is this `pim_enrich.controller.file` about? What's its job?

anaelChardan:
This controller is used at many places, with that we open to extension and it avoids to override the whole service

anaelChardan:
The idea come from @arnolanglade

PR #9024, opened 2018-10-24
PIM-7758: Fix product and product model deletion

SamirBoulil:
**Before this PR**:

The deletion of products and product models through the grid is handled
by a Tasklet called `DeleteProductsAndProductModelsTasklet` (https://github.com/akeneo/pim-community-dev/blob/f98e07d7fc7a33bdeb7705c2f927a98395e98c95/src/Pim/Component/Enrich/Job/DeleteProductsAndProductModelsTasklet.php#L128).

The way we fetch the product and product models to delete takes
advantage of the filters the users has set on the grid, passed onto the
tasklet and used along side the PQB.

In this version, we would also add a filter on the 'parent' property
to look for root product models (no parent) and simple products (no
parent as well).

Hence, the deletion would only work for those types of products and
product models.
The deletion of sub-product models and variant product models would
never delete anything.

**Solution**:

As the tasklet also counts the number of products and product models
deleted during its execution, the proposed solution involves deleting
the products/product models of the tree from the bottom up to the root.

Therefore, the deletion is made in 3 steps for the three levels:
1. Delete all the variant products
2. Delete all the sub product models
3. Delete all the root product models and simple products.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Pim/Component/Enrich/Job/DeleteProductsAndProductModelsTasklet.php
+++ b/src/Pim/Component/Enrich/Job/DeleteProductsAndProductModelsTasklet.php
@@ -90,8 +90,58 @@ class DeleteProductsAndProductModelsTasklet implements TaskletInterface
         $this->stepExecution->addSummaryInfo('deleted_products', 0);
         $this->stepExecution->addSummaryInfo('deleted_product_models', 0);
 
-        $products = $this->getProductsCursor();
+        $variantProducts = $this->variantProducts();
+        $this->delete($variantProducts);
 
+        $subProductModels = $this->findSubProductModels();
+        $this->delete($subProductModels);
+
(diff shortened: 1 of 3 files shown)


src/Pim/Component/Enrich/Job/DeleteProductsAndProductModelsTasklet.php
```diff
@@ -90,8 +90,58 @@ public function execute(): void
         $this->stepExecution->addSummaryInfo('deleted_products', 0);
         $this->stepExecution->addSummaryInfo('deleted_product_models', 0);
 
-        $products = $this->getProductsCursor();
+        $variantProducts = $this->variantProducts();
```
phaseinducer: Why not `findVariantProducts()` like other methods ?

PR #9031, opened 2018-10-25
PIM-7757: Reduce a lot the risk of inserting duplicate completenesses.

rouliane:
### The issue

We have a lot of concurrent `INSERT INTO` to save the completeness in the database that generates SQL errors because of duplicated keys. For example: Integrity constraint violation: 1062. Duplicate entry 'X-XX-XXXXXX' for key 'searchunique_idx'

Currently, when we save a product, we remove all the completeness in the database, we compute it again and we save its all in the database.

### The solution

In this PR, to avoid the concurrent `INSERT INTO` that could generate a duplicated key, we change the behavior of the completeness computing: we don't remove the existing product completenesses, but we update the existing ones by updating their ratio and we add or delete completenesses if we need.

The downside of this is that we had to make mutable the Completeness object.

### Performance measurement

| Tested with Icecat demo dev | Current 2.0 | With this PR |
|-----------------------------|-------------|--------------|
| Install                     |    180 s    |     183 s    |
| Product import              |     52 s    |      46 s    |

Volumetry is very small, but it seems the fix has no perceptible impacts on performances.

## Definition Of Done

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/CHANGELOG-2.0.md
+++ b/CHANGELOG-2.0.md
@@ -4,10 +4,6 @@
 
 - PIM-7757: Fix the risk of inserting duplicate completenesses.
 
-## BC Breaks
-
-- Methods `setRatio()`, `setMissingCount()` and `setRequiredCount()` have been added to `Pim\Component\Catalog\Model\CompletenessInterface`.
-
 # 2.0.41 (2018-10-23)
 
 ## Bug fixes
--- a/CHANGELOG-2.0.md
+++ b/CHANGELOG-2.0.md
@@ -1,5 +1,13 @@
 # 2.0.x
 
+## Bug fixes
+
+- PIM-7757: Fix the risk of inserting duplicate completenesses.
+
+## BC Breaks
+
+- Methods `setRatio()`, `setMissingCount()` and `setRequiredCount()` have been added to `Pim\Component\Catalog\Model\CompletenessInterface`.
+
 # 2.0.41 (2018-10-23)
 
 ## Bug fixes
--- a/src/Pim/Component/Catalog/Completeness/CompletenessGenerator.php
+++ b/src/Pim/Component/Catalog/Completeness/CompletenessGenerator.php
@@ -2,7 +2,9 @@
 
 namespace Pim\Component\Catalog\Completeness;
 
+use Doctrine\Common\Collections\Collection;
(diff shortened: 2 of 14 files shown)


CHANGELOG-2.0.md
```diff
@@ -1,5 +1,13 @@
 # 2.0.x
 
+## Bug fixes
+
+- PIM-7757: The risk of inserting duplicate completenesses is far more reduced.
```
willy-ahva: fix

src/Pim/Component/Catalog/Completeness/CompletenessGenerator.php
```diff
@@ -120,22 +123,95 @@ protected function createProductQueryBuilderForMissings(
...
-        $newCompletenesses = $this->completenessCalculator->calculate($product);
-        foreach ($newCompletenesses as $completeness) {
-            $completenessCollection->add($completeness);
+        $newLocalesChannels = [];
+        foreach ($newCompletenesses as $newCompleteness) {
+            $newLocalesChannels[] =
+                $newCompleteness->getLocale()->getId().'/'.$newCompleteness->getChannel()->getId();
+        }
+
+        $completenessesToAdd = array_diff($newLocalesChannels, $currentLocalesChannels);
```
ahocquard: it's a kind of mask actually. I don't like the naming of this variable: it does not contain the completeness objects.

src/Pim/Component/Catalog/Completeness/CompletenessGenerator.php
```diff
@@ -120,22 +123,95 @@ protected function createProductQueryBuilderForMissings(
...
+     * @param CompletenessInterface[] $newCompletenesses
+     * @param string[]                $completenessesToAdd
+     */
+    private function addNewCompletenesses(
+        Collection $completenessCollection,
+        array $newCompletenesses,
+        array $completenessesToAdd
+    ) {
+        foreach ($completenessesToAdd as $completenessToAdd) {
+            list($localeId, $channelId) = explode('/', $completenessToAdd);
```
ahocquard: coding style (just FYI): `[$localeId, $channelId]` works as well

PR #9032, opened 2018-10-25
PIM-7765: Add checks channel/currency in PriceCollectionFactory

SamirBoulil:

**Problem:**
When showing a product, and ultimately, hydrating its price collection values we do not check if the currencies coming from the database are still activated in the channel. hence, we hydrate those prices regardless.

When the user wants to do a modification on the product and saves it, those "old" prices are sent to the backend alongside with the updates, and a violation error occurs because 🎉  you cannot set a price on a product for a currency not activated by the channel :)

**The user cannot update the price anymore through the UI**

**Proposed solution:**
The solution is to:
- filter out the prices that have a currency not activated in the channel of the value at hydration time (for scopable prices attributes).
- filter out the prices that a currency not activated at hydration time (for non-scopable prices attributes)

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Pim/Bundle/CatalogBundle/Doctrine/ORM/Query/FindActivatedCurrencies.php
+++ b/src/Pim/Bundle/CatalogBundle/Doctrine/ORM/Query/FindActivatedCurrencies.php
@@ -58,12 +58,7 @@ class FindActivatedCurrencies implements FindActivatedCurrenciesInterface
             $this->activatedCurrenciesForChannels = $this->fetchActivatedCurrenciesForAllChannels();
         }
 
-        $currencies = [];
-        foreach ($this->activatedCurrenciesForChannels as $channel => $currenciesForChannel) {
-            $currencies = array_merge($currencies, $currenciesForChannel);
-        }
-
-        return array_unique($currencies);
+        return array_unique(array_merge(...array_values($this->activatedCurrenciesForChannels)));
     }
 
     /**
--- a/src/Pim/Bundle/CatalogBundle/Doctrine/ORM/Query/FindActivatedCurrencies.php
+++ b/src/Pim/Bundle/CatalogBundle/Doctrine/ORM/Query/FindActivatedCurrencies.php
@@ -1,0 +1,94 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Pim\Bundle\CatalogBundle\Doctrine\ORM\Query;
+
+use Doctrine\DBAL\DBALException;
+use Doctrine\ORM\EntityManagerInterface;
+use Pim\Component\Catalog\Channel\Query\FindActivatedCurrenciesInterface;
+
+/**
+ * @author    Samir Boulil <samir.boulil@akeneo.com>
+ * @copyright 2018 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+class FindActivatedCurrencies implements FindActivatedCurrenciesInterface
+{
+    /** @var array */
+    private $activatedCurrenciesForChannels = [];
+
+    /** @var EntityManagerInterface */
+    private $entityManager;
+
+    /**
+     * @param EntityManagerInterface $entityManager
+     */
+    public function __construct(EntityManagerInterface $entityManager)
+    {
+        $this->entityManager = $entityManager;
+    }
+
+    /**
+     * Method that returns a list of currencies codes activated for the given channel.
+     *
+     * @param string $channelCode
+     *
+     * @return array
+     *
+     * @throws DBALException
+     */
+    public function forChannel(string $channelCode): array
+    {
+        if (empty($this->activatedCurrenciesForChannels)) {
+            $this->activatedCurrenciesForChannels = $this->fetchActivatedCurrenciesForAllChannels();
+        }
+
+        return $this->activatedCurrenciesForChannels[$channelCode] ?? [];
+    }
+
+    /**
+     * {@inheritdoc}
+     *
+     * @throws DBALException
+     */
+    public function forAllChannels(): array
+    {
+        if (empty($this->activatedCurrenciesForChannels)) {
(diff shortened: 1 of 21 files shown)


src/Pim/Component/Catalog/Channel/Query/FindActivatedCurrenciesInterface.php
```diff
@@ -0,0 +1,26 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Pim\Component\Catalog\Channel\Query;
+
+interface FindActivatedCurrenciesInterface
```
rouliane: Is an interface really necessary ? Could we have another implementation of this query ?

SamirBoulil: I asked myself the same question. If we'd have another adapter of our application: yes (I know it's not the case today). 

In the PIM some query functions have interfaces some other don't. I figured I'll add mine but I can remove it.

src/Pim/Bundle/CatalogBundle/Doctrine/ORM/Query/FindActivatedCurrencies.php
```diff
@@ -0,0 +1,94 @@
...
+use Doctrine\DBAL\DBALException;
+use Doctrine\ORM\EntityManagerInterface;
+use Pim\Component\Catalog\Channel\Query\FindActivatedCurrenciesInterface;
+
+/**
+ * @author    Samir Boulil <samir.boulil@akeneo.com>
+ * @copyright 2018 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+class FindActivatedCurrencies implements FindActivatedCurrenciesInterface
```
anaelChardan: The service is stateful ?

SamirBoulil: Yes, as discussed with @BitOne

src/Pim/Bundle/CatalogBundle/Doctrine/ORM/Query/FindActivatedCurrencies.php
```diff
@@ -0,0 +1,94 @@
...
+     */
+    public function forAllChannels(): array
+    {
+        if (empty($this->activatedCurrenciesForChannels)) {
+            $this->activatedCurrenciesForChannels = $this->fetchActivatedCurrenciesForAllChannels();
+        }
+
+        $currencies = [];
+        foreach ($this->activatedCurrenciesForChannels as $channel => $currenciesForChannel) {
+            $currencies = array_merge($currencies, $currenciesForChannel);
```
anaelChardan: array_merge in a loop is not that good in term of performance, you could use ```array_merge(...array_values($this->activatedCurrenciesForChannel))``` I guess

anaelChardan: See here the explanation: https://github.com/ahocquard/blackfire-overhead-costs

SamirBoulil: I tried, but It didn't work when I tried it

src/Pim/Component/Catalog/Factory/Value/PriceCollectionValueFactory.php
```diff
@@ -28,16 +29,25 @@ class PriceCollectionValueFactory implements ValueFactoryInterface
...
+     * @param string                           $supportedAttributeType
+     * @param FindActivatedCurrenciesInterface $findActivatedCurrenciesForChannel
      */
-    public function __construct(PriceFactory $priceFactory, $productValueClass, $supportedAttributeType)
-    {
+    public function __construct(
+        PriceFactory $priceFactory,
+        $productValueClass,
+        $supportedAttributeType,
+        FindActivatedCurrenciesInterface $findActivatedCurrenciesForChannel = null
```
anaelChardan: Transform it as a nullable please ;)

src/Pim/Component/Catalog/Factory/Value/PriceCollectionValueFactory.php
```diff
@@ -125,11 +138,15 @@ protected function checkData(AttributeInterface $attribute, $data)
      *
      * @return PriceCollection
      */
-    protected function getPrices(AttributeInterface $attribute, array $data)
+    protected function getPrices(AttributeInterface $attribute, array $data, ?string $channelCode)
```
anaelChardan: what is $data and then everywhere what is filteredData? ProductData?

src/Pim/Component/Catalog/Factory/Value/PriceCollectionValueFactory.php
SamirBoulil: ah yes! thanks!

src/Pim/Component/Catalog/Channel/Query/FindActivatedCurrenciesInterface.php
jjanvier: think about decoupling please! domain vs infra is the key
you have no other choice apparently

your query function uses the EntityManager
if you have no interface, then you'll end up with infrastructure concerns in the ` src/Pim/Component/Catalog/Factory/Value/PriceCollectionValueFactory.php`

please read (and remember :p) that carefully https://matthiasnoback.nl/2018/08/when-to-add-an-interface-to-a-class/
> If the class depends on third-party code

that's the case here

src/Pim/Component/Catalog/Factory/Value/PriceCollectionValueFactory.php
SamirBoulil: $data contains the data part of the product value. In this case a list of array having the following form:`['amount' => ..., 'currency' => ... ]`.
The filteredData here corresponds to a filtered version of this list

PR #9035, opened 2018-10-26
Be able to validate an activated locale when the locale is a string

skeleton:
For the onboarder, we need validate a supplier which must contain an activated locale. Problem with the ActivatedLocaleValidator, it accepts only a Locale entity, while on the onboarder, we valide a DTO (which contains the locale code)

Edit:
I did not want to add an other validator like `ActivatedLocaleCodeValidator`. Ideally, in the future, we will only validate DTO, so we will have code to validate, not the entity linked

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Pim/Component/Catalog/Validator/Constraints/ActivatedLocaleValidator.php
+++ b/src/Pim/Component/Catalog/Validator/Constraints/ActivatedLocaleValidator.php
@@ -31,15 +31,23 @@ class ActivatedLocaleValidator extends ConstraintValidator
      */
     public function validate($locale, Constraint $constraint)
     {
-        if (!$locale instanceof LocaleInterface) {
-            return;
-        }
+        if ($locale instanceof LocaleInterface || (is_string($locale) && '' !== $locale)) {
+            if (is_string($locale)) {
+                $locale = $this->localeRepository->findOneByIdentifier($locale);
+                if (null === $locale) { // will be handled by another validator
+                    return;
+                }
+            }
 
-        if (!$locale->isActivated()) {
-            $this->context->buildViolation(
-                $constraint->message,
-                ['%locale%' => $locale]
-            )->addViolation();
+            if ($locale->isActivated()) {
+                return;
+            }
+
+            if ('' !== $constraint->propertyPath) {
+                $this->context->setNode($locale, $this->context->getObject(), $this->context->getMetadata(), $constraint->propertyPath);
+            }
+
+            $this->context->buildViolation($constraint->message, ['%locale%' => $locale])->addViolation();
         }
     }
 }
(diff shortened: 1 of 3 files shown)


src/Pim/Component/Catalog/Validator/Constraints/ActivatedLocaleValidator.php
```diff
@@ -31,15 +31,33 @@ public function __construct(IdentifiableObjectRepositoryInterface $localeReposit
      */
     public function validate($locale, Constraint $constraint)
     {
-        if (!$locale instanceof LocaleInterface) {
+        if (!$locale instanceof LocaleInterface && (null === $locale || '' === $locale)) {
```
AymericPlanche:
```suggestion
        if (!$locale instanceof LocaleInterface || empty($locale)) {
```

src/Pim/Component/Catalog/Validator/Constraints/ActivatedLocaleValidator.php
```diff
@@ -31,15 +31,33 @@ public function __construct(IdentifiableObjectRepositoryInterface $localeReposit
...
         }
 
-        if (!$locale->isActivated()) {
-            $this->context->buildViolation(
-                $constraint->message,
-                ['%locale%' => $locale]
-            )->addViolation();
+        if (is_string($locale)) {
+            $locale = $this->localeRepository->findOneByIdentifier($locale);
+            if (null === $locale) { // will be handle by an other validator
```
phaseinducer: another

src/Pim/Component/Catalog/spec/Validator/Constraints/ActivatedLocaleValidatorSpec.php
```diff
@@ -0,0 +1,118 @@
...
+        $locale = new Locale();
+        $context->buildViolation(
+            'The locale "%locale%" exists but has to be activated.',
+            ['%locale%' => $locale]
+        )->willReturn($violation);
+
+        $violation->addViolation()->shouldBeCalled();
+        $localeRepository->findOneByIdentifier(Argument::any())->shouldNotBeCalled();
+
+        $this->validate($locale, $constraint)->shouldReturn(null);
```
phaseinducer: Why do you need to test the return value ?

src/Pim/Component/Catalog/Validator/Constraints/ActivatedLocaleValidator.php
jjanvier:
```suggestion
            if (null === $locale) { // will be handled by another validator
```

src/Pim/Component/Catalog/Validator/Constraints/ActivatedLocaleValidator.php
AymericPlanche: if $locale = new stdClass(); it will not return and the code will raise an error line 45 $locale->isActivated() "unknown method" isn't it ?

skeleton: in your case, if a locale is a string, it will return. but yeah, there is a but if it's a stdClass, I will fix it

PR #9060, opened 2018-11-02
test cache lru vs doctrine cache

ahocquard:
```
ENRICHMENT=y
STRUCTURE=y
```

Just for fun 

`OK`: Done / Validated
`-`: Not needed


docteurklein:
how does it compare to https://www.doctrine-project.org/projects/doctrine-orm/en/2.6/reference/second-level-cache.html ?

ahocquard:
@docteurklein I didn't test it because it will be at least worst than our cache layer "Doctrine + our cache", as this cache is just doing `array_key_exists` once the entity is loaded, which is pretty fast.

But the whole purpose of the LRU cache is to be auto-garbaged, without any notion of "batch".
It's impossible to do that with Doctrine and the identity map pattern (because it keeps everything in memory). So, it implies to clear the cache quite often, which is very costly.

I did x10 with the LRU cache on the attribute entity.

src/Akeneo/Pim/Enrichment/Bundle/Sql/LRUCache.php
```diff
@@ -0,0 +1,82 @@
...
+     *
+     * @return mixed
+     */
+    public function get(string $key): ?object {
+        if (isset($this->data[$key])) {
+            $this->recordAccess($key);
+            return $this->data[$key];
+        }
+
+        return null;
```
ahocquard: we can't cache null value with this implementation

PR #9075, opened 2018-11-06
PIM-7621: Add average number of attributes per family into System Informations data

Doodoune:

Add new axis : 
Average number of attributes by family: For each family, you count its number of attributes. Then you compute the average for all the families.

It already exist in the Volume Monitoring Screen so already tested with integration tests. 

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Pim/Bundle/AnalyticsBundle/DataCollector/AttributeDataCollector.php
+++ b/src/Pim/Bundle/AnalyticsBundle/DataCollector/AttributeDataCollector.php
@@ -45,8 +45,12 @@ class AttributeDataCollector implements DataCollectorInterface
     /** @var AverageMaxQuery */
     private $localizableAndScopableAttributePerFamilyAverageMaxQuery;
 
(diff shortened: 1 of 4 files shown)


src/Pim/Bundle/AnalyticsBundle/DataCollector/AttributeDataCollector.php
```diff
@@ -103,6 +112,11 @@ public function collect(): array
             $numberOfLocalizableAndScopableAttributePerFamilyAverageMaxQuery = $this->localizableAndScopableAttributePerFamilyAverageMaxQuery->fetch()->getAverageVolume();
         }
 
+        $averageNumberOfAttributePerFamily = 0;
+        if (null !== $this->localizableAndScopableAttributePerFamilyAverageMaxQuery) {
```
ahocquard:
```suggestion
        if (null !== $this->attributePerFamilyAverageMaxQuery) {
```

PR #9087, opened 2018-11-08
Fix drag'n'drop of huge attribute group in family variant

damien-carcel:
## Description

In the variant family configuration, at the variant attributes screen, it is not possible to move a group of attributes if the group contains a large number of attributes:

fails

This drag'n'drop is managed by the `sortable` widget from JQuery UI. The issue is fixed by removing the use of its [`containment` option](https://api.jqueryui.com/sortable/#option-containment), as it is useless here: the sortable lists use all the window.

As a result, the drag'n'drop behaves as expected:

works

## Definition Of Done

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/CHANGELOG-2.3.md
+++ b/CHANGELOG-2.3.md
@@ -2,7 +2,8 @@
 
 ## Bug fixes
 
-- PIM-7771: Fix refresh versioning command about duplicate version's rule
+- PIM-7771: Fix refresh versioning command about duplicate version's rule.
+- PIM-7813: Fix a bug that prevents to drag'n'drop an attribute group containing a lot of attributes in the variant family configuration screen.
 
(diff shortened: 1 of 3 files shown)


CHANGELOG-2.3.md
```diff
@@ -2,7 +2,8 @@
 
 ## Bug fixes
 
-- PIM-7771: Fix refresh versioning command about duplicate version's rule
+- PIM-7771: Fix refresh versioning command about duplicate version's rule.
+- PIM-7813: Fix a bug that prevents to drag'n'drop a huge attribute group in variant family configuration screen.
```
grena: > (..) to drag'n'drop an attribute group containing a lot of attributes (...)

PR #9116, opened 2018-11-14
PIM-7853 - do not reload page if there is a form validation error

AymericPlanche:

When editing user information, we need to reload the page in order to have fresh information on UI.

**Problem**:
If there is validation errors (minimum number of characters not respected or so), the user is not saved and we want to display to the user that he entered wrong data in the form. But as we automatically reload the page the user didn't have the time to see his mistake.

**Solution**:
Check on the xhr response if we have validation error and reload or not the page.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Pim/Bundle/EnrichBundle/Resources/public/js/controller/user.js
+++ b/src/Pim/Bundle/EnrichBundle/Resources/public/js/controller/user.js
@@ -9,8 +9,11 @@ define([
             /**
              * {@inheritdoc}
              */
-            afterSubmit: function () {
-                window.location.reload(); //TODO nav: reload the page to update the menu
+            afterSubmit: function (xhr) {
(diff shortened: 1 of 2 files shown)


src/Pim/Bundle/EnrichBundle/Resources/public/js/controller/user.js
```diff
@@ -9,8 +9,11 @@ define([
             /**
              * {@inheritdoc}
              */
-            afterSubmit: function () {
-                window.location.reload(); //TODO nav: reload the page to update the menu
+            afterSubmit: function (xhr) {
+                if(xhr.responseText.indexOf('validation-tooltip') < 0)
```
pierallard: Please add comment to explain this

PR #9129, opened 2018-11-16
Merge 2.3 into master

Doodoune:


.gitignore
```diff
@@ -50,6 +50,7 @@ npm-debug.log
 !/docker/akeneo-behat.conf
 !/docker/httpd.conf
 .env
+docker-compose.yml
```
AymericPlanche: nope, the .yml have know to be versionned.
no .dist needed anymore because we have to use the .override.yml to "override" :)

Doodoune: Ok ! I change that :)

PR #9140, opened 2018-11-20
PIM-7855: Optional attributes are added when family column is not in the imported product file

grena:

Problem: When we import a product and the "family" column is not set, we can bypass the family attributes and are able to set optional attributes to product that have a family (_and thus **should not have optional attributes**_).

Solution: If the `family` column is not set in the imported file, we fetch the product by its identifier and then its family.

### DIFF
--- a/src/Pim/Bundle/CatalogBundle/tests/integration/Product/RemoveOptionalAttributeIntegration.php
+++ b/src/Pim/Bundle/CatalogBundle/tests/integration/Product/RemoveOptionalAttributeIntegration.php
@@ -15,7 +15,7 @@ class RemoveOptionalAttributeIntegration extends TestCase
     /**
      * @test
      */
-    function remove_an_attribute_on_a_product_without_family()
+    function test_remove_an_attribute_on_a_product_without_family()
     {
         /** @var EntityBuilder $entityBuilder */
         $entityBuilder = $this->getFromTestContainer('akeneo_integration_tests.catalog.fixture.build_entity');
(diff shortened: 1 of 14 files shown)


src/Pim/Component/Connector/Processor/Denormalization/ProductProcessor.php
```diff
@@ -188,7 +189,22 @@ protected function getIdentifier(array $item)
      */
     protected function getFamilyCode(array $item): string
     {
-        return $item['family'] ?? '';
+        if (key_exists('family', $item)) {
+            return $item['family'];
+        }
+
+        /** @var ProductInterface $product */
```
phaseinducer: no

src/Pim/Component/Connector/Processor/Denormalization/ProductProcessor.php
```diff
@@ -188,7 +189,21 @@ protected function getIdentifier(array $item)
      */
     protected function getFamilyCode(array $item): string
     {
-        return $item['family'] ?? '';
+        if (key_exists('family', $item)) {
+            return $item['family'];
+        }
+
+        $product = $this->repository->findOneByIdentifier($item['identifier']);
```
jmleroux: We will decrease performance. Is there really no other solution?

Doodoune:
You need to be rebased on 2.0 (changelog have change)

src/Pim/Bundle/CatalogBundle/tests/integration/Product/RemoveOptionalAttributeIntegration.php
```diff
@@ -0,0 +1,41 @@
...
+use PHPUnit\Framework\Assert;
+use Pim\Bundle\CatalogBundle\tests\fixture\EntityBuilder;
+use Pim\Component\Catalog\Model\ProductInterface;
+
+class RemoveOptionalAttributeIntegration extends TestCase
+{
+    /**
+     * @test
+     */
+    function remove_an_attribute_on_a_product_without_family()
```
mmetayer:
```suggestion
    function test_remove_an_attribute_on_a_product_without_family()
```
Otherwise it won' be run by the CI

PR #9157, opened 2018-11-26
Commit context map related images

jjanvier:
Small PR to add the material we used for the presentation about context map.
We will perform this context map presentation later. The materials are not finished. This is just to not lost them.


anaelChardan:
Maybe you could reference thoses images in "BOUNDED_CONTEXT.md" ? ;)

PR #9167, opened 2018-11-27
PIM-7660: get images of product models in the datagrid

ahocquard:

This is maybe one of the most complicated rule for for the read models.

When you display a product model, you have an image (attribute as image).

As the attribute as image is not necessary in the product model or its parent level, you have to get it from its first created child (a child can be a sub  product model or a variant product, depending of where is the attribute as image).

Our storage is perfect to handle product model feature when getting data from the bottom (variant  product) to the top (root product model).
But it is pretty complicated to do it from top to bottom (actually, it is not designed for it). 

I enumerate in the phpdoc all the possibilities. 

It replaces this "simple" class (but not that much when you dive into the code):
https://github.com/akeneo/pim-community-dev/blob/master/src/Akeneo/Pim/Enrichment/Component/Product/ProductModel/ImageAsLabel.php

It was totally inefficient because it uses a product or a product model entities, loading categories, group, applying permissions, etc...  and using `getImage` function (it should not be in the write model).

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Akeneo/Pim/Enrichment/Bundle/Storage/Sql/ProductGrid/ProductModelImagesFromCodes.php
+++ b/src/Akeneo/Pim/Enrichment/Bundle/Storage/Sql/ProductGrid/ProductModelImagesFromCodes.php
@@ -1,0 +1,340 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Akeneo\Pim\Enrichment\Bundle\Storage\Sql\ProductGrid;
+
+use Akeneo\Pim\Enrichment\Component\Product\Factory\Value\MediaValueFactory;
+use Akeneo\Pim\Enrichment\Component\Product\Factory\ValueCollectionFactoryInterface;
+use Doctrine\DBAL\Connection;
+
+/**
+ * When requesting the label of product model in the datagrid, you can get it from:
+ * - the parent product model
+ *      if the attribute as image is at the level 0 and the product model is not root
+ *
+ * - the current product model
+ *      if the attribute as image is at the level 0 and the current product model is root
+ *      OR
+ *      if the attribute as image is at the level 1 and the current product model is a sub product model
+ *
+ * - a sub product model
+ *      if the attribute as image is at the level 1 and the current product model is a root product model
+ *      it will be the image of the oldest sub product model by creation date with a non null image
+ *
+ * - a child variant product:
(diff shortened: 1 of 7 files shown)


src/Akeneo/Pim/Enrichment/Bundle/Storage/Sql/ProductGrid/ProductModelImagesFromCodes.php
```diff
@@ -0,0 +1,334 @@
...
+        $sql = <<<SQL
+            SELECT 
+                pm_root.code,
+                a_image.code as attribute_code,
+                a_image.is_localizable,
+                a_image.is_scopable,
+                product_child.raw_values,
+                JSON_EXTRACT(
+                    product_child.raw_values,
+                    CONCAT('$.', a_image.code, '.', IF(is_scopable = 1, '"$channelCode"', '"<all_channels>"'), '.', IF(is_localizable = 1, '"$localeCode"', '"<all_locales>"'))
```
jjanvier: is it safe to inject a variable like that?

> IF(is_scopable = 1, '"$channelCode"', '"<all_channels>"')

ahocquard: not at all...

ahocquard: fixed

src/Akeneo/Pim/Enrichment/Bundle/Storage/Sql/ProductGrid/ProductModelImagesFromCodes.php
```diff
@@ -0,0 +1,340 @@
...
+        return $images;
+    }
+
+    /**
+     * It gets image from the oldest child product model having a non null image.
+     *
+     * As we can't get easily the first row of each group when doing a group by,
+     * it executes one request per product model.
+     *
+     * @param array  $codes
```
BitOne: Useless?

ahocquard: It depends of the developer... we have no strict rule about it. I personally think that it's useless except for exceptions.

PR #9171, opened 2018-11-28
Remove deprecated

anaelChardan:

`OK`: Done / Validated
`-`: Not needed


src/Akeneo/Channel/Bundle/Controller/ExternalApi/ChannelController.php
```diff
@@ -259,28 +273,6 @@ public function partialUpdateListAction(Request $request)
...
-     * `conversion_units` are not well exposed through the api, on the api side it is an object while in ChannelInterface it is an array.
-     * To follow the API merge rules on object https://api-staging.akeneo.com/documentation/update.html#patch-rules,
-     * we are forced to process data before updating them, we cannot change this behavior everywhere to avoid BC breaks.
-     *
-     * @param ChannelInterface $channel
-     * @param array            $data
-     *
-     * @return array
-     */
-    protected function mergeAndFilterConversionUnits($channel, $data)
```
jjanvier: a deprecation about a method protected method that is not to implement by the client makes no sense imo

why don't you just turn it to private?

anaelChardan: I'm ok with that, should I remove the deprecated comment ?

PR #9176, opened 2018-11-29
Add docker support in makefile

anaelChardan:

This PR adds the support for docker in makefile

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,19 +1,26 @@
+DOCKER ?= false
+PHP_EXEC :=php
+
+ifeq ($(DOCKER),true)
+	PHP_EXEC=docker-compose exec -u docker fpm php
+endif
+
 ### Coupling detection
 .PHONY: structure-coupling
 structure-coupling:
-	vendor/bin/php-coupling-detector detect --config-file=src/Akeneo/Pim/Structure/.php_cd.php src/Akeneo/Pim/Structure
+	$(PHP_EXEC) vendor/bin/php-coupling-detector detect --config-file=src/Akeneo/Pim/Structure/.php_cd.php src/Akeneo/Pim/Structure
 
 .PHONY: user-management-coupling
 user-management-coupling:
-	vendor/bin/php-coupling-detector detect --config-file=src/Akeneo/UserManagement/.php_cd.php src/Akeneo/UserManagement
+	$(PHP_EXEC) vendor/bin/php-coupling-detector detect --config-file=src/Akeneo/UserManagement/.php_cd.php src/Akeneo/UserManagement
 
 .PHONY: channel-coupling
 channel-coupling:
-	vendor/bin/php-coupling-detector detect --config-file=src/Akeneo/Channel/.php_cd.php src/Akeneo/Channel
+	$(PHP_EXEC) vendor/bin/php-coupling-detector detect --config-file=src/Akeneo/Channel/.php_cd.php src/Akeneo/Channel
 
 .PHONY: enrichment-coupling
 enrichment-coupling:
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -1,19 +1,55 @@
...
+
+.PHONY: help-community
+help-community: ## Show the community help
+	@echo ""
+	@echo "Pim Community Dev available targets: "
+	@echo ""
+	@echo "You can use docker by using an environment variable, PLATFORM=docker"
+	@echo ""
+	@echo "========================================"
+	@awk 'BEGIN {FS = ":.*?## "} /^[a-zA-Z_-]+:.*?## / {printf "\033[36m%-$(HELP_TARGET_COLUMN_WIDTH)s\033[0m %s\n", $$1, $$2}' $(MAKEFILE_LIST) | sort
```
BitOne: Nop, please don't force color output...You know how it can end when you want to analyze it...

anaelChardan: Fixed ! ;) I totally agree :D

jjanvier:
This seems to add a lot of complexity :/ 

Can't the people who want to use the Makefile log into their container, and execute the Makefile?
What's the problem with that?

anaelChardan:
@jjanvier the goal is to be transparent for everyone, yes they can but they won't benefit from the autocompletion and so on, it's the matter of 2 variables :D

jjanvier:
When the Makefile was added, I saw it as a big opportunity to simplify our recurring tasks. But when I see that PR, I'm afraid I won't contribute as much or easily as I expected.

anaelChardan:
To add a task it's totally transparent what are you afraid of? Of adding `eval ${PHP_EXEC} ` in front of your task ?

BitOne:
I agree with JJ. With this PR, half of the makefile is to handle something that's not the purpose of the makefile...

jjanvier:
Not, it's not "2 variables". Yes, `eval ${PHP_EXEC} ` is much more complex than `php`. 

I expected we could have something like that https://speakerdeck.com/mykiwi/outils-pour-ameliorer-la-vie-des-developpeurs-symfony. (Or video https://www.youtube.com/watch?v=_OEDoPMvNY4).

Their Makefile is really great, I hope we'll have something like that. But with your solution it will soon be very complex to handle. 

Docker should be transparent right? Isn't it the original promise? Why should we change our code for Docker? It's a non sense imo...

anaelChardan:
@jjanvier In the link you provided the Makefile is simpler as the makefile uses docker directly ;)

Makefile
```diff
@@ -1,19 +1,28 @@
+DOCKER ?= false
+PHP_EXEC_CE :=""
```
BitOne: PHP_EXEC_CE => PHP_EXEC ;)

Makefile
```diff
@@ -1,19 +1,28 @@
...
+NODE_EXEC_CE :=""
```
BitOne: Same ;)

jjanvier:
In the link I provided, there are 2 parts:
1. the original makefile, without any Docker
2. the adaptation for Docker by using variables such as $(EXEC_PHP)

In their team, they all use Docker so yes it's more simple. 

Can we get rid of the `eval` at least? OK to write something like `$(EXEC_PHP) the command` if it's the compromise we have to do so that everybody can use the Makefile.

At least your PR is easier to understand now that it was before.

anaelChardan:
@jjanvier you can GTM it is as simple as you wish I hope ;)

damien-carcel:
Makefile cannot be executed in the containers. If, for instance, you want to add a `make front`, you need to call both PHP (for Symfony assets) and Yarn (for webpack), so 2 containers.
IMHO, Makefile has to be transparent, or it will be only for those who don't use Docker.

PR #9197, opened 2018-12-03
PIM-7897: Add cache to fetchAll in attribute group fetcher

Doodoune:

We have several requests that are sent 3 or 4 times, like this one:
[...]/rest/attribute-group/?options%5Blimit%5D=-1

screenshot 2018-11-29 at 15 47 02 1

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Pim/Bundle/EnrichBundle/Resources/public/js/fetcher/attribute-group-fetcher.js
+++ b/src/Pim/Bundle/EnrichBundle/Resources/public/js/fetcher/attribute-group-fetcher.js
@@ -36,18 +36,22 @@ define([
          *
          * {@inheritdoc}
          */
-        fetchAll: function () {
-            if (!_.has(this.options.urls, 'list')) {
-                return $.Deferred().reject().promise();
-            }
+        fetchAll: function (options) {
+            options = options || {};
 
-            this.entityListPromise = $.getJSON(
-                Routing.generate(this.options.urls.list, {
-                    options: {
-                        limit: -1
-                    }
-                })
-            ).then(_.identity).promise();
+            if (null === this.entityListPromise || false === options.cached) {
+                if (!_.has(this.options.urls, 'list')) {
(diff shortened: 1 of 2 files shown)


src/Pim/Bundle/EnrichBundle/Resources/public/js/fetcher/attribute-group-fetcher.js
```diff
@@ -36,18 +36,22 @@ define([
...
-                Routing.generate(this.options.urls.list, {
-                    options: {
-                        limit: -1
-                    }
-                })
-            ).then(_.identity).promise();
+                this.entityListPromise = $.getJSON(
+                    Routing.generate(this.options.urls.list, {
+                        options: {
+                            limit: -1
```
SamirBoulil: What is the limit for ?

src/Pim/Bundle/EnrichBundle/Resources/public/js/fetcher/attribute-group-fetcher.js
```diff
@@ -36,18 +36,22 @@ define([
...
          * {@inheritdoc}
          */
-        fetchAll: function () {
-            if (!_.has(this.options.urls, 'list')) {
-                return $.Deferred().reject().promise();
-            }
+        fetchAll: function (options) {
+            options = options || {};
+
+            if (null === this.entityListPromise || false === options.cached) {
```
SamirBoulil: where is `options.cached` initialized ?

Doodoune: see with yohan, 
- the cache should be enabled by default
- when it's not "initialize", it's undefined => so enable by default.
So we don't need to "initialize" it, it will always be cacheable and if you say "cache: false", it will be not cachable.

PR #9214, opened 2018-12-05
TIP-896: Fix some minor coupling issues

arnolanglade:

This PR fixes to small coupling issues, the first one is that we register a compiler pass in the wrong bounded context and the second is that we catch an exception (from the ES client) which are never thrown. 

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Akeneo/Pim/Enrichment/Bundle/Controller/ExternalApi/ProductController.php
+++ b/src/Akeneo/Pim/Enrichment/Bundle/Controller/ExternalApi/ProductController.php
@@ -34,7 +34,6 @@ use Akeneo\Tool\Component\StorageUtils\Remover\RemoverInterface;
 use Akeneo\Tool\Component\StorageUtils\Repository\IdentifiableObjectRepositoryInterface;
 use Akeneo\Tool\Component\StorageUtils\Saver\SaverInterface;
 use Akeneo\Tool\Component\StorageUtils\Updater\ObjectUpdaterInterface;
-use Elasticsearch\Common\Exceptions\ServerErrorResponseException;
 use Symfony\Component\HttpFoundation\JsonResponse;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
@@ -670,8 +669,6 @@ class ProductController
      * @param array                 $normalizerOptions
      *
      * @return array
-     *
-     * @throws ServerErrorResponseException
      */
     protected function searchAfterOffset(
         Request $request,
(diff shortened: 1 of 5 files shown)


src/Akeneo/Pim/Enrichment/Bundle/Controller/ExternalApi/ProductController.php
```diff
@@ -706,26 +703,12 @@ protected function searchAfterOffset(
...
-                throw new DocumentedHttpException(
-                    Documentation::URL_DOCUMENTATION . 'pagination.html#search-after-type',
-                    'You have reached the maximum number of pages you can retrieve with the "page" pagination type. Please use the search after pagination type instead',
-                    $e
-                );
-            }
-
-            throw new ServerErrorResponseException($e->getMessage(), $e->getCode(), $e);
-        }
+        $count = 'true' === $queryParameters['with_count'] ? $products->count() : null;
```
ahocquard: Should be validated by PO, as it was a request from @NolwennP

ahocquard: Also, I don't get the "which are never thrown.". Could you elaborate?

arnolanglade: It seems that the pagination does not throw this kind of exception because It only depends on the RouterInterface. I didn't find any normalizer which throw this exception. Maybe, I am wrong, do you know where are thrown those exceptions?

ahocquard: Pagination call the cursor, whih call the PQB, which call ES I suppose, which trigger the ServerException.

Then you have a normalizer for this exception:
https://github.com/akeneo/pim-community-dev/blob/master/src/Akeneo/Tool/Component/Api/Normalizer/Exception/DocumentedNormalizer.php


Note: I think this part has been done in a SLA. I don't think that's a good solution to do this kind of stuff this way. It should be documented in a troubleshooting guide. We will catch all errors from ES with a specific link...

arnolanglade: Do we speak about https://github.com/akeneo/pim-community-dev/blob/master/src/Akeneo/Tool/Component/Api/Pagination/OffsetHalPaginator.php? It seems it does not use the PQB!

ahocquard: let's see it IRL :)

PR #9272, opened 2018-12-18
TIP-899: Improve product export performance by computing headers at the end

BitOne:

In order to get all possible columns in a flat export file, we fill all the products loaded from the DB with all possible values but with empty data.
This is pretty inefficient, as this creates a lot of value objects that will generate at the end the exact same headers.

This PR changes the algorithm to generate the headers only at the end of the export by collecting all the families of the products exported and when the writer is flushed, by generating all the headers for the collected families attributes information.

The benchmark shows that even with products totally filled this solution is still faster (1.4x times) and goes up to 6x times faster with products lightly filled.

Benchmark result (Reference catalog with 10k products in DB, 8k products exported):
- 5% of product values filled: 6x faster (30s vs 3m30s)
- 50% of product values filled: 1.5x faster (3m56s vs 6m3s)
- 100% of product values filled: 1.4x faster (6m10s vs 8m20s)

About ProductModel support:
This PR still fills empty values for ProductModel by checking ProductModel in the reader, which breaks the genericity of the reader (maybe it should not be generic...).
Anyway; the first draft of this PR had a beginning of support for generating headers for ProductModel, so this line didn't exist.
But including in the same PR support for ProductModel was growing too much, because:
 - the business rules for generating headers for productModel flat files is more complex than with products as only headers belonging to the product model and its parents must be in the file. I cannot put the headers for the whole family, unless I change the functional behavior
 - the current state of the ProductModel and FamilyVariant modelling in DB makes it very difficult to list the attributes belonging to a FamilyVariant, and the tree structure itself of the product model doesn't help to have simple queries doing the job.

So I decided to work on the ProductModel side on another PR. Moreover the pain is mostly on Product export side for the moment, so that's where I put the priority.

TODO:
 - [x] EE PR for permissions on attributes
 - [x] Integration test for the queries
 
### DIFF
--- a/src/Akeneo/Pim/Enrichment/Bundle/Storage/Sql/Connector/Writer/File/Flat/GenerateHeadersFromAttributeCodes.php
+++ b/src/Akeneo/Pim/Enrichment/Bundle/Storage/Sql/Connector/Writer/File/Flat/GenerateHeadersFromAttributeCodes.php
@@ -1,0 +1,91 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Akeneo\Pim\Enrichment\Bundle\Storage\Sql\Connector\Writer\File\Flat;
+
+use Akeneo\Pim\Enrichment\Component\Product\Connector\Writer\File\FlatFileHeader;
+use Akeneo\Pim\Enrichment\Component\Product\Connector\Writer\File\GenerateFlatHeadersFromAttributeCodesInterface;
+use Doctrine\DBAL\Connection;
+
+/**
+ * @author    Benoit Jacquemont <benoit@akeneo.com>
+ * @copyright 2018 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+final class GenerateHeadersFromAttributeCodes implements GenerateFlatHeadersFromAttributeCodesInterface
+{
+    /** @var Connection */
+    private $connection;
+
+    public function __construct(Connection $connection)
+    {
+        $this->connection = $connection;
+    }
+
+    /**
+     * Generate all possible headers from the provided attribute codes
+     *
+     * @return FlatFileHeader[]
+     */
+    public function __invoke(
+        array $attributeCodes,
+        string $channelCode,
+        array $localeCodes
+    ): array {
+        $activatedCurrencyCodes = $this->connection->executeQuery(
+            "SELECT code FROM pim_catalog_currency WHERE is_activated = 1"
+        )->fetchAll(\PDO::FETCH_COLUMN, 0);
+
+        $channelCurrencyCodesSql = <<<SQL
+            SELECT currency.code
+            FROM pim_catalog_channel channel
+              JOIN pim_catalog_channel_currency cc ON cc.channel_id = channel.id
+              JOIN pim_catalog_currency currency ON currency.id = cc.currency_id
+            WHERE channel.code = :channelCode
+SQL;
+        $channelCurrencyCodes = $this->connection->executeQuery(
+            $channelCurrencyCodesSql,
+            ['channelCode' => $channelCode]
+        )->fetchAll(\PDO::FETCH_COLUMN, 0);
+
+        $attributesDataSql = <<<SQL
+            SELECT a.code,
+                   a.is_scopable,
+                   a.is_localizable,
+                   a.attribute_type,
+                   (
+                     SELECT JSON_ARRAYAGG(l.code)
+                     FROM pim_catalog_locale l
+                     JOIN pim_catalog_attribute_locale al ON al.locale_id = l.id
+                     WHERE al.attribute_id = a.id
+                   ) AS specific_to_locales
+            FROM pim_catalog_attribute a
+            WHERE a.code IN (:attributeCodes)
+            GROUP BY a.id;
+SQL;
+
+        $attributesData = $this->connection->executeQuery(
+            $attributesDataSql,
+            ['attributeCodes' => $attributeCodes],
+            ['attributeCodes' => \Doctrine\DBAL\Connection::PARAM_STR_ARRAY]
+        )->fetchAll();
+
+        $headers = [];
+        foreach ($attributesData as $attributeData) {
+            $headers[] = FlatFileHeader::buildFromAttributeData(
+                $attributeData["code"],
+                $attributeData["attribute_type"],
+                ("1" === $attributeData["is_scopable"]),
+                $channelCode,
+                ("1" === $attributeData["is_localizable"]),
+                $localeCodes,
+                $channelCurrencyCodes,
+                $activatedCurrencyCodes,
+                null !== $attributeData['specific_to_locales'] ? json_decode($attributeData['specific_to_locales'], true) : []
+            );
+        }
+
+        return $headers;
+    }
+}
--- a/src/Akeneo/Pim/Enrichment/Bundle/Storage/Sql/Connector/Writer/File/Flat/GenerateHeadersFromFamilyCodes.php
+++ b/src/Akeneo/Pim/Enrichment/Bundle/Storage/Sql/Connector/Writer/File/Flat/GenerateHeadersFromFamilyCodes.php
@@ -1,0 +1,93 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Akeneo\Pim\Enrichment\Bundle\Storage\Sql\Connector\Writer\File\Flat;
+
+use Akeneo\Pim\Enrichment\Component\Product\Connector\Writer\File\FlatFileHeader;
+use Akeneo\Pim\Enrichment\Component\Product\Connector\Writer\File\GenerateFlatHeadersFromFamilyCodesInterface;
+use Doctrine\DBAL\Connection;
+
+/**
+ * @author    Benoit Jacquemont <benoit@akeneo.com>
+ * @copyright 2018 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+final class GenerateHeadersFromFamilyCodes implements GenerateFlatHeadersFromFamilyCodesInterface
+{
+    /** @var Connection */
+    private $connection;
+
+    public function __construct(Connection $connection)
+    {
+        $this->connection = $connection;
+    }
+
+    /**
+     * Generate all possible headers from the provided family codes
+     *
+     * @return FlatFileHeader[]
+     */
+    public function __invoke(
+        array $familyCodes,
+        string $channelCode,
+        array $localeCodes
+    ): array {
+        $activatedCurrencyCodes = $this->connection->executeQuery(
+            "SELECT code FROM pim_catalog_currency WHERE is_activated = 1"
+        )->fetchAll(\PDO::FETCH_COLUMN, 0);
+
+        $channelCurrencyCodesSql = <<<SQL
+            SELECT currency.code
+            FROM pim_catalog_channel channel
+              JOIN pim_catalog_channel_currency cc ON cc.channel_id = channel.id
+              JOIN pim_catalog_currency currency ON currency.id = cc.currency_id
+            WHERE channel.code = :channelCode
+SQL;
+        $channelCurrencyCodes = $this->connection->executeQuery(
+            $channelCurrencyCodesSql,
+            ['channelCode' => $channelCode]
+        )->fetchAll(\PDO::FETCH_COLUMN, 0);
+
+        $attributesDataSql = <<<SQL
+            SELECT a.code,
+                   a.is_scopable,
+                   a.is_localizable,
+                   a.attribute_type,
+                   (
+                     SELECT JSON_ARRAYAGG(l.code)
+                     FROM pim_catalog_locale l
+                     JOIN pim_catalog_attribute_locale al ON al.locale_id = l.id
+                     WHERE al.attribute_id = a.id
+                   ) AS specific_to_locales
+            FROM pim_catalog_family f
+              JOIN pim_catalog_family_attribute fa ON fa.family_id = f.id
+              JOIN pim_catalog_attribute a ON a.id = fa.attribute_id
+            WHERE f.code IN (:familyCodes)
+            GROUP BY a.id;
+SQL;
+
+        $attributesData = $this->connection->executeQuery(
+            $attributesDataSql,
+            ['familyCodes' => $familyCodes],
+            ['familyCodes' => \Doctrine\DBAL\Connection::PARAM_STR_ARRAY]
+        )->fetchAll();
+
+        $headers = [];
+        foreach ($attributesData as $attributeData) {
+            $headers[] = FlatFileHeader::buildFromAttributeData(
+                $attributeData["code"],
+                $attributeData["attribute_type"],
+                ("1" === $attributeData["is_scopable"]),
+                $channelCode,
+                ("1" === $attributeData["is_localizable"]),
+                $localeCodes,
+                $channelCurrencyCodes,
+                $activatedCurrencyCodes,
+                null !== $attributeData['specific_to_locales'] ? json_decode($attributeData['specific_to_locales'], true) : []
+            );
+        }
+
+        return $headers;
+    }
+}
--- a/src/Akeneo/Pim/Enrichment/Component/Product/Connector/Processor/Normalization/ProductProcessor.php
+++ b/src/Akeneo/Pim/Enrichment/Component/Product/Connector/Processor/Normalization/ProductProcessor.php
@@ -55,7 +55,7 @@ class ProductProcessor implements ItemProcessorInterface, StepExecutionAwareInte
         IdentifiableObjectRepositoryInterface $channelRepository,
         AttributeRepositoryInterface $attributeRepository,
         BulkMediaFetcher $mediaFetcher,
-        EntityWithFamilyValuesFillerInterface $productValuesFiller
+        ?EntityWithFamilyValuesFillerInterface $productValuesFiller = null
     ) {
         $this->normalizer          = $normalizer;
         $this->channelRepository   = $channelRepository;
@@ -72,7 +72,9 @@ class ProductProcessor implements ItemProcessorInterface, StepExecutionAwareInte
         $parameters = $this->stepExecution->getJobParameters();
         $structure = $parameters->get('filters')['structure'];
         $channel = $this->channelRepository->findOneByIdentifier($structure['scope']);
-        $this->productValuesFiller->fillMissingValues($product);
+        if ($product instanceof ProductModelInterface) {
+            $this->productValuesFiller->fillMissingValues($product);
+        }
 
         $productStandard = $this->normalizer->normalize(
             $product,
--- a/src/Akeneo/Pim/Enrichment/Component/Product/Connector/Writer/File/Csv/ProductWriter.php
+++ b/src/Akeneo/Pim/Enrichment/Component/Product/Connector/Writer/File/Csv/ProductWriter.php
@@ -2,12 +2,20 @@
 
 namespace Akeneo\Pim\Enrichment\Component\Product\Connector\Writer\File\Csv;
 
+use Akeneo\Pim\Enrichment\Component\Product\Connector\Writer\File\GenerateFlatHeadersFromAttributeCodesInterface;
+use Akeneo\Pim\Enrichment\Component\Product\Connector\Writer\File\GenerateFlatHeadersFromFamilyCodesInterface;
+use Akeneo\Pim\Structure\Component\Repository\AttributeRepositoryInterface;
 use Akeneo\Tool\Component\Batch\Item\FlushableInterface;
 use Akeneo\Tool\Component\Batch\Item\InitializableInterface;
 use Akeneo\Tool\Component\Batch\Item\ItemWriterInterface;
+use Akeneo\Tool\Component\Batch\Job\JobParameters;
 use Akeneo\Tool\Component\Batch\Step\StepExecutionAwareInterface;
+use Akeneo\Tool\Component\Buffer\BufferFactory;
+use Akeneo\Tool\Component\Connector\ArrayConverter\ArrayConverterInterface;
 use Akeneo\Tool\Component\Connector\Writer\File\AbstractItemMediaWriter;
 use Akeneo\Tool\Component\Connector\Writer\File\ArchivableWriterInterface;
+use Akeneo\Tool\Component\Connector\Writer\File\FileExporterPathGeneratorInterface;
+use Akeneo\Tool\Component\Connector\Writer\File\FlatItemBufferFlusher;
 
 /**
  * Write product data into a csv file on the local filesystem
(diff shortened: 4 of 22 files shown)


src/Akeneo/Pim/Enrichment/Bundle/Storage/Sql/Connector/Writer/File/Flat/GenerateHeadersFromFamilyCodes.php
```diff
@@ -0,0 +1,167 @@
...
+namespace Akeneo\Pim\Enrichment\Bundle\Storage\Sql\Connector\Writer\File\Flat;
+
+use Akeneo\Pim\Enrichment\Component\Product\Connector\Writer\File\FlatFileHeader;
+use Akeneo\Pim\Enrichment\Component\Product\Connector\Writer\File\GenerateFlatHeadersInterface;
+use Akeneo\Pim\Structure\Component\AttributeTypes;
+use Doctrine\DBAL\Connection;
+
+/**
+ * @author    Benoit Jacquemont <benoit@akeneo.com>
+ * @copyright 2016 Akeneo SAS (http://www.akeneo.com)
```
pierallard: Nom de Zeus, Marty !

src/Akeneo/Pim/Enrichment/Bundle/Storage/Sql/Connector/Writer/File/Flat/GenerateHeadersFromFamilyCodes.php
```diff
@@ -0,0 +1,167 @@
...
+            $attributesDataSql,
+            ['familyCodes' => $familyCodes],
+            ['familyCodes' => \Doctrine\DBAL\Connection::PARAM_STR_ARRAY]
+        )->fetchAll();
+    }
+
+    /**
+     * Retrieve all activated currencies codes
+     */
+    protected function getActivatedCurrencyCodes(): array
```
pierallard: You already have it on the CurrencyRepository

src/Akeneo/Pim/Enrichment/Component/Product/Connector/Writer/File/GenerateFlatHeadersInterface.php
```diff
@@ -0,0 +1,28 @@
...
+declare(strict_types=1);
+
+namespace Akeneo\Pim\Enrichment\Component\Product\Connector\Writer\File;
+
+use Akeneo\Pim\Enrichment\Component\Product\Connector\Writer\File\FlatFileHeader;
+use Akeneo\Pim\Enrichment\Component\Product\Connector\Writer\File\GenerateFlatHeadersInterface;
+
+/**
+ * @author    Benoit Jacquemont <benoit@akeneo.com>
+ * @copyright 2016 Akeneo SAS (http://www.akeneo.com)
```
pierallard: Same

src/Akeneo/Pim/Enrichment/Bundle/Storage/Sql/Connector/Writer/File/Flat/GenerateHeadersFromFamilyCodes.php
BitOne: Yes, but it shouldn't be there as a Repository is not supposed to provide codes (it should only provide instances). And in term of coupling, it's not a good idea either.

src/Akeneo/Pim/Enrichment/Bundle/Storage/Sql/Connector/Writer/File/Flat/GenerateHeadersFromAttributeCodes.php
```diff
@@ -0,0 +1,97 @@
...
+use Akeneo\Pim\Enrichment\Component\Product\Connector\Writer\File\GenerateFlatHeadersFromAttributeCodesInterface;
+use Akeneo\Pim\Structure\Component\AttributeTypes;
+use Doctrine\DBAL\Connection;
+
+/**
+ * @author    Benoit Jacquemont <benoit@akeneo.com>
+ * @copyright 2018 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+class GenerateHeadersFromAttributeCodes implements GenerateFlatHeadersFromAttributeCodesInterface
```
ahocquard: missing integration test

src/Akeneo/Pim/Enrichment/Bundle/Storage/Sql/Connector/Writer/File/Flat/GenerateHeadersFromFamilyCodes.php
```diff
@@ -0,0 +1,98 @@
...
+use Akeneo\Pim\Enrichment\Component\Product\Connector\Writer\File\GenerateFlatHeadersFromFamilyCodesInterface;
+use Akeneo\Pim\Structure\Component\AttributeTypes;
+use Doctrine\DBAL\Connection;
+
+/**
+ * @author    Benoit Jacquemont <benoit@akeneo.com>
+ * @copyright 2018 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+class GenerateHeadersFromFamilyCodes implements GenerateFlatHeadersFromFamilyCodesInterface
```
ahocquard: missing integration test

src/Akeneo/Pim/Enrichment/Bundle/Storage/Sql/Connector/Writer/File/Flat/GenerateHeadersFromAttributeCodes.php
```diff
@@ -0,0 +1,97 @@
...
+    {
+        $this->connection = $connection;
+    }
+
+    /**
+     * Generate all possible headers from the provided attribute codes
+     *
+     * @return FlatFileHeader[]
+     */
+    public function __invoke(
```
ahocquard: FYI: https://github.com/akeneo/pim-community-dev/pull/7792#discussion_r173228392

My point of view: I'm ok with that as I see it as a kind of a typed callable. You don't implement several callable at a time, as a callable is by nature a function, which is unique.

src/Akeneo/Pim/Enrichment/Bundle/Storage/Sql/Connector/Writer/File/Flat/GenerateHeadersFromAttributeCodes.php
```diff
@@ -0,0 +1,97 @@
...
+            $channelCurrencyCodesSql,
+            ['channelCode' => $channelCode]
+        )->fetchAll(\PDO::FETCH_COLUMN, 0);
+
+        $attributesDataSql = <<<SQL
+            SELECT a.code,
+                   a.is_scopable,
+                   a.is_localizable,
+                   a.attribute_type,
+                   GROUP_CONCAT(l.code) AS specific_to_locales
```
ahocquard: I would use JSON_ARRAYAGG, as it's less a hack with a separator.
Also, there is no limit with JSON_ARRAYAGG.

src/Akeneo/Pim/Enrichment/Component/Product/Connector/Processor/Normalization/ProductProcessor.php
```diff
@@ -72,7 +72,9 @@ public function process($product)
         $parameters = $this->stepExecution->getJobParameters();
         $structure = $parameters->get('filters')['structure'];
         $channel = $this->channelRepository->findOneByIdentifier($structure['scope']);
-        $this->productValuesFiller->fillMissingValues($product);
+        if ($product instanceof ProductModelInterface) {
```
ahocquard: is it planned to do it for product models? This optional dependency is pretty weird

src/Akeneo/Pim/Enrichment/Component/Product/Connector/Writer/File/AbstractEntityWithFamilyFlatWriter.php
```diff
@@ -0,0 +1,136 @@
...
+use Akeneo\Tool\Component\Connector\Writer\File\AbstractItemMediaWriter;
+use Akeneo\Tool\Component\Connector\Writer\File\FileExporterPathGeneratorInterface;
+use Akeneo\Tool\Component\Connector\Writer\File\FlatItemBufferFlusher;
+
+/**
+ * @author    Benoit Jacquemont <benoit@akeneo.com>
+ * @copyright 2018 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+abstract class AbstractEntityWithFamilyFlatWriter extends AbstractItemMediaWriter
```
ahocquard: is it needed to extend the abstract? Is `AbstractItemMediaWriter` used as a type, or just as a way of avoiding code duplication?
If it's the last option, I generally prefer to copy the content of `AbstractItemMediaWriter` into this abstract to avoid complexity just for DRY.

src/Akeneo/Pim/Enrichment/Component/Product/Connector/Writer/File/FlatFileHeader.php
```diff
@@ -0,0 +1,159 @@
...
+
+/**
+ * @author    Benoit Jacquemont <benoit@akeneo.com>
+ * @copyright 2018 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+class FlatFileHeader
+{
+    /** @var string */
+    protected $code;
```
ahocquard: is there a need to extend it? Otherwise, final + private  in my opinion(I don't think there is a identified need to extend it).

tests/back/Pim/Enrichment/Specification/Component/Product/Connector/Writer/File/Csv/ProductWriterSpec.php
```diff
@@ -343,4 +351,147 @@ function it_writes_the_csv_file(
...
+        $bufferFactory,
+        $flusher,
+        $generateHeadersFromAttributeCodes,
+        FlatItemBuffer $flatRowBuffer,
+        StepExecution $stepExecution,
+        JobParameters $jobParameters,
+        JobExecution $jobExecution,
+        JobInstance $jobInstance,
+        ExecutionContext $executionContext,
+        FlatFileHeader $nameHeader,
```
ahocquard: if it's final, it will forces us to use the real object instead of mocking it. I think that it simplifies a lot the specification, and there are much more useful, as it's not a mock.

tests/legacy/features/pim/enrichment/product/mass-action/quick-export/quick_export_grid_context.feature
```diff
@@ -30,10 +30,10 @@ Feature: Quick export products according to the product grid context
     And the names of the exported files of "csv_product_grid_context_quick_export" should be "1_products_export_grid_context_en_US_tablet.csv,2_product_models_export_grid_context_en_US_tablet.csv"
     And first exported file of "csv_product_grid_context_quick_export" should contain:
     """
-    sku;color;description-en_US-tablet;family;groups;name-en_US;price-EUR;price-USD;size
-    boots;black;Mob;boots;;Amazing boots;20;25;40
-    sneakers;white;ylette;sneakers;;Sneakers;50;60;42
-    pump;blue;;;;Pump;15;20;41
+    sku;color;description-en_US-tablet;family;groups;name-en_US;price-EUR;price-USD;size;weight;weight-unit
```
ahocquard: Was it a bug?

tests/back/Pim/Enrichment/Specification/Component/Product/Connector/Writer/File/Csv/ProductWriterSpec.php
```diff
@@ -343,4 +351,147 @@ function it_writes_the_csv_file(
...
+        $bufferFactory,
+        $flusher,
+        $generateHeadersFromFamilyCodes,
+        FlatItemBuffer $flatRowBuffer,
+        StepExecution $stepExecution,
+        JobParameters $jobParameters,
+        JobExecution $jobExecution,
+        JobInstance $jobInstance,
+        ExecutionContext $executionContext,
+        FlatFileHeader $descHeader,
```
arnolanglade: You should not create a double for those objects.

src/Akeneo/Pim/Enrichment/Component/Product/Connector/Processor/Normalization/ProductProcessor.php
BitOne: Yes, it's definitely something dirty.
And the first draft of this PR had a beginning of support for generating headers for ProductModel, so this line didn't exist.
But including in the same PR support for ProductModel was growing too much, because:
 - the business rules for generating headers for productModel flat files is more complex than with products as only headers belonging to the product model and its parents must be in the file. I cannot put the headers for the whole family, unless I change the functional behavior
 - the current state of the ProductModel and FamilyVariant modelling in DB makes it very difficult to list the attributes belonging to a FamilyVariant, and the tree structure itself of the product model doesn't help to have simple queries doing the job.

So I decided to work on the ProductModel side on another PR. Moreover the pain is mostly on Product export side for the moment, so that's where I put the priority.

tests/legacy/features/pim/enrichment/product/mass-action/quick-export/quick_export_grid_context.feature
BitOne: Yes, it's not supposed to have all the currencies in case of an optional attribute.

src/Akeneo/Pim/Enrichment/Component/Product/Connector/Writer/File/FlatFileHeader.php
```diff
@@ -0,0 +1,159 @@
...
+        ?array $localeCodes,
+        bool $isMedia,
+        bool $usesUnit,
+        bool $usesCurrencies,
+        ?array $channelCurrencyCodes,
+        ?array $allCurrencyCodes,
+        bool $isLocaleSpecific,
+        ?array $specificToLocales = null
+    ) {
+        if ($isLocaleSpecific && empty($specificToLocales)) {
```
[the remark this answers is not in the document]
BitOne: No, it's not a read model. It's a domain object that contains the business logic of generating the headers. So I believe it must kind of protect itself from bad parameters combination?

src/Akeneo/Pim/Enrichment/Component/Product/Connector/Processor/Normalization/ProductProcessor.php
```diff
@@ -55,7 +55,7 @@ public function __construct(
         IdentifiableObjectRepositoryInterface $channelRepository,
         AttributeRepositoryInterface $attributeRepository,
         BulkMediaFetcher $mediaFetcher,
-        EntityWithFamilyValuesFillerInterface $productValuesFiller
+        ?EntityWithFamilyValuesFillerInterface $productValuesFiller = null
```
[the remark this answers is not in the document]
BitOne: Because the productValueFiller is used for ProductModel but not for Product anymore.

src/Akeneo/Pim/Enrichment/Component/Product/Connector/Writer/File/AbstractEntityWithFamilyFlatWriter.php
BitOne: :+1: I put everyting inside the concrete CSV and XLSX writer implementation.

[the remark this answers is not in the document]
BitOne: Yes, removed and put the code in the concrete implementations.

src/Akeneo/Pim/Enrichment/Bundle/Storage/Sql/Connector/Writer/File/Flat/GenerateHeadersFromAttributeCodes.php
BitOne: Yes, but @jjanvier argument makes sense: "This is a PHP dirty workaround (and still imo, we should not use any PHP workaround, only pure OOP stuff).". So what's the state of the decision about __invoke right now? I prefer homogenity, so if it's the "standard" way of working with query object, I will keep it like that.

jjanvier: (I didn't say we shouldn't use __invoke. I have no problem with that. I just have a problem with __invoke inside interfaces)

ahocquard: We avoid to use it in interface for now for all the PIM, except reference entities that use it.

Personally, I don't see it as interface, I see as typed function (as in Kotlin), and that's the way of doing it in PHP. 
But that's true it uses the "interface" keyword" and that the language does not allow typed function natively. But for a class it's the same fight: it uses "class" word but it's used as a class.

As a reminder, function typed in kotlin for example: (https://blog.kotlin-academy.com/kotlin-programmer-dictionary-function-type-vs-function-literal-vs-lambda-expression-vs-anonymous-edc97e8873e)

BitOne: The interface is a contract.

So if I use __invoke(), I shouldn't use it in the interface? Then the interface doesn't represent the contract anymore, which bothers quite a bit. In any other stronger typed language, that wouldn't be accepted... And in PHP, it may come soon that the language will refuse that as well.

So for me, it's either, we don't put invoke in the interface, so we must not use use invoke. And `fetch` is good as well.
Of we use __invoke() and it must be part of the contract defined by the interface.

src/Akeneo/Pim/Enrichment/Bundle/Storage/Sql/Connector/Writer/File/Flat/GenerateHeadersFromFamilyCodes.php
```diff
@@ -0,0 +1,93 @@
...
+            ['channelCode' => $channelCode]
+        )->fetchAll(\PDO::FETCH_COLUMN, 0);
+
+        $attributesDataSql = <<<SQL
+            SELECT a.code,
+                   a.is_scopable,
+                   a.is_localizable,
+                   a.attribute_type,
+                   (
+                     SELECT JSON_ARRAYAGG(l.code)
```
ahocquard: weird to aggregate without group by

src/Akeneo/Pim/Enrichment/Bundle/Storage/Sql/Connector/Writer/File/Flat/GenerateHeadersFromFamilyCodes.php
```diff
@@ -0,0 +1,93 @@
...
+                     FROM pim_catalog_locale l
+                     JOIN pim_catalog_attribute_locale al ON al.locale_id = l.id
+                     WHERE al.attribute_id = a.id
+                   ) AS specific_to_locales
+            FROM pim_catalog_family f
+              JOIN pim_catalog_family_attribute fa ON fa.family_id = f.id
+              JOIN pim_catalog_attribute a ON a.id = fa.attribute_id
+            WHERE f.code IN (:familyCodes)
+            GROUP BY a.id;
```
ahocquard: DISTINCT is really what you want

tests/back/Pim/Enrichment/Integration/Storage/Sql/Connector/Writer/File/Flat/AssertHeaders.php
```diff
@@ -0,0 +1,35 @@
...
+namespace AkeneoTest\Pim\Enrichment\Integration\Storage\Sql\Connector\Writer\File\Flat;
+
+use PHPUnit\Framework\Assert;
+use Akeneo\Pim\Enrichment\Component\Product\Connector\Writer\File\FlatFileHeader;
+
+final class AssertHeaders
+{
+    public static function same(array $expectedHeaders, array $actualHeaders): void
+    {
+        Assert::assertCount(count($expectedHeaders), $actualHeaders);
```
ahocquard: You could use Assert::equals from phpunit I think to avoid to do that. It asserts two objects are equals recursively.

src/Akeneo/Pim/Enrichment/Component/Product/Connector/Writer/File/Csv/ProductWriter.php
```diff
@@ -23,6 +31,121 @@ class ProductWriter extends AbstractItemMediaWriter implements
     StepExecutionAwareInterface,
     ArchivableWriterInterface
 {
+    /** @var array */
+    protected $familyCodes;
```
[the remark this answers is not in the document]
BitOne: Private without final is really a bad idea IMHO, as if somebody extends this class, two different variable with the same name will exists in two different scope: the private and the local. And this will create very weird behavior...

src/Akeneo/Pim/Enrichment/Component/Product/Connector/Writer/File/FlatFileHeader.php
```diff
@@ -0,0 +1,196 @@
...
+    private $allCurrencyCodes;
+
+    /** @var bool */
+    private $isLocaleSpecific;
+
+    /** @var array */
+    private $specificToLocales;
+
+    public function __construct(
+        string $code,
```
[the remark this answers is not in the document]
BitOne: It's not necessary coming from an attribute ;)

src/Akeneo/Pim/Enrichment/Component/Product/Connector/Writer/File/FlatFileHeader.php
```diff
@@ -0,0 +1,196 @@
...
+        ?bool $isScopable = false,
+        ?string $channelCode = null,
```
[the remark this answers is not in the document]
BitOne: Because the Header may reference a field that is not scopable. So the channelCode is useless.

src/Akeneo/Pim/Enrichment/Bundle/Storage/Sql/Connector/Writer/File/Flat/GenerateHeadersFromFamilyCodes.php
BitOne: How do you count the number of entries in a table :p

tests/back/Pim/Enrichment/Integration/Storage/Sql/Connector/Writer/File/Flat/AssertHeaders.php
BitOne: Yes, but in some case, the two objects may be different but can generate the same header strings.

ahocquard:
So, I did a benchmark of the requests on the reference catalog: 125 families, 550 attributes, 5 activated locale.

The fastest one that I did quickly:
```
            SELECT
				a.id,
				a.code,
				a.is_scopable,
				a.is_localizable,
				a.attribute_type,
				JSON_ARRAYAGG(l.code)
            FROM (
				SELECT DISTINCT fa.attribute_id 
				FROM pim_catalog_family f
				JOIN pim_catalog_family_attribute fa ON fa.family_id = f.id
			) as attribute
            JOIN pim_catalog_attribute a ON a.id = attribute.attribute_id
            LEFT JOIN pim_catalog_attribute_locale al ON al.attribute_id = a.id
			LEFT JOIN pim_catalog_locale l ON l.id = al.locale_id
            GROUP BY a.id;
```

There is not big `[null,null,null]` witht his query. It takes 10 ms (with all families).
The other one in your PR with the nested loop takes 40ms (the subquery is visible in the EXPLAIN and also in the profiling).
It takes 40ms.


My opinion:
It's maybe not perfect, but it's a premature optimization compared to the global time of a job (from 10 seconds to several hours).

PR #9293, opened 2018-12-24
Stabilize all the phpunit tests for circle ci

anaelChardan:
## What have been done

- Test dependent of the batch size are deleted, as they are useless and wrond as soon as yo modify the batch size.
- use refresh to wait the availability of the indexation when you serach in ES
- use sleep in one test, because the refresh was not enough for obscure reason (maybe a background process)

`OK`: Done / Validated
`-`: Not needed


.circleci/config.yml
```diff
@@ -3,19 +3,23 @@ jobs:
...
-            cp behat.yml.dist behat.yml
+          command: cp .env.dist .env
       - run:
           name: Setup tests results folder and log folder
           command: mkdir -p var/tests/phpunit var/tests/behat var/tests/phpspec var/tests/csfixer var/logs
       - run:
             name: Setup the parameters.yml files
             command: ./bin/docker/pim-setup.sh
+      - run:
+            name: Setup the batch size to 50
```
ahocquard: no, we should not modify it just to make the tests working.
It's a smell to indicate that the tests is not good.

Actually, the fix has to be done in the API: the batch size should not be used for the API. I did the same fix for the datagrid. 
Then, we can just delete failing tests...

https://github.com/akeneo/pim-community-dev/pull/8555

src/Akeneo/Pim/Enrichment/Component/Product/Query/ProductQueryBuilder.php
```diff
@@ -13,6 +13,7 @@
 use Akeneo\Pim\Structure\Component\Model\AttributeInterface;
 use Akeneo\Pim\Structure\Component\Repository\AttributeRepositoryInterface;
 use Akeneo\Tool\Component\StorageUtils\Cursor\CursorFactoryInterface;
+use function var_dump;
```
ahocquard: debug I suppose

tests/back/Pim/Enrichment/Integration/Doctrine/Common/Saver/SavingProductModelDescendantsIntegration.php
```diff
@@ -59,7 +59,7 @@ public function testIndexingProductModelDescendantsOnUnitarySave()
 
         // Since we set the "wait refresh" of ES to false, we have to wait for ES to actually refresh its shards
         // Cf. https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-refresh.html
-        sleep(3);
+        sleep(5);
```
ahocquard: no, you can wait until the next refresh of the index in the ES client.

tests/back/Pim/Enrichment/Integration/Elasticsearch/IndexConfiguration/SearchAncestors.php
```diff
@@ -9,7 +9,7 @@
  * @copyright 2017 Akeneo SAS (http://www.akeneo.com)
  * @license   http://opensource.org/licenses/osl-3.0.php Open Software License (OSL 3.0)
  */
-class SearchAncestorsIntegration extends AbstractPimCatalogProductModelIntegration
+class SearchAncestors extends AbstractPimCatalogProductModel
```
ahocquard: keepIntegration suffix, as it is used to search the tests.

anaelChardan: Ah oups PHPStorm did a mistake I just wanted to remove Integration from the abstract ;-).

tests/back/Pim/Enrichment/Integration/Doctrine/Common/Saver/SavingProductModelDescendantsIntegration.php
anaelChardan: I don’t know how to do that but that seems a better solution if we could get rid of those sleep

src/Akeneo/Pim/Enrichment/Component/Product/Query/ProductQueryBuilder.php
anaelChardan: Yes :-/

.circleci/config.yml
anaelChardan: Or just to change where we use this config in intégrations test. It is used for test about Pagination (via method getListSize) we can fix this number to 50 for example.

anaelChardan: But I do agree we shouldn’t change it for testing purpose. I wanted to introduce a new parameter as you said but on one PR it was maybe too big ;-)

anaelChardan: But it is the current behaviour on our current CI... we change this parameter in the dockerfile used by the CI

anaelChardan: I suspect the same. 😂

PR #9298, opened 2018-12-27
PIM-7926 - Set the parent attribute to null instead of an empty string

AymericPlanche:

In an import product definition you can "enable" or "disable" the comparison of the values of the product you try to update.

In "enable comparison" mode, if the value of an element is identical to the one in database, the product is skipped.

In "disable comparison" mode, we persist the data values as is.

For a regular "product" (not a variant product) the "parent" column in an import file must be absent or when the import file have mixed "product" and "variant product", the column "parent" is here but empty for "products".

**Problem**

When "disable comparison" is set to a product import job definition, and when you import a file containing mixed product, we try to update the product and set the parent value with an empty string.

**Solution**

In the `ProductAttributeFilter` class, if the parent value is empty, we set it's value to `null`.
In the `ParentFieldSetter`, if the value is `null` it's skipped.

I prefered set the value to `null` instead of `unset` the item in the array which have less impacts when dealing with mixed product/variant product.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/CHANGELOG-2.3.md
+++ b/CHANGELOG-2.3.md
@@ -3,6 +3,7 @@
 ## Bug fixes
 
 - PIM-7899: Remove Date of Birth field
+- PIM-7926: Fix the parent property setter when "enabled comparison" is set to false in an import job definition
 
 # 2.3.22 (2018-12-21)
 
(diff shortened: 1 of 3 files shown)


ahocquard:
Don't we have customers with invalid data in database?

AymericPlanche:
> Don't we have customers with invalid data in database?

Nope, when the problem occur, the simple product is skipped.

PR #9305, opened 2018-12-31
[behat] Dump SQL for catalogs initialization

pierallard:
Currently, each step `Given a xx catalog`
- loads the fixtures jobs
- loads the fixtures
- removes the fixtures jobs
For example, the bigger catalog (catalog_modeling) takes ~50s.

This PR dumps the SQL after a catalog initialization, in a .sql file. In the next catalog initialization, if this .sql already exists, it only reloads the SQL file.
For example, catalog_modeling is loaded in ~7s.

- [ ] To check before merge: Why we didn't put the .sql in the ~/var/cache folder ? Same question for the Integration tests .sql files.


pierallard:
https://github.com/akeneo/pim-community-dev/pull/9624

PR #9329, opened 2019-01-07
Aob 324 - Username in the PIM doesn't support the '@' and '.' character

AymericPlanche:

Some characters must be allowed into username field in order to ease Peter (the IT guy) administration tasks.

Tested without error:

- Create user (UI)
- Update user (UI)
- Login with username
- Update product (UI / API / Import)
- Create proposal (UI / API)

No errors, no notices, no UI glitches on username display.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Akeneo/UserManagement/Bundle/Controller/Rest/UserController.php
+++ b/src/Akeneo/UserManagement/Bundle/Controller/Rest/UserController.php
@@ -134,6 +134,7 @@ class UserController
      */
     public function getAction($identifier): JsonResponse
     {
+        $identifier = urldecode($identifier);
         $user = $this->repository->findOneByIdentifier($identifier);
 
         return new JsonResponse($this->normalizer->normalize($user, 'internal_api'));
@@ -265,6 +266,8 @@ class UserController
 
     private function getUserOr404($username): ?UserInterface
     {
+
+        $username = urldecode($username);
         $user = $this->repository->findOneByIdentifier($username);
 
         if (null === $user) {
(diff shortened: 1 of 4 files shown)


src/Akeneo/UserManagement/Bundle/Controller/Rest/UserController.php
```diff
@@ -134,6 +134,7 @@ public function getCurrentAction()
      */
     public function getAction($identifier): JsonResponse
     {
+        $identifier = urldecode($identifier);
```
AymericPlanche: manage the `%40` -> `@` conversion

src/Akeneo/UserManagement/Bundle/Resources/config/validation.yml
```diff
@@ -10,6 +10,9 @@ Akeneo\UserManagement\Component\Model\User:
             - Length:
                 min:        3
                 max:        255
+            - Regex:
```
skeleton: is it validated by PO to add constraints on the username? some users could not update their profil because of this new validation and will have to update their username (and lost in the same way their proposal :/ ) to be able to update it

AymericPlanche: 1. If they have today a username that doesn't respect the following pattern `[a-zA-Z0-9_]+` they can't access to their profile page so they can't update their username, as we add some constraints and allow them in the routing system it adds security, and control. (but it's a `pefize user` introduced issue - to discuss IRL)
2. As far as i've tested, if they update today their username, they loose their proposal, no behavior changes :)

skeleton: let's discuss about it tomorrow with Benj

AymericPlanche:
This PR will not be merged. Some behavior fixed here are due to the user page refactor. Some refactor of the user page will be minored. (Use technical DB id in url instead of username that could contain special characters)

PR #9333, opened 2019-01-08
Fix DSN to consider the port number on system information page

nei:

This PR includes the DB Port number on the DSN that is used on the system information page.

### DIFF
--- a/src/Akeneo/Platform/Bundle/AnalyticsBundle/DataCollector/StorageDataCollector.php
+++ b/src/Akeneo/Platform/Bundle/AnalyticsBundle/DataCollector/StorageDataCollector.php
@@ -16,6 +16,9 @@ class StorageDataCollector implements DataCollectorInterface
     /** @var string */
     protected $dbHost;
 
+    /** @var int */
+    protected $dbPort;
+
     /** @var string */
     protected $dbName;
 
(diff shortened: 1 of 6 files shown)


damien-carcel:
Before merging, we need to:
- rebase on master (cannot launch the CI without that)
- add the contribution in the 3.0 changelog
- add the BC-break in the 3.0 changelog

damien-carcel:
Hello @nei.

@BitOne just made me notice that this class was badly designed from the start (and I realized I was the one who made it this way :disappointed:). We create a new `\PDO` connection, but we should just use the doctrine connection that is already existing in the application.

Would you be interested to do it in this PR ? If you don't have the time for, it is no problem we will merge your PR as it is already an improvement :v:, then I'll perform the update afterward.

But if you are willing to do it, it is quite easy: you need to remove all parameters from the constructor to only inject the Doctrine connection instead (the Symfony service is `@database_connection`):
```php
use Doctrine\DBAL\Connection;
…
    /**
     * @param Connection $connection
     */
    public function __construct(Connection $connection)
    {
    }
```
Then use it in the `collect` method to get the server version:
```php
    public function collect()
    {
        return ['mysql_version' => $this->connection->getWrappedConnection()->getServerVersion()];
    }
```

Also, in any case, could you please rebase your PR on the latest version of master, so I can run the CI (mandatory to be able to merge :wink:).

Best regards,
Damien.

nei:
@damien-carcel cool no worries. I will do it tonight. thanks for the suggestion.

nei:
@damien-carcel done.

src/Akeneo/Platform/Bundle/AnalyticsBundle/DataCollector/StorageDataCollector.php
```diff
@@ -1,8 +1,9 @@
 <?php
 
-namespace Akeneo\Platform\Bundle\AnalyticsBundle\DataCollector;
+namespace Pim\Bundle\AnalyticsBundle\DataCollector;
```
damien-carcel: Oups, my bad, I copy/past you an error, this namespace does not exist anymore (I did the test on 2.3, this is master). Please revert the namespace change.

damien-carcel:
@nei Don't bother, I'll do the modification myself, and update the changelog in the process, in a dedicated commit :slightly_smiling_face:

damien-carcel:
One behat failing, know unstable, pass locally. Also, one fail on front stylelint, already fixed in another PR.
Merging.

PR #9341, opened 2019-01-09
Remove useless unstable tests of thepatch of channels

ahocquard:
This endpoint should not exist.
It is to patch in mass the channel... 

One of the test is unstable because a background process is triggered in the previous (remove completeness of channel and locale).
There is a conflict with the previous test, because it REPLACE a line after deletion of the whole database.
Then, when load the dump, there is a primary key error because the line already exist.

It's very tricky to fix for no value at all.

`OK`: Done / Validated
`-`: Not needed


grena:
> a background process is triggered in the previous
Could it be a problem for another test than the one you delete?

PR #9343, opened 2019-01-09
PIM-7889: Update ES indexes on category deletion

phaseinducer:

When a category is deleted the products and product models under it still contains the category in ES, so if it was the last category of the current tree they were linked to, you can never find them again because they won't appear in "unclassified". Same issue for the exports.

This PR fixes that by updating the indexes after the deletion.
For more details about the implementation, see the description of the subscriber.
We chose to call the ES API directly and not do a full save + reindex products because it's way faster and save us from adding a new job (very costly in a patch).

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Pim/Bundle/CatalogBundle/EventSubscriber/Category/UpdateIndexesOnCategoryDeletion.php
+++ b/src/Pim/Bundle/CatalogBundle/EventSubscriber/Category/UpdateIndexesOnCategoryDeletion.php
@@ -1,0 +1,103 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Pim\Bundle\CatalogBundle\EventSubscriber\Category;
+
+use Akeneo\Bundle\ElasticsearchBundle\Client;
+use Akeneo\Component\Classification\Model\CategoryInterface;
+use Akeneo\Component\StorageUtils\StorageEvents;
+use Pim\Component\Catalog\Category\GetDescendentCategoryCodes;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\EventDispatcher\GenericEvent;
+
+/**
+ * Subscribes to category deletion events and updates all ES indexes accordingly.
+ *
+ * This subscriber has to be stateful for consistency : we need to get all descendent codes before the category is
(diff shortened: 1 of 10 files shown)


src/Pim/Bundle/CatalogBundle/EventSubscriber/Category/UpdateIndexesOnCategoryDeletion.php
```diff
@@ -0,0 +1,103 @@
...
+use Symfony\Component\EventDispatcher\GenericEvent;
+
+/**
+ * Subscribes to category deletion events and updates all ES indexes accordingly.
+ *
+ * This subscriber has to be stateful for consistency : we need to get all descendent codes before the category is
+ * effectively removed (and the nested set is reindexed), but we also need to not perform the ES indexes update if the
+ * removal fails (because of a database constraint for example).
+ *
+ * Ideally this should be done by working more with domain events and aggregates, but as we don't have this for now and
```
ahocquard: :+1:

src/Pim/Component/Catalog/tests/integration/Category/GetDescendentCategoryCodesIntegration.php
```diff
@@ -0,0 +1,80 @@
...
+            ]
+        );
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function getConfiguration()
+    {
+        return $this->catalog->useFunctionalCatalog('catalog_modeling');
```
ahocquard: mhhhh, do we really need to use this catalog with 240 product models for that?

Maybe you can do the same as I did:
https://github.com/akeneo/pim-community-dev/blob/master/tests/back/Pim/Enrichment/Integration/Storage/ElasticsearchAndSql/CategoryTree/ListChildrenCategoriesWithCountIncludingSubCategoriesIntegration.php

phaseinducer: Nice ! I didn't know it existed :)

phaseinducer:
`features/mass-action/edit_common_attributes.feature:299` fails on the CI but passes locally.

PR #9375, opened 2019-01-15
PIM-7965: do not call the detacher anymore

rouliane:
This fix basically revert the fix done in this PR in the 2.0 branch : https://github.com/akeneo/pim-community-dev/pull/7368.

When a family is saved, a subscriber catch the save event and detach all the family variants, but it has a side effect on the API and generates this error :
```
request.CRITICAL: Uncaught PHP Exception Doctrine\ORM\ORMInvalidArgumentException: "A new entity was found through the relationship 'Pim\Bundle\CatalogBundle\Entity\Family#familyVariants' that was not configured to cascade persist operations for entity: Pim\Component\Catalog\Model\FamilyVariant@0000000053b43252000000001bcb5ec2. To solve this issue: Either explicitly call EntityManager#persist() on this unknown entity or configure cascade persist  this association in the mapping for example @ManyToOne(..,cascade={"persist"}). If you cannot find out which entity causes the problem implement 'Pim\Component\Catalog\Model\FamilyVariant#__toString()' to get a clue." at /home/akeneo/pim/vendor/doctrine/orm/lib/Doctrine/ORM/ORMInvalidArgumentException.php line 92 
```
With this PR, the error the PR in the 2.0 fixed cannot be reproduced in the 2.3 branch so the detacher can be safely removed.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/CHANGELOG-2.3.md
+++ b/CHANGELOG-2.3.md
@@ -1,5 +1,9 @@
 # 2.3.x
 
+## Bug fixes
+
+- PIM-7965: fix families patch endpoint when updating a family with a family variant 
+
 # 2.3.24 (2019-01-10)
  
 ## Bug fixes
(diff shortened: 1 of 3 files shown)


rouliane:
some behat and integration tests are failing on the CI, but they all pass locally, except 2 EE behats that are already failing during the nightly builds. I'll launch an empty build to be able to merge this PR.

PR #9378, opened 2019-01-16
PIM-7962: use baseRemover to delete user to do not use doctrine events

skeleton:

`OK`: Done / Validated
`-`: Not needed


jmleroux:
While using `remover` is fine, i don't understand how it would avoid doctrine events, because it's doing the same `remove` and `flush`

skeleton:
because our removers have their own events StorageEvents::PRE_REMOVE, etc.
This way, I can create a subscriber in my PR in EE instead of a listener. And why I don't want to create a listener is because I want to inject a query which is not allowed by symfony due to a ref circular (you listen a doctrine event, then you inject a doctrine service :/ )

PR #9383, opened 2019-01-16
Add Webmozart assert library dependency

jmleroux:
We use this library [here](https://github.com/akeneo/pim-community-dev/blob/master/src/Oro/Bundle/PimDataGridBundle/Normalizer/ProductAndProductModelRowNormalizer.php#L10) so we need it as a prod dependency.

**Note**: we didn't see the issue on CI because it is also an indirect dev dependency linked to phpspec.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -76,7 +76,8 @@
         "symfony/polyfill-apcu": "1.4.0",
         "twig/extensions": "1.3.0",
         "twig/twig": "1.36.0",
-        "box/spout": "2.7.2"
+        "box/spout": "2.7.2",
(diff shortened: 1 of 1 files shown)


fitn:
Why don't we change the code by instanceof like we are doing everywhere?

jmleroux:
> Why don't we change the code by instanceof like we are doing everywhere?

We are using a lot this assert library in EE for ReferenceEntities and we found it very usefull.
Anyway, we have to freeze to prepare the release and i'm just fixing dependencies, not code that was GTMed before :wink:

PR #9385, opened 2019-01-17
Remove end to end tests to add acceptance tests

skeleton:

Remove end to end tests on validation for resources: association type, family, group type & attribute.
To be honest, I expected to remove much more than 9 end to end tests but except for product, we don't test our validation :/ :/ 

The work is not finish for these resources because I'm facing the same problem as Arnaud about immutability of resources but I will make another PR about that.

`OK`: Done / Validated
`-`: Not needed


tests/back/Acceptance/AssociationType/Context/AssociationTypeValidation.php
```diff
@@ -0,0 +1,78 @@
...
+        $this->associationTypeBuilder = $associationTypeBuilder;
+        $this->validator = $validator;
+    }
+
+    /**
+     * @When I create an association type with a code :code
+     */
+    public function iCreateAnAssociationTypeWithCode(string $code)
+    {
+        $this->associationType = $this->associationTypeBuilder->build(['code' => $code], false);
```
arnolanglade: Weird, it sounds this class build valid entity https://github.com/akeneo/pim-community-dev/blob/2.3/tests/back/Common/EntityBuilder.php#L65
I think we should not use this class to build entity. We should create a dedicated builder for each entity because we should use "business oriented" builder instead of generic builder.

tests/back/Acceptance/Attribute/Context/AttributeValidation.php
```diff
@@ -0,0 +1,169 @@
...
+        $this->attributeSaver = $attributeSaver;
+        $this->validator = $validator;
+    }
+
+    /**
+     * @When I create an attribute with a code :code
+     */
+    public function iCreateAnAttributeWithACode(string $code)
+    {
+        $this->attribute = $this->attributeBuilder->build(['code' => $code], false);
```
arnolanglade: There is a specific builder for attribute: thttps://github.com/akeneo/pim-community-dev/blob/2.3/tests/back/Common/EntityBuilder.php#L65

BitOne:
Unfortunately too old to be mergeable.

PR #9423, opened 2019-01-25
Fix ES config loader

jmleroux:
ES config loader is not correct when merging mappings configuration (from extensions for example):
- properties must be replaced to allow override
- dynamic_templates must be appended to allow adding attribute types

It fixes this kind of error: https://github.com/akeneo/ExtendedAttributeTypeBundle/issues/64

And I already modified this loader, fixing properties merging and screwing dynamic_templates merging in #8807 . :sob: 
This time, i take care of the two use cases.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -231,9 +231,9 @@ workflows:
           - back_phpunit_end_to_end:
                 requires:
                     - back_phpunit_integration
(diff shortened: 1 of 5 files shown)


jjanvier:
it misses a spec JM ;)

PR #9447, opened 2019-01-29
PIM-8017: Fix PDF generation

Doodoune:
**Description**

In the PEF, there is a 500 internal error when trying to load the PDF.
 EE PR : https://github.com/akeneo/pim-enterprise-dev/pull/5613


Doodoune:
For 3.0.1

PR #9480, opened 2019-01-30
Stabilize the phpunit on 2.0

ahocquard:

`OK`: Done / Validated
`-`: Not needed


src/Pim/Bundle/CatalogBundle/tests/integration/PQB/ProductAndProductModelQueryBuilderIntegration.php
```diff
@@ -445,6 +453,8 @@ public function testStatusOnProductWithoutParent()
                 '1111111166',
                 '1111111167',
                 '1111111168',
+                '1111111169',
+                '1111111170',
```
ahocquard: it should be return actually, so even if it's skipped I prefer to fix was it is really excepted

PR #9490, opened 2019-01-31
PIM-8022: Fix the job status when using the batch command

damien-carcel:
**Description**

When running `akeneo:batch:job`, the status of the job on the job process page is always "Failed".
This is caused by the JobExecutionManager that is checking the health of the job. However, when launching the job with the batch command, the health check property stays to `null`, and the manager changes the status of the job to "failed".

This PR ensures that the health check property is filled before the process is started when using the job queue consumer. That way, a health check can be null only if the job is launched through the batch command, and then the manager keeps the current job execution status.

This PR also deprecates the batch command, as we should always use the queue and the publish command.

**Definition Of Done**

`OK`: Done / Validated
`-`: Not needed


src/Pim/Bundle/ImportExportBundle/Resources/config/routing/import_execution.yml
```diff
@@ -9,7 +9,6 @@ pim_importexport_import_execution_download_log:
 
 pim_importexport_import_execution_show:
     path: /{id}
-    defaults: { _controller: pim_import_export.controller.import_execution:showAction }
```
damien-carcel: useless, this is a front-end route

CHANGELOG-2.3.md
```diff
@@ -1,5 +1,7 @@
 # 2.3.x
 
+- PIM-8022: Fix the job status when using the batch command.
```
Doodoune: Add a ## Bug fixes

PR #9540, opened 2019-02-07
PIM-7939: Search on ancestors labels and codes

damien-carcel:
### The problem

When searching on "label or identifier" in the datagrid, and selecting all the results for a mass-action, the filters used are passed as they are to the batch job that will perform the mass action.

We then face an issue with product models and variant products, as the datagrid PQB returns only the elements of one level: those containing the value you are searching for.

But in the mass edit, we need to return both the element containing the value and its children, as the effect of the mass action could be performed on them (for instance, the search returns a product model, but the action is to activate products, so it will be performed on the variant products, not on their parents).

For instance, consider that the following example, where the label is on the sub product model:

| Data we search on | Root product model | Sub product model   | Variant product |
|-------------------|--------------------|---------------------|-----------------|
| code              | root               | sub                 | product         |
| label             |                    | A label not at root |                 |

- Search for "root":
    - datagrid will return the root product model (even if it is present in the label of the sub product model, it is also  the code of the root product model, so the highest level is returned)
    - mass edit should return everything
- Search for "label":
    - datagrid will return the sub product model
    - mass edit should return it along with all its children

### Solution

We already faced the same issue with IDs in the past (when selecting product models one by one or all visible) and introduced the "SelfAndAncestorFilter".

The same thing is done for "label or identifier" in this PR. This also implied indexing those new values in ES.

## Definition Of Done

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/CHANGELOG-2.3.md
+++ b/CHANGELOG-2.3.md
@@ -1,5 +1,10 @@
 # 2.3.x
 
+## Bug fixes
+
+- PIM-7939: Fix PQB search when an attribute as label is on an ancestor.
+  -> Not mandatory, you can re-index your products and product models to enjoy this fix with commands: `bin/console pim:product:index --all` and `bin/console pim:product-model:index --all`.
+
 # 2.3.29 (2019-02-11)
 
 ## Bug fixes
--- a/src/Pim/Component/Catalog/Normalizer/Indexing/ProductAndProductModel/ProductModelPropertiesNormalizer.php
+++ b/src/Pim/Component/Catalog/Normalizer/Indexing/ProductAndProductModel/ProductModelPropertiesNormalizer.php
@@ -7,6 +7,8 @@ namespace Pim\Component\Catalog\Normalizer\Indexing\ProductAndProductModel;
 use Pim\Component\Catalog\Model\ProductModelInterface;
 use Pim\Component\Catalog\Normalizer\Standard\Product\PropertiesNormalizer as StandardPropertiesNormalizer;
 use Pim\Component\Catalog\ProductAndProductModel\Query\CompleteFilterInterface;
+use Pim\Component\Catalog\Repository\ChannelRepositoryInterface;
+use Pim\Component\Catalog\Repository\LocaleRepositoryInterface;
 use Symfony\Component\Serializer\Normalizer\NormalizerInterface;
 use Symfony\Component\Serializer\SerializerAwareInterface;
 use Symfony\Component\Serializer\SerializerAwareTrait;
@@ -34,9 +36,23 @@ class ProductModelPropertiesNormalizer implements NormalizerInterface, Serialize
     /** @var CompleteFilterInterface */
     private $completenessGridFilterQuery;
 
-    public function __construct(CompleteFilterInterface $completenessGridFilterQuery)
-    {
+    /** @var ChannelRepositoryInterface */
+    private $channelRepository;
+
+    /** @var LocaleRepositoryInterface */
+    private $localeRepository;
+
+    /**
+     * @todo merge master: remove "= null"
+     */
+    public function __construct(
+        CompleteFilterInterface $completenessGridFilterQuery,
+        ChannelRepositoryInterface $channelRepository = null,
+        LocaleRepositoryInterface $localeRepository = null
+    ) {
         $this->completenessGridFilterQuery = $completenessGridFilterQuery;
+        $this->channelRepository = $channelRepository;
+        $this->localeRepository = $localeRepository;
     }
 
     /**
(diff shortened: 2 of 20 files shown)


ahocquard:
Before reading it: is it validated by PO?

damien-carcel:
@ahocquard 
```
* the categorization of a product model will be applied also on children products
```
No because this mass edit is not applied on children so the original filter will not be replaced with the "ancestors" one (same behavior that for the IDs).

For the associations, it is to be checked, but behavior should be the same.

CHANGELOG-2.3.md
```diff
@@ -10,6 +10,8 @@
 - PIM-8008: Fix attributes sort order in PEF.
 - PIM-8022: Fix the job status when using the batch command.
 - PIM-8050: Fix ElasticSearch mappings loader
+- PIM-7939: Fix PQB search when an attribute as label is on an ancestor.
```
damien-carcel: To move as 2.3.29 was tagged since the PR was open.

src/Pim/Component/Catalog/Normalizer/Indexing/ProductAndProductModel/ProductModelPropertiesNormalizer.php
```diff
@@ -34,9 +36,20 @@ class ProductModelPropertiesNormalizer implements NormalizerInterface, Serialize
...
-    {
+    /** @var ChannelRepositoryInterface */
+    private $channelRepository;
+
+    /** @var LocaleRepositoryInterface */
+    private $localeRepository;
+
+    public function __construct(
+        CompleteFilterInterface $completenessGridFilterQuery,
+        ChannelRepositoryInterface $channelRepository,
```
damien-carcel: Add `= null` to avoid BC break, and add a `@todo merge master: remove "= null"` in the PHPDoc

src/Pim/Component/Catalog/Normalizer/Indexing/ProductAndProductModel/ProductModelPropertiesNormalizer.php
```diff
@@ -192,4 +207,112 @@ private function getAncestorsCodes(ProductModelInterface $productModel): array
...
+     * @param ProductModelInterface $productModel
+     * @param string $attributeCodeAsLabel
+     *
+     * @return array
+     */
+    private function getLocalizableAndScopableLabels(
+        ProductModelInterface $productModel,
+        string $attributeCodeAsLabel
+    ): array {
+        $ancestorsLabels = [];
```
damien-carcel: Your repositories could be null in a project (once the constructor is fixed regarding my previous comment). If they are, you cannot get the label, so I think you should return an empty array. To validate with a PM.

Same applies to the two other methods below.

src/Pim/Component/Catalog/Normalizer/Indexing/ProductAndProductModel/ProductPropertiesNormalizer.php
```diff
@@ -32,6 +34,20 @@ class ProductPropertiesNormalizer implements NormalizerInterface, SerializerAwar
...
     private const FIELD_CATEGORIES_OF_ANCESTORS = 'categories_of_ancestors';
 
+    /** @var ChannelRepositoryInterface */
+    private $channelRepository;
+
+    /** @var LocaleRepositoryInterface */
+    private $localeRepository;
+
+    public function __construct(
+        ChannelRepositoryInterface $channelRepository,
```
damien-carcel: Same comments than for the `ProductModelPropertiesNormalizer`

src/Pim/Component/Catalog/Normalizer/Indexing/ProductAndProductModel/ProductModelPropertiesNormalizer.php
```diff
@@ -192,4 +210,119 @@ private function getAncestorsCodes(ProductModelInterface $productModel): array
...
+            return [];
+        }
+
+        $attributeAsLabel = $family->getAttributeAsLabel();
+        if (null === $attributeAsLabel) {
+            return [];
+        }
+
+        $hasRepositories = false;
+        if (null !== $this->channelRepository && null !== $this->localeRepository) {
```
anaelChardan: or simply `$hasRepositories = null !== $this->channelRepository && null !== $this->localeRepository`

src/Pim/Component/Catalog/Normalizer/Indexing/ProductAndProductModel/ProductModelPropertiesNormalizer.php
```diff
@@ -192,4 +210,119 @@ private function getAncestorsCodes(ProductModelInterface $productModel): array
...
+            $hasRepositories = true;
+        }
+
+        $ancestorsLabels = [];
+        $attributeCodeAsLabel = $attributeAsLabel->getCode();
+        switch (true) {
```
anaelChardan: ?

CHANGELOG-2.3.md
```diff
@@ -1,5 +1,10 @@
 # 2.3.x
 
+## Bug fixes
+
+- PIM-7939: Fix PQB search when an attribute as label is on an ancestor.
+  -> Please, re-index your products and product models with commands: `bin/console pim:product:index --all` and `bin/console pim:product-model:index --all`.
```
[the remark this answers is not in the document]
willy-ahva: Seen IRL, the reindexation is not mandatory.

PR #9558, opened 2019-02-11
Upgrade Doctrine ORM to latest patch

jjanvier:
As stated [here](https://github.com/doctrine/orm/releases/tag/v2.6.0), 2.5.14 is and will be the last patch of Doctrine ORM 2.5.

That would be cool if our 3.0 LTS could benefit of it.

Wait for https://github.com/akeneo/pim-community-dev/pull/9528 to be merged (To not conflict with the `composer.lock`)


ahocquard:
My PR is merged btw and that's why you have to regenerate the composer.lock

PR #9561, opened 2019-02-11
Improve makefile

arnolanglade:
### How Makefile works
A simple makefile consists of "rules" with the following shape:

```
target ... : prerequisites ...
        command
        ...
        ...
```

A **target** is usually the name of a file that is generated by a program; examples of targets are executable or object files. A target can also be the name of an action to carry out, such as `clean' (see section Phony Targets).

A **prerequisite** is a file that is used as input to create the target. A target often depends on several files.

A **command** is an action that make carries out. A rule may have more than one command, each on its own line. Please note: you need to put a tab character at the beginning of every command line! This is an obscurity that catches the unwary. 

Extra information [here](https://ftp.gnu.org/old-gnu/Manuals/make-3.79.1/html_chapter/make_2.html)

### What did I need to do have working PIM before the Makefile?
```bash
$ docker-compose up
$ bin/docker/pim-dependencies.sh
$ bin/docker/pim-setup.sh
$ bin/docker/pim-initialize.sh
```

### What do I need to do have working PIM after the Makefile 
```
make up
make init-pim
```

Makefile is more intelligent then a our script. For instance the target `up` depends on the target `vendor`. Make will install backend dependencies automatically if this directory does not exist. That means when I run this command I sure that my pim will be well setup because this target depend on every thing it needs.

We only need to run 2 commands. The target `up` may depend on `init-pim` but it means every time you run `make up` the database and run webpack.

### Note
I don't delete docker shell scripts so far. Moreover this makefile are the bare minimum to run a PIM, we need to improve it. Makefile will automatically include all file in make-file/*.mk. For now, I have added a file per bounded context with a single target (coupling rule). Each team should add its custom targets or you can add you custom targets. **Caution:** those targets are **optimized for docker**

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,7 @@ help:
 	@echo ""
 	@echo "Caution: those targets are optimized for docker"
 	@echo ""
-	@echo "Please add your custom Makefile in the dicretory "make-file". They will be automotically loaded!"
+	@echo "Please add your custom Makefile in the directory "make-file". They will be automatically loaded!"
 	@echo ""
 
 ## Include all *.mk files
--- a/Makefile
+++ b/Makefile
@@ -1,37 +1,90 @@
-DOCKER ?= false
-PHP_EXEC :=php
-
-ifeq ($(DOCKER),true)
-	PHP_EXEC=docker-compose exec -u docker fpm php
-endif
+DOCKER_COMPOSE = docker-compose
+PHP_EXEC=$(DOCKER_COMPOSE) exec -u docker fpm php
+YARN_EXEC=$(DOCKER_COMPOSE) run --rm node yarn
 
 .DEFAULT_GOAL := help
 
 .PHONY: help
 help:
 	@echo ""
-	@echo "Akeneo Pim Community Dev available targets:"
+	@echo "Caution: those targets are optimized for docker"
 	@echo ""
-	@echo "If you want to use docker, use an environment variable DOCKER=true"
+	@echo "Please add your custom Makefile in the dicretory "make-file". They will be automotically loaded!"
 	@echo ""
-	@grep -E '^.PHONY:.*##.*' $(MAKEFILE_LIST) | cut -c9- | sort | awk 'BEGIN {FS = " ## "}; {printf "%-30s %s\n", $$1, $$2}'
 
-### Coupling detection
-.PHONY: structure-coupling ## Run the coupling-detector on Structure
-structure-coupling:
-	$(PHP_EXEC) vendor/bin/php-coupling-detector detect --config-file=src/Akeneo/Pim/Structure/.php_cd.php src/Akeneo/Pim/Structure
+## Include all *.mk files
+include make-file/*.mk
 
-.PHONY: user-management-coupling ## Run the coupling-detector on User Management
-user-management-coupling:
-	$(PHP_EXEC) vendor/bin/php-coupling-detector detect --config-file=src/Akeneo/UserManagement/.php_cd.php src/Akeneo/UserManagement
+##
+## PIM initialization
+##
 
-.PHONY: channel-coupling ## Run the coupling-detector on Channel
-channel-coupling:
-	$(PHP_EXEC) vendor/bin/php-coupling-detector detect --config-file=src/Akeneo/Channel/.php_cd.php src/Akeneo/Channel
+composer.lock: composer.json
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -35,3 +38,66 @@ enrichment-coupling:
...
+
+node_modules: package.json
+	$(YARN_EXEC) install
+
+behat.yml:
+	cp ./behat.yml.dist ./behat.yml
+	sed -i "s/127.0.0.1\//httpd-behat\//g" ./behat.yml
+	sed -i "s/127.0.0.1/selenium/g" ./behat.yml
+
+app/config/parameters.yml:
```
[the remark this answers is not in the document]
arnolanglade: This is another way to do it but the best way is to use env vars. For now, I have used the command from the pim-setup.sh

Makefile
```diff
@@ -1,8 +1,11 @@
 DOCKER ?= false
-PHP_EXEC :=php
+PHP_EXEC = php
+YARN_EXEC = yarn
 
 ifeq ($(DOCKER),true)
+	DOCKER_COMPOSE = docker-compose
```
anaelChardan: Should be outside of the if as it is an invariant

Makefile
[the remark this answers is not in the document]
anaelChardan: This target works only for docker, am I right ?

Makefile
```diff
@@ -35,3 +38,66 @@ enrichment-coupling:
...
+	sed -i "s/127.0.0.1\//httpd-behat\//g" ./behat.yml
+	sed -i "s/127.0.0.1/selenium/g" ./behat.yml
+
+app/config/parameters.yml:
+	cp ./app/config/parameters.yml.dist ./app/config/parameters.yml
+	# Sed commands should be removed when env var will be introduce in the PIM
+	sed -i "s/database_host:.*localhost/database_host: mysql/g" ./app/config/parameters.yml
+	sed -i "s/localhost: 9200/elastic:changeme@elasticsearch:9200/g" ./app/config/parameters.yml
+
+app/config/parameters_test.yml:
```
anaelChardan: Same here

Makefile
```diff
@@ -35,3 +38,66 @@ enrichment-coupling:
...
+	sed -i "s/product_and_product_model_index_name:.*akeneo_pim_product_and_product_model/product_and_product_model_index_name:  test_akeneo_pim_product_and_product_model/g" ./app/config/parameters_test.yml
+	sed -i "s/record_index_name:.*akeneo_referenceentity_record/record_index_name:                     test_akeneo_referenceentity_record/g" ./app/config/parameters_test.yml
+	sed -i "s/product_proposal_index_name:.*akeneo_pim_product_proposal/product_proposal_index_name:           test_akeneo_pim_product_proposal/g" ./app/config/parameters_test.yml
+	sed -i "s/published_product_index_name:.*akeneo_pim_published_product/published_product_index_name:          test_akeneo_pim_published_product/g" ./app/config/parameters_test.yml
+	sed -i "s/published_product_and_product_model_index_name:.*akeneo_pim_published_product_and_product_model/published_product_and_product_model_index_name: test_akeneo_pim_published_product_and_product_model/g" ./app/config/parameters_test.yml
+
+docker-compose.override.yml:
+	cp docker-compose.override.yml.dist docker-compose.override.yml
```
anaelChardan: I would add the docker assertion here

Makefile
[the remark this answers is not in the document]
arnolanglade: yes, we may define some var make it more flexible.

Makefile
arnolanglade: this target only depend on `up`, we don't need to add an assertion. We need to keep it simple.

Makefile
```diff
@@ -1,37 +1,90 @@
...
 .DEFAULT_GOAL := help
 
 .PHONY: help
 help:
 	@echo ""
-	@echo "Akeneo Pim Community Dev available targets:"
+	@echo "Caution: those targets are optimized for docker"
 	@echo ""
-	@echo "If you want to use docker, use an environment variable DOCKER=true"
+	@echo "Please add your custom Makefile in the dicretory "make-file". They will be automotically loaded!"
```
anaelChardan:
```suggestion
	@echo "Please add your custom Makefile in the directory "make-file". They will be automatically loaded!"
```

PR #9566, opened 2019-02-12
Update dependencies

arnolanglade:

`OK`: Done / Validated
`-`: Not needed


composer.json
```diff
@@ -63,21 +63,19 @@
         "monolog/monolog": "1.23.0",
         "ocramius/proxy-manager": "2.1.1",
         "oneup/flysystem-bundle": "1.14.0",
-        "ramsey/uuid": "3.7.0",
-        "sensio/distribution-bundle": "5.0.20",
+        "ramsey/uuid": "^3.8",
```
jjanvier: not sure it's a good idea to use `^`
as we don't know use `composer.lock` for standard editions (same for ee dev), that means we don't know exactly which dependencies the customers will install...

ok, it's already the case now, as we don't precise exactly all dependencies (or dependencies of dependencies) in the `composer.json`

don't hear me wrong, what you did is exactly what I would do if we used properly `composer.lock`

PR #9569, opened 2019-02-13
Fix PhpUnit tests

jjanvier:
I don't know what happened, but in EE the build fails because of this.
Was it because of an upgrade of some library?


ahocquard:
It will not work. CE has a composer.lock, that's why it is working.
But here, you will fix EE (without composer.lock) and not CE (because it uses a dependency of the composer.lock).

Do you know which dependency is responsible of this problem?

PR #9571, opened 2019-02-13
Fix Build 3.0 by setting doctrine deps

jjanvier:
As those 2 dependencies were not explicitly set, and as we do not use a composer.lock in EE, the build EE was requiring other versions of those 2 libraries, which caused the build to fail.


ahocquard:
I don't understand this fix. We are going into the wrong direction in my opinion.

This is a subdependency, never used in our project.
If it breaks our tests, it's because we assert on JSON string instead of using a good assertion for JSON.

Having spaces in the JSON is not considered as a BC break by the library. And it is not.
Now, if this library is not used anymore, we will depend on it.


We can depend of all sub-dependencies with this kind of fix.

arnolanglade:
You may use this lib https://github.com/coduo/php-matcher to check our json response.

jjanvier:
@ahocquard we don't have the same version of `doctrine/dbal` in CE and EE. Don't you think it's a problem? 

> This is a subdependency, never used in our project.

Yes, but it's used by one the our doctrine dep. So yes, it's used by us.

> Having spaces in the JSON is not considered as a BC break by the library. And it is not.

Yes, ideally, we should change the tests. We should not care about the spaces in the tests, I agree.

Again, the main problem for me is that today, we don't have the same version of `doctrine/dbal` in CE and EE.

jjanvier:
To explain a bit more what happens:
- `doctrine/dbal` is required by `doctrine/orm`
- `doctrine/dbal` is not one of our requirement (not in the `require` section). This is just a sub dependency. 
- On CE, we have a `composer.lock`. So we use the `doctrine/dbal` version given by the lock: 2.5.13
- On EE, we don't have a `composer.lock`. So we use the [latest](https://github.com/doctrine/orm/blob/2.5/composer.json#L19) `doctrine/dbal` version compatible with our locked `doctrine/orm` version. So we get `doctrine/dbal` 2.6.3

PR #9574, opened 2019-02-13
GITHUB-9451: Use PHPUnit to list tests

anaelChardan:

We now decided to use PHPUnit to generate the tests we want to execute because by using "*Integration.php" it is very unconvenient, you can't have abstract that is named the same etc... And, moreover what is better than PHPUnit to list PHPUnit tests to launch? :P

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/.circleci/find_phpunit.sh
+++ b/.circleci/find_phpunit.sh
@@ -1,0 +1,24 @@
+#!/bin/sh
+
+set -eu
+
+if [[ -z "$1}" ]]; then
+    echo "Provide the argument SUITES you want to list-tests as first argument"
+fi
+
+SUITES=$@
+
+# Here is the output of phpunit
+
+# it gives a header + The namespaces of the test and the test so as we launch by file we have to work a little bit on it
+
+#PHPUnit 6.5.3 by Sebastian Bergmann and contributors.
+#
+#Available test(s):
+# - Akeneo\Bundle\RuleEngineBundle\tests\integration\ExecuteRuleWithoutPermissionsAppliedIntegration::testRuleExecutionOnAllProducts
+# - tests\integration\Akeneo\Bundle\RuleEngineBundle\Normalizer\Standard\RuleIntegration::testRule
+# - PimEnterprise\Bundle\ApiBundle\tests\EndToEnd\Controller\Asset\CreateAssetIntegration::testCreationOfAnAsset
+# - PimEnterprise\Bundle\ApiBundle\tests\EndToEnd\Controller\Asset\CreateAssetIntegration::testResponseWhenContentIsEmpty
+# - PimEnterprise\Bundle\ApiBundle\tests\EndToEnd\Controller\Asset\CreateAssetIntegration::testResponseWhenValidationFailed
+
+vendor/bin/phpunit -c app --testsuite=$SUITES --list-tests | tail -n +4 | cut -c4- | cut -d ':' -f 1 | sort | uniq | sed -e 's/\\/\\\\/g'
--- a/.circleci/find_phpunit.sh
+++ b/.circleci/find_phpunit.sh
@@ -2,7 +2,7 @@
 
 set -eu
 
-if [[ -z "$1}" ]]; then
+if [[ -z "${1}" ]]; then
     echo "Provide the argument SUITES you want to list-tests as first argument"
 fi
 
--- a/.circleci/find_phpunit.sh
+++ b/.circleci/find_phpunit.sh
@@ -22,4 +22,4 @@ SUITES=$@
 # - PimEnterprise\Bundle\ApiBundle\tests\EndToEnd\Controller\Asset\CreateAssetIntegration::testResponseWhenContentIsEmpty
 # - PimEnterprise\Bundle\ApiBundle\tests\EndToEnd\Controller\Asset\CreateAssetIntegration::testResponseWhenValidationFailed
 
-vendor/bin/phpunit -c app --testsuite=$SUITES --list-tests | tail -n +4 | cut -c4- | cut -d ':' -f 1 | sort | uniq | head -n10
+vendor/bin/phpunit -c app --testsuite=$SUITES --list-tests | tail -n +4 | cut -c4- | cut -d ':' -f 1 | sort | uniq | sed -e 's/\\/\\\\/g'
--- a/.circleci/find_phpunit.sh
+++ b/.circleci/find_phpunit.sh
@@ -2,8 +2,9 @@
 
 set -eu
 
-if [[ -z "${1}" ]]; then
+if [ -z "$1" ]; then
     echo "Provide the argument SUITES you want to list-tests as first argument"
+    exit 1
 fi
 
 SUITES=$@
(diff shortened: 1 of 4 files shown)


anaelChardan:
@skeleton yeah it wasn't perfect and sometimes we had the problem that we launched some tests that didn't exists (like abstract one). And there is nothing better than the tool itself to list tests to launch.

skeleton:
ok, so it could be nice to have this explanation in the description of the PR ;)

anaelChardan:
@skeleton you're totally right, I've forgot my copy paste :P :stvcqjvd:

.circleci/find_phpunit.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+
```
jjanvier: explain please
and exit -1 if you don't have the argument you expected

arnolanglade: :+1:

.circleci/run_phpunit.sh
```diff
@@ -6,7 +6,7 @@ TESTFILES=$@
 fail=0
 for TESTFILE in $TESTFILES; do
     echo $TESTFILE
-    docker-compose exec -T fpm ./vendor/bin/phpunit -c app --log-junit var/tests/phpunit/phpunit_$(uuidgen).xml $TESTFILE
+    docker-compose exec -T fpm ./vendor/bin/phpunit -c app --log-junit var/tests/phpunit/phpunit_$(uuidgen).xml --filter $TESTFILE
```
arnolanglade: Why do you need --filter ?

.circleci/find_phpunit.sh
anaelChardan: Good Idea.

.circleci/run_phpunit.sh
anaelChardan: Because phpunit is only able to launch file and not class without filter.

.circleci/find_phpunit.sh
```diff
@@ -0,0 +1,24 @@
...
+#PHPUnit 6.5.3 by Sebastian Bergmann and contributors.
+#
+#Available test(s):
+# - Akeneo\Bundle\RuleEngineBundle\tests\integration\ExecuteRuleWithoutPermissionsAppliedIntegration::testRuleExecutionOnAllProducts
+# - tests\integration\Akeneo\Bundle\RuleEngineBundle\Normalizer\Standard\RuleIntegration::testRule
+# - PimEnterprise\Bundle\ApiBundle\tests\EndToEnd\Controller\Asset\CreateAssetIntegration::testCreationOfAnAsset
+# - PimEnterprise\Bundle\ApiBundle\tests\EndToEnd\Controller\Asset\CreateAssetIntegration::testResponseWhenContentIsEmpty
+# - PimEnterprise\Bundle\ApiBundle\tests\EndToEnd\Controller\Asset\CreateAssetIntegration::testResponseWhenValidationFailed
+
+vendor/bin/phpunit -c app --testsuite=$SUITES --list-tests | tail -n +4 | cut -c4- | cut -d ':' -f 1 | sort | uniq | sed -e 's/\\/\\\\/g'
```
ahocquard: I think `sed -e 's/\\/\\\\/g'` is useless, as you can avoid it by surrounding your variable this way "$TESTFILES"

.circleci/run_phpunit.sh
ahocquard: Phpunit accept a classname. You have just to prefix it with "src".
Btw, I don't know how it works, I tried in local what you did and it doesn't work.

anaelChardan: Nop, if you give a class namespace, you have to give it through filter, it's in the doc ;)

ahocquard:
There is a drawback witht his method: if the classname is wrong (bad namespace), it will not execute any test.
When no test is executed, the return code is 0...

.circleci/run_phpunit.sh
anaelChardan: It can fail on your case depending on the behaviour of your terminal :/. I had this issue once.

anaelChardan:
@ahocquard if the namespace is wrong PHPUnit won't find it, so for me no problem on this point.

ahocquard:
That's the problem: the test is no executed as the namespace is wrong. But the return code is 0 so it skips this test silently...

.circleci/find_phpunit.sh
ahocquard: return code 1, no negative return code in shell

anaelChardan:
The developer is accountable to look at the namespaces and launch it if needed, we get the same problem with tests that don't start with `test` nor annotated by @test. We won't ensure those edge cases imho.

.circleci/find_phpunit.sh
anaelChardan: Every code that is different from 0 is good for returning a bad execution. But yes 1 is more conventional, I update it.

.circleci/run_phpunit.sh
anaelChardan: I may work with certain organisation of file

ahocquard:
I don't agree. What about moving tests? Changing autloloading, etc? You have to execute all your tests one after the other?
We are precisely doing this `list-tests` to improve the reliability of the command and to execute tests not ending by "Integration".

FYI, we already had this problem of tests that are not executed, and it will occur very often with big surprise if we don't handle it properly. It should not fail silently.

anaelChardan:
I don't see how we could prevent that. If the tool don't find a namespace, this is due to a misconfiguration problem and it is the accountability of the person who did this refactoring. How could we say "This namespace is missing". The only thing I can see is to look at the number of tests executed before and after. What would you suggest?

.circleci/find_phpunit.sh
anaelChardan: Phpunit needs double antislash for namespaces :/ : @see https://phpunit.readthedocs.io/en/7.0/textui.html#textui-examples-filter-patterns

.circleci/run_phpunit.sh
ahocquard: :+1:  for the documentation
I have to surround it with single quote both with bash and zsh.
We should do that, as otherwise it means the command works in circle ci but not in local.
Could you test please?

anaelChardan: I've tried but I didn't find a way to add single quotes.

.circleci/find_phpunit.sh
ahocquard: ok, not without `--filter` for your information ( I tested). Weird behavior

anaelChardan:
So, according to some tests using wrong namespaces PHPUnit does find them so the problem doesn't exist. We should have a script on the CI to find all bad namespaces.

A script like that can do that:

```php
<?php

$loader = require __DIR__.'/../app/autoload.php';

function phpUnitNamespaces() {
    $outputs = [];
    exec("vendor/bin/phpunit -c app --testsuite=PIM_Integration_Test --list-tests | tail -n +4 | cut -c4- | cut -d ':' -f 1 | sort | uniq", $outputs);

    $namespaces = [];
    foreach ($outputs as $output) {
        $namespaces[] = [
            'namespace' => $output,
            'file' => (new \ReflectionClass($output))->getFileName()
        ];
    }
    return $namespaces;
}



var_dump(phpUnitNamespaces());
```

.circleci/find_phpunit.sh
anaelChardan: Yes, it's weird, their not consistent

PR #9583, opened 2019-02-15
Clean stuff

arnolanglade:

`phpdoc.xml.dist` file is used anymore so I deleted them.
The cache generated by php-cs-fixer has been moved to the folder var like symfony cache or application logs. 

`OK`: Done / Validated
`-`: Not needed


.gitignore
```diff
@@ -44,7 +44,7 @@ behat.yml
 /phpspec.yml
 uploads_test/
 .phpspec/*.tpl
-.php_cs.cache
+var/php_cs.cache
```
damien-carcel: Shouldn't we rather ignore the whole `var` directory, except the one file we're only keeping? Like what is done for the `docker` directory.

grena: Yup I agree

.gitignore
```diff
@@ -13,10 +13,9 @@ web/app_local.php
 app/AppKernel_local.php
 app/console_local
 app/archive/*
-var/cache/*
+var/*
+!var/SymfonyRequirements.php
```
damien-carcel: :+1:

PR #9584, opened 2019-02-15
Remove old changelog.md and update.md

arnolanglade:

We don't need old changelog and ugrade file (legally speaking). To avoid visual pollution I propose to remove them and keep only the changelog and the upgrade instructions for the current branch.

`OK`: Done / Validated
`-`: Not needed


damien-carcel:
I don't disagree, but there is the issue of the bugs fixed by the maintenance on older branches (let say 2.3) that are fixed in this branch when merging 2.3 into 3.0 then into master.
How do we managed this aspect of the changelog? Should it appear on master, directly under the future 3.1.0 tag?

arnolanglade:
Each branch should owns its changelog. If you are using/working on 3.0 you don't need to know if a bug was fixed in 2.3 because the branch should be up to date. I have worked since 3 years I never look at a changelog from a previous version.

damien-carcel:
OK, let's take a concrete example that had already happen in maintenance. You fixed a bug in 2.0, that need integrators to create a new batch job in their PIM (this should not happen, but it did, several times). This info is in the 2.0 changelog.
You then merge in 2.3. This bug fixed in 2.0 was also present in 2.3, so merging 2.0 into 2.3 fixes it.
If you don't have the 2.0 changelog inside the 2.3 release, there is nothing telling you to add the new job.

As I said, I don't disagree with removing other changelogs, but we may need some data from them. How do we manage such a case?

arnolanglade:
This is the **dev** community I did not remove the changelog from the standard edition. By the way, an integrator work with a standard edition **not** a dev edition.

ahocquard:
@damien-carcel that's why I proposed to keep only the maintained branches.
It's done this way for the phpunit project, it's pretty clear and you have 3 changelogs max.

PR #9585, opened 2019-02-15
[POC] Reorganize the symfony enviroment

arnolanglade:

Currently, we have the following Symfony environment:  
* **prod:** we don't touch it.
* **dev:** we don't touch it.
* **behat:** it is used by behat and phpunit for tests which need a database. I propose to rename it to test_database because it is clearer!
* **test:** I propose to remove it from the pim because it is not used.
* **test_fake:** we don't touch it.

`OK`: Done / Validated
`-`: Not needed


app/config/config.yml
```diff
@@ -64,6 +64,18 @@ akeneo_pim_structure:
         color:
             class: Acme\Bundle\AppBundle\Entity\Color
             type: simple
+monolog:
```
jjanvier: why put that in config.yml?

jjanvier: is it normal you don't have the same change in CE?

app/config/config_test_fake.yml
```diff
@@ -8,9 +9,21 @@ framework:
...
 swiftmailer:
     disable_delivery: true
+
+pim_import_export:
+    record_mails: true
+
+parameters:
+    max_products_category_removal: 20
+    installer_data:       PimInstallerBundle:minimal
+    upload_dir:           '%kernel.root_dir%/../var/cache/uploads/product'
```
jjanvier: why do you change those parameters?
all the dir parameters already have the value you give them
or maybe I miss something oO

skeleton:
https://github.com/akeneo/pim-community-dev/blob/f0844929edadcfe4e6aee8c0723f7788d9c7a7fc/tests/back/Integration/TestCase.php#L37
change it too (and I think there is other AppKernelTest in out tests that hard code the env)

app/config/config_test_fake.yml
arnolanglade: All specfic config files "extend" now `config.yml`so you needed to copy/paste some of configuration from `config_behat` to this file. Each env define its own configuration.

app/config/config.yml
arnolanglade: Do you mean EE? the configuration is the same in EE because of EE feature.

app/config/config_test_fake.yml
arnolanglade: Those parameters are overridden for both tests env.

app/config/config_test_database.yml
```diff
@@ -1,25 +1,24 @@
...
 
-web_profiler:
-    toolbar: false
-    intercept_redirects: false
+doctrine:
+    orm:
+        entity_managers:
+            default:
+                metadata_cache_driver: array
+                query_cache_driver:    array
```
ahocquard: I though we had 3 parameters here :thinking: 
In prod env:
```
                metadata_cache_driver:
                    cache_provider: pim
                result_cache_driver:
                    cache_provider: pim
                query_cache_driver:
                    cache_provider: pim
```

arnolanglade:
Fixed Marie 528a18345c50db97634ca623baae3642e01db997

ahocquard:
That's really cool to clean it :+1:

skeleton:
you don't have to change https://github.com/akeneo/pim-community-dev/blob/528a18345c50db97634ca623baae3642e01db997/behat.yml.dist#L128 too ?

web/app_test_database.php
```diff
@@ -1,19 +1,19 @@
...
 ini_set('display_errors', '1');
 error_reporting(E_ALL);
 
 require __DIR__.'/../vendor/autoload.php';
 
-$env = getenv('BEHAT_ENV');
-
-if ($env == null) {
-    $env = 'behat';
+$envFile = __DIR__ . '/../.env';
```
anaelChardan: Need to verify if it works without and with docker, last time when I tried it in your project it was weird.

app/config/config_test_database.yml
arnolanglade: Copy/paste from another config.

web/app_test_database.php
arnolanglade: Copy/paste from the EE. I guess it should work.

arnolanglade:
Fixed Marie 0f664df86f07cb6b4bf1654fbb67c2a641723aae Good catch thank you!

Makefile
```diff
@@ -81,6 +81,17 @@ init-pim: clean docker-compose.override.yml app/config/parameters.yml app/config
 down:
 	$(DOCKER_COMPOSE) down -v
 
+## Enable xdebug
+.PHONY: enable-xdebug
```
anaelChardan: maybe enable-xdebug-ui

arnolanglade:
prod: mock nothing
dev: mock nothing?
test_legacy_behat: use every storages (mysql/ES/FileSystem) + mock Franklin library
test_storage: use every storages (mysql/ES/FileSystem) + mock Franklin library
test_in_memory: mock every storages (mysql/ES/FileSystem) + mock Franklin library

PR #9594, opened 2019-02-18
PIM-8129: Fix default system locale

pierallard:
Previously, default system locale was "Catalan", because "en" was not found.
Now, default system locale is "en_US" to be able to show "English (United States)" again.


pierallard:
> Not tests?

I think it's useless. It does not break anything, and it's on a screen we don't often use... If you really want a test, I can add a behat (I think it's the only way to check this, as it's just a Bundle configuration)

PR #9602, opened 2019-02-19
SLA-8054: Fix slow mysql query due to wrong path

anaelChardan:

Before we were using the Doctrine Query Builder and it was taking the wrong path.

Now we use a subquery and help him with a DISTINCT. (See the CODE)

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Pim/Bundle/CatalogBundle/Doctrine/ORM/Query/VariantProductRatio.php
+++ b/src/Pim/Bundle/CatalogBundle/Doctrine/ORM/Query/VariantProductRatio.php
@@ -3,6 +3,7 @@ declare(strict_types=1);
 
 namespace Pim\Bundle\CatalogBundle\Doctrine\ORM\Query;
 
+use Doctrine\DBAL\Driver\Statement;
 use Doctrine\ORM\EntityManagerInterface;
 use Pim\Component\Catalog\Model\ProductModelInterface;
 use Pim\Component\Catalog\ProductModel\Query\CompleteVariantProducts;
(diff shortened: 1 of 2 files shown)


src/Pim/Bundle/CatalogBundle/Doctrine/ORM/Query/VariantProductRatio.php
```diff
@@ -42,44 +43,82 @@ public function findComplete(
...
 
-        $queryBuilder
-            ->innerJoin('variant_product.completenesses', 'completeness')
-            ->innerJoin('completeness.locale', 'locale')
-            ->innerJoin('completeness.channel', 'channel')
-            ->setParameter(':product_model', $productModel->getId());
+        return $this->fetchResults($join, $productModel, $channel, $locale);
+    }
+
+    private function joinToProductWithTwoLevels(): string
```
ahocquard: it is far less readable than with two duplicate queries:
- one for 2 level
- one for 1 level

Note: I think there are 6 queries under the hood, because there should be 3 methods:
- findCompleteForLocale
- findCompleteFoChannel
- findCompleteFoChannelAndLocale

anaelChardan: At first I've done it in double but I found it more difficult to see where the difference was and why. And Yes under the hood there are 6 queries I do agree. But I won't change the public api now nor complexify the env.

anaelChardan:
It's in progress.

src/Pim/Bundle/CatalogBundle/Doctrine/ORM/Query/VariantProductRatio.php
```diff
@@ -42,44 +43,83 @@ public function findComplete(
         string $channel = '',
         string $locale = ''
     ): CompleteVariantProducts {
-        $queryBuilder = $this->entityManager->createQueryBuilder();
-        $queryBuilder->select(
-            'channel.code as channel_code, locale.code as locale_code, variant_product.identifier as product_identifier, CASE WHEN (completeness.ratio = 100) THEN 1 ELSE 0 END as complete'
-        );
+        $join = null;
```
ahocquard: add a comment to explain why we do a distinct.
The previous query was totally correct, but the mysql optimizer is bad at choosing the correct path.

The only way to force the optimizer to choose the correct path is to add the distinct.

PR #9606, opened 2019-02-19
Fix avatars migration

jmleroux:
We cannot update the user because of a migration error.
The `image` should be replaced by `file_info_id` and set to null.

### DIFF
--- a/upgrades/schema/Version_3_0_20180823184342_switch_to_utf8mb4.php
+++ b/upgrades/schema/Version_3_0_20180823184342_switch_to_utf8mb4.php
@@ -108,9 +108,7 @@ class Version_3_0_20180823184342_switch_to_utf8mb4 extends AbstractMigration
             );
         }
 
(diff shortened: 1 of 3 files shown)


upgrades/schema/Version_3_0_20180823184342_switch_to_utf8mb4.php
```diff
@@ -108,9 +108,7 @@ public function up(Schema $schema)
             );
         }
 
-        $this->addSql(
-            sprintf('ALTER TABLE pim_session CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_bin', $tableName)
-        );
+        $this->addSql('ALTER TABLE pim_session CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_bin');
```
jmleroux: boyscout mode, fix useless sprintf

PR #9614, opened 2019-02-20
PIM-8157: Errors on user edit

pierallard:
- Groups are not mandatory (a user can have 0 group)
- State was missing
- Avoid scrolltop when form is re-rendered (to be validated with PO)


src/Akeneo/Platform/Bundle/UIBundle/Resources/public/js/attribute/form/edit-form.js
```diff
@@ -38,19 +38,6 @@ define([
...
         getType: function () {
             return this.type;
-        },
-
-        /**
-         * {@inheritdoc}
-         *
-         * Little hack to restore the scroll position after a complete re-render of the form.
-         */
-        render: function () {
```
tamarasaurus: :fire:

PR #9615, opened 2019-02-20
PIM-8135: Fix cursor paginator sequence

LaurentPetard:

The cursor paginator used to mass publish products called the method `$cursor->next()` at a bad moment. It resulted that a product entity was hydrated but not returned by the paginator between each loop. The Doctrine entity manager is cleared between each page, so this product is detached when it should not, because it's returned in the next page.

I took the opportunity to fix the definition of the Paginator factory to use the same value for the page size that the cursor factory.

`OK`: Done / Validated
`-`: Not needed


LaurentPetard:
@damien-carcel I don't think that an integration test be really relevant for this edge-case. Ideally This "Paginator" should not exist at all. It adds complexity without real benefit.

PR #9621, opened 2019-02-20
PIM-8131 labels cannot be used for search in bulk action

anaelChardan:

I don't know why it was put at the end, but if the parameters `options` was passed to the controller in addition of another option if wasn't taken in account.

No spec existed before (should I add some?). I won't add behat as it is not critical at all.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Pim/Bundle/EnrichBundle/Controller/Rest/AttributeController.php
+++ b/src/Pim/Bundle/EnrichBundle/Controller/Rest/AttributeController.php
@@ -145,10 +145,14 @@ class AttributeController
      */
     public function indexAction(Request $request)
     {
-        $options = [];
+        $options = $request->request->get(
+            'options',
+            ['limit' => SearchableRepositoryInterface::FETCH_LIMIT, 'locale' => null]
+        );
 
         if ($request->request->has('identifiers')) {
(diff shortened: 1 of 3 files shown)


src/Pim/Bundle/EnrichBundle/Controller/Rest/AttributeController.php
```diff
@@ -145,7 +145,10 @@ public function __construct(
      */
     public function indexAction(Request $request)
     {
-        $options = [];
+        $options = $request->request->get(
```
anaelChardan: Add default value in case options wasn't passed

src/Pim/Bundle/EnrichBundle/Controller/Rest/AttributeController.php
```diff
@@ -145,10 +145,14 @@ public function __construct(
...
+            'options',
+            ['limit' => SearchableRepositoryInterface::FETCH_LIMIT, 'locale' => null]
+        );
 
         if ($request->request->has('identifiers')) {
             $options['identifiers'] = array_unique(explode(',', $request->request->get('identifiers')));
+            $options['limit'] = count($options['identifiers']);
```
anaelChardan: For the UI if we give the identifiers, we want all the results, imho.

PR #9637, opened 2019-02-22
Setup circleci for 1.7

anaelChardan:

22 tests are failing. Merging.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,167 @@
+version: 2
+jobs:
+  setup:
+    machine:
+      enabled: true
+      docker_layer_caching: true
+    steps:
+      - checkout
+      - run:
+            name: Rename docker-compose.yml.dist
+            command: mv docker-compose.yml.dist docker-compose.yml
+      - run:
+            name: Copy docker-compose.override.yml.dist
+            command: cp .circleci/docker-compose.override.yml.dist docker-compose.override.yml
+      - run:
+          name: Setup tests results folder and log folder
+          command: mkdir -p var/tests/phpunit var/tests/behat var/tests/phpspec var/tests/csfixer var/logs
+      - run:
+          name: Setup the parameters.yml files
+          command: ./bin/docker/pim-setup.sh
+      - run:
+          name: Change owner on project dir (default user = circleci (1001) and docker needs uid 1000)
+          command: sudo chown -R 1000:1000 ../project
+      - run:
+          name: Start containers
+          command: docker-compose up -d
+      - run:
+          name: Change owner on project dir after restoring cache
+          command: sudo chown -R 1000:1000 ../project
+      - run:
+          name: Install dependencies
+          command: ./bin/docker/pim-dependencies.sh
+      - run:
+          name: Pim database installation
+          command: docker-compose exec akeneo-behat app/console --env=behat pim:install --force --clean
+      - run:
+            name: Change owner on project dir after installing when there is no cache
+            command: sudo chmod -R 777 ../project
+      - persist_to_workspace:
+          root: ~/
+          paths:
+            - project
+
+  back_static:
+      machine:
+          enabled: true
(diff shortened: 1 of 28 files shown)


.circleci/config.yml
```diff
@@ -0,0 +1,175 @@
...
+      - run:
+          name: Change owner on project dir (default user = circleci (1001) and docker needs uid 1000)
+          command: sudo chown -R 1000:1000 ../project
+      - run:
+          name: Start containers
+          command: docker-compose up -d
+      - restore_cache:
+          name: Restore cache - vendor
+          keys:
+            - vendor-v1-{{ checksum "composer.lock" }}
```
damien-carcel: Add another entry with `- vendor-v1-`.
It provides a fallback for the cache when the content of composer.lock changes.

What's more, there is no composer.lock in the PIM, so not really usable, no?

behat.yml.dist
```diff
@@ -4,19 +4,21 @@ default:
     context:
         class:  Context\FeatureContext
         parameters:
-            base_url: http://akeneo-pim-behat.local/
+            base_url: http://akeneo-behat/
```
damien-carcel: I would put back `127.0.0.1` and use your `pim-setup.sh` (like I think it is done in 2.x)

composer.json
```diff
@@ -80,7 +80,7 @@
         "sebastian/exporter": "2.0.*",
         "akeneo/phpspec-skip-example-extension": "2.0.*",
         "akeneo/php-coupling-detector": "0.2",
-        "liuggio/fastest": "1.6.*"
+        "liuggio/fastest": "1.5.*"
```
damien-carcel: why do you need that?

anaelChardan: To list scenarios, this version isn't compatible with our behat version it looks like we've never used it :D

behat.yml.dist
anaelChardan: I dunno, this version will die soon and as everyone have to change it for now :/

.circleci/config.yml
anaelChardan: I dunno, could you show an example? ;)

behat.yml.dist
damien-carcel: Acceptable argument indeed.

.circleci/config.yml
damien-carcel: Why do you need an example? This is how the cache system of Circle is documented. The file you make the checksum has to exist, or it doesn't have any sense.

For the fall back, it works as follow: https://github.com/damien-carcel/app-skeleton/blob/master/.circleci/config.yml#L35

composer.json
damien-carcel: ><

.circleci/config.yml
damien-carcel: In any case, I think you should commit your composer.lock.
Or you use `composer.json` for the checksum, but it is less usefull, as it is not supposed to change in time, so the checksum used will never change either.

.circleci/config.yml
```diff
@@ -0,0 +1,175 @@
...
+      - run:
+          name: Pim database installation
+          command: docker-compose exec akeneo-behat app/console --env=behat pim:install --force --clean
+      - run:
+            name: Change owner on project dir after installing when there is no cache
+            command: sudo chmod -R 777 ../project
+      - save_cache:
+          paths:
+            - ./vendor
+          key: vendor-v7-{{ checksum "composer.lock" }}
```
damien-carcel: You store a `v7` here, but restore a `v1` just above. Not gonna work ;-)

anaelChardan: Copy paste from the 2.3 X)

.circleci/config.yml
anaelChardan: I will remove the cache

PR #9649, opened 2019-02-22
Add screenshots, logs and improve test render for Circle CI

ahocquard:
**Execute php cs and acceptance tests even if phpspec step is failing**
As documented in Circle CI, a failing step prevent the next one to be executed.
So we add this key to prevent that behavior:
 ```
when: always
````

**https://github.com/akeneo/pim-community-dev/issues/9506 https://github.com/akeneo/pim-community-dev/issues/9507 Improve the naming of the junit files by using scenario name instead of only an uuid**

Note: We still use uuid to avoid any conflict if some feature files have the same name. 

Before:
scenario_junit_before

After:
scenario_junit

**https://github.com/akeneo/pim-community-dev/issues/9466 Improve the render of the behat scenarios by putting all junit files in the same directory**

See screenshot above to know how it works.
Note: with the new suite Franklin in 3.0, behat generates two files `all.xml` and `franklin.xml`. It is already supported here.

Before this PR:

scenario_results_before

After:
scenario_results_after

**https://github.com/akeneo/pim-community-dev/issues/9458 Add logs in the artifacts**

logs

**https://github.com/akeneo/pim-community-dev/issues/9459 Add screenshots in the artifact**

screenshots

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -77,12 +77,15 @@ jobs:
           - run:
                 name: PhpCsFixer
                 command: docker-compose exec -T fpm vendor/bin/php-cs-fixer fix --diff --dry-run --config=.php_cs.php
+                when: always
           - run:
                 name: PhpCouplingDetector
                 command: docker-compose run -T fpm vendor/bin/php-coupling-detector detect --config-file=.php_cd.php src
+                when: always
           - run:
                 name: Acceptance back
                 command: docker-compose exec fpm vendor/bin/behat --strict -p acceptance -vv
+                when: always
           - store_test_results:
                 path: var/tests
           - store_artifacts:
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by jmleroux** [comment]
Awesome work! :+1: :+1:

File: .circleci/config.yml [hunk 1]
```diff
@@ -77,16 +77,21 @@ jobs:
           - run:
```
BitOne: Why not run all static analysis before phpSpec ? This way, they will be executed even if specs don't work.

ahocquard: We did that to have all feedbacks together as it's almost "free".
But it was not the expected behavior and I just fixed that. 

But maybe what we wanted at first is wrong: just having php cs before failing if it's not ok would force us to execute in local more often, which is a good thing.

PR #9654, opened 2019-02-25
PIM-8168: accept MySql version with suffix in requirements check

jmleroux:
MySql version can be in the form '5.7.22-log' and we should accept this valid version.

See https://dev.mysql.com/doc/refman/5.7/en/server-system-variables.html#sysvar_version


jmleroux:
I don't know how to test this. Is is necessary?

greg-akeneo:
The function is not unit testable in its actual state. Create a `sanitizeMySQLVersion(string $version): string` function so it is possible to add unit tests to it.

PR #9677, opened 2019-02-27
Improve makefile

arnolanglade:

### How Makefile works
A simple makefile consists of "rules" with the following shape:

```
target ... : prerequisites ...
        command
        ...
        ...
```

A **target** is usually the name of a file that is generated by a program; examples of targets are executable or object files. A target can also be the name of an action to carry out, such as `clean' (see section Phony Targets).

A **prerequisite** is a file that is used as input to create the target. A target often depends on several files.

A **command** is an action that make carries out. A rule may have more than one command, each on its own line. Please note: you need to put a tab character at the beginning of every command line! This is an obscurity that catches the unwary. 

Extra information [here](https://ftp.gnu.org/old-gnu/Manuals/make-3.79.1/html_chapter/make_2.html)

Interesting videos:
https://www.youtube.com/watch?v=_OEDoPMvNY4&feature=youtu.be
https://www.youtube.com/watch?v=8EiRoQBzxho

## How Makefile works
A simple makefile consists of "rules" with the following shape:

```
target ... : prerequisites ...
        command
        ...
        ...
```

A **target** is usually the name of a file that is generated by a program; examples of targets are executable or object files. A target can also be the name of an action to carry out, such as `clean' (see section Phony Targets).

A **prerequisite** is a file that is used as input to create the target. A target often depends on several files.

A **command** is an action that make carries out. A rule may have more than one command, each on its own line. Please note: you need to put a tab character at the beginning of every command line! This is an obscurity that catches the unwary. 

Extra information [here](https://ftp.gnu.org/old-gnu/Manuals/make-3.79.1/html_chapter/make_2.html)

Interesting videos:
https://www.youtube.com/watch?v=_OEDoPMvNY4&feature=youtu.be
https://www.youtube.com/watch?v=8EiRoQBzxho

### Install a PIM:
```
$ make up 
$ make install-pim
```
The first command will up the docker container. Then the second one will install a PIM for prod and behat environment. It will install the front and back dependencies if needed, the database for both environment, install all assets and build the front.

### Install database
If you only want to initialize database for prod or behat/test symfony environment.  

```
$ make install-database-test
$ make install-database-prod
```
 
### Initialize frontend
The following command will install all assets and then build frontend stuff.
```
$ make webpack-dev
$ make webpack-test
```

### Run tests
Lets take an example PhpSpec: if It will run all specs, run the following command:
```
$ make phpspec
```

If you need to run a single spec, run the following command:
```
$ make phpspec F=path/to/my/Spec
```

If you need to debug a spec
```
$ make phpspec-debug F=path/to/my/Spec
```

`behat-acceptance` and `behat-acceptance-debug` are working the same way.

**Caution**: for the phpunit and behat-legacy targets you need to up docker containers before (`make up`). If you need to debug an integration/end-to-end test you need to enable xdebug manually:
```
$ make xdebug-on
```

If you want to stop debugging, you can disable xdebug
```
$ make xdebug-off
```

### End of the day
This command will shutdown all docker containers.
```
$ make down
```

### Even more
The main makefile automatically include every files located in the `make-file` folder. By default, a makefile have been created for each bounded context. Moreover you can a custom makefile in this folder if you want add targets for your personal usage. **Caution:** this file must be suffixed by `.mk`.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -115,12 +115,12 @@ down:
 ## Enable Xdebug
 .PHONY: xdebug-on
 xdebug-on: docker-compose.override.yml
-	PHP_XDEBUG_ENABLED=1 make up
+	PHP_XDEBUG_ENABLED=1 $(MAKE) up
 
 ## Disable Xdebug
 .PHONY: xdebug-off
 xdebug-off: docker-compose.override.yml
-	PHP_XDEBUG_ENABLED=0 make up
+	PHP_XDEBUG_ENABLED=0 $(MAKE) up
 
 ##
 ## Run tests suite
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,7 @@
 DOCKER_COMPOSE = docker-compose
-PHP_EXEC=$(DOCKER_COMPOSE) exec -u docker fpm php
-YARN_EXEC=$(DOCKER_COMPOSE) run --rm node yarn
+YARN_EXEC = $(DOCKER_COMPOSE) run --rm node yarn
+PHP_RUN = $(DOCKER_COMPOSE) run -u docker --rm fpm php
+PHP_EXEC = $(DOCKER_COMPOSE) exec -u docker fpm php
 
 .DEFAULT_GOAL := help
 
@@ -15,18 +16,14 @@ help:
 ## Include all *.mk files
 include make-file/*.mk
 
+## Clean backend cache
+.PHONY: clean
+clean:
+	rm -rf var/cache
+
 ##
-## PIM initialization
+## PIM configuration
 ##
-
-composer.lock: composer.json
-	$(PHP_EXEC) /usr/local/bin/composer update
-
-vendor: composer.lock
-	$(PHP_EXEC) /usr/local/bin/composer install
-
-node_modules: package.json
-	$(YARN_EXEC) install
 
 behat.yml:
 	cp ./behat.yml.dist ./behat.yml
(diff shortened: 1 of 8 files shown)


Makefile
```diff
@@ -58,29 +56,90 @@ docker-compose.override.yml:
...
 	$(DOCKER_COMPOSE) down -v
 
+##
+## Xdebug
+##
+
+## Enable Xdebug
+.PHONY: xdebug-on
+xdebug-on: docker-compose.override.yml
+	PHP_XDEBUG_ENABLED=1 make up
```
anaelChardan: make up overrides the fact that you enabled docker no?

Makefile
```diff
@@ -89,7 +148,27 @@ down:
 .PHONY: coupling ## Run the coupling-detector on Everything
 coupling: structure-coupling user-management-coupling channel-coupling enrichment-coupling
 
-### Acceptance tests
-.PHONY: coupling ## Run all acceptance tests
-acceptance:
-	vendor/bin/behat --strict -p acceptance -vv
+.PHONY: phpspec
+phpspec: vendor
+	${PHP_RUN} vendor/bin/phpspec run ${F}
```
anaelChardan: What is `${F}`?

Makefile
```diff
@@ -89,7 +148,27 @@ down:
...
+behat-acceptance: behat.yml app/config/parameters_test.yml vendor
+	${PHP_RUN} vendor/bin/behat -p acceptance ${F}
+
+.PHONY: behat-acceptance-debug
+behat-acceptance-debug: behat.yml app/config/parameters_test.yml vendor
+	${DEBUG_PHP_RUN} vendor/bin/behat -p acceptance ${F}
+
+.PHONY: phpunit
+phpunit: app/config/parameters_test.yml vendor
+	${PHP_EXEC} vendor/bin/phpunit -c app ${F}
```
anaelChardan: no phpunit-debug? :)

docker-compose.override.yml.dist
```diff
@@ -9,7 +9,7 @@ services:
       COMPOSER_HOME: '/home/docker/.composer'
       BEHAT_TMPDIR: '/srv/pim/var/cache/tmp'
       PHP_IDE_CONFIG: 'serverName=pim-ce-cli'
-      PHP_XDEBUG_ENABLED: 0
+      PHP_XDEBUG_ENABLED: "${PHP_XDEBUG_ENABLED:-0}"
```
anaelChardan: :+1:

Makefile
arnolanglade: docker-composer does not have -e option, so you need to run `make debug-on` before.

Makefile
arnolanglade: I copy paste the PR description from EE to this one. You should have look to it :)

Makefile
arnolanglade: yes the docker-composer-overide.yml has been updated.

Makefile
anaelChardan: Perfect ;)

Makefile
anaelChardan: You didn't understood my comment ;)
`make up` has already the fact that you disable xdebug so how could it work if you do PHP_XDEBUG_ENABLED=1 PHP_XDEBUG_ENABLED=0 docker-compose up -d --remove-orphan

Makefile
anaelChardan: it could have been
```
phpunit-debug:
	${DEBUG_PHP_RUN} vendor/bin/phpunit -c app ${F}
```

Makefile
arnolanglade: ah right I need to check that.

Makefile
arnolanglade: We must up the docker containers for integration tests. If we use `docker-composer run` that means we will create a new php container but ftm is already up.

anaelChardan: ok ;)

Makefile
BitOne: So no way to use a target instead of using make? If no way, at least, use $(MAKE) to make sure to use proper make with the right options.

anaelChardan: maybe a parameter with a default value? (Dunno if it is possible)

arnolanglade:
I did find another way let's use $(MAKE)

PR #9681, opened 2019-02-28
PIM-8174: Fix too long URI for multi-select with a lots of options

jmleroux:
When we have a lots of options in the multi-select attribute, we can have an error because of the URI being too long. 

This fix aims to change the call for options from http GET to POST, when we will not have this kind of limitation.

Related to #3876

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Pim/Bundle/EnrichBundle/Resources/config/routing/attributeoption.yml
+++ b/src/Pim/Bundle/EnrichBundle/Resources/config/routing/attributeoption.yml
@@ -40,4 +40,4 @@ pim_enrich_attributeoption_get:
     defaults: { _controller: pim_enrich.controller.rest.attribute_option:getAction, _format: json }
     requirements:
         identifier: '[a-zA-Z0-9_]+'
-    methods: [GET]
+    methods: [GET, POST]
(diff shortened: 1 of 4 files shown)


src/Pim/Bundle/EnrichBundle/Resources/config/routing/attributeoption.yml
```diff
@@ -40,4 +40,4 @@ pim_enrich_attributeoption_get:
     defaults: { _controller: pim_enrich.controller.rest.attribute_option:getAction, _format: json }
     requirements:
         identifier: '[a-zA-Z0-9_]+'
-    methods: [GET]
+    methods: [GET, POST]
```
jmleroux: Should i restrict it to POST when it can work with both methods?

ahocquard: According to what @juliensnz said on Slack, I think we should only support one method.

jmleroux:
> Please remove the compatibility to GET method

I tried it and it never passed the CI.

src/Pim/Bundle/EnrichBundle/Resources/config/routing/attributeoption.yml
jmleroux: Will do in 2.3 pull up

PR #9691, opened 2019-03-01
PIM-8175: add the possiblity to filter on one or several index names when resetting ES indexes

ahocquard:

A customer has a problem on its proposal.
In order to fix that, we need to re-index all the ES of the PIM.

But he has several thousands of products and the problem is not on the product.
So, we added the possibility to re-index only a specific index to fix this kind of problems on a production environment.

```
bin/console akeneo:elasticsearch:reset-indexes -i akeneo_pim_product -i akeneo_pim_product_model
```

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Akeneo/Bundle/ElasticsearchBundle/Command/ResetIndexesCommand.php
+++ b/src/Akeneo/Bundle/ElasticsearchBundle/Command/ResetIndexesCommand.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Akeneo\Bundle\ElasticsearchBundle\Command;
 
+use Akeneo\Bundle\ElasticsearchBundle\Client;
 use Pim\Bundle\CatalogBundle\Command\IndexProductCommand;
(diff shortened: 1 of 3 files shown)


Doodoune:
Please add a Changelog ;)

Doodoune:
And think about changing the documentation if needed ^^

src/Akeneo/Bundle/ElasticsearchBundle/Command/ResetIndexesCommand.php
```diff
@@ -34,6 +35,7 @@ protected function configure()
                 InputOption::VALUE_NONE,
                 'Resets all registered ES indexes prior to reindex'
             )
+            ->addOption('index', 'i', InputOption::VALUE_REQUIRED | InputOption::VALUE_IS_ARRAY, 'ES index name to reset')
             ->setDescription('Resets all registered ES indexes');
```
jmleroux:
```suggestion
            ->setDescription('Resets registered ES indexes');
```
Removing 'all' because it's not always the case.

PR #9695, opened 2019-03-04
PIM-8178: fix family attributes selector

jmleroux:
When adding new attribute(s) to a family by search, we sent a request with the search term AND all the attributes already in the family in an `excluded_identifiers` array. But when there are a lot of attributes in the family, we end up with a very large array of POST parameters that goes beyond the PHP `max_input_vars` limit.

This PR will fix the issue by sending only the family code instead of all the family attributes. The attributes are then excluded server side.

`OK`: Done / Validated
`-`: Not needed


Doodoune:
Don't forget to add a Changelog

src/Akeneo/Pim/Structure/Bundle/Doctrine/ORM/Repository/InternalApi/AttributeSearchableRepository.php
```diff
@@ -70,6 +69,7 @@ protected function resolveOptions(array $options)
                 'is_locale_specific'     => null,
                 'useable_as_grid_filter' => null,
                 'families'               => null,
+                'exclude_identifiers_from_family' => null,
```
fitn: I am not comfortable with this naming but I cannot find a good name

PR #9696, opened 2019-03-04
AOB-376: Modify Oro Assetic to be able to compile several css stylesheets

phaseinducer:

To move the Onboarder UI to the pim-onboarder bundle we need to apply two different styles on the PIM (with different colors, etc.), the overridden style being in the bundle.
We didn't want to add a top-level css class because it would mean to duplicate the less in the bundle.
We chose instead to find a way to build several stylesheets (one with the native style, and another with the native style plus some overrides) and switch between them when needed.

The issue is that the less compilation is handled by the Oro Assetic bundle and has been designed to always generate only one css file from the less.
I chose to twist a concept introduced by Oro Assetic : the asset groups. Originally it's used for debugging purpose. You can specify some files to be in debug mode (so they won't be compressed in the final css file but dumped and linked as is). Spoiler alert : we never used that in the PIM.

Now these groups can also be used to define in which css file they will be compiled. For example if you have the following configuration :
```
css:
    group_one:
      - file_a.less
      - file_b.less
    group_two:
      - file_c.less

stylesheets:
    pim:
        groups: [group_one]
    michel:
        groups: [group_one, group_two]
```
You will get a `pim.css` file containing compiled less from `file_a.less` and `file_b.less`, and a `michel.css` file containing the compiled less from all three files.
Both css files will be dumped by the `oro:assetic:dump` command and linked from the Twig template in `<link...>` tags. Only the first stylesheet will be active by default. To switch the active stylesheet a bit of js will be needed.

**Drawbacks**
- If new groups are added or the existing groups are reorganized in the PIM we'll have to change the conf in the bundle, but we can expect it doesn't happen very often.
- It's not great to modify the native code for such a specific use case, but it's not a behavior we could have implemented in the bundle.
- It's complex.

**Advantages**
- Nothing will be duplicated in the bundle.
- It doesn't introduce any BC breaks. If no `stylesheets` entry is found in the conf the behavior will be the same as before.
- It works.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Oro/Bundle/AsseticBundle/Node/OroAsseticNode.php
+++ b/src/Oro/Bundle/AsseticBundle/Node/OroAsseticNode.php
@@ -6,33 +6,23 @@ use Assetic\Asset\AssetInterface;
 
 class OroAsseticNode extends \Twig_Node
 {
-    /**
-     * @var AssetInterface
-     */
-    protected $compressedAsset;
+    /** @var AssetInterface[] */
+    protected $compressAssets;
 
-    /**
-     * @var AssetInterface
-     */
-    protected $unCompressAsset;
-
-    protected $nameCompress;
-    protected $nameUnCompress;
+    /** @var AssetInterface[] */
+    protected $unCompressAssets;
 
     public function __construct(
         array $assets,
-        array $names,
-        $filters,
-        $inputs,
+        array $filters,
+        array $inputs,
         \Twig_NodeInterface $body,
         array $attributes = [],
-        $lineno = 0,
-        $tag = null
+        int $lineno = 0,
+        ?string $tag = null
     ) {
-        $this->nameCompress = $names['compress'];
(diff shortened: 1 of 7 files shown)


src/Oro/Bundle/AsseticBundle/Node/OroAsseticNode.php
```diff
@@ -48,121 +38,108 @@ public function __construct(
...
-        $compiler->raw('$this->env->getExtension(\'asset\')->getAssetUrl(')
+        $compiler
+            ->write("// asset \"$name\"\n")
+            ->write('$context[\'name\'] = ')
+            ->repr($name)
+            ->raw(";\n")
+            ->write('$context[')
+            ->repr($this->getAttribute('var_name'))
+            ->raw('] = ')
+            ->raw('$this->env->getExtension(\'asset\')->getAssetUrl(')
```
AymericPlanche: :heart:

PR #9708, opened 2019-03-06
DAPI-37: do not see black listed job in process tracker and widget

willy-ahva:

Move the LastOperationsWidget from `Akeneo\Platform\Bundle\DashboardBundle\Widget` to `Akeneo\Platform\Bundle\ImportExportBundle\Widget` to respect bounded context.

Remove old job execution manager that was used for the last operation widget.

Extract all data retrieving and format from the Widget to a fetcher to be able to remove easily the old Oro stack when we'll have time. See LastOperationsFetcher.

Add a service tag `akeneo_batch.job.not_visible` to jobs that we don't want to see in process tracker and last operation widget.

Add a grid listener to job grids to remove from retrieved data blacklisted jobs.

`OK`: Done / Validated
`-`: Not needed


src/Oro/Bundle/PimDataGridBundle/Resources/config/datagrid_listeners.yml
```diff
@@ -173,3 +173,16 @@ services:
...
             - { name: kernel.event_listener, event: oro_datagrid.datgrid.build.after.last-export-executions-grid, method: onBuildAfter }
+
+    pim_enrich.event_listener.black_list_job_codes_from_grid_listener:
+        class: 'Oro\Bundle\PimDataGridBundle\EventListener\BlackListJobsFromGridListener'
+        arguments:
+            -
+              - 'refresh_project_completeness_calculation'
+              - 'compute_product_models_descendants'
+              - 'compute_family_variant_structure_changes'
+              - 'compute_completeness_of_products_family'
```
fitn: Still thinking it's a bad idea to inject strings to the DI. I would have a registry of jobs instead of a collection of hardcoded strings.

tests/back/Platform/Specification/Bundle/ImportExportBundle/Query/GetLastOperationsSpec.php
```diff
@@ -0,0 +1,89 @@
...
+    {
+        $this->beConstructedWith($connection, $notVisibleJobs);
+    }
+
+    function it_is_a_last_operations_query()
+    {
+        $this->shouldImplement(GetLastOperationsInterface::class);
+    }
+
+    function it_executes_the_query_builder(
```
LaurentPetard: I don't think that to "spec" the methods `execute` and `getQueryBuilder` is necessary. The real value come from the integration test.

PR #9713, opened 2019-03-06
PIM-8187: Adds queries to fetch descendants product model ids and product ids

fitn:

`OK`: Done / Validated
`-`: Not needed


src/Pim/Bundle/CatalogBundle/tests/integration/Doctrine/ORM/Query/DescendantProductIdsIntegration.php
```diff
@@ -0,0 +1,36 @@
...
+use Akeneo\Test\Integration\TestCase;
+use Pim\Bundle\CatalogBundle\Entity\AttributeGroup;
+
+class DescendantProductIdsIntegration extends TestCase
+{
+    public function test_it_fetches_product_ids_from_product_model_ids()
+    {
+        $query = $this->get('pim_catalog.query.descendant_product_ids');
+
+        $resultRows = $query->fetchFromProductModelIds([1]);
```
mmetayer: I'd like it better if you fetched the product model and its descendants and compare ids, instead of using hard-coded ids (would prevent an error in case of a catalog update).

PR #9724, opened 2019-03-08
Reduce the number of errors with PhpStan level 2

jjanvier:
With this PR, we reduce the number of errors from 317 to 97 with PhpStan set to level 2.
As this PR is already getting big and as others problems are much harder to fix (possible bugs, dead code etc..), I prefer merge this PR as soon as possible.

### DIFF
--- a/src/Akeneo/Pim/Enrichment/Bundle/PdfGeneration/Builder/DompdfBuilder.php
+++ b/src/Akeneo/Pim/Enrichment/Bundle/PdfGeneration/Builder/DompdfBuilder.php
@@ -60,7 +60,7 @@ class DompdfBuilder implements PdfBuilderInterface
         }
 
         $this->dompdf = new \DOMPDF();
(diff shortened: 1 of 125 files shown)


src/Akeneo/Pim/Enrichment/Bundle/PdfGeneration/Builder/DompdfBuilder.php
```diff
@@ -60,7 +60,7 @@ protected function render($html)
         }
 
         $this->dompdf = new \DOMPDF();
-        $this->dompdf->set_paper(DOMPDF_DEFAULT_PAPER_SIZE);
```
anaelChardan: Are we sure it is defined?

jjanvier: yes I checked

PR #9752, opened 2019-03-13
Listen to SIGTERM and SIGINT to stop job queue runner after current job finished

stefandoorn:
I've been trying to use the job queue runner while also using autoscaling. That can lead to jobs being killed on downscale. This change listens to SIGTERM and SIGINT and stops the command after the current job has been finished. If there hasn't been any jobs then an TypeError is thrown due to `consume` not returning the proper type, but the command still still finish afterwards. This combined with a proper grace period on downscale should work. I haven't been able to test it that much on a large queue yet, as my local Docker set-up is somehow broken. I appreciate feedback and will try to deploy it myself in the meanwhile on a real installation to see how it develops.

### DIFF
--- a/src/Akeneo/Tool/Bundle/BatchQueueBundle/Command/JobQueueConsumerCommand.php
+++ b/src/Akeneo/Tool/Bundle/BatchQueueBundle/Command/JobQueueConsumerCommand.php
@@ -41,6 +41,9 @@ class JobQueueConsumerCommand extends ContainerAwareCommand
     /** Interval in microseconds before checking if the process is still running. */
     private const RUNNING_PROCESS_CHECK_INTERVAL = 200000;
 
+    /** Set to true on a kill signal, will finish current job and stop  */
+    private $shouldStop = false;
+
     /**
      * {@inheritdoc}
      */
@@ -57,6 +60,12 @@ class JobQueueConsumerCommand extends ContainerAwareCommand
      */
     protected function execute(InputInterface $input, OutputInterface $output)
     {
+        if (extension_loaded('pcntl')) {
+            pcntl_async_signals(true);
+            pcntl_signal(SIGTERM, [$this, 'stopCommand']);
+            pcntl_signal(SIGINT, [$this, 'stopCommand']);
+        }
+
         $errOutput = $output instanceof ConsoleOutputInterface ? $output->getErrorOutput() : $output;
 
         $consumerName = Uuid::uuid4();
@@ -114,7 +123,7 @@ class JobQueueConsumerCommand extends ContainerAwareCommand
 
                 sleep(self::EXCEPTION_WAIT_INTERVAL);
             }
-        } while (false === $input->getOption('run-once'));
+        } while (false === $this->shouldStop && false === $input->getOption('run-once'));
     }
 
     /**
@@ -168,5 +177,10 @@ class JobQueueConsumerCommand extends ContainerAwareCommand
     private function getJobExecutionManager(): JobExecutionManager
     {
         return $this->getContainer()->get('akeneo_batch_queue.manager.job_execution_manager');
+    }
+
+    public function stopCommand(): void
+    {
+        $this->shouldStop = true;
     }
 }


BitOne:
Thanks a lot @stefandoorn for you PR.

Unfortunately, I see several problems in the code you propose:
 - this makes the code have different behaviors depending on the presence of a PHP extension. And as this PHP extension is not part of the requirements of Akeneo, this means that both behaviors have to be tested on our CI, and our support team should be able to manage both behaviors, as we cannot add a new system requirement in a patch or minor versions, it's against our release policy.
 - the management of the signals becomes the responsibility of the script instead of being the responsibility of the PHP engine. In case of unwanted or unresponsive behavior of the script, it becomes unable to properly respond to these signals, while today it's managed by the low level PHP interpreter. So that would make the application more fragile and would force us to use the KILL signal to stop the process, meaning losing buffers content for example.

I think it can be managed differently. Maybe a flag that could be put somewhere to ask the process to stop for example.

Can you create another issue so we can discuss your needs, especially if you could detail your downscaling constraints (some idea of grace period duration, the interaction your cluster scheduler can have with the container/nodes that are downscaling, etc...)  , that would be very helpful to give you a proper answer.
We could has well integrate our cloud team in the discussion.

Thanks a lot!

stefandoorn:
@BitOne I'm not sure how we can listen both to KILL signal & set a flag (because basically that's what I do already). I don't know how we can make the PHP engine responsible for this, not sure PHP can work this way at all. 

The main issue is that I'm running Akeneo on Kubernetes, wanting to use auto-scaling on my CLI pods. When Kubernetes horizontal autoscaler decides to scale down, e.g. when CPU is too low on average, it will send a SIGINT to the main process the container is running (in my case the job runner). Normally if unhandled, it will terminate directly (now the case). So basically listening to SIGINT and gracefully stopping + a high grace period on Kubernetes side (default of 30 secs) makes the current job properly finish and therefore downscaling won't kill running jobs.

I do see your concerns, but couldn't get my head around a better solution. It all works as default without the extension, I didn't want to make the extension required for everyone. Obviously it could be put behind a feature flag (configuration) which defaults to being turned OFF.

BitOne:
You cannot listen for the KILL signal, that's the idea of this signal: it's handled by the kernel, not by the process. So the process cannot properly close the opened files and connection. And we want to avoid that ;)

For the grace period, I'm afraid that 30s is not enough for a lot of kind of job to finish (import, export and mass edit for example would take far more time).

Maybe if you could trigger the downscale only based on the fact there's no running job, that would be enough?

stefandoorn:
Grace period is actually configurable, so I'm combining this with a higher grace period (30 min in our set-up).

> So the process cannot properly close the opened files and connection. And we want to avoid that ;)

Isn't this better than not listening at all? Currently if I give SIGINT to my job runner, it dies instantly and nothing is cleaned up. This PR let's it wait until the job is properly done.

> Maybe if you could trigger the downscale only based on the fact there's no running job, that would be enough?

Yep, that could probably be, but that requires a specific set-up to inject metrics about amount of waiting/running jobs and let the autoscaler decide based on that. Issue is there still that I might have 3 pods running, 1 job left, the autoscaler kills 2 of them, but that might also kill basically the last job running.

I might be overlooking something, but I find it hard to come up with another solution that could be integrated into Akeneo providing this feature.

BitOne:
If you kill the queue runner, unless we made a misconfiguration on the process object, the job should continue and finish properly (unless your container is killed). So I don't understand exactly what you're trying to do.

stefandoorn:
True, that's basically the main issue, the main process gets killed on the container and therefore Kubernetes takes out the whole container at once (and subprocess also). On a long-running VM it will end properly I think, as you described.

BitOne:
Oh okay, I see. So your problem is your entry point. It shouldn't be the queue runner, but a process waiting for all PHP process to be finished on the container.
In this case, a flag to tell the daemon to stop would make it stop, then the entry point waits for the current job (if any) to stop.

The missing piece in this puzzle would be a way to tell the daemon to shutdown. But you can do that in your entry point, by capturing the PID of queue runner and trap the SIGINT in this entry point, sending it to the queue runner and waiting for the other PHP process to finish.

The bonus would be that it's easier to do that in shell and far less invasive for the application.

stefandoorn:
Yep, that would also be a proper solution I think. Not sure how to do it yet, but might do some digging around to see how to best write that.

BitOne:
I'm pretty sure you can leverage the `wait [PID]` function from the shell, as well as `trap` to execute function when receiving the SIGINT.

I'm still not sure how tolerant Kubernetes would be if you have a long-running process that would take hours to finish.

Maybe one direction would be to have a way for the queue runner to relaunch jobs that have been interrupted. But that's another kind of complexity, as we would need to differentiate jobs that have been interrupted by an external actor from jobs that crashed. And we maybe don't want to relaunch jobs that crashes as they may crash again and again (maybe implementing a retry policy ?).

stefandoorn:
Yep. Haven't tested it yet, but I think it might be something like this:

```
#!/usr/bin/env bash
set -x

pid=0

sigterm_handler() {
  if [ $pid -ne 0 ]; then
    # Kill daemon instantly, dont want to spawn new jobs
    kill -SIGTERM "$pid"
    wait "$pid"

    # Get all PHP process pids, wait for them to exit
    pgrep -f php | xargs wait
  fi
  exit 143; # SIGTERM
}

# Listen to SIGTERM
trap 'kill ${!}; sigterm_handler' SIGTERM

# Start job queue daemon in background, and get pid
/usr/bin/php /path/to/bin/console akeneo:batch:job-queue-consumer-daemon &
pid="$!"

# Wait forever
while true
do
  tail -f /dev/null & wait ${!}
done
```

stefandoorn:
Ok, this seems to be working:

```
#!/usr/bin/env bash
set -x

pid=0

sigterm_handler() {
  if [ $pid -ne 0 ]; then
    # Kill daemon instantly, dont want to spawn new jobs
    echo "Got SIGTERM, killing pid $pid"
    kill -SIGTERM "$pid"
    wait "$pid"

    # Get all PHP process pids, wait for them to exit
    echo "Search for other PHP processes"
    while pgrep -f php &> /dev/null; do echo 'Sleeping 30 seconds, PHP process found'; sleep 30; done
  fi
  exit 143; # SIGTERM
}

# Listen to SIGTERM
trap 'kill ${!}; sigterm_handler' SIGTERM

# Start job queue daemon in background, and get pid
/usr/bin/php /srv/pim/bin/console akeneo:batch:job-queue-consumer-daemon &
pid="$!"

# Wait forever
while true
do
  echo "Starting process"
  tail -f /dev/null & wait ${!}
done
```

Inspired by https://medium.com/@gchudnov/trapping-signals-in-docker-containers-7a57fdda7d86.

Waiting was less handy, simply sleeping and checking for another PHP process did the job. I wasn't sure if waiting for another spawned process was going to work, but this was easier to debug & test on a live container (by just starting a manual PHP process and kill it after a while).

stefandoorn:
Thanks @BitOne for the guidance.

PR #9757, opened 2019-03-14
AOB-404: SaaSify the PIM

damien-carcel:


.circleci/config.yml
```diff
@@ -40,7 +43,7 @@ jobs:
           command: ./bin/docker/pim-dependencies.sh
       - run:
           name: Pim database installation
-          command: docker-compose exec fpm bin/console --env=behat pim:install --force --clean
+          command: docker-compose exec fpm bin/console --env=prod pim:install --force --clean
```
damien-carcel: This way we don't use Google Cloud Storage at all in the setup job.

app/config/config_prod.yml
```diff
@@ -27,3 +27,6 @@ monolog:
 oro_assetic:
     css_debug:      ~
     css_debug_all:  false
+
+parameters:
+    installer_data: PimInstallerBundle:minimal
```
damien-carcel: To speed up `setup`

tests/legacy/features/Behat/Context/Domain/Spread/XlsxFileContext.php
```diff
@@ -235,9 +258,6 @@ protected function compareFile(array $expectedLines, array $actualLines, $path)
      */
     protected function compareLines(array $expectedLines, array $actualLines, $path)
     {
-        $expectedCount = count($expectedLines);
```
damien-carcel: not used…

PR #9766, opened 2019-03-15
PIM-8222: do not accept /draft in a product SKU

jmleroux:
This route is in conflict with an API route on EE: `pimee_api_product_model_draft_get`
The fix is quite ugly and will not allow the use of a draft suffix in SKU, but i cannot change an API route.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Pim/Bundle/ApiBundle/Resources/config/routing/product_model.yml
+++ b/src/Pim/Bundle/ApiBundle/Resources/config/routing/product_model.yml
@@ -8,7 +8,7 @@ pim_api_product_model_get:
     defaults: { _controller: pim_api.controller.product_model:getAction, _format: json }
     methods: [GET]
     requirements:
-        code: (.(?!/draft$))+
+        code: (.+$)(?<!/draft|/proposal)
 
 pim_api_product_model_create:
     path: /product-models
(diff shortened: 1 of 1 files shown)


src/Pim/Bundle/ApiBundle/Resources/config/routing/product_model.yml
```diff
@@ -8,7 +8,7 @@ pim_api_product_model_get:
     defaults: { _controller: pim_api.controller.product_model:getAction, _format: json }
     methods: [GET]
     requirements:
-        code: .+
+        code: (.(?!/draft$))+
```
skeleton: do the same as https://github.com/akeneo/pim-community-dev/blob/b004115819353128a630927dadfe15982046f8e7/src/Pim/Bundle/ApiBundle/Resources/config/routing/product.yml#L11

jmleroux: I'm not sure it's the same issue:
`(.+$)` means basically any character UNTIL the end of the line
`(?<!/draft|/proposal) ` means not preceded by `/draft` or `/proposal`

So i think this regex means "any string not starting by /draft and not starting by /proposal"

I'll test on an online tool to prove (or not) my point

jmleroux: and once again, you are right :wink:

PR #9776, opened 2019-03-18
PIM-8228: fix incorrect completeness

jmleroux:
When we change the requirement of an attribute in a family, it is not always removed from the table. It can be present with a `required = false` property.
But when we fetch requirements for completeness, we didn't test for `required` and these non required attributes were taken into account in the completeness.

This PR solves the issue by testing for the `required` property before adding the attribute in the completeness calculation.

`OK`: Done / Validated
`-`: Not needed


LaureBro:
@jmleroux can you confirm the changelog has also been updated? ;) thanks!

jmleroux:
ui: https://github.com/akeneo/pim-community-dev/blob/v1.7.40/CHANGELOG-1.7.md

:wink:

PR #9780, opened 2019-03-19
[2.3] attribute option form errors and validator

FabienSalles:
Hi,

I created a custom attribute in order to override default simple select attribute type.

When I submitted my first option the controller returned a 400 without any messages. The problem was due to the fact that the method `getFormErrors` only return errors for child forms so I overrided it to also return global form errors.

Edit : I also used parameter in the `AttributeTypeForOptionValidator` class to handle supported attribute types like a lot of your services because otherwise, we have to override your class instead of just the parameter of your service.

`OK`: Done / Validated
`-`: Not needed


FabienSalles:
I think this PR can be closed (cf  #10052)

PR #9815, opened 2019-03-25
Upgrade Phpstan to install Symfony's extension

jjanvier:
This extension allows us to identify : 
- undefined Symfony services
- services that are declared private but called publicly from the container

Also, this will be needed as soon as we'll try to reach level 2 of phpstan.

For instance, if have something like:

```php
$someService = $container->get('some-service-id');
$someService->someMethod();
```

Without the extension, phpstan will raise an error because it does not know what is `$someService` and therefore if the method `someMethod()` exists. With the extension, phpstan will detect `$someService` is an instance of the class `SomeAwesomeService` and therefore will be able to check if the method `someMethod` exists.

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -1,0 +1,6 @@
+includes:
+    - vendor/phpstan/phpstan-symfony/extension.neon
+parameters:
+    symfony:
+        container_xml_path: %rootDir%/../../../var/cache/dev/appDevDebugProjectContainer.xml
+
--- a/composer.json
+++ b/composer.json
@@ -95,7 +95,8 @@
         "squizlabs/php_codesniffer": "2.*",
         "friends-of-behat/symfony-extension": "^2.0",
         "liuggio/fastest": "1.6.*",
-        "phpstan/phpstan": "^0.11.1"
+        "phpstan/phpstan": "^0.11.1",
+        "phpstan/phpstan-symfony": "^0.11.1"
     },
     "suggest": {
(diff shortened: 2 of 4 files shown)


phpstan.neon
```diff
@@ -0,0 +1,5 @@
+includes:
+    - vendor/phpstan/phpstan-symfony/extension.neon
+parameters:
+    symfony:
+        container_xml_path: %rootDir%/../../../var/cache/dev/appDevDebugProjectContainer.xml
```
arnolanglade: Should we use the production container?

jjanvier: no idea 
we'll see what the CI says

jjanvier: I used the "recommended" configuration https://github.com/phpstan/phpstan-symfony#usage

BitOne:
Does it makes the https://github.com/akeneo/pim-community-dev/blob/master/bin/check-services-instantiability script obsolete?

phpstan.neon
arnolanglade: ok :+1:

jjanvier:
@BitOne hmmm,  I'd say no.
Phpstan checks only the PHP code. It checks the things like `$container->get(...)`.
Your script checks all the service definitions we have declared. And most of them are called through `$container->get(...)`.

BTW, your script currently detects "Found 14 non instantiable services on a total of 2260.". Is it normal? Is it supposed to be launched on the CI?

Issue #9800, opened 2019-03-22
Composer dependencies strictness


stefandoorn:
While using Akeneo now for a while, I ran into the issue that a recent Twig vulnerability (https://symfony.com/blog/twig-sandbox-information-disclosure) is preventing me from deploying Akeneo via CI now, as we by default integrate security checks in there. I would like to update such dependency, but the only way to do this currently is either PR (without knowing all the code is sometimes hard) or wait until a version bump has been done. Twig e.g. is required as `1.36.0`, while `^1.36` would prevent this as I can handle this on my side. Obviously we know less which version exactly is being used, but in case CI is running regularly we'll probably figure out rather soon. Also, the app gets tested against more versions in the outside world, therefore probably leading to more small bugs being resolved in good time and in the end having the app more resilient against future bugs. I would propose having more open version constraints, within limits of course. I use Sylius (https://github.com/sylius/sylius) a lot and in there this is being used, and it works rather fine as I can see. Obviously sometimes a minor issue, but that can easily be resolved if we keep up releasing as often as you currently do already.

jmleroux:
Hi @stefandoorn ,
you can do PR and i'll run them on our CI.

stefandoorn:
Ok, cool, will do that somewhere this week.

stefandoorn:
@jmleroux Could you run #9829 against CI?

PR #9829, opened 2019-03-26
Composer dependencies following semver

stefandoorn:
PoC for #9800.

I haven't removed any dependencies (as mentioned by @ahocquard in #9571) yet and tried to be strict where it was also strict now.


stefandoorn:
cc @jjanvier

PR #9854, opened 2019-04-01
DAPI-58: Make the grid filter column not specific to product

willy-ahva:

Needed for EE PR.

Two modules are used to display filters panel. One was dedicated to products and another one is used in all other grids. The product panel is more customizable and according to Pierre, the fact that we have two modules comes from a merge into master error. We should only have the product module because even if it was specific, it is customizable, paginable and runs with a query which does not hydrate all attributes.
That's why I decided to make it more generic and reuse it in proposal grid (CE PR). Next step is to remove the second module and replace it everywhere with the generic one.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Oro/Bundle/PimDataGridBundle/Resources/public/js/datafilter/filter/label_or_identifier-filter.js
+++ b/src/Oro/Bundle/PimDataGridBundle/Resources/public/js/datafilter/filter/label_or_identifier-filter.js
@@ -18,8 +18,8 @@ define(
             inputValueSelector: 'input[name="value"]',
 
             events: {
-                'keydown input[name="value"]': 'runTimeout',
-                'keypress input[name="value"]': 'runTimeout'
+                'keydown input[name=value]': 'runTimeout',
+                'keypress input[name=value]': 'runTimeout'
             },
 
             emptyValue: {
@@ -45,6 +45,8 @@ define(
                         label: __('pim_datagrid.search', {label: this.label})
                     })
                 );
+
+                this.delegateEvents();
             },
 
             /**
@@ -95,12 +97,17 @@ define(
 
             /**
              * Appends filter to a grid
-            * */
+             * If .search-zone is not in the element that is to say it could be somewhere in the page.
+             */
             moveFilter: function(collection, element) {
-                if (element) {
+                if (element.$('.search-zone').length !== 0) {
                     element.$('.search-zone')
                     .empty()
                     .append(this.$el.get(0));
+                } else if ($('.edit-form .search-zone').length !== 0) {
+                    $('.edit-form .search-zone')
+                        .empty()
+                        .append(this.$el.get(0));
                 }
             }
         });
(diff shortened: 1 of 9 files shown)


src/Oro/Bundle/PimDataGridBundle/Resources/public/js/datafilter/filter/label_or_identifier-filter.js
```diff
@@ -97,10 +99,14 @@ define(
              * Appends filter to a grid
             * */
             moveFilter: function(collection, element) {
-                if (element) {
+                if (element.$('.search-zone').length !== 0) {
                     element.$('.search-zone')
                     .empty()
                     .append(this.$el.get(0));
+                } else if ($('.search-zone').length !== 0) {
```
phaseinducer: Is it possible to use a more precise selector ?

phaseinducer: Also a little explanation in the method doc would be useful imo.

src/Oro/Bundle/PimDataGridBundle/Resources/public/js/datafilter/filters-column.ts
```diff
@@ -274,19 +246,19 @@ class FiltersColumn extends BaseView {
...
-    return _.template(this.filterGroupTemplate)({
+    return this.filterGroupTemplate({
       filters,
       groupName,
       ignoredFilters: this.ignoredFilters,
     });
   }
 
-  groupFilters(filters: GridFilter[]) {
+  groupFilters(filters: GridFilter[]): any {
```
phaseinducer: `Object` ?

src/Oro/Bundle/PimDataGridBundle/Resources/public/js/datafilter/filter/label_or_identifier-filter.js
willy-ahva: I didn't find one and change the class could impact a lot of functionalities.

src/Oro/Bundle/PimDataGridBundle/Resources/public/js/datafilter/filters-column.ts
willy-ahva: Seen irl

src/Oro/Bundle/PimDataGridBundle/Resources/public/js/datafilter/filter/label_or_identifier-filter.js
phaseinducer: You can surely find ancestors and use spaces. IRL if it's not clear enough?

tamarasaurus:
Can you add an explanation in the description please?

tamarasaurus:
It's not a merge error that we have two filter list modules. In the refactor we only did the product grid to improve performance. The other grids will need some changes - see https://github.com/akeneo/pim-community-dev/pull/8878/files

src/Oro/Bundle/PimDataGridBundle/Resources/public/js/datafilter/filter/label_or_identifier-filter.js
willy-ahva: done

PR #9877, opened 2019-04-05
PIM-8269: Do not display empty values from deleted attributes from fa…

anaelChardan:

The problem is that when an attribute is removed from a family and the value is empty from all channels and all locales we should hide them on the PEF. Indeed, we cannot have empty optionnal attributes.

The best thing we can do is to remove all empty values on the database but in a fix it cant be done, it is a huge subject!

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Pim/Component/Catalog/Factory/ValueCollectionFactory.php
+++ b/src/Pim/Component/Catalog/Factory/ValueCollectionFactory.php
@@ -48,7 +48,7 @@ class ValueCollectionFactory implements ValueCollectionFactoryInterface
     /**
      * {@inheritdoc}
      *
-     * Raw values that correspond to an non existing attribute (that was deleted
+     * Raw values that correspond to a non existing attribute (that was deleted
      * for instance) are NOT loaded.
      *
      * @see \Pim\Component\Catalog\Normalizer\Storage\Product\ProductValuesNormalizer.php
@@ -76,7 +76,17 @@ class ValueCollectionFactory implements ValueCollectionFactoryInterface
                         }
 
                         try {
-                            $values[] = $this->valueFactory->create($attribute, $channelCode, $localeCode, $data, true);
+                            $value = $this->valueFactory->create($attribute, $channelCode, $localeCode, $data, true);
+                            $productData = $value->getData();
+                            $isEmpty = (
+                                null === $productData ||
+                                (is_string($productData) && '' === trim($productData))  ||
+                                (is_array($productData) && 0 === count($productData))
+                            );
+
+                            if (!$isEmpty) {
+                                $values[] = $value;
+                            }
                         } catch (InvalidOptionException $e) {
                             $this->logger->warning(
                                 sprintf(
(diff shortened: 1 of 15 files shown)


src/Pim/Component/Catalog/Factory/ValueCollectionFactory.php
```diff
@@ -76,7 +76,11 @@ public function createFromStorageFormat(array $rawValues)
                         }
 
                         try {
-                            $values[] = $this->valueFactory->create($attribute, $channelCode, $localeCode, $data, true);
+                            $isEmpty = ($data === null || (is_string($data) && trim($data) === '') || empty($data) || (is_array($data) && count($data) === 0));
```
ahocquard: I would not check it here but but after creating the value can be empty due to a removed option, currency, etc

src/Pim/Component/Catalog/spec/Factory/ValueCollectionFactorySpec.php
```diff
@@ -32,7 +32,7 @@ function it_is_initializable()
         $this->shouldHaveType(ValueCollectionFactory::class);
     }
 
-    function it_creates_a_values_collection_from_the_storage_format(
+    function it_creates_a_values_collection_from_the_storage_format_and_removes_empty_one(
```
jjanvier: normally, you should not change the current tests
you should add a new one that explains your problem

src/Pim/Component/Catalog/Factory/ValueCollectionFactory.php
anaelChardan: Ok ;)

src/Pim/Component/Catalog/spec/Factory/ValueCollectionFactorySpec.php
anaelChardan: Fixed

src/Pim/Component/Catalog/spec/Factory/ValueCollectionFactorySpec.php
```diff
@@ -281,4 +284,36 @@ function it_create_empty_value_is_wrong_format_when_creating_a_values_collection
         $actualValues->shouldReturnAnInstanceOf(ValueCollection::class);
         $actualValues->shouldHaveCount(1);
     }
+
+    function it_does_not_return_empty_values(
+        $valueFactory,
+        $attributeRepository,
+        AttributeInterface $description,
+        ValueInterface $value1
```
ahocquard: you should create a value instead of mocking its behavior (same for $description attribute). As the type of the data is not exactly the same, the behavior between the implementation is not the same (getData returns different things).

src/Pim/Component/Catalog/spec/Factory/ValueCollectionFactorySpec.php
```diff
@@ -281,4 +284,36 @@ function it_create_empty_value_is_wrong_format_when_creating_a_values_collection
...
+    ) {
+        $description->getCode()->willReturn('description');
+        $description->isUnique()->willReturn(false);
+        $attributeRepository->findOneByIdentifier('description')->willReturn($description);
+
+        $value1->getAttribute()->willReturn($description);
+        $value1->getData()->willReturn('');
+
+        $valueFactory
+            ->create($description, 'ecommerce', 'en_US', '', true)
```
ahocquard: I would do a test for values returnning null and values returning an array also

PR #9916, opened 2019-04-17
InMemoryRoleRepository must implement RoleRepositoryInterface

AymericPlanche:

The InMemoryRoleRepository must implement the RoleRepositoryInterface which is used as type for the https://github.com/akeneo/pim-community-dev/blob/3.0/src/Akeneo/UserManagement/Component/Factory/UserFactory.php#L62

EDIT: we also removed the `strtolower(str_replace('ROLE_', '', $index))` in the `InMemoryRoleRepository::save` that is not compatible with the real behavior done in https://github.com/akeneo/pim-community-dev/blob/3.0/src/Akeneo/UserManagement/Component/Model/Role.php#L74

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/tests/back/Acceptance/spec/User/InMemoryRoleRepositorySpec.php
+++ b/tests/back/Acceptance/spec/User/InMemoryRoleRepositorySpec.php
@@ -2,11 +2,11 @@
 
 namespace spec\Akeneo\Test\Acceptance\User;
 
-use Akeneo\Tool\Component\StorageUtils\Repository\IdentifiableObjectRepositoryInterface;
 use Akeneo\Tool\Component\StorageUtils\Saver\SaverInterface;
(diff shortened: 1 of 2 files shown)


tests/back/Acceptance/spec/User/InMemoryRoleRepositorySpec.php
```diff
@@ -18,7 +18,7 @@ function it_is_initializable()
 
     function it_is_a_identifiable_object_repository()
```
phaseinducer: You can change this^^

pierallard: https://www.youtube.com/watch?v=otCpCn0l4Wo

PR #9885, opened 2019-04-09
#1062: Hide the file path property for cloud edition

arnolanglade:
**Description (for Contributor and Core Developer)**

We must not display the `File path` input (import profile) in the SAAS edition because this is not SAAS compliant. More information are available [here](https://akeneo.atlassian.net/secure/RapidBoard.jspa?rapidView=67&projectKey=TIP&view=planning&selectedIssue=TIP-1092&epics=visible&selectedEpic=TIP-1087).

To  hide this input, an new env var named EDITION is given to webpack:

```bash
$ yarn requirements && NODE_PATH=node_modules EDITION=cloud  webpack --config $npm_package_config_source/webpack.config.js
```
Webpack make this var global and make it accessible via `process.env.EDITION` in the application (example [here](https://github.com/akeneo/pim-community-dev/pull/9885/files#diff-0baa3075248b34fc2e17c4730dc78852R3)). 

The edition is not mandatory, you still can run webpack without specifying it:

```bash
$ yarn requirements && NODE_PATH=node_modules webpack --config $npm_package_config_source/webpack.config.js
```

**How to build front end for cloud edition:**

**Definition Of Done (for Core Developer only)**

| Q                                 | A
| --------------------------------- | ---
| Added Specs                       | -
| Added legacy Behats               | -
| Added acceptance tests            | -
| Added integration tests           | -
| Changelog updated                 | -
| Review and 2 GTM                  | Todo
| Micro Demo to the PO (Story only) | -
| Migration script                  | -
| Tech Doc                          | -

`Todo`: Pending / Work in progress
`OK`: Done / Validated
`-`: Not needed

PR #9922, opened 2019-04-19
TIP-1092 Hide import now button from the import profile export

arnolanglade:

We must not display the `Import now` switcher item (import profile) in the SAAS edition because this is not SAAS compliant. More information are available [here](https://akeneo.atlassian.net/secure/RapidBoard.jspa?rapidView=67&projectKey=TIP&view=planning&selectedIssue=TIP-1092&epics=visible&selectedEpic=TIP-1087).

Before:
before

After:
after

To hide this, an new env var named EDITION is given to webpack:

```bash
$ yarn requirements && NODE_PATH=node_modules EDITION=cloud  webpack --config $npm_package_config_source/webpack.config.js
```
Webpack make this var global and make it accessible via `process.env.EDITION` in the application (example [here](https://github.com/akeneo/pim-community-dev/pull/9885/files#diff-0baa3075248b34fc2e17c4730dc78852R3)). 

The edition is not mandatory, you still can run webpack without specifying it:

```bash
$ yarn requirements && NODE_PATH=node_modules webpack --config $npm_package_config_source/webpack.config.js
```

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Akeneo/Platform/Bundle/UIBundle/Resources/public/js/job/import/switcher.js
+++ b/src/Akeneo/Platform/Bundle/UIBundle/Resources/public/js/job/import/switcher.js
@@ -11,11 +11,13 @@ define(
     [
         'underscore',
         'pim/form',
+        'pim/edition',
         'pim/template/import/switcher'
     ],
     function (
         _,
         BaseForm,
+        pimEdition,
         template
     ) {
         return BaseForm.extend({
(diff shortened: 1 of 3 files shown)


anaelChardan:
So to be sure, they can only import a new file ?

src/Akeneo/Platform/Bundle/UIBundle/Resources/public/js/job/import/switcher.js
```diff
@@ -50,10 +52,13 @@ define(
                     this.setCurrentActionCode(_.first(this.actions).code);
                 }
 
-                this.$el.empty().append(this.template({
-                    actions: this.actions,
-                    current: this.currentActionCode
-                }));
+
+                if (this.actions.length > 1) {
```
tamarasaurus: so not for only 1 action?

arnolanglade: I don't want to display the action button if there is only one action!

arnolanglade:
Yes it is hard to test, there is not test on this part.
Yes they can only import files from the UI, in SAAS edition you can access hard drive.

jjanvier:
@ahocquard we'll try to test the cloud edition somehow (with a dedicated build for instance)

PR #9945, opened 2019-04-23
Fix exclude folders to make them relative

tamarasaurus:

We have a bug in the configuration of `ts-loader` in the webpack config. Since we tell ts-loader to exclude certain files from the build, e.g. a file with a path matching any of these terms: `/lib|node_modules|vendor|tests|src|packages/` it also excluded project folders like `/src/code/my-pim-project`, causing the build to break. This PR fixes this issue by making the excluded files relative to the project directory. 

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -183,7 +183,12 @@ const webpackConfig = {
           },
         ],
         include: /(web\/bundles)/,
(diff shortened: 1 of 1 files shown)


nei:
@tamarasaurus it worked for me. thank you.

PR #9946, opened 2019-04-24
 TIP-1126: Add a new storage for file uploaded from the PEF

arnolanglade:

Before product saving if an user uploads a file from the PEF, this file is stored in a temporal storage on the file system. This is not SAAS compliant, we want to avoid to store those files in the local sotrage! This PR changes the way we manage those files: instead of using "manual move" on the file system we use `Flystem` with a new "storage": `pef_tmp_storage`.

`OK`: Done / Validated
`-`: Not needed


src/Akeneo/Tool/Component/FileStorage/File/FileStorer.php
```diff
@@ -51,27 +51,20 @@ public function __construct(
...
         try {
-            $options = [];
-            $mimeType = $file->getMimeType();
-            if (null !== $mimeType) {
-                $options['ContentType'] = $mimeType;
-            }
-            $isFileWritten = $filesystem->writeStream($file->getKey(), $resource, $options);
-        } catch (FileExistsException $e) {
+            $isFileWritten = $this->mountManager->move(
+                (string) new Path('pefTmpStorage', $localFile->getPathname()),
```
arnolanglade: All files should go on a "global" tmp storage. it will be easy for the file storer to picker file from the storage.
It avoids accidental complicity.

src/Akeneo/Tool/Component/FileStorage/FileInfoFactory.php
```diff
@@ -46,11 +49,14 @@ public function createFromRawFile(\SplFileInfo $rawFile, $destFsAlias)
             $extension = $rawFile->getExtension();
         }
 
-        $size = filesize($rawFile->getPathname());
-        $mimeType = MimeTypeGuesser::getInstance()->guess($rawFile->getPathname());
+        $filesystem = $this->filesystemProvider->getFilesystem($destFsAlias);
+        $metadata = $filesystem->getMetadata($rawFile->getPathname());
+        $mimeType = $metadata['mimetype'];
+        $size = $metadata['size'];
+        $sha1 = sha1($filesystem->read($rawFile->getPathname()));
```
arnolanglade: is it the best to do that?
is file hash necessary ?

src/Akeneo/Platform/config/bundles/oneup_flysystem.yml
```diff
@@ -9,7 +9,13 @@ oneup_flysystem:
         archivist_adapter:
             local:
                 directory: '%archive_dir%'
+        pef_tmp_storage_adapter:
```
arnolanglade: Should we rename to tmp? should we use tmp_storage_adapter instead ?

PR #9960, opened 2019-04-26
#1092: Configure form extension for the cloud edition

arnolanglade:

Some form extensions were not well configured in this [PR](https://github.com/akeneo/pim-community-dev/pull/9958). This PR fixes that. We only allow the user to upload the file he want to import.

`OK`: Done / Validated
`-`: Not needed


fitn:
I don't understand why we call this property like that.
If functionally, the cloud is different and we should remove some buttons. Well, then make these buttons configurable is a good solution.
Anyway, I feel the technical part should not appear in the backend or frontend code. The configuration for SAAS/PAAS/On Premise should handle that depending on their needs.
If something is done for PAAS then we will add another parameter hideForPaas and we can add another specific parameter for on premise only too.

To me, the feature goal is to make this button **hidden** or not. Then the cloud configuration makes it hidden.

arnolanglade:
There is a single failure because of https://github.com/akeneo/pim-community-dev/pull/9992

arnolanglade:
We have 2 editions to maintain: cloud and on premise edition so we don't have the choice. We need to make some change like that to optimize the PIM for the cloud edition. Our integrator will take advantages of this change for hosting the PIM.

PR #9966, opened 2019-04-29
PIM-8318: Fix CE 2.3 CI

jmleroux:
We have intl issues on CE 2.3 which is linked to Intl.
See https://github.com/symfony/symfony/issues/31089 for more information.
We also must apply a patch done in 3.0 on AjaxFilter necessary with this Symfony version.
See #9528 
And thx @ahocquard  for helping me on the subject.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -27,10 +27,6 @@ jobs:
       - run:
           name: Start containers
           command: docker-compose up -d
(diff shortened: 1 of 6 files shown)


jmleroux:
Not sure this is the way to go: it fixes intl, but other issues appears.
Failure number bumped  from 7 to 32 :cry:

PR #9967, opened 2019-04-29
Merge 3.1 into master

jjanvier:


src/Akeneo/Platform/CommunityVersion.php
```diff
@@ -12,7 +12,7 @@
 class CommunityVersion
 {
     /** @staticvar string */
-    const VERSION = '3.1.0';
+    const VERSION = '3.1.0-BETA1';
```
jmleroux:
```suggestion
    const VERSION = '3.2.0';
```
Because master is for 3.2

mmetayer: maybe `3.2.x`?

PR #9987, opened 2019-05-02
PIM-8270: Update export jobs after a change on a channel category

jmleroux:

If the default channel category tree is changed, the export profiles are not updated.
The old tree remains, and is still used for the product exports: wrong products can be exported and so on...

Proposed solution was:
- Write a dedicated saver for the channel, that send a domain event when the category tree of a channel has been updated.
- Subscribe to this event and update the exports parameters.

For the reviewers:
- Check that the different classes are in the right bounded context.
- There is not many product exports so the update of the job instances is done synchronously. A proposed solution was to do this in a Job; but I don't thinks it's necessary ?

`OK`: Done / Validated
`-`: Not needed


src/Akeneo/Channel/Bundle/Doctrine/Saver/ChannelSaver.php
```diff
@@ -0,0 +1,137 @@
...
+        if (empty($channels)) {
+            return;
+        }
+
+        $data = \array_map(
+            function (ChannelInterface $channel) use ($commonOptions) {
+                return [
+                    $channel,
+                    \array_merge($commonOptions, ['is_new' => null === $channel->getId()]),
+                    $this->isChannelCategoryUpdated($channel->getCode(), $channel->getCategory()->getCode()),
```
ahocquard: I would avoid to do that in the saver as we are not in a SLA anymore, but I would stack events directly from the model.
It would be much more useful as we wanted to detcet other changes in other recent SLA:
- locale activated/deactivated for the channel
- currency added/removed for the chanel

This kind of changes are important as it impacts the whole catalog (values + completenessses) and we will need that kind of events anyway.

Then, I will pop this events in this saver.

fitn: To be honest, I am not sure this should be defined as an event. I am probably wrong by do not create any event without discussing with PO/PM ;) And a patch is not the best way to handle that.

We already did other solution in the past based on PRE/POST save events in the subscriber.

Why not doing like we did previously

ahocquard: It is not a patch, it is for the 3.2.

It is an event as you it is an intent from the user. I don't think that we should discuss with PO/PM for every events as it is an action doable from the API, from the import, and from the UI.

Other solutions are dirty workaround because you can't detect change easily (doctrine changeset or sql request...). and I would personally not do that in our enrichment context.

anaelChardan: I do agree with @ahocquard, I'm updating it, Alex is definitely right, let's add Business Events. POST/PRE are so meaningless (except for a patch) ;).

src/Akeneo/Pim/Enrichment/Bundle/EventSubscriber/ConfigureCategoryTreeForExportJobsAfterChangingTheChannelCategoryTree.php
```diff
@@ -0,0 +1,112 @@
...
+use Akeneo\Tool\Component\StorageUtils\Updater\ObjectUpdaterInterface;
+use Doctrine\Common\Persistence\ObjectRepository;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+
+/**
+ * @author Paul Chasle <paul.chasle@akeneo.com>
+ * @copyright 2019 Akeneo SAS (http://www.akeneo.com)
+ * @license http://opensource.org/licenses/osl-3.0.php Open Software License (OSL 3.0)
+ */
+final class ConfigureCategoryTreeForExportJobsAfterChangingTheChannelCategoryTree implements EventSubscriberInterface
```
anaelChardan: I would better put it in ImportExportBundle in Platform no?

src/Akeneo/Channel/Bundle/Doctrine/Saver/ChannelSaver.php
```diff
@@ -0,0 +1,122 @@
...
+use Doctrine\Common\Persistence\ObjectManager;
+use Symfony\Component\EventDispatcher\EventDispatcherInterface;
+use Symfony\Component\EventDispatcher\GenericEvent;
+
+/**
+ * @author Paul Chasle <paul.chasle@akeneo.com>
+ * @copyright 2019 Akeneo SAS (http://www.akeneo.com)
+ * @license http://opensource.org/licenses/osl-3.0.php Open Software License (OSL 3.0)
+ */
+final class ChannelSaver implements SaverInterface, BulkSaverInterface
```
ahocquard: can we not take advantage of it to introduce an interface `ChannelSaverInterface`? In order to not inject a server but use typehint. There is no BC and it's still better. 
I don't think we will do an  refactor with a repository for channel before a long time...

src/Akeneo/Channel/Component/Model/Channel.php
```diff
@@ -196,8 +201,18 @@ public function getCategory()
      */
     public function setCategory(CategoryInterface $category)
     {
+        $previousCategoryCode = null;
+
+        if ($this->getCategory() !== null) {
```
ahocquard: I would do two distinct branches.

```
if ($this->getCategory() === null) {
    $this->category = $category;
} else if ($this->getCategory()->getCode()  !== $category->getCode()) {
   $previousCategoryCode = $this->getCategory()->getCode();
   $this->addEvent(new ChannelCategoryHasBeenUpdated($this->code, $previousCategoryCode, $category->getCode()));
}
```

It's much more clear I think

src/Akeneo/Pim/Enrichment/Bundle/EventSubscriber/ConfigureCategoryTreeForExportJobsAfterChangingTheChannelCategoryTree.php
pchasle: It was my first approach but @arnolanglade suggested that it was more an enrichment context as export jobs are defined here; if I recall correctly.

ahocquard: Category is enrichment part. I think it's ok. It's the configuration of the jobs of the enrichment context that is modified.

src/Akeneo/Channel/Bundle/Doctrine/Saver/ChannelSaver.php
```diff
@@ -0,0 +1,122 @@
...
+        if (empty($channels)) {
+            return;
+        }
+
+        $data = \array_map(
+            function (ChannelInterface $channel) use ($commonOptions) {
+                return [
+                    $channel,
+                    \array_merge($commonOptions, ['is_new' => null === $channel->getId()]),
+                    $channel->popEvents()
```
ahocquard: it is not declared in the interface, isn't it? in that case, we should add a TIP card and check if the method exist, and return an empty array otherwise.
What do think about that?

src/Akeneo/Pim/Enrichment/Bundle/EventSubscriber/ConfigureCategoryTreeForExportJobsAfterChangingTheChannelCategoryTree.php
anaelChardan: That's weird because it is a consequencies of channel of an enrichment thing but the listener is from the import export black box :)

src/Akeneo/Channel/Component/Event/ChannelEvent.php
```diff
@@ -0,0 +1,14 @@
...
+namespace Akeneo\Channel\Component\Event;
+
+/**
+ * @author    Anael Chardan <anael.chardan@akeneo.com>
+ * @copyright 2019 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+interface ChannelEvent
+{
+    public function getName(): string;
```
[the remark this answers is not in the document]
anaelChardan: It's only to make the symfony dispatch works :/ as it does not work through types but through names :(. And yes we could add channel code ;)

src/Akeneo/Channel/Bundle/Doctrine/Saver/ChannelSaver.php
anaelChardan: I do agree so much <3

src/Akeneo/Channel/Component/Model/Channel.php
anaelChardan: it's indeed clearer.

src/Akeneo/Channel/Bundle/Doctrine/Saver/ChannelSaver.php
anaelChardan: I've added it to the interface. It's a BC.

src/Akeneo/Channel/Component/Model/Channel.php
anaelChardan: You forgot a little thing ;)

src/Akeneo/Channel/Bundle/Doctrine/Saver/ChannelSaver.php
ahocquard: :crossed_fingers: I can check if someone changed it in custom code of our EE customer in PAAS.

src/Akeneo/Pim/Enrichment/Bundle/EventSubscriber/ConfigureCategoryTreeForExportJobsAfterChangingTheChannelCategoryTree.php
ahocquard: Yes, all that stuff is weird. But I think the listener should be in enrichment also.

src/Akeneo/Channel/Bundle/Doctrine/Saver/ChannelSaver.php
anaelChardan: Perfect ;)

ahocquard: I confirm that no customer in PAAS or SAAS is extending the ChannelInterface.

src/Akeneo/Channel/Bundle/Storage/Orm/ChannelSaver.php
```diff
@@ -0,0 +1,120 @@
...
+    /**
+     * @param ChannelInterface[] $channels
+     */
+    private function doSaveAll(array $channels, array $commonOptions = []): void
+    {
+        if (empty($channels)) {
+            return;
+        }
+
+        $data = \array_map(
```
jmleroux:
```suggestion
        $data = array_map(
```

src/Akeneo/Channel/Bundle/Storage/Orm/ChannelSaver.php
```diff
@@ -0,0 +1,120 @@
...
+            function (ChannelInterface $channel) use ($commonOptions) {
+                return [
+                    $channel,
+                    \array_merge($commonOptions, ['is_new' => null === $channel->getId()]),
```
jmleroux:
```suggestion
                    array_merge($commonOptions, ['is_new' => null === $channel->getId()]),
```

src/Akeneo/Channel/Bundle/Doctrine/Saver/ChannelSaver.php
anaelChardan: Thanks! ;)

UPGRADE.md
```diff
@@ -292,34 +176,13 @@ Before updating the dependencies and migrating your data, please deactivate all
...
-
-    ```less
-        @import (less) "./web/bundles/yourbundle/assets/less/styles.css";
-        @import "./web/bundles/yourbundle/assets/less/bundle.less";
-    ```
+2. Adapt your custom codes to handle this breaking changes we introduced:
 
-    If you are importing a .css file, you must add `(less)` after the import, as above. If you only have .less files in your bundle's assets.yml, you can remove it.
+ - Service `pim_catalog.saver.channel` class has been changed to `Akeneo\Channel\Bundle\Storage\Orm\ChannelSaver`.
+ - Interface `Akeneo\Channel\Component\Model\ChannelInterface` has a new methods `popEvents(): array`
```
pchasle: Shouldn't be this only be in the CHANGELOG.md as it doesn't require a manual upgrade for the standard edition configuration ?

anaelChardan: It requires an upgrade for all customers which are actually extending the Channel (imho they shouldn't) ;)

UPGRADE.md
```diff
@@ -1,8 +1,8 @@
-# UPGRADE FROM 3.0 TO 3.1
+# UPGRADE FROM 3.1 TO 3.2
 
 ## Disclaimer
 
-> Please check that you're using Akeneo PIM v3.0.
+> Please check that you're using Akeneo PIM v3.2.
```
christophe-chausseray: Please check that you're using Akeneo PIM v3.1.

anaelChardan: Good catch! ;)

src/Akeneo/Channel/Component/Event/ChannelCategoryHasBeenUpdated.php
```diff
@@ -0,0 +1,54 @@
...
+
+use Symfony\Component\EventDispatcher\Event;
+
+/**
+ * @author Paul Chasle <paul.chasle@akeneo.com>
+ * @author Anaël Chardan <anael.chardan@akeneo.com>
+ * @copyright 2019 Akeneo SAS (http://www.akeneo.com)
+ * @license http://opensource.org/licenses/osl-3.0.php Open Software License (OSL 3.0)
+ */
+final class ChannelCategoryHasBeenUpdated extends Event implements ChannelEvent
```
SamirBoulil: I don't see the point of the interface here

src/Akeneo/Channel/Component/Event/ChannelCategoryHasBeenUpdated.php
```diff
@@ -0,0 +1,54 @@
...
+{
+    public const EVENT_NAME = 'CHANNEL_CATEGORY_HAS_BEEN_UPDATED';
```
SamirBoulil: I'd use the FQCN of the class as the name

src/Akeneo/Pim/Enrichment/Bundle/Resources/config/event_subscribers.yml
```diff
@@ -129,3 +129,15 @@ services:
...
             - { name: kernel.event_subscriber }
+
+    pim_catalog.event_subscriber.update_exports_on_channel_update:
+        class: 'Akeneo\Pim\Enrichment\Bundle\EventSubscriber\ConfigureCategoryTreeForExportJobsAfterChangingTheChannelCategoryTree'
+        arguments:
+            - '@akeneo_batch.job.job_instance_repository'
+            - '@akeneo_batch.updater.job_instance'
+            - '@akeneo_batch.saver.job_instance'
+            -
+                - 'csv_product_export'
```
SamirBoulil: This might be problematic for people having custom jobs no ? ( I don't see how to solve this ^^)

src/Akeneo/Channel/Component/Event/ChannelCategoryHasBeenUpdated.php
anaelChardan: Good idea, the problem is when you move the class, you move the listener.

src/Akeneo/Pim/Enrichment/Bundle/Resources/config/event_subscribers.yml
anaelChardan: It is only type of job not the job itself I guess.

src/Akeneo/Channel/Component/Event/ChannelCategoryHasBeenUpdated.php
anaelChardan: Just to gather identical type, I will remove it.

src/Akeneo/Pim/Enrichment/Bundle/Resources/config/event_subscribers.yml
ahocquard: It's the the responsibility of customers doing a job in my opinion.

src/Akeneo/Channel/Component/Event/ChannelCategoryHasBeenUpdated.php
SamirBoulil: We usually have dedicated sed commands that should fix the issue

src/Akeneo/Channel/Bundle/Storage/Orm/ChannelSaver.php
```diff
@@ -0,0 +1,120 @@
...
+    {
+        $commonOptions['unitary'] = false;
+
+        $this->doSaveAll($channels, $commonOptions);
+    }
+
+    /**
+     * @param ChannelInterface[] $channels
+     */
+    private function doSaveAll(array $channels, array $commonOptions = []): void
```
SamirBoulil: Honestly this function is doing multiple things that could have been split in multiple privates doing one job at a time

PR #9995, opened 2019-05-06
Clear Docker images cache

jmleroux:
- Added a docker-compose pull step to pull fresh images and avoid docker images cache issues
- added yarn.lock like we added composer.lock before
- use a Circle CI env var for our cache key. This way we will not have to modify the config to just invalidate cache. Changing the env var will be enough.


.circleci/config.yml
```diff
@@ -12,6 +12,9 @@ jobs:
       - run:
             name: Copy docker-compose.override.yml.dist
             command: cp .circleci/docker-compose.override.yml.dist docker-compose.override.yml
+      - run:
+            name: Pull the latest Docker images
+            command: docker-compose pull
```
ahocquard: is it really pulling the latest version?

.circleci/config.yml
```diff
@@ -43,8 +50,12 @@ jobs:
             name: Pim yarn test installation
             command: docker-compose run --rm node yarn run webpack-test
       - run:
-            name: Change owner on project dir after installing when there is no cache
+            name: Change owner on project dir after install
             command: sudo chmod -R 777 ../project
+      - save_cache:
```
grena: Don't you have permission problems by putting the cache after the chmod?
If I remember well, the permissions are "reset" when we put back the cache

.circleci/config.yml
jmleroux: Yep, you can see it it in the setup job of this build: https://circleci.com/gh/akeneo/pim-community-dev/8991

But i think i will have to add it on the behat job too because this is the one using the more containers and the cache seems to need refresh per container. Or machine. It's still not clear.

jmleroux:
> Missing EE also I think

Yes, i will dot it afterward.

.circleci/config.yml
jmleroux: I think i've done like in recent branches (3.0 and 3.1), but i can check

ahocquard: I think it's another problem.
We had a problem of permissions because we set-up a cache and then removed some lines about permissions.

As the cache keep the permissions (I think), we didn't notice that we removed a line to make it work when there is no cache.

We discovered it when we purged the cache: it was not working anymore. I don't know if it's understandable :smile:

.circleci/config.yml
ahocquard: A machine = a container in Circle CI. I think the cache is shared thanks to some shared filesystem. Maybe shared among all machine of a cluster behind the hood.

jmleroux: I added the pull step on behat and still see some containers pulling images when others do not, so i think it's necessary. And we lose only one minute at most.

PR #10002, opened 2019-05-06
TIP-1126: Uploading a file in the PEF

jjanvier:
## Story

Currently, the upload of medias is not SAAS compliant. Indeed, when we upload a media we:

1. perform a first request POST /media that stores the file on the local file system (on /tmp). The response looks like
```
{  
   "originalFilename":"Wallpaper-Ziggy-Boston-Nantes-D\u00fcsseldorf.jpg",
   "filePath":"\/tmp\/pim\/file_storage\/d\/b\/1\/e\/\/db1e48388d5a4a2b8fb038b1772855fac12d0d5a\/Wallpaper-Ziggy-Boston-Nantes-D\u00fcsseldorf.jpg"
}
```
This is also security flaw.

2. perform a second request POST /enrich/product/rest/1 that saves the product and moves the file from the local file system to the object storage. A file info object is created during this request.

This process is not SAAS compliant as any request should be stateless. We have several "frontend" PIM behind a load balancer. Here the first request could be handled by the PIM A while the second request could be handled by the PIM B. This would not work as the media is located on the file system of the PIM A and not on the file system of the PIM B.

## What we did here?

Two main ideas:

1. Store the media files in the object storage directly from the controller. So we have just one request now. And yes, that means we'll store useless files in case the user never clicks on "Save". But anyway, we already have those useless files. We have a dedicated story to handle them.

2. As the file is already stored when we save the product, that means `MediaProductValue` now handle `FileInfo` objects, and not a local file path. It is now the same mechanism as for `Option` or `ReferenceData`: in order to set an option or a reference data to a product, those objects must exist a priori. That's why we modified `MediaAttributeSetter` and thus, the places where it was used : the imports (products/product models/proposals) and the rules.

### DIFF
--- a/src/Akeneo/Pim/Enrichment/Bundle/Controller/InternalApi/MediaController.php
+++ b/src/Akeneo/Pim/Enrichment/Bundle/Controller/InternalApi/MediaController.php
@@ -4,6 +4,7 @@ namespace Akeneo\Pim\Enrichment\Bundle\Controller\InternalApi;
 
 use Akeneo\Pim\Enrichment\Component\FileStorage;
 use Akeneo\Tool\Component\FileStorage\File\FileStorer;
+use Akeneo\Tool\Component\FileStorage\Model\FileInfoInterface;
 use Akeneo\Tool\Component\FileStorage\PathGeneratorInterface;
 use Symfony\Component\HttpFoundation\File\UploadedFile;
 use Symfony\Component\HttpFoundation\JsonResponse;
@@ -85,7 +86,7 @@ class MediaController
         );
     }
 
-    protected function storeFile(UploadedFile $uploadedFile)
+    protected function storeFile(UploadedFile $uploadedFile): FileInfoInterface
     {
         return $this->fileStorer->store($uploadedFile, FileStorage::CATALOG_STORAGE_ALIAS);
     }
--- a/src/Akeneo/Pim/Enrichment/Bundle/Controller/InternalApi/MediaController.php
+++ b/src/Akeneo/Pim/Enrichment/Bundle/Controller/InternalApi/MediaController.php
@@ -2,8 +2,10 @@
 
 namespace Akeneo\Pim\Enrichment\Bundle\Controller\InternalApi;
 
+use Akeneo\Pim\Enrichment\Component\FileStorage;
+use Akeneo\Tool\Component\FileStorage\File\FileStorer;
 use Akeneo\Tool\Component\FileStorage\PathGeneratorInterface;
-use Symfony\Component\HttpFoundation\File\Exception\FileException;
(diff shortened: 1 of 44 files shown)


src/Akeneo/Pim/Enrichment/Component/Product/Connector/Processor/Denormalizer/ProductModelProcessor.php
```diff
@@ -127,6 +140,8 @@ public function process($standardProductModel): ?ProductModelInterface
             }
         }
 
+        $standardProductModel['values'] = $this->storeMedias($standardProductModel['values']);
```
arnolanglade: add a `if`

src/Akeneo/Pim/Enrichment/Component/Product/Connector/Processor/Denormalizer/MediaStorer.php
```diff
@@ -0,0 +1,48 @@
...
+    {
+        $mediaAttributes = $this->attributeRepository->findMediaAttributeCodes();
+
+        foreach ($rawProductValues as $attributeCode => $values) {
+            if (in_array($attributeCode, $mediaAttributes)) {
+                foreach ($values as $index => $value) {
+                    if (empty($value['data'])) {
+                        continue;
+                    }
+                    if (!is_file($value['data'])) {
```
BitOne: Maybe this should be the responsability of the FileStorer only ? As anyway in case of non readable file this will generate an error in the FileStorer.

src/Akeneo/Pim/Enrichment/Bundle/Controller/InternalApi/MediaController.php
```diff
@@ -69,23 +75,18 @@ public function postAction(Request $request)
...
             [
-                'originalFilename' => $file->getClientOriginalName(),
-                'filePath'         => $movedFile->getPathname()
+                'originalFilename' => $uploadedFile->getClientOriginalName(),
+                'filePath'         => $file->getKey()
             ]
         );
     }
+
+    protected function storeFile(UploadedFile $uploadedFile)
```
arnolanglade: return type hint

src/Akeneo/Pim/Enrichment/Component/Product/Connector/Processor/Denormalizer/MediaStorer.php
jjanvier: fixed

src/Akeneo/Pim/Enrichment/Bundle/Controller/InternalApi/MediaController.php
jjanvier: fixed

PR #10020, opened 2019-05-09
TIP-1157: Create real query for API List Product Models

mmetayer:

`OK`: Done / Validated
`-`: Not needed


src/Akeneo/Pim/Enrichment/Bundle/ProductModel/Query/Sql/GetGroupAssociationsByProductModelCodes.php
```diff
@@ -83,7 +83,12 @@ public function fromProductModelCodes(array $productModelCodes): array
         $results = [];
 
         foreach ($rows as $row) {
-            $results[$row['product_model_code']] = array_map('array_filter', json_decode($row['associations'], true));
+            $associations = json_decode($row['associations'], true);
```
ahocquard: FYI, I sort and get only values from array, it's easier to test as it helps to always return the same result.

mmetayer: :+1: I'll do it on TIP-1168

PR #10022, opened 2019-05-09
 AOB-479: Change relative resource paths in less files to absolute paths

phaseinducer:

**The problem**
As almost all resource paths in `url(...)` are relative in the PIM, when you `@import` PIM less files in another bundle (to override stuff for example), Assetic resolves paths wrongly.
We never saw the issue in the PIM because all the less is in UIBundle, but we had many issues in the Onboarder because of this.
This issue surely happened in customers projects too.

For example, a path declared as `url("../../images/jambon.png")` in the PIM could be resolved as `bundles/images/jambon.png` when compiled in a less importing the original file from the PIM, which doesn't exist at all.

**The fix**
- Always use absolute resource paths.
- Improve the `oro:assetic:dump` command to check if resource paths are absolute and if each path points to an existing file. It won't break the compilation because it could cause BC breaks, it will just display warnings.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Akeneo/Platform/Bundle/UIBundle/Resources/public/less/components/CatalogVolume.less
+++ b/src/Akeneo/Platform/Bundle/UIBundle/Resources/public/less/components/CatalogVolume.less
@@ -14,11 +14,7 @@
 
     &--count-products,
     &--average-max-records-per-reference-entity {
-      background-image: url("../../images/icon-product.svg");
-    }
-
-    &--count-channels {
-      background-image: url("../../images/icon-channel.svg");
+      background-image: url("/bundles/pimui/images/icon-product.svg");
     }
 
     &--average-max-attributes-per-family,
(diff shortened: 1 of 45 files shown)


src/Akeneo/Platform/Bundle/UIBundle/Resources/public/less/components/CatalogVolume.less
```diff
@@ -14,11 +14,7 @@
 
     &--count-products,
     &--average-max-records-per-reference-entity {
-      background-image: url("../../images/icon-product.svg");
-    }
-
-    &--count-channels {
```
willy-ahva: You removed a class here, is it ok ?

src/Oro/Bundle/AsseticBundle/Command/OroAsseticDumpCommand.php
```diff
@@ -127,7 +130,10 @@ private function doDump(AssetInterface $asset, OutputInterface $output)
                 )
             );
 
-            if (false === @file_put_contents($target, $asset->dump())) {
+            $fileContent = $asset->dump();
+            $this->checkResourcePaths($fileContent, $output);
+
+            if (false === @file_put_contents($target, $fileContent)) {
```
willy-ahva: is it compliant with SAAS ?

src/Oro/Bundle/AsseticBundle/Command/OroAsseticDumpCommand.php
```diff
@@ -140,4 +146,53 @@ private function getAssetVarCombinations(AssetInterface $asset)
...
+                    )
+                );
+
+                continue;
+            }
+
+            $path = $this->basePath.$url;
+            $cleanedPath = $this->cleanResourcePath($path);
+
+            if (!file_exists($cleanedPath)) {
```
willy-ahva: SAAS ?

src/Akeneo/Platform/Bundle/UIBundle/Resources/public/less/components/CatalogVolume.less
phaseinducer: It's overriden few lines below and the image doesn't exist anymore, so it's useless.

src/Oro/Bundle/AsseticBundle/Command/OroAsseticDumpCommand.php
phaseinducer: Nope, but if we need to rework this it will be done in a dedicated card imo.

PR #10038, opened 2019-05-14
PIM-8340: Allow to delete a user who authored or replied to a comment

mmetayer:
When a user comments on a product (or replies to a comment), it used to be impossible to remove him, because of the author foreign key. The easiest and cleanest solution would be to add an `ON DELETE SET NULL` clause to this foreign key, but we cannot do it in a 3.0 patch.

The proposed solution, which is far from ideal, is to add a (temporary) event subscriber which:
- starts a transaction and updates the concerned comments on PRE REMOVE
- and commits this transaction on POST REMOVE

`OK`: Done / Validated
`-`: Not needed


src/Akeneo/Pim/Enrichment/Bundle/Doctrine/ORM/EventListener/RemoveUserListener.php
```diff
@@ -0,0 +1,38 @@
...
+use Akeneo\Pim\Enrichment\Component\Comment\Model\Comment;
+use Akeneo\UserManagement\Component\Model\UserInterface;
+use Doctrine\ORM\Event\LifecycleEventArgs;
+
+/**
+ * @author    Mathias METAYER <mathias.metayer@akeneo.com>
+ * @copyright 2019 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ *
+ * @todo merge 3.1: Remove this listener, we should instead add an ON DELETE SET NULL on the author foreign key constraint
```
mmetayer: 3.1 or master?

src/Akeneo/Pim/Enrichment/Bundle/Doctrine/ORM/EventListener/RemoveUserListener.php
```diff
@@ -0,0 +1,38 @@
...
+     * @param LifecycleEventArgs $args
+     */
+    public function preRemove(LifecycleEventArgs $args): void
+    {
+        $user = $args->getObject();
+        if (!$user instanceof UserInterface) {
+            return;
+        }
+
+        $commentsToUpdate = $args->getObjectManager()->getRepository(Comment::class)->findBy(['author' => $user]);
```
mmetayer: kind of risky if there are a lot of comments, but it's the only way I found to update the comments in the same transaction

src/Akeneo/Pim/Enrichment/Component/Comment/Model/CommentInterface.php
```diff
@@ -45,11 +45,11 @@ public function setResourceName($resourceName);
     public function getResourceName();
 
     /**
-     * @param UserInterface $author
+     * @param UserInterface|null $author
      *
      * @return CommentInterface
      */
-    public function setAuthor(UserInterface $author);
+    public function setAuthor(?UserInterface $author);
```
mmetayer: also a side effect, an ON DELETE SET NULL clause would avoid to modify the interface :/

src/Akeneo/Pim/Enrichment/Bundle/Doctrine/ORM/EventListener/RemoveUserListener.php
Doodoune: 3.1 is released so master I think

src/Akeneo/Pim/Enrichment/Bundle/Doctrine/ORM/EventListener/RemoveUserListener.php
ahocquard: why not a classic SQL request? Also, is it needed to keep the comments? What is the expected behavior in the UI? Could we not have a 'DELETED_USER' author name in that case (a kind of default user)?

Has it been discussed with PM?

mmetayer: I didn't use a classic SQL request because I need the entities here, in order to update them in the same transaction as the user removal one (if the user cannot be deleted for any other reason, we don't want the comments to be updated)

src/Akeneo/Pim/Enrichment/Component/Comment/Model/CommentInterface.php
Doodoune: add a todo merge also here I think

mmetayer: I don't think so, this is already a BC break, we probably don't want to add another one in a later branch to revert it.

src/Akeneo/Pim/Enrichment/Bundle/Doctrine/ORM/EventListener/RemoveUserListener.php
jmleroux: Comments are "manual data". We are not (yet) as big as Facebook and i don't think it's a problem right now.

src/Akeneo/Pim/Enrichment/Bundle/Doctrine/ORM/EventListener/RemoveUserListener.php
fitn: Don't write "master" please. Write the number of the future release.

src/Akeneo/Pim/Enrichment/Bundle/Resources/config/savers.yml
```diff
@@ -83,10 +83,3 @@ services:
             - '@doctrine.orm.entity_manager'
             - '@event_dispatcher'
             - '%pim_comment.entity.comment.class%'
-
-    pim_comment.saver.comment:
```
mmetayer: duplicate service definition

src/Akeneo/Pim/Enrichment/Component/Comment/Model/CommentInterface.php
mmetayer: I updated the PR, this is not relevant anymore

PR #10039, opened 2019-05-14
Fix changelog for PIM-8214

Doodoune:
cf https://github.com/akeneo/pim-community-dev/pull/9740

We removed 3 validators without adding BC breaks.


pchasle:
What about `src/Pim/Component/Connector/Validator/Constraints/FilterDataValidator.php` that was removed too ?

Doodoune:
True, I gonna add it too. thanks @pchasle !

PR #10052, opened 2019-05-17
EE contribution - Attribute option form errors and validator

Doodoune:
I created a custom attribute in order to override default simple select attribute type.

When I submitted my first option the controller returned a 400 without any messages. The problem was due to the fact that the method getFormErrors only return errors for child forms so I overrided it to also return global form errors.

I also used parameter in the AttributeTypeForOptionValidator class to handle supported attribute types like a lot of your services because otherwise, we have to override your class instead of just the parameter of your service.

(cf https://github.com/akeneo/pim-community-dev/pull/9780)

### DIFF
--- a/src/Pim/Component/Catalog/Validator/Constraints/AttributeTypeForOptionValidator.php
+++ b/src/Pim/Component/Catalog/Validator/Constraints/AttributeTypeForOptionValidator.php
@@ -16,18 +16,29 @@ use Symfony\Component\Validator\ConstraintValidator;
  */
 class AttributeTypeForOptionValidator extends ConstraintValidator
 {
+    /** @var array */
+    protected $supportedAttributeTypes;
+
     /**
-     * @param object     $attributeOption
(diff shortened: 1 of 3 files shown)


src/Pim/Component/Catalog/Validator/Constraints/AttributeTypeForOptionValidator.php
```diff
@@ -16,16 +16,36 @@
...
 class AttributeTypeForOptionValidator extends ConstraintValidator
 {
+    /** @var array */
+    protected $supportedAttributeTypes;
+
+    /**
+     * AttributeTypeForOptionValidator constructor.
+     * @param array $supportedAttributeTypes
+     *
+     * TODO on merge 3.2, remove = null
```
jmleroux:
```suggestion
     * TODO on merge 3.2, remove = null and add BC BREAK in changelog
```

PR #10064, opened 2019-05-21
PIM-8351: Fix entity overriding priority

pierallard:
* What was the issue

Today, you can override basic entities by using ConfigureMappingsSubscriber and
- defining a new class
- defining a `mapping_override`
- defining a new ORM mapping.

An entity can have mapping (many to one, one to many, etc) to Interfaces, as we have a compiler pass to get real classes instead of these Interfaces (ResolveTargetEntityListener)

The issue is that the compilation to transform Interfaces to real classes was done before the override of entity. This implies an error during hydratation because Interfaces entities were not found.

* What causes the issue 

This issue exists since we upgraded Symfony from 3.4.4 to 3.4.26. We didn't find anything in the Symfony Changelog related to this issue.

* What is the solution

The solution is to change the priority of the ConfigureMappingsSubscriber to be executed before all the other subscribers.

| Added legacy Behats               | n
| Added acceptance tests            | n

### DIFF
--- a/src/Pim/Bundle/CatalogBundle/Doctrine/ORM/QueryBuilderUtility.php
+++ b/src/Pim/Bundle/CatalogBundle/Doctrine/ORM/QueryBuilderUtility.php
@@ -33,7 +33,7 @@ class QueryBuilderUtility
     }
 
     /**
-     * @todo Merge master: Dead code, to be removed.
+     * @deprecated Will be removed in 4.0
      *
      * Replaces name of tables in DBAL queries
      *
--- a/src/Pim/Bundle/CatalogBundle/Doctrine/ORM/QueryBuilderUtility.php
+++ b/src/Pim/Bundle/CatalogBundle/Doctrine/ORM/QueryBuilderUtility.php
@@ -33,6 +33,8 @@ class QueryBuilderUtility
     }
 
     /**
+     * @todo Merge master: Dead code, to be removed.
+     *
      * Replaces name of tables in DBAL queries
      *
      * @param EntityManager $em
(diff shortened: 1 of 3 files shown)


src/Pim/Bundle/CatalogBundle/Doctrine/ORM/QueryBuilderUtility.php
```diff
@@ -33,6 +33,8 @@ public static function removeExtraParameters(QueryBuilder $qb)
     }
 
     /**
+     * @todo Merge master: Dead code, to be removed.
```
fitn: It's a dead code in our code but as it is public, it does not mean it's not used in projects. It is a BC Break to remove it. Anyway, I agree it should be deprecated asap and removed for the future 4.0

pierallard: That's the meaning of the comment, to remove it only with BC in a next version.

jmleroux:
```suggestion
     * @deprecated Will be removed in 4.0
```

src/Pim/Bundle/CatalogBundle/Doctrine/ORM/QueryBuilderUtility.php
```diff
@@ -70,6 +72,8 @@ public static function prepareDBALQuery($em, $entityName, $sql)
...
```
jmleroux:
```suggestion
     * @deprecated Will be removed in 4.0
```

PR #10067, opened 2019-05-21
PIM-8257: Fix user grid filter set when creating a new user

pierallard:
The bug is described in the code in comment as it's a little bit tricky.

| Added legacy Behats               | n
| Added acceptance tests            | n


src/Oro/Bundle/PimDataGridBundle/Resources/public/templates/filter/search-filter.html
```diff
@@ -1 +1 @@
-<input class="AknFilterBox-search" maxlength="255" autocomplete="off" type="text" name="value" value="<%- (typeof value !== 'undefined') ? value : '' %>" placeholder="<%- label %>">
+<input class="AknFilterBox-search" maxlength="255" autocomplete="non" type="text" name="value" value="<%- (typeof value !== 'undefined') ? value : '' %>" placeholder="<%- label %>">
```
pchasle: You should keep `autocomplete="off"` which is a valid value event of some browsers choose to ignore it.

pierallard: Every browser will disable autocomplete if the value is not "on".
Chrome will not disable autocomplete with password if this is "off".

PR #10070, opened 2019-05-22
PIM-8364: Fix bootstrap modal having illustration

LaurentPetard:
This will fix the modal of the proposals screen in EE. The class "AknFullPage-content--withIllustration" must be added every time there will be a "AknFullPage-illustration" child.

It reports what has been done on 3.1, particularly this : https://github.com/akeneo/pim-community-dev/commit/c845eb9694eb979045f37ca27b67dfe8eec409bd#diff-d9e6bfe307e86d918a0bb77e6f941ca6

### DIFF
--- a/CHANGELOG-3.0.md
+++ b/CHANGELOG-3.0.md
@@ -3,6 +3,7 @@
 ## Bug fixes
 
 - PIM-8257: Fix user grid filter set when creating a new user
+- PIM-8364: Fix bootstrap modal having illustration
 
 # 3.0.19 (2019-05-21)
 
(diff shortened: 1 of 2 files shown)


pierallard:
I'm really not sure of this fix.
Can you check every modal using illustrationClass to see if they are displayed correctly ?

LaurentPetard:
> I'm really not sure of this fix.
> Can you check every modal using illustrationClass to see if they are displayed correctly ?

It's the contrary. It impacts modals that don't use illustrationClass. I didn't find other modals like that but it's hard to do, so I maybe wrong. But it's working on 3.1 so I'm confident about that.

LaurentPetard:
> And some other modals in full page ? (like the basket of associations, and the columns selector)

I checked, those modals don't have illustrations (div with class "AknFullPage-illustration") (it seems that they aren't built from `bootstrap-modal.js` )

LaurentPetard:
Wrong fix. It can be done on EE for the proposal modals.

PR #10080, opened 2019-05-23
PIM-8236: Fix search issue after navigation between tabs

pierallard:
# The issue

- Go to a family with variant edit
- Go to family variant tab
- Search with the quick search filter ; it works
- Go to another tab
- Go back to the family variant tab
- The quick search filter does not work anymore

# The fix

There is 2 fixes:

- The first one is to add a delegateEvent on the filters because it was missing and the user typing events are not triggered
- The second one is to not re-render the grid if it was already rendered, to keep all the events of the current grid.

### DIFF
--- a/CHANGELOG-2.3.md
+++ b/CHANGELOG-2.3.md
@@ -3,6 +3,7 @@
 ## Bug fixes
 
 - PIM-8351: Fix entity overriding priority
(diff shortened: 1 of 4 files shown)


jmleroux:
Won't fix

PR #10091, opened 2019-05-24
Add required extensions to composer.json

jmleroux:
We have a dedicated Symfony requirements checker, but it would be more usefull to check extensions at composer level:

- early errors
- easier maintenance (mbstring for example was not in the SF requirements check)
- easy for infra guys (cloud team) to understand and fix

`OK`: Done / Validated
`-`: Not needed


ahocquard:
Is it not a problem when doing the `composer install` on different server from the production server?

jmleroux:
> Is it not a problem when doing the `composer install` on different server from the production server?

I don't understand the question.

jmleroux:
> Is it not a problem when doing the `composer install` on different server from the production server?

OK, i got your point. You're right, it can be a problem, but we don't run the symfony requirements check anyway. And if we want to make it, we can do it with composer: https://getcomposer.org/doc/03-cli.md#check-platform-reqs. The only edge case would be when the `composer.json` is not deployed on the prod server.

Good point, let's wait for other devs opinion on it.

BitOne:
Good point from @ahocquard . We could add something in the documentation to run the composer command to check requirements, but composer is not supposed to be on the production env.
So maybe a tradeoff would be to keep the composer requirements AND the Symfony requirements?

jmleroux:
> So maybe a tradeoff would be to keep the composer requirements AND the Symfony requirements?

Which would mean TWO list of extensions to maintain. That is the opposite of my primary goal with this PR.

BitOne:
According to your PR description, the primary goal is to catch errors early. So it's still valid.

PR #10099, opened 2019-05-24
PIM-7942: Fix tooltip minimal width

pierallard:
The default tooltip was too short and was unreadable.
We added a minimal width to prevent this.


src/Pim/Bundle/UIBundle/Resources/config/assets.yml
```diff
@@ -11,6 +11,7 @@ css:
     lib_customization:
         - bundles/pimui/less/lib/bootstrap.switch.less
         - bundles/pimui/less/lib/bootstrap.dropdown.less
+        - bundles/pimui/less/lib/bootstrap.tooltip.less
```
tamarasaurus: we have to make sure when it reaches master it's moved to the index.less file instead

PR #10118, opened 2019-05-28
PIM-8386: Fix price filter default currency

jmleroux:
Price filter default was not currently initialized.
When we render the criteria for the first time, we must use the first currency available.

`OK`: Done / Validated
`-`: Not needed


src/Oro/Bundle/PimDataGridBundle/Resources/public/js/datafilter/filter/price-filter.js
```diff
@@ -191,6 +191,13 @@ define(
                 e.preventDefault();
             },
 
+            /**
+             * {@inheritdoc}
```
pchasle: You should remove the `{@inheritdoc}` here

PR #10125, opened 2019-05-28
PIM-8381: do not expose disabled locales in attribute options export

rouliane:
Same fix as https://github.com/akeneo/pim-community-dev/pull/8877 : we don't expose disabled locales in attribute options standard format.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Pim/Component/Catalog/Normalizer/Standard/AttributeOptionNormalizer.php
+++ b/src/Pim/Component/Catalog/Normalizer/Standard/AttributeOptionNormalizer.php
@@ -2,6 +2,7 @@
 
 namespace Pim\Component\Catalog\Normalizer\Standard;
 
+use Akeneo\Component\StorageUtils\Repository\IdentifiableObjectRepositoryInterface;
 use Pim\Component\Catalog\Model\AttributeOptionInterface;
 use Symfony\Component\Serializer\Normalizer\NormalizerInterface;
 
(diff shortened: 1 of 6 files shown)


src/Pim/Component/Catalog/Normalizer/Standard/AttributeOptionNormalizer.php
```diff
@@ -49,6 +58,14 @@ protected function normalizeLabels(AttributeOptionInterface $attributeOption, $c
 
         foreach ($attributeOption->getOptionValues() as $translation) {
             if (empty($locales) || in_array($translation->getLocale(), $locales)) {
+                // TODO merge: remove null in master
+                if (null !== $this->localeRepository) {
+                    $locale = $this->localeRepository->findOneByIdentifier($translation->getLocale());
```
pierallard: Instead, can you do a localeRepository->findBy(['locales' => $locales, 'activated' => true])
to avoid n+1 calls ?

rouliane: I could but by doing this I won't be able to use the cached repository. So i'm not sure we would have any gain.

rouliane: So it means that when we export 1000 products, we will do 1000 calls to the dabase. With my way, we will call the database only the number of languages used in attribute option translations.

jmleroux: Agree for using cached data

anaelChardan: I do agree with julian, we should as Pierre said do only one query but the cache repository is not able to fetch multiple identifier at a time, and we don't have that much locale anyway ;)

Issue #9868, opened 2019-04-03
Unable to override Category entity on Akeneo 3 due to hard-coded form type


peleq-smile:
Hi,

I'm trying to override Category entity to add a new property, using Akeneo CE 3.0.10, and I have some troubles to do it. 

I'm following official documentation, more or less because I know it's not always up-to-date.

I did override Category class, that's OK, but override of the Category form is not possible.
That's logic because I've just noticed that the parameter `pim_enrich.form.type.category.class` doesn't exist anymore ! The used Form Type  (`Akeneo\Pim\Enrichment\Bundle\Form\Type\CategoryType`) **is now hard-coded** :-1:  into this two YAML files :
- `Akeneo/Pim/Enrichment/Bundle/Resources/config/form_types.yml`
- `Akeneo/Pim/Enrichment/Bundle/Resources/config/controllers.yml`

... whereas in Akeneo 2 it was OK !

There is another issue about extending Category entity : #9833, I have no idea if this is linked or not.

Could you please fix this very shortly ?

Rgds,
Perrine

peleq-smile:
Hi,

To go forward, I updated both files by using a parameter for the class (in vendor, please don't judge me so fast... this is just for a training not a real project ^^), to be able to override it and it works.

But after that I was unable to add an MediaType form element for my new "image" property of a category, but this is another problem. I guess it's possible but I just don't know how to do that because there is no other sample of MediaType property with the "old twig front" system into Akeneo :'

BR,
Perrine

fodie:
Hi @peleq-smile,
How did you resolve the issue? I am new to symfony

jdeschl:
Hi @peleq-smile,
I had the same issue with customized category entity when upgrading from 2.3.x to 3.0.x. 
As of now it doesn't work following the guide from the official docs https://docs.akeneo.com/3.0/manipulate_pim_data/category/add_new_properties_to_a_category.html#define-the-category-form. As a quick fix I decided to override the service definition of the controller within my own bundle (e.g. in Resources/config/controllers.yml) using the parameterized FormType class (for which you probably have an override in place - following the docs):

```
    pim_enrich.controller.category_tree.product:
        class: 'Akeneo\Pim\Permission\Bundle\Controller\Ui\CategoryTreeController'
        arguments:
            - '@event_dispatcher'
            - '@pim_user.context.user'
            - '@pim_catalog.saver.category'
            - '@pim_catalog.remover.category'
[...]

PR #10148, opened 2019-05-29
Proposal issue 9868

jdeschl:
fixes services definition of pim_enrich.controller.category_tree.product to use param-based form-type

https://github.com/akeneo/pim-community-dev/issues/9868


jdeschl:
As we use Akeneo EE, I opened a ticket on EE Helpdesk in addition to this PR. I got the response, that the direkt use of classes-names instead of parameters in the dependency config was a technical decision. Thats why this code change is not valid and I will close the PR.

jmleroux:
Alas, all the DI is not migrated to "hard coded classes" but we will indeed stop using classname as parameters. More info can be found on Symfony Best practices: https://symfony.com/doc/3.4/best_practices/business-logic.html#service-no-class-parameter

jmleroux:
I made a PR to fix the documentation accordingly.

PR #10149, opened 2019-05-29
TIP-1171: Add event for Identifier Updated

pierallard:


src/Akeneo/Pim/Enrichment/Component/Product/Model/AbstractProduct.php
```diff
@@ -276,11 +277,17 @@ public function getIdentifier()
      */
     public function setIdentifier(ValueInterface $identifier)
     {
+        $previousIdentifier = $this->identifier;
```
ahocquard: could you add a spec on product please. As the POC will be bigger and bigger, adding some tests will be useful.

ahocquard:
Is is sure we modify the identifier this way and by not by updating the values?

PR #10158, opened 2019-05-29
AOB-311: Add a new attribute header cell type

phaseinducer:
This PR allows to customize header cells in the datagrid.
It gets the header attribute label from the database instead of translation file.

**Usage :**
```
datagrid:
    my-grid:
        ...
        columns:
            some_attribute_column:
                ...
                headerCell: attribute
                attributeCode: some_attribute_code
```

**Note :**

We need to validate the use of `catalogLocale` vs `uiLocale`


src/Oro/Bundle/PimDataGridBundle/Resources/public/js/datagrid/header-cell/attribute-header-cell.js
```diff
@@ -0,0 +1,47 @@
...
+                    throw new Error('The option "attributeCode" must be set for a header cell of type "attribute".');
+                }
+
+                this.attributeCode = options.column.get('extraOptions').attributeCode;
+
+                HeaderCell.prototype.initialize.apply(this, arguments);
+            },
+
+            render: function () {
+                FetcherRegistry.getFetcher('attribute').fetch(this.attributeCode)
```
pierallard: If you can, please don't fetch the data into the render() method and move it in the configure() method

src/Oro/Bundle/PimDataGridBundle/Resources/public/js/datagrid/header-cell/attribute-header-cell.js
```diff
@@ -0,0 +1,47 @@
+define(
```
pierallard: Can you transform this file into .typescript and put a description on it ?

src/Oro/Bundle/PimDataGridBundle/Resources/public/js/datagrid/header-cell/attribute-header-cell.js
ValentinMerlet: You're right, we'll fix it.

src/Oro/Bundle/PimDataGridBundle/Resources/public/js/datagrid/header-cell/attribute-header-cell.js
ValentinMerlet: Sure

src/Oro/Bundle/PimDataGridBundle/Resources/public/js/datagrid/header-cell/attribute-header-cell.js
ValentinMerlet: `configure()` is not called however `HeaderCell` extends Backbone.View so it should be called. Any idea ?

phaseinducer: The cell views are not instantiated by our builder so `configure()` is never called.

PR #10176, opened 2019-06-04
PIM-8410: Remove unused controller class parameter

jmleroux:
`pim_enrich.controller.category_tree.class` is not used anymore.
The best practice is to use FQDN in DI and override the whole service if needed.

`OK`: Done / Validated
`-`: Not needed


jmleroux:
> If we really want to help our integrators to migrate easily, we should add sed command in upgrade file

You're totally right, but then i will not do in a patch

PR #10183, opened 2019-06-04
Proposition to solve conflict on changelogs

ahocquard:

**Problem:**
We have a lot of changelog conflicts. As soon as you have it and you want to merge your PR, you can do two things:
- you have to launch a new CI: 1 hour, 20 containers, for... well a line.
It blocks all the other users that want to use the CI for a useful usecase .
- ask an admin user (again and again)

**Solution:**
- generate the changelog from the merge. Hardly doable with the current CI or when you merge technical cards (refacto, very well splitted tasks, etc). A big topic.

- use in a file `.gitattributes` the `CHANGELOG merge=union` startegy. Still not supported by github, I tested.

- use a trick used by gitlab: add spaces in the changelog
https://about.gitlab.com/2015/02/10/gitlab-reduced-merge-conflicts-by-90-percent-with-changelog-placeholders/
Yes, it's kind of dirty, I'm aware of it. But not as much as launching our very costly tests for nothing. 

So I propose to use the last one as a workaround. We can then think about the perfect solution.
If approved, I propose to update michel-tag to delete useless line before a release.

`OK`: Done / Validated
`-`: Not needed


jjanvier:
I don't like it much for older versions, but I think it's really cool for master.
On my side, I looked at several tools. The most interesting I found was https://github.com/github-changelog-generator/github-changelog-generator.

ahocquard:
@jjanvier most of the time, the conflicts are for older version.
You don't add it that much on master (you can have a look at it). 
So, I think it's interesting for all branches. 

@jmleroux what do you think about it?

jmleroux:
Conflicts can be an issue but we don't have much right now because we are a small team and don't often conflict. But the idea is interesting and can be tested.
It will not solve my other issue which is how to have a correct changelog when pulling up fixes.

- If i fix and tag a bug in 2.3.66, it will be in the 2.3 changelog. => OK
- After pull up on 3.0, it will be in 2.3-changelog file => not so OK. We have to check multiple files to know if a bug is fixed on 3.0.
- After pull on 3.1, it will simply not be there because we removed the older changelogs => not OK at all

That's why i was thinking about having just one big changelog file.

That being said, this PR is an interesting step and open the discussion, so i can just :+1: :+1:

ahocquard:
Seen with JM: to test on master first. I will do PR for it.

PR #10214, opened 2019-06-12
PIM-8383: Do not take in account product without family when filtering IS_EMPTY + fix count on bulk actions

anaelChardan:

This PR makes that the products without family are not returned while we are filtering on IS_EMPTY;
It also fixes the bug on the counter for bulk actions.

Why there are so many files changed:
Because a lot of our tests are using products without family which is a use case we want to avoid now.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Akeneo/Pim/Enrichment/Bundle/Doctrine/ORM/Query/CountImpactedProducts.php
+++ b/src/Akeneo/Pim/Enrichment/Bundle/Doctrine/ORM/Query/CountImpactedProducts.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Akeneo\Pim\Enrichment\Bundle\Doctrine\ORM\Query;
 
+use Akeneo\Pim\Enrichment\Bundle\Elasticsearch\ProductAndProductModelSearchAggregator;
 use Akeneo\Pim\Enrichment\Component\Product\Model\ProductInterface;
 use Akeneo\Pim\Enrichment\Component\Product\Query\Filter\Operators;
 use Akeneo\Pim\Enrichment\Component\Product\Query\ProductQueryBuilderFactoryInterface;
@@ -23,10 +24,6 @@ class CountImpactedProducts
     /** @var ProductQueryBuilderFactoryInterface */
     private $productQueryBuilderFactory;
 
-    /**
-     * @param ProductQueryBuilderFactoryInterface $productAndProductModelQueryBuilderFactory
-     * @param ProductQueryBuilderFactoryInterface $productQueryBuilderFactory
-     */
     public function __construct(
         ProductQueryBuilderFactoryInterface $productAndProductModelQueryBuilderFactory,
         ProductQueryBuilderFactoryInterface $productQueryBuilderFactory
@@ -157,6 +154,27 @@ class CountImpactedProducts
 
         $pqb = $this->productAndProductModelQueryBuilderFactory->create(['filters' => $filters]);
 
+        $attributeCodesWithIsEmptyOperator = $this->getAttributeCodesWithIsEmptyOperator($pqb->getRawFilters());
+        if (!empty($attributeCodesWithIsEmptyOperator)) {
+            $pqb->getQueryBuilder()->addFilter([
+                    'bool' => [
+                        'should' => [
+                            [
+                                'terms' => [
+                                    'attributes_for_this_level' => $attributeCodesWithIsEmptyOperator
+                                ],
+                            ],
+                            [
+                                'terms' => [
+                                    'attributes_of_ancestors' => $attributeCodesWithIsEmptyOperator
+                                ],
+                            ]
+                        ]
+                    ]
+                ]
+            );
+        }
+
         return $pqb->execute()->count();
     }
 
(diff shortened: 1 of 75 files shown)


tests/back/Pim/Enrichment/Integration/PQB/Filter/Attribute/Date/DateFilterIntegration.php
```diff
@@ -3,6 +3,7 @@
 namespace AkeneoTest\Pim\Enrichment\Integration\PQB\Filter\Date;
 
 use Akeneo\Pim\Enrichment\Component\Product\Query\Filter\Operators;
+use Akeneo\Pim\Structure\Component\AttributeTypes;
```
pchasle: Unused import?

src/Akeneo/Pim/Enrichment/Bundle/Elasticsearch/Filter/Attribute/DateFilter.php
```diff
@@ -22,18 +23,18 @@ class DateFilter extends AbstractAttributeFilter implements AttributeFilterInter
...
-     * @param array $supportedAttributeTypes
-     * @param array $supportedOperators
-     */
+    
+    /** @var FamilyFilter */
+    private $familyFilter;
+    
     public function __construct(
         AttributeValidatorHelper $attrValidatorHelper,
+        FamilyFilter $familyFilter,
```
mmetayer: This is a big BC break !!

src/Akeneo/Pim/Enrichment/Bundle/Elasticsearch/Filter/Attribute/DateFilter.php
```diff
@@ -134,6 +135,8 @@ public function addAttributeFilter(
                 ];
 
                 $this->searchQueryBuilder->addMustNot($existsClause);
+                $this->familyFilter->setQueryBuilder($this->searchQueryBuilder);
```
mmetayer: Shouldn't this filter be added by the PQB itself? Same remark for every other AttributeFilter in the PR

src/Akeneo/Pim/Enrichment/Bundle/Doctrine/ORM/Query/CountImpactedProducts.php
```diff
@@ -157,6 +154,27 @@ private function countAllProducts(array $filters): int
 
         $pqb = $this->productAndProductModelQueryBuilderFactory->create(['filters' => $filters]);
 
+        $attributeCodesWithIsEmptyOperator = $this->getAttributeCodesWithIsEmptyOperator($pqb->getRawFilters());
+        if (!empty($attributeCodesWithIsEmptyOperator)) {
```
mmetayer: This misses an integration test

src/Akeneo/Pim/Enrichment/Bundle/Elasticsearch/Filter/Attribute/DateFilter.php
LaurentPetard: If the filter on family is applied for every attribute filter, it seems less bad to do it in the PQB instead of adding a new dependency in every attribute filter.

anaelChardan: It’s not all AttributeFilters it is only those which are supporting the IS_EMPTY operator (so not all of them). Moreover if customers has implemented other type of filters with the same usage of operator but with a different goal, we can’t apply a business rule for them imho.

src/Akeneo/Pim/Enrichment/Bundle/Elasticsearch/Filter/Attribute/DateFilter.php
anaelChardan: Yes it is but I don’t know if a lot of customers are extending our Filters

src/Akeneo/Pim/Enrichment/Bundle/Doctrine/ORM/Query/CountImpactedProducts.php
anaelChardan: Good catch

src/Akeneo/Pim/Enrichment/Bundle/Elasticsearch/Filter/Attribute/DateFilter.php
ahocquard: You should not do it if you don't know.

src/Akeneo/Pim/Enrichment/Bundle/Elasticsearch/Filter/Attribute/DateFilter.php
ahocquard: Both are good I think. The first solution avoid a lot of injection in a patch.

Also, keep in mind that we will remove it soon or later by removing optional values.

anaelChardan: Personnaly even in a patch I prefer to isolate the goal of the PR on each filters instead of handling a very special use case on the root of PQB

mmetayer: Finally I think your solution  is better and provides more flexibility, I just updated to avoid BC breaks

PR #10239, opened 2019-06-17
SDS-11088: allow graceful shutdown for job consumer daemon

shavounet:

When deploying Akeneo (using tools like Capistrano, etc...), restarting the application may crash a running job. If there is a long queue, there's no way to stop it for a few minutes while deploying. It's especially blocking for continuous deployment.

This PR allow the use of the SIGQUIT posix signal to tell a daemon process to stop between jobs.

Usage example : 
```bash
$ ps faux | grep job-queue-consumer-daemon
vacla    29191  0.5  0.4 408540 66396 ?        S    17:53   0:00  |       \_ php bin/console --env=prod akeneo:batch:job-queue-consumer-daemon
vacla    30779  0.4  0.4 408540 66712 ?        S    17:53   0:00  |       \_ php bin/console --env=prod akeneo:batch:job-queue-consumer-daemon
vacla    31190  0.4  0.4 408540 66716 ?        S    17:53   0:00          \_ php bin/console --env=prod akeneo:batch:job-queue-consumer-daemon

$ kill -s SIGQUIT  29191 30779 31190
```

What's missing:
* proper handling from `\Akeneo\Bundle\BatchQueueBundle\Queue\DatabaseJobExecutionQueue::consume`
  - for now the waiting loop stays here and is only interrupted when `sleep !== 0` (which is not always the case when receiving a signal...)
    - I would propose some similar `stopConsuming` method on the interface
  - when it happens, an error occur due to the return type not matching method's signature : `Return value of Akeneo\Bundle\BatchQueueBundle\Queue\DatabaseJobExecutionQueue::consume() must be an instance of Akeneo\Component\BatchQueue\Queue\JobExecutionMessage, null returned`
    - note that this last error is only a verbosity problem and do not have an impact on jobs or functionnalities
* (for PaaS) allow to call something like `partners_supervisor graceful_shutdown` which would send the signal to each running daemon (or timeout ?), and do not restart

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -32,6 +32,7 @@
     },
     "require": {
         "php": "7.1.*",
+        "ext-pcntl": "*",
         "ass/xmlsecurity": "1.1.1",
         "doctrine/annotations": "1.3.1",
         "doctrine/cache": "1.6.1",
--- a/src/Akeneo/Bundle/BatchQueueBundle/Command/JobQueueConsumerCommand.php
+++ b/src/Akeneo/Bundle/BatchQueueBundle/Command/JobQueueConsumerCommand.php
@@ -41,6 +41,9 @@ class JobQueueConsumerCommand extends ContainerAwareCommand
     /** Interval in microseconds before checking if the process is still running. */
     private const RUNNING_PROCESS_CHECK_INTERVAL = 200000;
 
+    /** Flag used to stop the daemon */
+    private $shouldStop = false;
+
     /**
      * {@inheritdoc}
      */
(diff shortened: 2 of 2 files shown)


rybus:
Hello @shavounet !

Thanks for your contribution. This matter has already been discussed last March https://github.com/akeneo/pim-community-dev/pull/9752.

This PR was closed and the reasons why are explained. Also, a possible way to deal with it is provided. I hope you will find proper guidance there.

I want to thank you again for contributing to the PIM and I am sorry we cannot go any further in the process.

Best

PR #10248, opened 2019-06-19
TIP-1183: Updating the mapping and adding the doc

anaelChardan:

Add docs for elasticsearch mapping update

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/CHANGELOG-3.2.md
+++ b/CHANGELOG-3.2.md
@@ -6,11 +6,13 @@
 
 ## BC Breaks
 
+- The `Akeneo\Pim\Enrichment\Component\Product\Model\ValueCollectionInterface` interface has been renamed into `Akeneo\Pim\Enrichment\Component\Product\Model\WriteValueCollectionInterface`
 - Rename `Akeneo\Pim\Enrichment\Bundle\Elasticsearch\Filter\Field\AncestorFilter` to `Akeneo\Pim\Enrichment\Bundle\Elasticsearch\Filter\Field\AncestorIdFilter` 
-
  - DAPI-137: Fix the PQB to not aggregate results when there is a filter on id
 
+## Enhancements
 
+- The search on option.codes in ES is now case insensible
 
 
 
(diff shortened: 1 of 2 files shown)


CHANGELOG-3.2.md
```diff
@@ -6,8 +6,8 @@
 
 ## BC Breaks
 
+- The ValueCollection interface has been renamed into WriteValueCollectionInterface please apply `find ./src/ -type f -print0 | xargs -0 sed -i 's/ValueCollectionInterface/WriteValueCollectionInterface/g`
```
pierallard: Please set complete namespace and remove sed

UPGRADE.md
```diff
@@ -169,6 +169,16 @@ Before updating the dependencies and migrating your data, please deactivate all
     yarn install
     ```
 
+10. Migrate your Elasticsearch indices:
+
+    In case you updated the settings of Elasticsearch (like normalizers, filters and analyzers), please make sure you properly loaded your custom settings in the [Elasticsearch configuration](https://github.com/akeneo/pim-enterprise-standard/blob/3.1/app/config/pim_parameters.yml#L58-L68).
```
pierallard: I think we have to "force them" to run it because we changed it.

CHANGELOG-3.2.md
ahocquard: maybe we should add in improvement what we did about case insentivity of option in ES?

UPGRADE.md
ahocquard: We don't know why though. You can add a sentence:

- ES does not take in account case insensitivity of option codes when searching. As we modified the way products values are loaded from Mysql, ESresearch has to be case insensitive when searching on option codes.

CHANGELOG-3.2.md
anaelChardan: I'm adding it on the upgrade.

PR #10255, opened 2019-06-21
PIM-8462: Fix memory leak in ComputeFamilyVariantStructureChangesTasklet

willy-ahva:

We use PQB instead of repository to have pagination.
We use the cache clearer.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Pim/Component/Catalog/Job/ComputeFamilyVariantStructureChangesTasklet.php
+++ b/src/Pim/Component/Catalog/Job/ComputeFamilyVariantStructureChangesTasklet.php
@@ -5,15 +5,19 @@ declare(strict_types=1);
 namespace Pim\Component\Catalog\Job;
 
 use Akeneo\Component\Batch\Model\StepExecution;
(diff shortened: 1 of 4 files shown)


src/Pim/Component/Catalog/Job/ComputeFamilyVariantStructureChangesTasklet.php
```diff
@@ -103,61 +104,102 @@ public function execute()
...
+     * @param CursorInterface $cursor
      */
-    private function updateValuesOfEntities(array $entities): void
+    private function updateValuesOfEntities(CursorInterface $cursor): void
     {
-        $this->keepOnlyValuesForVariation->updateEntitiesWithFamilyVariant($entities);
+        $cursor->rewind();
+        $products = [];
+        $productModels = [];
+        while ($cursor->valid()) {
```
mmetayer: Just out of curiosity, why not use a `foreach` here instead of manually calling `rewind()`, `next()`, `valid()`...?

PR #10272, opened 2019-06-25
TIP-1190: Disable the wait of the refresh of the index when indexin in ES, by default

ahocquard:

We should not `wait_for` index refresh when indexing products in ES.
It means it wait the next refresh from ES, which is every 1 sec by default. 

Most of the time, it is disabled but not everywhere. It is not a best practice to use it, as it impacts performance.
So by default, we should disable it (it's disable by default in ES if not provided explicitely).

On Icecat catalog, we have a performance gain of at least 10 seconds during the installation (1m45 => 1m35).
We have a gain of 1 minute during integration tests also, for free.

On association step during product import, as the batch size is at 1 during import for this step, we were waiting most of the time a refresh in ES on icecat catalog (several calls per product with refresh, 2 hours on this step, I don't know how it's used in production...).
  

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/tests/back/Integration/IntegrationTestsBundle/Loader/FixturesLoader.php
+++ b/tests/back/Integration/IntegrationTestsBundle/Loader/FixturesLoader.php
@@ -6,6 +6,7 @@ namespace Akeneo\Test\IntegrationTestsBundle\Loader;
 
 use Akeneo\Test\Integration\Configuration;
 use Akeneo\Test\IntegrationTestsBundle\Security\SystemUserAuthenticator;
+use Akeneo\Tool\Bundle\ElasticsearchBundle\Refresh;
 use Symfony\Bundle\FrameworkBundle\Console\Application;
 use Symfony\Component\Console\Input\ArrayInput;
 use Symfony\Component\Console\Output\BufferedOutput;
@@ -86,7 +87,6 @@ class FixturesLoader implements FixturesLoaderInterface
         $fixturesHash = $this->getHashForFiles($files);
 
         $dumpFile = sys_get_temp_dir().self::CACHE_DIR.$fixturesHash.'.sql';
-
         if (file_exists($dumpFile)) {
             $this->databaseSchemaHandler->reset();
 
@@ -98,6 +98,7 @@ class FixturesLoader implements FixturesLoaderInterface
             $this->indexProductModels();
             $this->indexProducts();
 
+            $this->refreshES();
             return;
         }
 
@@ -105,6 +106,8 @@ class FixturesLoader implements FixturesLoaderInterface
         $this->clearAclCache();
 
         $this->loadData();
+        $this->refreshES();
+
         $this->dumpDatabase($dumpFile);
 
         $this->systemUserAuthenticator->createSystemUser();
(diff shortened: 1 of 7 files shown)


tests/back/Integration/IntegrationTestsBundle/Loader/FixturesLoader.php
```diff
@@ -379,4 +388,12 @@ protected function resetElasticsearchIndex(): void
             $client->resetIndex();
         }
     }
+
+    private function refreshES(): void
+    {
+        $this->container->get('akeneo_elasticsearch.client.product')->refreshIndex();
```
mmetayer: maybe you could use `$this->get('akeneo_elasticsearch.registry.clients')->getClients()` to list clients?

tests/back/Pim/Enrichment/Integration/Storage/Sql/ProductGrid/ProductGridFixturesLoader.php
```diff
@@ -313,4 +320,11 @@ private function createFamilyVariant(): void
         Assert::assertCount(0, $errors);
         $this->container->get('pim_catalog.saver.family_variant')->save($familyVariant);
     }
+
+    private function refreshEsIndex(): void
+    {
+        $this->container->get('akeneo_elasticsearch.client.product')->refreshIndex();
```
mmetayer: same question

tests/back/Pim/Enrichment/Specification/Bundle/Elasticsearch/Indexer/ProductIndexerSpec.php
```diff
@@ -149,7 +149,7 @@ function it_bulk_deletes_products_from_elasticsearch_index($productIndexClient,
         $this->removeAll([40, 33])->shouldReturn(null);
     }
 
-    function it_indexes_products_and_wait_for_index_refresh_by_default(
+    function it_indexes_products_and_wait_for_index_refresh(
```
mmetayer:
```suggestion
    function it_indexes_products_and_waits_for_index_refresh(
```

tests/back/Pim/Enrichment/Specification/Bundle/Elasticsearch/Indexer/ProductIndexerSpec.php
```diff
@@ -177,10 +177,10 @@ function it_indexes_products_and_wait_for_index_refresh_by_default(
             ['id' => 'bar', 'a key' => 'another value'],
         ], 'id', Refresh::waitFor())->shouldBeCalled();
 
-        $this->indexAll([$product1, $product2]);
+        $this->indexAll([$product1, $product2], ['index_refresh' => Refresh::waitFor()]);
     }
 
-    function it_indexes_products_and_disable_index_refresh(
+    function it_indexes_products_and_disable_index_refresh_by_default(
```
mmetayer:
```suggestion
    function it_indexes_products_and_disables_index_refresh_by_default(
```

tests/back/Pim/Enrichment/Specification/Bundle/Elasticsearch/Indexer/ProductModelIndexerSpec.php
```diff
@@ -174,7 +174,7 @@ function it_bulk_deletes_product_models_from_elasticsearch_index($productModelCl
         $this->removeAll([40, 33])->shouldReturn(null);
     }
 
-    function it_indexes_product_models_and_wait_for_index_refresh_by_default(
+    function it_indexes_product_models_and_disable_refresh_of_the_index_by_default(
```
mmetayer:
```suggestion
    function it_indexes_product_models_and_disables_refresh_of_the_index_by_default(
```

tests/back/Integration/IntegrationTestsBundle/Loader/FixturesLoader.php
ahocquard: It's done on purpose as we are not doing anything on draft and proposal. It would impact badly the performance of the integration tests as it adds 0,5 sec of waiting time in average per index.

When we will merge indexes, it will be even better (it has a real impact).

tests/back/Pim/Enrichment/Integration/Storage/Sql/ProductGrid/ProductGridFixturesLoader.php
ahocquard: same response, not needed for draft and proposal and ref entities, etc

PR #10283, opened 2019-06-26
DAPI: make catalog volume monitoring section more generic

willy-ahva:

Make catalog volume monitoring section more generic.
From Franklin context, we would like to use exactly the same section as catalog volume monitoring. Js is already generic, but not the naming.
This PR proposes to move files and rename some class and less.

`OK`: Done / Validated
`-`: Not needed


ahocquard:
Before doing this kind of PR, you should check with the PM what is plan in the future.
If it's nothing special:
- you can make it generic
- otherwise, don't do it

tests/front/acceptance/cucumber/decorators/catalog-volume/volume.decorator.js
```diff
@@ -7,9 +7,9 @@ const Volume = async (nodeElement) => {
   };
 
   const getValue = async () => {
-    const value = await nodeElement.$('.AknCatalogVolume-value');
-    const mean = await value.$('.AknCatalogVolume-value span:nth-child(1) div');
-    const max = await value.$('.AknCatalogVolume-value span:nth-child(2) div');
+    const value = await nodeElement.$('.AknShowFigures-value');
```
ahocquard: why don't we copy/past? 

Is is that important to make it generic? Generic = coupling.
I don't want to break Franklin by modifying something in the catalog volume monitoring. We are already struggling with our legacy, please stop to do it for new stuff...

mmetayer: I agree with Alex here, I have no problem with renaming modules in `pim/form/common`, but I feel it would be more future-proof to duplicate the ones in `pim/catalog-volume/` in case the needs evolve differently for the 2 contexts

PR #10303, opened 2019-07-01
PIM-8480: Remove the job execution message orphans after a job execution purge

LaurentPetard:

The old job executions can be deleted from the database with the command `akeneo:batch:purge-job-execution`. The problem is that it doesn't delete the messages related to these job executions in the queue.

To fix this problem I added a query function that deletes job execution message orphans, and I call it in the purge command. I think that it's better to keep the queue system decoupled instead of adding a foreign key with a "delete cascade" constraint.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Akeneo/Bundle/BatchBundle/Command/PurgeJobExecutionCommand.php
+++ b/src/Akeneo/Bundle/BatchBundle/Command/PurgeJobExecutionCommand.php
@@ -50,6 +50,9 @@ class PurgeJobExecutionCommand extends ContainerAwareCommand
             $this->getJobExecutionRepository()->remove($jobsExecutions);
             $output->write(sprintf("%s jobs execution deleted ...\n", count($jobsExecutions)));
         }
+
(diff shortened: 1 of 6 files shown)


src/Akeneo/Bundle/BatchBundle/Command/PurgeJobExecutionCommand.php
```diff
@@ -50,6 +50,9 @@ protected function execute(InputInterface $input, OutputInterface $output)
             $this->getJobExecutionRepository()->remove($jobsExecutions);
             $output->write(sprintf("%s jobs execution deleted ...\n", count($jobsExecutions)));
         }
+
+        // Purge the message queue, even if there was no job executions to remove this time.
```
jmleroux: Not necessary.

LaurentPetard:
CI green before pushing the last commit that just remove a comment line.

PR #10332, opened 2019-07-05
Fixes object storage configuration

BitOne:

Fixes configuration issues that created problems with the new version of the S3 library.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/.circleci/local-object-storage-setup.sh
+++ b/.circleci/local-object-storage-setup.sh
@@ -12,7 +12,7 @@ services:
         factory_method: factory
         arguments:
             -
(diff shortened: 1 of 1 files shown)


jjanvier:
but where does that problem come from?

BitOne:
@jjanvier : the problem came from the configuration, but was not visible before, as the S3 library ignored this part of the configuration. But a change in this library made it stricter so it now fails because of this configuration.
And this library has been updated because of the changes we've made in composer.json to validate the right PHP extension, that needed to regenerate a fresher composer.lock.

PR #10395, opened 2019-07-18
[POC 1] DO NOT MERGE Upgrade to SF4

jjanvier:
DO NOT MERGE THIS PR

This is just a POC to help identify the difficulties we'll meet when upgrading to Symfony 4.

My goal was to be able to:
- search in the products grid
- update a product via the PEF

I've taken some shortcuts to make it work. The only important thing is what I learnt from this POC.

The biggest difficulty we'll encounter when upgrading is [about the DI](https://symfony.com/doc/current/service_container/3.3-di-changes.html). We need to find a smart way to do it.


src/Oro/Bundle/DataGridBundle/Resources/config/services.yml
```diff
@@ -47,12 +50,13 @@ services:
             - { name: oro_datagrid.configuration.provider, priority: -10 }
 
     oro_datagrid.configuration.provider.chain:
+        public: true
         class: '%oro_datagrid.configuration.provider.chain.class%'
 
     oro_datagrid.datagrid.metadata_parser:
```
christophe-chausseray: It doesn't need to be public also ?

jjanvier: my goal with this POC was just to update a product via the UI and display the product grid
apparently for those 2 operations, this service doesn't need to be public

PR #10397, opened 2019-07-18
Tip 1159 simplify normalizer

pierallard:
The internal ProductCompletenessCollection Normalizer was too complex because it loads a lot of useless things just for getting labels.
We extracted it into 2 queries to only channel and attribute labels.


src/Akeneo/Pim/Enrichment/Component/Product/Normalizer/InternalApi/ProductCompletenessCollectionNormalizer.php
```diff
@@ -181,78 +234,86 @@ private function average(array $completenesses)
...
+    }
+
+    /**
+     * @param string $localeCode
+     *
+     * @return string|null
+     */
+    private function getLocaleName(string $localeCode): ?string
+    {
+        $locale = new Locale();
```
mmetayer: Maybe you could directly use `\Locale::getDisplayName($localeCode)` here? It would avoid instantiating a Locale object (and avoid coupling :) )

PR #10406, opened 2019-07-19
TIP-1159: Adapter to save product completenesses

mmetayer:
The goal of this PR is to move the completeness calculation out of the Product saver (here, in an event subscriber, but it could be elsewhere as well), and use the new SaveProductCompletenessQuery to save it (i.e not relying on ORM anymore in order to save the completeness)

`OK`: Done / Validated
`-`: Not needed


src/Akeneo/Pim/Enrichment/Bundle/EventSubscriber/ComputeProductCompletenessSubscriber.php
```diff
@@ -0,0 +1,79 @@
...
+
+    public function computeProductCompleteness(GenericEvent $event): void
+    {
+        $product = $event->getSubject();
+
+        if (!$product instanceof ProductInterface) {
+            return;
+        }
+
+        $completenesses = $this->completenessCalculator->calculate($product);
```
SamirBoulil: Maybe it's too hard to do in your PR, but wouldn't it be easier if the completenessCalculator returns a ProductCompletenessCollection directly ?

pierallard: Yeah Samir but it's not the goal of this PR for now, but it's planned, don't worry :)

tests/back/Pim/Enrichment/Integration/Completeness/CompletenessFloorRoundedRatioIntegration.php
```diff
@@ -50,7 +50,8 @@ public function testFloorRoundedRatio()
 
         // the real ratio would 1/3 = 33.33333...%
         // here we want to floor to 33%
-        $this->assertEquals(33, $firstProduct->getCompletenesses()->first()->getRatio());
+        $completeness = $this->getProductCompletenesses()->fromProductId($firstProduct->getId())->getIterator()->current();
```
ahocquard: mhhh we have a lot of tests that we could change to be unit test, basically this one

src/Akeneo/Pim/Enrichment/Bundle/EventSubscriber/ComputeProductCompletenessSubscriber.php
```diff
@@ -0,0 +1,79 @@
...
+use Akeneo\Tool\Component\StorageUtils\StorageEvents;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\EventDispatcher\GenericEvent;
+
+/**
+ * @author    Mathias METAYER <mathias.metayer@akeneo.com>
+ * @copyright 2019 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+final class ComputeProductCompletenessSubscriber implements EventSubscriberInterface
```
ahocquard: `ComputeAndPersistProductCompletenessSubscriber` 

Personally, I never know the difference between compuete and calculate, if it persists or not

src/Akeneo/Pim/Enrichment/Bundle/EventSubscriber/ComputeProductCompletenessSubscriber.php
```diff
@@ -0,0 +1,79 @@
...
+use Akeneo\Pim\Enrichment\Component\Product\Model\Projection\ProductCompleteness;
+use Akeneo\Pim\Enrichment\Component\Product\Model\Projection\ProductCompletenessCollection;
+use Akeneo\Pim\Enrichment\Component\Product\Query\SaveProductCompletenesses;
+use Akeneo\Pim\Structure\Component\Model\AttributeInterface;
+use Akeneo\Tool\Component\StorageUtils\StorageEvents;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\EventDispatcher\GenericEvent;
+
+/**
+ * @author    Mathias METAYER <mathias.metayer@akeneo.com>
```
ahocquard: This subscriber should be executed before the subscriber to index the products In Elasticsearch, as the completeness is also indexed.

src/Akeneo/Pim/Enrichment/Bundle/EventSubscriber/ComputeProductCompletenessSubscriber.php
mmetayer: :+1: I actually struggled to find a good naming for this class, your solution is better

src/Akeneo/Pim/Enrichment/Bundle/EventSubscriber/ComputeProductCompletenessSubscriber.php
ahocquard: Good comment. Basically, we are detaching the completeness from the product for now (doctrine mapping, etc).
Once it's done, we will be free to do what you suggested (which is an adapter) and then improve the algorithm (which means changing the adapter implementation).

PR #10429, opened 2019-07-23
TIP-1198 Don't need to check if web/bundle exist

arnolanglade:

This PR optimizes the class autoloading: we don't need to load classes in production if we only use them for test and there were some duplicated loading rules too.

Some post scripts are removed because we don't need them. Our install command do the same job, it is a pity to do two times the same job to install the PIM. We will save a little bit time to build our PIM :)

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -86,6 +86,9 @@ jobs:
                 name: Start containers
                 command: docker-compose up -d fpm
           - run:
+                name: Generate production environment cache
+                command: docker-compose exec fpm bin/console cache:clear -e dev
+          - run:
                 name: PhpSpec
                 command: docker-compose exec -T fpm vendor/bin/phpspec run --format=junit > var/tests/phpspec/specs.xml
           - run:
--- a/composer.json
+++ b/composer.json
@@ -15,7 +15,10 @@
         "psr-4": {
             "Pim\\Upgrade\\": "upgrades/"
         },
-        "classmap": [ "app/AppKernel.php", "app/AppCache.php" ]
+        "classmap": [
+            "app/AppKernel.php",
+            "app/AppCache.php"
+        ]
     },
     "autoload-dev": {
         "psr-0": {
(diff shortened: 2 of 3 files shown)


.circleci/config.yml
```diff
@@ -85,6 +85,9 @@ jobs:
           - run:
                 name: Start containers
                 command: docker-compose up -d fpm
+          - run:
+                name: Generate production environment cache
+                command: docker-compose exec fpm bin/console cache:clear -e dev
```
skeleton: dev env?

mmetayer: yes, it's for phpstan

arnolanglade: Phpstan need a an xml file, this file is only available in dev env. you should have a look [that](https://github.com/phpstan/phpstan-symfony#configuration) to have more info.

skeleton: so you should change the name from "Generate production environment cache" to "Generate dev environment cache"

composer.json
```diff
@@ -25,9 +28,7 @@
             "Pim\\Behat\\": "tests/legacy/features/Behat",
             "AkeneoTest\\": "tests/back",
             "Akeneo\\Test\\IntegrationTestsBundle\\": "tests/back/Integration/IntegrationTestsBundle/",
-            "Akeneo\\Test\\Integration\\": "tests/back/Integration/",
```
anaelChardan: Wasn't it needed?

.circleci/config.yml
anaelChardan: Same comment as @skeleton

arnolanglade: ah ok, didn't see that! If the CI is green I will merge this PR but I will change that in my next PR.

ahocquard:
I have one question: is the merge with the standard planned in the future?

Generally, when you release in production, you rely on composer to initialize all the codebase you need on a dedicated server.
It's possible to not have access to composer anymore in a production environment., but only an archive.

If you prevent to use composer for the installation, it means that you can't prepare the codebase by only using composer, without having a database up, an ES up, etc...

For dev environment, we don't care, but for the install of the PIM in the future, I'm not sure. When we migrate a standard version, we don't call `pim:install`.

arnolanglade:
First we plan to remove standard edition, we want to generate them. You should have a look to https://github.com/akeneo/pim-enterprise-dev/pull/6532
Then, you don't run composer in production but you need to package your application before.
Finally, We will work our tools to easily migrate because we plan to deliver the next major every month.

PR #10431, opened 2019-07-23
TIP-901: no direct access to container

jjanvier:
In Symfony 4, we cannot access it from the container anymore.

1. Doctrine circular dependencies

Sometimes, the container was injected to avoid circular dependencies with Doctrine subscriber. 
It's recommended to use event listener instead. It can load lazily the event listener in that case, which avoid circular dependencies. 

It's also recommended for performance reason.
https://symfony.com/doc/current/doctrine/event_listeners_subscribers.html#performance-considerations

2. Remove of the dead code

Some Oro classes were using the container. These classes were not used sometime.
I removed them and a lot of other Oro classes not used.

Source: https://ahocquard.github.io/CE_coverage/Oro/Bundle/DataGridBundle/index.html

3. What's left?
All the command that are container aware should be declared as services.
As there are more than 20 commands to change, it will be done in a dedicated PR.

Also, unfortunately, 4 classes in CE still use the container with logic to get services according to a parameter passed in to the function.
It's not possible to avoid the use of the container in these services in an easy way.
We will have to declare them public when migrating to SF4.

`OK`: Done / Validated
`-`: Not needed


src/Akeneo/UserManagement/Bundle/Resources/config/event_subscribers.yml
```diff
@@ -5,16 +5,20 @@ parameters:
     pim_user.event_subscriber.remove_role.class:      Akeneo\UserManagement\Bundle\EventListener\RemoveRoleSubscriber
     # Move this class
     pim_locale.locale_subscriber.class:               Akeneo\UserManagement\Bundle\EventListener\LocaleSubscriber
-    pim_locale.locale_listener.class:                 Pim\Bundle\UserBundle\EventListener\LocaleListener
     pim_user.event_listener.create_user_system.class: Akeneo\UserManagement\Bundle\EventListener\CreateUserSystemListener
 
 services:
     pim_user.event_subscriber.user_preferences:
```
jjanvier: should the class/service be renamed as user preferences listener?

ahocquard: I was lazy. I did it :smile:

tests/back/UserManagement/Specification/Bundle/EventListener/UserPreferencesListenerSpec.php
```diff
@@ -3,46 +3,39 @@
...
 use Akeneo\Channel\Component\Model\ChannelInterface;
-use Akeneo\Channel\Component\Model\LocaleInterface;
 use Akeneo\Channel\Component\Repository\ChannelRepositoryInterface;
+use Akeneo\Channel\Component\Repository\LocaleRepositoryInterface;
 use Akeneo\Tool\Component\Classification\Model\CategoryInterface;
 use Akeneo\Tool\Component\Classification\Repository\CategoryRepositoryInterface;
-use Akeneo\UserManagement\Bundle\EventListener\UserPreferencesSubscriber;
-use Akeneo\UserManagement\Bundle\Manager\UserManager;use Akeneo\UserManagement\Component\Model\UserInterface;use Akeneo\UserManagement\Component\Repository\UserRepositoryInterface;use Doctrine\Common\EventSubscriber;
+use Akeneo\UserManagement\Bundle\EventListener\UserPreferencesListener;
+use Akeneo\UserManagement\Component\Model\UserInterface;use Akeneo\UserManagement\Component\Repository\UserRepositoryInterface;
```
LeoBenoist: Two use on the same line

jjanvier:
GTM

PR #10432, opened 2019-07-23
TIP-901: update dependencies

jjanvier:
This PR simply bumps a few dependencies to the last patches.
This is necessary in order to be able to upgrade Symfony to the version 4.

See https://github.com/akeneo/pim-community-dev/pull/10395

### DIFF
--- a/app/AppKernel.php
+++ b/app/AppKernel.php
@@ -34,7 +34,6 @@ class AppKernel extends Kernel
 
         if (in_array($this->getEnvironment(), array('dev', 'test', 'behat'))) {
             $bundles[] = new Sensio\Bundle\DistributionBundle\SensioDistributionBundle();
-            $bundles[] = new Sensio\Bundle\GeneratorBundle\SensioGeneratorBundle();
             $bundles[] = new Symfony\Bundle\DebugBundle\DebugBundle();
             $bundles[] = new Symfony\Bundle\WebProfilerBundle\WebProfilerBundle();
             $bundles[] = new Symfony\Bundle\WebServerBundle\WebServerBundle();
--- a/app/AppKernel.php
+++ b/app/AppKernel.php
@@ -33,7 +33,6 @@ class AppKernel extends Kernel
         $bundles = $this->registerProjectBundles();
 
         if (in_array($this->getEnvironment(), array('dev', 'test', 'behat'))) {
-            $bundles[] = new Sensio\Bundle\DistributionBundle\SensioDistributionBundle();
             $bundles[] = new Symfony\Bundle\DebugBundle\DebugBundle();
             $bundles[] = new Symfony\Bundle\WebProfilerBundle\WebProfilerBundle();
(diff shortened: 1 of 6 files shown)


app/AppKernel.php
```diff
@@ -33,8 +33,6 @@ public function registerBundles()
         $bundles = $this->registerProjectBundles();
 
         if (in_array($this->getEnvironment(), array('dev', 'test', 'behat'))) {
-            $bundles[] = new Sensio\Bundle\DistributionBundle\SensioDistributionBundle();
```
arnolanglade: Are you sure? is it needed to generate bootstrap.php.cache ?

app/AppKernel.php
```diff
@@ -33,8 +33,6 @@ public function registerBundles()
...
-            $bundles[] = new Sensio\Bundle\GeneratorBundle\SensioGeneratorBundle();
```
arnolanglade: :+1:

composer.json
```diff
@@ -50,38 +50,36 @@
...
-        "sensio/framework-extra-bundle": "3.0.26",
-        "sensio/generator-bundle": "3.1.6",
+        "sensio/framework-extra-bundle": "5.2.4",
         "symfony/monolog-bundle": "3.1.2",
-        "symfony/swiftmailer-bundle": "3.0.3",
-        "symfony/security-acl": "3.0.0",
+        "symfony/swiftmailer-bundle": "3.2.4",
+        "symfony/security-acl": "3.0.2",
         "symfony/symfony": "3.4.28",
         "symfony/thanks": "^1.0",
```
arnolanglade: We should remove this dependencies

app/AppKernel.php
jjanvier: @arnolanglade this bundle is not compatible with Symfony 4

jjanvier: cf https://github.com/sensiolabs/SensioDistributionBundle/blob/master/composer.json

arnolanglade: yes but do we need to until dumping Symfony to V4?

composer.json
jjanvier: I'll do in another PR.
The CI is green for now

app/AppKernel.php
jjanvier: what is used `bootstrap.php.cache` for? 
if the CI is green, then it's OK no?

composer.json
```diff
@@ -50,38 +50,36 @@
         "doctrine/cache": "1.8.0",
         "doctrine/collections": "1.5.0",
         "doctrine/common": "2.10.0",
-        "doctrine/data-fixtures": "1.2.2",
```
fitn: Do we still use it? Same for doctrine-fixtures-bundle?
That's not directly linked to your PR anyway ;)

jjanvier: I'll check !
and change in another PR if needed

PR #10459, opened 2019-07-25
TIP-901: make controllers public

jjanvier:
With Symfony4, all services are private by default. Including the controllers. Which means, if you update the PIM to Symfony4, you'll have this kind of error:

`[2019-07-25 19:21:36] request.CRITICAL: Uncaught PHP Exception InvalidArgumentException: "Controller "pim_user.controller.security_rest" cannot be fetched from the container because it is private. Did you forget to tag the service with "controller.service_arguments"?" at /srv/pim/vendor/symfony/symfony/src/Symfony/Component/HttpKernel/Controller/ContainerControllerResolver.php line 71 {"exception":"[object] (InvalidArgumentException(code: 0): Controller \"pim_user.controller.security_rest\" cannot be fetched from the container because it is private. Did you forget to tag the service with \"controller.service_arguments\"? at /srv/pim/vendor/symfony/symfony/src/Symfony/Component/HttpKernel/Controller/ContainerControllerResolver.php:71, Error(code: 0): Class 'pim_user.controller.security_rest' not found at /srv/pim/vendor/symfony/symfony/src/Symfony/Component/HttpKernel/Controller/ControllerResolver.php:133)"} []`

We have to make controllers public in order to work with Symfony4.

### DIFF
--- a/src/Akeneo/Channel/Bundle/Resources/config/controllers.yml
+++ b/src/Akeneo/Channel/Bundle/Resources/config/controllers.yml
@@ -1,6 +1,7 @@
 services:
 
     pim_enrich.controller.rest.locale:
+        public: true
         class: 'Akeneo\Channel\Bundle\Controller\InternalApi\LocaleController'
         arguments:
             - '@pim_catalog.repository.locale'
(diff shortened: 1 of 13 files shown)


jjanvier:
@fitn why do you want to remove controllers as services? What do you want exactly?

PR #10465, opened 2019-07-26
TIP-901: Small fixes to ease SF4 support

jjanvier:
This PR fixes 3 things needed to upgrade to Symfony 4:
- sets the missing security provider for API firewall
- fixes a few BC breaks
- removes the (useless) `Message` class. Sometimes, we passed a `Message` to `ControllerTrait::addFlash`. It was OK because the method was not type-hinted. [Now it accepts only a string](https://github.com/symfony/symfony/blob/4.0/src/Symfony/Bundle/FrameworkBundle/Controller/ControllerTrait.php#L153).

`OK`: Done / Validated
`-`: Not needed


ahocquard:
There are BC break. I don't think it's extended though...

PR #10470, opened 2019-07-29
Fix remove orphan archives

fitn:
The remove of the job directory deletes files with existing jobs execution (so that should not be removed).
It comes from the fact we use job instance code instead of job name. The directory name is based on the job name not on the job instance code.
So, if the job instance code is not the same than the job name, it removes the directory.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/upgrades/schema/remove_orphan_archives.php
+++ b/upgrades/schema/remove_orphan_archives.php
@@ -86,12 +86,11 @@ SQL;
 function removeOrphanJobDirectories(Connection $database, string $archiveDir): void
 {
     $existingJobsWithJobExecutions = <<<SQL
-SELECT ji.job_name
+SELECT DISTINCT ji.job_name
 FROM akeneo_batch_job_execution je 
 INNER JOIN akeneo_batch_job_instance ji 
     ON ji.id = je.job_instance_id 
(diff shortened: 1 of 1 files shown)


upgrades/schema/remove_orphan_archives.php
```diff
@@ -86,11 +86,12 @@ function removeOrphanJobExecutionDirectories(Connection $database, string $archi
...
 {
     $existingJobsWithJobExecutions = <<<SQL
-SELECT ji.code
+SELECT ji.job_name
 FROM akeneo_batch_job_execution je 
 INNER JOIN akeneo_batch_job_instance ji 
     ON ji.id = je.job_instance_id 
-WHERE ji.type IN ('import', 'export');
+WHERE ji.type IN ('import', 'export')
+GROUP BY ji.job_name;
```
ahocquard: DISTINCT, there is no aggregation function (tell what you want)

I said during it the training :smile:

fitn: Haha I was not there :p

fitn: It's ok :)

PR #10478, opened 2019-07-29
TIP-1196: make Behat env closer to prod's one

jjanvier:
The goal is still to make the enterprise edition (and our tests) closer to what we have in production.
I removed all the conf that was not necessary in our Behat env.

### DIFF
--- a/app/config/config_behat.yml
+++ b/app/config/config_behat.yml
@@ -3,39 +3,9 @@ imports:
 
 framework:
     test: ~
-    csrf_protection: true
-
-doctrine:
-    orm:
-        entity_managers:
-            default:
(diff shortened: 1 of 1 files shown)


ahocquard:
Btw, I don't know what is doing the key:
```
pim_import_export:
    record_mails: true
```

But the only class using this config seems a test class not used anymore (dead code). I saw that but I don't really know what is doing. If you have clue...

anaelChardan:
@alex maybe it was only for asset notifications as user config but still it’s weird 😂

jjanvier:
it used here https://github.com/akeneo/pim-enterprise-dev/pull/6600#discussion_r309460224

PR #10483, opened 2019-07-30
TIP-1192 Make the symfony kernel compatible with SF4

jjanvier:
For more information, please read http://fabien.potencier.org/symfony4-directory-structure.html and https://symfony.com/doc/current/bundles.html.

This is another step forward SF4 (and maybe Flex).

Next move is to remove those `AppKernelTest` classes :)

### DIFF
--- a/app/AppKernel.php
+++ b/app/AppKernel.php
@@ -1,7 +1,12 @@
 <?php
 
+declare(strict_types=1);
+
+use Symfony\Bundle\FrameworkBundle\Kernel\MicroKernelTrait;
 use Symfony\Component\Config\Loader\LoaderInterface;
 use Symfony\Component\HttpKernel\Kernel;
+use Symfony\Component\Routing\RouteCollectionBuilder;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
 
 /**
  * PIM AppKernel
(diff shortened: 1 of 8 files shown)


app/AppKernel.php
```diff
@@ -12,194 +15,58 @@
...
     }
 
     /**
-     * @return string
+     * {@inheritdoc}
      */
     public function getLogDir(): string
     {
-        return dirname(__DIR__) . DIRECTORY_SEPARATOR . 'var' . DIRECTORY_SEPARATOR . 'logs';
+        return $this->getProjectDir().'/var/logs';
```
ahocquard: `DIRECTORY_SEPARATOR`?

anaelChardan: Don’t know if it is useful as we only target linux.

app/AppKernel.php
```diff
@@ -12,194 +15,58 @@
...
-            new Acme\Bundle\AppBundle\AcmeAppBundle(),
-        ];
-    }
+    use MicroKernelTrait;
 
     /**
      * {@inheritdoc}
      */
-    public function registerBundles()
+    public function registerBundles(): iterable
```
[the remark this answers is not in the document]
arnolanglade: you should do that in EE too :)

app/AppKernel.php
arnolanglade: Symfony does not use them neither!

tests/back/Integration/AppKernelTest.php
```diff
@@ -12,41 +10,14 @@
...
     /**
-     * {@inheritdoc}
+     * {@inheritDoc}
+     *
+     * Necessary to make gedmo extension tree work. Otherwise the path located in
+     * "vendor/akeneo/pim-community-dev/src/Akeneo/Platform/config/bundles/gedmo_doctrine_extensions.yml"
+     * is never the right one...
      */
-    protected function registerProjectBundles(): array
+    public function getRootDir()
```
christophe-chausseray:
```suggestion
    public function getRootDir(): string
```

PR #10485, opened 2019-07-30
PIM-8601: Fix purge of the job execution according to the date of creation and not deletion

ahocquard:

When you purge the job executions, some job execution are not purged because it failed.
The `end_time` is not set and the query does not delete jobs without `end_time`.

We could remove jobs without `end_time`, but it means that a job executed today that failed would be purged also, whatever the parameter of the command (older than X days).

The most simple thing is just to be based on the date of creation.

**Cons:**
As a job could be put in the queue and executed only some hours later, I forced to purge at a minimum of 1 day.
I don't think there is any usecase at allowing 0 day. 

It means also if a job last 24 hours you can have a problem if you execute the command with "older than 1 day".
But I don't think that's a good practice to purge jobs older than 1 days and also and if a job last 24 hours, we have bigger problem.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Akeneo/Platform/Bundle/ImportExportBundle/Command/PurgeJobExecutionCommand.php
+++ b/src/Akeneo/Platform/Bundle/ImportExportBundle/Command/PurgeJobExecutionCommand.php
@@ -26,7 +26,7 @@ class PurgeJobExecutionCommand extends ContainerAwareCommand
     {
         $this->setName('akeneo:batch:purge-job-execution');
         $this->setDescription(
-            'Purge jobs execution older than number of days you want except the last one, by default 90 days'
+            'Purge jobs execution older than number of days you want except the last one, by default 90 days, minimum is 1 day'
         );
         $this->addOption(
             'days',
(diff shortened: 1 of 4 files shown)


src/Akeneo/Platform/Bundle/ImportExportBundle/Command/PurgeJobExecutionCommand.php
```diff
@@ -43,9 +43,9 @@ protected function configure()
     protected function execute(InputInterface $input, OutputInterface $output)
     {
         $days = $input->getOption('days');
-        if (!is_numeric($days)) {
+        if (!(is_numeric($days) && $days > 0)) {
             $output->writeln(
-                sprintf('<error>Option --days must be a number, "%s" given.</error>', $input->getOption('days'))
+                sprintf('<error>Option --days must be a number strictly superior to 0, "%s" given.</error>', $input->getOption('days'))
```
jmleroux: > It means also if a job last 24 hours you can have a problem if you execute the command with "older than 1 day".
> But I don't think that's a good practice to purge jobs older than 1 days and also and if a job last 24 hours, we have bigger problem.

If it's not a good practice and could generate problems, why not asking for a parameter > 1 ?

ahocquard: You still have the problem: instead of 24 hours, it's 48 hours. I just choosed a limit. 24 hours is reasonable I think.

ahocquard:
@fitn updated date is not filled when it is not started for any reason

LeoBenoist:
https://akeneo.atlassian.net/projects/SDS/issue/SDS-11582

PR #10495, opened 2019-07-31
PIM-8603: Sort attribute columns by code if no order is defined

jmleroux:
Add a sort on attribute code to have a constant order even when all attributes order and attribute groups order are set to 0.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/CHANGELOG-3.1.md
+++ b/CHANGELOG-3.1.md
@@ -2,7 +2,7 @@
 
 ## Bug fixes
 
-- PIM-8603: Sort attribute columns by code if no order is defined
+- PIM-8603: Sort attribute columns by label if no order is defined
 
 # 3.1.17 (2019-07-30)
 
--- a/CHANGELOG-3.1.md
+++ b/CHANGELOG-3.1.md
@@ -1,5 +1,9 @@
 # 3.1.x
 
+## Bug fixes
+
+- PIM-8603: Sort attribute columns by code if no order is defined
+
 # 3.1.17 (2019-07-30)
 
 ## Bug fixes
--- a/src/Oro/Bundle/PimDataGridBundle/Query/Sql/ListProductGridAvailableColumns.php
+++ b/src/Oro/Bundle/PimDataGridBundle/Query/Sql/ListProductGridAvailableColumns.php
@@ -129,7 +129,7 @@ FROM pim_catalog_attribute AS att
 INNER JOIN pim_catalog_attribute_group AS g ON att.group_id = g.id
 LEFT JOIN pim_catalog_attribute_translation AS att_trans ON att.id = att_trans.foreign_key AND att_trans.locale = :locale
 WHERE att.useable_as_grid_filter = 1 AND att.code NOT IN (:attributesToExclude) $sqlSearch
-ORDER BY g.sort_order ASC, att.sort_order ASC, att.code ASC
+ORDER BY g.sort_order ASC, att.sort_order ASC, label ASC
 LIMIT $limit OFFSET $offset
 SQL;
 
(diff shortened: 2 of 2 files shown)


jmleroux:
> Ok but why not ordering by labels instead of codes?

I would say perfs and because it has not be asked till now. But i can make it on labels.

Doodoune:
You can order by label, as it's already specify in the select
-> COALESCE(att_trans.label, CONCAT('[', att.code, ']')) AS label

So you can add an "order by label"

src/Oro/Bundle/PimDataGridBundle/Query/Sql/ListProductGridAvailableColumns.php
```diff
@@ -129,7 +129,7 @@ private function fetchAttributesAsColumn(string $locale, int $limit, int $offset
 INNER JOIN pim_catalog_attribute_group AS g ON att.group_id = g.id
 LEFT JOIN pim_catalog_attribute_translation AS att_trans ON att.id = att_trans.foreign_key AND att_trans.locale = :locale
 WHERE att.useable_as_grid_filter = 1 AND att.code NOT IN (:attributesToExclude) $sqlSearch
-ORDER BY g.sort_order ASC, att.sort_order ASC
+ORDER BY g.sort_order ASC, att.sort_order ASC, att.code ASC
```
Doodoune:
```suggestion
ORDER BY g.sort_order ASC, att.sort_order ASC, label ASC
```

CHANGELOG-3.1.md
```diff
@@ -1,5 +1,9 @@
 # 3.1.x
 
+## Bug fixes
+
+- PIM-8603: Sort attribute columns by code if no order is defined
```
Doodoune:
```suggestion
- PIM-8603: Sort attribute columns by label if no order is defined
```

src/Oro/Bundle/PimDataGridBundle/Query/Sql/ListProductGridAvailableColumns.php
```diff
@@ -129,7 +129,7 @@ private function fetchAttributesAsColumn(string $locale, int $limit, int $offset
 INNER JOIN pim_catalog_attribute_group AS g ON att.group_id = g.id
 LEFT JOIN pim_catalog_attribute_translation AS att_trans ON att.id = att_trans.foreign_key AND att_trans.locale = :locale
 WHERE att.useable_as_grid_filter = 1 AND att.code NOT IN (:attributesToExclude) $sqlSearch
-ORDER BY g.sort_order ASC, att.sort_order ASC
+ORDER BY g.sort_order ASC, att.sort_order ASC, label ASC
```
ahocquard: is it not mandatory to have an order?

Doodoune: seems that no... this problem came from a Saas user ^^'

ahocquard: It is mandatory: the correct solution is to determine why it happens.

Doodoune:
It's a ticket raised by a Saas user, so they don't have access to the database. 
=> https://akeneo.atlassian.net/browse/SDS-11429

src/Oro/Bundle/PimDataGridBundle/Query/Sql/ListProductGridAvailableColumns.php
Doodoune: g.sort_order and att.sort_order are are all equals to 0 in the client case

jmleroux: I think it's possible to have the order at 0 if you create your attributes by import and do not edit them via the UI

PR #10497, opened 2019-07-31
TIP-1193 Run behat on the right env using makefile

arnolanglade:
This PR adds a comment to explain a little trick on the CI and it fixes makefile to run behat on the right env.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -188,7 +188,7 @@ jobs:
     steps:
       - attach_workspace:
           at: ~/
-        # The version of docker compose used by circle ci does not support -`e` option (`docker-composer exec -e ENV_VAR=value`) to set environment variables to containers
+        # The version of docker compose used by circle ci does not support `-e` option (`docker-composer exec -e ENV_VAR=value`) to set environment variables to containers.
         # To load .env.behat (app/bootstrap.php) we need to set APP_ENV=behat but it is not possible because of the version of docker-composer.
         # To avoid to upgrade it, we use this little trick
       - run:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -188,6 +188,9 @@ jobs:
     steps:
       - attach_workspace:
           at: ~/
+        # The version of docker compose used by circle ci does not support -`e` option (`docker-composer exec -e ENV_VAR=value`) to set environment variables to containers
```
jjanvier:
```suggestion
        # The version of docker compose used by circle ci does not support `-e` option (`docker-composer exec -e ENV_VAR=value`) to set environment variables to containers.
```

.circleci/config.yml
```diff
@@ -188,6 +188,9 @@ jobs:
...
+        # To load .env.behat (app/bootstrap.php) we need to set APP_ENV=behat but it is not possible because of the version of docker-composer.
```
jjanvier:
```suggestion
        # To load .env.behat (from app/bootstrap.php) we need to set APP_ENV=behat but it is not possible because of the version of docker-compose.
```

.circleci/config.yml
```diff
@@ -188,6 +188,9 @@ jobs:
     steps:
       - attach_workspace:
           at: ~/
+        # The version of docker compose used by circle ci does not support `-e` option (`docker-composer exec -e ENV_VAR=value`) to set environment variables to containers.
```
jmleroux: And what about setting `APP_ENV` directly in Circle CI?
It's an environment var after all.

PR #10498, opened 2019-07-31
TIP-1193 Add default values to docker-composer.yml

arnolanglade:

To avoid this kind of problem when you up your containers without a `.env` file, default env vars values have been added .

```
jjanvier:pcd$ docker-compose up -d                                                                       
WARNING: The APP_DATABASE_PASSWORD variable is not set. Defaulting to a blank string.
WARNING: The APP_DATABASE_USER variable is not set. Defaulting to a blank string.
WARNING: The APP_DATABASE_NAME variable is not set. Defaulting to a blank string.
Recreating pcd_mysql-behat_1 ... 
Recreating pcd_mysql_1 ... 
Starting pcd_node_1 ... 
pcd_fpm_1 is up-to-date
pcd_selenium_1 is up-to-date
pcd_object-storage_1 is up-to-date
pcd_elasticsearch_1 is up-to-date
Recreating pcd_httpd-behat_1 ... 
Recreating pcd_mysql_1 ... done
```

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -62,9 +62,9 @@ services:
     image: 'mysql:5.7'
     environment:
       MYSQL_ROOT_PASSWORD: 'root'
-      MYSQL_USER: '${APP_DATABASE_USER}'
-      MYSQL_DATABASE: '${APP_DATABASE_NAME}'
-      MYSQL_PASSWORD: '${APP_DATABASE_PASSWORD}'
+      MYSQL_USER: '${APP_DATABASE_USER:-akeneo_pim}'
+      MYSQL_DATABASE: '${APP_DATABASE_NAME:-akeneo_pim}'
+      MYSQL_PASSWORD: '${APP_DATABASE_PASSWORD:-akeneo_pim}'
     networks:
       - 'akeneo'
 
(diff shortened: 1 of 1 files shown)


arnolanglade:
- `.env` file is not mandatory
- Everyone does not use make
- Onboarder works this way
- It is the step have a versioned `.env` and stick to symfony convention.

jmleroux:
> Why don't we just force the user to have .env

>  It's already its job to have values for your environment.

I agree with Alex... kind of. We should force the user to have values for his environment.
Whether he is using `.env` or not, `make` or not.
`.env` is a convenient way of having a working environment and if the user don't want to use it, it is his responsibility to build the env.

> It is the step have a versioned .env and stick to symfony convention.

That's why i'm divided :smiley:

arnolanglade:
> .env is a convenient

arnolanglade:
`.env` is not convenient for docker-composer because all `DOCKER_PORT_HTTP*` are useless because you just need to add default values : 8080/8081/8082 for the EE and 8090/8091/8092 for the CE. We have `BLACKFIRE_*` env vars which are used by a single team if I correctly understand. It seems they are not used by the application nor the docker-composer.

ahocquard:
-     .env file is not mandatory

Yes, that's just for development. Our CI should not use it btw.

-     Everyone does not use make

This is another topic, but it's not linked to my comment. 

-     Onboarder works this way

It does not mean they are doing it correctly. You should not have 2 distinct ways of setting your environment variables.

```
We have BLACKFIRE_* env vars which are used by a single team if I correctly understand. It seems they are not used by the application nor the docker-composer.
```

It's used to test performance of the PIM. It is used by a single team for now, but I hope it will change. It is used by the application 


We had the opportunity to have a common way to define environment variable, but it's now read in three ways:
- variables defined in ".env" and read by the application and docker-compose (should not occur in a CI environment, `.env` files should be delete)
- variables exported (should be done in production and so, in our CI)
- variables defined in our docker-compose

Anyway, we disagree and you have your 2 GTM. We will see in the future at which point it will be tricky to detect bugs due to all this configuration and overloading. We have already one bug in our standard about it.

PR #10501, opened 2019-07-31
[POC2] DO NOT MERGE Upgrade to SF4

jjanvier:
The first POC made [here](https://github.com/akeneo/pim-community-dev/pull/10395) allowed to me determine that:
- updating dependencies was complex
- we called a lot of services directly from the container, which Symfony4 does not allow anymore 

All those problems have been solved by the PRs https://github.com/akeneo/pim-community-dev/pull/10432 and https://github.com/akeneo/pim-community-dev/pull/10431. Other minor [PRs also have been merged](https://github.com/akeneo/pim-community-dev/pulls?q=is%3Apr+is%3Aclosed+label%3A%22towards+SF4%22) to facilitate the upgrade.

This new POC goes a step further:
1. it upgrades to Symfony4 libraries
2. it introduces Flex
3. it makes the `config` directory structure compatible with Symfony4 (and Flex). Please note that I didn't completely removed the `app` folder in this POC. But it should be done in the real PR.

Work/do not work:
- :heavy_check_mark:  product grid works (no need to hack the controllers or some missing public services anymore)
- :heavy_check_mark: PEF works (no need to hack the controllers or some missing public services anymore)
- :red_circle: the PIM is not yet installable via `bin/console pim:install`. It's a pity because I'd like to launch a CI ASAP and see how many tests are broken.

Lessons learnt:
- changing the directory structure is really complex. It takes patience and strictness. I don't know what is the best way we could help our customers and integrators to move to Symfony4. Any idea is welcome.
- about commands : 
   - they are services like others, which means they should be declared as public services (same as controllers)
   - they access a lot the container directly :( At some point we'll have to refactor them. And maybe start by deleting the ones we judge useless. 
- we still have a lot of work to make our PIM compatible with Symfony4 without any dirty hack

Coming next:
- I'll open a PR that will facilitate the upgrade with a lot of "little things" done in this POC
- delete useless commands
- fix Symfony4 BC breaks
- I'll focus on making the command `bin/console pim:install` so that I can launch a CI


config/packages/prod/doctrine.yaml
```diff
@@ -0,0 +1,10 @@
+doctrine:
+    orm:
+        entity_managers:
+            default:
+                metadata_cache_driver:
+                    cache_provider: pim
```
ahocquard: things to do after symfony 4 update: profiling the PIM to detect if we still use everything properly. I remember I did some things about these cache drivers that were not used in 2.0.

config/packages/swiftmailer.yaml
```diff
@@ -0,0 +1,3 @@
+swiftmailer:
+    url: '%env(MAILER_URL)%'
+    spool: { type: 'memory' }
```
ahocquard: what is that?

config/packages/prod/doctrine.yaml
jjanvier: @ahocquard indeed
and the default configuration provided by SF4 is not the same
maybe we have a few things to dig in

config/packages/swiftmailer.yaml
jjanvier: no idea
it was there before
it's still in the default SF4 configuration, so I let it

config/bundles.php
```diff
@@ -56,9 +57,8 @@
     Akeneo\Tool\Bundle\VersioningBundle\AkeneoVersioningBundle::class => ['all' => true],
 
     // Dev related bundles
-    Symfony\Bundle\DebugBundle\DebugBundle::class => ['dev' => true, 'test' => true, 'behat' => true],
     Symfony\Bundle\WebProfilerBundle\WebProfilerBundle::class => ['dev' => true, 'test' => true, 'behat' => true],
-    Symfony\Bundle\WebServerBundle\WebServerBundle::class => ['dev' => true, 'test' => true, 'behat' => true],
+    Symfony\Bundle\MakerBundle\MakerBundle::class => ['dev' => true],
```
anaelChardan: Do we need it?

config/packages/swiftmailer.yaml
anaelChardan: The spool is something that retains email instead of sending them and another process reads the spool in order to send them. https://symfony.com/doc/4.1/email/spool.html

config/services.yaml
```diff
@@ -84,3 +87,26 @@ parameters:
     count_product_model_values_limit: -1
     count_product_and_product_model_values_limit: -1
     count_users_limit: -1
+
+    pim_ce_dev_src_folder_location: '%kernel.project_dir%'
+
+services:
```
anaelChardan: :+1:

PR #10505, opened 2019-08-01
TIP-901: Small fixes towards SF4

jjanvier:
Easier to read commit by commit. 

The kernel is now located in `src/Kernel.php` for Symfony4.
ACL support is removed from Symfony. We need an extra `AclBundle`.

### DIFF
--- a/app/bundles.php
+++ b/app/bundles.php
@@ -10,6 +10,7 @@ return [
     Symfony\Bundle\SecurityBundle\SecurityBundle::class => ['all' => true],
     Symfony\Bundle\SwiftmailerBundle\SwiftmailerBundle::class => ['all' => true],
     Symfony\Bundle\TwigBundle\TwigBundle::class => ['all' => true],
+    Symfony\Bundle\AclBundle\AclBundle::class => ['all' => true],
 
     // ORO dependencies
     Doctrine\Bundle\FixturesBundle\DoctrineFixturesBundle::class => ['all' => true],
--- a/app/config/pim_parameters.yml
+++ b/app/config/pim_parameters.yml
@@ -16,9 +16,9 @@ parameters:
     websocket_host:       "127.0.0.1"
     websocket_port:       8080
 
-    upload_dir:           '%kernel.root_dir%/uploads/product'
-    archive_dir:          '%kernel.root_dir%/archive'
-    catalog_storage_dir:  '%kernel.root_dir%/file_storage/catalog'
+    upload_dir:           '%kernel.project_dir%/app/uploads/product'
+    archive_dir:          '%kernel.project_dir%/app/archive'
+    catalog_storage_dir:  '%kernel.project_dir%/app/file_storage/catalog'
     tmp_storage_dir:      '/tmp/pim/file_storage'
     upload_tmp_dir:       '/tmp/pim/upload_tmp_dir'
 
--- a/app/config/security.yml
+++ b/app/config/security.yml
@@ -28,11 +28,13 @@ security:
             pattern:                        ^/user/(login|reset-request|send-email|check-email)$
             provider:                       chain_provider
             anonymous:                      true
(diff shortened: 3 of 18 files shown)


src/Akeneo/Tool/Bundle/ElasticsearchBundle/tests/integration/UpdateIndexMappingIntegration.php
```diff
@@ -16,7 +16,7 @@ class UpdateIndexMappingIntegration extends TestCase
 {
     public function test_it_moves_the_index_and_the_pqb_still_works()
     {
-        $indexHost = $this->getParameter('index_hosts');
+        $indexHost = getenv('APP_INDEX_HOSTS');
```
AymericPlanche: :warning: this will not work in the future, in the new version of dotenv component the variables defined in .env are not set as real system environment variables. https://github.com/symfony/dotenv/blob/master/CHANGELOG.md#500 up to discuss about this IRL ;)

jjanvier: @AymericPlanche but isn't `.env` just a "hack"? shouldn't we use real env vars, like in production? I'm quite lost about what we should or should not do about env vars...

I can revert this commit, no problem btw

anaelChardan: You could set the parameter index host by fetching the env directly in the config file no?

app/config/config.yml
```diff
@@ -52,5 +52,3 @@ akeneo_pim_structure:
 
 parameters:
     pim_ce_dev_src_folder_location: '%kernel.project_dir%'
-    index_hosts: '%env(APP_INDEX_HOSTS)%'
```
anaelChardan: I would let it like that

src/Akeneo/Tool/Bundle/ElasticsearchBundle/tests/integration/UpdateIndexMappingIntegration.php
AymericPlanche: First: Why do we want to use environment variable based configuration value ?
-> Cloud deployment is based on production docker images. We don't want to generate and maintain configuration files that could be mounted as volumes in docker images when we have to deploy them in production. 
The cloud way for application configuration is environment variables everywhere, the one used in production are real / system environment variables, if you do a `export` command in a production container, you'll be able to see the environment variables.

-> Regarding the dev environment(s), as we can/have to manage multiple application instances in one laptop, it's easier to be able to have a `.env` file that will be `loaded` by the `symfony\DotEnv` component within the application instead of exporting env var each time we need to switch from a ee-3.0 and a ee-3.1 for instance.

to sum up: the fact that we manage default values via a `.env` file is great for dev/ci environment and allow to be production ready, but it does not mean that the application need to read the value directly from the `getenv()` php method, we need to have the configuration keys in configuration yaml files, the value will be dynamic using `%env(...)%` syntax.

jjanvier: OK, I'll revert this commit
thanks @AymericPlanche

ahocquard: > but it does not mean that the application need to read the value directly from the getenv()

In my opinion it means that the code will work in production but not in development...

AymericPlanche: I think it's really important that everyone have to be convinced about this subject, maybe we could make a presentation to all the team. cc @jjanvier

PR #10508, opened 2019-08-02
PIM-8611: Fix filters applied when a product model is excluded from a product-grid selection

LaurentPetard:
There's a bug when a user selects all the products in a filtered product-grid to export them, but un-select manually one (or several) product-model (by clicking on the checkbox). 
The export will well exclude the product-model, but will add its variants.

It's because in this case, the ES query is built with the filters of the product-grid chosen by the user, and will add a filter to exclude the id of the products that have been unselected. In the case of a product-model, it results that the variants of the excluded product-model match the ES query, when they should be excluded.

To fix that I added a filter on the field "ancestor.id" to exclude the variants of the excluded product-models.
To do that I had to change the `AncestorFilter` application, to remove a clause on the existence of the field "ancestor" to not exclude the "simple" products when the operator `NOT IN` is used for this filter. I don't know why this clause was here, especially as it's not applied with the operator `IN`

### DIFF
--- a/src/Akeneo/Pim/Enrichment/Bundle/Doctrine/ORM/Repository/ProductMassActionRepository.php
+++ b/src/Akeneo/Pim/Enrichment/Bundle/Doctrine/ORM/Repository/ProductMassActionRepository.php
@@ -42,6 +42,13 @@ class ProductMassActionRepository implements ProductMassActionRepositoryInterfac
         if (!empty($values)) {
             $condition = $inset ? Operators::IN_LIST : Operators::NOT_IN_LIST;
             $queryBuilder->addFilter('id', $condition, $values);
+
+            $productModelIds = array_values(array_filter($values, function ($id) {
+                return 0 === strpos($id, 'product_model_');
+            }));
+            if (false === $inset && !empty($productModelIds)) {
+                $queryBuilder->addFilter('ancestor.id', Operators::NOT_IN_LIST, $productModelIds);
+            }
         }
     }
 
--- a/src/Akeneo/Pim/Enrichment/Bundle/Doctrine/ORM/Repository/ProductMassActionRepository.php
+++ b/src/Akeneo/Pim/Enrichment/Bundle/Doctrine/ORM/Repository/ProductMassActionRepository.php
@@ -4,6 +4,7 @@ namespace Akeneo\Pim\Enrichment\Bundle\Doctrine\ORM\Repository;
 
 use Akeneo\Pim\Enrichment\Component\Product\Query\Filter\Operators;
 use Akeneo\Pim\Enrichment\Component\Product\Query\ProductQueryBuilder;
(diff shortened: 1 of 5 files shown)


src/Akeneo/Pim/Enrichment/Bundle/Elasticsearch/Filter/Field/AncestorFilter.php
```diff
@@ -86,13 +86,7 @@ public function addFieldFilter($field, $operator, $values, $locale = null, $chan
                         self::ANCESTOR_ID_ES_FIELD => $values,
                     ],
                 ];
-                $filterClause = [
-                    'exists' => [
-                        'field' => self::ANCESTOR_ID_ES_FIELD,
```
LaurentPetard: In my opinion, this clause was illogic for the operator "NOT_IN_LIST", it could make sense for the operator "IN_LIST", but it's not applied for it. I didn't find any other case that uses this filter with the operator "NOT_IN_LIST"

src/Akeneo/Pim/Enrichment/Bundle/Doctrine/ORM/Repository/ProductMassActionRepository.php
```diff
@@ -42,6 +42,13 @@ public function applyMassActionParameters($queryBuilder, $inset, array $values)
         if (!empty($values)) {
             $condition = $inset ? Operators::IN_LIST : Operators::NOT_IN_LIST;
             $queryBuilder->addFilter('id', $condition, $values);
+
+            $productModelIds = array_values(array_filter($values, function ($id) {
```
ahocquard: you can add a private function if you think it can improve the "why"

src/Akeneo/Pim/Enrichment/Bundle/Doctrine/ORM/Repository/ProductMassActionRepository.php
```diff
@@ -42,6 +42,13 @@ public function applyMassActionParameters($queryBuilder, $inset, array $values)
...
+                return 0 === strpos($id, 'product_model_');
+            }));
+            if (false === $inset && !empty($productModelIds)) {
+                $queryBuilder->addFilter('ancestor.id', Operators::NOT_IN_LIST, $productModelIds);
```
ahocquard: does it work correctly with both condition (in list, not in list)?

LaurentPetard: In the case of `false === $inset` it's necessarly a "not in list" condition

LaurentPetard: I re-worked the methods to clarify

PR #10512, opened 2019-08-02
PIM-8614: Fix empty variant axes validation

LaurentPetard:
The PHP function `empty()` returns `true` for the value `'0'`, but it's a valid value for an option code, and so for an axis value.


src/Akeneo/Pim/Enrichment/Component/Product/Validator/Constraints/NotEmptyVariantAxesValidator.php
```diff
@@ -63,7 +63,7 @@ public function validate($entity, Constraint $constraint)
             $isEmptyMetricValue = (null !== $value && $value->getData() instanceof MetricInterface &&
                 null === $value->getData()->getData());
 
-            if ((null === $value || (empty($value->getData()) && !is_bool($value->getData()))) || $isEmptyMetricValue) {
```
pierallard: Why removing the is_bool ?

LaurentPetard: Because it was here only for the case of `$value->getData()` equals `false`. ( `empty(false)` returns `true`)

PR #10520, opened 2019-08-05
PIM-8576: Fix attribute options in case of very long strings

pierallard:
Before
Selection_108
Selection_107

After

Selection_105
Selection_106


pierallard:
@tamarasaurus  It depends if someone in the validation process ask me do to this. The PR will pass from 6 lines to 300.

PR #10525, opened 2019-08-05
TIP-1185: Add migration script

tamarasaurus:

This PR adds a migration script to delete the alias + index associated with `akeneo_pim_product` and `akeneo_pim_product_model`. 

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/upgrades/3.2-4.0/es_delete_unused_indexes.php
+++ b/upgrades/3.2-4.0/es_delete_unused_indexes.php
@@ -21,20 +21,19 @@ $hosts = [$container->getParameter('index_hosts')];
 
 $client = $builder->setHosts($hosts)->build()->indices();
 
-echo 'Deleting alias akeneo_pim_product and index.';
-if ($client->existsAlias(['name' => 'akeneo_pim_product' ])) {
-    $productAlias = $client->getAlias(['name' => 'akeneo_pim_product' ]);
-    $productIndex = array_keys($productAlias)[0];
-    $client->deleteAlias([ 'index' => $productIndex, 'name' => 'akeneo_pim_product' ]);
-    return $client->delete(['index' => $productIndex]);
+function deleteAliasAndIndex($client, $aliasName) {
+    if ($client->existsAlias(['name' => $aliasName ])) {
+        echo 'Deleting alias '.$aliasName.' and index.';
+        $alias = $client->getAlias(['name' => $aliasName ]);
+        $index = array_keys($alias)[0];
+        $client->deleteAlias([ 'index' => $index, 'name' => $aliasName ]);
+        return $client->delete(['index' => $index]);
+    }
+
+    echo 'The alias "'.$aliasName.'"" does not exist. This could be because it\'s part of the Enterprise Edition.';
 }
 
-echo 'Deleting alias akeneo_pim_product_model and index.';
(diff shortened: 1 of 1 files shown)


upgrades/3.2-4.0/es_delete_unused_indexes.php
```diff
@@ -0,0 +1,40 @@
...
+$hosts = [$container->getParameter('index_hosts')];
+
+$client = $builder->setHosts($hosts)->build()->indices();
+
+echo 'Deleting alias akeneo_pim_product and index.';
+if ($client->existsAlias(['name' => 'akeneo_pim_product' ])) {
+    $productAlias = $client->getAlias(['name' => 'akeneo_pim_product' ]);
+    $productIndex = array_keys($productAlias)[0];
+    $client->deleteAlias([ 'index' => $productIndex, 'name' => 'akeneo_pim_product' ]);
+    return $client->delete(['index' => $productIndex]);
```
ahocquard: I think you should do a function and delete both CE and EE indexes.
`
$this->deleteIndexAndAlias('product......');
$this->deleteIndexAndAlias('published...');
`
If the index or the alias does not exist, it just print a message that the index does not exist. SOmething like:
`The index does not exist. Maybeyou have a Community Edition and this index is part of the Enterprise Edition.`

It avoids to have two migrations scripts and the result is the same for the community.

Otherwise, :+1: for the implementation

upgrades/3.2-4.0/es_delete_unused_indexes.php
```diff
@@ -0,0 +1,43 @@
+<?php
+
+if (!file_exists(__DIR__ . '/../../app/AppKernel.php')) {
+    die("Please run this command from your Symfony application root.");
+}
+
+require __DIR__ . '/../../vendor/autoload.php';
+require __DIR__ . '/../../app/AppKernel.php';
```
anaelChardan: This one has moved to `src/Kernel`

upgrades/3.2-4.0/es_delete_unused_indexes.php
```diff
@@ -0,0 +1,43 @@
...
+
+$envFile = __DIR__ . '/../.env';
+if (file_exists($envFile)) {
```
anaelChardan: `instead I think you can require the bootstrap file located: `app/bootstrap.php`

upgrades/3.2-4.0/es_delete_unused_indexes.php
```diff
@@ -0,0 +1,43 @@
...
+
+    echo sprintf(
+        "The alias \"%s\" does not exist. This could be because it's part of the Enterprise Edition.\n",
+        $aliasName
+    );
+}
+
+deleteAliasAndIndex($client, 'akeneo_pim_product');
+deleteAliasAndIndex($client, 'akeneo_pim_product_model');
+deleteAliasAndIndex($client, 'akeneo_pim_published_product_and_product_model');
```
anaelChardan: This one shouldn't be in the EE ?

tamarasaurus: Alex suggested to only have one script to handle both

anaelChardan: ok ;)

PR #10527, opened 2019-08-05
TIP-901: fix SF4 deprecations

jjanvier:
A few SF4 deprecations fixes here:
- bundle overrides are not allowed anymore (but we didn't really use them)
- missing public controllers
- remove JSON DBAL type that has became useless
- apply strict option on Choices forms
- fix twig deprecation


src/Akeneo/Tool/Bundle/StorageUtilsBundle/Doctrine/DBAL/Types/JsonType.php
```diff
@@ -1,45 +0,0 @@
...
-        return 'json';
-    }
-
-    /**
-     * {@inheritdoc}
-     */
-    public function convertToPHPValue($value, AbstractPlatform $platform)
-    {
-        // workaround to fix mysql bug. see https://bugs.mysql.com/bug.php?id=86898
-        $value = preg_replace('/[\x00-\x1F]/', '\\u001f', $value);
```
ahocquard: For reviewers: it has been fixed in Mysql 5.7.21

PR #10534, opened 2019-08-06
PIM-8313: Do not display already added attributes in the family attributes selector dropdown

pierallard:
- The query was wrong
- The controller doesn't take in account the parameter from the front


src/Akeneo/Pim/Structure/Bundle/Doctrine/ORM/Repository/InternalApi/AttributeSearchableRepository.php
```diff
@@ -178,9 +178,13 @@ protected function findBySearchQb($search, array $options)
         }
 
         if (null !== $options['excluded_family']) {
-            $qb->leftJoin('a.families', 'xf');
-            $qb->andWhere('xf.code IS NULL OR xf.code <> :excluded_family');
+            $subQuery = $this->entityManager->createQueryBuilder()
```
ahocquard: if it's wrong, you should modify a test. If it's missing, you should add it.

PR #10538, opened 2019-08-06
PIM-8624: Fix default product grid view selector in user profile when there is more than 20 views

pierallard:
The previous field was loading "all" the grid views before displaying it.
2 issues:
- it's not scalable (imagine 10000 views)
- the query automatically strip to the 20 firsts ones.

So I transformed it into an "ajax-way", to load them only when user click, and 20 by 20.


src/Oro/Bundle/PimDataGridBundle/Repository/DatagridViewRepository.php
```diff
@@ -51,6 +51,11 @@ public function findDatagridViewBySearch(UserInterface $user, $alias, $term = ''
         $options += ['limit' => 20, 'page' => 1];
         $offset = (int) $options['limit'] * ((int) $options['page'] - 1);
 
+        $identifiers = null;
```
ahocquard: please add a test at least on queries

PR #10546, opened 2019-08-07
PIM-8631: Fix column selector in the case of an attribute code as integer

pierallard:
| Added integration tests           | y


tests/legacy/features/pim/enrichment/product/datagrid/display_attributes_in_the_grid.feature
```diff
@@ -54,3 +54,18 @@ Feature: Display product attributes in the grid
       | column    | title       |
       | Side view | **empty**   |
       | Top view  | akeneo2.jpg |
+
+  @jira https://akeneo.atlassian.net/browse/PIM-8631
+  Scenario: Successfully display attributes in the datagrid with code as integer
```
ahocquard: This is an edge case, we don't need an end-to-end test on it I think. Is a spec doable?

PR #10554, opened 2019-08-08
TIP-901: use SF4 config directory

jjanvier:
One step forward into Symfony 4 :) Here it's about the configiration directory structure. No more `/app/config`, please welcome `/config`. If you want to know more about it please read [this article from FabPot](http://fabien.potencier.org/symfony4-directory-structure.html) and the doc [here](https://symfony.com/doc/current/quick_tour/the_architecture.html#development-versus-production-environments) and [there](https://symfony.com/doc/current/configuration.html#configuration-environments).

```
jjanvier:pcd$ tree config
config
├── bootstrap.php
├── bundles.php
├── packages
│   ├── config.yaml
│   ├── security.yaml
│   ├── behat
│   │   └── config.yaml
│   ├── dev
│   │   └── config.yaml
│   ├── prod
│   │   └── config.yaml
│   ├── test
│   │   ├── config.yaml
│   │   └── security
│   │       └── security.yaml
│   └── test_fake
│       └── config.yaml
├── routes
│   ├── dev
│   │   └── routes.yaml
│   └── routes.yaml
└── services
    ├── pim_parameters.yaml
    ├── services.yaml
    └── test_fake
        └── services.yaml

11 directories, 15 files
```

Personnally, I really like this new structure. It's easier to understand what's an environment, how you can configure a package for a specific environment and the differences between the DI (`services` folder), the routes (`routes` folder) and the configuration of the libraries we use (`packages` folder)

A difference to note is that now, environments are not meant to be extended anymore. For instance `test` env does not extend anymore `dev`. They are completely independant. 

Please note that in this PR:
- ~`.yml` are now extended `.yaml` (yes, that's the new SF4 convention...)~ reverted
- `app/config` has not been removed yet. I'll do it in another PR
- all the files present in `packages` should be split ideally. This could allow us to use Flex full possibilities. For instance, `packages/prod/config.yaml` should be split into `packages/prod/doctrine.yaml` and `packages/prod/monolog.yaml`. The goal is that if you uninstall monolog for instance, Flex will automatically drop its configuration file. I'll do it in another PR. We have to find a smart way to do that without copying/pasting everything in EE.
- `pim_parameters.yml` has been kept as is. I need to find something smart to not copy everything in EE by removing this file.


config/packages/behat/services.yaml
```diff
@@ -1,6 +1,3 @@
-imports:
-    - { resource: config.yml }
-
 framework:
```
arnolanglade: Nope, these are not services but package configuration. I guess you should name it `config.yaml` for now.

config/packages/test_fake/services.yaml
```diff
@@ -0,0 +1,24 @@
+imports:
```
arnolanglade: Idem those import should go in `config/services_test_fake.yaml`

bin/check-services-instantiability
```diff
@@ -7,7 +7,7 @@ if (!file_exists('src/Kernel.php')) {
     die("Please run this command from your Symfony application root.");
 }
 
-require 'app/bootstrap.php';
+require 'config/bootstrap.php';
```
arnolanglade: This file already includes 'vendor/autoload.php';

.circleci/local-object-storage-setup.sh
```diff
@@ -3,9 +3,8 @@
 SCRIPT_DIR=$(dirname $0)
 
 
-cat >> $SCRIPT_DIR/../app/config/config.yml <<OBJECT_STORAGE_CONF
+cat >> $SCRIPT_DIR/../config/services/services.yaml <<OBJECT_STORAGE_CONF
 
-services:
```
skeleton: normal?

jjanvier: yep, it's pushed into `services.yaml` that already has the `services:` key

PR #10552, opened 2019-08-08
PIM-8632: Fix empty column in the product datagrid when product model doest not have any shared value

pierallard:
In the case of
- a product model without any value
- a product variant of this PM with a value on it

The query will do a JSON_MERGE of `[]`, `{ sku: ... }`, which result in `[{}, { sku: ... }]`.

If I have to add a test, feel free to indicate what and where.

| Q                                 | A
| --------------------------------- | ---
| Added Specs                       | Todo
| Added legacy Behats               | Todo
| Added acceptance tests            | Todo
| Added integration tests           | Todo
| Changelog updated                 | Todo
| Review and 2 GTM                  | Todo
| Micro Demo to the PO (Story only) | Todo
| Migration script                  | -
| Tech Doc                          | -

PR #10560, opened 2019-08-09
PIM-8633: Fix query to get product models when root product model does not have any shared value

pierallard:
Follow this fix:
https://github.com/akeneo/pim-community-dev/pull/10552

As @ahocquard said, there is another fail with this case with another query. Fixed!


pierallard:
@ahocquard Totally agree. But it needs also a migration script.

PR #10567, opened 2019-08-12
TIP-1159: Remove CompletenessRemover

pierallard:
Remove useless service and commands about CompletenessRemover


tests/back/Pim/Enrichment/Integration/Completeness/CalculateCompletenessCommandIntegration.php
```diff
@@ -59,4 +59,14 @@ private function getCountRowCompleteness()
 
         return (int) $stmt->fetch()['count'];
     }
+
+    private function removeAllCompletenesses($product)
+    {
+        $sql = 'DELETE FROM pim_catalog_completeness';
+        $this->get('doctrine.orm.entity_manager')->getConnection()->prepare($sql)->execute();
```
ahocquard: you can get directly the connection I think

PR #10588, opened 2019-08-14
TIP-1159: Transform indexing commands to be container aware

pierallard:


ahocquard:
Missing EE to avoid to inject permission PQB

PR #10602, opened 2019-08-19
PIM-8661: Fix merging values from a variant product when one of its ancestors has empty values

mmetayer:
Empty values are stored as an empty JSON array in `raw_values` column, which used to conflict with the JSON_MERGE MySQL function, and return an array instead of an object. 
ex:
`SELECT JSON_MERGE('[]', '{"foo": "bar"}', '{"bar": "baz"}');` 
will return
 `[{"foo": "bar"}, {"bar": "baz"}]`
whereas we expect 
`{"foo": "bar", "bar": "baz"}`

This PR ensures that we merge only objects in order to get the expected format for the results.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Akeneo/Pim/Enrichment/Bundle/Product/Query/Sql/GetValuesAndPropertiesFromProductIdentifiers.php
+++ b/src/Akeneo/Pim/Enrichment/Bundle/Product/Query/Sql/GetValuesAndPropertiesFromProductIdentifiers.php
@@ -27,6 +27,8 @@ final class GetValuesAndPropertiesFromProductIdentifiers
 
     public function fetchByProductIdentifiers(array $productIdentifiers): array
     {
+        // TODO TIP-1231: remove the 'IF' conditions in the JSON_MERGE_PRESERVE statement,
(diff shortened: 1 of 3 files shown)


src/Akeneo/Pim/Enrichment/Bundle/Product/Query/Sql/GetValuesAndPropertiesFromProductIdentifiers.php
```diff
@@ -37,7 +39,11 @@ public function fetchByProductIdentifiers(array $productIdentifiers): array
     p.updated,
     f.code AS family_code,
     JSON_ARRAYAGG(g.code) AS group_codes,
-    JSON_MERGE(COALESCE(pm1.raw_values, '{}'), COALESCE(pm2.raw_values, '{}'), p.raw_values) as raw_values
```
mmetayer: `JSON_MERGE` is deprecated, see https://dev.mysql.com/doc/refman/5.7/en/json-modification-functions.html#function_json-merge

ahocquard:
@pierallard you did in datagrid I think, not in API

PR #10606, opened 2019-08-19
TIP-1202: remove the old app/ directory

jjanvier:
This PR removes the `app/` directory and cleans ther work started by https://github.com/akeneo/pim-community-dev/pull/10554:
- app/archive, app/file_storage and app/uploads are now located in var/
- PIM requirements are moved to var/. This is not where they should belong, it's just temporary. (var/ should contain files to which the system writes data during the course of its operation). But as Symfony completely changed the way to check its requirement, it requires more work and thinking.
- phpunit's conf has been placed at the root of the repository. This is not ideal, but I did the same as our other tests libraries and tools. Maybe we should clean that someday.
- several other minor stuff...


var/SymfonyRequirements.php
```diff
@@ -1,5 +1,7 @@
 <?php
 
+// TODO: This file should not be in var/. Handle this properly via TIP-1232
```
arnolanglade: Agree!

PR #10616, opened 2019-08-20
TIP-1204 Rework docker-compose configuration

arnolanglade:

This PR update update the `docker-compose.yml` because we created new docker php images few days ago: one for production (SAAS) and another for dev purpose.

First, you need to build php images locally because they are not pushed in a docker registry:

```bash
$ make php-image-dev
```
You need to run the following command if you need to build production image:

```bash
$ make php-image-prod
```

Then, to set up your local environment you need to run: 

```bash
$ APP_ENV=behat make up
```
**Allowed values for APP_ENV** are `prod`, `dev`, `test_fake` and `test` regarding on the symfony env you want to work. For you information, `behat` and `production` containers are not provided by default. Please run the previous command to boot the PIM on the right Symfony environment.

`OK`: Done / Validated
`-`: Not needed


.circleci/wait_docker_up.sh
```diff
@@ -19,7 +19,7 @@ echo "MySQL server is running!"
 
 COUNTER=1
 echo "Waiting for Elasticsearch server…"
-while ! docker-compose exec fpm curl http://elastic:changeme@elasticsearch:9200/_cat/health | grep green > /dev/null 2>&1; do
+while ! docker-compose run --rm php curl http://elastic:changeme@elasticsearch:9200/_cat/health | grep green > /dev/null 2>&1; do
```
arnolanglade: Todo: we need to install curl for dev image:
```
/usr/local/bin/docker-php-entrypoint: 8: exec: curl: not found
```

src/Akeneo/Tool/Component/Console/CommandLauncher.php
```diff
@@ -53,14 +53,9 @@ protected function getPhp()
...
-            '%s %s%sconsole --env=%s %s',
-            $this->getPhp(),
-            sprintf('%s%s..%sbin', escapeshellarg($this->rootDir), DIRECTORY_SEPARATOR, DIRECTORY_SEPARATOR),
-            DIRECTORY_SEPARATOR,
-            $this->environment,
-            $command
-        );
+        $memoryLimit = ini_get('memory_limit');
+
+        return "{$this->getPhp()} -d memory_limit={$memoryLimit} {$this->rootDir}/../bin/console --env={$this->environment} {$command}";
```
jjanvier: everywhere we use `sprintf`
why change that?

jjanvier:
> For you information, there is no more a behat and a production environment by default.

what do you mean?
there no more 2 different containers that's it? (not Symfony environment)

src/Akeneo/Tool/Component/Console/CommandLauncher.php
arnolanglade: because of `-d memory_limit={$memoryLimit}` . Without it, jobs crash because they don't have enough memory.

arnolanglade: BTW, this syntax is already used in the PIM. Moreover sprint is not readable!

arnolanglade:
> > For you information, there is no more a behat and a production environment by default.
> 
> what do you mean?
> there no more 2 different containers that's it? (not Symfony environment)

You're right, the description has been updated.

.circleci/config.yml
```diff
@@ -37,22 +39,18 @@ jobs:
...
-          name: Pim yarn dev installation
-          command: docker-compose run --rm node yarn run webpack-dev
+          command: |
+              sudo chown -R 1000:1000 ../project
+              sudo chown -R 1000:1000 ~/.composer
       - run:
-          name: Pim yarn test installation
-          command: docker-compose run --rm node yarn run webpack-test
+          name: Install PIM
+          command: |
```
BitOne: I would prefer to use different steps in order to have a better view of the timing of each part.

arnolanglade: The CI is green. I guess it can be done in another PR!

PR #10618, opened 2019-08-21
TIP-1159: fix coupling between enrichment and structure we introduced with the completeness

ahocquard:
The family mask was coupled to the product completeness mask.
We should not have such coupling. The coupling hsould be just product -> structure, not the opposite.

So, I just added the calculation to return a product completeness in the product completeness mask (instead of family mask).

Also, I moved every models in the same directory for consistency. 
And I renamed completeness family mask to required attribute mask: there is no notion of completeness in structure.

| Added legacy Behats               | Never, ever

`OK`: Done / Validated
`-`: Not needed


src/Akeneo/Pim/Structure/Component/Query/PublicApi/Family/RequiredAttributesMaskForChannelAndLocale.php
```diff
@@ -4,20 +4,17 @@
...
-class CompletenessFamilyMaskPerChannelAndLocale
+class RequiredAttributesMaskForChannelAndLocale
 {
     /**
      * This separator should not be allowed in attribute codes
      *
      * @var string
      */
-    private const ATTRIBUTE_CHANNEL_LOCALE_SEPARATOR = '-';
+    public const ATTRIBUTE_CHANNEL_LOCALE_SEPARATOR = '-';
```
pierallard: We have to check that a locale can not have '-' in its code. Create card plz :heart:

src/Akeneo/Pim/Structure/Component/Query/PublicApi/Family/RequiredAttributesMask.php
```diff
@@ -0,0 +1,56 @@
...
+declare(strict_types=1);
+
+namespace Akeneo\Pim\Structure\Component\Query\PublicApi\Family;
+
+/**
+ * This mask is done to gather all the masks for a given family
+ * e.g:
+ *     Given a channel "ecommerce" with locales "en_US", "fr_FR"
+ *     AND a channel "tablet" with locales "en_UK"
+ *     AND a family "t-shirts" with attributes "size"
```
mmetayer: (scopable and localizable)

PR #10621, opened 2019-08-21
PIM-8673: Add a fallback to get the mime-type of files loaded without metadata.

pchasle:

When using a cloud storage solution (AWS S3, Google Cloud Storage). You need to send the `Content-Type` metadata along with your file or it will be stored as `application/octet-stream`.

The goal of this PR is to:
- Set the metadata option required by each Flysystem adapter (dirty fix, but a rewrite to handle properly each adapter configuration would probably be too much).
- Fallback to load the mime-type of a file from the FileInfo repository if we get a `Content-Type` `application/octet-stream` when loading a file from the storage.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Akeneo/Platform/Bundle/UIBundle/Imagine/FlysystemLoader.php
+++ b/src/Akeneo/Platform/Bundle/UIBundle/Imagine/FlysystemLoader.php
@@ -3,6 +3,7 @@
 namespace Akeneo\Platform\Bundle\UIBundle\Imagine;
 
 use Akeneo\Tool\Component\FileStorage\FilesystemProvider;
+use Akeneo\Tool\Component\FileStorage\Repository\FileInfoRepositoryInterface;
 use Liip\ImagineBundle\Binary\Loader\LoaderInterface;
 use Liip\ImagineBundle\Exception\Binary\Loader\NotLoadableException;
 use Liip\ImagineBundle\Model\Binary;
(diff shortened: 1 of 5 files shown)


src/Akeneo/Platform/Bundle/UIBundle/Imagine/FlysystemLoader.php
```diff
@@ -85,6 +86,13 @@ protected function retrieveContentFileFromVfs($path)
             throw new NotLoadableException(sprintf('Unable to read the file "%s" from the filesystem.', $path));
         }
 
+        if ('application/octet-stream' === $mimeType && null !== $this->fileInfoRepository) {
```
ahocquard: not tested (a specification exist)

src/Akeneo/Platform/Bundle/UIBundle/Imagine/FlysystemLoader.php
```diff
@@ -23,14 +24,14 @@ class FlysystemLoader implements LoaderInterface
...
 
-    /**
-     * @param FilesystemProvider $filesystemProvider
-     * @param array              $filesystemAliases
-     */
-    public function __construct(FilesystemProvider $filesystemProvider, array $filesystemAliases)
+    /** @var FileInfoRepositoryInterface | null */
+    protected $fileInfoRepository;
+
+    public function __construct(FilesystemProvider $filesystemProvider, array $filesystemAliases, FileInfoRepositoryInterface $fileInfoRepository = null)
```
ahocquard: could you add a comment  to remove null during merge please

mmetayer: please add a TODO to remove the `= null` for the last argument on 4.0

src/Akeneo/Platform/Bundle/UIBundle/Imagine/FlysystemLoader.php
mmetayer: Could you add a test? Also, add a TODO to remove the ` null !== $this->fileInfoRepository` test on 4.0

src/Akeneo/Platform/Bundle/UIBundle/Imagine/FlysystemLoader.php
pchasle: This fallback could stay optional as it's mainly a patch to fix a wrong mime-type on existing objects. What do you think?

ahocquard: if it's for a patch and useless later on: the argument should be removed on master
if it's mandatory to do it later on: remove null

But definitively not an optional dependency on master :)

mmetayer: I don't like optional arguments in constructors, it forces us to test the existence of the property before being able to use it (see L89: `&& null !== $this->fileInfoRepository`)

pchasle: Thanks, I will add the comments to make the dependency mandatory on master.

PR #10625, opened 2019-08-22
TIP-1231: Remove [] in SQL

pierallard:


tests/back/Pim/Enrichment/Specification/Component/Product/Model/ProductModelSpec.php
```diff
@@ -384,4 +384,10 @@ function it_gets_label_when_casting_object_as_string(
 
         $this->__toString()->shouldReturn('shovel');
     }
+
+    function it_saves_empty_raw_values()
```
ahocquard: :smile: 

I don't know how we could test it without reflection...  Let's say it's covered by integration tests...

pierallard: I think we can't test it in a better way than this :) But we saw that integration tests covers well this use case with the php type casts.

PR #10641, opened 2019-08-26
PIM-8678: Change the updated property value of the Family entity when the attribute requirements are updated

olivierpontier:
This pull request aims to change the value of the updated property from the Family entity when the user updates the attribute requirements (SETTINGS > FAMILIES > ${FAMILY} > Attributes)

This is done "automatically" if the family implement the Timestampable interface. Interface methods already are implemented in the Family model, but the the interface was not declared.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/CHANGELOG-3.0.md
+++ b/CHANGELOG-3.0.md
@@ -3,6 +3,7 @@
 ## Bug fixes
 
 - PIM-8615: Fix issue issue with boolean attribute used as variant axis
+- PIM-8678: Fill the updated property on Family entity update
 
 # 3.0.38 (2019-08-20)
 
--- a/CHANGELOG-3.0.md
+++ b/CHANGELOG-3.0.md
@@ -2,8 +2,11 @@
 
 ## Bug fixes
 
-- PIM-8615: Fix issue issue with boolean attribute used as variant axis
+- PIM-8615: Fix issue with boolean attribute used as variant axis
 - PIM-8678: Fill the updated property on Family entity update
+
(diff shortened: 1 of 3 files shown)


src/Akeneo/Pim/Structure/Component/Model/FamilyInterface.php
```diff
@@ -14,7 +15,11 @@
  * @copyright 2014 Akeneo SAS (http://www.akeneo.com)
  * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
  */
-interface FamilyInterface extends TranslatableInterface, ReferableInterface, VersionableInterface
+interface FamilyInterface extends
+    TranslatableInterface,
+    ReferableInterface,
+    VersionableInterface,
+    TimestampableInterface
```
mmetayer: This is a BC break (even if our concrete Family implem already has the methods), you should document it

arnolanglade: yes you should add something in the changelog

CHANGELOG-3.0.md
```diff
@@ -3,6 +3,7 @@
 ## Bug fixes
 
 - PIM-8615: Fix issue issue with boolean attribute used as variant axis
+- PIM-8678: Fill the updated property on Family entity update
```
jmleroux: Please add a BC break to explain why we add the Timestampable interface.

PR #10643, opened 2019-08-26
TIP-1204 Rework docker-compose configuration

arnolanglade:

`DockerFile`, `php` and `apache` configuration have been copied / pasted from EE to CE. The purpose is to make the CI passes with the new php images. Then, we need to think what should be the best solution to avoid this ducplication.

`OK`: Done / Validated
`-`: Not needed


.circleci/config.yml
```diff
@@ -37,22 +39,18 @@ jobs:
...
-      - run:
-          name: Pim yarn dev installation
-          command: docker-compose run --rm node yarn run webpack-dev
+          command: |
+              sudo chown -R 1000:1000 ../project
+              sudo chown -R 1000:1000 ~/.composer
       - run:
-          name: Pim yarn test installation
-          command: docker-compose run --rm node yarn run webpack-test
+          name: Install PIM
```
BitOne: I think multiple steps is better than a suite of command. This way, we can monitor the duration of each step.

jjanvier: ok, I'll fix in another PR
here the CI is green

PR #10647, opened 2019-08-27
TIP-1213 Clean .env files

arnolanglade:

This PR cleans some little glitches like `.env` files because we don't need to override env vars if their values are the same than in `.env`.

| Added Specs                       |  -
| Added legacy Behats               |  -
| Added acceptance tests            |  -
| Added integration tests           |  -
| Changelog updated                 |  -
| Micro Demo to the PO (Story only) |  -

`OK`: Done / Validated
`-`: Not needed


docker-compose.yml
```diff
@@ -95,7 +95,7 @@ services:
       MINIO_ACCESS_KEY: 'AKENEO_OBJECT_STORAGE_ACCESS_KEY'
       MINIO_SECRET_KEY: 'AKENEO_OBJECT_STORAGE_SECRET_KEY'
     ports:
-      - '${DOCKER_PORT_OBJECT_STORAGE:-9010}:9000'
+      - '${DOCKER_PORT_OBJECT_STORAGE:-9020}:9000'
```
BitOne: 9010, no? As we decided to keep the same ports between CE and EE?

arnolanglade:
1 unstable but it passes locally

PR #10653, opened 2019-08-28
TIP-1204 Fix little glitches

arnolanglade:

We make some targets (which run test) more generic because they can be easily configurable through env vars. Don't you still can add you custom target in the directory `make-file`. The installation step has been split because we want better timing information.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -43,14 +43,26 @@ jobs:
               sudo chown -R 1000:1000 ../project
               sudo chown -R 1000:1000 ~/.composer
       - run:
-          name: Install PIM
-          command: |
-              make vendor
-              make node_modules
-              make assets
-              make css
-              make javascript-test
-              make javascript-dev
+            name: Install back dependencies
+            command: |
+                make vendor
+      - run:
+            name: Install front dependencies
+            command: |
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -43,14 +43,26 @@ jobs:
...
-              make assets
-              make css
-              make javascript-test
-              make javascript-dev
+            name: Install back dependencies
+            command: |
+                make vendor
+      - run:
+            name: Install front dependencies
+            command: |
```
BitOne: You can remove the "|" ;)

.circleci/config.yml
```diff
@@ -43,14 +43,26 @@ jobs:
...
+                make assets
+      - run:
+            name: Build CSS
+            command: |
+                make css
+      - run:
+            name: Build Javascript
+            command: |
+                make javascript-test
+                make javascript-dev
```
BitOne: Why javascript-dev is required? javascript-test should be enough, no?

arnolanglade: no, javascript-dev generates javascript files needed to run the application. This is a little weird!

PR #10657, opened 2019-08-29
PIM-8710: New converter to flatten invalid import values

jmleroux:
This PR fixes an error raised when trying to write the invalid data file for an import with warnings.

We tried to import a file with a valid date field and another invalid attribute field.

- the row is read from the XLSX file and the date field is automtically casted into a DateTime object
- the row is invalidated because of the invalid field value
- the row is written in the invalid data file.

BUT, because the date field is still a DateTime, the JSON File buffer [will encode it into an array](https://github.com/akeneo/pim-community-dev/blob/3.0/src/Akeneo/Tool/Component/Buffer/JSONFileBuffer.php#L58) which is not writable.

With this fix, the row is buffered into a into a pure flat key => string value array.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Akeneo/Tool/Component/Connector/ArrayConverter/StringArrayConverter.php
+++ b/src/Akeneo/Tool/Component/Connector/ArrayConverter/StringArrayConverter.php
@@ -1,0 +1,63 @@
+<?php
+
+namespace Akeneo\Tool\Component\Connector\ArrayConverter;
+
+use Akeneo\Tool\Component\Connector\Exception\DataArrayConversionException;
+
+/**
+ * String array converter.
+ * It flattens an array to a key => string value array.
+ *
+ * @author JM Leroux <jean-marie.leroux@akeneo.com>
+ * @copyright 2019 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php Open Software License (OSL 3.0)
+ */
+class StringArrayConverter implements ArrayConverterInterface
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function convert(array $item, array $options = [])
+    {
+        $convertedItem = [];
+
+        foreach ($item as $property => $data) {
+            $convertedItem[$property] = $this->convertToString($property, $data);
+        }
+
+        return $convertedItem;
+    }
+
+    /**
+     * @throws DataArrayConversionException
+     */
+    private function convertToString(string $property, $data): string
+    {
+        $convertedData = null;
+
+        if ($data instanceof \DateTime) {
+            $convertedData = $data->format('Y-m-d H:i:s');
+        } elseif (is_array($data)) {
+            if (count($data) !== count($data, COUNT_RECURSIVE)) {
+                throw new DataArrayConversionException(
+                    sprintf('The property "%s" could not be converted into string.', $property)
+                );
+            }
+            $convertedData = implode(',', $data);
+        } elseif (is_object($data) && !method_exists($data, '__toString')) {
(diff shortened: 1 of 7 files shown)


src/Akeneo/Tool/Component/Connector/ArrayConverter/StringArrayConverter.php
```diff
@@ -0,0 +1,61 @@
...
+        return $convertedItem;
+    }
+
+    /**
+     * @throws DataArrayConversionException
+     */
+    private function convertToString($property, $data): string
+    {
+        if ($data instanceof \DateTime) {
+            $convertedData = $data->format('Y-m-d H:i:s');
```
skeleton: dunno about that because if a user tries to reimport invalid data, this value won't be as before. maybe it's not a problem, you should check with po

src/Akeneo/Tool/Component/Connector/ArrayConverter/StringArrayConverter.php
```diff
@@ -0,0 +1,61 @@
...
+            $convertedItem[$property] = $this->convertToString($property, $data);
+        }
+
+        return $convertedItem;
+    }
+
+    /**
+     * @throws DataArrayConversionException
+     */
+    private function convertToString($property, $data): string
```
skeleton: string $property

src/Akeneo/Tool/Component/Connector/ArrayConverter/StringArrayConverter.php
```diff
@@ -0,0 +1,61 @@
...
+            $convertedData = implode(',', $data);
+        } elseif (is_object($data) && !method_exists($data, '__toString')) {
+            throw new DataArrayConversionException(
+                    sprintf('The property "%s" could not be converted into string.', $property)
+                );
+        } else {
+            $convertedData = (string) $data;
+        }
+
+        if (!is_string($convertedData)) {
```
skeleton: `if (null === $convertedData)` + initialize variable at the beginning of the method? would be clearer

jmleroux:
> about the naming maybe something like `ConvertInvalidDataToString` ?

No because it is just a converter. Currently i use it only for invalid data, but it's only purpose is to flatten an array into a simpler key-value array.

src/Akeneo/Tool/Component/Connector/ArrayConverter/StringArrayConverter.php
jmleroux: Good catch, will check.

src/Akeneo/Tool/Component/Connector/ArrayConverter/StringArrayConverter.php
jmleroux: I initialized it because it's obviously clearer. But no need to test for `null` : `is_string(null) = false` and its sufficient.

skeleton:
> > about the naming maybe something like `ConvertInvalidDataToString` ?
> 
> No because it is just a converter. Currently i use it only for invalid data, but it's only purpose is to flatten an array into a simpler key-value array.

what I don't like with StringArrayConverter, it's too technical and to precise. If later, you have to add a new conversion in your class, the naming will be wrong. The functional purpose is to handle the invalid data, not to convert data in a classical import way, we already have converter for that.
But I already approved your PR, do what you want ;)

jmleroux:
> The functional purpose is to handle the invalid data, not to convert data in a classical import way, we already have converter for that.

I get your point. This converter is shitty and should not be used elsewhere than to convert invalid data. I will rename. Thx.

PR #10661, opened 2019-08-29
TIP-1249: upgrade to MySQL 8

jjanvier:
Let's use MySQL 8 with the PIM :)

### DIFF
--- a/src/Akeneo/Pim/Enrichment/Bundle/Storage/Sql/ProductGrid/FetchProductRowsFromIdentifiers.php
+++ b/src/Akeneo/Pim/Enrichment/Bundle/Storage/Sql/ProductGrid/FetchProductRowsFromIdentifiers.php
@@ -324,7 +324,7 @@ SQL;
         $sql = <<<SQL
             SELECT 
                 p.identifier,
-                JSON_ARRAYAGG(COALESCE(ft.label, CONCAT("[", g.code, "]"))) as groups 
+                JSON_ARRAYAGG(COALESCE(ft.label, CONCAT("[", g.code, "]"))) AS product_groups 
             FROM
                 pim_catalog_product p
                 JOIN pim_catalog_group_product gp ON gp.product_id = p.id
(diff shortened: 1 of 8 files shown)


src/Akeneo/Pim/Enrichment/Bundle/Storage/Sql/ProductGrid/FetchProductRowsFromIdentifiers.php
```diff
@@ -324,7 +324,7 @@ private function getGroups(array $identifiers, string $localeCode): array
         $sql = <<<SQL
             SELECT 
                 p.identifier,
-                JSON_ARRAYAGG(COALESCE(ft.label, CONCAT("[", g.code, "]"))) as groups 
+                JSON_ARRAYAGG(COALESCE(ft.label, CONCAT("[", g.code, "]"))) AS "groups"
```
BitOne: Is it because groups is reserved word? If that's the case, we shouldn't use it at all. product_groups maybe ?

tests/back/Pim/Enrichment/Integration/Storage/Sql/ProductGrid/AssertRows.php
```diff
@@ -17,6 +17,24 @@ public static function same(array $expectedRows, array $rows): void
...
+            $actualRow = null;
+
+            foreach ($rows as $row) {
+                if ($row->identifier() === $identifier) {
+                    $actualRow = $row;
+                    break;
+                }
+            }
+
+            self::assertSameRow($expectedRow, $actualRow);
```
mmetayer: :+1:
I would just add an assertion to check that `$acualRow` was found before calling `self::assertSameRow($expectedRow, $actualRow);`

src/Akeneo/Pim/Enrichment/Bundle/Storage/Sql/ProductGrid/FetchProductRowsFromIdentifiers.php
jjanvier: yes you're right, it became a reserved keywords https://dev.mysql.com/doc/refman/8.0/en/keywords.html#keywords-in-current-series

PR #10678, opened 2019-09-03
PIM-8712: Fix folder prefix for temporary files

tamarasaurus:
This PR fixes the FileFetcher when using a remote filesystem to store files like Amazon S3.
It started with a fix for assets mass upload where we seen that there are issues with remote filesystem and local temporary storage dir. https://github.com/akeneo/pim-enterprise-dev/pull/6807
This PR introduces the use of a remote filesystem for dev and CI and uses the application config to know the correct temporary storage.

Files are stored on the remote fs and fetched into a local temporary directory per app configuration: https://github.com/akeneo/pim-community-dev/blob/3.2/app/config/pim_parameters.yml#L23

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Akeneo/Tool/Component/FileStorage/spec/File/FileFetcherSpec.php
+++ b/src/Akeneo/Tool/Component/FileStorage/spec/File/FileFetcherSpec.php
@@ -10,12 +10,16 @@ use PhpSpec\ObjectBehavior;
 
 class FileFetcherSpec extends ObjectBehavior
 {
-    function let(Filesystem $tmpFilesystem)
+    function let(Filesystem $tmpFilesystem, Local $adapter)
     {
+        $tmpFilesystem->getAdapter()->willReturn($adapter);
+        $prefix = sys_get_temp_dir() . '/spec/';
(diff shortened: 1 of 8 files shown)


src/Akeneo/Tool/Component/FileStorage/spec/File/FileFetcherSpec.php
```diff
@@ -10,12 +10,16 @@
 
 class FileFetcherSpec extends ObjectBehavior
 {
-    function let(Filesystem $tmpFilesystem)
+    function let(Filesystem $tmpFilesystem, Local $adapter)
```
jjanvier: it's a string in the constructor of the class
how can the spec pass oO ?

docteurklein: it's not passed to the constructor `$this->beConstructedWith($tmpFilesystem);`

jmleroux:
We should change the way we handle local storage on master. We will not do it in a patch to not introduce BC breaks.

PR #10680, opened 2019-09-04
Fix import products test

jmleroux:
I removed an unstable test because it's already tested in this critical scenario: https://github.com/akeneo/pim-community-dev/blob/3.2/tests/legacy/features/pim/enrichment/product/import/import_products.feature#L29


tests/legacy/features/pim/enrichment/product/import/import_products.feature
```diff
@@ -84,30 +84,6 @@ Feature: Import products coming from an external application
     And the english localizable value name of "SKU-001" should be "Donec"
     And the english tablet description of "SKU-001" should be "dictum magna. Ut tincidunt orci quis lectus. Nullam suscipit, est"
 
-  @javascript
```
tamarasaurus: :fire:

PR #10688, opened 2019-09-05
PIM-8734: Change label for default channel in minimal catalog

jmleroux:
When a new client opens for the first time its PIM, some entities are created by default (a category tree, a channel, a locale) 

The existing channel code is "ecommerce" but its label is "Default". 
It can be a bit confusing for the customer, especially if the Used change the channel locale and keep en_US for its UI language. The label Default is then displayed, and the user does not understand where does it come from. 

Expected: change the label "Default" for "Ecommerce" to be compliant with the channel's code.
 
`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/CHANGELOG-3.2.md
+++ b/CHANGELOG-3.2.md
@@ -4,6 +4,7 @@
 
 - PIM-8719: Fix Mink Selenium dependency
 - PIM-8677: Purge all job executions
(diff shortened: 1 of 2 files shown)


jjanvier:
(you could that in icecat too)

jmleroux:
> (you could that in icecat too)

If i'm not mistaken, it's already the case.

PR #10690, opened 2019-09-05
TIP-1241: Use new completeness algorithm to compute missing required attributes in the PEF

mmetayer:
As we now calculate a product's completeness before rendering in in the PEF, we can take advantage to use this accurate completeness to compute the missing required attributes, and we do not need the IncompleteValuesNormalizer anymore. It also allows having a single algorithm for completeness calculation in the PIM.

Also, as we need to calculate missing required attributes for the PEF even for a product model, the new class `MissingRequiredAttributesCalculator` was created, and uses exactly the same algorithm as the CompletenessCalulator. This service should only be used to compute missing attributes for the PEF (in internal_api normalizers)

Note that legacy `IncompleteValuesNormalizer` will be removed in another PR.

`OK`: Done / Validated
`-`: Not needed


src/Akeneo/Pim/Enrichment/Component/Product/Normalizer/InternalApi/MissingRequiredAttributesNormalizer.php
```diff
@@ -0,0 +1,38 @@
...
+namespace Akeneo\Pim\Enrichment\Component\Product\Normalizer\InternalApi;
+
+use Akeneo\Pim\Enrichment\Component\Product\Completeness\Model\ProductCompletenessWithMissingAttributeCodes;
+use Akeneo\Pim\Enrichment\Component\Product\Completeness\Model\ProductCompletenessWithMissingAttributeCodesCollection;
+
+/**
+ * @copyright 2019 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+class MissingRequiredAttributesNormalizer implements MissingRequiredAttributesNormalizerInterface
```
nmarniesse: Is there a reason to implements a `MissingRequiredAttributesNormalizerInterface` instead of the symfony's `NormalizerInterface`?  
I think using the symfony's one will be more re-usable, but maybe overkill for now

mmetayer: @nmarniesse I didn't use Symfony's `NormalizerInterface` because:
- I didn't need to implement `supportsNormalization()` method (useless since I don' use the serializer/normalizer pattern)
- it allowed me to strong-type my first argument, and I didn't need the 2 other arguments (`$format` and `$context`)

I wouldn't have added an interface at all but for the EE (I decorated the service in EE to apply permissions)

PR #10698, opened 2019-09-09
TIP-1222: Create a model for product indexation

nmarniesse:

[TIP-1222](https://akeneo.atlassian.net/browse/TIP-1222)

- Create the `IndexableProduct` model
- Update `ProductIndexer` to use this model

Done by Alex:
- immutable model
- move the model in bundle instead of component as it's infrastructure stuff
- rename it as ElasticsearchProductProjection instead of IndexableProduct as it's infrastructure stuff and just for ES

`OK`: Done / Validated
`-`: Not needed


tests/back/Pim/Enrichment/Integration/Product/IndexableProductIntegration.php
```diff
@@ -0,0 +1,84 @@
...
+use Akeneo\Test\Integration\TestCase;
+
+/**
+ * Class IndexableProductIntegration
+ *
+ * @author    Nicolas Marniesse <nicolas.marniesse@akeneo.com>
+ * @copyright 2019 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+class IndexableProductIntegration extends TestCase
```
ahocquard: The correct level of abstraction to test is to test ORM query in TIP-1223 with sevral integration tests instead. This way your test will not change when the implementation of the query will change: that's the whole goal of a test that guarantee the behavior of you interface implementation.

src/Akeneo/Pim/Enrichment/Component/Product/Connector/ReadModel/IndexableProduct.php
```diff
@@ -0,0 +1,446 @@
...
+            $product->isVariant() && $product->getParent() ? $product->getParent()->getCode() : null,
+            $normalizedValues,
+            static::getAncestors($product, $activatedLocaleCodes, $channelCodes),
+            $label,
+            self::getAttributeCodesOfAncestors($product, $attributesProvider),
+            self::getAttributeCodesForOwnLevel($product, $attributesProvider)
+        );
+    }
+
+    private static function getUpdatedAt(ProductInterface $product): \DateTime
```
ahocquard: I had in mind to put all these private function logic in the ORM query for TIP-1223, and to keep this model quite dumb. It had several pros:
- you only have to spec this model to test that the array is correctly returned as expected, here it's not testable without integration test, which is weird for a model
- you would test the implementation of the ORM query and then change the implementation (without changing the test). It guarantee regression: you should neither change nor remove test when changing implementation details (ORM to SQL). It will guarantee safely that product without family, without groups, etc, are safely handled. 

Actually, with this card, you should not even use this model for now.

nmarniesse: Thanks for your comments. Yes I realize that I did a part of TIP-1223. To be honest the boundaries of TIP-1222 was not very clear for me, but now I get it better.  
A lot of this code will be moved or deleted in next cards

By the way this PR is not finished yet, I work to add a normalizer: there will be no more `toArray` method. I know you prefer the `toArray` method but I find this solution better in term of single of responsability and I found the normalizer more extendable for EE needs. And the integration test will be on the normalizer and not on the model. This will take more sense I guess.

ahocquard: In term of responsibility, it's the same in my opinion (final normalizer that have a static method vs a method to serialize your object as an array). In term of DX, it's easier to have `toArray` and easier to debug. It avoids to leak X getter/setter also. 

For the extensibility part for EE, I don't get what you plan. 

All your data (CE + EE) should already be in your modeI. Your normalizer should not have any logic to enrich data with EE data. That's harder to debug and to test, and it's not its responbility (this is the responsibility of the query). Your normlizer should not have any logic in it. It just transforms data (that's why I think it's  useless and add another layer of complexity instead of `toArray`).

ahocquard: For the `toArray` argument, there was a discussion about it on Twitter for the Symfony client doing it: https://symfony.com/doc/current/components/http_client.html#basic-usage

As you, some people said it leaks things. The oppponent (Symfony + some other people) said that's is better for the DX (I agree). So it's a matter of taste, as you wish. The second part of my comment about EE is more important.

nmarniesse: Ok `toArray` and `additionalFields` look great for me.

PR #10701, opened 2019-09-10
TIP-1254 Uses test file instead of filter for better perf and reliability

BitOne:
Uses test file instead of filter for better perf and reliability


jjanvier:
I'm not a big fan of merging the 2 phpunit steps. The original idea was to say: "we split our tests into suites to have the shortest feedback loop". I still believe in this idea: CS then unit then acceptance then integration then end to end. I'm not willing to sacrifice this idea for a faster build.

That being said, maybe those `phpunit_end_to_end` are not end to end. And that fixes the problem. If it's the case, then we should explain why they are not end to end and maybe rename the step `phpunit_integration`. But this seems more related to the renaming of SF envs than to launching properly the phpunits.

jjanvier:
OK I didn't understand what was done. Forget about my comment.

BitOne:
By merging End to End and Integration, we went from 14minutes (see nighlty https://circleci.com/workflow-run/a6bfd43e-d0a4-4635-8ad5-c821b92a6070) to 11mi46s. So a more than 2 minutes reduction.
Moreover it was the first build using filenames, the timing were not yet known by the CI.

I will monitor it, but it should get better in time.

In all cases, this PR was mainly to workaround a segfault in PHP 7.3 that seems to occur only with --filter. But we will see how it goes later ;)

Now, enjoy your holidays and stop looking at this repo! :)

PR #10705, opened 2019-09-11
Mount the yarn cache in the node container

arnolanglade:

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -36,8 +36,11 @@ services:
   node:
     image: 'akeneo/node:10'
     user: 'node'
+    environment:
+      YARN_CACHE_FOLDER: '/home/node/.cache/yarn'
     volumes:
       - './:/srv/pim'
+      - '${HOST_YARN_CACHE_FOLDER:-~/.cache/yarn}:/home/node/.cache/yarn'
     working_dir: '/srv/pim'
     networks:
       - 'pim'


docker-compose.yml
```diff
@@ -36,8 +36,11 @@ services:
   node:
     image: 'akeneo/node:10'
     user: 'node'
+    environment:
+      YARN_CACHE_FOLDER: '/home/node/.cache/yarn'
     volumes:
       - './:/srv/pim'
+      - '${HOST_YARN_CACHE_FOLDER:-~/.cache/yarn}:/home/node/.cache/yarn'
```
pchasle: For me, this kind of configuration should be a deliberate choice from the developer as you are explicitly referencing/writing files on the system outside of the project scope (aka. in the user home).
It has more its place in user-defined `docker-compose.override.yml`.
It's the same issue for the composer cache volume.

damien-carcel: You can override the value thanks to the environment variable. Plus, you can define this variable in your profile directly, this way it applies to all your projects. And you have cache by default for most devs, without the need to add an override.

pchasle: But what if you don't want any cache outside the docker container?

pchasle: I'm not nitpicking here, I have the issue on my home computer with files permissions for docker in my /home folder.

damien-carcel: Then the real issue seems more to me with the user permissions than with the cache itself, no?
And this can be fixed by setting the container user at runtime.

PR #10723, opened 2019-09-16
PIM-8750: Fix wrong re-render on simple-select-async

pchasle:

If you press tab on the identifier field to select the next field family (keyboard navigation) and you start typing to search for a family; then the select2 field misbehave and jump in the top-left corner of the screen:
bug

Fix:
fix

I fixed it by luck, as it seems to be a race condition between a re-render of the template and the select2 dynamic dropdown positioning.

Removing the re-render doesn't break/change everything on the behavior of the form/select2 on this page and it should not break anything elsewhere (but we never know).

(If someone with extensive knowledge of the frontend stack / select2 have a better explanation I'm interested)

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Akeneo/Platform/Bundle/UIBundle/Resources/public/js/form/common/fields/simple-select-async.js
+++ b/src/Akeneo/Platform/Bundle/UIBundle/Resources/public/js/form/common/fields/simple-select-async.js
@@ -28,13 +28,6 @@ function (
   Routing
 ) {
   return BaseField.extend({
-    events: {
-        'change input': function (event) {
-            this.errors = [];
-            this.updateModel(this.getFieldValue(event.target));
-            this.getRoot().render();
-        }
-    },
     template: _.template(template),
     choiceUrl: null,
     resultsPerPage: 20,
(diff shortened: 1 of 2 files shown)


Doodoune:
Don't forget the changelog, please :)

pchasle:
> This is a common module, it's a very risky solution.
> The goal of this `render()` is to update a form when the behavior of some fields depends on other ones (like you select a choice, another field must become hidden).
> I'm almost sure there will be regressions on other screens. We must find another fix.

The **change** event is triggered even if you are just searching in the select2. Maybe we should use a **on select** instead (or something like that) to only re-render when the user actually choose an option in the select.

src/Akeneo/Platform/Bundle/UIBundle/Resources/public/js/form/common/fields/simple-select-async.js
```diff
@@ -81,7 +74,17 @@ function (
...
      */
     postRender(templateContext) {
-      initSelect2.init(this.$('.select2'), this.getSelect2Options(templateContext));
+      const $select2 = this.$('.select2');
+      
+      initSelect2.init($select2, this.getSelect2Options(templateContext));
+
+      const onChange = () => {
+        this.errors = [];
+        this.updateModel($select2.val());
```
phaseinducer: Does it work with `this.updateModel(this.getFieldValue($select2));` ?
`getFieldValue()` can be important for override purposes.

pchasle: It does :+1:

jmleroux:
Test frozen

PR #10724, opened 2019-09-16
TIP-1160: Add migration

pierallard:
We forbid the use of null values in the domain, as it does not make sense (the value object should not accept null values anymore). It was the cause of multiple bug.

So, we should not have any null value anymore in the storage, which was not the case.
The gaol of this script is to clean the database.

The completeness is not impacted, no need to relaculate it.
We will have to re-index only the impacted product. Do note that the indexation will be with sql queries as we are currently reworking it.

### DIFF
--- a/tests/back/Pim/Enrichment/Upgrade/RemoveProductModelEmptyRawValuesIntegration.php
+++ b/tests/back/Pim/Enrichment/Upgrade/RemoveProductModelEmptyRawValuesIntegration.php
@@ -1,5 +1,6 @@
 <?php
 
+use Akeneo\Test\Integration\Configuration;
 use Akeneo\Test\Integration\TestCase;
 use Akeneo\Tool\Component\Console\CommandLauncher;
 use Doctrine\DBAL\Connection;
@@ -9,9 +10,9 @@ use Doctrine\DBAL\Connection;
  */
 class RemoveProductModelEmptyRawValuesIntegration extends TestCase
 {
-    protected function getConfiguration()
+    protected function getConfiguration(): Configuration
     {
-        $this->catalog->useTechnicalCatalog();
+        return $this->catalog->useTechnicalCatalog();
     }
 
     public function testItRemovesAllEmptyValues()
@@ -29,7 +30,6 @@ INSERT INTO pim_catalog_product_model VALUES
     (NULL, NULL, :familyId, 'pm6', '{"name": {"<all_channels>": {"fr_FR": "", "en_US": "bar"}}}', NOW(), NOW(), 0, 0, 0, 0),
     (NULL, NULL, :familyId, 'pm7', '{"name": {"ecommerce": {"<all_locales>": ""}, "mobile": {"<all_locales>": "bar"}}}', NOW(), NOW(), 0, 0, 0, 0)
 SQL;
-
         $this->getConnection()->executeQuery($sql, ['familyId' => $familyId]);
 
         $this->reExecuteMigration('_4_0_20190917080512_remove_product_model_empty_raw_values');
--- a/tests/back/Pim/Enrichment/Upgrade/RemoveProductModelEmptyRawValuesIntegration.php
+++ b/tests/back/Pim/Enrichment/Upgrade/RemoveProductModelEmptyRawValuesIntegration.php
@@ -1,0 +1,71 @@
+<?php
+
+use Akeneo\Test\Integration\TestCase;
+use Akeneo\Tool\Component\Console\CommandLauncher;
+use Doctrine\DBAL\Connection;
+
+/**
+ * This class will be removed after 4.0 version
+ */
+class RemoveProductModelEmptyRawValuesIntegration extends TestCase
+{
+    protected function getConfiguration()
+    {
+        $this->catalog->useTechnicalSqlCatalog();
+    }
+
+    public function testItRemovesAllEmptyValues()
+    {
+        $this->getConnection()->executeQuery('DELETE FROM pim_catalog_product_model');
+        $sql = <<<SQL
+INSERT INTO pim_catalog_product_model VALUES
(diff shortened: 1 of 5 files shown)


upgrades/schema/Version_4_0_20190916122239_remove_empty_raw_values.php
```diff
@@ -0,0 +1,143 @@
...
+/**
+ * This migration will delete the empty values from raw values of product and product models.
+ * For example, the value {attr: {<all_channels>: {<all_locales: []}}} will be removed from the raw_values field.
+ */
+final class Version_4_0_20190916122239_remove_empty_raw_values extends AbstractMigration implements ContainerAwareInterface
+{
+    /** @var ContainerInterface */
+    private $container;
+
+    const BATCH_SIZE = 100;
```
ahocquard: you can inrcrease it at 1000 or even 10000 I think

upgrades/schema/Version_4_0_20190916122239_remove_empty_raw_values.php
```diff
@@ -0,0 +1,143 @@
...
+
+    private function cleanProducts()
+    {
+        $productsToProcess = true;
+        $productIdentifiersToIndex = [];
+        $page = 0;
+        while ($productsToProcess) {
+            $productsToProcess = false;
+            $sql = sprintf(
+                "SELECT identifier, raw_values FROM pim_catalog_product LIMIT %d, %s",
```
ahocquard: you should not use LIMIT, because it's very slow: it will load data and then do the limit. You should use the previous identifier to request the next page.

upgrades/schema/Version_4_0_20190916122239_remove_empty_raw_values.php
```diff
@@ -0,0 +1,143 @@
...
+                $page * self::BATCH_SIZE,
+                self::BATCH_SIZE
+            );
+            $rows = $this->connection->executeQuery($sql)->fetchAll();
```
ahocquard: I think you could `yield` in a private function to iterate over all products (it avoid double foreach and put the cursor logic in a private function).

Then you will have just `foreach ($this->getRawValues() as $row))`

upgrades/schema/Version_4_0_20190916122239_remove_empty_raw_values.php
```diff
@@ -0,0 +1,143 @@
...
+
+            foreach ($rows as $row) {
+                $productsToProcess = true;
+                $rawValues = json_decode($row['raw_values'], true);
+                $cleanRawValues = $this->getValueCleaner()->cleanAllValues(['ID' => $rawValues])['ID'];
+                if ($rawValues !== $cleanRawValues) {
```
ahocquard: I have not the implementation detail from getValueCleaner. Is it sure it keeps the order?

upgrades/schema/Version_4_0_20190916122239_remove_empty_raw_values.php
```diff
@@ -0,0 +1,143 @@
...
+                        [
+                            'rawValues' => json_encode($cleanRawValues),
+                            'identifier' => $row['identifier']
+                        ], [
+                            'rawValues' => Type::STRING,
+                            'identifier' => Type::STRING
+                        ]
+                    );
+                    $productIdentifiersToIndex[] = $row['identifier'];
+                    if (count($productIdentifiersToIndex) % self::BATCH_SIZE === 0) {
```
ahocquard: you should do it a the end of the loop also

upgrades/schema/Version_4_0_20190916122239_remove_empty_raw_values.php
```diff
@@ -0,0 +1,143 @@
...
+     * {@inheritdoc}
+     */
+    public function setContainer(ContainerInterface $container = null)
+    {
+        $this->container = $container;
+    }
+
+    public function up(Schema $schema) : void
+    {
+        $this->cleanProducts();
```
ahocquard: I would do 2 scripts

upgrades/schema/Version_4_0_20190916122239_remove_empty_raw_values.php
pierallard: Yes.

upgrades/schema/Version_4_0_20190916122239_remove_empty_raw_values.php
pierallard: Already done

tests/back/Pim/Enrichment/Upgrade/RemoveProductModelEmptyRawValuesIntegration.php
```diff
@@ -0,0 +1,71 @@
...
+use Doctrine\DBAL\Connection;
+
+/**
+ * This class will be removed after 4.0 version
+ */
+class RemoveProductModelEmptyRawValuesIntegration extends TestCase
+{
+    protected function getConfiguration()
+    {
+        $this->catalog->useTechnicalSqlCatalog();
```
ahocquard: this catalog should not be used anymore I deprecate it in another PR

upgrades/schema/Version_4_0_20190917080512_remove_product_model_empty_raw_values.php
```diff
@@ -0,0 +1,90 @@
...
+                            'rawValues' => json_encode($cleanRawValues),
+                            'code' => $row['code']
+                        ], [
+                            'rawValues' => Type::STRING,
+                            'code' => Type::STRING
+                        ]
+                    );
+                    $productModelCodesToIndex[] = $row['code'];
+                    if (count($productModelCodesToIndex) % self::BATCH_SIZE === 0) {
+                        $this->getProductModelIndexer()->indexFromProductModelCodes($productModelCodesToIndex);
```
ahocquard: you should index children products also

PR #10729, opened 2019-09-16
TIP-1252 Removes deprecated docker layer caching

BitOne:

Docker layer caching will be removed for the container plan and very expensive for the Performance Plan.
So we remove it and will work on an alternative solution.

`OK`: Done / Validated
`-`: Not needed


BitOne:
3.1 not maintained anymore

PR #10742, opened 2019-09-18
PIM-7890: fix non removed category filter in job instances

nmarniesse:

Fix [PIM-7890](https://akeneo.atlassian.net/browse/PIM-7890)  
When a category or parent category is removed, the filter stay in the job instance and there is no way to remove it.  
Now when we remove a category, we also remove the filter in job instances.  

`OK`: Done / Validated
`-`: Not needed

Before, one category was still selected in the filters when we remove the category:
PIM-7890_before

After, the filter is removed automatically when the category is removed:
PIM-7890_after

### DIFF
--- a/src/Pim/Bundle/EnrichBundle/EventListener/Storage/RemoveCategoryFilterInJobInstanceSubscriber.php
+++ b/src/Pim/Bundle/EnrichBundle/EventListener/Storage/RemoveCategoryFilterInJobInstanceSubscriber.php
@@ -1,0 +1,170 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Pim\Bundle\EnrichBundle\EventListener\Storage;
+
+use Akeneo\Component\Batch\Model\JobInstance;
+use Akeneo\Component\StorageUtils\Saver\BulkSaverInterface;
+use Akeneo\Component\StorageUtils\StorageEvents;
+use Doctrine\ORM\EntityRepository;
+use Pim\Component\Catalog\Model\CategoryInterface;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\EventDispatcher\GenericEvent;
+
+/**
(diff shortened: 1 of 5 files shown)


Doodoune:
Please add a changelog ;)

src/Pim/Bundle/EnrichBundle/EventListener/Storage/RemoveCategoryFilterInJobInstanceSubscriber.php
```diff
@@ -0,0 +1,174 @@
...
+     * We can not get the child's codes of a category after delete. So when a category is going to be deleted,
+     * we get all the child's codes linked to it and keep them in $computedCodes property.
+     *
+     * @param GenericEvent $event
+     * @return RemoveCategoryFilterInJobInstanceSubscriber
+     */
+    public function computeAndHoldCategoryTreeCodes(GenericEvent $event): RemoveCategoryFilterInJobInstanceSubscriber
+    {
+        $subject = $event->getSubject();
+        if ($subject instanceof CategoryInterface) {
```
jmleroux:
```suggestion
        if (!$subject instanceof CategoryInterface) {
            return;
        }
        $this->computedCodes[$subject->getCode()] = $this->getCodeAndChildrenCodes($subject);
```
I find it simpler to early retrun and not have to test for the interface in the rest of the code.

nmarniesse: I prefer early return too, but here we can't do that. The subject can be a category or an array of categories, the method handles both cases.

pchasle: The check will be more complicated as it apparently can be an array of CategoryInterface too, but I agree with you, early return are almost always easier to understand.

PR #10745, opened 2019-09-18
TIP-1221: implement elasticsearch product projection in SQL

ahocquard:

SQL implementation of indextaion of products.

What's missing:
- [ ] optimize the creation of the value collection

`OK`: Done / Validated
`-`: Not needed


src/Akeneo/Pim/Enrichment/Bundle/Storage/Sql/ElasticsearchProjection/GetElasticsearchProductProjection.php
```diff
@@ -0,0 +1,329 @@
...
+    ) {
+        $this->connection = $connection;
+        $this->valuesNormalizer = $valuesNormalizer;
+        $this->valueCollectionFactory = $valueCollectionFactory;
+        $this->additionalDataProviders = $additionalDataProviders;
+    }
+
+    public function fromProductIdentifier(string $productIdentifier): ElasticsearchProductProjection
+    {
+        return $this->fromProductIdentifiers([$productIdentifier])[$productIdentifier];
```
mmetayer: could lead to an undefined index error if the identifier does not exist. BTW do we really need this method (seems like we do not use it anywhere)?

PR #10749, opened 2019-09-19
PIM-8777: Forbid space in username at creation

jmleroux:
Space in username can trigger bugs in the application. We found one with EE rules and chose to forbid spaces in usernames.
Ideally, we should add a validation at user level, but this cannot be done in a patch, so we decided to not create new users with spaces.

| Added acceptance tests            | ok

`OK`: Done / Validated
`-`: Not needed


tests/back/UserManagement/Integration/Bundle/Command/CreateUserCommandIntegration.php
```diff
@@ -0,0 +1,162 @@
...
+
+use Akeneo\Test\Integration\TestCase;
+use Akeneo\UserManagement\Bundle\Command\CreateUserCommand;
+use Akeneo\UserManagement\Component\Model\UserInterface;
+use Exception;
+use Symfony\Bundle\FrameworkBundle\Console\Application;
+use Symfony\Component\Console\Tester\CommandTester;
+use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
+
+class CreateUserCommandIntegration extends TestCase
```
ahocquard: this is a validation rule, it should not be tested with an end-to-end-test ideally but with acceptance test. Did you try to do it?

And it should not be tested through the cli command.

src/Akeneo/UserManagement/Bundle/Command/CreateUserCommand.php
```diff
@@ -91,6 +92,9 @@ private function askForUsername(InputInterface $input, OutputInterface $output):
             if (empty($answer)) {
                 throw new \InvalidArgumentException("The username is mandatory.");
             }
+            if (strstr($answer, ' ') !== false) {
```
ahocquard: why if it's already validated?

nmarniesse: This does not handle other "spaces" like tabs, carriage return, new lines, etc...  
Maybe something like this could be better:

```php
if (preg_match('/\s/', $answer) !== 0) {
```

tests/features/user-management/create_user.feature
```diff
@@ -0,0 +1,9 @@
+Feature: Create a new user
+  In order to manage application users
+  As an admin
+  I need to be able to create a new user
+
+  @acceptance-back
+  Scenario: Validate the username upon user creation
```
ahocquard: `
```suggestion
  Scenario: A user cannot be created when there i space in the username
```

src/Akeneo/UserManagement/Bundle/Command/CreateUserCommand.php
```diff
@@ -10,6 +10,7 @@
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Console\Question\Question;
+use Symfony\Component\Validator\ConstraintViolation;
```
pchasle: Unused import ?

src/Akeneo/UserManagement/Bundle/Resources/config/validation.yml
```diff
@@ -1,5 +1,6 @@
 Akeneo\UserManagement\Component\Model\User:
     constraints:
+        - Akeneo\UserManagement\Bundle\Validator\Constraints\CreateUser: ~
```
pchasle: Is there a reason for using a class constraint instead of a simple Regex property constraint ?

jmleroux: Because we have to add some logic to test if is a created user.

pchasle: This constraint should probably be enforced at all time, not just for the user creation. What happens if I edit the username afterward and add some spaces?

PR #8496, opened 2018-07-18
PIM-7499 - Dashboard perf

Doodoune:
**Description (for Contributor and Core Developer)**

**Definition Of Done (for Core Developer only)**

| Q                                 | A
| --------------------------------- | ---
| Added Specs                       | Todo
| Added legacy Behats               | Todo
| Added acceptance tests            | Todo
| Added integration tests           | Todo
| Changelog updated                 | Todo
| Review and 2 GTM                  | Todo
| Micro Demo to the PO (Story only) | Todo
| Migration script                  | -
| Tech Doc                          | -

`Todo`: Pending / Work in progress
`OK`: Done / Validated
`-`: Not needed

PR #10750, opened 2019-09-19
PIM-7762: Fix completeness widget

nmarniesse:

Fix [PIM-7762](https://akeneo.atlassian.net/browse/PIM-7762) completeness widget when there is a lot of products in catalog.
This is a backport of PIM-7499 (see https://github.com/akeneo/pim-community-dev/pull/8496), differences are:
- we can not use JSON_ARRAYAGG in SQL query (added in MySQL 5.7.22), so it's replaced by GROUP_CONCAT ans some adjustments
- `mquery()` on ES client is not existed here, so we launch query by query  

`OK`: Done / Validated
`-`: Not needed


src/Pim/Bundle/EnrichBundle/Storage/ElasticsearchAndSql/FollowUp/GetCompletenessPerChannelAndLocale.php
```diff
@@ -0,0 +1,306 @@
...
+                                        ],
+                                    ],
+                                ],
+                            ],
+                        ],
+                    ],
+                ],
+            ];
+
+            $result = $this->client->search($this->indexType, $body);
```
nmarniesse: Seems we can do that in one query, instead of multiple. We have to implement the `mquery()` in `Akeneo\Bundle\ElasticsearchBundle\Client` like in 3.0+

ahocquard:
I'm really sorry you worked on this backfix but I've seen it with @NolwennP and it's not validated by PO. We will not backport it.

PR #10752, opened 2019-09-20
TIP-1280: Refactor completeness tests

mmetayer:
Move tests about completeness from legacy behat to PHPUnit

`OK`: Done / Validated
`-`: Not needed


tests/back/Pim/Enrichment/Integration/Completeness/Import/ComputeCompetenessOnProductModelImportIntegration.php
```diff
@@ -0,0 +1,102 @@
...
+
+class ComputeCompetenessOnProductModelImportIntegration extends TestCase
+{
+    /**
+     * @group critical
+     */
+    public function test_that_completeness_is_computed_when_product_models_are_imported(): void
+    {
+        $tshirtBlueM = $this->get('pim_catalog.repository.product')->findOneByIdentifier('tshirt-divided-navy-blue-m');
+        Assert::assertNull($tshirtBlueM->getValue('composition'));
```
mmetayer: Note: the removed behat also tested indexation, but it's not the purpose of the test, and will be covered in another test

ahocquard:
@mmetayer I think this test is not that critical, as we have already one end-to-end in the API for that and the dedicated logic does not exist anymore in import/export. I would just delete the behats actually...

PR #10773, opened 2019-09-25
TIP-1263: upgrade Symfony to v4.3

arnolanglade:
(Post from JJ)

This PR upgrades Symfony to 4.3. Please keep in mind that our objective is to use Symfony 4.4 (which is LTS). It will be released in November. Going from 4.3 to 4.4 should be easy. The complex/time consuming migration has been done here.

## What's done here?

- upgrade all Symfony libraries to `^4.3` (we'll set them precisely if needed with the version `4.4`)
- fix circular dependency about indexing normalizers (appeared with the Symfony's upgrade)
- "clean" PhpUnit tests: 
    - no more `new WhateverService()`, we now try to use dependency injection everywhere 
    - API tests (and PhpUnit in general) do not use anymore the [debug](https://symfony.com/doc/current/configuration/front_controllers_and_kernel.html#debug-mode) mode when booting the kernel
- make the PhpUnit pass with Symfony 4: we needed to make some services public (see below)... Also, we needed to make some tests compliant with the [UrlGenerator](https://github.com/symfony/symfony/blob/4.3/src/Symfony/Component/Routing/Generator/UrlGenerator.php#L30-L41). 
- make the Behat pass with Symfony 4: we needed to make some services public (see below)...

## Weird stuff with this migration

There are a few weird stuff we have encountered with this migration:

1. `debug` mode of the kernel seems hardly usable for us sometimes. I don't why, but when you launch a PhpUnit with the kernel set to `debug=true`, compiling the container will use all the memory of your machine. Everything works perfectly fine in `dev` mode :man_shrugging: 
2. A circular dependency related to `indexing` normalizers appeared. So I guess they changed something about how the container is compiled (I don't know what exactly). The good thing is that if forced us to clean those indexers definitions (because yes, we were doing some really strange things here...) 
3. The most weird stuff is about the public/private services. As you know (well, I hope^^), services are private by default with Symfony 4. Whereas we use the container in test mode for PhpUnit (which should render all services public), we were forced to render explicitly some services `public`. For instance, imagine you have the following test:
```php
class MyTestCase extends TestCase {
   function test_it_does_something() {
       $this->get('foo')->foo(); // the service foo is private (by default) but accessible for test purpose
       $this->get('bar')->bar(); // the service bar is private (by default), but NOT accessible for test purpose.. 
   }
}
```

At first we didn't understand what happened. Then we discovered [this issue](https://github.com/symfony/symfony/issues/28528). From our understanding, it seems Symfony removes some services from the container (the ones that are not used, the ones that are inlined and the ones that are can be removed). Those removed/inlined services are not directly accessible via `$container->get()` even with `debug:true` in the `framework` configuration.

To be able to merge this PR, we decided to adopt the easy (and dirty method) : we made the service `bar` public. We'll fix later. We'll probably render all services `public` in `test` env to not bother about this kind of problems.

## Side note about private services

Services private by default has been a huge BC break for us. But I'm sure it's a bad for a greater good. It'll force to think more our design. And talking about design, tests are no exceptions. We should not access directly the container for stuff we use in every tests. We should instead create small test services. 

For instance, take a look at the [base API test case](https://github.com/akeneo/pim-community-dev/blob/master/src/Akeneo/Tool/Bundle/ApiBundle/tests/integration/ApiTestCase.php#L294-L325). The method `createAdminUser` should be itself a service (defined in `test_services.yml` for instance).
This would allow us to remove the direct accesses to container.


composer.json
```diff
@@ -124,5 +138,12 @@
         "akeneo/measure-bundle": "0.6.0-dev",
         "akeneo/batch-bundle": "0.6.0-dev"
     },
-    "minimum-stability": "stable"
+    "minimum-stability": "stable",
+    "scripts": {
```
arnolanglade: to remove ?

composer.json
```diff
@@ -72,29 +72,40 @@
...
+        "behat/behat": "3.4.3",
         "behat/gherkin":"4.5.1",
         "behat/mink":"1.7.1",
         "behat/transliterator":"1.2.0",
         "behat/mink-browserkit-driver": "1.3.3",
         "behat/mink-extension": "2.3.1",
-        "behat/mink-selenium2-driver": "^1.3.1",
-        "behat/symfony2-extension": "2.1.5",
+        "behat/mink-selenium2-driver": "1.3.x-dev",
+        "behat/symfony2-extension": "2.1.4",
```
arnolanglade: revert

.circleci/config.yml
```diff
@@ -35,7 +35,10 @@ jobs:
               sudo chown -R 1000:1000 ~/.cache/yarn
       - run:
             name: Install back dependencies
-            command: make vendor
+            command: |
+                make up
```
jjanvier: fix it

.circleci/config.yml
```diff
@@ -35,7 +35,10 @@ jobs:
               sudo chown -R 1000:1000 ~/.cache/yarn
       - run:
             name: Install back dependencies
-            command: make vendor
+            command: |
```
arnolanglade: should be reverted

composer.json
```diff
@@ -71,28 +71,39 @@
...
+        "symfony/dotenv": "^4.3",
+        "symfony/acl-bundle": " 1.0.0",
+        "symfony/debug-bundle": "^4.3",
+        "symfony/web-profiler-bundle": "^4.3",
+        "symfony/web-server-bundle": "^4.3"
     },
     "require-dev": {
         "akeneo/php-coupling-detector": "0.3.*",
-        "behat/behat": "3.5.0",
+        "behat/behat": "3.4.3",
```
arnolanglade: should be reverted

config/bootstrap.php
```diff
@@ -14,7 +14,7 @@
     throw new RuntimeException('Please run "composer require symfony/dotenv" to load the ".env" files configuring the application.');
 } else {
     $path = dirname(__DIR__).'/.env';
-    $dotenv = new Dotenv(false);
+    $dotenv = new Dotenv(true);
```
arnolanglade: true is allready the default value

PR #10775, opened 2019-09-25
TIP-1284: Create a service to compute completeness and index product models, children and ancestors

nmarniesse:

[TIP-1284](https://akeneo.atlassian.net/browse/TIP-1284): Create a service to compute completeness and index product models, children and ancestors  
Add also batch in ProductIndexer  
The class is here but the service is not declared in the DI, as the service/class will be probably an EventSubscriber in the next card.  

`OK`: Done / Validated
`-`: Not needed


src/Akeneo/Pim/Enrichment/Bundle/ProductModel/RefreshProductCompletenessesAndIndex.php
```diff
@@ -0,0 +1,63 @@
...
+     */
+    public function fromProductModelCodes(array $productModelCodes): void
+    {
+        if (empty($productModelCodes)) {
+            return;
+        }
+
+        $variantProductIdentifiers = $this->getDescendantVariantProductIdentifiers->fromProductModelCodes(
+            $productModelCodes
+        );
```
nmarniesse: The process is not very optimized as this service calls the `GetDescendantVariantProductIdentifiers::fromProductModelCodes()` query, and the service ProductModelDescendantsAndAncestorsIndexer below calls the same query. 
We can refactor in order to call the query only once, or implements a cache in the query. Don't know what is the best way.

src/Akeneo/Pim/Enrichment/Bundle/Elasticsearch/Indexer/ProductIndexer.php
```diff
@@ -60,19 +61,19 @@ public function indexFromProductIdentifiers(array $productIdentifiers, array $op
...
         $indexRefresh = $options['index_refresh'] ?? Refresh::disable();
 
-        $elasticsearchProductProjections = $this->getElasticsearchProductProjection->fromProductIdentifiers($productIdentifiers);
-        $normalizedProductProjections = array_map(
-            function (ElasticsearchProductProjection $indexableProduct) {
-                return $indexableProduct->toArray();
-            },
-            $elasticsearchProductProjections
-        );
+        foreach (array_chunk($productIdentifiers, self::BATCH_SIZE) as $batchProductIdentifiers) {
```
ahocquard: don't hesitate to do a variable for `array_chunk($productIdentifiers, self::BATCH_SIZE)`. Inlining everything is pretty difficult to debug.

src/Akeneo/Pim/Enrichment/Bundle/ProductModel/RefreshProductCompletenessesAndIndex.php
```diff
@@ -0,0 +1,63 @@
...
+ * Orchestrator for below jobs:
+ *  - Computes and saves the completenesses for the variant products of the given product models.
+ *  - Indexes these variant products
+ *  - Indexes the product models impacted by the new completeness (= ancestors and descendants of the given product models)
+ *
+ * @author    Nicolas Marniesse <nicolas.marniesse@akeneo.com>
+ * @copyright 2019 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+class RefreshProductCompletenessesAndIndex
```
ahocquard: final

I propose `ComputeProductCompletenessAndIndexProductAndAncestors` Long but explicit. I don't like refresh, which means nothing in term of business (the PM never talk refresh of the completeness)

src/Akeneo/Pim/Enrichment/Bundle/ProductModel/RefreshProductCompletenessesAndIndex.php
ahocquard: I think the best way is to wait if it's costly and benchmark, I think it's not a big deal at all. We have several thousands queries to avoid before reaching this kind of optimization.

 Adding a cache will lead to more problems than it will solve (think about long running process and memory leaks, which are very time consuming in term of maintenance).

tests/back/Pim/Enrichment/Specification/Bundle/Elasticsearch/Indexer/ProductIndexerSpec.php
```diff
@@ -160,4 +191,14 @@ private function getElasticSearchProjection(string $identifier): ElasticsearchPr
             []
         );
     }
+
+    private function getRangeIdentifiers(int $start, int $end): array
+    {
+        return array_map(
```
ahocquard: `return preg_filter('/^/', 'p_', range($start, $end));` do the job also FYI (to win one line contest :smile: )

src/Akeneo/Pim/Enrichment/Bundle/ProductModel/RefreshProductCompletenessesAndIndex.php
```diff
@@ -0,0 +1,63 @@
...
+        ProductModelDescendantsAndAncestorsIndexer $productModelDescendantsAndAncestorsIndexer,
+        GetDescendantVariantProductIdentifiers $getDescendantVariantProductIdentifiers
+    ) {
+        $this->computeAndPersistProductCompletenesses = $computeAndPersistProductCompletenesses;
+        $this->productModelDescendantsAndAncestorsIndexer = $productModelDescendantsAndAncestorsIndexer;
+        $this->getDescendantVariantProductIdentifiers = $getDescendantVariantProductIdentifiers;
+    }
+
+    /**
+     * We can make some optimizations here: if no variant products, we can only index product models
```
ahocquard: It's more or less true, but I would rather say that the killer optimization is to:
- avoid to calculate the completeness if no required attribute has changed: so no indexation of the ancestors also
- avoid to calculate the products if no values, association, categories, etc has changed

So, it is to have business events. Then, with some discussions with PM, the optimization will be pretty obvious I think. But adding business events is a big topic. I think we should not try to optimize another way, it will only be "bricole", adding more complexity and *very big* maintenance effort + hard refactor later on (here it is very simple).

src/Akeneo/Pim/Enrichment/Bundle/ProductModel/RefreshProductCompletenessesAndIndex.php
```diff
@@ -0,0 +1,63 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Akeneo\Pim\Enrichment\Bundle\ProductModel;
```
ahocquard: Btw I think this is component stuff. And  Akeneo\Tool\Component\StorageUtils\Indexer\ProductIndexerInterface should in component of the enrichment.

nmarniesse: I refactor that in new commit. Please check if all is correct

nmarniesse: Finally rollback the namespace change for ComputeProductCompletenessAndIndexProductAndAncestors. There is some coupling issues so I guess we need to harmonize all at the end of the sprint.

tests/back/Pim/Enrichment/Specification/Bundle/Elasticsearch/Indexer/ProductIndexerSpec.php
```diff
@@ -160,4 +191,9 @@ private function getElasticSearchProjection(string $identifier): ElasticsearchPr
             []
         );
     }
+
+    private function getRangeIdentifiers(int $start, int $end): array
+    {
+        return preg_filter('/^/', 'p_', range($start, $end));
```
mmetayer: :smile:

PR #10786, opened 2019-09-26
TIP-1262 Make context services public to avoid failure in SF4.3

arnolanglade:

In symfony 4.0 all services are private by default that means our Behat services are private too. Behat Symfony extension need to get those services that why we declare them as public.

`OK`: Done / Validated
`-`: Not needed


ahocquard:
Just an idea: is it not possible to use test kernel instead to have everything public?

https://github.com/symfony/framework-bundle/blob/master/Test/WebTestCase.php#L50

PR #10792, opened 2019-09-26
TIP-1297: Use LRU cache for valueCollectionNormalization

pierallard:


src/Akeneo/Pim/Structure/Bundle/Query/PublicApi/Attribute/Sql/SqlGetAttributes.php
```diff
@@ -51,7 +51,8 @@ public function forCodes(array $attributeCodes): array
                 boolval($rawAttribute['is_localizable']),
                 boolval($rawAttribute['is_scopable']),
                 $rawAttribute['metric_family'],
-                boolval($rawAttribute['decimals_allowed'])
+                boolval($rawAttribute['decimals_allowed']),
+                $rawAttribute['backend_type']
```
ahocquard: you can add comma in function call to avoid this kind of diff now: https://kinsta.com/blog/php-7-3/

I don't know if it's really useful though

PR #10797, opened 2019-09-27
TIP-1262: proper dependency injection in the test services

jjanvier:
When trying to upgrade to SF4, we face a problem: all services are now private by default. In the test, you can bypass that problem by setting `test: true` in the `framework` configuration.

In our PhpUnit tests, we have a weird "double kernel booted". We sometimes inject `@service_container` or `@kernel` to retrieve some services inside the test services. It seems, but we are not sure, that those injected containers are not built with the `test` configuration set to `true`.

Anyway, even if this PR doesn't fix our problem, it's better to properly inject only what we need.

### DIFF
--- a/src/Akeneo/Tool/Bundle/ApiBundle/tests/integration/ApiTestCase.php
+++ b/src/Akeneo/Tool/Bundle/ApiBundle/tests/integration/ApiTestCase.php
@@ -45,16 +45,17 @@ abstract class ApiTestCase extends WebTestCase
     protected function setUp(): void
     {
         static::bootKernel(['debug' => false]);
-        $authenticator = new SystemUserAuthenticator(static::$kernel->getContainer());
-        $authenticator->createSystemUser();
 
         $this->testKernel = new \AppKernelTest('test', false);
         $this->testKernel->boot();
 
-        $this->catalog = $this->testKernel->getContainer()->get('akeneo_integration_tests.configuration.catalog');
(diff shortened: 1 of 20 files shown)


src/Akeneo/Tool/Bundle/ApiBundle/tests/integration/ApiTestCase.php
```diff
@@ -45,16 +45,22 @@ abstract class ApiTestCase extends WebTestCase
...
     {
         static::bootKernel(['debug' => false]);
-        $authenticator = new SystemUserAuthenticator(static::$kernel->getContainer());
-        $authenticator->createSystemUser();
 
         $this->testKernel = new \AppKernelTest('test', false);
         $this->testKernel->boot();
 
-        $this->catalog = $this->testKernel->getContainer()->get('akeneo_integration_tests.configuration.catalog');
+        $authenticator = new SystemUserAuthenticator(
```
ahocquard: I think you should do it *after* the initialization of your catalog.
Reason: when you use the user, it load the activated locale for this user from database. And so, the locale of the previous catalog, leading to weird bug that can take days to solve.

See https://github.com/akeneo/pim-community-dev/blob/master/tests/back/Integration/TestCase.php#L50

But the user default locale is not used in the API, so not a big deal.

jjanvier: ok fixed

PR #10808, opened 2019-09-30
PIM-8820: Avoid multiple refresh of the product grid on product delete

pierallard:
If you go to the product grid, then PEF, then product grid, it will register 2 times the events of the datagrid and refresh 2 times the grid on product deletion.
This PR removes the previous events before adding new ones.


src/Oro/Bundle/PimDataGridBundle/Resources/public/js/datagrid/grid.js
```diff
@@ -287,6 +287,11 @@ define(
             _listenToCommands: function () {
```
tamarasaurus: When does this function get called?

pierallard: initialize()

PR #10811, opened 2019-09-30
AOB-638: Upgrade UI

momoss:

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Akeneo/Platform/Bundle/UIBundle/Resources/public/less/components/grid/Grid.less
+++ b/src/Akeneo/Platform/Bundle/UIBundle/Resources/public/less/components/grid/Grid.less
@@ -585,3 +585,10 @@
     }
   }
 }
+
+.AknGrid-headerCell,
+.AknGrid-bodyCell {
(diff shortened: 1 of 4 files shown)


src/Akeneo/Platform/Bundle/UIBundle/Resources/public/less/components/grid/Grid.less
```diff
@@ -585,3 +585,10 @@
     }
   }
 }
+
+.AknGrid-headerCell,
+.AknGrid-bodyCell {
+  &-important {
+    background-color: lighten(@AknBlue, 41%);
+  }
+}
```
phaseinducer: Are you sure we need to do this in the CE ?

momoss: Yes, because in this screen it's the pim css who is loaded, it has been seen with the BEM master Peter, it's ok for him to do things like that ;)

PR #10814, opened 2019-10-01
PIM-8838: Force display of label and image in gallery mode even if they are not in the column list

pierallard:
It does not change the display of the standard grid because only the "values" are displayed.
It only always return the image and label "properties" in the product grid.


jmleroux:
check frozen

PR #10815, opened 2019-10-01
TIP-1262: remove AppKernelTest

jjanvier:
Currently, when we launch phpunit tests, we use a double kernel (and thus container) system:
- the first kernel, the class `AppKernelTest`, is used _mostly_ to load fixtures in the `setup` method
- the second, the regular `Kernel` class, is used for almost everything else (but it depends how the test is written)

This trick is not really needed and makes the understanding/debugging of the tests complex. By reading a few tests, I feel it was even understood how to use it. This PR aims to remove it.

The first commits are about removing this `AppKernelTest` class. The others are about fixing the failing tests due to this change. The failing tests were mostly about:
- Now, as we have only one kernel, that means we have only one container. Which means we have only one Doctrine manager. Which means some data we have loaded in the fixtures are now maybe in a cache somewhere (like a cached repository for instance) when we really execute the test.
- Permissions problems. Fixtures were setup as an admin. And the test was supposed to be executed with another user. 
- ...

Another thing done in this PR is that I used real injection (instead of injecting the container directly) for the `akeneo_test_integration*` services.


jjanvier:
GTM for your commits @arnolanglade :)

PR #10818, opened 2019-10-01
PIM-8331: Fix display of category tree in the product grid when user does not have access to its default tree

pierallard:
The called method was not aware of any permission to get the default catalog tree.
It implies a bug on front.


src/Akeneo/Pim/Enrichment/Component/Category/CategoryTree/UseCase/ListRootCategoriesWithCountHandler.php
```diff
@@ -56,7 +56,7 @@ public function handle(ListRootCategoriesWithCount $query): array
             $this->categoryRepository->find($query->categoryIdSelectedAsFilter()) : null;
 
         if (null === $categorySelectedAsFilter) {
-            $categorySelectedAsFilter = $this->userContext->getUserProductCategoryTree();
+            $categorySelectedAsFilter = $this->userContext->getAccessibleUserTree();
```
ahocquard: This method does not exist in CE. UserContext is not overrider in EE... (it should be a dedicated query with an interface in my opinion).

PR #10821, opened 2019-10-01
TIP-1262 catalog installation

arnolanglade:

`OK`: Done / Validated
`-`: Not needed


jjanvier:
shouldn't it be an env var instead?
because it's kind independent of the SF env

arnolanglade:
Dunno, we would introduce an env but do we need to load ice cat elsewhere in DEV env?

arnolanglade:
Close in favor of https://github.com/akeneo/pim-community-dev/pull/10860

PR #10827, opened 2019-10-02
PIM-7332: show an error message when a number attribute field reaches the PHP_INT_MAX

willy-ahva:
It adds an error message when you edit a product and you try to put a number higher than PHP_INT_MAX.
It adds an error message when you edit an attribute and you try to put as the maximum a number higher than PHP_INT_MAX. (It was the case before but now you have an explicit error message) .

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Pim/Component/Catalog/Validator/ConstraintGuesser/RangeGuesser.php
+++ b/src/Pim/Component/Catalog/Validator/ConstraintGuesser/RangeGuesser.php
@@ -49,6 +49,10 @@ class RangeGuesser implements ConstraintGuesserInterface
             }
         }
 
+        if (AttributeTypes::NUMBER === $attribute->getType() && (null === $max || PHP_INT_MAX < $max)) {
+            $max = PHP_INT_MAX;
+        }
+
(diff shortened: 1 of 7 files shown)


src/Pim/Component/Catalog/Validator/ConstraintGuesser/RangeGuesser.php
```diff
@@ -49,6 +49,10 @@ public function guessConstraints(AttributeInterface $attribute)
             }
         }
 
+        if (AttributeTypes::NUMBER === $attribute->getType() && (null === $max || PHP_INT_MAX < $max)) {
+            $max = PHP_INT_MAX;
```
pierallard: Do you think you can have issue with minimal at -PHP_INT_MAX ?

willy-ahva: Nope because if the minimum is greater or equal to the maximum an error is raised. The maximum is PHP_INT_MAX even if you add nothing.

jmleroux:
check frozen

PR #10842, opened 2019-10-04
PIM-8866: Merge 2.3 into 3.0

jmleroux:


src/Pim/Bundle/CatalogBundle/Doctrine/ORM/Query/GetChannelActiveLocaleCodes.php
```diff
@@ -0,0 +1,47 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of the Akeneo PIM Enterprise Edition.
```
pierallard: Wrong header

PR #10889, opened 2019-10-15
PIM-7691: Users were able to edit their own roles

tseho:

Users without the permissions `Edit roles` or `Edit user groups` could escalate their own role from `User` to `Administrator`. Submitted roles and groups were only ignored in the controller of `updateProfileAction`.
The solution is to use the SecurityFacade service to check if the logged user has the correct permissions, if it's not the case, submitted roles & groups are ignored.

Same as https://github.com/akeneo/pim-community-dev/pull/10868 but the current PR targets 3.0.x

| Added legacy Behats               | yes

### DIFF
--- a/src/Akeneo/UserManagement/Bundle/Controller/Rest/UserController.php
+++ b/src/Akeneo/UserManagement/Bundle/Controller/Rest/UserController.php
@@ -11,6 +11,7 @@ use Akeneo\UserManagement\Component\Model\UserInterface;
 use Doctrine\Common\Persistence\ObjectManager;
 use Doctrine\Common\Persistence\ObjectRepository;
 use Oro\Bundle\SecurityBundle\Annotation\AclAncestor;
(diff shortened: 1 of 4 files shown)


src/Akeneo/UserManagement/Bundle/Controller/Rest/UserController.php
```diff
@@ -80,10 +81,13 @@ class UserController
     /** @var TranslatorInterface */
     private $translator;
 
+    /** @var SecurityFacade|null */
+    private $securityFacade;
+
     /**
      * @todo merge 3.2:
```
jmleroux: `merge master`. 3.2 is already tagged and should not have BC breaks

jmleroux:
Little comment then GTM

src/Akeneo/UserManagement/Bundle/Controller/Rest/UserController.php
tseho: done

PR #10894, opened 2019-10-16
AOB-661: allow prefix of ES document _id

docteurklein:


docteurklein:
> you don't have spec to change on the client?

not even! seems like the default value of the constructor argument is enough.

docteurklein:
> As seen together IRL, it will not work in this current state with the search after method if you expose the API later on.

in 3.2:
maybe. it's still not clear where it poses problem, can't find any problem by reading the code, but we need to test it in CE already.

docteurklein:
tested on CE with a prefix: seems OK

PR #10910, opened 2019-10-18
PIM-7963: fix datepicker width not adapting to the dropdown

ValentinMumble:

The datepicker was not adapting its width when changing the month or the year, I added a lib customization to fix it.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Pim/Bundle/UIBundle/Resources/public/less/lib/bootstrap.datepicker.less
+++ b/src/Pim/Bundle/UIBundle/Resources/public/less/lib/bootstrap.datepicker.less
@@ -1,0 +1,3 @@
+.datepicker table {
+  width: 100%;
+}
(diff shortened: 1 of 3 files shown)


src/Pim/Bundle/UIBundle/Resources/public/less/lib/bootstrap.datepicker.less
```diff
@@ -0,0 +1,3 @@
+.datepicker table {
```
tseho: does the table element has a css class ?

ValentinMumble: yes but it can be specific, like `.table-condensed` for instance

PR #10911, opened 2019-10-18
PIM-8893: update dompdf because the pdf export fail on some products

tseho:

We update dompdf because the pdf export fail on some products on a known error fixed in 0.8.3,
see https://github.com/dompdf/dompdf/issues/902

`OK`: Done / Validated
`-`: Not needed


src/Akeneo/Pim/Enrichment/Bundle/PdfGeneration/Builder/DompdfBuilder.php
```diff
@@ -2,6 +2,8 @@
 
 namespace Akeneo\Pim\Enrichment\Bundle\PdfGeneration\Builder;
 
+use Dompdf\Dompdf;
```
LaurentPetard: You have to add a rule in `src/Akeneo/Pim/Enrichment/.php_cd.php` to avoid a coupling error (see the CI result)

PR #10943, opened 2019-10-22
Fix integration tests

fitn:

`OK`: Done / Validated
`-`: Not needed


src/Akeneo/Apps/back/tests/Integration/Client/Fos/CreateClientIntegration.php
```diff
@@ -0,0 +1,35 @@
+<?php
+declare(strict_types=1);
+
+namespace Akeneo\Apps\back\tests\Integration\Client\Fos;
```
pchasle: What is the Fos namespace?

fitn: It means FriendOfSymfony. It represents the current implementation of client generation. I am not a big fan of this name but however, I feel it is the right one. What's your opinion?

pchasle: It's what I though, fine by me but not very explicit especially as the test doesn't mention directly the Fos client.

PR #10946, opened 2019-10-23
PIM-8906: Fix file storage UUID generator

jmleroux:
Actual storage UUID is `sha1(filename . microsecond())`

We ran into issue with API with more than 1000 calls per second because of duplicated UUID.

This PR uses real random UUID and we will not rely on the creation time anymore.

### DIFF
--- a/CHANGELOG-3.0.md
+++ b/CHANGELOG-3.0.md
@@ -1,5 +1,9 @@
 # 3.0.x
 
+# Improvements
+
+- PIM-8906: Fix file storage UUID generator
+
 # 3.0.48 (2019-10-23)
 
 ## Bug fixes
--- a/src/Akeneo/Tool/Component/FileStorage/PathGenerator.php
+++ b/src/Akeneo/Tool/Component/FileStorage/PathGenerator.php
@@ -2,6 +2,7 @@
 
 namespace Akeneo\Tool\Component\FileStorage;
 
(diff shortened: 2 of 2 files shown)


ahocquard:
I personally don't GTM because I'm very skeptical about the usecase. Is it raised by a customer at least?

The ticket should be "Try to generate more than 1000 time the SAME assets in the same seconds with the API". I don't even know how it's possible to handle such an amount of requests for a classical PIM installation...

jmleroux:
> I personally don't GTM because I'm very skeptical about the usecase. We are opening the door of weird fixes. Is it raised by a customer at least?
> 
> The ticket should be "Try to generate more than 1000 time the SAME assets in the same seconds with the API". I don't even know how it's possible to handle such an amount of requests for a classical PIM installation...

Yes it is weird and yes it is raised by a customer.
However, it's not required at Service Level Agreement because we simply advised to not do so much request for the same asset.
So this PR is just a suggestion and you're totally justified at refusing it.
Let's see if we have other opinions.

ahocquard:
@jmleroux I read the whole SDS ticket and I'm pretty sure that it's not the cause of the problem. The problem occurs in the same request. I would bet on a persitence problem :wink: (Doctrine, etc)

jmleroux:
@ahocquard 
What would you think if i do this PR on master ?

PR #9582, opened 2019-02-15
PIM-8127: Be able to save and launch job even if products filters are broken

willy-ahva:
**Description (for Contributor and Core Developer)**

Today, if you add a family in a filter and remove the family, the job can not be editable and can not be launched.

Seen with PO, filters of Job Parameters must not be validated. If a filter has an object which is not found (a family code or whatever) this filter must be ignored and job launched.
The job instance must be editable too.

It has already been done for attributes here: https://github.com/akeneo/pim-community-dev/pull/8667

**Definition Of Done (for Core Developer only)**

| Q                                 | A
| --------------------------------- | ---
| Added Specs                       | OK
| Added legacy Behats               | -
| Added acceptance tests            | -
| Added integration tests           | OK
| Changelog updated                 | OK
| Review and 2 GTM                  | Todo
| Micro Demo to the PO (Story only) | Todo
| Migration script                  | -
| Tech Doc                          | -

`Todo`: Pending / Work in progress
`OK`: Done / Validated
`-`: Not needed

PR #10954, opened 2019-10-24
PIM-8903: Fix export job instances filters when a family is deleted

LaurentPetard:
The problem occurs when a family is deleted and there are export profiles that have this family as content's filter.

When these export profiles are launched, the filter on families is ignored even if there are other families in it. Moreover, the code of the deleted family is kept in the filter even if the profile is updated,  whereas the deleted family is not displayed.

This is a consequence of this previous fix: https://github.com/akeneo/pim-community-dev/pull/9582

The chosen solution (discussed with PM) is to fix the affected job instances when a family is deleted. It's acceptable because it's not a common use case, and there can't be thousands of export profiles having this family as filter at the same time.


src/Akeneo/Pim/Structure/Bundle/EventSubscriber/RemoveFamilyFromJobInstanceFiltersOnFamilyDeletionSubscriber.php
```diff
@@ -0,0 +1,94 @@
...
+        if (!$family instanceof FamilyInterface) {
+            return;
+        }
+
+        $familyCode = $family->getCode();
+        $jobInstances = $this->jobInstanceRepository->findBy(['type' => JobInstance::TYPE_EXPORT]);
+        $jobsToUpdate = [];
+
+        /** @var JobInstance $jobInstance */
+        foreach ($jobInstances as $jobInstance) {
```
LaurentPetard: Ugly algo, but it allows to loop only once on the job instances and their filters

src/Akeneo/Pim/Structure/Bundle/EventSubscriber/RemoveFamilyFromJobInstanceFiltersOnFamilyDeletionSubscriber.php
```diff
@@ -0,0 +1,94 @@
...
+            $rawParameters = $jobInstance->getRawParameters();
+            foreach ($rawParameters['filters']['data'] ?? [] as $filterIndex => $filter) {
+                if ($this->isFamilyInFilter($familyCode, $filter)) {
+                    $updatedRawParameters = $this->removeFamilyCodeFromRawParameters($familyCode, $rawParameters, $filterIndex);
+                    $jobInstance->setRawParameters($updatedRawParameters);
+                    $jobsToUpdate[] = $jobInstance;
```
nmarniesse: There is no case where we need to remove the job? (If there is no more family code for example)

LaurentPetard: Nope, an export job can have no filters (even if it's not recommended)

PR #10959, opened 2019-10-24
TIP-1327 Fix docker script (wait for elasticsearch)

arnolanglade:

Since few day we have used `wait_docker_up.sh` in the MakeFile but some targets have been broken (pim-XXX, for example). This script checks that MySQL and Elasticsearch are alive. To know if ES is up we need to call an URL: `http://elasticsearch:9200/_cat/health` and find the word green. ES provides a better service: "http://elasticsearch:9200/_cluster/health?wait_for_status=green`. We don't need to use `grep` anymore. 

The main problem is that we use `docker-composer run` to call this URL instead  `docker-composer exec`. `docker-composer run` boots a container but  `docker-composer exec` uses the existing container.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/docker/wait_docker_up.sh
+++ b/docker/wait_docker_up.sh
@@ -19,7 +19,7 @@ echo "MySQL server is running!"
 
 COUNTER=1
 echo "Waiting for Elasticsearch server…"
-while ! docker-compose run --rm elasticsearch curl http://elasticsearch:9200/_cat/health | grep green > /dev/null 2>&1; do
+while ! docker-compose exec elasticsearch curl -s -k --fail "http://elasticsearch:9200/_cluster/health?wait_for_status=green&timeout=1s" > /dev/null; do
     COUNTER=$((${COUNTER} + 1))
     if [ ${COUNTER} -gt ${MAX_COUNTER} ]; then
         echo "We have been waiting for Elasticsearch too long already; failing." >&2


docker/wait_docker_up.sh
```diff
@@ -19,7 +19,7 @@ echo "MySQL server is running!"
 
 COUNTER=1
 echo "Waiting for Elasticsearch server…"
-while ! docker-compose run --rm elasticsearch curl http://elasticsearch:9200/_cat/health | grep green > /dev/null 2>&1; do
+while ! docker-compose exec elasticsearch curl "http://elasticsearch:9200/_cluster/health?wait_for_status=green&timeout=1s" > /dev/null; do
```
ahocquard: timeout=1s seems too low, I would keept the default value 30sec no ?

docker/wait_docker_up.sh
```diff
@@ -19,7 +19,7 @@ echo "MySQL server is running!"
 
 COUNTER=1
 echo "Waiting for Elasticsearch server…"
-while ! docker-compose run --rm elasticsearch curl http://elasticsearch:9200/_cat/health | grep green > /dev/null 2>&1; do
+while ! docker-compose exec elasticsearch curl "http://elasticsearch:9200/_cluster/health?wait_for_status=green" > /dev/null; do
```
nmarniesse: Here potentially we can wait 45*30sec = 22 minutes!  
IMHO either we use timeout 1 sec and we keep the while loop (like you did before), or we use timeout=45s and remove the loop.

arnolanglade: OK we need to keep the loop because if the container is not up the script will fail.

ahocquard:
@arnolanglade I doubt it works as it is. The command `docker-compose exec elasticsearch curl "http://elasticsearch:9200/_cluster/health?wait_for_status=green"` will always return a code error 0, even if the status is red. Because it's the return code of "curl". We still need the grep I think.

arnolanglade:
You're right I will add the option `--fail` to make the command fails.

PR #10963, opened 2019-10-24
PIM-8894: Change product identifier validation to forbid surrounding spaces

LaurentPetard:
All the characters except comma “,“ and semicolon “;“ are accepted for the product identifiers. If there is a space at the beginning or end of an identifier, it will be saved with it in the database.

The problem, in this case, is that the trailing spaces are ignored when we search a product by its identifier. It’s due to MySQL where all CHAR, VARCHAR, and TEXT values are compared without regard to any trailing spaces (it's according to the collation, and all the collations in MySQL 5.x have this behavior).

For instance, with an existing product with the identifier "foo". If a product is imported with the identifier " foo", a new product will be created, but with an identifier "foo ", the product "foo" will be updated. If there's a space at the beginning, the product repository doesn't find the existing product. But if there's a trailing space, the product repository finds the existing product.

As discussed with the PMs, the chosen solution is to change the validation of the product identifier (and the product model code) to forbid spaces at the beginning and at the end. The existing products having this problem can no longer be updated without fixing their identifiers. We chose to not fix that ourselves by a database migration script, to let the clients decide what to do with these identifiers (especially if there is duplication)

### DIFF
--- a/tests/legacy/features/pim/enrichment/product/create_product.feature
+++ b/tests/legacy/features/pim/enrichment/product/create_product.feature
@@ -32,26 +32,3 @@ Feature: Product creation
     Then I should be on the product "caterpillar_1" edit page
     And I should see the text "caterpillar_1"
     And I should see the text "Family sandals"
-
-  Scenario: Fail to create a product with an already used code
-    Given I create a product
-    And I fill in the following information in the popin:
-      | SKU | sandals |
-    And I press the "Save" button in the popin
-    Then I should see validation error "The same identifier is already set on another product"
-
-  @jira https://akeneo.atlassian.net/browse/PIM-4706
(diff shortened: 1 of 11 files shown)


tests/legacy/features/pim/enrichment/product/create_product.feature
```diff
@@ -32,26 +32,3 @@ Feature: Product creation
     Then I should be on the product "caterpillar_1" edit page
     And I should see the text "caterpillar_1"
     And I should see the text "Family sandals"
-
-  Scenario: Fail to create a product with an already used code
```
LaurentPetard: Now that the validation of the product identifier is covered by an acceptance test, there's no need to keep integration and legacy end-to-end tests for that.

PR #10974, opened 2019-10-28
TIP-1350: Warmup cached queries in external API controllers

mmetayer:

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Akeneo/Pim/Enrichment/Bundle/Storage/Sql/Cache/DummyWarmupQueryCache.php
+++ b/src/Akeneo/Pim/Enrichment/Bundle/Storage/Sql/Cache/DummyWarmupQueryCache.php
@@ -1,0 +1,22 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Akeneo\Pim\Enrichment\Bundle\Storage\Sql\Cache;
+
+use Akeneo\Tool\Bundle\ApiBundle\Cache\WarmupQueryCache;
+use Symfony\Component\HttpFoundation\Request;
+
+/**
+ * @copyright 2019 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+final class DummyWarmupQueryCache implements WarmupQueryCache
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function fromRequest(Request $request): void
+    {
+    }
+}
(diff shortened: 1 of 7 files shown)


src/Akeneo/Tool/Bundle/ApiBundle/Cache/WarmupQueryCache.php
```diff
@@ -0,0 +1,21 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Akeneo\Tool\Bundle\ApiBundle\Cache;
+
+use Symfony\Component\HttpFoundation\Request;
+
+/**
+ * Warms up LRU cached queries in order to minimize the number of SQL queries performed towards the DB.
```
ahocquard: I propose this to expose the drawbacks:

```suggestion
 * When importing product/product model, a lot of Database queries are triggered. We suffer of the 1+n queries problem which drastically impact the performance.
 * @ see https://www.sitepoint.com/silver-bullet-n1-problem/
 * 
 * One way to avoid that is to pre-load the need data in 1 query instead of 1 +n queries. This trick increase drastically the performance. There are two drawbacks though:
 * - it means we know the needed subrequests in sub-layers, meaning that we have knowledge of an internal behavior. Therefore, changing the implementation of the internal behavior can make this cache warmup obsolete.
 
Our abstraction to import products is leaky: https://www.joelonsoftware.com/2002/11/11/the-law-of-leaky-abstractions/
 
 * - the queries are cached, and so stateful. Do note that this was already the case even before this class, which mitigate this drawback
 
 Early analyzis of the initial request allows to gather the queries' arguments (e.g attribute codes).
```

src/Akeneo/Pim/Enrichment/Bundle/Storage/Sql/Cache/DummyWarmupQueryCache.php
```diff
@@ -0,0 +1,22 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Akeneo\Pim\Enrichment\Bundle\Storage\Sql\Cache;
```
ahocquard: I would not choose this namespace, as we rely on Request and it's not really an adapter. I propose this namespace even if it's not a controller:

```
src/Akeneo/Pim/Enrichment/Bundle/Controller/ExternalApi/
``` 

I didn't find a better place.

src/Akeneo/Tool/Bundle/ApiBundle/Cache/WarmupQueryCache.php
```diff
@@ -0,0 +1,29 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Akeneo\Tool\Bundle\ApiBundle\Cache;
```
ahocquard: Personally, I would keep it in `Pim\Enrichment\Controller\ExternalApi`, it's not code that we want to be shared with other teams

PR #10984, opened 2019-10-28
TIP-1327: Introduce new targets to run our tests (locally and on the CI)

arnolanglade:
### Description

This PR introduces some new targets to run our tests, they are located in `make-file/test.mk`. At the end the purpose of this PR and the next ones is to be able to run tests the same way on the CI and locally. We won't rework the targets phpunit and behat-legacy in this PR. We want to keep it as simpler as possible for now.

For your information this new file contains the following targets:

   * coupling: checks if the bounded context are well "de-coupled" (moved from `make-file/dev.mk`)
   * check-pullup: checks is there are not mistake during the pull up
   * lint-back and lint-front analyzes source code to flag programming errors, bugs, stylistic errors, and suspicious constructs. [Why this name?](https://en.wikipedia.org/wiki/Lint_(software))
   * unit-back and unit-front: run all unit tests (back and front)
   * acceptance-back and acceptance-front run all acceptance tests (back and front)
   * integration-front: run front integration tests

Those targets will be used on the CI to run tests but you can use them locally (you might need to run unit tests, for instance).

### How should we build those targets?

`make-file/test.mk` contains the main target (which run all tests) and there are one other file per bounded context (`asset-manager.mk`, `franklin-insights.mk`, etc).

Let's take an example: `coupling-back`.
Each team needs to build its own target named `XXX-coupling-back` where XXX is the name of the bounded context.

For instance, asset manager:
```
.PHONY: asset-manager-coupling-back
asset-manager-coupling-back:
	$(PHP_RUN) vendor/bin/php-coupling-detector detect "command option"
```
Then the target coupling-back will depend on all bounded context coupling targets.
```
.PHONY: coupling
coupling: reference-entity-coupling asset-manager-coupling franklin-insights-coupling #etc.
```
### How should we use those targets?

For instance, I am a member of the raccoon team and I am working on the asset manager bounded context. I am designing domain objects so I only want to run my unit tests. So I will define a target if it does not already exit in the make file `make-file/asset-manage.mk` named asset-manager-unit-back which will run my specs. Then I just need to use make asset-manager-unit-back to run the pecs of my bounded context. If my bounded context is well decoupled I don't need to run unit tests from other bounded contexts.

### What is the perfect world?

For the next days we will introduce the rest of the missing targets in the `make-file/test.mk` and then we/you will create the missing targets in the file bounded context make files.

### Are our tools perfect?

No, for instance PHPSpec does not manage multiple output formats like Behat does. To be able to enable junit format only on the CircleCi an new environment variable named CI have been added. Its default value is 0.

This var will allow to configure our test tools on the CI. To ease the creation of the Makefile targets functions have been created. Please, have a look to https://github.com/akeneo/pim-enterprise-dev/pull/7155/files#diff-2f76e25665bd6978e8952ff248835785R43

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/make-file/test.mk
+++ b/make-file/test.mk
@@ -1,0 +1,41 @@
+.PHONY: coupling
+coupling: structure-coupling user-management-coupling channel-coupling enrichment-coupling
+
+.PHONY: check-pullup
+check-pullup:
+	${PHP_RUN} bin/check-pullup
+
+### Lint tests
+.PHONY: lint-back
+lint-back: var/cache/dev
+	${PHP_RUN} vendor/bin/phpstan analyse src/Akeneo/Pim -l 1
+	${PHP_RUN} vendor/bin/php-cs-fixer fix --diff --dry-run --config=.php_cs.php
+
+.PHONY: lint-front
+lint-front:
+	$(YARN_RUN) lint
+
+### Unit tests
+.PHONY: unit-back
+unit-back:
+	${PHP_RUN} vendor/bin/phpspec run --format=junit > var/tests/phpspec/specs.xml
+	./.circleci/find_non_executed_phpspec.sh
+
+.PHONY: unit-front
+unit-front:
+	$(YARN_RUN) unit
+
+### Acceptance tests
+.PHONY: acceptance-back
+acceptance-back:
+	${PHP_RUN} vendor/bin/behat --strict -p acceptance -vv
+
+.PHONY: acceptance-front
+acceptance-front:
+	MAX_RANDOM_LATENCY_MS=100 $(YARN_RUN) acceptance run acceptance ./tests/features
+
+### Integration tests
+.PHONY: integration-front
+integration-front:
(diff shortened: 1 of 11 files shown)


make-file/test.mk
```diff
@@ -0,0 +1,77 @@
...
+	${PHP_RUN} vendor/bin/phpspec run $(O)
+	$(call execute_on_ci_only, .circleci/find_non_executed_phpspec.sh)
+
+.PHONY: unit-front
+unit-front:
+	$(YARN_RUN) unit
+
+### Acceptance tests
+.PHONY: acceptance-back
+acceptance-back: apps-acceptance-back
```
fitn: What about moving the "apps-acceptance-back" call in a new line and align with all the stuff we launched for acceptance-back. It makes things more readable to me.
And same remarks for all the commands (apps-unit-back, etc.). I did not find it directly when looking at the PR.

PR #10987, opened 2019-10-29
PIM-7027: fix completeness visibility on PEF

nmarniesse:

When we change locale on PEF and the ratio can not be calculated on this locale, completeness widget element was removed. So it can be display again when we change back to a locale with ratio. The element was removed to fix design issues. Here the solution is to hide it.

`OK`: Done / Validated
`-`: Not needed


nmarniesse:
Waiting for a PM validation on some point

PR #11014, opened 2019-11-04
Fix circular reference

fitn:

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -98,7 +98,8 @@
         "symfony/web-profiler-bundle": "^4.3",
         "symfony/web-server-bundle": "^4.3",
         "symfony/lock": "^4.3",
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -98,7 +98,8 @@
         "symfony/web-profiler-bundle": "^4.3",
         "symfony/web-server-bundle": "^4.3",
         "symfony/lock": "^4.3",
-        "symfony/yaml": "^4.3"
+        "symfony/yaml": "^4.3",
+        "symfony/proxy-manager-bridge": "^4.3"
```
nmarniesse: Think update the composer.lock should be done too

fitn: Yes I've just done it! Thanks :)

PR #11017, opened 2019-11-05
PIM-7851: Disable session on stateless firewalls

pchasle:

Sets the session in the request only if the security firewall is not stateless.

The default SessionListener doesn't care about the authentication and only use the 'session' setting of the framework.

In the same way, the 'stateless' setting of the firewall only affect the use of the session for authentication purpose.

`OK`: Done / Validated
`-`: Not needed


pchasle:
Closing this as the source issue is already fixed on v2.3 and the enhancement necessary for a true stateless API need further investigation.

PR #11021, opened 2019-11-05
TIP-1327: Use make to run behat legacy tests

arnolanglade:

Please have a look to this [PR](https://github.com/akeneo/pim-community-dev/pull/10984).

This PR adds a new target `end-to-end-legacy-back` to run behat legacy and end to end tests with behat.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -87,17 +87,17 @@ jobs:
                 when: always
           - run:
                 name: Analyzes source code to flag programming errors, bugs, stylistic errors, and suspicious constructs
-                command: make lint-back CI=1
+                command: make lint-back
           - run:
                 name: Code Coupling Detection
-                command: make coupling-back CI=1
+                command: make coupling-back
                 when: always
           - run:
                 name: Unit tests
-                command: make unit-back CI=1
+                command: make unit-back
           - run:
                 name: Acceptance tests
-                command: make acceptance-back CI=1
+                command: make acceptance-back
 
                 when: always
           - store_test_results:
@@ -128,10 +128,10 @@ jobs:
                 command: APP_ENV=test make database
           - run:
                 name: PHPunit Integration
-                command: make integration-back CI=1
+                command: make integration-back
           - run:
                 name: PHPunit End to end
-                command: make end-to-end-back CI=1
+                command: make end-to-end-back
           - store_test_results:
                 path: var/tests/phpunit
           - store_artifacts:
@@ -169,7 +169,7 @@ jobs:
             command: APP_ENV=behat make database
       - run:
             name: Behat legacy
-            command: make end-to-end-legacy SUITE=$TESTSUITE CI=1
+            command: make end-to-end-legacy SUITE=$TESTSUITE
       - run:
           name: Gather Junit test result files in the same directory to improve the render of failing tests
           command: |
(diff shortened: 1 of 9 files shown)


make-file/apps.mk
```diff
@@ -1,3 +1,57 @@
...
+#
+# Which kind of tests do/should we have in the PIM?
+# =================================================
+# - lint (back and front)
+# - unit (back and front)
+# - acceptance (back and front)
+# - integration (back and front)
+# - end to end
+#
+# You should at least define a target for every kind of tests we previously list with the following pattern:
```
jjanvier:
```suggestion
# You should at least define a target for every kind of tests we previously listed with the following pattern:
```

make-file/test.mk
```diff
@@ -51,14 +51,33 @@ integration-back: var/tests/phpunit apps-integration-back
...
 
-### Integration tests
+### End to end tests
 .PHONY: end-to-end-back
 end-to-end-back: var/tests/phpunit
 ifeq ($(CI),1)
 	.circleci/run_phpunit.sh . .circleci/find_phpunit.php End_to_End
 else
-	${PHP_RUN} vendor/bin/phpunit -c . --testsuite End_to_End $(O)
+	@echo Run end to end test locally is too long, please use the target define for your bounded context (ex: bounded-context-end-to-end-back)
```
jjanvier:
```suggestion
	@echo Run end to end test locally is too long, please use the target defined for your bounded context (ex: bounded-context-end-to-end-back)
```

.circleci/config.yml
```diff
@@ -87,17 +87,17 @@ jobs:
                 when: always
           - run:
                 name: Analyzes source code to flag programming errors, bugs, stylistic errors, and suspicious constructs
-                command: make lint-back
+                command: make lint-back CI=1
```
vincbe: Strange that you have to add `CI=1` as this is a CircleCI built-in variable 
If it doesn't work without it perhaps it's because you test the value `1` instead of `true` .. so you have to override the builtin (no boolean, env var are always strings)
> CI	Boolean	true (represents whether the current environment is a CI environment)
https://circleci.com/docs/2.0/env-vars/#built-in-environment-variables

.circleci/config.yml
```diff
@@ -87,17 +87,17 @@ jobs:
...
           - run:
                 name: Code Coupling Detection
-                command: make coupling-back
+                command: make coupling-back CI=1
```
vincbe: same

.circleci/config.yml
```diff
@@ -87,17 +87,17 @@ jobs:
...
                 when: always
           - run:
                 name: Unit tests
                 command: make unit-back CI=1
           - run:
                 name: Acceptance tests
-                command: make acceptance-back
+                command: make acceptance-back CI=1
```
vincbe: same... (won't do for each)

make-file/apps.mk
```diff
@@ -1,3 +1,57 @@
...
+#
+# bounded-context-(lint|unit|acceptance|integration|end-to-end)-(back|front)
+#
+# Examples: asset-unit-back, asset-acceptance-front, asset-integration-back.
+#
+# How to define targets with a specfic configuration for the CI?
+# ==============================================================
+# For instance phpspec does not support multiple formats that means you can run the same command on the CI and locally.
+# If you need to do that you can use an environment variable named CI which is a "boolean".
```
vincbe: Environments variables are always string, there is no boolean in fact... it'sprobably why you override the CI built-in with 1 when it is already defined as `true`

PR #11030, opened 2019-11-06
PIM-8941: [experimental] generate typescript interfaces from JSON schema

tseho:

This is experimental.

We want to define JSON schema files for validating API responses.
The validation will be then simplified, when used in the fetchers, with a library like https://github.com/epoberezkin/ajv
We also want to generate non-versioned typescript interfaces from the same JSON schema.

This PR includes 2 webpack changes:
- a **resolver plugin**: when one of the generated typescript interface is imported somewhere in the code, the resolver will ensure this generated file exists instead of throwing a "module not found" error.
- a **loader**: every time a `.schema.json` file is imported somewhere in the code, the loader will generate its typescript interface counterpart. This is really useful for webpack watch.

Since we are hooking into the resolver, I've been doing some performance checks to ensure we are not slowing down webpack too much.

```
existing webpack config + no cache
time for i in {1..10}; do echo ${i} && rm -rf web/cache && run node yarn run webpack-dev > /dev/null; done
7m16,659s

resolver plugin enabled + no cache
time for i in {1..10}; do echo ${i} && rm -rf web/cache && run node yarn run webpack-dev > /dev/null; done
8m7,952s

existing webpack config + cache
time for i in {1..10}; do echo ${i} && run node yarn run webpack-dev > /dev/null; done
6m56,124s

resolver plugin enabled + cache
time for i in {1..10}; do echo ${i} && run node yarn run webpack-dev > /dev/null; done
6m56,998s

resolver plugin enabled + cache + force recompile
time for i in {1..10}; do echo ${i} && rm -f web/js/json-schema/model/family.ts && run node yarn run webpack-dev > /dev/null; done
6m53,628s
```

There is an average increase of 11% when removing the cache. Otherwise, the difference is unnoticeable.

### DIFF
--- a/frontend/webpack/json-schema/compile.js
+++ b/frontend/webpack/json-schema/compile.js
@@ -1,0 +1,9 @@
+/* eslint-env es6 */
+const fs = require('fs');
+const { compileFromFile } = require('json-schema-to-typescript');
+const colors = require('colors');
+
+module.exports = async function compile(schemaPath, typescriptPath) {
+  const content = await compileFromFile(schemaPath);
+  fs.writeFileSync(typescriptPath, content);
+};
--- a/frontend/webpack/json-schema/json-schema-loader.js
+++ b/frontend/webpack/json-schema/json-schema-loader.js
@@ -1,0 +1,22 @@
+/* eslint-env es6 */
+const path = require('path');
+const fs = require('fs');
+const utils = require('loader-utils');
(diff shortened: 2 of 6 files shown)


ahocquard:
SOAP is back :trollface:

tseho:
This has been replaced by an simplier version in https://github.com/akeneo/pim-enterprise-dev/pull/7301

PR #11033, opened 2019-11-07
PIM-8943: Display validation messages on family translations

pierallard:
Before, the back was sending validation errors in this format:

    [
    'translations[0].label': { message: 'Too long' },
    ]

And the front was expecting 

    [
      'translations': { en_US: { message; 'Too long' } }
    ]

After investigation, the [0] comes from the translation relation. Once validation are created, they does not include at all any information about the failing locale. So we choose to index the translations with the current locale (with indexBy).
From the front point of view, we transform "translations[en_US].label: { ... }" into "translations: { en_US : ... } }"

### DIFF
--- a/src/Akeneo/Pim/Structure/Bundle/Resources/config/model/doctrine/Family.orm.yml
+++ b/src/Akeneo/Pim/Structure/Bundle/Resources/config/model/doctrine/Family.orm.yml
@@ -49,6 +49,7 @@ Akeneo\Pim\Structure\Component\Model\Family:
                 - detach
                 - remove
             orphanRemoval: true
+            indexBy: locale
         requirements:
             targetEntity: Akeneo\Pim\Structure\Component\Model\AttributeRequirementInterface
             mappedBy: family
--- a/src/Akeneo/Pim/Structure/Component/Model/Family.php
+++ b/src/Akeneo/Pim/Structure/Component/Model/Family.php
@@ -305,10 +305,8 @@ class Family implements FamilyInterface
         if (null === $locale) {
             return null;
         }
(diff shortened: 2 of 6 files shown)


src/Akeneo/Pim/Structure/Bundle/Resources/config/model/doctrine/Family.orm.yml
```diff
@@ -49,6 +49,7 @@ Akeneo\Pim\Structure\Component\Model\Family:
                 - detach
                 - remove
             orphanRemoval: true
+            indexBy: locale
```
mmetayer: You'll also have to update the `\Akeneo\Pim\Structure\Component\Model\Family::addTranslation()` method to set the correct key in the translations collection

src/Akeneo/Pim/Structure/Component/Model/Family.php
```diff
@@ -305,10 +305,8 @@ public function getTranslation(?string $locale = null)
         if (null === $locale) {
             return null;
         }
-        foreach ($this->getTranslations() as $translation) {
-            if ($translation->getLocale() == $locale) {
-                return $translation;
-            }
+        if (isset($this->getTranslations()[$locale])) {
+            return $this->getTranslations()[$locale];
         }
```
mmetayer:
```php
if ($this->translations->containsKey($locale)) {
    return $this->translations->get($locale);
}
```

pierallard: :woman_shrugging: if you want

pierallard:
@mmetayer Seen with support. Only family is required for this card.

PR #11034, opened 2019-11-07
PIM-6902: Add requirement for ES aliases

pierallard:
https://akeneo.atlassian.net/browse/PIM-6902

### DIFF
--- a/CHANGELOG-3.2.md
+++ b/CHANGELOG-3.2.md
@@ -4,6 +4,7 @@
 
 - PIM-8943: Display validation messages on family translations
 - PIM-8953: User without "list users" permission can not access to other user pages
+- PIM-6902: Forbid usage of uppercase in Elasticsearch indexes (see https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-create-index.html)
 
 # 3.2.17 (2019-10-30)
 
--- a/CHANGELOG-3.2.md
+++ b/CHANGELOG-3.2.md
@@ -4,7 +4,7 @@
 
 - PIM-8943: Display validation messages on family translations
 - PIM-8953: User without "list users" permission can not access to other user pages
-- PIM-6902: Forbid usage of uppercase in Elasticsearch indexes (see https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-create-index.html)
+- PIM-6902: Forbid usage of uppercase in Elasticsearch aliases (see https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-create-index.html)
 
 # 3.2.17 (2019-10-30)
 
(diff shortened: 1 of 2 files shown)


CHANGELOG-3.2.md
```diff
@@ -1,5 +1,9 @@
 # 3.2.x
 
+## Bug fixes:
+
+- PIM-6902: Forbid usage of uppercase in Elasticsearch indexes (see https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-create-index.html)
```
ahocquard: it's alias now, not index

src/Akeneo/Platform/CommunityRequirements.php
```diff
@@ -254,6 +239,34 @@ protected function getConnection() : PDO
         );
     }
 
+    protected function getEsIndexNamesWithUppercase(): array {
+        // We included the EE index names to avoid too much complexity for this use case.
+        // TODO Merge master; the index name list was updated.
+        $esNames = [
```
ahocquard: we will for sure forget it when adding new functionnality, we can get index name from the registry of the client I think

pierallard: It's impossible, Kernel and autload are not loaded with this class call.

src/Akeneo/Platform/CommunityRequirements.php
```diff
@@ -254,6 +239,35 @@ protected function getConnection() : PDO
...
+            'record_index_name',
+            'asset_index_name',
+        ];
+
+        $parameters = $this->getParameters();
+        $esNamesWithUppercase = [];
+        foreach ($esNames as $esName) {
+            if (isset($parameters['parameters'][$esName])) {
+                $value = $parameters['parameters'][$esName];
+                if (strtolower($value) !== $value) {
```
mmetayer: Shouldn't we also validate the other naming rules ? As per https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-create-index.html#indices-create-api-path-params, index (and alias) names should:
- Lowercase only
- Cannot include `\`, `/`, `*`, `?`, `"`, `<`, `>`, `|`, ` ` (space character), `,`, `#`
- Indices prior to 7.0 could contain a colon (:), but that’s been deprecated and won’t be supported in 7.0+
- Cannot start with `-`, `_`, `+`
- Cannot be `.` or `..`
- Cannot be longer than 255 bytes (note it is bytes, so multi-byte characters will count towards the 255 limit faster)

pierallard: Yeah I know, but it's overkill for a little use case. In every cases, it will fail at the installation. (And the card only talk about uppercase)

PR #11047, opened 2019-11-08
Remove values filtering before indexing

nmarniesse:

Values come from source of truth (DB) and we don't need to filter the values before send data to Elasticsearch. For performance reason we now drop the filtering.

There is several ways to do it, none of them seem perfect to me.

Variant solution:
We can consider it's not the responsibility of the `ReadValueCollectionFactory` to filter, so maybe we can directly remove the filtering from `createMultipleFromStorageFormat()`.  
The services that use this method must decide if they have to filter the values or not before calling it. This seem to be cleaner in term of responsibility but not in term of security: we can easily forget the step to filter the values.
Others solution: we can create a InternalReadValueCollectionFactory. 

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Akeneo/Pim/Enrichment/Component/Product/Factory/ReadValueCollectionFactory.php
+++ b/src/Akeneo/Pim/Enrichment/Component/Product/Factory/ReadValueCollectionFactory.php
@@ -39,30 +39,22 @@ class ReadValueCollectionFactory
         return $this->createMultipleFromStorageFormat([$notUsedIdentifier => $rawValues])[$notUsedIdentifier];
     }
 
+    /**
+     * @param array $rawValueCollections
+     * @return ReadValueCollection[]
+     */
     public function createMultipleFromStorageFormat(array $rawValueCollections): array
     {
         $filteredRawValuesCollection = $this->chainedNonExistentValuesFilter->filterAll($rawValueCollections);
-        $valueCollections = $this->createValues($filteredRawValuesCollection);
 
-        return $valueCollections;
+        return $this->createMultipleFromStorageFormatWithoutFilter($filteredRawValuesCollection);
     }
 
-    private function getAttributesUsedByProducts(array $rawValueCollections): array
-    {
-        $attributeCodes = [];
-
-        foreach ($rawValueCollections as $productIdentifier => $rawValues) {
-            foreach (array_keys($rawValues) as $attributeCode) {
-                $attributeCodes[] = (string) $attributeCode;
-            }
-        }
-
-        $attributes = $this->getAttributeByCodes->forCodes(array_unique($attributeCodes));
-
-        return $attributes;
-    }
-
-    private function createValues(array $rawValueCollections): array
+    /**
+     * @param array $rawValueCollections
+     * @return ReadValueCollection[]
+     */
+    public function createMultipleFromStorageFormatWithoutFilter(array $rawValueCollections): array
     {
         $entities = [];
         $attributes = $this->getAttributesUsedByProducts($rawValueCollections);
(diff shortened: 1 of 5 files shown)


src/Akeneo/Pim/Enrichment/Component/Product/Factory/ReadValueCollectionFactory.php
```diff
@@ -39,30 +39,25 @@ public function createFromStorageFormat(array $rawValues): ReadValueCollection
...
-        }
-
-        $attributes = $this->getAttributeByCodes->forCodes(array_unique($attributeCodes));
-
-        return $attributes;
-    }
-
-    private function createValues(array $rawValueCollections): array
+    /**
+     * Warning: use this method only for internal operations. Data can be inconsistent
```
ahocquard: Big warning: this method does not filter inconsistent data. For example, if a locale is removed, the values about this locale will be present in the value collection, despite this one does not exist anymore.

You should carefully use this method if you know that the inconsistency does not matter, as it's the case for the Elasticsearch indexation for example. You should not use it if you expose it publicly, such as in the API and very probably in the UI.
It is mainly skipped for performance purpose: you can get until a performance gain of 30% to hydrate the values, which is not negligible.

src/Akeneo/Pim/Enrichment/Component/Product/Factory/ReadValueCollectionFactory.php
```diff
@@ -39,30 +39,25 @@ public function createFromStorageFormat(array $rawValues): ReadValueCollection
...
+     * (cf. ChainedNonExistentValuesFilterInterface)
+     *
+     * @param array $rawValueCollections
+     * @return ReadValueCollection[]
+     */
+    public function createMultipleFromStorageFormatWithoutFilter(array $rawValueCollections): array
```
ahocquard:
```suggestion
    public function createMultipleFromStorageFormatWithoutFilteringInconsistentData(array $rawValueCollections): array
```

tests/back/Pim/Enrichment/Specification/Component/Product/Factory/ReadValueCollectionFactorySpec.php
```diff
@@ -88,4 +88,45 @@ function it_creates_a_values_collection_from_the_storage_format_for_one_entity(
             ]
         ));
     }
+
+    function it_creates_multiple_value_collections_without_filter(GetAttributes $getAttributeByCodes)
```
ahocquard:
```suggestion
    function it_creates_multiple_value_collections_without_filtering_inconsistent_data(GetAttributes $getAttributeByCodes)
```

tests/back/Pim/Enrichment/Specification/Component/Product/Factory/ReadValueCollectionFactorySpec.php
```diff
@@ -88,4 +88,45 @@ function it_creates_a_values_collection_from_the_storage_format_for_one_entity(
...
+                    'tablet' => [
+                        'en_US' => 'a text area for tablets in English',
+                        'fr_FR' => 'une zone de texte pour les tablettes en français',
+
+                    ],
+                ],
+            ],
+        ];
+
+        $getAttributeByCodes->forCodes(['sku', 'description'])->willReturn(['sku' => $sku, 'description' => $description]);
```
ahocquard: could you add a non existing attribute also, to test that it's correctly skipped

ahocquard:
Seen together: we keep in mind this improvement. But it complicates the things. We will see how it's used in production. The real problem is how we currently handle the deletion, and it's probably here where should invest time, instead of workaround to filter non existent data when reading in database.

We have a gain from 68 to 56 products/sec on the reference catalog in local (1m12 for 4100 products to 1min). It's not negligible, but it's not a x2. Also, this improvement is only about the indexation, which represents 20% max of the import time. So it's 20% improvement on 20% = 4% max.

PR #11055, opened 2019-11-13
API: Cleaning react Reducer / Actions / Types

willy-ahva:

`OK`: Done / Validated
`-`: Not needed


src/Akeneo/Apps/front/src/application/apps/components/CreateAppForm.tsx
```diff
@@ -0,0 +1,176 @@
...
+    const translate = useContext(TranslateContext);
+    const history = useHistory();
+    const isFirstRender = useRef(true);
+
+    useEffect(() => {
+        if (isFirstRender.current) {
+            isFirstRender.current = false;
+
+            return;
+        }
```
pchasle: What is this for?

src/Akeneo/Apps/front/src/application/apps/components/CreateAppForm.tsx
```diff
@@ -0,0 +1,176 @@
...
+            if (false === input.checkValidity()) {
+                if (input.validity.valueMissing) {
+                    dispatch(setError(name, `akeneo_apps.constraint.${name}.required`));
+                }
+                if (input.validity.patternMismatch) {
+                    dispatch(setError(name, `akeneo_apps.constraint.${name}.invalid`));
+                }
+            }
+        });
+    }, [state.controls.code.value, state.controls.label.value, dispatch]);
```
pchasle: Eslint might warn on this, you are missing `codeInputRef` and `labelInputRef` dependencies

src/Akeneo/Apps/front/src/application/common/components/button/Button.tsx
```diff
@@ -21,7 +21,7 @@ export const Button = ({
     }
 
     return (
-        <button onClick={onClick} className={classNames.join(' ')} disabled={disabled}>
+        <button type='button' onClick={onClick} className={classNames.join(' ')} disabled={disabled}>
             {children}
             {undefined !== count && <span className='AknButton--withSuffix'>{count}</span>}
         </button>
```
pchasle: You will need to regenerate the snapshot tests

pchasle: `make apps-front-tests-watch` and `u` to update when prompted

src/Akeneo/Apps/front/src/application/apps/components/CreateAppForm.tsx
```diff
@@ -0,0 +1,176 @@
...
+            return;
+        }
+        [codeInputRef, labelInputRef].forEach(inputRef => {
+            const input = inputRef.current;
+            if (null === input) {
+                return;
+            }
+
+            const name = input.name;
+            if (false === input.checkValidity()) {
```
pchasle: You might be able to do this check to avoid re-validating. If you have errors on an invalid input, then it's already validated.
```suggestion
            if (false === input.checkValidity() && 0 === Object.keys(state.controls[name].errors).length) {
```

src/Akeneo/Apps/front/src/application/apps/components/CreateAppForm.tsx
willy-ahva: The first time we render the useEffect will check validity of inputs. As they are not valid (code is null for example) errors are shown when the form is rendered. To avoid this you implemented a flag in the state with action/reducer (SET_VALIDATED), which is heavier than having a useRef flag. This way to do was advised by François during the training.

pchasle: I didn't know of this usage of useRef :+1: 

But there is still the issue that all errors will be triggered on the second render even if you didn't touched the 2nd input.

e.g. editing **code** => trigger validation => error "**label** is required" is displayed (but user didn't interact with it)

An easy fix will be to only valid if the control is `touched` (or `dirty`) and it's generally the go to solution, you can take a look at [Formik](https://jaredpalmer.com/formik/docs/guides/validation) or [Angular](https://angular.io/guide/form-validation)

PR #11069, opened 2019-11-15
Pull-up: Fix Asset Manager integration tests

SamirBoulil:

In 3.2, we experienced an issue on the asset manager BC regarding the integration tests of controllers. Every 404th+ tests would fail, but they would pass if run one by one.

We figured this issue happened because the SSO Repository had no InMemory implementation for the `test_fake` environment, so it would call the database anyway.
This repo was called for each HTTP request made to the PIM.

To fix this issue, we decided to close openned connection even on the for kernel built with the `test_fake` env.

On master, this service was not part of the `test_fake` env (_as it should_).

Today, we are currenlty blocked by this in the 3.2 -> master pull-up.

**Solution:**
Add the `akeneo_integration_tests.doctrine.connection.connection_closer` service into the `test_fake` environement.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/config/services/test_fake/services.yml
+++ b/config/services/test_fake/services.yml
@@ -11,3 +11,9 @@ imports:
     - { resource: ../../../src/Akeneo/Apps/back/tests/Acceptance/Resources/config/queries.yml }
     - { resource: ../../../src/Akeneo/Apps/back/tests/Acceptance/Resources/config/repositories.yml }
     - { resource: ../../../src/Akeneo/Apps/back/tests/Acceptance/Resources/config/services.yml }
+
+services:
+    akeneo_integration_tests.doctrine.connection.connection_closer:
+        class: Akeneo\Test\IntegrationTestsBundle\Doctrine\Connection\ConnectionCloser
+        arguments:
+            - '@doctrine'


config/services/test_fake/services.yml
```diff
@@ -11,3 +11,12 @@ imports:
     - { resource: ../../../src/Akeneo/Apps/back/tests/Acceptance/Resources/config/queries.yml }
     - { resource: ../../../src/Akeneo/Apps/back/tests/Acceptance/Resources/config/repositories.yml }
     - { resource: ../../../src/Akeneo/Apps/back/tests/Acceptance/Resources/config/services.yml }
+
+parameters:
+    akeneo_integration_tests.doctrine.connection.connection_closer.class: Akeneo\Test\IntegrationTestsBundle\Doctrine\Connection\ConnectionCloser
+
+services:
+    akeneo_integration_tests.doctrine.connection.connection_closer:
+        class: '%akeneo_integration_tests.doctrine.connection.connection_closer.class%'
```
pierallard: Don't hesitate to put directly the class instead of setting a parameter, if you don't want anybody to override this behavior.

SamirBoulil: VENDU!

PR #11080, opened 2019-11-18
 PIM-8954: Forbid access to any user data when the permission 'list users' is not granted

LaurentPetard:
Every user can access to the data of any other user with the end-point "pim_user_user_rest_get", even if the permission "List users" is not granted. 

It's a backport of https://github.com/akeneo/pim-community-dev/pull/11036 for the 2.3 with the difference that in 2.3, the user identifier of the end-point is not the integer id, but the user name or its email.

### DIFF
--- a/CHANGELOG-2.3.md
+++ b/CHANGELOG-2.3.md
@@ -3,7 +3,7 @@
 ## Bug fixes
 
 - PIM-7027: fix completeness visibility on product edit form
-- PIM-8954: Forbid user without "list users" permission to access to other user data
+- PIM-8954: Forbid user without "list users" permission to access other user data
 
 # 2.3.69 (2019-10-24)
 
--- a/CHANGELOG-2.3.md
+++ b/CHANGELOG-2.3.md
@@ -3,6 +3,7 @@
 ## Bug fixes
 
 - PIM-7027: fix completeness visibility on product edit form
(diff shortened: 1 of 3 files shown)


CHANGELOG-2.3.md
```diff
@@ -3,6 +3,7 @@
 ## Bug fixes
 
 - PIM-7027: fix completeness visibility on product edit form
+- PIM-8954: Forbid user without "list users" permission to access to other user data
```
SamirBoulil:
```suggestion
- PIM-8954: Forbid user without "list users" permission to access other user data
```

src/Pim/Bundle/UserBundle/Controller/UserRestController.php
```diff
@@ -65,6 +73,16 @@ public function getAction($identifier)
     {
         $user = $this->userRepository->findOneByIdentifier($identifier);
 
+        $token = $this->tokenStorage->getToken();
+        $currentUserId = null !== $token ? $token->getUser()->getId() : null;
+
+        if (null !== $this->securityFacade) {
```
SamirBoulil: Add a comment to discard this fix as its already fixed in 3.2

LaurentPetard: Fixed

PR #11106, opened 2019-11-21
PIM-8990: Improve the performance of the creation of the attributes requirements

LaurentPetard:

When a channel is created, all the attribute requirements are created and persisted in the same process, e. So one line is inserted per attribute of each family.

For a catalog with a lot of families and a lot of attributes per family, the creation of a channel reaches the maximum time execution.

The proposed solution is simply to create only the attribute requirements for the identifier.

`OK`: Done / Validated
`-`: Not needed


jmleroux:
After some discussion, this PR is closed and will be reworked in a much simpler way.

PR #11115, opened 2019-11-22
PIM-8990: Fix channel creation taking too much time

LaurentPetard:

When a channel is created, all the attribute requirements are created and persisted in the same process. So one line is inserted per attribute of each family.

For a catalog with a lot of families and a lot of attributes per family, the creation of a channel reaches the maximum time execution. (PS: it's not a memory leak. The process could run for hours if we don't limit it)

The proposed solution is simply to create only the attribute requirements for the identifier.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/CHANGELOG-3.0.md
+++ b/CHANGELOG-3.0.md
@@ -1,5 +1,9 @@
 # 3.0.x
 
+## Bug fixes
+
+- PIM-8990: Create only the attributes requirements of the identifier attribute when a channel is created
+
 # 3.0.55 (2019-11-22)
 
 ## Bug fixes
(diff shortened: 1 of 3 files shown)


CHANGELOG-3.0.md
```diff
@@ -1,5 +1,9 @@
 # 3.0.x
 
+## Bug fixes
+
+- PIM-8990: Create only the attributes requirements of the identifier attribute when a channel is created
```
rybus: Fix the memory leak due to the creation of too many attributes requirements when creating a channel.

LaurentPetard: I misphrased the commit message, but it's not a memory leak. It's "just" that it took too much time to execute all the queries.

rybus: Awww okay! "Reduce time execution" then? :stuck_out_tongue:

mmetayer: or neutral "Fix initialization of attribute requirements on channel creation"

PR #11119, opened 2019-11-22
PIM-8997: Replace versioning cursor used for the purge

LaurentPetard:
The existing cursor used for purging the versions has a memory leak.

This PR add a new cursor using generator & doctrine iterate to reduce the memory footprint.
It also add an option `--batch-size` to the command when it's not enough.

This way, when they have a smaller amount of versions but that are really heavy, users can reduce the batch size as a workaround for the `out of memory`.

To do:
- [ ] tests


src/Pim/Bundle/VersioningBundle/Doctrine/ORM/VersionRepository.php
```diff
@@ -145,7 +147,7 @@ public function findPotentiallyPurgeableBy(array $options = [])
             $qb->setParameter('limit_date', $options['limit_date'], Type::DATETIME);
         }
 
-        return $this->cursorFactory->createCursor($qb);
+        return $this->cursorFactory->createCursor($qb, ['page_size' => $options['batch_size'] ?? 100]);
```
jmleroux: I don't think you should pass a default value.

src/Pim/Bundle/VersioningBundle/Purger/VersionPurger.php
```diff
@@ -22,8 +22,6 @@
  */
 class VersionPurger implements VersionPurgerInterface
 {
-    const BULK_THRESHOLD = 1000;
```
pierrejolly: Maybe we should keep this constant to avoid BC break.

jmleroux:
Way cleaner. I like it.

PR #11132, opened 2019-11-26
PIM-8950: PoC on Content Security Policy

jmleroux:
This PR add a Content Security Policy to prevent the execution of javascript `<script></script>` tags in the PIM, mainly to void direct execution in WYSIWYG editors. Sadly, we still have some legacy and legit scripts tags that still must be executed. For these tags, CSP allows to identify them as legit by adding them a nonce string that must correspond to the same nonce in the CSP headers.

To do that, we add a very simple nonce generator using Uuid. This Uuid is the same for all the request. It is used by a Twig extension to add it in templates and by a response subscriber to add it in the headers.

To sum up the request process:
- twig templates are rendered and the twig extension call the nonce generator. First call will generate it, subsequent calls will just use the cached value.
- before sending the response, the subscriber inject the CSP header. The headers call the nonce generator, again using the cached value
- the response is sent with the same nonce in twig templates and in the headers

### DIFF
--- a/src/Akeneo/Platform/Bundle/UIBundle/EventListener/AddContentSecurityPolicyListener.php
+++ b/src/Akeneo/Platform/Bundle/UIBundle/EventListener/AddContentSecurityPolicyListener.php
@@ -1,0 +1,45 @@
+<?php
+
+namespace Akeneo\Platform\Bundle\UIBundle\EventListener;
+
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\HttpKernel\Event\FilterResponseEvent;
+use Symfony\Component\HttpKernel\KernelEvents;
+
+/**
+ * Inject CSP headers in response object
+ *
+ * @author JM Leroux <jean-marie.leroux@akeneo.com>
+ * @copyright 2019 Akeneo SAS (http://www.akeneo.com)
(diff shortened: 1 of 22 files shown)


src/Akeneo/Platform/config/pim.yml
```diff
@@ -29,6 +29,8 @@ imports:
 parameters:
     pim_job_product_batch_size: 100
     env(AKENEO_PIM_URL): ''
+    env(UNIQUE_ID): '2726c7f26c'
```
jmleroux: Default value for the UNIQUE_ID env var in case it is not set.

src/Akeneo/Platform/Bundle/UIBundle/EventListener/AddContentSecurityPolicyListener.php
```diff
@@ -0,0 +1,42 @@
...
+    public static function getSubscribedEvents(): array
+    {
+        return [
+            KernelEvents::RESPONSE => 'addCspHeaders',
+        ];
+    }
+
+    public function addCspHeaders(FilterResponseEvent $event): void
+    {
+        $policy = "frame-src *.youtube.com";
```
jmleroux: This whitelist should be discussed about what we can accept

Fatma3011:
Hi, i have this problem, i can't solve it, any idea please ?


image

Fatma3011:
i coud'nt handle this error, any solution ?

image

PR #11144, opened 2019-11-29
API-780: Be able to edit an App to add an image.

willy-ahva:

`OK`: Done / Validated
`-`: Not needed


src/Akeneo/Apps/front/src/application/apps/components/ImageUploader.tsx
```diff
@@ -0,0 +1,77 @@
...
+        <>
+            <div className={containerClassName}>
+                <input type='file' accept='image/*' onChange={onInputChange} className='AknImage-updater' />
+
+                {null === previewImage && (
+                    <div className='AknImage-uploader'>
+                        <span className='AknImage-uploaderIllustration' />
+                        <span className='AknImage-uploaderHelper'>
+                            <Translate id={'pim_apps.image_helper.upload'} />{' '}
+                            <HelperLink href='#'>
```
pchasle: You need a translation key for the link too.

src/Akeneo/Apps/back/Application/Command/UpdateAppHandler.php
```diff
@@ -44,6 +46,8 @@ public function handle(UpdateAppCommand $command): void
 
         $app->setLabel(new AppLabel($command->label()));
         $app->setFlowType(new FlowType($command->flowType()));
+        $app->setImage(null !== $command->image() ? new AppImage($command->image()) : null);
```
fitn: As we don't deal with image deletion, it would be better to call setImage only when you have an image. And that's more readable than having a ternary operator in a setter.
```suggestion
        if (null !== $command->image()) {
            $app->setImage(new AppImage($command->image()));
        }
```

src/Akeneo/Apps/back/Domain/Model/Write/App.php
```diff
@@ -80,4 +91,9 @@ public function setFlowType(FlowType $flowType): void
     {
         $this->flowType = $flowType;
     }
+
+    public function setImage(?AppImage $image): void
```
fitn: According to a previous comment. If we want to remove the image, a method removeImage can be added :)
```suggestion
    public function setImage(AppImage $image): void
```

src/Akeneo/Apps/back/Infrastructure/Symfony/Resources/config/validation/UpdateAppCommand.yml
```diff
@@ -6,3 +6,5 @@ Akeneo\Apps\Application\Command\UpdateAppCommand:
             - Callback: [Akeneo\Apps\Domain\Validation\App\AppLabelMustBeValid, validate]
         flowType:
             - Callback: [Akeneo\Apps\Domain\Validation\App\FlowTypeMustBeValid, validate]
+        image:
+            - Callback: [Akeneo\Apps\Domain\Validation\App\AppImageMustBeValid, validate]
```
fitn: Should not we validate that the image exists?
As we have no link between media file and app tables and we use a simple string to store the media, I feel we should enforce that point. Otherwise you can save your AppImage with "toto" instead of a real path.
file_key column is indexed so you can query on it

src/Akeneo/Apps/back/Application/Command/UpdateAppHandler.php
fitn: According to the discussion we had with Franck, that's ok like that :+1:

src/Akeneo/Apps/back/Domain/Model/Write/App.php
fitn: Idem here :+1:

src/Akeneo/Apps/front/src/application/apps/components/ImageUploader.tsx
```diff
@@ -0,0 +1,115 @@
...
+            return null;
+        }
+
+        if (isOk(result)) {
+            return result.data.filePath;
+        }
+        return null;
+    };
+
+    const onInputChange = async (event: React.ChangeEvent<HTMLInputElement>) => {
```
pchasle: Juste a note, the convention is to use `handle` as prefix, rather than `on` for event handler, aka. `handleInputChange`

src/Akeneo/Apps/front/src/application/apps/use-image-uploader.ts
```diff
@@ -0,0 +1,50 @@
...
+
+export const useImageUploader = () => {
+    const url = useRoute('pim_enrich_media_rest_post');
+
+    return async (file: File) => {
+        if (!allowedExtensions.includes(file.type)) {
+
+            return err({
+                'extension': {
+                    'message': 'Extension not allowed.',
```
pchasle: Missing translation

src/Akeneo/Apps/front/src/application/common/components/Figure.tsx
```diff
@@ -6,9 +6,15 @@ export const Figure = styled.div`
     cursor: pointer;
 `;
 
+export const ImageContainer = styled.div`
```
pchasle:
```suggestion
const ImageContainer = styled.div`
```

src/Akeneo/Apps/front/src/application/shared/router/use-routing.ts
```diff
@@ -0,0 +1,10 @@
+import {useContext} from 'react';
+import {RouterContext} from './router-context';
+
+export const useRouting = () => {
+    const {generate} = useContext(RouterContext);
+
+    return (route: string, parameters?: object) => {
+            return generate(route, parameters);
+        };
+};
```
pchasle: To remove ?

src/Akeneo/Apps/back/Infrastructure/Symfony/Resources/config/validation/UpdateAppCommand.yml
willy-ahva: done

PR #11167, opened 2019-12-03
TIP-1265: deactivate debug mode by default

ahocquard:

When the debug is activated, the CLI is leaking in term of memory because it collects a lot of data for the profiling.
Here is a PR to deactivate it by default.

We removed the option `--no-debug` to reduce the number of way to specify activate the debug mode. Now, it's only odable through `.env` file, or via environment variable `docker-compose run -e APP_DEBUG=1`.

`OK`: Done / Validated
`-`: Not needed


bin/console
```diff
@@ -3,7 +3,8 @@
 use Symfony\Bundle\FrameworkBundle\Console\Application;
 use Symfony\Component\Console\Input\ArgvInput;
 use Symfony\Component\Debug\Debug;
-
+pcntl_async_signals(true);
+pcntl_signal(SIGINT, function() {meminfo_dump(fopen('/srv/pim/var/cache/dump.json', 'w')); });
```
BitOne: Oops ;)

PR #11171, opened 2019-12-03
Fishing hook: print out failed integration tests

SamirBoulil:
Some back integration tests fails because they endup with a segmentation fault (like this one https://circleci.com/gh/akeneo/pim-enterprise-dev/22764#tests/containers/3)

This PR attempts to print out the integration test file that failed directly after it failed.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/.circleci/run_phpunit.sh
+++ b/.circleci/run_phpunit.sh
@@ -18,7 +18,11 @@ for TEST_FILE in $TEST_FILES; do
 
     set +e
     docker-compose exec -u www-data -T fpm ./vendor/bin/phpunit -c $CONFIG_DIRECTORY --log-junit var/tests/phpunit/phpunit_$(uuidgen).xml $TEST_FILE
-    fail=$(($fail + $?))
+    TEST_RESULT=$?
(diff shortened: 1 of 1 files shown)


.circleci/run_phpunit.sh
```diff
@@ -18,7 +18,11 @@ for TEST_FILE in $TEST_FILES; do
 
     set +e
     docker-compose exec -u www-data -T fpm ./vendor/bin/phpunit -c $CONFIG_DIRECTORY --log-junit var/tests/phpunit/phpunit_$(uuidgen).xml $TEST_FILE
-    fail=$(($fail + $?))
+    TEST_RESULT=$?
+    if [ $TEST_RESULT -ne 0 ]; then
```
BitOne: Could be replaced by ` [ $TEST_RESULT ] || echo "Test has failed: $TEST_FILE"`

jjanvier: I prefer the way Samir wrote it
easier to understand for non bash fans

PR #11173, opened 2019-12-03
API-769: Adds type column in user table

fitn:

I moved the migration script before the apps' creation as our migration script calls the Apps' creation command.
As it uses the new user it eases the migration process.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/upgrades/schema/Version_4_0_20191013162543_update_user_table.php
+++ b/upgrades/schema/Version_4_0_20191013162543_update_user_table.php
@@ -1,0 +1,35 @@
+<?php declare(strict_types=1);
+
+namespace Pim\Upgrade\Schema;
+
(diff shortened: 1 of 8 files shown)


upgrades/schema/Version_4_0_20191013162543_update_user_table.php
```diff
@@ -0,0 +1,35 @@
...
+    }
+
+    public function up(Schema $schema) : void
+    {
+        $this->addSql(<<<SQL
+ALTER TABLE oro_user ADD COLUMN type varchar(20) NOT NULL DEFAULT 'user'
+SQL);
+    }
+
+    public function down(Schema $schema) : void
```
willy-ahva: Shouldn't you throw an exception ?

PR #11176, opened 2019-12-03
PIM-9007: [SLA] Cancelling category selection in export profile does not work

rouliane:
This PR fix a bug on product export when cancelling a category selection. The chosen categories were synced by reference to the backbone "state" when the user selected / unselected a category, so the cancel operation didn't erase the user choices.

This fix is pretty simple : I send to the modal only a copy of the categories in the current backbone state, so the user clicks won't be synced. And only the modal validation button will sync the chosen categories.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/CHANGELOG-3.0.md
+++ b/CHANGELOG-3.0.md
@@ -1,5 +1,9 @@
 # 3.0.x
 
+## Bug fixes
+
+- PIM-9007: Fix cancelling category selection in export profile
+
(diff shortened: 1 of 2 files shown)


rouliane:
I had to push force to update then changelog, but the CI was green before.

jmleroux:
Pro tip: no need to push force as we can squash on Github

PR #11191, opened 2019-12-05
API-845: Avoid app code & label that have less than 3 characters

pierrejolly:

`OK`: Done / Validated
`-`: Not needed


src/Akeneo/Apps/back/Domain/Model/ValueObject/AppCode.php
```diff
@@ -17,10 +17,13 @@ class AppCode
...
         $code = trim($code);
-        
+
         if (empty($code)) {
             throw new \InvalidArgumentException(sprintf(self::CONSTRAINT_KEY, 'required'));
         }
+        if (strlen($code) < 3) {
+            throw new \InvalidArgumentException(sprintf(self::CONSTRAINT_KEY, 'too_short'));
+        }
         if (strlen($code) > 100) {
```
pchasle: We should use `mb_strlen` to avoid issue with unicode characters (symfony validator already use this)

PR #11204, opened 2019-12-06
PIM-8950: Store script in session for ajax calls

jmleroux:
The nonce used to authorize our inline scripts must not be regenerated on ajax calls.
We regenerate it on every first request and store it in session to retrieve it for ajax requests.

### DIFF
--- a/src/Akeneo/Platform/Bundle/UIBundle/EventListener/ScriptNonceGenerator.php
+++ b/src/Akeneo/Platform/Bundle/UIBundle/EventListener/ScriptNonceGenerator.php
@@ -3,6 +3,8 @@
 namespace Akeneo\Platform\Bundle\UIBundle\EventListener;
 
 use Ramsey\Uuid\Uuid;
+use Symfony\Component\HttpFoundation\RequestStack;
+use Symfony\Component\HttpFoundation\Session\Session;
 
 /**
  * Generate and return the CSP javascript nonce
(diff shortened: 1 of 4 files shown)


src/Akeneo/Platform/Bundle/UIBundle/EventListener/ScriptNonceGenerator.php
```diff
@@ -15,11 +17,35 @@ class ScriptNonceGenerator
...
     {
+        if (!$this->request->getCurrentRequest()->isXmlHttpRequest()) {
+            $this->session->set('nonce', null);
+        }
+
+        if (null === $this->generatedNonce) {
+            $this->generatedNonce = $this->session->get('nonce', null);
+        }
+
         if (null === $this->generatedNonce) {
```
pyklein: THis if can be removed as it'll be covered by the one before :)

jmleroux: Nope. `$this->generatedNonce` can be set in the first if.

pyklein: It can execute this if before the previous one? 
I knew this could happen in JS not in PHP... Is it due to the time it take to read the session? :thinking:

PR #11211, opened 2019-12-09
API-862: Update audit table + Create Read Model

fitn:

This PR:
- fixes audit table removing keyword used as column name and change table constraint on app table to replace it with username in user table (versioning uses username, not app code)
- creates read models to be used with audit table
- adds a cli command to populate audit table

`OK`: Done / Validated
`-`: Not needed


src/Akeneo/Apps/back/Infrastructure/Cli/PopulateAuditTableCommand.php
```diff
@@ -0,0 +1,80 @@
...
+use Akeneo\Apps\Application\Query\FetchAppsHandler;
+use Akeneo\Apps\Application\Query\FindAnAppHandler;
+use Akeneo\Apps\Application\Query\FindAnAppQuery;
+use Doctrine\DBAL\Connection;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+
+/**
+ * @author Romain Monceau <romain@akeneo.com>
```
pierrejolly: Maybe you can add comment to not forget to delete this command

```
TODO: To remove
@deprecated
```

src/Akeneo/Apps/back/Infrastructure/Install/Query/CreateAppAuditTableQuery.php
```diff
@@ -14,12 +14,12 @@ final class CreateAppAuditTableQuery
     const QUERY = <<<SQL
 CREATE TABLE IF NOT EXISTS akeneo_app_audit(
     id INT NOT NULL AUTO_INCREMENT,
-    app_code VARCHAR(100) NOT NULL,
+    app_username VARCHAR(255) NOT NULL,
```
pierrejolly: IMO we should keep using app_code, indeed we get an username from versioning but it's more related to the way we choose to implement our solution but what we want to store is an app_code

fitn: Ok I move back this part.

PR #11217, opened 2019-12-10
TIP 1346 - Prepare CE standard edition

BitOne:

This PR prepares a script to generate the PIM Community Standard distribution.

The standard distribution must only contain a composer.json that will do everything needed after installation to create the right directories and copies the file.

This will ease maintenance of the standard distribution as well as the migrations.

The script responsible for this is https://github.com/akeneo/pim-community-dev/blob/9f51c96c7422c56d12f5324d70e8f6f1e46a0af9/std-build/copy-required-files.sh

And it takes its data from this directory: https://github.com/akeneo/pim-community-dev/blob/9f51c96c7422c56d12f5324d70e8f6f1e46a0af9/std-build/

`OK`: Done / Validated
`-`: Not needed


std-build/copy-required-files.sh
```diff
@@ -0,0 +1,36 @@
...
+
+mkdir -p $STANDARD_DISTRIB_DIR/src \
+         $STANDARD_DISTRIB_DIR/bin \
+         $STANDARD_DISTRIB_DIR/public \
+         $STANDARD_DISTRIB_DIR/config/packages \
+         $STANDARD_DISTRIB_DIR/config/services \
+         $STANDARD_DISTRIB_DIR/docker
+
+cp $DEV_DISTRIB_DIR/CHANGELOG*.md $STANDARD_DISTRIB_DIR
+cp $DEV_DISTRIB_DIR/UPGRADE*.md $STANDARD_DISTRIB_DIR
```
jmleroux: I don't think we need to copy CHANGELOG and UPGRADE

std-build/Kernel.php
```diff
@@ -0,0 +1,124 @@
...
+ *
+ * @author Nicolas Dupont <nicolas@akeneo.com>
+ */
+class Kernel extends BaseKernel
+{
+    use MicroKernelTrait;
+
+    public function registerBundles(): iterable
+    {
+        $contents = require $this->getProjectDir() . '/config/bundles.php';
```
BitOne: Load bundles from the CE then the one from the standard.

std-build/bundles.php
```diff
@@ -0,0 +1,63 @@
+<?php
```
BitOne: Provide an empty file (bundles should be loaded from CE then standard)

.circleci/config.yml
```diff
@@ -45,9 +45,6 @@ jobs:
       - run:
             name: Install asset
             command: make assets
-      - run:
-            name: Build CSS
-            command: make css
```
jjanvier: is it for an optimization? or because it's done by something else?

Dockerfile
```diff
@@ -1,7 +1,3 @@
-#
-# This first image will be use as a base
-# for production and development images
-#
 FROM debian:buster-slim AS base
```
jjanvier: the first 40 lines of this Dockerfile are the same than the Dockerfile.base
I don't understand oO

Makefile
```diff
@@ -15,153 +10,8 @@ help:
...
 	@echo ""
 
-## Include all *.mk files
 include make-file/*.mk
 
-##
-## Front
-##
-
-yarn.lock: package.json
```
jjanvier: the cool thing in being here is that you get the autocompletion directy

jjanvier: almost all commands here are useful from a development point of view
so why not giving the whole file to the integrators? (and maybe we could move some of the less useful commands)

config/packages/prod/oneup_flysystem.yml
```diff
@@ -0,0 +1,11 @@
+oneup_flysystem:
```
jjanvier: ah yes, we forgot that

docker/akeneo.conf
```diff
@@ -10,7 +10,7 @@
         RewriteEngine On
         RewriteCond %{REQUEST_FILENAME} !-f
         SetEnv APP_ENV ${APP_ENV}
-        RewriteRule ^(.*)$ app.php [QSA,L]
+        RewriteRule ^(.*)$ index.php [QSA,L]
```
jjanvier: maybe we have some pim-docs to change

std-build/composer.json
```diff
@@ -0,0 +1,39 @@
...
+        "psr-4": {
+          "Pim\\Upgrade\\": "upgrades/"
+        },
+        "exclude-from-classmap": [
+            "vendor/akeneo/pim-community-dev/src/Kernel.php"
+        ]
+
+    },
+    "require": {
+        "akeneo/pim-community-dev": "dev-TIP-1346@dev"
```
jjanvier: to change afterwards

std-build/copy-required-files.sh
```diff
@@ -0,0 +1,36 @@
...
+cp $DEV_DISTRIB_DIR/docker/akeneo.conf $STANDARD_DISTRIB_DIR/docker/
+cp $DEV_DISTRIB_DIR/docker/httpd.conf $STANDARD_DISTRIB_DIR/docker/
+cp $DEV_DISTRIB_DIR/docker/wait_docker_up.sh $STANDARD_DISTRIB_DIR/docker/
+cp $DEV_DISTRIB_DIR/config/bootstrap.php $STANDARD_DISTRIB_DIR/config/
+cp $DEV_DISTRIB_DIR/bin/console $STANDARD_DISTRIB_DIR/bin/
+cp $DEV_DISTRIB_DIR/public/index.php $STANDARD_DISTRIB_DIR/public/
+cp $DEV_DISTRIB_DIR/std-build/Kernel.php $STANDARD_DISTRIB_DIR/src
+cp $DEV_DISTRIB_DIR/std-build/docker-compose.yml $STANDARD_DISTRIB_DIR/
+
+# The following files are only copied if the destination does not exist
```
jjanvier: is it a good idea? can't it lead to problems?
maybe we should explain to the integrators that they should use a VCS and analyze the differences (if any) resulting of the copy

std-build/docker-compose.yml
```diff
@@ -0,0 +1,98 @@
+version: '3.4'
+
+services:
+  php:
+    image: 'akeneo/pim-php-dev:4.0'
```
jjanvier: is this the only difference?
couldn't we do something else?

.circleci/config.yml
BitOne: Done by webpack already.

config/packages/dev/oneup_flysystem.yml
```diff
@@ -1,14 +1,11 @@
 oneup_flysystem:
     adapters:
         catalog_storage_adapter:
-            awss3v3:
-                client: 'Aws\S3\S3Client'
-                bucket: 'catalog'
+            local:
```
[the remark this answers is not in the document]
BitOne: Went to far, rollbacked.

.circleci/detect_structure_changes.sh
```diff
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 
+exit 0
```
BitOne: This is needed to go through the CI as the changes in Docker configuration (public image vs locally built one) makes installing master fails.
This will be reverted before merging onto master.

ahocquard:
We could provide a README also to explain how to initialize the standard edition.

PR #11224, opened 2019-12-10
PIM-8950: Use BAPID cookie to generate JS nonce

jmleroux:
The PR to re-re-re-re-implement Content Security Policy.

**Previously on PIM Content Security Policy, season 4**

*To put a strong defense against the evil forces of XSS, the PIM team implemented the ultimate CSP weapon (similar to the Doom's BFG) in order to eradicate these little gremlins in one shot.*

But our heroes faced multiple obstacles in this fierce battle: they had to generate a incredibly complicated nonce (well... a UUID) to protect all our precious (but dirty) inline JS from the evil XSS forces and inject this antidote in every response before releasing it in the wild world

This was without counting on the forces of AJAX horde, firing request in every direction at every moment. The nonce was re-generated too many times and the great CSP weapon was completely screwed.

To regain control on the XHTML galaxy, PIM batalions enforced a session implementation to check that this nonce was calculated only for original requests, leaving the AJAX rebels use the same nonce stored in session. And it was good

But then came the thumbnail problem, media sources requesting for thumbnails in pure non ajax requests, thus resetting this so precious unique nonce each time they were called in the field.

To alleviate this issue, the PIM alliance decided to enforce a simplest yet strong version of the CSP solution: the nonce will be generated (again with quantic computers generating an incredibly complicated... uuid) only **once per user sesssion**. This way all requests of the universe will use the same golden number, no matter their origin or acceptable content. There is a (little) drawback to use always the same number for all the session, but it's also way simpler and less prone to errors.

All was well in the best of all worlds.

Almost...

Blackhawk inspection squadron captain Alex Popeye saw this artillery and invocated the "I don't like statefull services" galactic clause.

CSP must be reworked.

**The final battle**

In order to put an end to this endless war, PIM governance, speaking with the voice of Benoit Palpatine supported by captain Popeye, decided to enforce an even simplest yet stil strong version of the CSP solution: the nonce will be not be generated anymore, saving billions of quantic blips. The golden number will be computed from the BAPID cookie, which is common to every user request too. But this way we will not use a statefull service, nor will we risk to lock the database because of session access.

**What's next: kill'em all!**

All these efforts are made to protect the inline JS gobelins, small creatures very helpful but tricky. We should get rid of them as soon as possible to have a more secure universe.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Akeneo/Platform/Bundle/UIBundle/EventListener/ScriptNonceGenerator.php
+++ b/src/Akeneo/Platform/Bundle/UIBundle/EventListener/ScriptNonceGenerator.php
@@ -42,16 +42,8 @@ class ScriptNonceGenerator
     public function getGeneratedNonce(): string
     {
         $request = $this->request->getCurrentRequest();
-        $bapid = $request->cookies->get('BAPID');
+        $bapId = $request->cookies->get('BAPID');
 
-        // If the BAPID id null because of a tricked request, we could have the same nonce opening a security hole
-        // because the generated nonce would always be the same.
-        // We can prevent it by generationg a random UUID string. It will probably break the PIM inline scripts,
-        // but also protect us against other inline scripts (like in wysiwig)
-        if (null === $bapid || '' === trim($bapid)) {
-            return Uuid::uuid4()->toString();
-        }
-
-        return hash_hmac('ripemd160', $bapid, $this->kernelSecret);
+        return hash_hmac('ripemd160', $bapId, $this->kernelSecret);
     }
 }
--- a/src/Akeneo/Platform/Bundle/UIBundle/EventListener/ScriptNonceGenerator.php
+++ b/src/Akeneo/Platform/Bundle/UIBundle/EventListener/ScriptNonceGenerator.php
@@ -9,20 +9,31 @@ use Symfony\Component\HttpFoundation\Session\Session;
 /**
  * Generate and return the CSP javascript nonce.
  *
+ * The nonce is a generated string used to identify valid inline scripts used in the PIM.
+ * Every inline script that will not match this nonce will be blocked for execution.
+ * This is mainly used to avoid inline scripts in our wysiwyg editor,
+ * but also every text input that could import malicious javascript.
+ *
+ * @see https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/script-src
+ *
+ * The generated nonce is used in the AddContentSecurityPolicyListener to add the CSP to the HTTP response.
+ * It is also used in Twig template to set the nonce in inline script tags.
+ *
  * @author JM Leroux <jean-marie.leroux@akeneo.com>
  * @copyright 2019 Akeneo SAS (http://www.akeneo.com)
  * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
  */
 class ScriptNonceGenerator
 {
-    /** @var string */
-    private $generatedNonce;
     /** @var RequestStack */
     private $request;
+    /** @var string */
+    private $kernelSecret;
 
-    public function __construct(RequestStack $request)
+    public function __construct(RequestStack $request, string $kernelSecret)
     {
         $this->request = $request;
+        $this->kernelSecret = $kernelSecret;
     }
 
     /**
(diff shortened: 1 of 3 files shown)


tests/back/Platform/Specification/Bundle/UIBundle/EventListener/ScriptNonceGeneratorSpec.php
```diff
@@ -18,18 +18,15 @@ function let(RequestStack $requestStack, Request $request)
...
         $request->getSession()->willReturn($session);
-        $session->set('nonce', null)->shouldBeCalledTimes(1);
         $session->get('nonce', null)->willReturn(null);
         $session->set('nonce', Argument::type('string'))->shouldBeCalledTimes(1);
 
         $this->getGeneratedNonce()->shouldMatch('/\w{8}-\w{4}-\w{4}-\w{4}-\w{8}/');
     }
 
-    function it_get_the_nonce_from_session_for_xmlhttprequest(Request $request, SessionInterface $session)
+    function it_get_the_nonce_from_session(Request $request, SessionInterface $session)
```
ahocquard:
```suggestion
    function it_gets_the_nonce_from_session(Request $request, SessionInterface $session)
```

ahocquard:
```suggestion
    function it_gets_the_nonce_from_session(Request $request, SessionInterface $session)
```

src/Akeneo/Platform/Bundle/UIBundle/EventListener/ScriptNonceGenerator.php
```diff
@@ -30,19 +30,13 @@ public function __construct(RequestStack $request)
     */
     public function getGeneratedNonce(): string
     {
-        if (!$this->request->getCurrentRequest()->isXmlHttpRequest()) {
-            $this->request->getCurrentRequest()->getSession()->set('nonce', null);
-        }
-
         if (null === $this->generatedNonce) {
```
ahocquard: why is it stateful? Why do we put it in a variable? It's onlu used once per request in theory.

src/Akeneo/Platform/Bundle/UIBundle/EventListener/ScriptNonceGenerator.php
```diff
@@ -30,19 +30,13 @@ public function __construct(RequestStack $request)
...
             $this->generatedNonce = $this->request->getCurrentRequest()->getSession()->get('nonce', null);
         }
 
         if (null === $this->generatedNonce) {
             $this->generatedNonce = Uuid::uuid4()->toString();
-            if (!$this->request->getCurrentRequest()->isXmlHttpRequest()) {
-                $this->request->getCurrentRequest()->getSession()->set('nonce', $this->generatedNonce);
-            }
+            $this->request->getCurrentRequest()->getSession()->set('nonce', $this->generatedNonce);
```
tseho: What happens in you use getMasterRequest instead ?

src/Akeneo/Platform/Bundle/UIBundle/EventListener/ScriptNonceGenerator.php
jmleroux: It's generated only once, but used many times by the Twig extension that add the nonce in every inline JS in our templates

tseho: Could you just store it in the master request the first time and retrieve it from there after ?

ahocquard: ok :+1: I would not use it from master request, as otherwise it will open the session each time you try to get it. And it adds a lock in database, preventing AJAX concurrency.

jmleroux:
> I agree that storing that in master request (paramater bag) is more elegant as it avoids a statefull service.

I will have a look at it

BitOne:
Not really comfortable with the master request parameter bag. It's basically a $GLOBALS in disguise. It's too generic. I don't like stateful services as well, but at least we know the data is there and it has a proper location.

ahocquard:
@BitOne I dislike statefull services, and it's particularly dangerous inside the listeners. The "statefulness" of request listeners is error-prone because they are executed very often. I already fought against it several times actually (locale listener, etc). Here, the session is bind to the request, therefore it makes sense to put information inside the instance of the object. It's different from $GLOBAL, because it's bind an object instance.

BitOne:
@ahocquard let's with the associated PR. a parameter bag is still a throw everything in it bag without structure

src/Akeneo/Platform/Bundle/UIBundle/EventListener/ScriptNonceGenerator.php
```diff
@@ -9,42 +9,49 @@
...
-            $this->generatedNonce = $this->request->getCurrentRequest()->getSession()->get('nonce', null);
-        }
-
-        if (null === $this->generatedNonce) {
-            $this->generatedNonce = Uuid::uuid4()->toString();
-            if (!$this->request->getCurrentRequest()->isXmlHttpRequest()) {
-                $this->request->getCurrentRequest()->getSession()->set('nonce', $this->generatedNonce);
-            }
+        $request = $this->request->getCurrentRequest();
+        $bapid = $request->cookies->get('BAPID');
```
BitOne: bapId

jmleroux: I have BAPID in all of my requests, even on demo.akeneo.com

image

PR #11256, opened 2019-12-17
PIM-9030: Fix date formatting for non UTC values

jmleroux:
We can have values for date that are not formatted for UTC. When using API for example.
We need to convert them to UTC before making any operations on them.
We didn't have the issue with values for product because Doctrine knows how to do the conversion, but we have other side effects, like in versioning where the value in history is not converted to UTC.

### DIFF
--- a/src/Akeneo/Pim/Enrichment/Component/Product/Factory/Value/DateValueFactory.php
+++ b/src/Akeneo/Pim/Enrichment/Component/Product/Factory/Value/DateValueFactory.php
@@ -34,18 +34,17 @@ class DateValueFactory extends AbstractValueFactory
             );
         }
 
-
         try {
-            $date = new \DateTime($data);
-
             if (!preg_match('/^\d{4}-\d{2}-\d{2}/', $data)) {
                 $this->throwsInvalidDateException($attribute, $data);
             }
+            $date = new \DateTime($data);
+            $utcDate = \DateTime::createFromFormat('U', $date->format('U'));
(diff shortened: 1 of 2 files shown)


src/Akeneo/Pim/Enrichment/Component/Product/Factory/Value/DateValueFactory.php
```diff
@@ -34,18 +34,17 @@ protected function prepareData(AttributeInterface $attribute, $data, bool $ignor
...
         }
 
-
         try {
-            $date = new \DateTime($data);
-
             if (!preg_match('/^\d{4}-\d{2}-\d{2}/', $data)) {
                 $this->throwsInvalidDateException($attribute, $data);
             }
+            $date = new \DateTime($data);
```
ahocquard: Better to do it this way I think: `new DateTime('2013-03-29 04:13:35', new DateTimeZone('America/New_York')`

jmleroux: `$data` is a string, so i don't know the timezone at this point.
But i can know it after instantiation.

jmleroux: Plus there is an issue with `\DateTime::format` ignoring the timezone : https://www.php.net/manual/fr/datetime.format.php#114366

That is the main reason I want to convert it to UTC.

PR #11263, opened 2019-12-17
PIM-9030: Fix date formatting for non UTC values

jmleroux:
Date attributes values are stored in UTC format for the product raw_values.
Versioning was only storing `Y-m-d` dates and we could have a difference between history and the actual value of the attribute because of a difference timezone.
By storing dates versions in UTC format, we ensure to have the same value in history and current value.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Akeneo/Tool/Bundle/VersioningBundle/tests/integration/Manager/VersionManagerIntegration.php
+++ b/src/Akeneo/Tool/Bundle/VersioningBundle/tests/integration/Manager/VersionManagerIntegration.php
@@ -119,6 +119,11 @@ class VersionManagerIntegration extends TestCase
         $this->productUpdater->update($product, $updates);
         $this->productSaver->save($product);
 
+        $this->assertEquals(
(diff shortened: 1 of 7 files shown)


BitOne:
Can you provide a test showing it fixes the issue, please?

src/Akeneo/Tool/Bundle/VersioningBundle/tests/integration/Manager/VersionManagerIntegration.php
```diff
@@ -134,7 +139,7 @@ public function testCreateProductVersionOnUpdate()
             'groups'     => 'groupB',
             'categories' => '',
             'parent'     => '',
-            'a_date'     => '2017-02-01',
+            'a_date'     => '2017-02-01T00:00:00+01:00',
```
BitOne: Doesn't seem right for me. The previous value "2017-02-01" is the right value.

PR #11272, opened 2019-12-18
PIM-9028: Fix error on JSON_EXTRACT where a_image.code is numerical

jmleroux:
In the fixed request, if `a_image.code` is numerical (ex: 001), it raises a MySQL error 

`Invalid JSON path expression. The error is around character position 5.`

It is because of the following request

```
SELECT JSON_EXTRACT(
               raw_values,
               '$.001."<all_channels>"."<all_locales>"'
           )
FROM pim_catalog_product_model
where id=2;
```

It failed with the same error
If I replace 001 by "001", it will be executed without problems.

### DIFF
--- a/CHANGELOG-3.2.md
+++ b/CHANGELOG-3.2.md
@@ -3,6 +3,7 @@
 ## Bug fixes:
 
 - PIM-9027: Translations missing on metrics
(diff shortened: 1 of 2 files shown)


ahocquard:
:fouet: @jmleroux

PR #11282, opened 2019-12-19
API-892: Update audit table with fresh data

fitn:

`OK`: Done / Validated
`-`: Not needed


src/Akeneo/Apps/back/Domain/Audit/Persistence/Query/ExtractAppsEventCountQuery.php
```diff
@@ -0,0 +1,17 @@
...
+namespace Akeneo\Apps\Domain\Audit\Persistence\Query;
+
+/**
+ * @author Romain Monceau <romain@akeneo.com>
+ * @copyright 2019 Akeneo SAS (http://www.akeneo.com)
+ * @license http://opensource.org/licenses/osl-3.0.php Open Software License (OSL 3.0)
+ */
+interface ExtractAppsEventCountQuery
+{
+    public function extractCreatedProducts(string $date): array;
```
pchasle: - Maybe we should have a Query for each instead of two methods in the same Query, especially if we plan to add other event type in the future.
- The name of the method is not clear enough for me, in relation to the date/time-span (not clear that it's only for the day).
Maybe we should assume that it's the responsibility of the caller to give an interval (startDate, endDate) **OR** add the notion of day to the method name (extractCreatedProductsByDay)

src/Akeneo/Apps/back/Application/Audit/Service/UpdateProductEventCountService.php
```diff
@@ -0,0 +1,45 @@
...
+
+use Akeneo\Apps\Domain\Audit\Persistence\Query\ExtractAppsEventCountQuery;
+use Akeneo\Apps\Domain\Audit\Persistence\Repository\EventCountRepository;
+
+/**
+ * @author Romain Monceau <romain@akeneo.com>
+ * @copyright 2019 Akeneo SAS (http://www.akeneo.com)
+ * @license http://opensource.org/licenses/osl-3.0.php Open Software License (OSL 3.0)
+ */
+class UpdateProductEventCountService
```
pchasle: This should probably be a Command+CommandHandler gluing the different services calls.

PR #11289, opened 2019-12-20
PIM-9036: Rework version purger to use DBAL instead of ORM.

jmleroux:
The goal is to fix two problems on the versioning purge for huge volumes. 

1. There's still a memory leak 
2. The purge takes too much time.

The solution is to not use Doctrine ORM anymore and delete directly the versions from their id with an SQL query. We also change the "purger" to always purge by type of resource to be able to easily find a way to optimize the purge.

We implemented a `key - set` cursor (kind of "search after" cursor) to iterate over the versioning table. It is optimized to use a new index without having to fetch the whole table data that can be too big and impact the performances.
In order to have correct performances, we had to add an new index in a migration file.

Query example:
```
SELECT id FROM pim_versioning_version WHERE resource_name = 'Akeneo\\Pim\\Enrichment\\Component\\Product\\Model\\Product' AND logged_at < '2019-12-22' AND id > 221134 ORDER BY resource_name,  logged_at, id limit 10;

SELECT id, logged_at FROM pim_versioning_version 
WHERE resource_name = 'Akeneo\\Pim\\Enrichment\\Component\\Product\\Model\\Product'  
  AND (
      (logged_at = '2019-12-22' AND id < 999) 
      OR logged_at < '219-12-22'
  ) 
ORDER BY logged_at DESC, id DESC LIMIT 100
```

The query will use the `resource_name_resource_id_version_idx` index to fetch the purgeable ids by batch of 100 items.

We also reworked the Advisors to make them work with DBAL. An advisor can either `remove` versions that should not be purged, like the version 1, or it can `keep` versions that are validated to purge, like all versions but the last.

As a result, this command does not leak anymore and is quicker. It can still take hours on huge versioning tables, but we tested it and it finished to purge a 23 millions versions table (6 month + old versions) in 4-5 hours.


jjanvier:
I remember my old times with Oracle. One of our recommendation was to never use `OR`. (And to favor `UNION` instead for example). It was not effective at all. I have no idea if it's the same or not for MySQL. But I give you this lead in case you can/want to test it.

ahocquard:
@jjanvier it's not doable as soon to do UNION as soon as you want to do a cursor to fetch X entities.
Actually, here it's not a problem. 'OR' is most of the time a problem as soon as it's on different columns. It prevents to use the index. Here, it's on the same column: "access predicate" is done on "logged_at, id". Then, it filters every row by iterating on the leaf of the index.

src/Akeneo/Tool/Bundle/VersioningBundle/Resources/config/model/doctrine/Version.orm.yml
```diff
@@ -3,22 +3,21 @@ Akeneo\Tool\Component\Versioning\Model\Version:
     table: pim_versioning_version
     repositoryClass: Akeneo\Tool\Bundle\VersioningBundle\Doctrine\ORM\VersionRepository
     indexes:
-        resource_name_resource_id_idx:
+        resource_name_resource_id_version_idx:
```
ahocquard:
```suggestion
       # this index is useful to get the last versions for a given entity, without doing row lookup, during the purge
        resource_name_resource_id_version_idx:
```

src/Akeneo/Tool/Bundle/VersioningBundle/Resources/config/model/doctrine/Version.orm.yml
```diff
@@ -3,22 +3,21 @@ Akeneo\Tool\Component\Versioning\Model\Version:
...
         pending_idx:
             columns:
                 - pending
-        version_idx:
-            columns:
-                - version
         logged_at_idx:
             columns:
                 - logged_at
+        resource_name_logged_at_idx:
```
ahocquard: This index is useless, only `logged_at` is useful if I remember well. Also not included in the migration script if it;s still useful!

src/Akeneo/Tool/Bundle/VersioningBundle/spec/Purger/KeepLastVersionPurgerAdvisorSpec.php
```diff
@@ -0,0 +1,43 @@
...
+    {
+        $this->shouldHaveType(KeepLastVersionPurgerAdvisor::class);
+    }
+
+    function it_implements_purger_interface()
+    {
+        $this->shouldImplement(VersionPurgerAdvisorInterface::class);
+    }
+
+    function it_supports_versions_types_only(PurgeableVersionList $versionList)
```
ahocquard: no mock on objects, use the real behavior. The more you are doing that, the more you mimic a behavior. The day it changes, everything should be mocked differently.

src/Akeneo/Tool/Bundle/VersioningBundle/Purger/PurgeableVersionList.php
```diff
@@ -0,0 +1,57 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Akeneo\Tool\Bundle\VersioningBundle\Purger;
+
+class PurgeableVersionList implements \Countable
```
ahocquard: final, it would avoid to use mock in tests ;)

src/Akeneo/Tool/Bundle/VersioningBundle/spec/Purger/SkipFirstVersionPurgerAdvisorSpec.php
```diff
@@ -19,20 +27,18 @@ function it_is_a_version_purger_advisor()
         $this->shouldImplement(VersionPurgerAdvisorInterface::class);
     }
 
-    function it_supports_versions_types_only(VersionInterface $version, $notAVersionObject)
+    function it_supports_versions_types_only(PurgeableVersionList $versionList)
```
ahocquard: same

src/Akeneo/Tool/Bundle/VersioningBundle/tests/integration/Command/PurgeVersioningCommandIntegration.php
```diff
@@ -0,0 +1,138 @@
...
+        $this->givenFamilyVersionsAtLeastAsYoungAs($limitDate, 2, 35, 9);
+        $this->givenFamilyVersionsAtLeastAsYoungAs($limitDate, 3, 44, 13);
+
+        $versions = $this->getConnection()->executeQuery('SELECT id, resource_id, version FROM pim_versioning_version')->fetchAll();
+        $this->assertCount(25, $versions);
+
+        $output = $this->runPurgeCommand();
+        $result = $output->fetch();
+
+        $this->assertContains('Start purging versions of Akeneo\Pim\Structure\Component\Model\Family (1/1)', $result);
```
ahocquard: I would use a entity string dedicated for the test. This way, the day you refactor the Family class by moving, you don't break this test.

src/Akeneo/Tool/Bundle/VersioningBundle/tests/integration/Command/PurgeVersioningCommandIntegration.php
```diff
@@ -0,0 +1,138 @@
...
+    {
+        $limitDate = new \DateTime('now');
+
+        $this->givenFamilyVersionsOlderThan($limitDate, 8, 35);
+        $this->givenFamilyVersionsOlderThan($limitDate, 12, 44);
+        $this->givenFamilyVersionsAtLeastAsYoungAs($limitDate, 2, 35, 9);
+        $this->givenFamilyVersionsAtLeastAsYoungAs($limitDate, 3, 44, 13);
+
+        $versions = $this->getConnection()->executeQuery('SELECT id, resource_id, version FROM pim_versioning_version')->fetchAll();
+        $this->assertCount(25, $versions);
```
ahocquard: `Assert::`, it's a static method and it is deprecated to use it with `$this` (with a real language, it would not work :trollface: )

src/Akeneo/Tool/Bundle/VersioningBundle/tests/integration/Command/PurgeVersioningCommandIntegration.php
```diff
@@ -0,0 +1,138 @@
...
+
+        $output = $this->runPurgeCommand();
+        $result = $output->fetch();
+
+        $this->assertContains('Start purging versions of Akeneo\Pim\Structure\Component\Model\Family (1/1)', $result);
+        $this->assertContains('Successfully deleted 18 versions', $result);
+
+        $versions = $this->getConnection()->executeQuery('SELECT id, resource_id, version FROM pim_versioning_version')->fetchAll();
```
ahocquard: `count(*)` ?

src/Akeneo/Tool/Bundle/VersioningBundle/Resources/config/model/doctrine/Version.orm.yml
```diff
@@ -3,22 +3,21 @@ Akeneo\Tool\Component\Versioning\Model\Version:
...
                 - resource_name
                 - resource_id
-        resource_name_idx:
-            columns:
-                - resource_name
+                - version
         pending_idx:
             columns:
                 - pending
-        version_idx:
```
ahocquard: it's is not removed in the migration script

ahocquard:
Could you remove the behat that test this command also please?

src/Akeneo/Tool/Bundle/VersioningBundle/Doctrine/ORM/VersionRepository.php
```diff
@@ -130,28 +130,31 @@ public function createDatagridQueryBuilder(array $parameters = [])
      */
     public function findPotentiallyPurgeableBy(array $options = [])
     {
-        $qb = $this->createQueryBuilder('v');
+        $connection = $this->_em->getConnection();
```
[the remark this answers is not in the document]
jmleroux: In this case, i found the query builder useful to add "where" conditions.

src/Akeneo/Tool/Bundle/VersioningBundle/Purger/VersionPurger.php
```diff
@@ -22,125 +19,95 @@
...
     }
 
-    /**
-     * {@inheritdoc}
-     */
-    public function addVersionPurgerAdvisor(VersionPurgerAdvisorInterface $versionPurgerAdvisor)
+    private function purgeVersionsByResourceName($resourceName, array $options, OutputInterface $output): int
     {
-        $this->versionPurgerAdvisors[] = $versionPurgerAdvisor;
+        $versionsToPurgeCount = $this->countVersionsToPurge($resourceName, $options);
```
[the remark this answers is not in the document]
jmleroux: It is an approximation and i will document it, but i need a count for the progress bar

jmleroux:
> Could you remove the behat that test this command also please? Thanks!

I thought i removed it. Will check.

src/Akeneo/Tool/Bundle/VersioningBundle/Resources/config/model/doctrine/Version.orm.yml
jmleroux: I don't want to remove it because other scripts could rely on it.

src/Akeneo/Tool/Bundle/VersioningBundle/Doctrine/ORM/VersionRepository.php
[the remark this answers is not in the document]
ahocquard: Ok, matter of style I suppose. What is the purpose of this function btw?

jmleroux:
> Could you remove the behat that test this command also please? Thanks!

Already done: tests/legacy/features/pim/structure/family/purge_versionning.feature is deleted

src/Akeneo/Tool/Bundle/VersioningBundle/Purger/PurgeableVersionList.php
ahocquard: I insist  on that point :wink:

src/Akeneo/Tool/Bundle/VersioningBundle/Doctrine/Query/SqlDeleteVersionsByIdsQuery.php
```diff
@@ -0,0 +1,38 @@
...
+    {
+        $this->dbConnection = $dbConnection;
+    }
+
+    /**
+     * @param int[] $versionIds
+     */
+    public function execute(array $versionIds): void
+    {
+        $sql = <<<SQL
```
nmarniesse: If `$versionIds` is empty, the query will fail. Could you add an early return if `$versionIds` is empty

src/Akeneo/Tool/Bundle/VersioningBundle/Doctrine/Query/SqlGetAllButLastVersionIdsByIdsQuery.php
```diff
@@ -0,0 +1,49 @@
...
+        $query = <<<SQL
+SELECT current.id 
+FROM pim_versioning_version AS current
+WHERE current.id IN (:version_ids)
+AND EXISTS(
+    SELECT 1 FROM pim_versioning_version AS latest 
+    WHERE latest.resource_name = current.resource_name
+        AND latest.resource_id = current.resource_id
+        AND latest.version > current.version
+)
```
nmarniesse: This query seems complicated to me. I see something like:

```sql
SELECT current.id 
FROM pim_versioning_version
WHERE id IN (:version_ids)
ORDER BY version DESC
OFFSET 1
```
(To be tested)

src/Akeneo/Tool/Bundle/VersioningBundle/Doctrine/Query/SqlGetPurgeableVersionListQuery.php
```diff
@@ -0,0 +1,117 @@
...
+     * As most of the versions are for products, the rows are not filtered by "resource_name" filter most of the time.
+     * It's therefore not a big deal to do row lookup.
+     */
+    public function youngerThan(string $resourceName, \DateTime $date, int $listSize): iterable
+    {
+        $query = <<<SQL
+SELECT id, logged_at FROM pim_versioning_version 
+WHERE resource_name = :resource_name  
+  AND (
+      (logged_at = :logged_at AND id > :last_id)
```
nmarniesse: In Where clause, it's better to put the condition using an index before the other condition.
For example here, if we have 1000 lines, the engine will check all the 1000 lines to check the `logged_at` condition (very slow), then it uses the index on `id` (quick). It's better to set the id condition first, this way instead of parsing 1000 lines to check the `logged_at` condition, it will parse less lines.  

Here I think we will never have 1000 lines so it sounds ok

src/Akeneo/Tool/Bundle/VersioningBundle/Doctrine/Query/SqlGetAllButLastVersionIdsByIdsQuery.php
ahocquard: It will not work, you want to return several rows, not just one. EXIST is encourage to have an early return (but it's not totally what is doing Mysql with the `>` operator, it cannot uses it as an access predicate, limitation of the engine)

src/Akeneo/Tool/Bundle/VersioningBundle/Doctrine/Query/SqlGetPurgeableVersionListQuery.php
ahocquard: The order of the condition is not important. It's a declarative language. The engine will choose the JOIN/condition that is the best at first to eliminate a maximum of rows. Here, it uses the index on "logged_at" (which contain also the primary key `id` as Mysql is a clustered table). So, it's very fast, even on 22 millions of lines. (https://use-the-index-luke.com/sql/partial-results/fetch-next-page)

src/Akeneo/Tool/Bundle/VersioningBundle/Doctrine/Query/SqlGetAllButLastVersionIdsByIdsQuery.php
nmarniesse: OFFSET means we want all lines except the first one, and as we order by version desc, all versions will be returned except the latest one.

PR #11302, opened 2019-12-27
Improve tests for the datagrid

ahocquard:
**Refactoring of the one test**

The tests that are testing the filters of the datagrid take 20% of the time on the CI in CE (10% in EE).
I improved for now only 1 test: in this PR this test takes 47 sec (still very slow, I know) instead of 324 sec (7 times faster). If it's the same for the other tests, it means we can reduce from 20% of the time to 3% (20%/7).

**Before:** 

For each filter:
- remove the previous filter
- display the filter
- set the operator
- check datagrid

**Now**

- display all filters for all operators (1 attribute = 1 operator)
- set the operators
- check datagrid

The products created for the test is a matrix:
- 1 product that is eligible for every filters 
- X products, each one being eligible to be returned in the datagrid except for one filter

This way, we have the same degree of confidence: we can detect which filter is broken easily

I know that this kind of tests is not ideal and still too slow. Ideally, we should do unit tests for the front-end or integration tests as Tam did but:
- I doubt about the trustability of these tests with the current front-end architecture
- I doubt about the maintenance, as almost no people is involved at improving the tests.

And we can't delete these tests as it concerns the datagrid, which is the most important screen.

**Deletion of datagrid tests**

There are a lot of tests that are testing the datagrid with two times the same attribute type. For example, filtering with two attributes of type price, first one with equal operator, and second one with all operators (so, all combination).
This is very probably legacy tests, during the time where the integration tests didn't exist for the PQB.

All the filters (price, textarea, etc...) are already tested in the datagrid, no need to test it with the combination of another attribute.
All the filters are also very well tested in integration tests.

`OK`: Done / Validated
`-`: Not needed


tests/legacy/features/pim/enrichment/product/datagrid/filtering/filter_products_per_number_fields.feature
```diff
@@ -8,35 +8,37 @@ Feature: Filter products by number field
     Given the "default" catalog configuration
     And I am logged in as "Mary"
 
-  Scenario: Successfully filter products by empty value for number attributes
+  Scenario: Successfully filter products by empty value for number attributes 2
```
nmarniesse: 2?

ahocquard:
```suggestion
  Scenario: Successfully filter products by empty value for number attributes
```

PR #11313, opened 2019-12-31
PIM-9035: Fix search in the textarea for the very long texts

pierallard:
# Description

If you search through the UI a product containing a word at the end of a text area, filled with a long text, you cannot get the related product.
This is because the CONTAINS use the textarea tokenizer textarea_preprocessed_analyzer, which truncate the textarea to the 100 first chars.

This truncate was done to be able to sort results (c.f. https://github.com/akeneo/pim-community-dev/pull/8045 ). It indicates that "preprocessed" analyzer should only be used with EQUALS filter, not CONTAINS, but CONTAINS still used it.
The integration test was wrong, because it searched a word included in the 100 first characters.

The original solution was to use the standard analyzer, but it can not be done in a patch because it changes a lot of things:
- if you search "un chien", it will return every document with "un" or "chien"
- if you search "t-shirt", you will not find the text because of the tokenization of "t-shirt" in "t", "shirt"
- if you search "un chien", you will not find the documents with "un <b>chien</b>"
etc...

The decision was taken with @BitOne to increase the preprocessed truncate limit to 10000. Not the idea solution, but will fix this specific use case.


src/Akeneo/Pim/Enrichment/Bundle/Elasticsearch/Filter/Attribute/TextAreaFilter.php
```diff
@@ -71,7 +71,7 @@ public function addAttributeFilter(
                 break;
 
             case Operators::CONTAINS:
-                $attributePath .= '.preprocessed';
+                $attributePath .= '';
```
nmarniesse: Is there a reason to keep this line?

src/Akeneo/Pim/Enrichment/Bundle/Elasticsearch/Filter/Attribute/TextAreaFilter.php
```diff
@@ -82,7 +82,7 @@ public function addAttributeFilter(
                 break;
 
             case Operators::DOES_NOT_CONTAIN:
-                $attributePath .= '.preprocessed';
+                $attributePath .= '';
```
nmarniesse: same

src/Akeneo/Pim/Enrichment/Bundle/Elasticsearch/Filter/Attribute/TextAreaFilter.php
pierallard: yes

src/Akeneo/Pim/Enrichment/Bundle/Elasticsearch/Filter/Attribute/TextAreaFilter.php
pierallard: same

src/Akeneo/Pim/Enrichment/Bundle/Elasticsearch/Filter/Attribute/TextAreaFilter.php
jmleroux: Discussed IRL, it's easier for merging in upper branches

PR #11315, opened 2019-12-31
PIM-9041: Fix performance issue on "Add to existing product model" mass action

pierallard:
# Description

The mass edit action "Add to existing product model" is quite slow with a huge set of families.
The query is OK (very quick and not optimizable), but the normalization is still complex (it normalize all the family, attributes, requirements, etc).
It exists an option in the normalizer to only return code and labels, this PR use it.

With icecat, the previous query was 250ms ~, now it's 60ms ~.

### DIFF
--- a/CHANGELOG-3.0.md
+++ b/CHANGELOG-3.0.md
@@ -5,6 +5,7 @@
 ## Bug fixes
 
 - PIM-9030: Fix date formatting for non UTC values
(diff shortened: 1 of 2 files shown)


ahocquard:
Actually, the query is optimizable, because it's doing 1 +n queries behind the hood, "thanks to" the lazy loading :) (https://github.com/ahocquard/pim-community-dev/pull/4/files#diff-d2623d982002950a4b47f56283ec9ded)

PR #11323, opened 2020-01-02
PIM-6386: Add exception when trying to update the same product value on an attribute twice, with the same scope and locale

pierallard:


CHANGELOG-2.3.md
```diff
@@ -1,5 +1,9 @@
 # 2.3.x
 
+## Bug fixes
+
+PIM-6386: Add exception when trying to update the same product value on the "a_metric" attribute twice, with the same scope and locale
```
nmarniesse: It works only with "a_metric" attribute? :trollface:

pierallard: Haha yeah

PR #11339, opened 2020-01-06
DAPI-710 add jwt library dependency and ext-openssl requirement

AymericPlanche:

`OK`: Done / Validated
`-`: Not needed


composer.json
```diff
@@ -67,6 +68,7 @@
         "friendsofsymfony/oauth-server-bundle": "1.6.2",
         "friendsofsymfony/rest-bundle": "2.6.0",
         "gedmo/doctrine-extensions":"v2.4.37",
+        "lcobucci/jwt": "3.3.1",
```
olivierpontier: Why do we have to define the JWT dependency as a requirement of the CE too? Could we talk about that together, please?

PR #11348, opened 2020-01-07
Set default value to suggested confirm action

FrankStappers:
Matches the default value to the suggested default action


jmleroux:
CLA check

PR #11380, opened 2020-01-10
Fix migration job execution

ahocquard:
The goal of this migration is to delete the compute product model descendant job. Before doing that, we need to get all pending jobs to index product models.

We were using a search after after (key set pagination):
1. It cannot work as it is designed to search after a specific id. It means you have to fetch the whole product model table to apply the search after.
2. It cannot work as it does not filter on jobs that are pending (consumer is NULL in the queue). It filters on all jobs... 

So, it was taking too much time to be executed.

I tested the query on 800 000 jobs and it's pretty fast (4 seconds). It returns 120 000 product model codes (thanks to the DISTINCT), which is ok in term of memory. So, no need to have a search after pagination in my opinion.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/upgrades/schema/Version_4_0_20191004145507_remove_compute_model_descendant_jobs.php
+++ b/upgrades/schema/Version_4_0_20191004145507_remove_compute_model_descendant_jobs.php
@@ -16,8 +16,6 @@ use Symfony\Component\DependencyInjection\ContainerInterface;
 final class Version_4_0_20191004145507_remove_compute_model_descendant_jobs extends AbstractMigration implements
     ContainerAwareInterface
 {
-    private const BATCH_SIZE = 10000;
-
     /** @var ContainerInterface */
     private $container;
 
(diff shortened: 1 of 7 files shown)


upgrades/schema/Version_4_0_20191004145507_remove_compute_model_descendant_jobs.php
```diff
@@ -31,85 +29,70 @@ public function setContainer(ContainerInterface $container = null)
...
-        $this->addSql($sql, ['jobInstanceId' => $jobInstanceId], ['jobInstanceId' => \PDO::PARAM_INT]);
+        $sql = <<<SQL
+            DELETE job_instance
+            FROM 
+                akeneo_batch_job_instance job_instance
+            WHERE job_instance.code = 'compute_product_models_descendants'
+        SQL;
 
-        $this->addSql("DELETE FROM akeneo_batch_job_instance WHERE code = 'compute_product_models_descendants'");
+        $this->addSql($sql);
```
jjanvier: on line [113](https://github.com/akeneo/pim-community-dev/blob/a867d4771fbf5dd309087496a0ec83936731827f/upgrades/schema/Version_4_0_20191004145507_remove_compute_model_descendant_jobs.php#L113-L114), there is a `pim_catalog.elasticsearch.indexer.product_model_descendants_and_ancestors->indexFromProductModelCodes()`

maybe this is just because the name is missleading, but, do we really need to reindex the descendants whereas it has been done [5 lines before](https://github.com/akeneo/pim-community-dev/blob/a867d4771fbf5dd309087496a0ec83936731827f/upgrades/schema/Version_4_0_20191004145507_remove_compute_model_descendant_jobs.php#L107)

ahocquard: compute and persist = completeness computed and persisted in mysql
indexation = index product and product models in ES

You need to recalculate the completeness before doing. Basically, that's the logic of the compute product model descendant job.

PR #11385, opened 2020-01-13
Pim 9032 - Fix specs with prophecy 1.10

anaisbaunelemaire:


tests/back/Pim/Enrichment/Specification/Component/Product/EntityWithFamilyVariant/CheckAttributeEditableSpec.php
```diff
@@ -49,18 +49,6 @@ function it_is_editable_if_the_attribute_is_part_of_the_family_and_the_product_i
         $this->isEditable($product, $attribute)->shouldReturn(true);
     }
 
-    function it_throws_an_exception_if_the_variant_product_has_no_family_variant(
```
jmleroux: We delete it because it's not a real use case.

PR #9624, opened 2019-02-21
Dump SQL for features

pierallard:
Currently, each step `Given a xx catalog`
- loads the fixtures jobs
- loads the fixtures
- removes the fixtures jobs
For example, the bigger catalog (catalog_modeling) takes ~50s.

This PR dumps the SQL after a catalog initialization, in a .sql file. In the next catalog initialization, if this .sql already exists, it only reloads the SQL file.
For example, catalog_modeling is loaded in ~7s.

Previous timings : 34m08, 34m14
https://circleci.com/gh/akeneo/pim-community-dev/4526#build-timing
https://circleci.com/gh/akeneo/pim-community-dev/4991#build-timing

New timing: 28m57
https://circleci.com/gh/akeneo/pim-community-dev/5106#build-timing

- [x] To check before merge: Why we didn't put the .sql in the ~/var/cache folder ? Same question for the Integration tests .sql files.

PR #11395, opened 2020-01-14
Add a dump when loading the catalog in the legacy behats

ahocquard:
When we load the fixtures in the behat, we are using basic import. When we load several times the same catalog, it's pretty useless as we could use a SQL dump instead.

Actually this is what is done for the integration tests. We didn't want to complexify the stack one year ago with a cache layer with the new CI. Now that it's pretty "stable" (the infra, not the tests :trollface: ) and mature, I think it's time to use a cache.

We already did an experiment one year ago with Pierre (https://github.com/akeneo/pim-community-dev/pull/9624), but it was a lot of copy/past. A better solution is to use the same way to load a catalog between behat and integration tests.

I had to adapt tests on reference data, as it's now loaded by default with all the reference data (as in the integration tests). Before, it was not loaded.

Note: I don't think it will lead to more unstability in our tests as our issues occur **after** loading the catalog (race condition on the front 99% of the time).

**CE Results**
```
curl "https://circleci.com/api/v1.1/project/github/akeneo/pim-community-dev/27765/tests?circle-token=token_circle_ci" | jq '.tests |= sort_by(.run_time) | .tests[].run_time' | awk '{sum+=$1} END{printf("%f\n",sum)}'
```

Total: 13460 seconds (3h40)
After: 11375 seconds  (3h09)

Gain -15% time

**EE Results**

Before: 35633 seconds (9h53)
After: 26468 seconds (7h21)
Gain: -25% time

`OK`: Done / Validated
`-`: Not needed


tests/legacy/features/pim/enrichment/reference-data/import/import_products.feature
```diff
@@ -33,22 +30,22 @@ Feature: Execute a job
     And the family of the product "SKU-001" should be "new_heels"
     And product "SKU-002" should be enabled
     And the product "SKU-001" should have the following values:
-      | heel_color               | [Red]          |
-      | sole_fabric              | [Nylon], [PVC] |
-      | lace_fabric-en_US-tablet |                |
-      | lace_fabric-en_US-mobile | [Kevlar]       |
+      | heel_color               | [red]              |
```
jjanvier: are all those kind of things normal?
we should we change the tests if we just change the way to load the fixtures?

jjanvier:
(Nice PRs by the way)

ahocquard:
@jjanvier I relaunched it something like 3 times and in EE same. I think it's stable because our issues are about the race condition in the UI.

tests/legacy/features/pim/enrichment/reference-data/import/import_products.feature
ahocquard: did you read the descrioption :wink: ? Reference data are loaded with the ctaalog by default now, as for the integration tests. It was not loaded before, so I had issues about duplicated keys. It does not change the behavior of the test.

jjanvier: my comment was more about: before we had a label and now we have a code
this kind of refactoring should not change those tests (except the ones that loads data that are now here with the dump of course)

ahocquard: It was a code in both cases, but before it was a code with an uppercase ;)

PR #11399, opened 2020-01-14
PIM-9036: Update versioning indexes

jmleroux:
I think i can copy / paste this migration in previous versions, because it can be executed whatever the database state.

### DIFF
--- a/src/Akeneo/Tool/Bundle/VersioningBundle/Resources/config/model/doctrine/Version.orm.yml
+++ b/src/Akeneo/Tool/Bundle/VersioningBundle/Resources/config/model/doctrine/Version.orm.yml
@@ -3,13 +3,11 @@ Akeneo\Tool\Component\Versioning\Model\Version:
     table: pim_versioning_version
     repositoryClass: Akeneo\Tool\Bundle\VersioningBundle\Doctrine\ORM\VersionRepository
     indexes:
-        resource_name_resource_id_idx:
+        resource_name_resource_id_version_idx:
             columns:
                 - resource_name
                 - resource_id
-        resource_name_idx:
-            columns:
-                - resource_name
+                - version
         pending_idx:
             columns:
                 - pending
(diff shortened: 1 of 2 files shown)


src/Akeneo/Tool/Bundle/VersioningBundle/Resources/config/model/doctrine/Version.orm.yml
```diff
@@ -3,13 +3,11 @@ Akeneo\Tool\Component\Versioning\Model\Version:
     table: pim_versioning_version
     repositoryClass: Akeneo\Tool\Bundle\VersioningBundle\Doctrine\ORM\VersionRepository
     indexes:
-        resource_name_resource_id_idx:
+        resource_name_resource_id_version_idx:
```
ahocquard: please document it as good practice of the why of this index. Otehrwise, it's pretty impossible to know the consequence of changing/removing an index...

```suggestion
        # this index is useful to get the maximum version for a given entity when doing the purge
        # it allows to do an index-only scan, which improved greatly the performance
        # @see https://use-the-index-luke.com/sql/clustering/index-only-scan-covering-index
        resource_name_resource_id_version_idx:
```

upgrades/schema/Version_4_0_202001131458_fix_versioning_indexes.php
```diff
@@ -0,0 +1,63 @@
...
+class Version_4_0_202001131458_fix_versioning_indexes extends AbstractMigration
+{
+    public function up(Schema $schema)
+    {
+        $this->updateVersioningIndexes();
+    }
+
+    /**
+     * Try to drop and create versioning indexes.
+     * Does nothing if they are already dropped/created
```
ahocquard:
```suggestion
     * Does nothing if they are already dropped/created, as the migration could be applied from a patch in 3.2
```

jmleroux:
```suggestion
     * Does nothing if they are already dropped/created, as the migration could be already applied from a patch in 3.x
```

PR #11408, opened 2020-01-15
PIM-6160: Fix update of 'unique' property in attribute updater

mmetayer:
Fix update of 'unique' property in attribute updater when updating both unique and type properties

In AttributeUpdater, when updating first the 'unique' property then the 'type' property (this can happen with imports), the 'unique' value is overridden by the attribute type's default `isUnique()` value.
This PR fixes that:
- if the attribute type must be unique (`AttributeType::isUnique()` is true, e.g identifier attribute type), then the attribute's `unique` property is forced to `true`
- else if the attribute's `unique` property is not set (attribute creation), the `unique` property is initialized with the attribute type's default value
- else we do not update the `unique` property

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Akeneo/Pim/Structure/Component/Updater/AttributeUpdater.php
+++ b/src/Akeneo/Pim/Structure/Component/Updater/AttributeUpdater.php
@@ -299,7 +299,9 @@ class AttributeUpdater implements ObjectUpdaterInterface
 
         $attribute->setType($attributeType->getName());
         $attribute->setBackendType($attributeType->getBackendType());
-        $attribute->setUnique($attributeType->isUnique());
+        if (true === $attributeType->isUnique() || null === $attribute->isUnique()) {
+            $attribute->setUnique($attributeType->isUnique());
+        }
     }
 
     /**
--- a/CHANGELOG-3.0.md
+++ b/CHANGELOG-3.0.md
@@ -3,6 +3,7 @@
 ## Bug fixes
 
 - PIM-8163: Display messages when toggling curencies
+- PIM-6160: Fix update of 'unique' property in attribute updater when updating both unique and type properties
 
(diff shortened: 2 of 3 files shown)


src/Akeneo/Pim/Structure/Component/Updater/AttributeUpdater.php
```diff
@@ -299,7 +299,11 @@ protected function setType($attribute, $data)
 
         $attribute->setType($attributeType->getName());
         $attribute->setBackendType($attributeType->getBackendType());
-        $attribute->setUnique($attributeType->isUnique());
+        if (true === $attributeType->isUnique()) {
+            $attribute->setUnique(true);
+        } elseif (null === $attribute->isUnique()) {
+            $attribute->setUnique($attributeType->isUnique());
```
tseho: here, `$attributeType->isUnique()` will always be `false` ?

jmleroux: > here, `$attributeType->isUnique()` will always be `false` ?

No, it is the default value value for the type and it can be true.

tseho: line 302, you already handle the case where it's true.

mmetayer: @tseho you're right, in the `elseif` it will always be `false`. I will refactor to make it more understandable

mmetayer: BTW I find the `AttributeType::isUnique()` method poorly named. should be smth like `mustBeUnique()` IMO

PR #11427, opened 2020-01-17
API-946: Add e2e on fetch Audit events

pchasle:

Add e2e on fetch Audit events

I had to add the end_date as http query parameter to be able to do those tests.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Akeneo/Connectivity/Connection/back/Infrastructure/InternalApi/Controller/AuditController.php
+++ b/src/Akeneo/Connectivity/Connection/back/Infrastructure/InternalApi/Controller/AuditController.php
@@ -28,10 +28,13 @@ class AuditController
     public function sourceConnectionsEvent(Request $request): JsonResponse
     {
         $eventType = $request->get('event_type', '');
-        $startPeriod = new \DateTime('-7 days', new \DateTimeZone('UTC'));
-        $today = new \DateTime('now', new \DateTimeZone('UTC'));
+        $endDate = $request->get(
+            'end_date',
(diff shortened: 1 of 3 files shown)


src/Akeneo/Connectivity/Connection/back/Infrastructure/InternalApi/Controller/AuditController.php
```diff
@@ -28,10 +28,13 @@ public function __construct(CountDailyEventsByConnectionHandler $countDailyEvent
...
-        $startPeriod = new \DateTime('-7 days', new \DateTimeZone('UTC'));
-        $today = new \DateTime('now', new \DateTimeZone('UTC'));
+        $endDate = $request->get(
+            'end_date',
+            (new \DateTime('now', new \DateTimeZone('UTC')))->format('Y-m-d')
+        );
+        $startDate = (new \DateTime($endDate, new \DateTimeZone('UTC')))->modify('-7 day')->format('Y-m-d');
 
-        $query = new CountDailyEventsByConnectionQuery($eventType, $startPeriod->format('Y-m-d'), $today->format('Y-m-d'));
+        $query = new CountDailyEventsByConnectionQuery($eventType, $startDate, $endDate);
```
fitn: Seems less readable like this from my point of view. What does it change?

pchasle: If the `$endDate` is already given as query parameter then it's already a string and we don't need/can't call format on it.

PR #11436, opened 2020-01-17
Octopus 4.0.x fixes

pchasle:


fitn:
Everything has already been reviewed so that's ok to not have 2 approvals :)

PR #11463, opened 2020-01-22
API-952: Adds phpstan static analysis with different levels depending of the stack

fitn:

`OK`: Done / Validated
`-`: Not needed


make-file/connectivity-connection.mk
```diff
@@ -82,6 +83,9 @@ else
 	APP_ENV=test ${PHP_RUN} vendor/bin/phpunit -c . --testsuite Akeneo_Connectivity_Connection_EndToEnd $(0)
 endif
 
+connectivity-connection-cs-fix:
```
pierrejolly: Maybe we should not define a new make command since she do the exact same thing than the one in the Makefile

fitn:
Last commit just removed a useless make command for cs-fix so I don't relaunch the CI as it was ok before :)

PR #11482, opened 2020-01-24
Revert "AST-232: Remove GD from requirements"

jmleroux:
GD is required when there is an image in the PDF

See https://github.com/dompdf/dompdf/wiki/Requirements


BitOne:
Can you add a test on the PDF Render with an image? In order to avoid being fooled next time... Thank you!

pierallard:
Miss Changelog !

jmleroux:
> Miss Changelog !

Do we really need a changelog entry when reverting a requirement that was not documented?

PR #11502, opened 2020-01-28
PIM-9067: fix product filter on mass edit action

nmarniesse:

On product grid when we filter on parent code and we select all rows, the variant product are not taken in account in bulk action.  
This is because we filter on a root product model and the parent of the variant products is the sub product model, not the root. So the products are not retrieved by the query.  
The selected solution is to use `ancestors.codes` (that contains all the parents) instead of `parent`.  

This will fix:
- number of products that are concerned by the bulk action bfore launch the job
- product fetching during the job

**Before**

No product number displayed  

**After**

image

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Oro/Bundle/PimDataGridBundle/Extension/MassAction/MassActionDispatcher.php
+++ b/src/Oro/Bundle/PimDataGridBundle/Extension/MassAction/MassActionDispatcher.php
@@ -122,6 +122,8 @@ class MassActionDispatcher
             }
         }
 
+        $filters = $this->convertParentFieldIfAllRowsAreSelected($filters);
+
         return $filters;
     }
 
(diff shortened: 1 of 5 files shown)


src/Oro/Bundle/PimDataGridBundle/Extension/MassAction/MassActionDispatcher.php
```diff
@@ -287,4 +289,44 @@ public function getMassActionByNames($actionName, $datagridName)
 
         return $this->getMassActionByName($actionName, $datagrid);
     }
+
+    /**
+     * When all rows are selected and the filter is activated on a root product model, then the query
+     * returns the sub product model bit not the variant product.
```
jmleroux:
```suggestion
     * returns the sub product model but not the variant product.
```

Issue #11481, opened 2020-01-24
Fix random: tests/legacy/features/pim/enrichment/product/datagrid/datagrid_views.feature:50


fitn:
And I should see the text "Some shoes": Spin : timeout of 20 excedeed, with message : Cannot find the text "Some shoes" (Context\Spin\TimeoutException)

https://circleci.com/gh/akeneo/pim-community-dev/26721#tests/containers/1

ahocquard:
After investigation, it's because we reach the timeout during the last "And I should see the text "Some shoes"".

The method to get the text takes too much time: 
```
    And I should see the text "Some shoes"                                  # Context\AssertionContext::assertPageContainsText()
      │ float(2.2598869800568)
      │ float(3.8870730400085)
      │ float(3.779876947403)
```

As we reduced the timeout... it's unstable.

ahocquard:
Maybe fixed by incrementing the timeout...

PR #11517, opened 2020-01-30
Stabilize behat again

ahocquard:
Fix some unstable:
(tests/legacy/features/pim/enrichment/product/datagrid/datagrid_views.feature:50)

https://github.com/akeneo/pim-community-dev/issues/11481

The method to assert that a text is in the page take  A LOT of time. So when it's spinning, it can reach the timeout that I reduced to 20 sec.

I reduced it because a lot of time we wait something does not happen during the time of the timeout (a text does not appear, during 20 seconds for example).

Let's try with 25...

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/behat.yml
+++ b/behat.yml
@@ -1,7 +1,7 @@
 contexts: &context
     -   Context\FeatureContext:
             -   base_url: 'http://httpd-behat/'
-                timeout: 20000
+                timeout: 25000
                 window_width: 1280
                 window_height: 1024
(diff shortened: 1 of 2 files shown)


ahocquard:
@mmetayer yes but I will let people improve it if they want, because I'm pretty exhausted that only a few parts of the team (you and me, let's be honest) do it.

PR #11565, opened 2020-02-12
API-844: As Peter, I want to be warned when I use a wrong user with the client API

willy-ahva:
Since we moved from API connections to Connections, we generate a Client and a new associated User (a special one for the API) for each new Connection. But it is still possible to connect through the API with client A and a User B.

If you to do this you lose the Audit feature because we are not able to monitor data and you lose permissions associated with user A.

As the long term goal is to forbid this behavior and to force the login with Client A / User A, we want to display in Connection screen (List + Edit) a warning as long as you log in with a wrong credentials combination.

| Added integration tests           | - Replaced by E2E backend
| Migration script                  | Done

`OK`: Done / Validated
`-`: Not needed


src/Akeneo/Connectivity/Connection/back/Infrastructure/Cli/GenerateAuditDataCommand.php
```diff
@@ -0,0 +1,63 @@
...
+declare(strict_types=1);
+
+namespace Akeneo\Connectivity\Connection\Infrastructure\Cli;
+
+use Doctrine\DBAL\Connection;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+
+class GenerateAuditDataCommand extends Command
```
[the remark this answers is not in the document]
willy-ahva: yes I shouldn't have commited it.

src/Akeneo/Tool/Bundle/ApiBundle/Security/OAuth2.php
```diff
@@ -14,15 +20,54 @@
...
+        $authHeaders = $this->getAuthorizationHeader($request);
+        $clientCredentials = $this->getClientCredentials($inputData, $authHeaders);
+        $clientId = substr($clientCredentials[0], 0, strpos($clientCredentials[0], '_'));
+
+        $this->eventDispatcher->dispatch(
+            new ApiAuthenticationEvent($request->get('username'), $clientId)
+        );
+
+        return $response;
+    }
```
pchasle: Was there an issue with using the [fos_oauth_server.post_authorization_process](https://github.com/FriendsOfSymfony/FOSOAuthServerBundle/blob/master/Resources/doc/the_oauth_event_class.md) event instead ? To avoid extending the OAuth2 class internals.

src/Akeneo/Connectivity/Connection/front/src/settings/reducers/wrong-credentials-combination-reducer.ts
```diff
@@ -0,0 +1,15 @@
+import {Reducer} from 'react';
+import {WrongCredentialsCombinations} from '../../model/wrong-credentials-combinations';
+import {
+    WRONG_CREDENTIALS_COMBINATIONS_FETCHED,
+    WrongCredentialsCombinationsActions,
+} from '../actions/wrong-credentials-combinations-actions';
+
+export const reducer: Reducer<WrongCredentialsCombinations, WrongCredentialsCombinationsActions> = (_, action) => {
```
pchasle: No need to change it, but using a reducer is a little overkill here and having a state/setState inside a context would have be fine.

src/Akeneo/Connectivity/Connection/front/src/common/components/SmallHelper.tsx
```diff
@@ -29,9 +35,14 @@ const HintTitle = styled.div`
...
-    <SubsectionHint className='AknSubsection'>
-        <HintIcon />
+type Props = {
+    children: ReactNode;
+    level?: 'info' | 'warning';
+};
+
+export const SmallHelper = ({children, level = 'info'}: Props) => (
+    <SubsectionHint className='AknSubsection' level={level}>
+        <HintIcon level={level} />
```
pchasle: It's syntactic sugar, but we can have something like that:
```tsx
type Props = PropsWithChildren<{ warning: boolean } | { info: boolean }>

export const SmallHelper = ({children, ...props}: Props) => {
    const level = 'warning' in props ? 'warning' : 'info';
    return (<SubsectionHint className='AknSubsection' level={level}>
        <HintIcon level={level} />
```
To be able to use it like this:
```tsx
<SmallHelper warning>...</SmallHelper>
<SmallHelper info>...</SmallHelper>
```

src/Akeneo/Connectivity/Connection/front/src/settings/components/ConnectionCredentials.tsx
```diff
@@ -35,6 +39,23 @@ export const ConnectionCredentials: FC<Props> = ({code, label, credentials: cred
...
+                                translate={translate}
+                            />
+                        ) : (
+                            <SeveralWrongCombinationsWarning
+                                combinations={wrongCombination}
+                                goodUsername={credentials.username}
+                                translate={translate}
+                            />
+                        )}
+                    </SmallHelper>
```
pchasle: This should be split in (at least) another file/component.
And it would probably be even lighter/clearer to have something like `<DocumentationLink />` and `<WrongCredentialsCombinationsList />` outside `<OneWrongCombinationWarning />` and `<SeveralWrongCombinationsWarning />`

NolwennP:
Link Help center PR: https://github.com/akeneo/pim-helpcenter/pull/484

src/Akeneo/Connectivity/Connection/back/tests/Unit/spec/Domain/WrongCredentialsConnection/Model/Read/WrongCredentialsCombinationSpec.php
```diff
@@ -0,0 +1,86 @@
...
+        $this->users()->shouldReturn([]);
+    }
+
+    public function it_adds_and_provides_users(): void
+    {
+        $firstDate = new \DateTime('2019-05-15 16:25:00');
+        $this->addUser('bynder', $firstDate);
+        $this->users()->shouldReturn(['bynder' => $firstDate]);
+
+        $dadadate = new \DateTime();
```
pchasle: `$secondDate` ? :smile:

PR #11599, opened 2020-02-19
API-995: Dispatch event on products read

fitn:
We need to track products that have been sent through the API to get some statistics around that.

It introduces the event dispatcher inside the query/handler.

`OK`: Done / Validated
`-`: Not needed


src/Akeneo/Pim/Enrichment/Component/Product/Event/ProductsReadEvent.php
```diff
@@ -0,0 +1,29 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Akeneo\Pim\Enrichment\Component\Product\Event;
```
nmarniesse: If you receive an `Akeneo\Pim\Enrichment\Component\Product\Event\ProductsReadEvent` event, how do you know the source of the action is the API? Do you consider it's an event for all use case, or for API only?  

If the class is in `Akeneo\Pim\Enrichment\Component\Product\Connector\Event`, I guess it will be easier to create multiple read events (by use case for example).

src/Akeneo/Pim/Enrichment/Component/Product/Event/ProductsReadEvent.php
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * @author Romain Monceau <romain@akeneo.com>
+ * @copyright 2020 Akeneo SAS (http://www.akeneo.com)
+ * @license http://opensource.org/licenses/osl-3.0.php Open Software License (OSL 3.0)
+ */
+final class ProductsReadEvent
```
nmarniesse: Maybe `ReadProductsEvent` should be better?

src/Akeneo/Pim/Enrichment/Component/Product/Event/ProductsReadEvent.php
fitn: In fact, the goal is not to know there this kind of information. It's the job of the listener/subscriber.
We will get this information from the UserContext :)

We create this event for all use cases. We can also move it into the Controller if you feel it's better :+1: Let me know.

I don't have other use cases than this one and I feel we can do almost everything based on this one for reading purpose. We have the product id, then we can track everything.

src/Akeneo/Pim/Enrichment/Component/Product/Event/ProductsReadEvent.php
fitn: Will do! :+1:

PR #11611, opened 2020-02-21
#11583: Remove useless GLOB_BRACE flag from standard Kernel

### LINKED ISSUE #11583: GLOB_BRACE used in Kernel.php is not compatible with all Linux based OS's
**I'm reporting a Bug**

I am installing a Akeneo 4.0.5 on php 7.3.14 on a Alpine distribution and having the fallowing warning

```
Warning: use of undefined constant GLOB_BRACE
```
This is due to the usage of this constant in the [Kernel.php](https://github.com/akeneo/pim-community-dev/blob/4.0/std-build/Kernel.php#L115)

According to the PHP [documentation](https://www.php.net/manual/fr/function.glob.php)
> Note: The GLOB_BRACE flag is not available on some non GNU systems, like Solaris.

The symfony load using the finder works perfectly as it has a workaround. So IMHO that should be used instead of using glob directly.

Regards,

jmleroux:
Fixes #11583

> the braces could be removed altogether, as they only list one variant. "/{packages}/" is equivalent to "/packages/" as there is only the variant "packages". It would be a different story if it was "/{packages,}/" (notice the empty variant) or "/{packages,routes}/".

### DIFF
--- a/std-build/Kernel.php
+++ b/std-build/Kernel.php
@@ -112,9 +112,9 @@ class Kernel extends BaseKernel
     private function loadPackagesConfigurationExceptSecurity(LoaderInterface $loader, string $confDir, string $environment): void
     {
         $files = array_merge(
-            glob($confDir . '/{packages}/*.yml', GLOB_BRACE),
-            glob($confDir . '/{packages}/' . $environment . '/*.yml', GLOB_BRACE),
-            glob($confDir . '/{packages}/' . $environment . '/**/*.yml', GLOB_BRACE)
+            glob($confDir . '/{packages}/*.yml),
+            glob($confDir . '/{packages}/' . $environment . '/*.yml'),
+            glob($confDir . '/{packages}/' . $environment . '/**/*.yml')
         );
 
         $files = array_filter(
--- a/std-build/Kernel.php
+++ b/std-build/Kernel.php
@@ -112,9 +112,9 @@ class Kernel extends BaseKernel
     private function loadPackagesConfigurationExceptSecurity(LoaderInterface $loader, string $confDir, string $environment): void
     {
         $files = array_merge(
-            glob($confDir . '/{packages}/*.yml),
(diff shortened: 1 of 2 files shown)


std-build/Kernel.php
```diff
@@ -112,9 +112,9 @@ private function loadPackagesConfiguration(LoaderInterface $loader, string $conf
     private function loadPackagesConfigurationExceptSecurity(LoaderInterface $loader, string $confDir, string $environment): void
     {
         $files = array_merge(
-            glob($confDir . '/{packages}/*.yml', GLOB_BRACE),
-            glob($confDir . '/{packages}/' . $environment . '/*.yml', GLOB_BRACE),
-            glob($confDir . '/{packages}/' . $environment . '/**/*.yml', GLOB_BRACE)
+            glob($confDir . '/{packages}/*.yml),
```
l-lotz: There is a syntax error the terminating ' is missing. Also /{packages}/ should then be converted to /packages/ if GLOB_BRACE is removed.

jmleroux: Oops, done too quick on GIthub :)

std-build/Kernel.php
```diff
@@ -112,9 +112,9 @@ private function loadPackagesConfiguration(LoaderInterface $loader, string $conf
     private function loadPackagesConfigurationExceptSecurity(LoaderInterface $loader, string $confDir, string $environment): void
     {
         $files = array_merge(
-            glob($confDir . '/{packages}/*.yml', GLOB_BRACE),
-            glob($confDir . '/{packages}/' . $environment . '/*.yml', GLOB_BRACE),
-            glob($confDir . '/{packages}/' . $environment . '/**/*.yml', GLOB_BRACE)
+            glob($confDir . '/packages/*.yml'),
```
ahocquard: using the `$loader` is better I think

l-lotz: I think the loader is not used, because all security.yml files are filtered out later in the array_filter.

PR #11628, opened 2020-02-24
PIM-9112: Add a migration for removing empty metric values

tseho:

If the user creates a metric attribute in a PIM 3.2, links it to an existing family (that contains already products), by default the metric attribute is filled like that:

```
"my_metric":[{
  "locale":null,
  "scope":null,
  "data": {"amount": null, "unit": null}
}]
```
When upgrading to 4.0+, an error is thrown because null values are not allowed anymore.
This case was not handled by the existing `remove_product_model_empty_raw_values` migration.

This PR adds a migration for removing empty metric values.

### DIFF
--- a/upgrades/schema/Version_5_0_20200224123916_remove_product_empty_metric_values.php
+++ b/upgrades/schema/Version_5_0_20200224123916_remove_product_empty_metric_values.php
@@ -131,7 +131,7 @@ final class Version_5_0_20200224123916_remove_product_empty_metric_values
             $rows = $this->connection->executeQuery($sql)->fetchAll();
 
             if (count($rows) === 0) {
-                return null;
+                break;
             }
 
             foreach ($rows as $row) {
--- a/upgrades/schema/Version_5_0_20200224123916_remove_product_empty_metric_values.php
+++ b/upgrades/schema/Version_5_0_20200224123916_remove_product_empty_metric_values.php
@@ -1,0 +1,154 @@
+<?php declare(strict_types=1);
+
+namespace Pim\Upgrade\Schema;
+
+use Akeneo\Pim\Enrichment\Bundle\Elasticsearch\Indexer\ProductIndexer;
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\DBAL\Types\Types;
+use Doctrine\Migrations\AbstractMigration;
+use Symfony\Component\DependencyInjection\ContainerAwareInterface;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * This migration will delete the empty values from metric values of products.
+ * For example, the value {attr: {<all_channels>: {<all_locales>: {"amount": null, "unit": null}}}} will be removed
+ * from the raw_values field.
+ */
+final class Version_5_0_20200224123916_remove_product_empty_metric_values
+    extends AbstractMigration
+    implements ContainerAwareInterface
+{
+    private const MYSQL_BATCH_SIZE = 1000;
+    private const ELASTICSEARCH_BATCH_SIZE = 100;
+
+    /** @var ContainerInterface */
+    private $container;
+
+    /**
+     * {@inheritdoc}
+     */
+    public function setContainer(ContainerInterface $container = null)
+    {
+        $this->container = $container;
+    }
+
+    public function up(Schema $schema): void
+    {
+        $this->addSql('SELECT "disable migration warning"');
+
+        /** @var string[] $metricAttributesCodes */
+        $metricAttributesCodes = $this->findMetricAttributesCodes();
+
+        $productIdentifiersToIndex = [];
+
+        $rows = $this->getAllProducts();
+        foreach ($rows as $i => $row) {
+            $values = json_decode($row['raw_values'], true);
+
+            if (empty(array_intersect(array_keys($values), $metricAttributesCodes))) {
+                continue;
+            }
+
+            $cleanValues = $this->cleanMetricValues($values, $metricAttributesCodes);
+
+            if ($values !== $cleanValues) {
+                $this->connection->executeQuery(
+                    'UPDATE pim_catalog_product SET raw_values = :rawValues WHERE identifier = :identifier',
+                    [
+                        'rawValues' => json_encode($cleanValues, JSON_FORCE_OBJECT),
+                        'identifier' => $row['identifier'],
+                    ], [
+                        'rawValues' => Types::STRING,
+                        'identifier' => Types::STRING,
+                    ]
+                );
+
+                $productIdentifiersToIndex[] = $row['identifier'];
+                if (count($productIdentifiersToIndex) % self::ELASTICSEARCH_BATCH_SIZE === 0) {
+                    $this->getProductIndexer()->indexFromProductIdentifiers($productIdentifiersToIndex);
+                    $productIdentifiersToIndex = [];
+                }
+            }
+        }
+
+        if (!empty($productIdentifiersToIndex)) {
+            $this->getProductIndexer()->indexFromProductIdentifiers($productIdentifiersToIndex);
+        }
+    }
+
+    private function cleanMetricValues(array $values, array $metricAttributesCodes): array
+    {
+        $results = [];
+
+        foreach ($values as $attributeCode => $channelValues) {
+            foreach ($channelValues as $channel => $localeValues) {
+                foreach ($localeValues as $locale => $data) {
+                    if (!in_array($attributeCode, $metricAttributesCodes)) {
+                        $results[$attributeCode][$channel][$locale] = $data;
+                        continue;
+                    }
+
+                    if ($this->isMetricFilled($data)) {
+                        $results[$attributeCode][$channel][$locale] = $data;
+                    }
+                }
+            }
+        }
+
+        return $results;
+    }
+
+    private function isMetricFilled($data)
+    {
+        if (null === $data) {
+            return false;
+        }
+
+        if (!is_array($data)) {
+            return false;
+        }
+
+        return isset($data['unit']) && isset($data['amount']);
+    }
+
+    private function findMetricAttributesCodes(): array
+    {
+        $sql = "SELECT code FROM pim_catalog_attribute WHERE attribute_type = 'pim_catalog_metric'";
+
+        return $this->connection->executeQuery($sql)->fetchAll(\PDO::FETCH_COLUMN);
+    }
+
+    private function getAllProducts(): \Generator
+    {
+        $lastId = null;
+
+        while (true) {
+            $sql = sprintf(
+                "SELECT identifier, raw_values FROM pim_catalog_product %s ORDER BY identifier LIMIT %d",
+                $lastId !== null ? sprintf('WHERE identifier > "%s"', $lastId) : '',
+                self::MYSQL_BATCH_SIZE
+            );
+
+            $rows = $this->connection->executeQuery($sql)->fetchAll();
+
+            if (count($rows) === 0) {
+                return null;
+            }
+
+            foreach ($rows as $row) {
+                yield $row;
+                $lastId = $row['identifier'];
+            }
+        }
+    }
+
+    private function getProductIndexer(): ProductIndexer
+    {
+        return $this->container->get('pim_catalog.elasticsearch.indexer.product');
+    }
+
+    public function down(Schema $schema): void
+    {
+        $this->throwIrreversibleMigrationException();
+    }
+}
--- a/upgrades/schema/Version_5_0_20200224123916_remove_product_empty_metric_values.php
+++ b/upgrades/schema/Version_5_0_20200224123916_remove_product_empty_metric_values.php
@@ -19,7 +19,7 @@ final class Version_5_0_20200224123916_remove_product_empty_metric_values
     implements ContainerAwareInterface
 {
     private const MYSQL_BATCH_SIZE = 1000;
-    private const ELASTICSEARCH_BATCH_SIZE = 100;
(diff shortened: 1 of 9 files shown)


upgrades/schema/Version_5_0_20200224123916_remove_product_empty_metric_values.php
```diff
@@ -0,0 +1,154 @@
...
+ * This migration will delete the empty values from metric values of products.
+ * For example, the value {attr: {<all_channels>: {<all_locales>: {"amount": null, "unit": null}}}} will be removed
+ * from the raw_values field.
+ */
+final class Version_5_0_20200224123916_remove_product_empty_metric_values
+    extends AbstractMigration
+    implements ContainerAwareInterface
+{
+    private const MYSQL_BATCH_SIZE = 1000;
+    private const ELASTICSEARCH_BATCH_SIZE = 100;
```
ahocquard: 1000 is ok too

upgrades/schema/Version_5_0_20200224123916_remove_product_empty_metric_values.php
```diff
@@ -0,0 +1,154 @@
...
+        /** @var string[] $metricAttributesCodes */
+        $metricAttributesCodes = $this->findMetricAttributesCodes();
+
+        $productIdentifiersToIndex = [];
+
+        $rows = $this->getAllProducts();
+        foreach ($rows as $i => $row) {
+            $values = json_decode($row['raw_values'], true);
+
+            if (empty(array_intersect(array_keys($values), $metricAttributesCodes))) {
```
ahocquard: FYI, array_interset_key is much more efficient (O(n) vs O(n*m). Maybe it matters for big raw value collections. 

With Pierre, we did a x10 when working on benchmarks on their new topic on associations. Choosing the correct data structure is important when we manipulate raw values.

upgrades/schema/Version_5_0_20200224123916_remove_product_empty_metric_values.php
```diff
@@ -0,0 +1,154 @@
...
+        if (!empty($productIdentifiersToIndex)) {
+            $this->getProductIndexer()->indexFromProductIdentifiers($productIdentifiersToIndex);
+        }
+    }
+
+    private function cleanMetricValues(array $values, array $metricAttributesCodes): array
+    {
+        $results = [];
+
+        foreach ($values as $attributeCode => $channelValues) {
```
ahocquard: Not very optimal. You iterate over all the values of all the products, whatever the number of the metricAttributeCode. It's O(m*n), where n is the number of values and m the number of products.
In the code I propose you, it O(m\*y),  where y is the number metric attribute code and m the number of products. y <= n, and is very small most of the time.

```
foreach ($metricAttributesCodes as $metricAttributeCode) {
    if (!isset($values[$metricAttributeCode]) {
           continue; // no need to do array_diff anymore before this function
     }

    $newValue = [];
    foreach ($values[$metricAttributeCode] as $channel => $locale) {
         foreach ($localeValues as $locale => $data) {
                 if ($this->isMetricFilled($data)) {
                    $newValue[$channel][$locale] = $data;
                 }
          }
    }
     if (!empty($newValue)) {
            $values[$metricAttributeCode] =  $newValue;
      }
}
```

upgrades/schema/Version_5_0_20200224123916_remove_product_empty_metric_values.php
```diff
@@ -0,0 +1,154 @@
...
+            $sql = sprintf(
+                "SELECT identifier, raw_values FROM pim_catalog_product %s ORDER BY identifier LIMIT %d",
+                $lastId !== null ? sprintf('WHERE identifier > "%s"', $lastId) : '',
+                self::MYSQL_BATCH_SIZE
+            );
+
+            $rows = $this->connection->executeQuery($sql)->fetchAll();
+
+            if (count($rows) === 0) {
+                return null;
```
ahocquard: return an empty array, it's typehint with `array`

tseho: Would `return;` be better ? The function returns a Generator, it should yield the results and exit where there is none anymore.

ahocquard: actually, this check is not needed. It will foreach on nothing: end of the generator. I checked some code examples and I never found any `return`.

tseho: But I have a `while (true)`. The result is the same but would it make sens to have a `break` ?

ahocquard: :+1:  `break` seems ok in that case.

jjanvier:
Before merging, we should test on a real size catalog. We (black hawks) can help you for that.

Doodoune:
FYI - Fix tested on tdf-14 with the database of which production deployed on it. Don't know if it's a "real size catalog".

tseho:
@jjanvier As Elodie said, I did run it on a copy of a customer database (~850Mb), it took ~10s.
If you want a test on a bigger one, I will see with you how to do it.

jjanvier:
@BitOne this is more or less what was done for https://github.com/akeneo/pim-community-dev/blob/master/upgrades/schema/Version_4_0_20200116122239_remove_product_empty_raw_values.php. And that worked well on our SaaS customers.

tseho:
Exactly, It's really inspired by this migration.

jjanvier:
@tseho what's the number of products and products models on this database? I would indeed be more confident if we could try on other envs.

tseho:
@jjanvier 10255 products, 0 product models.

jjanvier:
So, that's not significant at all. We need to find a catalog with a high number of products/models and that would be impacted by your fix.

BitOne:
The previous migration that used to go through all products did it because it was meant to clean up null values for all kind of attribute type. So it was a good compromise between efficiency and complexity.

Here, we are targeting only metric attributes, so we must be able to find exactly the products that should be impacted by the migration.

Moreover, having the SQL query that is able to find the products that should be fixed will help us find the most relevant catalog to test this migration.

tseho:
Proposed query:
```
SELECT JSON_EXTRACT(raw_values, '$.*.*.*.unit') AS value
FROM pim_catalog_product
WHERE JSON_CONTAINS_PATH(raw_values, 'one', '$.tech_spec_weight_with_keyboard')
AND JSON_CONTAINS(JSON_EXTRACT(raw_values, '$.*.*.*.unit'), 'null')
ORDER BY identifier;
```
Where $.tech_spec_weight_with_keyboard is the list of attribute codes which is a metric.

ahocquard:
I'm dubitative at being  faster with a single SQL query if the goal is to reduce the selection part (number of products). It's not very different from doing it in PHP except some IO... I'm curious in tern of speed if you have number, but PHP is very fast at handling JSON. Sometimes, it's even faster than pure Mysql... 
Actually, the algorithm is even worse as you use wildcard `JSON_EXTRACT(raw_values, '$.*.*.*.unit')`. It means that it has to check in every values, which is not the case of the PHP algorithm.

If it's slow or speed is almost the same:
I would do an ES research. Why? Because a big SQL query will do a FULL TABLE SCAN exactly like the current state of this PR, and I already had an experience on another query to do such thing on a real customer catalog. The speed was approximately 2min for 100 000 products with Mysql 5.7.

Even if Mysql 8.0 is maybe 2 time faster to handle JSON, it's 10min for 1 million products. I don't think that it's fast enough. We could select product ids in the database with ES having null currencies. Everything is already indexed. I should take only few hundred ms.

tseho:
I'm aware that this query is worst, the time was higher on my demo database.
It was just to try to return only affected products but both `JSON_CONTAINS_PATH` with the full list of metrics and `'$.*.*.*.unit'` are far from efficient.
Btw, @BitOne took the time to look into it and there are more products affected than expected, in big databases, something like up to 30% of products.
With this new information, I think @BitOne want to try the current version on a duplicated environment.

If it's not good enough, we can look into an ES query, as you suggest.

tseho:
I've updated both the code and the tests, we now also cleanup the price collections.

upgrades/test_schema/Version_5_0_20200224123916_remove_product_empty_amount_values_Integration.php
```diff
@@ -21,19 +21,19 @@ public function testItRemovesAllEmptyValues()
...
-    (NULL, NULL, NULL, NULL, 1, 'product7', '{"a_price": {"<all_channels>": {"<all_locales>": null}}}', NOW(), NOW()),
-    (NULL, NULL, NULL, NULL, 1, 'product8', '{"a_price": {"<all_channels>": {"<all_locales>": []}}}', NOW(), NOW()),
-    (NULL, NULL, NULL, NULL, 1, 'product9', '{"a_price": {"<all_channels>": {"<all_locales>": [{"amount": null, "currency": "USD"}]}}}', NOW(), NOW()),
-    (NULL, NULL, NULL, NULL, 1, 'product10', '{"a_price": {"<all_channels>": {"<all_locales>": [{"amount": "1329.00", "currency": "USD"}]}}}', NOW(), NOW()),
-    (NULL, NULL, NULL, NULL, 1, 'product11', '{"a_price": {"<all_channels>": {"<all_locales>": [{"amount": "1329.00", "currency": "USD"}, {"amount": "1329.00", "currency": "EUR"}]}}}', NOW(), NOW()),
-    (NULL, NULL, NULL, NULL, 1, 'product12', '{"a_price": {"<all_channels>": {"<all_locales>": [{"amount": "1329.00", "currency": "USD"}, {"amount": null, "currency": "EUR"}]}}}', NOW(), NOW())
+INSERT INTO pim_catalog_product (is_enabled, identifier, raw_values, created, updated) VALUES
```
tseho, line 24: :+1:

ahocquard:
@LaurentPetard We discussed about the selection of the products that should be done with ES. Also, is there a good reason to not re-index products? There can be weird behaviour with null values with optional values/ product without families / is empty filter during a research , so re-indexing seems cleaner.

LaurentPetard:
@ahocquard The problem with indexing the products (using the service ProductIndexer) is that it could crash in the future if the indexation call some code whose changes need a migration to work (this migration will not be executed before the migration of this PR). This problem already occurred in the past.
So if the re-ndexation is mandatory we will have to find a solution to do avoid this problem.

ahocquard:
@LaurentPetard the issue is more complex. The dependencies could change also. Anything in the code can impact the migration actually... 

It means that we can't index products in any migration scripts. That's not really acceptable.

Personally, I think that avoiding to re-index the products because the code could change is the wrong direction to take. Here, we ensure that Mysql is clean. But what if we assume that ES is consistent also in the future, whereas it's not? For example, we could read the products from ES instead of Mysql in the API in the future, and it will break in production.

There should be a clear direction from the @akeneo/squad-blackhawks about it, without weird workaround.

BitOne:
The good practice about migration has been given several times in the past: a migration must not rely on code outside of the migration itself. So this exclude indexing for example.
And indexing is a good example, because today we cannot migrate customers from 3.2 to Serenity because a migration is re-indexing some products; and of course the new code is relying on table that doesn't exist yet at the moment the migration is executed.

BitOne:
Of couse, I know that some migration relied on deep and complex business code (and some of them fails because of that by the way), and it's not acceptable as well to copy and paste the full business inside the migration.

But in this case, we are taking a big risk and I think this kind of migration must be handled in a different and very specific way.

tseho:
> a migration must not rely on code outside of the migration itself.
:+1: 

> today we cannot migrate customers from 3.2 to Serenity because we have a migration that is re-indexing some products
Yeah, the issue with the new measurement table breaking this migration https://github.com/akeneo/pim-community-dev/blob/89c1521b81ea2ad487083b2100fb35da5a780763/upgrades/schema/Version_4_0_20200116122239_remove_product_empty_raw_values.php was discovered after this PR was created and luckily before it was merged. We agreed it was not good to reproduce the same mistake and the current ticket got stuck at the maintenance.

With the lastest version proposed by our dear @LaurentPetard, there is no call to any business code done during the migration.
But yes, as @ahocquard said, products are not updated in elasticsearch. It will not affect things like `product without families` because we are only cleaning 2 types of attributes, measurements & prices. It far from perfect, but it's a compromise.

> But in this case, we are taking a big risk and I think this kind of migration must be handled in a different and very specific way.
What do you mean ?

BitOne:
I was talking about migration depending on outside code: it's bad practice and dangerous, but sometimes difficult to avoid. And in this case, these migrations are considered exceptional and usually it means we need to execute them with a version of the code matching the version of the migration. It's doable and quite painful and hard to do it.

But it seems that's not the case for this ticket, so that was more a general remark on the subject of migrations relying on external code.

ahocquard:
In general, it means that we have no easy way to be consistent between ES index and Mysql during a migration, and that's a quite big issue as it concerns the storage.

For this particular PR, not indexing is a workaround that could be costly. This is no an acceptable trade-off in my opinion. I already worked on tickets with null values/empty filters on ES queries from the datagrid, and it's a nightmare (several days of investigation from different people). Imho, ES should reflect Mysql data.

BitOne:
Yes, Elasticsearch should reflect MySQL data. But reindexing product is not the only way to do it.

ahocquard:
@BitOne are you thinking about an update of ES document? It could be quite tricky for the metrics but maybe doable (I don't know at which point you can update complex structures like metric values with ES).

BitOne:
If there"s a risk with ES data no updated, I want to know what it is, and what should be the data to avoid that risk.

Then a solution, or no solution is not needed, can be design.

PR #11634, opened 2020-02-25
[akeneo-design-system]: init lib and storybook configuration

nicomaligne:

This is a first step to a shared components library.
The Akeneo shared lib runs with react / styled components / react testing library.

You can run the storybook with yarn / npm storybook. 
The ExampleButton is just here to showcase storybook and the mdx format.
There is some typescript errors due to [duplicate identifier](https://github.com/styled-components/styled-components/issues/2842) in styled components types dependencies.
A yarnclean file has beed added to remove unnecessary types React native and avoid the console spam.

jest.config.js is still a work in progress.

`OK`: Done / Validated
`-`: Not needed


akeneo-design-system/.prettierrc
```diff
@@ -0,0 +1,13 @@
+{
+    "arrowParens": "avoid",
+    "bracketSpacing": false,
+    "jsxBracketSameLine": false,
+    "jsxSingleQuote": true,
+    "printWidth": 120,
+    "quoteProps": "as-needed",
+    "semi": true,
+    "singleQuote": true,
+    "tabWidth": 4,
```
ValentinMumble:
```suggestion
    "tabWidth": 2,
```
As in https://github.com/akeneo/pim-community-dev/blob/master/.prettierrc.json

akeneo-design-system/src/components/ExampleButton/ExampleButton.tsx
```diff
@@ -0,0 +1,54 @@
...
+  height: 3rem;
+  width: 10rem;
+  border-radius: 3%;
+  &:hover {
+    cursor: ${props => (props.disabled ? "not-allowed" : "pointer")};
+  }
+`;
+
+type ExampleButtonProps = {
+    children: any,
```
ValentinMumble:
```suggestion
    children: React.ReactNode,
```
?

akeneo-design-system/.prettierrc
nicomaligne: We have to decide about this ;) 
https://github.com/akeneo/pim-community-dev/blob/master/src/Akeneo/Connectivity/Connection/front/.prettierrc.json#L10
https://github.com/akeneo/pim-community-dev/blob/master/src/Akeneo/Connectivity/Connection/front/.editorconfig#L6
make one file to rule them all, if everyone is ok.

akeneo-design-system/src/components/ExampleButton/ExampleButton.tsx
nicomaligne: Yep sure 👍 
Like I said this code will be delete in the next pr

akeneo-design-system/.prettierrc
ValentinMumble: AH! I wasn't even aware that they had a custom prettier config in `Connectivity/Connection` ;)

.gitignore
```diff
@@ -42,3 +42,6 @@ coverage/*
 ###> symfony/web-server-bundle ###
 /.web-server-pid
 ###< symfony/web-server-bundle ###
+
+
+
```
tseho: ?

akeneo-design-system/extend-jest.d.ts
```diff
@@ -0,0 +1,5 @@
+declare namespace jest {
+    interface Matchers<R> {
+      toHaveStyleRule: import("jest-styled-components").jest.Matchers<R>["toHaveStyleRule"]
+    }
+  }
```
tseho: formatting ?

akeneo-design-system/Makefile
```diff
@@ -0,0 +1,18 @@
+
+DOCKER_PORT_STORYBOOK = 6006
+_YARN = docker run -u node -v `pwd`:/srv/ads -w /srv/ads -p $(DOCKER_PORT_STORYBOOK):80 --rm -it node:13 yarn
+
+# Install dependencies
+_yarn:
+	$(_YARN)
```
tseho: IMHO, I would rather specify `install --frozen-lockfile` instead of the fallback on the default behavior (which is `install`)

akeneo-design-system/jest.config.js
```diff
@@ -0,0 +1,179 @@
+// For a detailed explanation regarding each configuration property, visit:
```
tseho: IMHO, you can cleanup this file and only keep used options, the link to the doc is enough.

akeneo-design-system/src/components/Buttons/CoreButton/CoreButton.test.tsx
```diff
@@ -0,0 +1,74 @@
+import React from 'react';
+import 'jest-styled-components';
+
+import { CoreButton } from './CoreButton';
+import { render, fireEvent } from '@testing-library/react';
+
+describe('testing button', () => {
+    test('should display a button with the given children', () => {
+        // Given
```
tseho: We can discuss about `Given/When/Then` but honestly, they don't bring a real value to the test readability.

pchasle: I think it does, and it eases the reading/mental workload, but it's maybe not obvious on simple components like this one.

akeneo-design-system/Makefile
pchasle: Good catch, it's important for the CI.

PR #11638, opened 2020-02-25
 PIM-9114: Fix "EMPTY" operator for ParentFilter on mass actions

tseho:

The fix introduced by https://github.com/akeneo/pim-community-dev/pull/11502 does not work for the quick export, which is a special mass action.
In this previous fix, the filter 'parent' is dynamically replaced by 'ancestor.code' when all the rows are selected.
This is not working when the operator is 'EMPTY', because the `FilterParent` is supporting the operator 'EMPTY' but `AncestorCodeFilter` does not.

The dirty solution proposed is to only apply this replacement for other filters than 'EMPTY'. 

### DIFF
--- a/src/Akeneo/Pim/Enrichment/Bundle/Elasticsearch/Filter/Field/AncestorCodeFilter.php
+++ b/src/Akeneo/Pim/Enrichment/Bundle/Elasticsearch/Filter/Field/AncestorCodeFilter.php
@@ -70,14 +70,6 @@ class AncestorCodeFilter extends AbstractFieldFilter
                 ];
                 $this->searchQueryBuilder->addMustNot($mustNotClause);
                 break;
-            case Operators::IS_EMPTY:
-                $clause = [
-                    'exists' => [
-                        'field' => self::ANCESTOR_CODES_ES_FIELD,
-                    ],
-                ];
-                $this->searchQueryBuilder->addMustNot($clause);
-                break;
             default:
                 throw InvalidOperatorException::notSupported($operator, static::class);
         }
--- a/src/Akeneo/Pim/Enrichment/Bundle/Elasticsearch/Filter/Field/AncestorCodeFilter.php
+++ b/src/Akeneo/Pim/Enrichment/Bundle/Elasticsearch/Filter/Field/AncestorCodeFilter.php
@@ -76,7 +76,12 @@ class AncestorCodeFilter extends AbstractFieldFilter
                         'field' => self::ANCESTOR_CODES_ES_FIELD,
                     ],
                 ];
-                $this->searchQueryBuilder->addFilter($clause);
+                $this->searchQueryBuilder->addMustNot($clause);
+
+                $familyExistsClause = [
+                    'exists' => ['field' => 'family.code']
+                ];
+                $this->searchQueryBuilder->addFilter($familyExistsClause);
                 break;
             default:
                 throw InvalidOperatorException::notSupported($operator, static::class);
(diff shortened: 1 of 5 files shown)


src/Akeneo/Pim/Enrichment/Bundle/Elasticsearch/Filter/Field/AncestorCodeFilter.php
```diff
@@ -70,6 +70,14 @@ public function addFieldFilter($field, $operator, $values, $locale = null, $chan
                 ];
                 $this->searchQueryBuilder->addMustNot($mustNotClause);
                 break;
+            case Operators::IS_EMPTY:
+                $clause = [
+                    'exists' => [
+                        'field' => self::ANCESTOR_CODES_ES_FIELD,
+                    ],
+                ];
+                $this->searchQueryBuilder->addFilter($clause);
```
tseho: I'm aware there is an error here. Working on it.

mmetayer: Shouldn't it be `$this->searchQueryBuilder->addMustNot($clause);` ?

tseho: yes, I saw it when re-reading my PR. Thanks for the catch @mmetayer

tseho:
@mmetayer I have a doubt about this:
https://github.com/akeneo/pim-community-dev/blob/c55496af2e735abab465b99ef318396560c3e768/src/Akeneo/Pim/Enrichment/Bundle/Elasticsearch/Filter/Field/ParentFilter.php#L62-L70
VS
https://github.com/akeneo/pim-community-dev/blob/c55496af2e735abab465b99ef318396560c3e768/src/Akeneo/Pim/Enrichment/Bundle/Elasticsearch/Filter/Attribute/TextFilter.php#L124-L136

Why does we have to add a clause on the family code there ?

tseho:
Added the missing tests.

mmetayer:
@tseho please also add integration tests in `tests/back/Pim/Enrichment/Integration/PQB/Filter/Field/AncestorFilterIntegration.php` or create `AncestorCodesFilterIntegration` (they're way more important than unit tests in this case :wink:)

mmetayer:
> @mmetayer I have a doubt about this:
> https://github.com/akeneo/pim-community-dev/blob/c55496af2e735abab465b99ef318396560c3e768/src/Akeneo/Pim/Enrichment/Bundle/Elasticsearch/Filter/Field/ParentFilter.php#L62-L70
> 
> 
> VS
> https://github.com/akeneo/pim-community-dev/blob/c55496af2e735abab465b99ef318396560c3e768/src/Akeneo/Pim/Enrichment/Bundle/Elasticsearch/Filter/Attribute/TextFilter.php#L124-L136
> 
> Why does we have to add a clause on the family code there ?

Because we don't want to fetch products without family in this case (which have a great probability of being matched by the query, see https://github.com/akeneo/pim-community-dev/pull/10214).

tseho:
I saw the related PR, I'm having trouble to know if I should do the same here or not.
What do you mean by `in this case` ?

mmetayer:
Sorry, I was not clear enough :slightly_smiling_face:,  I meant in the latter case, when you filter on a value. 
In your PR you don't need it, the `parent is empty` should return:
- root product models (which ALWAYS have a family)
- and simple (non variant) products (whether they have a family or not)

tseho:
thanks you @mmetayer for the help on this PR.

PR #11641, opened 2020-02-26
TIP-1392: Move 3.2 to 4.0 migration script in a dedicated folder

pierrejolly:

`OK`: Done / Validated
`-`: Not needed


ahocquard:
@jjanvier should we not purge the database to avoid warning? The directory and the database will nt be up to date. You can define several directories as in the documentation (https://symfony.com/doc/master/bundles/DoctrineMigrationsBundle/index.html), but it's defined by namespace :/

jjanvier:
@ahocquard 

> should we not purge the database to avoid warning

I don't understand what you mean. AFAIK, this will work.

ahocquard:
Seen IRL with JJ, it can be a problem because the filename is still in database. It will generate a ton of warnings for nothing.

jjanvier:
If we merge that, the nightly step `deploy_upgrade_pim3_environment` won't work. Closing.

PR #11643, opened 2020-02-26
PIM-9113: [SLA] Locale Specific attribute breaks product grid

anaisbaunelemaire:
Let's take this attribute with the following settings:
1- Attribute Date ("Test-AB" on the gif)
2- Value Per Locale = Yes
3- Read Only = Yes
4- Usable in grid = Yes
5- Locale Specific = Yes
6- Locale selected = "US"

When we activated this attribute as a filter in the product grid, and when the selected locale was not "US" (German, for example), the product grid was broken (infinite loader).

This PR adds a dedicated Elasticsearch filter validator helper which catches the Logic Exception when the attribute is locale specific but its locale is not available.

locale_specific

`OK`: Done / Validated
`-`: Not needed


src/Akeneo/Pim/Enrichment/Component/Product/Validator/AttributeValidatorHelper.php
```diff
@@ -86,7 +86,7 @@ public function validateLocale(AttributeInterface $attribute, $locale)
             );
         }
 
-        if ($attribute->isLocaleSpecific() && !in_array($locale, $attribute->getAvailableLocaleCodes())) {
+        if (!$attribute->isLocalizable() && $attribute->isLocaleSpecific() && !in_array($locale, $attribute->getAvailableLocaleCodes())) {
```
ahocquard: The fix is not correct I think.
The goal of the method is to validate that the locale associated to an attribute is correct.
Locale specific attribute is a setting that you can apply for both localizable and non localizable attributes.

You modified this part of validation on locale specific attributes to only occur on non localizable attributes.
It means that if you pass in the method a localizable attribute with a locale that is not part of the activated locales, it will not trigger an exception. It should.

What is the expected ehavior on the UI side? It's weird to have a locale specific filter displayed for a locale "fr_FR" if it's locale specific to "en_US". It means that you can research on it but actually... not.

To confirm with PM, but modifying the behavior of this function is not the good way to do it.

src/Akeneo/Pim/Enrichment/Component/Product/Validator/ElasticsearchFilterValidator.php
```diff
@@ -0,0 +1,53 @@
...
+        }
+
+        return true;
+    }
+
+    public function isChannelValidForAttribute(string $attributeCode, ?string $channelCode): bool
+    {
+        $attribute = $this->attributeRepository->findOneByIdentifier($attributeCode);
+
+        $this->attributeValidator->validateScope($attribute, $channelCode);
```
ahocquard: why don't you catch exception in that case?

ValentinMumble:
> I think this implementation is much better 👍 (maybe the `isValid` is not the best naming, as you don't return true/false but true or exception, my bad when I suggested that)

@ahocquard Yeah I renamed them to be more consistent, in the Channel case the validation logic is the same I guess so no Exception to be caught.

src/Akeneo/Pim/Enrichment/Component/Product/Validator/ElasticsearchFilterValidator.php
ValentinMumble: > I think this implementation is much better 👍 (maybe the `isValid` is not the best naming, as you don't return true/false but true or exception, my bad when I suggested that)

@ahocquard Yeah I renamed them to be more consistent, in the Channel case the validation logic is the same I guess so no Exception to be caught.

PR #11653, opened 2020-02-28
PIM-9110: mitigate again more the problem of deadlock by splitting the transcations

ahocquard:

We tried a lot of things:
- increasing the `sleep` drastically at each retry  (:red_circle: )
- increasing the number of retry drastically (20)  (worked with 10 thread, ? with 30 threads)
- to delete only unwanted row + INSERT (:red_circle: )
- SELECT id from rows and then only DELETE by id to know if it changes the locking strategy  (:red_circle: )
- DELETE index `locale_id` and `channel_id` in case of locking  (:red_circle: )
- use INSERT ON DUPLICATE KEY UPDATE (:heavy_check_mark: with 10 threads, :red_circle: with 30)
- use two different transactions(:heavy_check_mark: with 10 threads, :red_circle: with 30)

We chose the last one because the ON DUPLICATE KEY solution does not clean the table. If you deactivate a local, it would mean you still have the row in the completeness table... and when you reactivate it you have a wrong completeness... etc

The final solution would be to put in JSON. We still need to test though. It sacrifices consistency for concurrency. As it's not the completeness is not the source of truth, only a projection, it's not a big deal. It implies a migration though, etc. So, more work.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Akeneo/Pim/Enrichment/Bundle/Storage/Sql/Completeness/SqlSaveProductCompletenesses.php
+++ b/src/Akeneo/Pim/Enrichment/Bundle/Storage/Sql/Completeness/SqlSaveProductCompletenesses.php
@@ -47,69 +47,57 @@ final class SqlSaveProductCompletenesses implements SaveProductCompletenesses
      */
     public function saveAll(array $productCompletenessCollections): void
     {
-        $retry = 0;
-        $isError = true;
-        while (true === $isError) {
-            try {
-                $this->connection->transactional(function (Connection $connection) use ($productCompletenessCollections) {
-                    $productIds = array_unique(array_map(function (ProductCompletenessWithMissingAttributeCodesCollection $productCompletenessCollection) {
-                        return $productCompletenessCollection->productId();
-                    }, $productCompletenessCollections));
+        $deleteFunction = function () use ($productCompletenessCollections) {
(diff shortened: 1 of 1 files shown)


src/Akeneo/Pim/Enrichment/Bundle/Storage/Sql/Completeness/SqlSaveProductCompletenesses.php
```diff
@@ -138,4 +126,26 @@ private function channelIdsIndexedByChannelCodes(): array
...
+        $isError = true;
+        while (true === $isError) {
+            try {
+                $this->connection->transactional($function);
+
+                $isError = false;
+            } catch (DeadlockException $e) {
+                $retry += 1;
+
+                usleep(300000);
```
jjanvier: could you comment that please? The 2 usleep
(or better, extract them in a method with a proper name and phpdoc)

ahocquard: It was already explained in the phpdoc of the method.

ahocquard:
Close in favor of https://github.com/akeneo/pim-community-dev/pull/11661

PR #11654, opened 2020-02-28
RUL-39: add converters for dates and metrics

mmetayer:

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Akeneo/Pim/Enrichment/Component/Product/Updater/Converter/DateToStringDataConverter.php
+++ b/src/Akeneo/Pim/Enrichment/Component/Product/Updater/Converter/DateToStringDataConverter.php
@@ -1,0 +1,36 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Akeneo\Pim\Enrichment\Component\Product\Updater\Converter;
+
+use Akeneo\Pim\Enrichment\Component\Product\Model\ValueInterface;
+use Akeneo\Pim\Structure\Component\AttributeTypes;
+use Akeneo\Pim\Structure\Component\Model\AttributeInterface;
+use Webmozart\Assert\Assert;
+
+/**
+ * @copyright 2020 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+final class DateToStringDataConverter implements ValueDataConverter
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function supportsAttributes(AttributeInterface $sourceAttribute, AttributeInterface $targetAttribute): bool
+    {
+        return AttributeTypes::DATE === $sourceAttribute->getType() &&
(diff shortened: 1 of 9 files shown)


src/Akeneo/Pim/Enrichment/Component/Product/Updater/Converter/DateToStringDataConverter.php
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function convert(ValueInterface $sourceValue, AttributeInterface $targetAttribute)
+    {
+        Assert::isInstanceOf($sourceValue->getData(), \DateTimeInterface::class);
+
+        return $sourceValue->getData()->format('c');
```
nmarniesse: Are you sure for the `c` format? The users fill a date with day month and year. Not sure it's important to then to have the hour, minute second and the timezone. For the concatenate, the default format is `Y-m-d` (US date format) but they could specify the format in the settings.  
To check with Morgane.

tests/back/Pim/Enrichment/Specification/Component/Product/Updater/Converter/MetricToStringDataConverterSpec.php
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    function it_converts_a_metric_to_a_string()
+    {
+        $sourceValue = MetricValue::value(
+            'weight_in_grams',
+            new Metric('weight', 'GRAM', 23.55, 'KILOGRAM', .02355)
+        );
+
+        $this->convert($sourceValue, new Attribute())->shouldReturn('23.5500 GRAM');
```
nmarniesse: Number format must be validate with Morgane. If I remember well, we store `23.5500` in DB but users see `23,55` in UI. Seem to be more logic for users to have the second one when they copy (that's we did for concatenate)

src/Akeneo/Pim/Enrichment/Component/Product/Updater/Converter/DateToStringDataConverter.php
mmetayer: That's the expected format, as stated in https://akeneo.atlassian.net/browse/RUL-39

tests/back/Pim/Enrichment/Specification/Component/Product/Updater/Converter/MetricToStringDataConverterSpec.php
mmetayer: The point of this class is to copy attribute values, not transforming them. IMO the output should be as close as possible as the input.

mmetayer: Seen with @morganen: keep it like that for now, and we can iterate later if needed

PR #11656, opened 2020-02-28
PIM-9111: Fix the variant list when the channel does not support the selected locale

tseho:

How to reproduce:
- create a new channel with only one locale (eg: fr_FR)
- open a product model PEF
- switch to a channel with several locales
- switch to a locale not used in the new channel (eg: en_US)
- open the variant list (everything is ok)
- switch to the new channel (selected locale is still en_US)
- open the variant list 

There is a fatal js error in the dropdown because the dropdown tries to retrieve the completeness using the selected locale, see https://github.com/akeneo/pim-community-dev/blob/f2be06998b7477eccead072a6d016d774a9f52ef/src/Akeneo/Platform/Bundle/UIBundle/Resources/public/js/product/form/variant-navigation.js#L343

But this locale does not exists in this channel.

The proposed fix is to simply check if the selected locale exists for this channel, and if not, fallback to a completeness ratio of `0`.

### DIFF
--- a/src/Akeneo/Platform/Bundle/UIBundle/Resources/public/js/product/form/variant-navigation.js
+++ b/src/Akeneo/Platform/Bundle/UIBundle/Resources/public/js/product/form/variant-navigation.js
@@ -335,7 +335,7 @@ define(
 
                 if ('product' === entity.model_type) {
                     const channelCompletenesses = _.findWhere(entity.completeness, {channel: catalogScope});
-                    if (channelCompletenesses === undefined) {
+                    if (channelCompletenesses === undefined || channelCompletenesses.locales[catalogLocale] === undefined) {
                         return {
                             ratio: 0
                         };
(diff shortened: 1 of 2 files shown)


src/Akeneo/Platform/Bundle/UIBundle/Resources/public/js/product/form/variant-navigation.js
```diff
@@ -335,7 +335,8 @@ define(
 
                 if ('product' === entity.model_type) {
                     const channelCompletenesses = _.findWhere(entity.completeness, {channel: catalogScope});
-                    if (channelCompletenesses === undefined) {
+                    if (channelCompletenesses === undefined ||
```
ValentinMumble: Yoda style? :trollface:

tseho: you're right.

PR #11661, opened 2020-03-02
PIM-9110: lock tables to insert completeness if we still have deadlocks after 5 attempts

ahocquard:
When loading data with the API with 30 threads, we had some deadlocks despite the retry strategy, during insertion of completeness.
We tried a lot of things to mitigate this risk of deadlock during insertion of the completeness:

- increasing the `sleep` drastically at each retry  (:red_circle: )
- increasing the number of retry drastically (20)  (worked with 10 thread, ? with 30 threads)
- to delete only unwanted row + INSERT (:red_circle: )
- SELECT id from rows and then only DELETE by id to know if it changes the locking strategy  (:red_circle: )
- DELETE index `locale_id` and `channel_id` in case of locking  (:red_circle: )
- use INSERT ON DUPLICATE KEY UPDATE (:heavy_check_mark: with 10 threads, :red_circle: with 30)
- use two different transactions(:heavy_check_mark: with 10 threads, :red_circle: with 30)
- use lock tables (:heavy_check_mark: with 10 threads, :heavy_check_mark:  with 30)

I close the PR https://github.com/akeneo/pim-community-dev/pull/11653 in favor of this one.

When it fails despite 5 attempts, we lock the table. This way, we serialize the transaction. It guarantees it works whatever is the number of threads at the cost of the performance when the table is locked. It means that all concurrent requests are pending.

**Note**: we can't test this as we can't simulate deadlock, so I deactivated the `retry` function to test that the lock tables behaves the same way when running the integration test. The test was green.
 

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Akeneo/Pim/Enrichment/Bundle/Storage/Sql/Completeness/SqlSaveProductCompletenesses.php
+++ b/src/Akeneo/Pim/Enrichment/Bundle/Storage/Sql/Completeness/SqlSaveProductCompletenesses.php
@@ -9,6 +9,7 @@ use Akeneo\Pim\Enrichment\Component\Product\Query\SaveProductCompletenesses;
 use Doctrine\DBAL\Connection;
 use Doctrine\DBAL\Exception\DeadlockException;
 use Doctrine\DBAL\ParameterType;
+use Psr\Log\LoggerInterface;
 
 /**
  * @author    Mathias METAYER <mathias.metayer@akeneo.com>
@@ -20,9 +21,13 @@ final class SqlSaveProductCompletenesses implements SaveProductCompletenesses
     /** @var Connection */
     private $connection;
 
-    public function __construct(Connection $connection)
+    /** @var LoggerInterface */
+    private $logger;
+
+    public function __construct(Connection $connection, LoggerInterface $logger)
     {
         $this->connection = $connection;
+        $this->logger = $logger;
     }
 
     /**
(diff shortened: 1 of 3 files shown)


src/Akeneo/Pim/Enrichment/Bundle/Storage/Sql/Completeness/SqlSaveProductCompletenesses.php
```diff
@@ -138,4 +131,48 @@ private function channelIdsIndexedByChannelCodes(): array
...
+
+        $formerAutocommitValue = (int) $value['@@autocommit'];
+        try {
+            $this->connection->executeQuery('SET autocommit=0');
+            $this->connection->executeQuery('LOCK TABLES pim_catalog_completeness WRITE');
+            $function();
+            $this->connection->executeQuery('COMMIT');
+            $this->connection->executeQuery('UNLOCK TABLES');
+        } finally {
+            $this->connection->executeQuery(sprintf('SET autocommit=%s', $formerAutocommitValue));
```
ahocquard: Changing a parameter does not support prepared statement (I could not use `:autocommit`). So I checked the value before to be sure it's either `1` or `0`.

src/Akeneo/Pim/Enrichment/Bundle/Storage/Sql/Completeness/SqlSaveProductCompletenesses.php
```diff
@@ -40,75 +40,68 @@ public function save(ProductCompletenessWithMissingAttributeCodesCollection $com
...
-                        $numberCompletenessRow += count($productCompletenessCollection);
-                    }
-                    $placeholders = implode(',', array_fill(0, $numberCompletenessRow, '(?, ?, ?, ?, ?)'));
-
-                    if (empty($placeholders)) {
-                        return;
-                    }
-
-                    $insert = <<<SQL
+        // we get it before to avoid to lock the tables "pim_catalog_locale" and "pim_catalog_channel"
```
jjanvier: "we get it"
what is "it"?

src/Akeneo/Pim/Enrichment/Bundle/Storage/Sql/Completeness/SqlSaveProductCompletenesses.php
```diff
@@ -138,4 +131,48 @@ private function channelIdsIndexedByChannelCodes(): array
...
+        $value = $this->connection->executeQuery('SELECT @@autocommit')->fetch();
+        if (!isset($value['@@autocommit']) && ((int) $value['@@autocommit'] !== 1 || (int) $value['@@autocommit'] !== 0)) {
+            throw new \LogicException('Error when getting autocommit parameter from Mysql.');
+        }
+
+        $formerAutocommitValue = (int) $value['@@autocommit'];
+        try {
+            $this->connection->executeQuery('SET autocommit=0');
+            $this->connection->executeQuery('LOCK TABLES pim_catalog_completeness WRITE');
+            $function();
```
jjanvier: no catch?

src/Akeneo/Pim/Enrichment/Bundle/Storage/Sql/Completeness/SqlSaveProductCompletenesses.php
jjanvier: %d ?

src/Akeneo/Pim/Enrichment/Bundle/Storage/Sql/Completeness/SqlSaveProductCompletenesses.php
jjanvier: we should at least log no?

src/Akeneo/Pim/Enrichment/Bundle/Storage/Sql/Completeness/SqlSaveProductCompletenesses.php
```diff
@@ -138,4 +131,48 @@ private function channelIdsIndexedByChannelCodes(): array
...
+    {
+        $retry = 0;
+        $isError = true;
+        while (true === $isError) {
+            try {
+                $this->connection->transactional($function);
+
+                $isError = false;
+            } catch (DeadlockException $e) {
+                $retry += 1;
```
jjanvier: shouldn't we log that?

src/Akeneo/Pim/Enrichment/Bundle/Storage/Sql/Completeness/SqlSaveProductCompletenesses.php
ahocquard: as `%s` in sprintf but for decimal

src/Akeneo/Pim/Enrichment/Bundle/Storage/Sql/Completeness/SqlSaveProductCompletenesses.php
ahocquard: no catch because in this case, that's the ultimate attempt. If we have an error, we should throw an exception.

src/Akeneo/Pim/Enrichment/Bundle/Storage/Sql/Completeness/SqlSaveProductCompletenesses.php
ahocquard: good idea, I will do that in another PR

ahocquard:
@nmarniesse `'INSERT ON DUPLICATE KEY UPDATE'` has two drawbacks:
- we still have some fails with 30 threads, so it does not fix the problem
- it does not delete the useless rows, so our completeness is dirty. It means that you have to do a lot of JOIN on activated locale and channel as soon as you access the completeness table. In other term, the table is not clean.
- the documented solution is to do table lock as a last attempt

src/Akeneo/Pim/Enrichment/Bundle/Storage/Sql/Completeness/SqlSaveProductCompletenesses.php
jjanvier: maybe you should comment that

src/Akeneo/Pim/Enrichment/Bundle/Storage/Sql/Completeness/SqlSaveProductCompletenesses.php
ahocquard: I added logs and tested it: far better (warning as it's the level where we trigger it in production AFAIK).

PR #11664, opened 2020-03-02
[akeneo-design-system]: Moving to babel loader

nicomaligne:
POC in progress, here we are changing typescript loader to babel loader with typescript plugins.

`OK`: Done / Validated
`-`: Not needed


akeneo-design-system/tsconfig.json
```diff
@@ -26,6 +25,4 @@
         "suppressImplicitAnyIndexErrors": true,
         "target": "es5"
     },
-    "include": ["src/**/*"],
-    "exclude": ["node_modules"]
```
pchasle: I think that we still need the include/exclude for development / IDE

akeneo-design-system/tsconfig.json
```diff
@@ -17,7 +14,9 @@
         "noImplicitThis": true,
         "noUnusedLocals": true,
         "noUnusedParameters": true,
+        "outDir": "lib",
         "rootDir": "src",
+        "skipLibCheck": true,
```
pchasle: I wouldn't use the flag unless we effectively have errors in some malformed libs/dt.ts and/or this improve significantly the build time.

nicomaligne: Yep I need to investigate on this, I've got error when I build on the pim project but not on a test one. Need to check dependencies.

akeneo-design-system/tsconfig.json
nicomaligne: I think with the rootDir do the trick.

akeneo-design-system/tsconfig.json
pchasle: Keep it this way for now then ;)

PR #11665, opened 2020-03-02
TIP-1394: use env var for mailer_from_address

jjanvier:
Needed for our SaaS instances.


src/Akeneo/UserManagement/Bundle/DependencyInjection/Configuration.php
```diff
@@ -30,7 +30,6 @@ public function getConfigTreeBuilder()
                     ->addDefaultsIfNotSet()
                     ->children()
                         ->scalarNode('address')
-                            ->defaultValue('no-reply@example.com')
```
jjanvier: default value is now given by the env var here https://github.com/akeneo/pim-community-dev/blob/master/config/packages/akeneo_pim_user.yml#L3

PR #11669, opened 2020-03-03
[akeneo-design-system]: Add styled components tests

nicomaligne:
Adding some styled components / css oriented tests.
The jest-styled-components encounters some typescript [issues](https://github.com/styled-components/jest-styled-components/issues/264).
A file extend-jest.d.ts has been added to extend properly the jest types, with a new function "toHaveStyleRule".
Doc about jest-styled-components is [here](https://github.com/styled-components/jest-styled-components).
Also for now I encounter a problem with the setupTests.ts file to avoid importing "import 'jest-styled-components';" in all tests file. It will be fixed in an another pr.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/akeneo-design-system/src/components/Buttons/CoreButton/CoreButton.test.tsx
+++ b/akeneo-design-system/src/components/Buttons/CoreButton/CoreButton.test.tsx
@@ -1,75 +1,104 @@
 import React from 'react';
-import { CoreButton } from "./CoreButton";
-import {
-    render, fireEvent
-} from "@testing-library/react";
+import 'jest-styled-components';
 
-describe("testing button", () => {
-    test("should display a button with the given children", () => {
+import { CoreButton } from './CoreButton';
+import { render, fireEvent } from '@testing-library/react';
+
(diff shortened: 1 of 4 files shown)


akeneo-design-system/src/components/Buttons/CoreButton/CoreButton.test.tsx
```diff
@@ -1,75 +1,104 @@
...
+            .c0 {
+              border-radius: 16px;
+              cursor: pointer;
+              font-size: 13px;
+              font-weight: 400;
+              height: 32px;
+              line-height: 32px;
+              padding: 0 15px;
+              text-transform: uppercase;
+            }
```
pchasle: Does this mean we need to duplicate all the CSS inside the tests?

nicomaligne: For atomic components (and as start) I think it could be a good idea. It may be too much for molecule / compound component

PR #11677, opened 2020-03-04
Remove nested set pattern in product models

ahocquard:
Fix performance issues when deleting product model and also deadlock when inserting product model with parallelism through the PAI

We have two issues:
- performance problem when deleting a product model (and I suppose when inserting product models also...). An index is missing on `left` `right` `root`.
- deadlock when inserting product model

All of this is due to the nested set pattern we introduced for the product models. It is never used. Let's drop that, this dead code is not really dead code as there are bugs due to that.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/tests/back/Pim/Enrichment/Specification/Component/Product/ProductModel/ImageAsLabelSpec.php
+++ b/tests/back/Pim/Enrichment/Specification/Component/Product/ProductModel/ImageAsLabelSpec.php
@@ -54,7 +54,7 @@ class ImageAsLabelSpec extends ObjectBehavior
         $familyVariant->getVariantAttributeSets()->willReturn($attributeSets);
         $productModel->getFamily()->willReturn($family);
         $productModel->getFamilyVariant()->willReturn($familyVariant);
-        $productModel->getLevel()->willReturn(1);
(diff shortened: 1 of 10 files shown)


tests/back/Pim/Enrichment/Specification/Component/Product/ProductModel/ImageAsLabelSpec.php
```diff
@@ -124,14 +124,14 @@ function it_gets_the_attribute_as_image_value_of_a_product_model_coming_from_a_p
         $attributeSetTwo->getAttributes()->willReturn($attributeCollectionTwo);
         $attributeSetTwo->getLevel()->willReturn(2);
 
-        $attributeCollectionOne->contains($attributeAsImage)->willReturn(false);
-        $attributeCollectionTwo->contains($attributeAsImage)->willReturn(true);
+        $attributeCollectionOne->contains($attributeAsImage)->willReturn(true);
+        $attributeCollectionTwo->contains($attributeAsImage)->willReturn(false);
```
ahocquard: the test was wrong...

PR #11701, opened 2020-03-06
[akeneo-design-system]: Add ally addon to storybook

nicomaligne:
Add ally addon to storybook
https://www.npmjs.com/package/@storybook/addon-a11y
I think this could be cool.
Even if we are not aiming to be fully accessible, the plug in could help to apply some very simple practice.
It's not a PR / compilation blocker.

| Added Specs                       |-

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/akeneo-design-system/.storybook/main.js
+++ b/akeneo-design-system/.storybook/main.js
@@ -2,7 +2,7 @@ const path = require('path');
 
 module.exports = {
     stories: ['../src/**/*.stories.mdx'],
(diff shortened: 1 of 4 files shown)


pchasle:
Could be nice, but I'm wondering - as accessibility is not a priority (for now) - if we are not going to have a lot of warning on all the components - background noise - and ultimately ignore them.
I think it would be nice to clarify our stance with UX (and PM) on this first.

nicomaligne:
If i'm not mistaken, it will just add a tab with accessibility warning, if you don't open the tabs you will not see the warning. So not very intrusive. Also a lots of accessibility rules are juste proper html markup (important on complex component / forn), so it should be mandatory.

PR #11704, opened 2020-03-08
TIP-1395: Adds a webpack option to run typescript without type checking

tseho:

This is an attempt to improve webpack build times.

The **webpack build time is increasing** and the main culprit is `typescript` with its `ts-loader`, 
mostly on the **type-checking** part.

For a lot of daily tasks, when we are building the PIM, we don't need to run the type-checking. It should only be executed when needed and on the CI, as we already do for the back with `make lint-back`.

This PR introduces a new option for our webpack config: `--strict`. When enabled, webpack will run as before, with the type-checking.  When omitted, typescript will only transpile from `ts` to `js`.

#### What are the performances changes (CE) ? (TL;DR: -30%)

For the benchmark, I've used a small script to cleanup the cache between builds, to check the performances on a "cold" webpack:
```
#!/bin/bash

rm -rf node_modules/.cache public/cache
yarn run webpack
```

Webpack **before**:
```
$ perf stat -r 10 -B ./perf.sh

 Performance counter stats for './perf.sh' (10 runs):
      32,897314634 seconds time elapsed                                          ( +-  1,35% )
```

Webpack **after**, without `--strict`:
```
perf stat -r 10 -B ./perf.sh

 Performance counter stats for './perf.sh' (10 runs):
      22,984510380 seconds time elapsed                                          ( +-  2,01% )
```

#### What about the CI ?

For the CI, I introduced a new make command: `make javascript-dev-strict` and a new step `Front type checking` inside `test_front_static_acceptance_and_integration`.

#### And if I want the type-checking ?

You can keep the previous behavior by running `yarn run webpack-dev --strict`.

Also, `yarn run webpack-watch` is unaffected, `--strict` is enabled by default when using watch, as we usually want type-checking when working actively on the frontend code.

#### Overall benefits

This commands are **slightly faster**:

- `make pim-dev`
- `make pim-behat`
- `make javascript-dev`
- `make front`
- `yarn run webpack`
- `yarn run webpack-dev`
- `yarn run webpack-test`

And the CI differences (approximatly, I don't want to run a lot of jobs for this):

| step | before | after | diff |
|---|---:|---:|---:|
| build_dev | 4m 4s | 3m 51s | **-13s** |
| test_front_static_acceptance_and_integration | 3m 8s | 3m 46s | **+38s** | 

Note: The differences are insignificant regarding the CI for the CE but are more valuable in EE.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -194,6 +194,9 @@ jobs:
               name: Change owner on project dir (default user = circleci (1001) and docker needs uid 1000)
               command: sudo chown -R 1000:1000 ../project ~/.cache/yarn
         - run:
+              name: Front type checking
+              command: make javascript-dev-strict
+        - run:
               name: Front linter
               command: make lint-front
         - run:
--- a/Makefile
+++ b/Makefile
@@ -45,7 +45,7 @@ javascript-dev:
 	$(YARN_RUN) run webpack-dev
 
 .PHONY: javascript-dev-strict
-javascript-dev:
+javascript-dev-strict:
 	$(DOCKER_COMPOSE) run -u www-data --rm php rm -rf public/dist
 	$(YARN_RUN) run webpack-dev --strict
 
--- a/Makefile
+++ b/Makefile
@@ -44,6 +44,11 @@ javascript-dev:
 	$(DOCKER_COMPOSE) run -u www-data --rm php rm -rf public/dist
 	$(YARN_RUN) run webpack-dev
 
+.PHONY: javascript-dev-strict
+javascript-dev:
+	$(DOCKER_COMPOSE) run -u www-data --rm php rm -rf public/dist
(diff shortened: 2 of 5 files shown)


.circleci/config.yml
```diff
@@ -193,6 +193,9 @@ jobs:
         - run:
               name: Change owner on project dir (default user = circleci (1001) and docker needs uid 1000)
               command: sudo chown -R 1000:1000 ../project ~/.cache/yarn
+        - run:
+              name: Front type checking
+              command: make javascript-dev-strict
```
pchasle: Maybe we should just run a `yarn tsc -p tsconfig.json --noEmit` in the `Front linter` step?

tseho: does this command work on yours ?

pchasle: I have 2 small import errors on CE that can be easily fixed. Didn't try it on EE.

PR #11715, opened 2020-03-10
API-1016: Purge audit data

willy-ahva:

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Akeneo/Connectivity/Connection/back/Domain/Audit/Persistence/Query/PurgeAuditProductQuery.php
+++ b/src/Akeneo/Connectivity/Connection/back/Domain/Audit/Persistence/Query/PurgeAuditProductQuery.php
@@ -1,0 +1,21 @@
+<?php
+declare(strict_types=1);
+
+namespace Akeneo\Connectivity\Connection\Domain\Audit\Persistence\Query;
+
+/**
+ * @author    Willy Mesnage <willy.mesnage@akeneo.com>
+ * @copyright 2020 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+interface PurgeAuditProductQuery
+{
+    /**
+     * Returns the number of rows deleted.
+     *
+     * @param \DateTimeImmutable $before Delete rows that have been saved strictly before this datetime.
+     *
+     * @return int
+     */
+    public function execute(\DateTimeImmutable $before): int;
+}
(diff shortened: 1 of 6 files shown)


src/Akeneo/Connectivity/Connection/back/Infrastructure/Persistence/Dbal/Query/DbalPurgeAuditProductQuery.php
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * @author    Willy Mesnage <willy.mesnage@akeneo.com>
+ * @copyright 2020 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+class DbalPurgeAuditProductQuery implements PurgeAuditProductQuery
+{
+    /** @var Connection */
+    private $connection;
```
pierrejolly: Since we are in a context where `connection` mean `api connection` I think we should named this variable `dbalConnection`

src/Akeneo/Connectivity/Connection/back/Domain/Audit/Persistence/Query/PurgeAuditProductQuery.php
```diff
@@ -0,0 +1,21 @@
...
+declare(strict_types=1);
+
+namespace Akeneo\Connectivity\Connection\Domain\Audit\Persistence\Query;
+
+/**
+ * @author    Willy Mesnage <willy.mesnage@akeneo.com>
+ * @copyright 2020 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+interface PurgeAuditProductQuery
```
pchasle: I'm not confident that the action of deleting Event Counts has his place in the Domain (or it would rather be a Command than a Query).
Calling the repository to delete all the Event Counts before X DateTime would be sufficient. What do you think?

willy-ahva: In a repository, you must not have business, but only the crud. The purge is representative of a business need. That's why it is represented in the domain layer. Moreover, maybe tomorrow we'll need during this purge to delete another entity which is in another table, or the table we'll be split in two, or move in ES, if the query is in the repository it's not consistent. That's why we must not put this query in the repository.

PR #11719, opened 2020-03-10
PIM-9137: do not export headers if no data to export and attribute used as filters

rouliane:
This PR fixes a regression in 3.0 in a very specific use case : if a product export has no data and one or several attributes are used as filters, the expected behavior is to NOT export the headers (so the file should be empty).

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/tests/back/Pim/Enrichment/Specification/Component/Product/Connector/Writer/File/Csv/ProductWriterSpec.php
+++ b/tests/back/Pim/Enrichment/Specification/Component/Product/Connector/Writer/File/Csv/ProductWriterSpec.php
@@ -500,4 +500,65 @@ class ProductWriterSpec extends ObjectBehavior
         ]);
         $this->flush();
     }
+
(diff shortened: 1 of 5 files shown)


tests/back/Pim/Enrichment/Specification/Component/Product/Connector/Writer/File/Csv/ProductWriterSpec.php
```diff
@@ -500,4 +500,66 @@ function it_writes_the_csv_file_with_headers_and_selected_attributes(
...
+
+        $bufferFactory->create()->willReturn($flatRowBuffer);
+        $flusher->flush(
+            $flatRowBuffer,
+            Argument::type('array'),
+            Argument::type('string'),
+            -1
+        )->willReturn([
+            $this->directory . 'CSV_Product_export_product1.csv',
+            $this->directory . 'CSV_Product_export_product2.csv'
```
mmetayer: there shouldn't be several files if there is no content :stuck_out_tongue_winking_eye:

PR #11720, opened 2020-03-10
[akeneo-design-system]: Add main theme

nicomaligne:

Add main theme

| Added Specs                       |  -

`OK`: Done / Validated
`-`: Not needed


tseho:
Why is it called "main theme" ?

nicomaligne:
@tseho The idea with the theme is to have one main / core theme and when it will be needed some new specific theme. For example some tabs selection and filter text have different color between pim / onboarder.
So in a near future we could have a onBoarder theme importing some value from the mainTheme and exposing a 'tabs-selected-color' variable in the dedicated tabs components.
This will helps us to have multiple theme in the styled component theme provider.
mainTheme may not be the ideal name (maybe coreTheme ?), open to any suggestion about the naming ;)

tseho:
@nicomaligne thanks.
Yeah, the naming seems off to me, I'm aware of the differences between the
PIM and the Shared Catalog, for example.
Maybe a naming like PIMTheme would be a better fit.

Le mer. 11 mars 2020 à 13:05, nicomaligne <notifications@github.com> a
écrit :

nicomaligne:
@tseho The goal was to have some basic / core value that could be shared to many theme like OnBoaderTheme / SharedCatalogTheme / PIMTheme ? (not sure for this one :p).
It's why the name is generic. It's more an AkeneoTheme, in which other theme pick the attributes values they need. But the PIM is so huge and the main product, so yes PIM theme could be an idea. You can suggest a poll on slack for a new naming. I will make the change.

PR #11744, opened 2020-03-13
MET-105: Fix CreateMeasurementFamily validation

tseho:

2 issues fixed:
- The constraints `NotBlank` and `Length(min=1)` on the code are doing the same thing, when the field is empty, 2 validation errors are returned. It was useless.
- the `Create` endpoint should not use the `SaveMeasurementFamilyCommand` because it's an upsert. I added a `CreateMeasurementFamilyCommand` with proper validation and the new `CodeMustBeUnique`.

### DIFF
--- a/src/Akeneo/Tool/Bundle/MeasureBundle/Application/CreateMeasurementFamily/CreateMeasurementFamilyCommand.php
+++ b/src/Akeneo/Tool/Bundle/MeasureBundle/Application/CreateMeasurementFamily/CreateMeasurementFamilyCommand.php
@@ -1,0 +1,24 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Akeneo\Tool\Bundle\MeasureBundle\Application\CreateMeasurementFamily;
+
+/**
+ * @copyright 2020 Akeneo SAS (https://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php Open Software License (OSL 3.0)
+ */
+class CreateMeasurementFamilyCommand
+{
+    /** @var string */
+    public $code;
+
+    /** @var array */
+    public $labels;
+
+    /** @var string */
+    public $standardUnitCode;
+
+    /** @var array */
+    public $units = [];
+}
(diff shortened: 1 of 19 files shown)


src/Akeneo/Tool/Bundle/MeasureBundle/Application/CreateMeasurementFamily/CreateMeasurementFamilyCommand.php
```diff
@@ -0,0 +1,24 @@
...
+    public $code;
+
+    /** @var array */
+    public $labels;
+
+    /** @var string */
+    public $standardUnitCode;
+
+    /** @var array */
+    public $units = [];
```
SamirBoulil: Are you sure you need the whole units ?
Maybe standardUnitCode, symbol (the input we have in the create form)
then directly set the convert_from_standard to Multiply by one in the handler ?

SamirBoulil: I mean there is no way, there are multiple units in this variable right ?
If it's too difficult to do for now, we can leave it as is.

src/Akeneo/Tool/Bundle/MeasureBundle/Resources/translations/validators.en_US.yml
```diff
@@ -6,5 +6,7 @@ pim_measurements:
                 limit_reached: 'You cannot create the "%measurement_family_code%" measurement family  because you have reached the limit of %limit% measurement families'
         measurement_family:
             standard_unit_code_should_exist: 'The "%standard_unit_code%" standard unit code does not exist in the list of units for the measurement family.'
+            code:
+                must_be_unique: 'This measurement family code is already used.'
```
SamirBoulil: The measurement family already exists
```suggestion
                must_be_unique: 'This measurement family code already exists.'
```

src/Akeneo/Tool/Bundle/MeasureBundle/Application/CreateMeasurementFamily/CreateMeasurementFamilyCommand.php
tseho: I need only one unit in this array.
Rewriting this means I have to diverge more regarding the validation, it's not complicated, it's just more work (and I don't mind).
And you are right, this command **should** reflect its real job, not a copy-pasted job.

At the same time, the form, in the frontend, does create an object respecting the MeasurementFamily model, not a new `MeasurementFamilyCreation`. This would mean we would need changes there too.

SamirBoulil: I agree let's keep it this way

PR #11754, opened 2020-03-16
PIM-9140: Translate error message from backend in dialogs

mmetayer:

Translate error messages coming from backend in delete dialogs

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Akeneo/Platform/Bundle/UIBundle/Resources/public/js/pim-init.js
+++ b/src/Akeneo/Platform/Bundle/UIBundle/Resources/public/js/pim-init.js
@@ -2,6 +2,7 @@ define(
     [
     'jquery',
     'underscore',
+    'oro/translator',
     'backbone',
     'oro/mediator',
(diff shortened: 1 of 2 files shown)


src/Akeneo/Platform/Bundle/UIBundle/Resources/public/js/pim-init.js
```diff
@@ -161,7 +163,7 @@ define(
                                 messenger.notify(
                                     'error',
                                     (xhr.responseJSON && xhr.responseJSON.message) ?
-                                        xhr.responseJSON.message :
+                                        __(xhr.responseJSON.message) :
```
pchasle: I guess that if the message is plain English, the translator would not find the 'key' and return the message unmodified...

mmetayer: yes, exactly

PR #11762, opened 2020-03-16
Fix changelog files

jmleroux:
CHANGELOG-5.0 was a bad naming, we still don't know if there will be a 5.0

I merged the two in one CHANGELOG file


CHANGELOG.md
```diff
@@ -1,6 +1,84 @@
+# 5.0.0
```
nmarniesse: `5.0.0` does not exist ;) To remove IMO

PR #11771, opened 2020-03-18
PIM-9152: Handle old versioning date format

jmleroux:
Fixes issues when products saved with 'date' are modified to 'datetime'. The PIM recognizes those fields as being updated (even if the date remains the same) and creates versions in the product history.

### DIFF
--- a/src/Akeneo/Tool/Bundle/VersioningBundle/Builder/VersionBuilder.php
+++ b/src/Akeneo/Tool/Bundle/VersioningBundle/Builder/VersionBuilder.php
@@ -172,8 +172,52 @@ class VersionBuilder
         return array_filter(
             $changeset,
             function ($item) {
-                return $item['old'] !== $item['new'];
+                return $this->hasValueChanged($item['old'], $item['new']);
             }
         );
+    }
+
+    private function hasValueChanged($old, $new): bool
+    {
+        if (null !== $hasChanged = $this->hasLegacyDateChanged($old, $new)) {
+            return $hasChanged;
+        }
+
+        return $old !== $new;
+    }
+
+    /**
+     * We need to handle date comparison for old versioning format 'Y-m-d' in place of the new 'Y-m-d\TH:i:sP'.
+     *
+     * To determine that we are comparing date from the old versioning format:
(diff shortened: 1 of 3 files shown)


src/Akeneo/Tool/Bundle/VersioningBundle/Builder/VersionBuilder.php
```diff
@@ -172,8 +172,51 @@ protected function filterChangeset(array $changeset)
...
+        if (false === $oldDateTime) {
+            return null;
+        }
+
+        $newDateTime = \DateTimeImmutable::createFromFormat(\DateTimeInterface::ATOM, $new);
+        if (false === $newDateTime) {
+            return null;
+        }
+
+        return $oldDateTime->format('u') !== $newDateTime->format('u');
```
mmetayer: Nope, you're only comparing the microseconds part here. You probably meant 'U' :wink:

src/Akeneo/Tool/Bundle/VersioningBundle/spec/Builder/VersionBuilderSpec.php
```diff
@@ -77,4 +77,44 @@ function it_compare_versions(Version $pending, Version $previousPending)
...
+            'name' => 'foo',
+            'date' => '2020-01-01T00:00:00+00:00',
+            'date_with_old_format' => '2020-01-01'
+        ]);
+
+        $version->setVersion(2)->willReturn($version);
+        $version->setSnapshot([
+            'name' => 'bar',
+            'date' => '2020-01-01T00:00:00+00:00',
+            'date_with_old_format' => '2020-01-01T00:00:00+00:00'
```
mmetayer: you should add a test with a different timezone (e.g '2020-01-01T12:00:00+12:00'), and also a test which proves that 2 different dates are well handled by your solution

src/Akeneo/Tool/Bundle/VersioningBundle/Builder/VersionBuilder.php
pchasle: Thanks, this was a bug :/

PR #11774, opened 2020-03-18
Pim 9147 create completeness clean command

greg-akeneo:
**Description**
Since the foreign key between completeness and product has been removed, the delete of products does not cascade to completeness. This PR introduces the job that removes the completeness records of products that do not exist anymore.

See [JIRA ticket](https://akeneo.atlassian.net/browse/PIM-9147)

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Akeneo/Pim/Enrichment/Bundle/Command/CleanCompletenessForNonExistingProducts.php
+++ b/src/Akeneo/Pim/Enrichment/Bundle/Command/CleanCompletenessForNonExistingProducts.php
@@ -20,7 +20,7 @@ use Symfony\Component\Console\Output\OutputInterface;
  */
 
 class CleanCompletenessForNonExistingProducts extends Command
-{ 
+{
     protected static $defaultName = 'pim:completeness:clean';
 
     /** @var Connection */
@@ -56,4 +56,3 @@ SQL;
         return 0;
     }
 }
-
--- a/src/Akeneo/Pim/Enrichment/Bundle/Command/CleanCompletenessForNonExistingProducts.php
+++ b/src/Akeneo/Pim/Enrichment/Bundle/Command/CleanCompletenessForNonExistingProducts.php
@@ -1,0 +1,59 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Akeneo\Pim\Enrichment\Bundle\Command;
+
+use Doctrine\DBAL\Connection;
+use Symfony\Component\Console\Command\Command;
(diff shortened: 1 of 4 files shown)


ahocquard:
Why is in a job and not in a dedicated `remover` for the products? It would be much more clean.

src/Akeneo/Pim/Enrichment/Bundle/Command/CleanCompletenessForNonExistingProducts.php
```diff
@@ -0,0 +1,58 @@
...
+        $this->setDescription('Clean orphan completeness entries.');
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        $sql = <<<SQL
+DELETE FROM pim_catalog_completeness AS pcc
```
ahocquard: I think it will lock all the rows of this table...

greg-akeneo: true but this is not a command that is intended to be launched in parallel.

BitOne: It could be launched at the same time as a massive API import. So this may be lock the table too long for the API calls.

We discussed with Greg about using a first SELECT to find out what are the rows of this table that should be removed. Then use the resulting ids to remove the rows. This would avoid locking other rows (because by primary key), and we could even batch them.

The remover() part is good suggestion as well. This would avoid adding another job.

BitOne:
We are missing the deployement configuration in this PR, but this needs to be done in the EE.

I would try to do a PR with a specific product remover.

src/Akeneo/Pim/Enrichment/Bundle/Command/CleanCompletenessForNonExistingProducts.php
ahocquard: Yes, this is the same thing that I proposed to calculate the completeness in 1.7 and that Pipou implemented (SELECT + DELETE).

Best solution is to handle it in an event I think: as for the indexation actually, which is also a projection. 
(https://github.com/akeneo/pim-community-dev/blob/master/src/Akeneo/Pim/Enrichment/Bundle/EventSubscriber/Product/OnDelete/ComputeProductsAndAncestorsSubscriber.php)

PR #11785, opened 2020-03-20
MET-124: Split models

ValentinMumble:

This PR:
- Splits model in dedicated files
- Add a `autofocus` prop in TextField component
- Disables the add operation button instead of hiding it

`OK`: Done / Validated
`-`: Not needed


tseho:
@SamirBoulil we should, at least for everything that can be easily unit tested.

PR #11791, opened 2020-03-20
RUL-89: add query to fetch all association type codes

nmarniesse:

Add a query to fetch all association type codes.
- add a SQL query that returns a generator (to prevent memory problem)
- add a cache query. The cache is automatically disabled if there is too many elements to prevent memory problem too.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Akeneo/Pim/Structure/Bundle/Query/PublicApi/AssociationType/Cache/CachedGetAssociationTypeCodes.php
+++ b/src/Akeneo/Pim/Structure/Bundle/Query/PublicApi/AssociationType/Cache/CachedGetAssociationTypeCodes.php
@@ -8,7 +8,7 @@ use Akeneo\Pim\Structure\Component\Query\PublicApi\AssociationType\GetAssociatio
 
 /**
  * Cache for association type codes fetching. If there is too much codes to set in cache, no cache is done
- * in order to prevent memory problem (see static::CACHED_ITEM_LIMIT).
+ * in order to prevent memory problem (see CACHED_ITEM_LIMIT).
  *
  * @author    Nicolas Marniesse <nicolas.marniesse@akeneo.com>
  * @copyright 2020 Akeneo SAS (http://www.akeneo.com)
--- a/src/Akeneo/Pim/Structure/Bundle/Query/PublicApi/AssociationType/Cache/CachedGetAssociationTypeCodes.php
+++ b/src/Akeneo/Pim/Structure/Bundle/Query/PublicApi/AssociationType/Cache/CachedGetAssociationTypeCodes.php
@@ -1,0 +1,61 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Akeneo\Pim\Structure\Bundle\Query\PublicApi\AssociationType\Cache;
+
+use Akeneo\Pim\Structure\Component\Query\PublicApi\AssociationType\GetAssociationTypeCodes;
+
+/**
+ * Cache for association type codes fetching. If there is too much codes to set in cache, no cache is done
+ * in order to prevent memory problem (see static::CACHED_ITEM_LIMIT).
+ *
+ * @author    Nicolas Marniesse <nicolas.marniesse@akeneo.com>
+ * @copyright 2020 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+final class CachedGetAssociationTypeCodes implements GetAssociationTypeCodes
+{
+    const CACHED_ITEM_LIMIT = 10000;
+
+    /** @var GetAssociationTypeCodes */
+    private $getAssociationTypeCodes;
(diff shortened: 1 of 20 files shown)


src/Akeneo/Pim/Structure/Bundle/Query/PublicApi/AssociationType/Sql/SqlGetAssociationTypeCodes.php
```diff
@@ -0,0 +1,57 @@
...
+    {
+        $this->connection = $connection;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function findAll(): \Iterator
+    {
+        $limitByQuery = 1000;
```
pierallard: pageSize

src/Akeneo/Pim/Structure/Bundle/Query/PublicApi/AssociationType/Cache/CachedGetAssociationTypeCodes.php
```diff
@@ -0,0 +1,61 @@
...
+ * Cache for association type codes fetching. If there is too much codes to set in cache, no cache is done
+ * in order to prevent memory problem (see static::CACHED_ITEM_LIMIT).
+ *
+ * @author    Nicolas Marniesse <nicolas.marniesse@akeneo.com>
+ * @copyright 2020 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+final class CachedGetAssociationTypeCodes implements GetAssociationTypeCodes
+{
+    const CACHED_ITEM_LIMIT = 10000;
```
pierallard: Is really 10k strings too much for the memory? (Real question)

nmarniesse: It depends of the strings lengths. See [strlen](https://www.php.net/manual/en/function.strlen.php). I think it's a safeguard if someone tests the limits of the PIM or want it to crash. We never had more than 50 association types IMO.

src/Akeneo/Pim/Structure/Component/Query/PublicApi/AssociationType/GetAssociationTypeCodes.php
```diff
@@ -0,0 +1,18 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Akeneo\Pim\Structure\Component\Query\PublicApi\AssociationType;
```
mmetayer: Associations are part of the `Enrichment` BC, not `Structure`

src/Akeneo/Pim/Structure/Bundle/Query/PublicApi/AssociationType/Cache/CachedGetAssociationTypeCodes.php
mmetayer: So not needed IMO, I think we should just return all the association type codes at once

tests/back/Acceptance/AssociationType/InMemoryGetAssociationTypeCodes.php
```diff
@@ -0,0 +1,32 @@
...
+    private $associationTypeRepository;
+
+    public function __construct(InMemoryAssociationTypeRepository $associationTypeRepository)
+    {
+        $this->associationTypeRepository = $associationTypeRepository;
+    }
+
+    public function findAll(): \Iterator
+    {
+        /** @var AssociationType $associationType */
```
mmetayer: please remove this line

src/Akeneo/Pim/Structure/Bundle/Query/PublicApi/AssociationType/Cache/CachedGetAssociationTypeCodes.php
nmarniesse: I don't see any problem to keep the code like that. If I remove this, the code is simpler but there is potentially a vulnerability: if someone imports several association types he can break the rule engine.  
Maybe it's over-engineering but I prefer security.

src/Akeneo/Pim/Structure/Component/Query/PublicApi/AssociationType/GetAssociationTypeCodes.php
nmarniesse: Good catch! I don't remember why I put that in Structure...

nmarniesse:
> Please move to the right bounded context. Also, I'm not sure we even need this query thinking

I fix the bounded context. We need this query for EE in order to clear the associations. Instead of fetching and hydrating all association type entities, here we fetch only the code and we can use cache. 
I proposed a solution where we can clear the associations without fetching association type codes but it was not very clean and with Pierre we decide to use this solution instead

PR #11811, opened 2020-03-23
PIM-9157: Fix performance issue when loading the data of a product group

LaurentPetard:
There's a performance issue when displaying the edit page of a group that has a lot of products.

The problem comes from the `GroupNormalizer` that iterate on all the products of a group to retrieve the product identifiers. It makes Doctrine loads and hydrates all the products and it's very costly.

The list of product identifiers is necessary for the frontend, and it would be complicated to change that. So the simplest and cheapest solution to fix this problem is to add a SQL query to retrieve the product identifiers.

Execution time of the end-point that loads a group with 1000 products (on a PIM installed locally with dev env and Icecat catalog)

- before: ~900ms
- after: ~200ms (same time with only one product)

### DIFF
--- a/src/Akeneo/Pim/Enrichment/Component/Product/Normalizer/InternalApi/GroupNormalizer.php
+++ b/src/Akeneo/Pim/Enrichment/Component/Product/Normalizer/InternalApi/GroupNormalizer.php
@@ -3,6 +3,7 @@
 namespace Akeneo\Pim\Enrichment\Component\Product\Normalizer\InternalApi;
 
 use Akeneo\Pim\Enrichment\Component\Product\Model\GroupInterface;
+use Akeneo\Pim\Enrichment\Component\Product\Query\GetGroupProductIdentifiers;
 use Akeneo\Platform\Bundle\UIBundle\Provider\StructureVersion\StructureVersionProviderInterface;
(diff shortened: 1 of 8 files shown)


src/Akeneo/Pim/Enrichment/Component/Product/Normalizer/InternalApi/GroupNormalizer.php
```diff
@@ -32,22 +33,21 @@ class GroupNormalizer implements NormalizerInterface, CacheableSupportsMethodInt
...
+    /** @var GetGroupProductIdentifiers */
+    private $getGroupProductIdentifiers;
+
     public function __construct(
         NormalizerInterface $groupNormalizer,
         StructureVersionProviderInterface $structureVersionProvider,
         VersionManager $versionManager,
-        NormalizerInterface $versionNormalizer
+        NormalizerInterface $versionNormalizer,
+        GetGroupProductIdentifiers $getGroupProductIdentifiers
```
jmleroux: BC break to document in CHANGELOG

LaurentPetard: Already done. Or I forgot something?

PR #11815, opened 2020-03-23
API-970: Audit Destination connections

willy-ahva:

`OK`: Done / Validated
`-`: Not needed


src/Akeneo/Connectivity/Connection/front/src/audit/components/Charts.tsx
```diff
@@ -32,14 +28,34 @@ export const Charts = () => {
...
+            </>;
     }
 
-    return (
-        <>
-            <DataSourceCharts />
-            <UserSurvey />
-        </>
-    );
+    return <>{render}</>;
```
pchasle: I would like to propose something simpler, because it is really hard to understand the logic here.
```tsx
    if (sourceConnections.length == 0 && destinationConnections.length == 0) {
        return <NoConnection />;
    }

    const Charts = [<DataSourceCharts />, <DataDestinationCharts />];
    if (sourceConnections.length === 0 && destinationConnections.length > 0) {
        Charts.reverse();
    }

    return (
        <>
            {Charts}
            <UserSurvey />
        </>
    );
```

src/Akeneo/Connectivity/Connection/front/src/audit/useConnectionSelector.ts
```diff
@@ -0,0 +1,33 @@
+import {Dispatch, SetStateAction, useEffect, useState} from 'react';
```
pchasle: File name should be `useConnectionSelect.ts`

src/Akeneo/Connectivity/Connection/front/src/common/components/PageError.tsx
```diff
@@ -1,6 +1,6 @@
 import React, {FC, ReactNode} from 'react';
 import styled from 'styled-components';
-import defaultImgUrl from '../assets/illustrations/api.svg';
+import defaultImgUrl from '../assets/illustrations/charts.svg';
```
pchasle: This error page is also used by the settings, and I think we should keep the API svp anyway as it's used when we have no connections yet.
If you want to use another illustration there is a imgUrl props too.

PR #11826, opened 2020-03-25
API-1029: Format Audit dates on the backend

pchasle:

`OK`: Done / Validated
`-`: Not needed


willy-ahva:
I think we use the date formater elsewhere. Are you sure we can remove it?

pchasle:
> I think we use the date formater elsewhere. Are you sure we can remove it?

I checked on 4.0. Didn't do it on master (will check now).
But we shouldn't use it at all considering the potential issues, so I will replace it elsewhere if necessary.

PR #11838, opened 2020-03-26
PIM-9163: total_fields limit of elasticsearch should be configurable

jmleroux:
Total_fields limit of elasticsearch should be configurable on serenity instances

As per this documentation, this is something that can be needed by some clients:

https://docs.akeneo.com/4.0/maintain_pim/common_issues/index.html#the-limit-of-total-fields-is-reached-in-elasticsearch

### DIFF
--- a/.env
+++ b/.env
@@ -11,4 +11,4 @@ APP_INDEX_HOSTS=elasticsearch:9200
 APP_PRODUCT_AND_PRODUCT_MODEL_INDEX_NAME=akeneo_pim_product_and_product_model
 MAILER_URL=null://localhost?encryption=tls&auth_mode=login&username=foo&password=bar&sender_address=no-reply@example.com
 AKENEO_PIM_URL=http://localhost:8080
-APP_ELASTICSEARCH_TOTAL_FIELDS_LIMIT=1000
+APP_ELASTICSEARCH_TOTAL_FIELDS_LIMIT=10000
--- a/.env
+++ b/.env
@@ -11,3 +11,4 @@ APP_INDEX_HOSTS=elasticsearch:9200
 APP_PRODUCT_AND_PRODUCT_MODEL_INDEX_NAME=akeneo_pim_product_and_product_model
 MAILER_URL=null://localhost?encryption=tls&auth_mode=login&username=foo&password=bar&sender_address=no-reply@example.com
 AKENEO_PIM_URL=http://localhost:8080
+APP_ELASTICSEARCH_TOTAL_FIELDS_LIMIT=1000
--- a/config/services/pim_parameters.yml
+++ b/config/services/pim_parameters.yml
@@ -31,8 +31,6 @@ parameters:
     # Maximum number of resources when updating a list of resources
     api_input_max_resources_number: 100
 
-    # Default value if env var is not defined
-    env(APP_ELASTICSEARCH_TOTAL_FIELDS_LIMIT): 10000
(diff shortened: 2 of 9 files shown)


.env
```diff
@@ -11,3 +11,4 @@ APP_INDEX_HOSTS=elasticsearch:9200
 APP_PRODUCT_AND_PRODUCT_MODEL_INDEX_NAME=akeneo_pim_product_and_product_model
 MAILER_URL=null://localhost?encryption=tls&auth_mode=login&username=foo&password=bar&sender_address=no-reply@example.com
 AKENEO_PIM_URL=http://localhost:8080
+APP_ELASTICSEARCH_TOTAL_FIELDS_LIMIT=1000
```
jjanvier: it was set to 10 000
not 1000

config/services/pim_parameters.yml
```diff
@@ -31,6 +31,10 @@ parameters:
     # Maximum number of resources when updating a list of resources
     api_input_max_resources_number: 100
 
+    # Default value if env var is not defined
+    env(APP_ELASTICSEARCH_TOTAL_FIELDS_LIMIT): 1000
```
jjanvier: same here

config/services/pim_parameters.yml
```diff
@@ -31,6 +31,10 @@ parameters:
...
+    elasticsearch_total_fields_limit: '%env(APP_ELASTICSEARCH_TOTAL_FIELDS_LIMIT)%'
```
nmarniesse: The last 2 lines could be replaced by:

```
    elasticsearch_total_fields_limit: '%env(default:10000:APP_ELASTICSEARCH_TOTAL_FIELDS_LIMIT)%'
```

I didn't know that, just found it in https://symfony.com/doc/4.4/configuration/env_var_processors.html

.env
jmleroux: Do not hesitate to use the "suggestion" github feature for this kind of fix :wink:

config/services/pim_parameters.yml
jmleroux: I tested it, it doesn't work:
> env(default:fallback_param:BAR)
> Retrieves the value of the parameter fallback_param when the BAR env var is not available

It's not a default value, but a default parameter and that's not what we want

nmarniesse: Ah ok, so it should work like that:

```
    elasticsearch_total_fields_limit_default: 10000 
    elasticsearch_total_fields_limit: '%env(default:elasticsearch_total_fields_limit_default:APP_ELASTICSEARCH_TOTAL_FIELDS_LIMIT)%'
```

You can keep your solution, they are both equal IMO.

make-file/test.mk
```diff
@@ -32,7 +32,7 @@ ifeq ($(CI),true)
 	$(DOCKER_COMPOSE) run -T -u www-data --rm php php vendor/bin/phpspec run --format=junit > var/tests/phpspec/specs.xml
 	.circleci/find_non_executed_phpspec.sh
 else
-	${PHP_RUN} vendor/bin/phpspec run
+	${PHP_RUN} vendor/bin/phpspec run -v
```
jjanvier: to revert I suppose

config/services/pim_parameters.yml
```diff
@@ -31,6 +31,10 @@ parameters:
     # Maximum number of resources when updating a list of resources
     api_input_max_resources_number: 100
 
+    # Default value if env var is not defined
+    env(APP_ELASTICSEARCH_TOTAL_FIELDS_LIMIT): 10000
```
BitOne: Use the .env file to define default values.

PR #11844, opened 2020-03-26
PIM-9164: Fix build property path for localized attributes validation

LaurentPetard:
A client encountered this problem:
When attempting to save a product with a comma-separated decimal value (i.e 13,24) inside of a localized price attribute, the PIM fails leaving in a blank page. It should display an error message on the price field.

The blank page is due to a Javascript error because the locale is defined instead of the channel in the JSON response of the product edition.
The reason is that Backend side, the property path set in the validation constraint is malformed. It contains only the attribute code and the locale code separated by a dash. There is no channel because the attribute is not scopable, but it should replace the channel by "<all_channels>" to have the right format : "attribute-channel-locale". By having a property path like "attribute-locale", the violation normalizer interprets the locale as the channel.

The problem concerns only the localizable but not scopable attributes, and only with two constraints specific to the PIM: https://github.com/akeneo/pim-community-dev/blob/4.0/src/Akeneo/Tool/Component/Localization/Validator/Constraints/NumberFormat.php and https://github.com/akeneo/pim-community-dev/blob/4.0/src/Akeneo/Tool/Component/Localization/Validator/Constraints/DateFormatValidator.php because they set the property path of the violation themselves. For the native Symfony constraints, the path is defined differently.

There's maybe a better solution that would involve changing the way we define the property path of a constraint violation, but it would need a lot of rework. And anyways, the property path built in https://github.com/akeneo/pim-community-dev/blob/4.0/src/Akeneo/Pim/Enrichment/Component/Product/Localization/Localizer/AttributeConverter.php is wrong, because it cannot be interpreted for the attributes that are only localizable or scopable.

### DIFF
--- a/src/Akeneo/Pim/Enrichment/Component/Product/Localization/Localizer/AttributeConverter.php
+++ b/src/Akeneo/Pim/Enrichment/Component/Product/Localization/Localizer/AttributeConverter.php
@@ -175,23 +175,16 @@ class AttributeConverter implements AttributeConverterInterface
      * Build the property path of the attribute
      *
      * @param array  $data
-     * @param string $code
+     * @param string $attributeCode
      *
      * @return string
      */
-    protected function buildPropertyPath(array $data, $code)
+    protected function buildPropertyPath(array $data, $attributeCode)
     {
-        $path = $code;
+        $channelCode = isset($data['scope']) && '' !== $data['scope'] ? $data['scope'] : '<all_channels>';
+        $localeCode = isset($data['locale']) && '' !== $data['locale'] ? $data['locale'] : '<all_locales>';
 
-        if (isset($data['scope']) && '' !== $data['scope']) {
-            $path .= sprintf('-%s', $data['scope']);
-        }
-
-        if (isset($data['locale']) && '' !== $data['locale']) {
-            $path .= sprintf('-%s', $data['locale']);
-        }
-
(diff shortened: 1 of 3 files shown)


src/Akeneo/Pim/Enrichment/Component/Product/Localization/Localizer/AttributeConverter.php
```diff
@@ -175,23 +175,16 @@ protected function convertToDefaultFormat(LocalizerInterface $localizer, array $
...
+     * @param string $attributeCode
      *
      * @return string
      */
-    protected function buildPropertyPath(array $data, $code)
+    protected function buildPropertyPath(array $data, $attributeCode)
     {
-        $path = $code;
+        $channelCode = isset($data['scope']) ? $data['scope'] : '<all_channels>';
+        $localeCode = isset($data['locale']) ? $data['locale'] : '<all_locales>';
```
nmarniesse: If $data['locale'] is an empty string, the isset will return true. So the $localeCode will be an empty string too. Think its not what we want, we want `<all_locales>` in this case

LaurentPetard: In practice the locale and the channel cannot be an empty string, but yes it's more rigorous to add a check on that.

PR #11834, opened 2020-03-26
MET-138: Change external GET endpoint to return units as object

ValentinMumble:
**Description (for Contributor and Core Developer)**

**Definition Of Done (for Core Developer only)**

| Q                                 | A
| --------------------------------- | ---
| Added Specs                       | Todo
| Added legacy Behats               | Todo
| Added acceptance tests            | Todo
| Added integration tests           | Todo
| Changelog updated                 | Todo
| Review and 2 GTM                  | Todo
| Micro Demo to the PO (Story only) | Todo
| Migration script                  | -
| Tech Doc                          | -

`Todo`: Pending / Work in progress
`OK`: Done / Validated
`-`: Not needed

PR #11858, opened 2020-03-30
MET-137: move constraint validation into respective directory

StevenVAIDIE:

- We splitted the Validators that were applied to the command classes in their respective directories : `Validation/CreateMeasurementFamily` and `Validation/SaveMeasurementFamily` instead of only `Validation/MeasurementFamily`.
- We added the method `MeasurementFamily::normalizeWithIndexedUnits` into the model.
- We updated the Save controller to switch between a `CreateMeasurementFamilyCommand` and `SaveMeasurementFamilyCommand` depending on the request.
- The Save controller supports partial updates
- We rewrited the tests for the External API Save endpoint with the following use cases:
  - it_create_a_measurement_family_when_it_does_not_exists
  - it_returns_an_error_when_the_measurement_family_does_not_have_a_code
  - it_returns_an_error_when_the_units_are_not_correctly_indexed
  - it_does_nothing_when_the_measurement_family_already_exists_and_is_exactly_the_same
  - it_add_an_unit_when_it_does_not_exist
  - it_add_a_label_translation
  - it_update_a_label_translation
  - it_update_an_unit_operations
  - it_creates_multiple_measurement_families
  - it_returns_an_error_when_the_measurement_family_list_does_not_have_the_right_structure
  - it_returns_an_error_when_the_measurement_family_creation_does_not_have_the_right_structure
  - it_returns_an_error_when_the_measurement_family_update_does_not_have_the_right_structure

### DIFF
--- a/src/Akeneo/Tool/Bundle/MeasureBundle/Controller/ExternalApi/SaveMeasurementFamiliesAction.php
+++ b/src/Akeneo/Tool/Bundle/MeasureBundle/Controller/ExternalApi/SaveMeasurementFamiliesAction.php
@@ -22,8 +22,8 @@ use Symfony\Component\HttpFoundation\JsonResponse;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\HttpKernel\Exception\BadRequestHttpException;
-use Symfony\Component\Validator\Validator\ValidatorInterface;
 use Symfony\Component\Validator\Constraints as Assert;
+use Symfony\Component\Validator\Validator\ValidatorInterface;
 
 /**
  * @author    Samir Boulil <samir.boulil@akeneo.com>
--- a/src/Akeneo/Tool/Bundle/MeasureBundle/Controller/ExternalApi/SaveMeasurementFamiliesAction.php
+++ b/src/Akeneo/Tool/Bundle/MeasureBundle/Controller/ExternalApi/SaveMeasurementFamiliesAction.php
@@ -14,7 +14,7 @@ use Akeneo\Tool\Bundle\MeasureBundle\Controller\ExternalApi\JsonSchema\Measureme
 use Akeneo\Tool\Bundle\MeasureBundle\Exception\MeasurementFamilyNotFoundException;
 use Akeneo\Tool\Bundle\MeasureBundle\Model\MeasurementFamily;
 use Akeneo\Tool\Bundle\MeasureBundle\Model\MeasurementFamilyCode;
-use Akeneo\Tool\Bundle\MeasureBundle\Persistence\MeasurementFamilyRepository;
+use Akeneo\Tool\Bundle\MeasureBundle\Persistence\MeasurementFamilyRepositoryInterface;
 use Akeneo\Tool\Component\Api\Exception\ViolationHttpException;
 use Akeneo\Tool\Component\Api\Normalizer\Exception\ViolationNormalizer;
 use Symfony\Component\HttpFoundation\JsonResponse;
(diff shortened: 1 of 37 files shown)


src/Akeneo/Tool/Bundle/MeasureBundle/Controller/ExternalApi/SaveMeasurementFamiliesAction.php
```diff
@@ -38,18 +48,30 @@ class SaveMeasurementFamiliesAction
...
+        MeasurementFamilyRepository $measurementFamilyRepository
     ) {
         $this->measurementFamilyListValidator = $measurementFamilyListValidator;
+        $this->measurementFamilyCommonStructureValidator = $measurementFamilyCommonStructureValidator;
         $this->measurementFamilyStructureValidator = $measurementFamilyStructureValidator;
         $this->validator = $validator;
         $this->violationNormalizer = $violationNormalizer;
         $this->saveMeasurementFamilyHandler = $saveMeasurementFamilyHandler;
+        $this->createMeasurementFamilyHandler = $createMeasurementFamilyHandler;
+        $this->measurementFamilyRepository = $measurementFamilyRepository;
```
tseho: it is ok if we inject the repository there ? We need to check if the measurement family code exists to switch between `create` and `save` commands, and also guess the appropriate return code.

ValentinMumble: I guess so but why not the interface?

tseho: You are right, this is a mistake.

src/Akeneo/Tool/Bundle/MeasureBundle/Model/MeasurementFamily.php
```diff
@@ -62,6 +62,20 @@ function (Unit $unit) {
         ];
     }
 
+    public function normalizeWithIndexedUnits(): array
```
ValentinMumble: We decided to not put it in the model in this PR https://github.com/akeneo/pim-community-dev/pull/11834/files#diff-c2b785254b6fefa7e19c523bdb106c65L60 but I guess if we do, we should update it there too

tseho: I do agree we should also update for the GET.
Why didn't you put it in the model ?

ValentinMumble: As it was only for the external API we assumed it was more an infra layer-related issue than a model one but I really don't mind either solution

StevenVAIDIE:
Currently, the API allow scientific annotation for convert_from_standard[value] on internal and external controller (we only have validation on front-end for internal controller with digit regex). 

When we add unit with scientific annotation we save the scientific annotation inside database. But i think that we cannot save the family on interface.
When we update scientific annotation with the same scientific annotation there is no specific problem, we save the scientific annotation inside database even if the measurement family is used on a attribute.
But, when we update float convert_from_standard[value], with scientific annotation and the unit is already assign to attribute, we throw an error because we consider that the scientific annotation is not the same than previous float.

Currently we have the problem when we launch the [migration project](https://github.com/akeneo/YamlToMeasurement) with this file https://github.com/akeneo/MeasureBundle/blob/master/Resources/config/measure.yml
In the "Length" family the "MIL" unit have a convert_from_standard[value]=0.0000254. The project YamlToMeasurement cast this float to 2.54E-5.

tseho:
@StevenVAIDIE good catch, can you create a ticket in the sprint ?

PR #11888, opened 2020-04-03
PIM-9174: PHP_IDE_CONFIG is not dependant from the PIM edition

jmleroux:


Doodoune:
You will have to do the same in EE - to be consistent.

PR #11910, opened 2020-04-08
PIM-9177: Fix showing/updating attributes in attribute group when too many attributes

SamirBoulil:

When there is too many attributes within a group, and a user tries to list all the attributes for this group. The UI freezes.

Solution:

We only display the first 500 attributes of this group, when the number of attributes exceeds 500. We display a helper message at the end of the link and a link to the attribute list screen pre-filtered with the current attribute group code.

Screen Shot 2020-04-08 at 11 29 19

EDIT 10/04/2020:

In this new version, we filter the 500 attributes to show in the front-end (before we did this in the back-end). This way, we avoid updating the attribute with only 500 attributes instead of the whole attribute group entity.

I introduced a query function to speed things up during the update in the attribute group controller to avoid falling into a php maximum execution time error.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Akeneo/Pim/Structure/Bundle/Controller/InternalApi/AttributeGroupController.php
+++ b/src/Akeneo/Pim/Structure/Bundle/Controller/InternalApi/AttributeGroupController.php
@@ -4,6 +4,7 @@ namespace Akeneo\Pim\Structure\Bundle\Controller\InternalApi;
 
 use Akeneo\Pim\Enrichment\Bundle\Filter\CollectionFilterInterface;
 use Akeneo\Pim\Structure\Bundle\Event\AttributeGroupEvents;
+use Akeneo\Pim\Structure\Bundle\Query\InternalAPI\AttributeGroup\Sql\FindAttributeCodesForAttributeGroup;
 use Akeneo\Pim\Structure\Component\Model\AttributeGroupInterface;
 use Akeneo\Pim\Structure\Component\Model\AttributeInterface;
 use Akeneo\Tool\Component\StorageUtils\Factory\SimpleFactoryInterface;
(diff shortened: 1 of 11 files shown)


src/Akeneo/Platform/Bundle/UIBundle/Resources/public/js/attribute-group/form/tab/attribute.js
```diff
@@ -12,26 +12,31 @@ define(
...
+        Router,
         BaseForm,
         i18n,
         UserContext,
         FetcherRegistry,
         SecurityContext,
         Dialog,
-        template
+        template,
+        jqueryUi, // TODO: remove if not used
```
jmleroux, line 38: I think we can remove it.

src/Akeneo/Pim/Structure/Bundle/Controller/InternalApi/AttributeGroupController.php
```diff
@@ -79,23 +80,28 @@ class AttributeGroupController
...
+     * @param SaverInterface                      $saver
+     * @param RemoverInterface                    $remover
+     * @param EntityRepository                    $attributeRepository
+     * @param ObjectUpdaterInterface              $attributeUpdater
+     * @param SaverInterface                      $attributeSaver
+     * @param SecurityFacade                      $securityFacade
+     * @param SimpleFactoryInterface              $attributeGroupFactory
+     * @param EventDispatcherInterface            $eventDispatcher
+     * @param CollectionFilterInterface           $inputFilter
+     * @param FindAttributeCodesForAttributeGroup $attributeCodesForAttributeGroup
```
jmleroux, line 104: I would simply remove all the phpDoc as it is not necessary anymore with type hinting

src/Akeneo/Pim/Structure/Bundle/Controller/InternalApi/AttributeGroupController.php
```diff
@@ -440,13 +426,25 @@ protected function getAttributeGroupOr404(string $identifier): AttributeGroupInt
...
+        // TODO pull-up: Remove this conditional, use the attributeCodesForAttributeGroup query function (else part)
+        if (null === $this->attributeCodesForAttributeGroup) {
+            $attributeCodesBefore = array_map(
+                function (AttributeInterface $attribute) {
+                    return $attribute->getCode();
+                },
+                $attributeGroup->getAttributes()->toArray()
+            );
+        } else {
+            $attributeCodesBefore =$this->attributeCodesForAttributeGroup->execute($newAttributeGroup['code']);
```
SamirBoulil, line 445:
```suggestion
            $attributeCodesBefore =$this->findAttributeCodesForAttributeGroup->execute($newAttributeGroup['code']);
```

PR #11917, opened 2020-04-09
API-1065: Create ConnectionContext service

pierrejolly:

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Akeneo/Connectivity/Connection/back/Infrastructure/ConnectionContext.php
+++ b/src/Akeneo/Connectivity/Connection/back/Infrastructure/ConnectionContext.php
@@ -1,0 +1,54 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Akeneo\Connectivity\Connection\Infrastructure;
+
+use Akeneo\Connectivity\Connection\Domain\Settings\Model\Read\Connection;
+
+/**
+ * @author Pierre Jolly <pierre.jolly@akeneo.com>
+ * @copyright 2020 Akeneo SAS (http://www.akeneo.com)
+ * @license http://opensource.org/licenses/osl-3.0.php Open Software License (OSL 3.0)
+ */
+class ConnectionContext
+{
+    /** @var Connection */
+    private $connection;
+
+    /** @var bool */
+    private $collectable;
+
+    /** @var bool */
+    private $areCredentialsValidCombination;
+
+    public function getConnection(): Connection
+    {
+        return $this->connection;
+    }
+
+    public function setConnection(Connection $connection): void
+    {
+        $this->connection = $connection;
+    }
+
+    public function isCollectable(): bool
+    {
+        return $this->collectable;
+    }
+
+    public function setCollectable(bool $collectable): void
+    {
+        $this->collectable = $collectable;
+    }
+
+    public function areCredentialsValidCombination(): bool
+    {
+        return $this->areCredentialsValidCombination;
+    }
+
+    public function setAreCredentialsValidCombination(bool $areCredentialsValidCombination): void
+    {
+        $this->areCredentialsValidCombination = $areCredentialsValidCombination;
+    }
+}
--- a/src/Akeneo/Connectivity/Connection/back/Infrastructure/ConnectionContext.php
+++ b/src/Akeneo/Connectivity/Connection/back/Infrastructure/ConnectionContext.php
@@ -85,7 +85,7 @@ class ConnectionContext
         }
 
         if (null === $this->getConnection()) {
-            throw new \Exception('You must initialize client id and username before using this service.');
+            throw new \LogicException('You must initialize client id and username before using this service.');
         }
 
         return $this->collectable = $this->getConnection()->auditable() && $this->areCredentialsValidCombination();
@@ -98,7 +98,7 @@ class ConnectionContext
         }
 
         if (null === $this->clientId || null === $this->username) {
-            throw new \Exception('You must initialize client id and username before using this service.');
+            throw new \LogicException('You must initialize client id and username before using this service.');
         }
 
         $this->areCredentialsValidCombination = $this->areCredentialsValidCombinationQuery
--- a/src/Akeneo/Connectivity/Connection/back/Infrastructure/ConnectionContext.php
+++ b/src/Akeneo/Connectivity/Connection/back/Infrastructure/ConnectionContext.php
@@ -4,7 +4,10 @@ declare(strict_types=1);
 
 namespace Akeneo\Connectivity\Connection\Infrastructure;
 
-use Akeneo\Connectivity\Connection\Domain\Settings\Model\Read\Connection;
+use Akeneo\Connectivity\Connection\Domain\Settings\Model\Write\Connection;
+use Akeneo\Connectivity\Connection\Domain\Settings\Persistence\Repository\ConnectionRepository;
+use Akeneo\Connectivity\Connection\Domain\WrongCredentialsConnection\Persistence\Query\AreCredentialsValidCombinationQuery;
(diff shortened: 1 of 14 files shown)


src/Akeneo/Connectivity/Connection/back/Infrastructure/ConnectionContext.php
```diff
@@ -0,0 +1,97 @@
...
+        $this->username = $username;
+    }
+
+    public function getConnection(): Connection
+    {
+        if (null !== $this->connection) {
+            return $this->connection;
+        }
+
+        $connectionCode = $this->selectConnectionCodeByClientIdQuery->execute($this->clientId);
```
pchasle, line 69: We need to check if the clientId is not null here **and** if `selectConnectionCodeByClientIdQuery` doesn't return null.
(We should probably fix `DbalSelectConnectionCodeByClientIdQuery` too, that cast a `(string)` on a potentially false value)

It would be ideal if `selectConnectionCodeByClientIdQuery->execute` return null when no connection is found
and that `ConnectionContext->getConnection` return null too.
In case you call this service on a non-api request.

`isCollectable` and `areCredentialsValidCombination` should also handle the case when the service is not initialized and there is no Connection, probably by throwing Exceptions this time.

src/Akeneo/Connectivity/Connection/back/tests/Unit/spec/Infrastructure/ConnectionContextSpec.php
```diff
@@ -0,0 +1,32 @@
...
+use Akeneo\Connectivity\Connection\Domain\WrongCredentialsConnection\Persistence\Query\SelectConnectionCodeByClientIdQuery;
+use Akeneo\Connectivity\Connection\Infrastructure\ConnectionContext;
+use PhpSpec\ObjectBehavior;
+
+/**
+ * @author Pierre Jolly <pierre.jolly@akeneo.com>
+ * @copyright 2020 Akeneo SAS (http://www.akeneo.com)
+ * @license http://opensource.org/licenses/osl-3.0.php Open Software License (OSL 3.0)
+ */
+class ConnectionContextSpec extends ObjectBehavior
```
pchasle, line 18: You are missing a lot of tests here

src/Akeneo/Connectivity/Connection/back/tests/Unit/spec/Infrastructure/EventSubscriber/ConnectionContextEventSubscriberSpec.php
```diff
@@ -0,0 +1,28 @@
...
+use Akeneo\Connectivity\Connection\Infrastructure\ConnectionContext;
+use Akeneo\Tool\Bundle\ApiBundle\EventSubscriber\ApiAuthenticationEvent;
+use PhpSpec\ObjectBehavior;
+
+/**
+ * @author Pierre Jolly <pierre.jolly@akeneo.com>
+ * @copyright 2020 Akeneo SAS (http://www.akeneo.com)
+ * @license http://opensource.org/licenses/osl-3.0.php Open Software License (OSL 3.0)
+ */
+class ConnectionContextEventSubscriberSpec extends ObjectBehavior
```
pchasle, line 16: Missing tests too

src/Akeneo/Connectivity/Connection/back/Infrastructure/ConnectionContext.php
```diff
@@ -0,0 +1,97 @@
...
+use Akeneo\Connectivity\Connection\Domain\Settings\Persistence\Repository\ConnectionRepository;
+use Akeneo\Connectivity\Connection\Domain\WrongCredentialsConnection\Persistence\Query\AreCredentialsValidCombinationQuery;
+use Akeneo\Connectivity\Connection\Domain\WrongCredentialsConnection\Persistence\Query\SelectConnectionCodeByClientIdQuery;
+
+/**
+ * @author Pierre Jolly <pierre.jolly@akeneo.com>
+ * @copyright 2020 Akeneo SAS (http://www.akeneo.com)
+ * @license http://opensource.org/licenses/osl-3.0.php Open Software License (OSL 3.0)
+ */
+class ConnectionContext
```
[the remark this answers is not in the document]
pierrejolly: I can't mock this service in my test if I defined the class as final :/

src/Akeneo/Connectivity/Connection/back/Infrastructure/Persistence/Dbal/Query/DbalSelectConnectionCodeByClientIdQuery.php
```diff
@@ -17,14 +17,22 @@ public function __construct(Connection $dbalConnection)
...
 FROM akeneo_connectivity_connection
 WHERE client_id = :client_id
 SQL;
 
-        return (string) $this->dbalConnection->executeQuery($sqlQuery, ['client_id' => $clientId])->fetchColumn();
+        $connectionCode = $this->dbalConnection
+            ->executeQuery($sqlQuery, ['client_id' => $clientId])
+            ->fetchColumn();
+
+        if (null === $connectionCode) {
```
pchasle, line 32: `fetchColumn` return false if there is not result.

Also, try to use the early return syntax to avoid branching.
```php
if (false === $connectionCode) {
    return null;
}
```

src/Akeneo/Connectivity/Connection/back/tests/Unit/spec/Infrastructure/EventSubscriber/ConnectionContextEventSubscriberSpec.php
```diff
@@ -0,0 +1,38 @@
...
+    ): void {
+        $this->beConstructedWith($connectionContext);
+    }
+
+    public function it_provides_subscribed_events(): void
+    {
+        $this->getSubscribedEvents()->shouldReturn([ApiAuthenticationEvent::class => ['initializeConnectionContext', 1000]]);
+    }
+
+    public function it_initialize_connection_context($connectionContext): void
```
pchasle, line 29:
```suggestion
    public function it_initializes_connection_context($connectionContext): void
```

src/Akeneo/Connectivity/Connection/back/Infrastructure/ConnectionContext.php
```diff
@@ -0,0 +1,109 @@
...
+        $connectionCode = $this->selectConnectionCodeByClientIdQuery->execute($this->clientId);
+
+        if (null === $connectionCode) {
+            return null;
+        }
+
+        return $this->connection = $this->connectionRepository->findOneByCode($connectionCode);
+    }
+
+    public function isCollectable(): bool
```
pchasle, line 81: Add some phpdoc for the exception `@throw` and I think that using a sub type like `LogicException` could be better.

src/Akeneo/Connectivity/Connection/back/tests/Unit/spec/Infrastructure/ConnectionContextSpec.php
```diff
@@ -0,0 +1,132 @@
...
+        $this->setClientId('client_id');
+        $this->setUsername('username');
+
+        $areCredentialsValidCombinationQuery->execute('client_id', 'username')->willReturn(true);
+        $areCredentialsValidCombinationQuery->execute('client_id', 'username')->shouldBeCalled();
+
+        $this->areCredentialsValidCombination()->shouldReturn(true);
+    }
+
+    public function it_throws_an_exception_during_is_collectable(): void
```
pchasle, line 118: I think that you should also add a test for when the client_id is not defined and GetConnection return null

PR #11946, opened 2020-04-17
PIM-9197: make the InMemoryGetAttributes case insensitive

nmarniesse:

The SQL implementation of GetAttributes is case insensitive. To reflect this in our tests, we have to do the same in InMemory one.  

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/tests/back/Acceptance/Attribute/InMemoryGetAttributes.php
+++ b/tests/back/Acceptance/Attribute/InMemoryGetAttributes.php
@@ -24,13 +24,18 @@ class InMemoryGetAttributes implements GetAttributes
 
     public function forCodes(array $attributeCodes): array
     {
+        $attributesIndexedByCode = [];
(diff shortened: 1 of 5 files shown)


tests/back/Acceptance/Attribute/InMemoryGetAttributes.php
```diff
@@ -24,13 +24,18 @@ public function __construct(InMemoryAttributeRepository $attributeRepository)
 
     public function forCodes(array $attributeCodes): array
     {
+        $attributesIndexedByCode = [];
```
jmleroux, line 27: We could have coded this array building in the `findAll`

nmarniesse: The findAll returns an array but I want specially index the attributes by code in lower case. The findAll does not do that

PR #11947, opened 2020-04-17
PIM-9195: Add extra ImageMagick library to handle SVG files

jmleroux:
This lib allows us to generate preview for SVG files


tseho:
Should we add a test in EE to check if the SVG is correctly supported ?

jmleroux:
> Should we add a test in EE to check if the SVG is correctly supported ?

Do we have these tests for other formats?

jmleroux:
@tseho 
We will add them in an Improvement ticket because this PR is a bit urgent... :wink:

PR #11953, opened 2020-04-21
RUL-168: Fetch default metric unit in PublicApi attribute

mmetayer:

`OK`: Done / Validated
`-`: Not needed


CHANGELOG.md
```diff
@@ -54,6 +54,8 @@
     - add `Psr\Log\LoggerInterface $logger`
 - Change constructor of `Akeneo\Pim\Enrichment\Component\Product\Normalizer\InternalApi\GroupNormalizer` to
     - add `Akeneo\Pim\Enrichment\Component\Product\Query\GetGroupProductIdentifiers`
+- Change constructor of `Akeneo\Pim\Structure\Component\Query\PublicApi\AttributeType\Attribute` to
```
mmetayer, line 57: Some thoughts:
- maybe we should call the new property `defaultMeasurementUnit` ?
- instead of fetching it as a property of the `Attribute` class, maybe we could put it in the `properties` array (would be less consistent with the attribute table, but it would avoid BC breaks)

pierallard: - I was wondering the same question. Maybe you have to see this directly with Raccoons if they planned to rename it elsewhere.
- For me this BC is acceptable, keep it here. As the constructor should not be "available from the exterior" (it's on the Query part, you should use the query to get it, not use the constructor), this is OK for me.

PR #11957, opened 2020-04-21
TIP-1456: launch EE tests for a CE PR \o/

jjanvier:
With this PR, EE tests will be launched from a CE branch. That means, when you have a CE branch only, you don't have anymore to create a fake EE branch to launch an EE build. And thus, you have less risks to break the EE when merging a CE PR.

As of now, the nightly CE build will launch only CE tests. Indeed, with this PR given, in CE we:
- checkout a CE for nightly builds
- checkout a EE for PR buils

The CE and EE `.circle/config.yml` are almost the same, but not exactly (some jobs/steps are not present in CE, the checkout part is different etc...). That means, when you'll change one of them, you'll likely have to open a PR to change the other.


.circleci/config.yml
```diff
@@ -1,65 +1,91 @@
-version: 2
+version: 2.1
+
+aliases:
```
jjanvier, line 3: check if useful

.circleci/config.yml
```diff
@@ -316,9 +603,68 @@ workflows:
...
+                    - test_front_static_acceptance_and_integration
+                    - test_back_phpunit
+                    - test_onboarder_bundle
+                    - test_back_performance
+                    - back_behat_legacy
+                    - deploy_master_environment
+                    - deploy_upgrade_oldest_prod_environment
+                    - test_back_data_migrations
+
+commands:
```
jjanvier, line 630: check if useful

jjanvier:
> I would personally use the API
https://circleci.com/docs/api/v2/#trigger-a-new-pipeline to trigger EE build + wait for the end of the build in a CE job. It avoids copy/past and to maintain two circle CI files.

@ahocquard that's a solution we thought about too. But that implies creating a kind of proxy for CircleCI/Github, host it, maintain it and so on. So that's why choose this solution instead. We already maintain 2 circle configs anyway :man_shrugging: 

> when you have a EE branch, you will execute two times the EE tests?

Indeed, if you have a CE and a EE branch, you'll launch twice the EE tests. But currently, I don't have the feeling it's a problem. People seem to work either on CE or on EE, but rarely on both.

ahocquard:
@jjanvier why do you need a proxy? you can call the API from a CE job directly, and print the link in the build.
Then you can wait for the end for this job in this step. It's easily doable I think.

We already maintain two files, but we can see in this PR that we added a lot of things that are specific to the EE (deployment, etc).  I bet that maintaining both files will be pretty complicated.

jjanvier:
> why do you need a proxy? you can call the API from a CE job directly, and print the link in the build. Then you can wait for the end for this job in this step. It's easily doable I think.

So that means in Github, you have one job instead of the (useful imo) list of jobs we have here. We were also afraid that people would be a bit lost by navigating from CE to EE in CircleCI. From a user point of view, it really bad imo. From a maintenance point of view, yes it's probably better.

> We already maintain two files, but we can see in this PR that we added a lot of things that are specific to the EE (deployment, etc). I bet that maintaining both files will be pretty complicated.

As long as the tests jobs remain the same (or quite almost the same), I'm personally OK with that. But yes, I understand your concerns.

ahocquard:
@jjanvier ok I better understand and I agree the DX is less good. 

> We were also afraid that people would be a bit lost by navigating from CE to EE in CircleCI

I would personally test it at first, because doing such PR is approximately 1/2 day (maybe 1), so pretty easy to have a quick feedback from the users and then iterate. 

My concern is that people are not maintaining the Circle CI config, and the additional complexity will not help. But maybe I'm wrong :) Wait & see :man_shrugging: 

Anyway, good job chef JJ :+1: :poivrier:

PR #11972, opened 2020-04-23
PIM-9192: Fix error being printed in the response of partial update of products API

tseho:

A deadlock can occurs during the completeness calculation, on the `postSave` event in partial update of products.

This produce the following output:
```
...
{"line":93,"identifier":"1111111204","status_code":204}
{"line":94,"identifier":"1111111205","status_code":204}
{"line":95,"identifier":"1111111206","status_code":204}
{"line":96,"identifier":"1111111207","status_code":204}
{"line":97,"identifier":"1111111208","status_code":204}
{"line":98,"identifier":"1111111209","status_code":204}
{"line":99,"identifier":"1111111210","status_code":204}
{"line":100,"identifier":"1111111211","status_code":204}{"code":500,"message":"Internal Server Error"}
```

The proposed solution is to log the error instead of breaking the API response because it's already too late to change the response status, and also because, from a functional point of view, a completeness error should not block the product update.

This has been confirmed by our dear Performance Director @ahocquard 

### DIFF
--- a/CHANGELOG-4.0.md
+++ b/CHANGELOG-4.0.md
@@ -2,7 +2,7 @@
 
 ## Bug fixes
 
-- PIM-9192: Fix deadlock error being printed in the response of partial update of products API
+- PIM-9192: Fix error being printed in the response of partial update of products API
 
 # 4.0.18 (2020-04-23)
 
--- a/CHANGELOG-4.0.md
+++ b/CHANGELOG-4.0.md
@@ -1,5 +1,9 @@
 # 4.0.x
 
+## Bug fixes
+
+- PIM-9192: Fix deadlock error being printed in the response of partial update of products API
+
 # 4.0.18 (2020-04-23)
 
 ## Bug fixes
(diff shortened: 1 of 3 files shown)


src/Akeneo/Pim/Enrichment/Bundle/Controller/ExternalApi/ProductController.php
```diff
@@ -424,7 +430,16 @@ public function partialUpdateListAction(Request $request): Response
         $resource = $request->getContent(true);
         $this->apiAggregatorForProductPostSave->activate();
         $response = $this->partialUpdateStreamResource->streamResponse($resource, [], function () {
-            $this->apiAggregatorForProductPostSave->dispatchAllEvents();
+            try {
+                $this->apiAggregatorForProductPostSave->dispatchAllEvents();
+            } catch (\Exception $exception) {
```
ahocquard, line 435: why not Throwable?

CHANGELOG-4.0.md
```diff
@@ -1,5 +1,9 @@
 # 4.0.x
 
+## Bug fixes
+
+- PIM-9192: Fix deadlock error being printed in the response of partial update of products API
```
BitOne, line 5: The deadlock was just an example, this should apply on all 500 errors (let's say, MySQL connection limit reached, or an exception launched in the application).

PR #11977, opened 2020-04-24
PIM-9219: Fix duplicated categories in the product API when present in both Product Model and Variant

tseho:

If one category is assigned first to a variant, then to the product model, in the product external API, this category is returned twice.

Instead of trying to search and update the database for this case, we simply hide the duplicates from the product API response. The reasons are:

- checking/removing duplicate categories from children while saving a product model would induce a huge performance impact
- it’s not that important/bad to have a category for both the product and the parent product model in BDD, because functionally speaking the child does have the category (it inherits it from its parent)

### DIFF
--- a/CHANGELOG-3.2.md
+++ b/CHANGELOG-3.2.md
@@ -1,5 +1,9 @@
 # 3.2.x
 
+## Bug fixes:
+
+- PIM-9219: Fix duplicated categories in the product API when present in both Product Model and Variant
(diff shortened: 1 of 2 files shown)


ahocquard:
> it’s not that important/bad to have a category for both the product

Yes it is. If you uncategorize the product model, the product will still be categorized. It's unpredictible for the user.

PR #11985, opened 2020-04-27
API-1096: Store collected errors inside ES index

pchasle:

`OK`: Done / Validated
`-`: Not needed


src/Akeneo/Connectivity/Connection/back/Infrastructure/ErrorManagement/CollectApiError.php
```diff
@@ -0,0 +1,76 @@
...
+use Akeneo\Connectivity\Connection\Domain\ErrorManagement\Persistence\Repository\BusinessErrorRepository;
+use Akeneo\Connectivity\Connection\Domain\Settings\Model\ValueObject\FlowType;
+use Akeneo\Connectivity\Connection\Infrastructure\ConnectionContext;
+use Symfony\Component\HttpKernel\Exception\HttpException;
+
+/**
+ * @copyright 2020 Akeneo SAS (http://www.akeneo.com)
+ * @license http://opensource.org/licenses/osl-3.0.php Open Software License (OSL 3.0)
+ */
+final class CollectApiError
```
pierrejolly, line 17: What about `ApiErrorCollector` ?

src/Akeneo/Connectivity/Connection/back/Infrastructure/ErrorManagement/CollectApiError.php
```diff
@@ -0,0 +1,76 @@
...
+            return;
+        }
+
+        if (false === $this->connectionContext->isCollectable() || FlowType::DATA_SOURCE !== (string) $connection->flowType()) {
+            return;
+        }
+
+        $this->errors = array_merge(
+            $this->errors,
+            array_map(function (string $error) use ($connection) {
```
willy-ahva, line 71: You should always prefer `foreach` over `array_*` because it has the best perf and IMHO array_map is less readable.

src/Akeneo/Connectivity/Connection/back/Infrastructure/ErrorManagement/CollectApiError.php
```diff
@@ -0,0 +1,76 @@
...
+        $this->extractErrorsFromHttpException = $extractErrorsFromHttpException;
+    }
+
+    public function collectFromHttpException(HttpException $httpException): void
+    {
+        $errors = $this->extractErrorsFromHttpException->extractAll($httpException);
+        $this->collect($errors);
+    }
+
+    public function save(): void
```
willy-ahva, line 46: Just a suggestion: I would say `flush`

src/Akeneo/Connectivity/Connection/back/Infrastructure/EventSubscriber/ApiErrorEventSubscriber.php
```diff
@@ -0,0 +1,76 @@
...
+        $request = $terminateEvent->getRequest();
+
+        if (false === $this->isMonitoredRoute($request)) {
+            return;
+        }
+
+        $this->collectApiError->save();
+    }
+
+    private function isMonitoredRoute(Request $request): bool
```
willy-ahva, line 68:
```suggestion
    private function isRouteMonitored(Request $request): bool
```

src/Akeneo/Connectivity/Connection/back/Infrastructure/ErrorManagement/CollectApiError.php
```diff
@@ -0,0 +1,76 @@
...
+
+        $this->repository->bulkInsert($this->errors);
+    }
+
+    /**
+     * @param string[] $errors
+     */
+    private function collect(array $errors): void
+    {
+        $connection = $this->connectionContext->getConnection();
```
willy-ahva, line 60: I think that the connection context should be put in the subscriber. Once we are in the `collect from HTTP exception` we should not early return because of wrong API Connection, IMHO it has no sense when you don't know the business.

src/Akeneo/Connectivity/Connection/back/Infrastructure/Persistence/Elasticsearch/Repository/ElasticsearchBusinessErrorRepository.php
```diff
@@ -0,0 +1,40 @@
...
+use Akeneo\Connectivity\Connection\Domain\ErrorManagement\Model\Write\BusinessError;
+use Akeneo\Connectivity\Connection\Domain\ErrorManagement\Persistence\Repository\BusinessErrorRepository;
+use Akeneo\Tool\Bundle\ElasticsearchBundle\Client;
+
+/**
+ * @author    Willy Mesnage <willy.mesnage@akeneo.com>
+ * @copyright 2020 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+class ElasticsearchBusinessErrorRepository implements BusinessErrorRepository
```
willy-ahva, line 16: Just a suggestion, replace Elasticsearch by ES? We already do it with DBAL.

src/Akeneo/Connectivity/Connection/back/Infrastructure/ErrorManagement/ExtractErrorsFromHttpException.php
```diff
@@ -0,0 +1,64 @@
...
+    public function extractAll(HttpException $httpException): array
+    {
+        if (
+            false === $httpException instanceof UnprocessableEntityHttpException
+            && false === $httpException instanceof NotFoundHttpException
+        ) {
+            return [];
+        }
+
+        $json = $this->serializer->serialize($httpException, 'json', new Context());
```
willy-ahva, line 41: It's weird to rely on this serializer to transform a class into json that we decode, then reencode. Here, we serialize then decode, whereas we just need to normalize.

src/Akeneo/Connectivity/Connection/back/Infrastructure/ErrorManagement/CollectApiError.php
pchasle: The goal is to avoid business logic in the Subscriber. But I agree with you, it's not really explicit.
Maybe we could have a different naming for the method that suggest it will Collect IF the conditions match connection+auditable?

src/Akeneo/Connectivity/Connection/back/Infrastructure/Persistence/Elasticsearch/Repository/ElasticsearchBusinessErrorRepository.php
pchasle: Honestly Elasticsearch is more explicit than ES that is quite short/broad. At least DBAL is somewhat unique as an acronym.

src/Akeneo/Connectivity/Connection/back/Infrastructure/ErrorManagement/ExtractErrorsFromHttpException.php
pchasle: Yes, but unfortunately there is not public method to only normalize and we need to use the serializer to have all the right conf tightly coupled to it.

pchasle: Bonus is we alaways have the same serialization for a given error, it was not the case before with the kernel.response that had different format between the PartialUpdateList and the rest of the methods

src/Akeneo/Connectivity/Connection/back/Infrastructure/ErrorManagement/CollectApiError.php
pchasle: I asked Willy too and I think we both prefer to avoid service name with `-or`. If you think about the name here: "it's the service that collect api error."

src/Akeneo/Connectivity/Connection/back/Infrastructure/ErrorManagement/CollectApiError.php
willy-ahva: The name of the class CollectApiError, so we already know that it's for API. Let it like that. Finally, I think you're right. :+1:

src/Akeneo/Connectivity/Connection/back/Infrastructure/ErrorManagement/ExtractErrorsFromHttpException.php
willy-ahva: Ok, I didn't work on it so I didn't know about the needed conf. Really too bad there is no normalizer.

PR #12008, opened 2020-04-29
Lock Symfony version on 4.4.7 because of validation issues with 4.4.8

jmleroux:


composer.json
```diff
@@ -82,46 +82,45 @@
...
+        "symfony/options-resolver": "4.4.7",
+        "symfony/intl": "4.4.7",
+        "symfony/finder": "4.4.7",
+        "symfony/templating": "4.4.7",
+        "symfony/proxy-manager-bridge": "4.4.7"
     },
     "require-dev": {
         "akeneo/php-coupling-detector": "0.3.*",
+        "symfony/debug-bundle": "4.4.7",
+        "symfony/web-profiler-bundle": "4.4.7",
```
jmleroux, line 123: These are dev dependencies, not prod

config/bundles.php
```diff
@@ -59,5 +59,4 @@
     // Dev related bundles
     Symfony\Bundle\DebugBundle\DebugBundle::class => ['dev' => true, 'test' => true, 'behat' => true],
     Symfony\Bundle\WebProfilerBundle\WebProfilerBundle::class => ['dev' => true, 'test' => true, 'behat' => true],
-    Symfony\Bundle\WebServerBundle\WebServerBundle::class => ['dev' => true, 'test' => true, 'behat' => true],
```
jmleroux, line 62: Removed because deprecated AND not used

jmleroux: It seems to be used. No idea why but i re  enabled it.

PR #12011, opened 2020-04-29
MET-237: Add the feature flag FLAG_QUANTIFIED_ASSOCIATION_ENABLED

tseho:

### DIFF
--- a/.env
+++ b/.env
@@ -12,3 +12,4 @@ APP_PRODUCT_AND_PRODUCT_MODEL_INDEX_NAME=akeneo_pim_product_and_product_model
 MAILER_URL=null://localhost?encryption=tls&auth_mode=login&username=foo&password=bar&sender_address=no-reply@example.com
 AKENEO_PIM_URL=http://localhost:8080
 APP_ELASTICSEARCH_TOTAL_FIELDS_LIMIT=10000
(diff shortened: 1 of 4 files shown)


.env
```diff
@@ -12,3 +12,4 @@ APP_PRODUCT_AND_PRODUCT_MODEL_INDEX_NAME=akeneo_pim_product_and_product_model
 MAILER_URL=null://localhost?encryption=tls&auth_mode=login&username=foo&password=bar&sender_address=no-reply@example.com
 AKENEO_PIM_URL=http://localhost:8080
 APP_ELASTICSEARCH_TOTAL_FIELDS_LIMIT=10000
+FLAG_QUANTIFIED_ASSOCIATION_ENABLED=0
```
SamirBoulil, line 15: ?
```suggestion
FLAG_QUANTIFIED_ASSOCIATION_ENABLED=false
```

tseho: env variables as boolean should use 0/1 since an env variable is always a string.
This is what we already do in the PIM:
https://github.com/akeneo/pim-community-dev/blob/eda2373ec1b9de73f133690a54e7cd17363367d9/.env#L2
https://github.com/akeneo/pim-community-dev/blob/eda2373ec1b9de73f133690a54e7cd17363367d9/make-file/dev.mk#L46

PR #12052, opened 2020-05-07
API-1090: Add an UnknownAttributeException

pchasle:
Add an `UnknownAttributeException` (in place of `UnknownPropertyException`) to identify more precisely this kind of error.

```
Akeneo\Pim\Enrichment\Component\Product\Exception\UnknownAttributeException
```

### DIFF
--- a/src/Akeneo/Pim/Enrichment/Bundle/Controller/ExternalApi/ProductController.php
+++ b/src/Akeneo/Pim/Enrichment/Bundle/Controller/ExternalApi/ProductController.php
@@ -526,15 +526,6 @@ class ProductController
             } else {
                 $data['values'] = [];
             }
-        } catch (UnknownPropertyException $exception) {
-            throw new DocumentedHttpException(
-                Documentation::URL . 'patch_products__code_',
-                sprintf(
-                    'Property "%s" does not exist. Check the expected format on the API documentation.',
-                    $exception->getPropertyName()
-                ),
-                $exception
-            );
         } catch (PropertyException $exception) {
             throw new DocumentedHttpException(
                 Documentation::URL . 'patch_products__code_',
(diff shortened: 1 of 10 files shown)


src/Akeneo/Pim/Enrichment/Bundle/Controller/ExternalApi/ProductController.php
```diff
@@ -526,15 +526,6 @@ protected function filterEmptyValues(ProductInterface $product, array $data): ar
...
             }
-        } catch (UnknownPropertyException $exception) {
-            throw new DocumentedHttpException(
-                Documentation::URL . 'patch_products__code_',
-                sprintf(
-                    'Property "%s" does not exist. Check the expected format on the API documentation.',
-                    $exception->getPropertyName()
-                ),
-                $exception
-            );
```
pchasle, line 537: Redundant, the `UnknownPropertyException` already display this message and is handled by the `catch PropertyException` below to create the `DocumentedHttpException`.

tests/back/Pim/Enrichment/Specification/Component/Product/Comparator/Filter/EntityWithValuesFilterSpec.php
```diff
@@ -227,6 +228,36 @@ function it_throws_an_exception_if_code_is_not_found(
             );
     }
 
+    function it_throws_an_unknown_attribute_exception_if_an_attribute_code_is_not_found(
+        $normalizer,
+        $attributeRepository,
```
mmetayer, line 233:
```suggestion
        NormalizerInterface $normalizer,
        AttributeRepositoryInterface $attributeRepository,
```

mmetayer:
Looks good, you probably forgot to update error messages in API e2e test though

mmetayer:
Another question: don't you want to apply this change to the ProductModelUpdater as well?

pchasle:
> Another question: don't you want to apply this change to the ProductModelUpdater as well?

We might do this in another PR (with the same logic). Unless you think we should ship the two together?

PR #12062, opened 2020-05-12
PIM-9245: Fix relative JS files path for less compilation

jmleroux:
Seen here in action: https://regex101.com/r/4A6zOF/5 (you can open the substitution panel to see the result)

It respect and improve the two previous examples:
https://regex101.com/r/4A6zOF/2
https://regex101.com/r/vdDaWy/5

One big error was the use of `[^vendor]` instead of `(?!=vendor)`

Regex explanation of `^.+(?=vendor).*?(?=src)\/`

`^` : easy, the string start
`.+(?=vendor)` : everything that is not followed by "vendor". Greedy mode, will go the the last _vendor_ occurrence
`.*?(?=src)` : everything that is not followed by "src". Non greedy mode, will go to the first _src_ occurrence
`\/` : require a slash behind _src_, not needed IMHO but i wanted to stay in the same line than previous code.

To sum up:
> remove everything before the FIRST 'src/' that is after the LAST 'vendor' occurrence

### DIFF
--- a/frontend/build/compile-less.js
+++ b/frontend/build/compile-less.js
@@ -45,9 +45,14 @@ function getFileContents(filePath) {
  */
 function collectBundleImports(bundlePaths) {
     // Make each path relative
+    /*
+    * regex explanation: remove everything before the FIRST 'src' that is after the LAST 'vendor' occurrence
+    * /srv/pim/vendor/my-domain/my-bundle/src/foo/bundle/resources => src/foo/bundle/resources
+    * /srv/subfolder/src/pim/vendor/my-domain/my-bundle/src/bar/bundle/resources => src/bar/bundle/resources
+    */
     const indexFiles = bundlePaths.map(bundlePath => {
         return `${bundlePath}/${BUNDLE_LESS_INDEX_PATH}`
-                .replace(/(^.+)[^vendor](?=\/src|\/vendor)\//gm, '')
+                .replace(/^(.+)(?=vendor).*?(?=src)\//gm, '');
     })
 
     const bundleImports = []
--- a/frontend/build/compile-less.js
+++ b/frontend/build/compile-less.js
@@ -52,7 +52,7 @@ function collectBundleImports(bundlePaths) {
     */
     const indexFiles = bundlePaths.map(bundlePath => {
         return `${bundlePath}/${BUNDLE_LESS_INDEX_PATH}`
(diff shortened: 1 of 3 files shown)


frontend/build/compile-less.js
```diff
@@ -45,9 +45,14 @@ function getFileContents(filePath) {
...
     // Make each path relative
+    /*
+    * regex explanation: remove everything before the FIRST 'src' that is after the LAST 'vendor' occurrence
+    * /srv/pim/vendor/my-domain/my-bundle/src/foo/bundle/resources => src/foo/bundle/resources
+    * /srv/subfolder/src/pim/vendor/my-domain/my-bundle/src/bar/bundle/resources => src/bar/bundle/resources
+    */
     const indexFiles = bundlePaths.map(bundlePath => {
         return `${bundlePath}/${BUNDLE_LESS_INDEX_PATH}`
-                .replace(/(^.+)[^vendor](?=\/src|\/vendor)\//gm, '')
+                .replace(/^.+(?=vendor).*?(?=src)\//gm, '');
```
ahocquard, line 55: A more simple way to avoid to use negative look forward, which is complicated to undertsand, would be:
```suggestion
                .replace(/^.+vendor.+?src/src/gm, '');
```

ahocquard: I added a fix, it should be "/vendor" and not only vendor in your research. A pakage could contain the same vendor or src in the namespace I suppose

jmleroux: I'm not sure: what if there is NO _vendor_ in the searched string?

jmleroux: I tried your last suggestion here: https://regex101.com/r/AOUBhA/2

:pensive:

ahocquard: Your version has a bug. Also, the `\/` at the end of your code prevents it to work (and that's logic).

See https://regex101.com/r/4A6zOF/8 (last line). You can also add `\/` at the end of the regex to see that it's an error.
My version does not have this bug: https://regex101.com/r/4A6zOF/8

My example didn't work as it is, because I didn't know the JS syntax to replace (replaced string is the second argument of the JS function, which I didn't know). 
With the example, I think you will better understand the regex.

Also, if there is no vendor, it does nothing. According to the JS comment, it is what we expect no?
Therefore it should be:
```
.replace("/^.+\/vendor\/.+?\/src\//src\//gm", '');
```

PR #12068, opened 2020-05-13
PIM-9245: Fix relative JS files path for less compilation

jmleroux:
**EDIT: this description is not right anymore. See https://github.com/akeneo/pim-community-dev/pull/12068#issuecomment-628094055 comment below for the correction.**

A simpler version of #12062 , thanks to @ahocquard .

The replace is slightly different, but simpler: instead of removing a string, it cleans and replaces it.
The search is also simpler to understand because positive lookahead can sometimes disturb people.

So now, the description of the regex is:

**Replace everything until and including the FIRST 'src/ that is after the LAST 'vendor' occurrence by `src/`**

The new regex behavior can be tested here: https://regex101.com/r/4A6zOF/9

Note from Alex: it fixes a bug also, if a directory is named "foo-vendor" or "foo-src" for example.

### DIFF
--- a/frontend/build/compile-less.js
+++ b/frontend/build/compile-less.js
@@ -51,7 +51,6 @@ function collectBundleImports(bundlePaths) {
     * /srv/subfolder/src/pim/vendor/my-domain/my-bundle/src/bar/bundle/resources => src/bar/bundle/resources
     */
     const indexFiles = bundlePaths.map(bundlePath => {
-        console.log(`${bundlePath}/${BUNDLE_LESS_INDEX_PATH}`)
         return `${bundlePath}/${BUNDLE_LESS_INDEX_PATH}`
                 .replace(/^(.+?(?=\/vendor\/)|.+(?=\/src\/))\//gm, '');
     })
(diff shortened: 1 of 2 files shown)


jmleroux:
Another link, with real examples to validate the fix: https://regex101.com/r/4A6zOF/18

jmleroux:
The fix is not good and we had luck so far it works. I understand better what is the purpose of the comment "// Make each path relative". It's not to remove the vendor part or else, it's just to remove the absolute path part on top of the root folder.

We have several case to manage and I tried to list them in this example: https://regex101.com/r/4A6zOF/21

Finally, I have to use positive lookahead because i do not want to use capturing groups in such a regex.

Let's try to explain this regex. I put arbitrary blank spaces to better show the different parts:

`^(   .+?(?=\/vendor\/)   |   .+(?=\/src\/)   )`

`^` : the string starts, let's go !

`(` : enclose two conditions, see below

`.+(?=\/vendor\/)` : all the characters before a _/vendor/_ occurence. I added the non greedy mode to address the last (tricky) example in the link provided

`|` : an OR condition, if the previous part is not matched

`.+(?=\/src\/)` : everything before the last _/src/_ occurence. Greedy mode because i can think about a use case of non greedy. (remember, we are sure at this point there is no vendor to look for)

`)` : end of the conditionnal group

PR #12102, opened 2020-05-19
PIM-9055: Allow not to drop an existing database during install

jmleroux:
This PR adds an option in the pim:installer:db command in order to allow the user not to drop an existing database, but rather empty it.

For PaaS hosts such as Platform.sh, the database is provided by the host and we cannot delete it.

### DIFF
--- a/CHANGELOG-4.0.md
+++ b/CHANGELOG-4.0.md
@@ -14,6 +14,10 @@
 - PIM-9445: Fix boolean attribute is broken on compare/translate when the attribute is localisable or scopable
 - PIM-9439: Fix PEF shakes on product with lot of simple/multi select
 
+## Technical Improvements
+
+- PIM-9055: Allow not to drop an existing database during the install proces
+
 # 4.0.56 (2020-09-09)
 
 ## Bug fixes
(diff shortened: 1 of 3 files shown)


CHANGELOG-4.0.md
```diff
@@ -1,5 +1,9 @@
 # 4.0.x
 
+## Technical Improvements
+
+- PIM-9055: Allow not to drop an existing database during the install proces
```
AymericPlanche, line 5:
```suggestion
- PIM-9055: Disallow to drop an existing database during the install proces
```

src/Akeneo/Platform/Bundle/InstallerBundle/Command/InstallCommand.php
```diff
@@ -43,7 +43,13 @@ protected function configure()
             ->setDescription('Akeneo PIM Application Installer.')
             ->addOption('force', null, InputOption::VALUE_NONE, 'Force installation')
             ->addOption('symlink', null, InputOption::VALUE_NONE, 'Install assets as symlinks')
-            ->addOption('clean', null, InputOption::VALUE_NONE, 'Clean previous install');
+            ->addOption('clean', null, InputOption::VALUE_NONE, 'Clean previous install')
+            ->addOption(
+                'doNotDropDatabase',
+                null,
+                InputOption::VALUE_NONE,
+                'Try to use an existing database if it already exists. Beware, the database data will still be deleted'
```
ahocquard, line 51:
```suggestion
                'Try to use an existing database if it already exists. It does not drop the database schema but only the tables. Therefore, beware that the database data will still be deleted.'
```

ahocquard: I still don't know the difference with dropping the schema, except for providers where you can't remove a schema (missing right)... So I would not propose it as option but just use `drop table` by default to avoid to complexify the command :thinking:

PR #12103, opened 2020-05-19
PIM-9260: do not use FPM memory_limit for CommandLauncher

jmleroux:
We should not fix the memory limit of a batch command FROM the web context because they may not have the same memory_limit setting.

### DIFF
--- a/src/Akeneo/Tool/Component/Console/CommandLauncher.php
+++ b/src/Akeneo/Tool/Component/Console/CommandLauncher.php
@@ -53,9 +53,14 @@ class CommandLauncher
      */
     protected function buildCommandString($command)
     {
-        $memoryLimit = ini_get('memory_limit');
+        $phpCommand = $this->getPhp();
 
-        return "{$this->getPhp()} -d memory_limit={$memoryLimit} {$this->rootDir}/../bin/console --env={$this->environment} {$command}";
+        if ('cli' === php_sapi_name()) {
+            $memoryLimit = ini_get('memory_limit');
+            $phpCommand = "{$this->getPhp()} -d memory_limit={$memoryLimit}";
+        }
+
+        return "{$phpCommand} {$this->rootDir}/../bin/console --env={$this->environment} {$command}";
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/Akeneo/Tool/Component/Console/CommandLauncher.php
```diff
@@ -53,9 +53,14 @@ protected function getPhp()
      */
     protected function buildCommandString($command)
     {
-        $memoryLimit = ini_get('memory_limit');
+        $phpCommand = $this->getPhp();
 
-        return "{$this->getPhp()} -d memory_limit={$memoryLimit} {$this->rootDir}/../bin/console --env={$this->environment} {$command}";
+        if ('cli' === php_sapi_name()) {
+            $memoryLimit = ini_get('memory_limit');
+            $phpCommand = "{$this->getPhp()} -d memory_limit={$memoryLimit}";
```
rybus, line 60: imho this is not necessary since memory_limit will apply and get its value from CLI config.

src/Akeneo/Tool/Component/Console/CommandLauncher.php
```diff
@@ -53,9 +53,14 @@ protected function getPhp()
...
+        }
+
+        return "{$phpCommand} {$this->rootDir}/../bin/console --env={$this->environment} {$command}";
```
rybus, line 63: no memory_limit when triggered via fpm?

src/Akeneo/Tool/Component/Console/CommandLauncher.php
jmleroux: Agree. I think i'll remove it.

src/Akeneo/Tool/Component/Console/CommandLauncher.php
jmleroux: Nope. FPM just build the command string, but the command itself is executed in a CLI environment. I can't think of any use case where a CLI command would need the memory_limit of the web context.

PR #12110, opened 2020-05-20
Update PHPStan

pchasle:
Update PHPStan from `0.11` to `0.12`
- https://phpstan.org/blog/phpstan-0-12-released
- https://phpstan.org/blog/from-minutes-to-seconds-massive-performance-gains-in-phpstan

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -3,4 +3,5 @@ includes:
 parameters:
     symfony:
         container_xml_path: %rootDir%/../../../var/cache/dev/srcKernelDevDebugContainer.xml
-
+    ignoreErrors:
+        - '#Unsafe usage of new static().#'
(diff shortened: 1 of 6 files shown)


phpstan.neon
```diff
@@ -3,4 +3,5 @@ includes:
 parameters:
     symfony:
         container_xml_path: %rootDir%/../../../var/cache/dev/srcKernelDevDebugContainer.xml
-
+    ignoreErrors:
+        - '#Unsafe usage of new static().#'
```
pchasle, line 7: This new error appeared a lot in the code base (10+ on CE & EE).
The PHPStan suggested fix is to change each impacted class to be `final`, something I didn't want to do here has the impact could be huge.

src/Akeneo/Pim/Structure/Component/Normalizer/Versioning/AttributeNormalizer.php
```diff
@@ -54,7 +54,7 @@ public function normalize($attribute, $format = null, array $context = [])
         $flatAttribute['available_locales'] = implode(self::ITEM_SEPARATOR, $standardAttribute['available_locales']);
         $flatAttribute['locale_specific'] = $attribute->isLocaleSpecific();
 
-        unset($flatAttribute['labels']);
+        unset($flatAttribute['labels']); /** @phpstan-ignore-line */
```
pchasle, line 57: No easy fix for this, PHPStan doesn't know the `$standardAttribute` type
> Cannot unset offset 'labels' on array('available_locales' => string, 'locale_specific' => mixed).

PR #12119, opened 2020-05-24
CXP-285: new connection error chart

pchasle:
- Created a data_source error chart
- Reorganize Victory base chart (regroup into a `Chart` folder & cleanup themes)
- Remove chart ordering between data_source / data_destination product events
- Created the EmptyState components (following the pattern https://www.carbondesignsystem.com/patterns/empty-states-pattern/)
- Added NotAuditable screen on the monitoring page
- Added NotDataSource screen on the monitoring page


src/Akeneo/Connectivity/Connection/back/Infrastructure/Symfony/Resources/translations/jsmessages.en_US.yml
```diff
@@ -172,6 +172,12 @@ akeneo_connectivity.connection:
                 during_the_last_seven_days: 'Total over the last 7 days:'
                 today: Today
         error_management:
+            data_source_error_chart:
```
willy-ahva, line 175: As it is in the audit domain, shouldn't it be in charts?

src/Akeneo/Connectivity/Connection/front/src/audit/components/Chart/themes.ts
```diff
@@ -0,0 +1,88 @@
+import {VictoryThemeDefinition} from 'victory';
+
+const blue = '#3b438c';
```
willy-ahva, line 3: why don't you use styled-with-theme.ts?

pchasle: These colors are for charts only, I guess we could add a dedicated section inside the theme but I would have to check with the designers for the naming of each color :thinking:

src/Akeneo/Connectivity/Connection/back/Infrastructure/Symfony/Resources/translations/jsmessages.en_US.yml
pchasle: We have:
- dashboard
  - charts
  - error_management
    - data_source_error_chart
    - widget
  - user_survey

All the product event charts are in `charts` and they should probably be separated.
I didn't want to add another one.

pchasle:
> Far better like this! Last thing why DataDestination and DataSource charts are not in components//Chart/ ?

`Chart` is really the _generic_ part of the weekly charts, whereas DataDestination & DataSource are the real components, so it put some strong boundary between the two parts

PR #12122, opened 2020-05-25
PIM-9184: Fix error count

nmarniesse:

When no error are found on the connection, the `errorCounts` property was not initialized as an array. So we had the error `Invalid argument supplied for foreach() in...` and a 500.  

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Akeneo/Connectivity/Connection/back/Domain/Audit/Model/Read/ErrorCountPerConnection.php
+++ b/src/Akeneo/Connectivity/Connection/back/Domain/Audit/Model/Read/ErrorCountPerConnection.php
@@ -19,6 +19,7 @@ final class ErrorCountPerConnection
      */
     public function __construct(array $errorCounts)
     {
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -14,6 +14,7 @@
 - PIM-9197: Make queries in InMemoryGetAttributes case insensitive
 - PIM-9213: Fix tooltip hover on Ellipsis for Family Name on creating product
 - PIM-9184: API - Fix dbal query group by part for saas instance
+- PIM-9184: API - Fix empty error count
```
pierrejolly, line 17: It's normal that this is the same id `PIM-9184` than the previous change ?

nmarniesse: I saw this bug when I worked on this issue, I assume this fix is a part of the 9184. We could create another issue to have a distinct number if it's needed

PR #12125, opened 2020-05-25
DAPI-1008: Add versioning on attribute options

LaurentPetard:
As part of the Data-quality-insight features, we need to determine when an attribute option has been modified (to check the spelling). 

To do so, the chosen solution is to add versioning on the attribute options. It will be imperceptible for the users because the current history of the attributes will be unchanged. It will add entries in the versioning table when an option is created, deleted, sorted, and when a value is updated (for the entity `Akeneo\Pim\Structure\Component\Model\AttributeOption`)

### DIFF
--- a/src/Akeneo/Tool/Bundle/VersioningBundle/UpdateGuesser/AttributeOptionUpdateGuesser.php
+++ b/src/Akeneo/Tool/Bundle/VersioningBundle/UpdateGuesser/AttributeOptionUpdateGuesser.php
@@ -35,8 +35,8 @@ class AttributeOptionUpdateGuesser implements UpdateGuesserInterface
 
         if ($entity instanceof AttributeOptionInterface) {
             $pendings[] = $entity->getAttribute();
-        } elseif ($entity instanceof AttributeOptionValueInterface) {
-            $pendings[] = $entity->getOption()->getAttribute();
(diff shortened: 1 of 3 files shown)


src/Akeneo/Tool/Bundle/VersioningBundle/UpdateGuesser/AttributeOptionUpdateGuesser.php
```diff
@@ -35,8 +35,8 @@ public function guessUpdates(EntityManager $em, $entity, $action)
 
         if ($entity instanceof AttributeOptionInterface) {
             $pendings[] = $entity->getAttribute();
-        } elseif ($entity instanceof AttributeOptionValueInterface) {
-            $pendings[] = $entity->getOption()->getAttribute();
+        } elseif ($entity instanceof AttributeOptionValueInterface && $action === UpdateGuesserInterface::ACTION_UPDATE_ENTITY) {
```
LaurentPetard, line 38: Adding the attribute in the pending entities was useless because a modification on an option value doesn't trigger any change in the attribute itself (it doesn't modify the change-set).
The option must be added only when the action is an update because an option value can't be deleted by itself (it occurs only when the entire option is deleted)

olivierpontier:
Should we add a line about it in the change log?

LaurentPetard:
@olivierpontier I don't think so, because the need is purely internal and it's transparent for the users

PR #12135, opened 2020-05-26
PIM-9255: Reload completeness on associated product when channel is switched on association view

anaisbaunelemaire:

**Issue:**
When the channel was changed on a product, the associated products grid was not reloaded.
The completeness value could be wrong.

**Solution:**
We have made the associated products reloaded when the user changes the channel in the association view.

PIM-9255

Changelog updated                 | Yes

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Akeneo/Platform/Bundle/UIBundle/Resources/public/js/product/form/associations.js
+++ b/src/Akeneo/Platform/Bundle/UIBundle/Resources/public/js/product/form/associations.js
@@ -82,6 +82,7 @@ define([
             };
             params[this.datagrids.products.paramName] = this.datagrids.products.getParamValue(associationType);
             params.dataLocale = UserContext.get('catalogLocale');
+            params.dataScope = UserContext.get('catalogScope');
 
             return params;
           }.bind(this),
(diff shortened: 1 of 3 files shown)


src/Akeneo/Platform/Bundle/UIBundle/Resources/public/js/product/form/associations.js
```diff
@@ -393,6 +394,9 @@ define([
       let urlParams = params;
       urlParams.alias = gridName;
       urlParams.params = _.clone(params);
+      if (params.dataScope) {
```
SamirBoulil, line 397: are we sure we need this check since it seems to be created above ?

jmleroux: Yep, not useful imho too

anaisbaunelemaire: You seem to be right ;-) I fix that!

PR #12167, opened 2020-05-28
RAC-26: Flat files import

ValentinMumble:

- Flat files import

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Akeneo/Pim/Enrichment/Component/Product/Comparator/Filter/ProductAssociationFilter.php
+++ b/src/Akeneo/Pim/Enrichment/Component/Product/Comparator/Filter/ProductAssociationFilter.php
@@ -39,8 +39,9 @@ class ProductAssociationFilter implements FilterInterface
     {
         $originalAssociations = $this->normalizer->normalize($product, 'standard');
         $hasAssociation = $this->hasNewAssociations($newValues);
+        $hasQuantifiedAssociation = $this->hasNewQuantifiedAssociations($newValues);
 
-        if (!$hasAssociation && empty($originalAssociations)) {
+        if (!$hasAssociation && !$hasQuantifiedAssociation && empty($originalAssociations)) {
             return [];
         }
 
@@ -55,16 +56,17 @@ class ProductAssociationFilter implements FilterInterface
             }
         }
 
+        foreach ($newValues[ProductNormalizer::FIELD_QUANTIFIED_ASSOCIATIONS] as $type => $field) {
+            foreach ($field as $key => $association) {
+                $data = $this->compareQuantifiedAssociation($originalAssociations, $association, $type, $key);
+                $result[ProductNormalizer::FIELD_QUANTIFIED_ASSOCIATIONS][$type][$key] = $data ?? [];
+            }
+        }
+
         return $result;
     }
 
-    /**
-     * Has association(s) in new values ?
-     *
-     * @param array $convertedItem
-     *
-     * @return bool
-     */
+    /* Has association(s) in new values? */
     protected function hasNewAssociations(array $convertedItem): bool
     {
         if (!isset($convertedItem['associations'])) {
@@ -80,19 +82,30 @@ class ProductAssociationFilter implements FilterInterface
         return false;
     }
 
+    /* Has quantified association(s) in new values? */
+    protected function hasNewQuantifiedAssociations(array $convertedItem): bool
+    {
+        if (!isset($convertedItem['quantified_associations'])) {
+            return false;
+        }
+
+        foreach ($convertedItem['quantified_associations'] as $association) {
+            if (!empty($association['products']) || !empty($association['product_models'])) {
+                return true;
+            }
+        }
+
+        return false;
+    }
+
     /**
      * Compare product's association
-     *
-     * @param array  $originalAssociations original associations
-     * @param array  $associations         product's associations
-     * @param string $type                 type of association (PACK, SUBSTITUTION, etc)
-     * @param string $key                  key of group (products or groups)
      *
      * @throws \LogicException
      *
      * @return array|null
      */
-    protected function compareAssociation(array $originalAssociations, array $associations, $type, $key): ?array
+    protected function compareAssociation(array $originalAssociations, array $associations, string $type, string $key): ?array
     {
         $comparator = $this->comparatorRegistry->getFieldComparator(ProductNormalizer::FIELD_ASSOCIATIONS);
         $diff = $comparator->compare($associations, $this->getOriginalAssociation($originalAssociations, $type, $key));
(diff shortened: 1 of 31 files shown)


src/Akeneo/Pim/Enrichment/Component/Product/Comparator/Filter/ProductAssociationFilter.php
```diff
@@ -105,13 +118,25 @@ protected function compareAssociation(array $originalAssociations, array $associ
...
-     * @param string $key                  key of group (products or groups)
+     * Compare product's quantified association
+     *
+     * @throws \LogicException
      *
-     * @return array
+     * @return array|null
      */
-    protected function getOriginalAssociation(array $originalAssociations, $type, $key): array
+    protected function compareQuantifiedAssociation(array $originalAssociations, array $associations, string $type, string $key): ?array
```
SamirBoulil, line 127:
```suggestion
    protected function compareQuantifiedAssociations(array $originalAssociations, array $associations, string $type, string $key): ?array
```

src/Akeneo/Pim/Enrichment/Component/Product/Connector/ArrayConverter/FlatToStandard/Product.php
```diff
@@ -222,14 +222,18 @@ protected function mapFields(array $item, array $options): array
     protected function filterFields(array $mappedItem, $withAssociations): array
     {
         if (false === $withAssociations) {
-            $isGroupAssPattern = '/^\w+'.AssociationColumnsResolver::GROUP_ASSOCIATION_SUFFIX.'$/';
-            $isProductAssPattern = '/^\w+'.AssociationColumnsResolver::PRODUCT_ASSOCIATION_SUFFIX.'$/';
-            $isProductModelAssPattern = '/^\w+'.AssociationColumnsResolver::PRODUCT_MODEL_ASSOCIATION_SUFFIX.'$/';
+            $isGroupAssPattern = '/^\w+' . AssociationColumnsResolver::GROUP_ASSOCIATION_SUFFIX . '$/';
+            $isProductAssPattern = '/^\w+' . AssociationColumnsResolver::PRODUCT_ASSOCIATION_SUFFIX . '$/';
+            $isProductModelAssPattern = '/^\w+' . AssociationColumnsResolver::PRODUCT_MODEL_ASSOCIATION_SUFFIX . '$/';
+            $isProductAssQuantityPattern = '/^\w+' . AssociationColumnsResolver::PRODUCT_ASSOCIATION_SUFFIX . AssociationColumnsResolver::QUANTITY_SUFFIX . '$/';
```
SamirBoulil, line 228:
```suggestion
            $isProductAsQuantityPattern = '/^\w+' . AssociationColumnsResolver::PRODUCT_ASSOCIATION_SUFFIX . AssociationColumnsResolver::QUANTITY_SUFFIX . '$/';
```

SamirBoulil: isGroupAssPattern, isProductAssPattern,  isProductModelAssPattern etc.

src/Akeneo/Pim/Enrichment/Component/Product/Comparator/Filter/ProductAssociationFilter.php
ValentinMumble: Well the purpose is to compare to 1 association so I guess the singular makes sense, we kept the name of the original method for regular associations.

src/Akeneo/Pim/Enrichment/Component/Product/Connector/ArrayConverter/FlatToStandard/Product.php
ValentinMumble: `isGroupAssociationPattern` instead no?

StevenVAIDIE:
You should also modify src/Akeneo/Pim/Enrichment/Component/Product/Connector/Processor/Denormalizer/ProductModelAssociationProcessor.php

src/Akeneo/Pim/Enrichment/Component/Product/Comparator/Filter/ProductAssociationFilter.php
```diff
@@ -55,16 +56,17 @@ public function filter(EntityWithValuesInterface $product, array $newValues): ar
             }
         }
 
+        foreach ($newValues[ProductNormalizer::FIELD_QUANTIFIED_ASSOCIATIONS] as $type => $field) {
+            foreach ($field as $key => $association) {
+                $data = $this->compareQuantifiedAssociation($originalAssociations, $association, $type, $key);
```
StevenVAIDIE, line 61: originalAssociations doesn't contain quantified associations, it only contain normalized association because the normalizer injected and used line 40 is "AssociationNormalizer" not "QuantifiedAssociationNormalizer".

IMHO, we should inject quantifiedAssociationsNormalizer and call function normalize like l40. By the way I think we never enter into the early return (l44) because $originalAssociation is never empty due to MissingAssociationAdder. The best way should be use hasNewQuantifiedAssociations (rename function ?) on $originalAssociations instead of checking if empty but clearly I might be wrong

src/Akeneo/Pim/Enrichment/Component/Product/Connector/ArrayConverter/FlatToStandard/ColumnsMerger.php
```diff
@@ -185,4 +244,22 @@ protected function mergePriceData(array $resultRow, array $collectedPrices)
...
+     * Merge quantified associations in result rows
+     *
+     * @param array $resultRow
+     * @param array $collectedQuantifiedAssociations
+     *
+     * @return array
+     */
+    protected function mergeQuantifiedAssociationData(array $resultRow, array $collectedQuantifiedAssociations)
+    {
+        foreach ($collectedQuantifiedAssociations as $associationTypeCode => $quantifiedAssociation) {
```
StevenVAIDIE, line 258: Problem is that we never check if number of values in quantity and identifier is same. Maybe we should handle this in this function. 

```
    private function collectQuantifiedIdentifierAssociationData(array $collectedQuantifiedAssociations, string $fieldName, $fieldValue)
    {
        list($associationTypeCode, $productType) = explode('-', $fieldName);
        if (!isset($collectedQuantifiedAssociations[$associationTypeCode])) {
            $collectedQuantifiedAssociations[$associationTypeCode] = ['products' => ['identifiers' => []], 'product_models' => ['identifiers' => []]];
        }

        if (empty($fieldValue)) {
            return $collectedQuantifiedAssociations;
        }

        $collectedQuantifiedAssociations[$associationTypeCode] = [
            $productType => [
                'identifiers' => [
                    explode(ProductAssociation::IDENTIFIER_SEPARATOR, $fieldValue)
                ]
            ]
        ];

        return $collectedQuantifiedAssociations;
    }
```

Same for quantity and in this function merge into one array after checking that number on element in identifiers and quantities are the same

src/Akeneo/Pim/Enrichment/Component/Product/Connector/ArrayConverter/FlatToStandard/ColumnsMerger.php
```diff
@@ -185,4 +232,27 @@ protected function mergePriceData(array $resultRow, array $collectedPrices)
 
         return $resultRow;
     }
+
+    private function mergeQuantifiedAssociationData(array $resultRow, array $collectedQuantifiedAssociations): array
+    {
+        foreach ($collectedQuantifiedAssociations as $associationTypeCode => $quantifiedAssociation) {
+            foreach (['products', 'product_models'] as $entityType) {
+                if (!empty($quantifiedAssociation[$entityType])) {
```
StevenVAIDIE, line 240: We should fill the result in the else condition by an empty array. Otherwise the value is empty string instead of empty array

 With early return it makes something like this :
```
            foreach (['products', 'product_models'] as $entityType) {
                if (empty($quantifiedAssociation[$entityType])) {
                    $resultRow[sprintf('%s%s%s', $associationTypeCode, AttributeColumnInfoExtractor::FIELD_SEPARATOR, $entityType)] = [];

                    continue;
                }

                if (
                    count($quantifiedAssociation[$entityType]['identifiers']) !==
                    count($quantifiedAssociation[$entityType]['quantities'])
                ) {
                    throw new \LogicException('Inconsistency detected: the count of identifiers and quantities is not the same');
                }

                $resultRow[sprintf('%s%s%s', $associationTypeCode, AttributeColumnInfoExtractor::FIELD_SEPARATOR, $entityType)] =
                array_map(function ($identifier, $quantity) {
                    return ['identifier' => $identifier, 'quantity' => (int) $quantity];
                }, $quantifiedAssociation[$entityType]['identifiers'], $quantifiedAssociation[$entityType]['quantities']);
            }

```

PR #12169, opened 2020-05-28
PIM-9276: fix values merge during indexation

nmarniesse:

https://akeneo.atlassian.net/browse/PIM-9276  

A problem occurred when we change the level of an attribute in a family variant. A cron is launched to move the values from a product/product model to another product/product model, but during this operation there is a time where the value is at 2 levels (for example the value is in the root product model and the sub product model). That is because we cannot save product models and products at same time.  
In the indexation process, we merge all the values of root product model, sub product model and variant products, and this operation does not make it right because of the `JSON_MERGE_PRESERVE`.

For example a date value is present in the root PM and the sub PM. When we merge the value becomes an array of string date. When we hydrate object just after, the process crashes because it tries to hydrate an array instead of a string.

```sql
SELECT JSON_MERGE_PRESERVE('{"my_date": {"<all_channels>": {"<all_locales>": "2020-05-28T00:00:00+00:00"}}}', '{"my_date": {"<all_channels>": {"<all_locales>": "2020-05-28T00:00:00+00:00"}}}');
-- Result: {"my_date": {"<all_channels>": {"<all_locales>": ["2020-05-28T00:00:00+00:00", "2020-05-28T00:00:00+00:00"]}}}
```

The solution here is to replace `JSON_MERGE_PRESERVE` by `JSON_MERGE_PATCH`. The last one does not convert string in array. It just keep the value in the last parameter.  

```sql
SELECT JSON_MERGE_PATCH('{"my_date": {"<all_channels>": {"<all_locales>": "2020-05-28T00:00:00+00:00"}}}', '{"my_date": {"<all_channels>": {"<all_locales>": "2020-05-28T00:00:00+00:00"}}}');
-- Result: {"my_date": {"<all_channels>": {"<all_locales>": "2020-05-28T00:00:00+00:00"}}}
```

After discusion with Chipmunks team, we think this solution is not very risky because:

- The value must be at one level only. This case does not appear often
- The `JSON_MERGE_PRESERVE` modifies the value formats that it worst

Others solutions are discussed but more complicated:
- Do not dispatch events just after each product save. Dispatch when all product and product models are saved
- Dispatch events but block the indexation and keep in mind the product/product model to index. At the end of the job when all products and product models are saved, launch the indexation

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/tests/back/Pim/Enrichment/Integration/Product/Job/ComputeFamilyVariantStructureChangesTaskletIntegration.php
+++ b/tests/back/Pim/Enrichment/Integration/Product/Job/ComputeFamilyVariantStructureChangesTaskletIntegration.php
@@ -1,0 +1,145 @@
+<?php
+
+
+declare(strict_types=1);
+
+/*
+ * This file is part of the Akeneo PIM Enterprise Edition.
+ *
+ * (c) 2020 Akeneo SAS (http://www.akeneo.com)
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace AkeneoTest\Pim\Enrichment\Integration\Product\Job;
+
+use Akeneo\Pim\Enrichment\Component\Product\Job\ComputeFamilyVariantStructureChangesTasklet;
+use Akeneo\Pim\Enrichment\Component\Product\Model\ProductModelInterface;
+use Akeneo\Pim\Structure\Component\Model\FamilyVariant;
+use Akeneo\Test\Integration\Configuration;
+use Akeneo\Test\Integration\TestCase;
+use Akeneo\Tool\Component\Batch\Job\JobParameters;
+use Akeneo\Tool\Component\Batch\Model\JobExecution;
+use Akeneo\Tool\Component\Batch\Model\StepExecution;
+use PHPUnit\Framework\Assert;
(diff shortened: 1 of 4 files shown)


mmetayer:
No test?

nmarniesse:
> No test?

Just added one for the job (ComputeFamilyVariantStructureChangesTaskletIntegration).

tests/back/Pim/Enrichment/Integration/Product/Job/ComputeFamilyVariantStructureChangesTaskletIntegration.php
```diff
@@ -0,0 +1,145 @@
...
+        $this->assertNotNull($value = $subProductModel->getValue('a_number_float'));
+        $this->assertEqualsCanonicalizing(5.5, $value->getData());
+    }
+
+    private function moveAttributeAtLevel1InFamilyVariant(string $familyVariantCode, string $attributeCode): void
+    {
+        $familyVariant = $this->getFamilyVariant($familyVariantCode);
+        $content = $this->get('pim_enrich.normalizer.family_variant')->normalize($familyVariant);
+        $content['variant_attribute_sets'][0]['attributes'][] = $attributeCode;
+        unset($content['meta']);
```
mmetayer, line 83: you can just use `pim_catalog.normalizer.standard.family_variant` :wink:

PR #12174, opened 2020-05-29
RUL-170: Fix InvalidItemFromViolationsException when invalid value is a collection of objects

mmetayer:

During an import, when the invalid value of a constraint violation is a collection of objects (without magic `__toString()` method), the instantiation of InvalidItemFromViolationsException of will fail with a `Catchable Fatal Error: Object of class Foo\Bar\Baz could not be converted to string`.
This PR fixes this issue.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Akeneo/Tool/Component/Connector/Exception/InvalidItemFromViolationsException.php
+++ b/src/Akeneo/Tool/Component/Connector/Exception/InvalidItemFromViolationsException.php
@@ -3,7 +3,6 @@
 namespace Akeneo\Tool\Component\Connector\Exception;
 
 use Akeneo\Pim\Enrichment\Component\Product\Model\ProductPriceInterface;
-use Akeneo\Pim\Enrichment\Component\Product\Model\WriteValueCollection;
 use Akeneo\Tool\Component\Batch\Item\InvalidItemException as BaseInvalidItemException;
 use Akeneo\Tool\Component\Batch\Item\InvalidItemInterface;
 use Symfony\Component\Validator\ConstraintViolationInterface;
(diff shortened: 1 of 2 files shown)


src/Akeneo/Tool/Component/Connector/Exception/InvalidItemFromViolationsException.php
```diff
@@ -43,21 +42,11 @@ public function __construct(
         foreach ($violations as $violation) {
             // TODO: re-format the message, property path doesn't exist for class constraint
             // TODO: for instance cf VariantGroupAxis
-            $invalidValue = $violation->getInvalidValue();
-            if ($invalidValue instanceof ProductPriceInterface) {
-                $invalidValue = sprintf('%s %s', $invalidValue->getData(), $invalidValue->getCurrency());
-            } elseif (is_object($invalidValue) && method_exists($invalidValue, '__toString')) {
-                $invalidValue = (string)$invalidValue;
-            } elseif (is_object($invalidValue)) {
-                $invalidValue = get_class($invalidValue);
```
mmetayer, line 52: I also got rid of this behaviour (display FQCN when invalidValue is an object), it really doesn't bring any value IMO, and makes the error message hardly readable

src/Akeneo/Tool/Component/Connector/Exception/InvalidItemFromViolationsException.php
```diff
@@ -68,4 +57,30 @@ public function __construct(
...
+        if (is_scalar($invalidValue)) {
+            return (string)$invalidValue;
+        }
+        if ($invalidValue instanceof ProductPriceInterface) {
+            return sprintf('%s %s', $invalidValue->getData(), $invalidValue->getCurrency());
+        }
+        if (is_object($invalidValue) && method_exists($invalidValue, '__toString')) {
+            return (string) $invalidValue;
+        }
+        if (is_array($invalidValue) || $invalidValue instanceof \Traversable) {
```
nmarniesse, line 72: Could be replace by `if (is_iterable($invalidValue)) {`

https://www.php.net/manual/fr/function.is-iterable.php

PR #12196, opened 2020-06-02
AOB-968: fix product label rendering, return false if values are undefined

momoss:

The `getLabelFromAttribute` method tries to find  the `attributeAsLabelIdentifier` value in form values, on Onboarder side this value is not sent to the front,
so it throws an exception when trying to apply to find method on something `undefined`, 
in that case, we return false to display the product identifier. 

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/CHANGELOG-4.0.md
+++ b/CHANGELOG-4.0.md
@@ -1,5 +1,9 @@
 # 4.0.x
 
+## Bug fixes
+
+- AOB-968: Fix product label rendering on edit form
+
 # 4.0.31 (2020-06-01)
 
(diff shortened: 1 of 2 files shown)


CHANGELOG-4.0.md
```diff
@@ -1,5 +1,9 @@
 # 4.0.x
 
+## Bug fixes
+
+- AOB-968: Return false if values are undefined
```
ValentinMerlet, line 5: I think "Fix duplicate message display on product save" or something like that is more explicit

PR #12200, opened 2020-06-02
PIM-9273: Allow to save product after locale deletion

LaurentPetard:
Please focus your review on the second commit. The first commit has been backported from master

Since the 4.0, the validation of the product values is stricter. It's no more possible to update a value for a locale not activated for the given channel  (but activated for another channel). It's was possible in 3.2, but it was a mistake.

The side effect is that if a locale is deactivated for a channel, all the products that had a value on a scopable and localizable attribute for this locale cannot be updated anymore from the UI. It's because the values of the deactivated locale are still submitted and don't pass the validation.

The solution is to filter these values when they are fetched from the database, as it has been done for the other similar cases (e.g. deleted options or deleted reference data)

### DIFF
--- a/src/Akeneo/Pim/Enrichment/Component/Product/Factory/NonExistentValuesFilter/NonExistentChannelLocaleValuesFilter.php
+++ b/src/Akeneo/Pim/Enrichment/Component/Product/Factory/NonExistentValuesFilter/NonExistentChannelLocaleValuesFilter.php
@@ -1,0 +1,69 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Akeneo\Pim\Enrichment\Component\Product\Factory\NonExistentValuesFilter;
+
+use Akeneo\Channel\Component\Query\PublicApi\ChannelExistsWithLocaleInterface;
+
+/**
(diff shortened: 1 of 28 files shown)


src/Akeneo/Pim/Enrichment/Component/Product/Factory/NonExistentValuesFilter/NonExistentChannelLocaleValuesFilter.php
```diff
@@ -0,0 +1,69 @@
...
+    {
+        $this->channelsLocales = $channelsLocales;
+    }
+
+    public function filter(OnGoingFilteredRawValues $onGoingFilteredRawValues): OnGoingFilteredRawValues
+    {
+        $filteredRawValues = [];
+        foreach ($onGoingFilteredRawValues->nonFilteredRawValuesCollectionIndexedByType() as $type => $typeAttributeProducts) {
+            $filteredRawValues[$type] = [];
+            foreach ($typeAttributeProducts as $attributeCode => $products) {
```
mmetayer, line 28:
```suggestion
            foreach ($typeAttributeProducts as $attributeCode => $rawValuesIndexedByIdentifier) {
```

PR #12214, opened 2020-06-03
CXP-322: Create VO to structure DocumentedErrorInterface

willy-ahva:

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Akeneo/Pim/Enrichment/Component/Error/Documented/DocumentationCollection.php
+++ b/src/Akeneo/Pim/Enrichment/Component/Error/Documented/DocumentationCollection.php
@@ -1,0 +1,57 @@
+<?php
+declare(strict_types=1);
+
+namespace Akeneo\Pim\Enrichment\Component\Error\Documented;
+
+/**
+ * @author    Willy Mesnage <willy.mesnage@akeneo.com>
+ * @copyright 2020 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+class DocumentationCollection
+{
+    /** @var Documentation[] */
+    private $collection;
+
+    public function __construct(array $documentations)
+    {
+        foreach ($documentations as $documentation) {
+            if (!$documentation instanceof Documentation) {
+                throw new \InvalidArgumentException(sprintf(
+                    'Class "%s" can only contain collection of "%s", instance of "%s" given.',
+                    self::class,
+                    Documentation::class,
+                    get_class($documentation)
+                ));
+            }
+        }
+        $this->collection = $documentations;
+    }
+
+    /**
+     * @return array like
+     * [
+     *     [
+     *         'message' => 'Please check your {attribute_settings}.',
+     *         'params' => [
+     *             '{attribute_settings}' => [
+     *                 'route' => 'pim_enrich_attribute_index',
+     *                 'params' => [],
+     *                 'title' => 'Attributes settings',
+     *                 'type' => 'route',
+     *                 'needle' => '{attribute_settings}'
+     *             ],
+     *         ],
+     *     ],
+     * ]
+     */
+    public function normalize(): array
+    {
+        $normalizedCollection = [];
+        foreach ($this->collection as $documentation) {
+            $normalizedCollection[] = $documentation->normalize();
+        }
+
+        return $normalizedCollection;
+    }
+}
--- a/src/Akeneo/Pim/Enrichment/Component/Error/Documented/DocumentationCollection.php
+++ b/src/Akeneo/Pim/Enrichment/Component/Error/Documented/DocumentationCollection.php
@@ -29,17 +29,17 @@ class DocumentationCollection
     }
 
     /**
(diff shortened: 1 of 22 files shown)


src/Akeneo/Pim/Enrichment/Component/Error/Documented/MessageParameterInterface.php
```diff
@@ -0,0 +1,19 @@
...
+ * @author    Willy Mesnage <willy.mesnage@akeneo.com>
+ * @copyright 2020 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+interface MessageParameterInterface
+{
+    /**
+     * @return string String to be replaced. Must be a string surrounded by "{needle}".
+     */
+    public function needle(): string;
```
pchasle, line 16:
```suggestion
```
I don't think that the Parameter should be aware of the key by which it is indexed in his parent; That should only be the responsibility of the message (Documentation).

Also, the needle term is often used in the context of search like functions (needle in a haystack) and it feels strange to use it like this. `key` or `parameterKey` would be more explicit (original I know :/)

src/Akeneo/Pim/Enrichment/Component/Product/Exception/UnknownAttributeException.php
```diff
@@ -17,11 +21,41 @@ final class UnknownAttributeException extends PropertyException implements
...
+                    new HrefMessageParameter(
+                        'What is an attribute?',
+                        'https://help.akeneo.com/pim/serenity/articles/what-is-an-attribute.html',
+                        '{what_is_attribute}'
+                    ),
+                    new HrefMessageParameter(
+                        'Manage your attributes',
+                        'https://help.akeneo.com/pim/serenity/articles/manage-your-attributes.html',
+                        '{manage_attribute}'
+                    )
```
pchasle, line 45: We can drop completely the parameter key from the Parameter objects and index them on the message directly.

Also, it feels weird (redundant) to have to give the parameter key with `{ }` if we enforce a special kind of quotation in the Documentation message:
```suggestion
                     'what_is_attribute' => new HrefMessageParameter(
                        'What is an attribute?',
                        'https://help.akeneo.com/pim/serenity/articles/what-is-an-attribute.html'
                    ),
                    'manage_attribute' => new HrefMessageParameter(
                        'Manage your attributes',
                        'https://help.akeneo.com/pim/serenity/articles/manage-your-attributes.html'
                    )
```
The alternative would be to not enforce any special quotation, then it would be logical to add them to the parameter key.

src/Akeneo/Pim/Enrichment/Component/Product/Exception/UnknownAttributeException.php
```diff
@@ -17,11 +21,41 @@ final class UnknownAttributeException extends PropertyException implements
...
 {
+    /** @var DocumentationCollection */
+    private $documentation;
+
     public function __construct(string $attributeName, string $message = '', int $code = 0, \Exception $previous = null)
     {
         parent::__construct($message, $code, $previous);
 
         $this->propertyName = $attributeName;
+        $this->documentation = new DocumentationCollection([
```
pchasle, line 32: Any reason for putting this in the constructor in place of the method `getDocumentation`? I think it's generally easier to read the code when the constructor only take care of his parameters and when static-like functions (`getErrorIdentifier`, `getDocumention`, ...) take care of their own data.

src/Akeneo/Pim/Enrichment/Component/Error/Documented/DocumentationCollection.php
```diff
@@ -0,0 +1,57 @@
...
+                    Documentation::class,
+                    get_class($documentation)
+                ));
+            }
+        }
+        $this->collection = $documentations;
+    }
+
+    /**
+     * @return array like
```
pierrejolly, line 32: @pchasle can we use [Phpstan array shape](https://phpstan.org/writing-php-code/phpdoc-types#array-shapes) to define this kind of array ?

pchasle: Sure, it would be something like this:
```
array{
  message: string,
  params: array<string, array{
    route: string,
    params: array<string, string>,
    title: string,
    type: string
  }>
}
```

pchasle: But we can have an example too.

src/Akeneo/Pim/Enrichment/Component/Product/Exception/UnknownAttributeException.php
willy-ahva: If there is an error you will not be able to instantiate the Exception using the VO inside the __construct. If you put it in a dedicated method, you throw exception only when you call it.

pchasle: Make sense to spot the error asap :+1:

PR #12224, opened 2020-06-05
CXP-278: Add info about connections in the PIM tracker

pierrejolly:

Add information to know how many connections have a PIM and how many are tracked or not

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Akeneo/Platform/Bundle/AnalyticsBundle/Query/Sql/ApiConnectionCount.php
+++ b/src/Akeneo/Platform/Bundle/AnalyticsBundle/Query/Sql/ApiConnectionCount.php
@@ -1,0 +1,53 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Akeneo\Platform\Bundle\AnalyticsBundle\Query\Sql;
+
+use Akeneo\Connectivity\Connection\Domain\Settings\Model\ValueObject\FlowType;
+use Akeneo\Tool\Component\Analytics\ApiConnectionCountQuery;
+use Doctrine\DBAL\Connection;
+use Doctrine\DBAL\FetchMode;
+
+/**
+ * @author Pierre Jolly <pierre.jolly@akeneo.com>
+ * @copyright 2020 Akeneo SAS (http://www.akeneo.com)
+ * @license http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+class ApiConnectionCount implements ApiConnectionCountQuery
+{
+    /** @var Connection */
+    private $connection;
+
+    public function __construct(Connection $connection)
+    {
+        $this->connection = $connection;
+    }
+
+    public function fetch(): array
+    {
+        $query = <<<SQL
+SELECT flow_type, count(flow_type) AS "flow_type_count", auditable
+FROM akeneo_connectivity_connection
+GROUP BY flow_type, auditable
+SQL;
+
+        $data = $this->connection->executeQuery($query)->fetchAll(FetchMode::ASSOCIATIVE);
+
+        return $this->normalize($data);
+    }
+
+    private function normalize(array $data): array
+    {
+        $normalized[FlowType::DATA_SOURCE] = ['tracked' => 0, 'untracked' => 0];
+        $normalized[FlowType::DATA_DESTINATION] = ['tracked' => 0, 'untracked' => 0];
+        $normalized[FlowType::OTHER] = ['tracked' => 0, 'untracked' => 0];
+
+        foreach ($data as $row) {
+            $status = $row['auditable'] ? 'tracked':'untracked';
+            $normalized[$row['flow_type']][$status] = $row['flow_type_count'];
+        }
+
+        return $normalized;
+    }
+}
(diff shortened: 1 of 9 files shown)


src/Akeneo/Platform/Bundle/AnalyticsBundle/Query/Sql/ApiConnectionCount.php
```diff
@@ -0,0 +1,53 @@
...
+SQL;
+
+        $data = $this->connection->executeQuery($query)->fetchAll(FetchMode::ASSOCIATIVE);
+
+        return $this->normalize($data);
+    }
+
+    private function normalize(array $data): array
+    {
+        $normalized[FlowType::DATA_SOURCE] = ['tracked' => 0, 'untracked' => 0];
```
pchasle, line 42: You need to initialize your array, that is the kind of error that phpstan will raise (if we up the strictness on the PIM)

ahocquard: No, this is when the path in the code could generate an undefined variable, as in a foreach. Here, there is no way to have an undefined variable.
Tested on a mini-project with phpstan.

```
vendor/bin/phpstan --level 7 analyse src
```

```
<php

$toto['tata'] = 1;
```

src/Akeneo/Platform/Bundle/AnalyticsBundle/Query/Sql/ApiConnectionCount.php
```diff
@@ -0,0 +1,53 @@
...
+        $normalized[FlowType::DATA_DESTINATION] = ['tracked' => 0, 'untracked' => 0];
+        $normalized[FlowType::OTHER] = ['tracked' => 0, 'untracked' => 0];
+
+        foreach ($data as $row) {
+            $status = $row['auditable'] ? 'tracked':'untracked';
```
ahocquard, line 47:
```suggestion
            $status = (bool) $row['auditable'] ? 'tracked':'untracked';
```

As any data returned by PDO is a string, it's a little bit better for readability: it says you explicitely expects a boolean.

tests/back/Platform/Integration/Analytics/Query/ApiConnectionCountIntegration.php
```diff
@@ -0,0 +1,56 @@
...
+        $this->connectionLoader->createConnection('csv', 'CSV', FlowType::DATA_SOURCE, true);
+        $this->connectionLoader->createConnection('magento', 'Magento Connector', FlowType::DATA_DESTINATION, false);
+        $this->connectionLoader->createConnection('priint', 'Priint Connector', FlowType::DATA_DESTINATION, true);
+        $this->connectionLoader->createConnection('marketplace', 'Marketplace Connector', FlowType::DATA_DESTINATION, false);
+        $this->connectionLoader->createConnection('translation', 'Translation tool', FlowType::OTHER, false);
+
+        $result = $this->apiConnectionCountQuery->fetch();
+
+        $expectedResult = [
+            FlowType::DATA_SOURCE => ['tracked' => 2, 'untracked' => 0],
```
ahocquard, line 41: are we sure to use constant here? I means it's a contract to the outside world (the tracker). The pros of using a constant is DRY. But here, the test should fail if the returned response is modified (and you can add a comment tht it's used by the tracker and should not be modified).

Otherwise, it will break in another system of another team without any warning.

tests/back/Platform/Specification/Bundle/AnalyticsBundle/DataCollector/DBDataCollectorSpec.php
```diff
@@ -73,7 +77,8 @@ function it_collects_database_statistics(
         $productValueCountQuery,
         $productValueAverageMaxQuery,
         $productValuePerFamilyAverageMaxQuery,
-        $emailDomains
+        $emailDomains,
```
ahocquard, line 80: you can prefix with the type now with phpspec.

tests/back/Platform/Specification/Bundle/AnalyticsBundle/DataCollector/DBDataCollectorSpec.php
```diff
@@ -91,26 +96,36 @@ function it_collects_database_statistics(
...
+                'max_category_in_one_category' => 25,
+                'max_category_levels' => 6,
+                'nb_product_values' => 254897,
                 'avg_product_values_by_product' => 7,
                 'avg_product_values_by_family' => 10,
                 'max_product_values_by_family' => 12,
-                'email_domains' => 'example.com,other-example.com'
+                'email_domains' => 'example.com,other-example.com',
+                'api_connection' => [
+                    FlowType::DATA_SOURCE => ['tracked' => 0, 'untracked' => 0],
```
ahocquard, line 125: the real API contract is here, so you can keep constants in the integration test and hardcode values here otherwise (with a comment to explain why).

src/Akeneo/Platform/Bundle/AnalyticsBundle/Query/Sql/ApiConnectionCount.php
pchasle: Thanks, I didn't know this. Even if that seems weird in comparisons to other languages :thinking:

PR #12235, opened 2020-06-05
RUL-276: ObjectNotFoundException extends \LogicException

mmetayer:

The phpDoc of [`ProductQueryBuilderInterface::addFilter()`](https://github.com/akeneo/pim-community-dev/blob/55841bceae131dbe475042f45fcf1d86ffb75745/src/Akeneo/Pim/Enrichment/Component/Product/Query/ProductQueryBuilderInterface.php#L30) states that it can only throw `\LogicException`, but some filters throw `ObjectNotFoundException`s, which directly extends the base `\Exception`

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Akeneo/Pim/Enrichment/Component/Product/Exception/ObjectNotFoundException.php
+++ b/src/Akeneo/Pim/Enrichment/Component/Product/Exception/ObjectNotFoundException.php
@@ -9,7 +9,7 @@ namespace Akeneo\Pim\Enrichment\Component\Product\Exception;
  * @copyright 2014 Akeneo SAS (http://www.akeneo.com)
  * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
  */
-class ObjectNotFoundException extends \Exception
+class ObjectNotFoundException extends \LogicException
 {
     /**
      * @param string     $message


src/Akeneo/Pim/Enrichment/Component/Product/Exception/ObjectNotFoundException.php
```diff
@@ -9,7 +9,7 @@
  * @copyright 2014 Akeneo SAS (http://www.akeneo.com)
  * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
  */
-class ObjectNotFoundException extends \Exception
+class ObjectNotFoundException extends \LogicException
```
nmarniesse, line 12: Not very convinced by this. Maybe `\InvalidArgumentException`? (that extends LogicException by the way)

mmetayer: :thinking: I don't see your point, '\InvalidArgumentException` should be thrown if a function/method argument has an unexpected type. The `ObjectNotFoundException` is more generic IMO (it could be thrown by a method without argument for instance)

nmarniesse: You're right, the \InvalidArgumentException is not a good idea.

PR #12250, opened 2020-06-08
GITHUB-APD-158: hardcode version name as master to allow correct analytics

ahocquard:
https://github.com/akeneo/actionable-product-data/issues/158

It's currently not possible to know how many CE users installed master version.
Actually, it's mixed with 4.0.X, which is not good good for analytics.

So, the solution is to hardcode it. I added a test to avoid a mistake during a pull-up. 

*Note*: I changed the `detect_structure_changes.sh` script. Actually, changing the version name has an impact on this script, even if it's run only in EE.
It's because when we create an ES index, we prefix it with the version name from the CE. We were somehow "lucky" it worked.

It's totally useless to put the version name when creating an index. I did that when we switched from index to alias. It was to help at finding what was the previous index and the target index. But in serenity, we have index prefixed with `3_2` and it's confusing.

Now, when we create an index, it's not prefixed anymore. There is no impact on the installed database.

*Note 2*: I changed the regex in `detect_structure_changes.sh` script, both for 4.0 and current branch. In theory, it should not for the 4.0, but there is an index created during a migration 
 (so, without the version name prefix): `akeneo_connectivity_connection_error_e6026edc-2a98-4893-b7b4-7871d7327edf`

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Akeneo/Tool/Bundle/ElasticsearchBundle/Client.php
+++ b/src/Akeneo/Tool/Bundle/ElasticsearchBundle/Client.php
@@ -261,7 +261,7 @@ class Client
         $body['aliases'] = [$this->indexName => (object) []];
 
         $params = [
-            'index' => strtolower($this->indexName . '_' . str_replace('.', '_', CommunityVersion::VERSION) . '_' . Uuid::uuid4()),
+            'index' => strtolower($this->indexName . '_' . Uuid::uuid4()),
             'body' => $body,
         ];
 
--- a/tests/back/Platform/Specification/CommunityVersionSpec.php
+++ b/tests/back/Platform/Specification/CommunityVersionSpec.php
@@ -1,0 +1,22 @@
+<?php
+
+namespace Specification\Akeneo\Platform;
+
+use Akeneo\Platform\CommunityVersion;
+use PhpSpec\ObjectBehavior;
+use PHPUnit\Framework\Assert;
+
+class CommunityVersionSpec extends ObjectBehavior
+{
+    /**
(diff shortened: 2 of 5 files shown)


tests/back/Platform/Specification/CommunityVersionSpec.php
```diff
@@ -0,0 +1,22 @@
+<?php
+
+namespace Specification\Akeneo\Platform;
+
+use Akeneo\Platform\CommunityVersion;
+use PhpSpec\ObjectBehavior;
+use PHPUnit\Framework\Assert;
+
+class CommunityVersionSpec extends ObjectBehavior
```
jjanvier, line 9: nice :+1:

src/Akeneo/Tool/Bundle/ElasticsearchBundle/Client.php
```diff
@@ -261,7 +261,7 @@ public function createIndex()
         $body['aliases'] = [$this->indexName => (object) []];
 
         $params = [
-            'index' => strtolower($this->indexName . '_' . str_replace('.', '_', CommunityVersion::VERSION) . '_' . Uuid::uuid4()),
```
docteurklein, line 264: potential bc break?

ahocquard: no, there is already an UUID, so it's impossible to guess the name of the index (you have to get it from the alias).
It was just for a kind of DX, but it was a mistake from me because it's more confusing than useful.

PR #12263, opened 2020-06-10
PIM-9280: Fix SqlGetConnectorProduct

jmleroux:
We have an issue on MySQL 8.0.20 on a GROUP BY clause returning weird (and wild results).
We found with Popeye that the inference was better resolved using the `id` primary key and the `identifier` unique index, but don't have real explanation of the fix.

It may be a mysql bug. A 8.0.18 or a 8.0.20 bug, i don't know.

Believe me, it works... Good luck for the review...


ahocquard:
See https://bugs.mysql.com/bug.php?id=99846&thanks=4 for the issue.

PR #12270, opened 2020-06-11
RUL-271: Allow relative dates for DateFilter and DateTimeFilter

mmetayer:

Allows to use a relative datetime format in the ES' DateFilter and DateTimeFilter. 
The format is stricter than PHP allows, it must be: `now` or `+/-$count $unit` (`$count` is an integer and `$unit` is one of "day", "week", "month" or "year", with an optional final "s"; the DateTimeFilter also accepts "minute" and "hour")

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Akeneo/Pim/Enrichment/Bundle/Elasticsearch/Filter/Field/DateTimeFilter.php
+++ b/src/Akeneo/Pim/Enrichment/Bundle/Elasticsearch/Filter/Field/DateTimeFilter.php
@@ -22,7 +22,7 @@ use Akeneo\Tool\Component\StorageUtils\Repository\IdentifiableObjectRepositoryIn
 class DateTimeFilter extends AbstractFieldFilter implements FieldFilterInterface
 {
     const DATETIME_FORMAT = 'Y-m-d H:i:s';
-    const RELATIVE_DATETIME_FORMAT = '/^now|([+-])([0-9]+)\s?(second|minute|hour|day|week|month|year)s?$/';
+    const RELATIVE_DATETIME_FORMAT = '/^(now|[+-][0-9]+\s?(second|minute|hour|day|week|month|year)s?)$/';
 
     /** @var IdentifiableObjectRepositoryInterface */
     protected $jobInstanceRepository;
--- a/src/Akeneo/Pim/Enrichment/Bundle/Elasticsearch/Filter/Field/DateTimeFilter.php
+++ b/src/Akeneo/Pim/Enrichment/Bundle/Elasticsearch/Filter/Field/DateTimeFilter.php
@@ -22,6 +22,7 @@ use Akeneo\Tool\Component\StorageUtils\Repository\IdentifiableObjectRepositoryIn
 class DateTimeFilter extends AbstractFieldFilter implements FieldFilterInterface
 {
     const DATETIME_FORMAT = 'Y-m-d H:i:s';
+    const RELATIVE_DATETIME_FORMAT = '/^now|([+-])([0-9]+)\s?(second|minute|hour|day|week|month|year)s?$/';
 
     /** @var IdentifiableObjectRepositoryInterface */
     protected $jobInstanceRepository;
@@ -55,7 +56,9 @@ class DateTimeFilter extends AbstractFieldFilter implements FieldFilterInterface
         if (null === $this->searchQueryBuilder) {
             throw new \LogicException('The search query builder is not initialized in the filter.');
         }
-
+        // For now, we only allow relative dates for "simple" operators (excluding BETWEEN, NOT BETWEEN, and obviously
+        // SINCE LAST N DAYS and SINCE LAST JOB)
+        $value = $this->convertRelativeDate($value);
         $this->checkValue($operator, $field, $value);
 
         switch ($operator) {
@@ -285,5 +288,21 @@ class DateTimeFilter extends AbstractFieldFilter implements FieldFilterInterface
         $dateTime->setTimezone($utcTimeZone);
 
         return $dateTime->format('c');
+    }
+
+    /**
+     * Converts a "relative date" string ("now" | "+/- {amount} {unit}") to an actual DateTime object
+     */
+    protected function convertRelativeDate($value)
+    {
+        if (is_string($value) && 1 === preg_match(self::RELATIVE_DATETIME_FORMAT, trim($value))) {
+            try {
+                return new \DateTime(trim($value), new \DateTimeZone('UTC'));
+            } catch (\Exception $e) {
+                return $value;
+            }
+        }
+
+        return $value;
     }
 }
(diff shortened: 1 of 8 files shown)


src/Akeneo/Pim/Enrichment/Bundle/Elasticsearch/Filter/Field/DateTimeFilter.php
```diff
@@ -286,4 +289,20 @@ protected function getFormattedDate($field, $value)
...
     }
+
+    /**
+     * Converts a "relative date" string ("now" | "+/- {amount} {unit}") to an actual DateTime object
+     */
+    final protected function convertRelativeDate($value)
+    {
+        if (is_string($value) && 1 === preg_match(self::RELATIVE_DATETIME_FORMAT, trim($value))) {
+            try {
+                return new \DateTime(trim($value), new \DateTimeZone('UTC'));
```
mmetayer, line 300: it would be better if we introduced a `Clock` interface, but I don't know where

src/Akeneo/Pim/Enrichment/Bundle/Resources/config/localization/presenters.yml
```diff
@@ -60,7 +60,7 @@ services:
         class: '%pim_catalog.localization.presenter.date.class%'
         arguments:
             - '@pim_catalog.localization.factory.datetime'
-            - ['created_at', 'updated_at']
+            - ['created', 'updated']
```
mmetayer, line 63: not 100% sure about that, it might be safer to let `created_at` and `updated_at`, and simply add `created` and `updated`

src/Akeneo/Pim/Enrichment/Bundle/Elasticsearch/Filter/Field/DateTimeFilter.php
```diff
@@ -22,6 +22,7 @@
 class DateTimeFilter extends AbstractFieldFilter implements FieldFilterInterface
 {
     const DATETIME_FORMAT = 'Y-m-d H:i:s';
+    const RELATIVE_DATETIME_FORMAT = '/^now|([+-])([0-9]+)\s?(second|minute|hour|day|week|month|year)s?$/';
```
pierallard, line 25: In EE, why do you don't use this variable?

tests/back/Pim/Enrichment/Specification/Bundle/Elasticsearch/Filter/Field/DateTimeFilterSpec.php
```diff
@@ -296,6 +296,65 @@ function it_adds_a_filter_with_operator_since_last_job_with_not_existing_job(
...
+                    ],
+                ],
+            ]
+        )->shouldBeCalled();
+
+        $this->setQueryBuilder($sqb);
+        $this->addFieldFilter('updated', '>', '-1 week');
+    }
+
+    function it_thrwws_an_exception_for_relative_dates_on_unsupported_operators(SearchQueryBuilder $sqb)
```
pierallard, line 317: thtrhhwhss

src/Akeneo/Pim/Enrichment/Bundle/Elasticsearch/Filter/Field/DateTimeFilter.php
mmetayer: they're not exactly the same (more captures in EE), plus I don't want to couple the EE with a filter implem

src/Akeneo/Pim/Enrichment/Bundle/Elasticsearch/Filter/Field/DateTimeFilter.php
```diff
@@ -286,4 +289,20 @@ protected function getFormattedDate($field, $value)
...
     }
+
+    /**
+     * Converts a "relative date" string ("now" | "+/- {amount} {unit}") to an actual DateTime object
+     */
+    protected function convertRelativeDate($value)
+    {
+        if (is_string($value) && 1 === preg_match(self::RELATIVE_DATETIME_FORMAT, trim($value))) {
+            try {
+                return new \DateTime(trim($value), new \DateTimeZone('UTC'));
```
nmarniesse, line 300: The timezone is it needed? I mean `new \DateTime('now', new \DateTimeZone('UTC'));`,  `new \DateTime('now', new \DateTimeZone('Pacific/Nauru'));` and `new \DateTime('now'));` represents the same time.  

If we need that because the date is formatted somewhere else, I think we must handled this in this somewhere else and not here. 

By the way if php is configured with another timezone by default, I would use this one instead of UTC (so I would remove the timezone parameter).

src/Akeneo/Pim/Enrichment/Bundle/Elasticsearch/Filter/Field/DateTimeFilter.php
nmarniesse: maybe in `Akeneo\Tool\Component\ClockInterface.php`?

src/Akeneo/Pim/Enrichment/Bundle/Elasticsearch/Filter/Field/DateTimeFilter.php
mmetayer: all datetimes are stored in UTC timezone in the PIM (MySQL / ES), hence the UTC. However you're right, conversion is already handled in the `getFormattedDate()` method, I'll remove it

src/Akeneo/Pim/Enrichment/Bundle/Elasticsearch/Filter/Field/DateTimeFilter.php
mmetayer: maybe, I might do it in another PR

tests/back/Pim/Enrichment/Integration/PQB/Filter/Field/DateTimeFilterIntegration.php
```diff
@@ -128,6 +128,57 @@ public function testOperatorNotBetween()
         $this->assert($result, ['bar', 'baz', 'foo']);
     }
 
+    public function testFilterWithRelativeDates()
+    {
+        $this->createProductWithFollowingUpdatedDate('test', new \DateTime('3 minutes ago', new \DateTimeZone('UTC')));
```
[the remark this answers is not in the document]
mmetayer: I don't understand what that would change here, a datetime is a datetime, whatever its timezone, e.g
```php
$date1 = new \DateTime('now', new \DateTimezone('UTC'));
$date2 = new \DateTime('now', new \DateTimezone('Europe/Paris'));
var_dump($date1->getTimestamp() === $date2->getTimestamp()); // bool(true)
```
And in the PIM, every datetime is stored in UTC (in ES and MySQL), so every datetime you pass for comparison is also converted in UTC. So here I can DateTime with any timezone, (even a different one for every product), the test should pass anyway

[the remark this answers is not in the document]
nmarniesse: Comparing 2 PHP datetimes is not a problem. The problem is that mysql lose the timezone.
I didn't know we convert automatically to fetch/send date in DB. This is nice. So the datetime we fetch is exactly the same than the datetime we save. (A lot of softwares store the date in the timezone configured by the server, so if we convert in a way we need to convert again in the other way)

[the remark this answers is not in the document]
mmetayer: :+1: Ok, I get what you meant now :slightly_smiling_face:

PR #12295, opened 2020-06-15
PIM-9294: Fix the removal of a validation rule in attribute edit form

mmetayer:

Fixes an issue where an empty string was sent to the back when clearing the `Validation rule` simple select field, and was then rejected by the validation.

`OK`: Done / Validated
`-`: Not needed


src/Akeneo/Pim/Structure/Bundle/Resources/config/validation/attribute.yml
```diff
@@ -292,13 +292,11 @@ Akeneo\Pim\Structure\Component\Model\Attribute:
                     - pim_reference_data_simpleselect
         validationRule:
             - Choice:
-                strict: true
```
mmetayer, line 295: default value. Passing `false` to the `strict` option was deprecated in Symfony 3.2, the option is not even documented anymore for Symfony 4.x

PR #12302, opened 2020-06-16
Fix InstallStatusManager to expect column in the same case as requested

mkilmanas:
InstallStatusManager wrongly assumed that PIM is not installed because of the flawed way of validating it - it was doing an SQL query with lowercase column name but then checking results and expecting the column name in uppercase. While this might work with some specific versions of MySQL, but MariaDB (at least 10.3) is case-sensitive in the sense that it returns the result the same way as it was requested - i.e. lowercase in this case.

This change only makes the selection and verification to use the same constant value, so that case sensitivity become irrelevant and the solution can work equally well on MySQL and MariaDB

`OK`: Done / Validated
`-`: Not needed


mkilmanas:
Is there anything I can do more to get this merged?

PR #12346, opened 2020-06-23
CXP-355: create custom regex validation constraint

pchasle:
image

### DIFF
--- a/src/Akeneo/Pim/Enrichment/Component/Product/Validator/Constraints/RegexValidator.php
+++ b/src/Akeneo/Pim/Enrichment/Component/Product/Validator/Constraints/RegexValidator.php
@@ -1,0 +1,46 @@
+<?php
+
+namespace Akeneo\Pim\Enrichment\Component\Product\Validator\Constraints;
+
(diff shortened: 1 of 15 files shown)


src/Akeneo/Pim/Enrichment/Component/Product/Validator/Constraints/RegexValidator.php
```diff
@@ -0,0 +1,46 @@
...
+            throw new UnexpectedValueException($value, 'string');
+        }
+
+        $value = (string) $value;
+
+        if (null !== $constraint->normalizer) {
+            $value = ($constraint->normalizer)($value);
+        }
+
+        if ($constraint->match xor preg_match($constraint->pattern, $value)) {
```
tomglvng, line 37: Is there a reason here to use 'xor' instead of 'or' ?

pchasle: I should have mentioned that the code is a copy of the original Regex validator from Symfony.
I guess the xor allow to inverse the condition if $match is set to false

PR #12370, opened 2020-06-29
Fix unstable legacy behats

mmetayer:

Fixes 2 unstable legacy behats

### DIFF
--- a/tests/legacy/features/pim/enrichment/product/pef/edit_product_with_localized_attributes.feature
+++ b/tests/legacy/features/pim/enrichment/product/pef/edit_product_with_localized_attributes.feature
@@ -40,10 +40,10 @@ Feature: Edit a product with localized attributes
 
   Scenario: Successfully view and edit localized price
     Given I am on the "foo" product page
-    Then the field Prix should contain "1000,50"
+    Then the product Prix in EUR should be "1000,50"
     When I change the "Prix" to "1200,50 EUR"
     And I save the product
-    Then the field Prix should contain "1200,50"
+    Then the product Prix in EUR should be "1200,50"
     And the product "foo" should have the following values:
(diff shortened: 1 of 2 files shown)


tests/legacy/features/user-management/user/edit_user.feature
```diff
@@ -46,13 +46,13 @@ Feature: Edit a user
...
-    When I edit the "mary" user
+    And I save the role
+    And I edit the "mary" user
     And I visit the "Groups and roles" tab
     And I fill in the following information:
       | Role | Tata role |
-    Then I save the user
-    And I logout
+    When I save the user
+    Then I should not see the text "There are unsaved changes"
```
mmetayer, line 55: - the logout used to be executed too quickly (so the alert 'You will lose changes to the attribute if you leave the page' was displayed)
- the `I am logged in as` step also performs a logout, no need to do it twice

tests/legacy/features/pim/enrichment/product/pef/edit_product_with_localized_attributes.feature
```diff
@@ -40,10 +40,10 @@ Feature: Edit a product with localized attributes
 
   Scenario: Successfully view and edit localized price
     Given I am on the "foo" product page
-    Then the field Prix should contain "1000,50"
+    Then the product Prix in EUR should be "1000,50"
```
mmetayer, line 43: I only managed to reproduce this fail once, but as there are 2 'Prix' fields (one for EUR and one for USD), I guess this fix should help Behat get the right one

ahocquard: Maybe an "id" order by issue (1,10,9) in the UI.

PR #12382, opened 2020-06-30
TIP-1493: launch a "regular CE build" from the same PR workflow

jjanvier:
## What?

This PR comes from a discussion we had with Juju. Sometimes the squads need to launch a regular CE build from a CE PR, just to ensure they didn't break some weird stuff in the Community Edition (probably due to a "sensitive" CE/EE override problem).

Currently, as I hope you know, a CE build actually checkouts the EE master and the proper CE branch. All EE tests are launched. And normally, EE tests also contains CE tests. A build is considered green when all EE tests are green. Which is really good, as our goal is to not break the SaaS offer when we merge a CE PR.

With this PR, you have a new approval step `ready_to_build_only_ce?`. This branch of the workflow will checkout the CE branch and launch only CE tests. **A PR is still considered mergeable when `pull_request_success` is green. That means this branch of the workflow will have no effect on our merges.** This step is **not** required to merge. This is just an help for people who want to check sensitive CE/EE overrides' problems.

ce_build

## How?

Duplicating the jobs in the CircleCI config file was out of question. Instead, I used a very handy feature of CircleCI. You can reuse the same job several times in the same workflow as long as you explicitly give a specific name. For instance:

```
workflow:
    pull_request:
        jobs:
            test_backend: ~
            test_backend:
                 name: second_test_backend
 ```

This workflow would launch the job `test_backend` twice under the names `test_backend` and `second_test_backend`.

And that works perfectly for us as our CircleCI jobs do the same between CE and EE. Indeed, all the logic is done in our Makefiles.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -556,17 +556,17 @@ workflows:
                 name: test_back_phpunit_ce
                 requires:
                     - test_back_static_and_acceptance_ce
-                    - test_front_static_acceptance_and_integration
+                    - test_front_static_acceptance_and_integration_ce
           - test_back_missing_structure_migrations:
                 name: test_back_missing_structure_migrations_ce
                 requires:
                     - test_back_static_and_acceptance_ce
-                    - test_front_static_acceptance_and_integration
+                    - test_front_static_acceptance_and_integration_ce
           - test_back_data_migrations:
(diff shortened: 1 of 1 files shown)


tseho:
Really helpful, thanks JJ !

PR #12391, opened 2020-07-01
PIM-9315: Improve the error message when you have no rights on a product

anaisbaunelemaire:

When trying (for example) to modify a product on which you do not have rights on, the PIM displayed  **"Product XXX does not exist."**, but the real problem here was that you didn't have permission to access the product XXX.

For security reasons, it was decided that this message should be improved as follows (for products and product models) :
**“Product XXX does not exist or you do not have permission to access it”.**

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/CHANGELOG-4.0.md
+++ b/CHANGELOG-4.0.md
@@ -1,5 +1,9 @@
 # 4.0.x
 
+## Improvement
+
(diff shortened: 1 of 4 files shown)


anaisbaunelemaire:
Thanks Steven for you feedback. It was a mistake, I corrected it !

PR #12392, opened 2020-07-01
Use a separate database for the tests

tseho:
**What's the issue ?**

For us developers working on the PIM daily with docker, when we launch locally any test using the database (integration, end2end, legacy behat), the database `akeneo_pim` is droped.

It is annoying because we loose each time everything we could have configured/created in our local instance of the PIM during development.

**What's the proposed solution ?**

For tests, we use another database called `akeneo_pim_test`.
By configuring the env variable `APP_DATABASE_NAME` to this dedicated database for the tests, the default one is preserved.

**What's the trick ?**

The mysql docker image support, by default, only one database. It's the one specified in `MYSQL_DATABASE`.

I'm thinking of 2 solutions:
- use a second mysql container
- create a second database in the existing mysql container

I went with the second solution, because the impact on the configuration is lighter. There is no conflict on ports, there is no need to change any configuration other than the `APP_DATABASE_NAME`, it will also consume less resources.
To automatically create this database when the container start, it's quit easy, we just need to add a script to `/docker-entrypoint-initdb.d`.

**What are the impacts for the team ?**

A fresh install of the PIM will have 0 issues with it.
A `make down && make pim-dev` will have 0 issues with it.

The only issue I'm thinking of, is when a developer will pull this change without recreating his mysql container, it will encounters an error when it will run one of those test locally. Something like `the database "akeneo_pim_test" does not exists`.

When it happens the first time, it's really easy to fix it without droping anything, with the following new command:
```
make database-test
```


config/services/test/test_services.yml
```diff
@@ -23,7 +23,7 @@ services:
         class: 'Akeneo\Tool\Component\Console\CommandLauncher'
         arguments:
             - '%kernel.root_dir%'
-            - 'prod'
+            - 'test'
```
tseho, line 26: IMHO, this is a controversial one.
There was an explicit intention to use prod for running migrations when using the migrations tests.
see https://github.com/akeneo/pim-community-dev/pull/12028

The issue we have, when prod and test does not share the database, the migrations which are run in prod env are not visible in test ...

.circleci/detect_structure_changes.sh
```diff
@@ -67,6 +67,9 @@ fi
 echo "Export env vars from .env..."
 export $(cat .env)
 
+echo "Use the database akeneo_pim_test..."
+echo "APP_DATABASE_NAME=akeneo_pim_test" >> .env.test.local
+
```
tseho, line 72: detect_structure_change load the codebase of 4.0 before applying migrations.
In the 4.0 codebase, the test database is still configured to `akeneo_pim` (and I have no intention to change that)
With the trick `.env.test.local`, the 4.0 codebase will use the expected database.

ValentinMumble (approved):

tseho:
check frozen

PR #12430, opened 2020-07-07
PIM-9344:  Correctly convert variable to string in pdf renderer

tseho:

The pdf renderer expect every attribute value to be an object with `__toString`.
In some cases, it does not work because the content is already a string (`pim_catalog_simpleselect` and `pim_catalog_multiselect` are already converted to string before that line).

A simple fix is to use `|trim` instead of `.__toString()` because the twig trim filter always cast the value to a string with `(string)$value` and it will work for both cases (objects implementing `__toString` and strings)


tseho:
check frozen

PR #12443, opened 2020-07-10
GITHUB-APD-150: Add queries to fetch data from the external service

christophe-chausseray:

Ticket : https://github.com/akeneo/actionable-product-data/issues/150

## The goal of this PR

The communication panel displays announcements. This announcements were loaded from a JSON file in the PIM before this PR. It means that each month, we had to add it in the JSON file of the PIM repository, manually, from the content of the helpcenter.

In order to get directly the announcements of the pim communication panel from the content of the help center, we must request an external API that we did (https://pim-comm-panel-staging.akeneo.com/announcements).

To test that we request correctly the external API, it does not request directly firebase but it mocks an external HTTP server to test all the layers (mainly HTTP layer) of the PIM codebase. This approach is already done on the PHP client and I did it already on several projects: it works pretty well.

The url of the external service should be an overridable parameter with an environment variable to be able to call a staging external service.

Note:
- the limit parameter is removed: it's the external service that will return a default limit instead. We can easily change the limit without releasing a new PIM version this way.
- the service returns an "endDateNotifcation" instead of "notificationDuration", that's why I changed the format

`OK`: Done / Validated
`-`: Not needed


ahocquard:
Merged with one GTM, validated with Nico for this feature

PR #12457, opened 2020-07-13
RAC-120: refactor: Rename QuantifiedAssociations by QuantifiedAssociationCollection

StevenVAIDIE:
In this PR:
- Rename quantifiedAssociations by QuantifiedAssociationCollection
- Do not rename all occurence of quantifiedAssociations by QuantifiedAssociationCollection => too much modification (https://github.com/akeneo/pim-community-dev/pull/12446/files)

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Akeneo/Pim/Enrichment/Bundle/Resources/config/validation/product.yml
+++ b/src/Akeneo/Pim/Enrichment/Bundle/Resources/config/validation/product.yml
@@ -31,8 +31,6 @@ Akeneo\Pim\Enrichment\Component\Product\Model\Product:
         associations:
             - Symfony\Component\Validator\Constraints\All:
                 - Akeneo\Pim\Enrichment\Component\Product\Validator\Constraints\AssociationTypeIsNotQuantified: ~
-        quantifiedAssociations:
-            - Symfony\Component\Validator\Constraints\Valid: ~
     getters:
         values:
             - Symfony\Component\Validator\Constraints\Valid:
@@ -41,3 +39,5 @@ Akeneo\Pim\Enrichment\Component\Product\Model\Product:
             - Akeneo\Pim\Enrichment\Component\Product\Validator\Constraints\ReferenceDataOptionsExist: ~
             - Akeneo\Pim\Enrichment\Component\Product\Validator\Constraints\ScopableValues: ~
             - Akeneo\Pim\Enrichment\Component\Product\Validator\Constraints\LocalizableValues: ~
+        quantifiedAssociations:
+            - Symfony\Component\Validator\Constraints\Valid: ~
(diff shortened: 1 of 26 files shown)


src/Akeneo/Pim/Enrichment/Bundle/Resources/config/validation/product.yml
```diff
@@ -31,8 +31,6 @@ Akeneo\Pim\Enrichment\Component\Product\Model\Product:
         associations:
             - Symfony\Component\Validator\Constraints\All:
                 - Akeneo\Pim\Enrichment\Component\Product\Validator\Constraints\AssociationTypeIsNotQuantified: ~
-        quantifiedAssociations:
-            - Symfony\Component\Validator\Constraints\Valid: ~
```
tseho, line 35: it's not needed ?

src/Akeneo/Pim/Enrichment/Component/Product/Validator/Constraints/QuantifiedAssociationsValidator.php
```diff
@@ -94,7 +93,7 @@ private function validateAssociationType(
 
         if (null === $associationType) {
             $this->context->buildViolation(
-                QuantifiedAssociationsConstraint::ASSOCIATION_TYPE_DOES_NOT_EXIST_MESSAGE
+                QuantifiedAssociations::ASSOCIATION_TYPE_DOES_NOT_EXIST_MESSAGE
```
tseho, line 96: IMHO, it's confusing to go back to the real class name.
For normal constraints, like `NotBlank` it's easy to read when you call any of its static property.
For this one, since the constraint was named like the model class, it was kinda confusing. It could have been `ValidQuantifiedAssociationCollection` or something like this.
Anyway, I'm not saying we should rename the validation class, but I would rather keep `use Akeneo\Pim\Enrichment\Component\Product\Validator\Constraints\QuantifiedAssociations as QuantifiedAssociationsConstraint`, it's a little easier to read.

src/Akeneo/Pim/Enrichment/Bundle/Resources/config/validation/product.yml
SamirBoulil: I think it moved below ⬇️  but I don't know what the "getters" thing is :/

tseho: Ah yes I've missed it !
With the symfony validator, you can put a constraint on a property. Like it was done here for `quantifiedAssociations`. Since the property is private, symfony call the property accessor that will guess the correct getter for accessing the value, usually `getQuantifiedAssociations()` and will apply the validation on this value.

Since the internal property has been renamed to `quantifiedAssociationCollection`, if the validation is set to `quantifiedAssociationCollection`, symfony will not find it because it expect `getQuantifiedAssociationCollection()` which does not exists.

You can have a validation directly on the getter https://symfony.com/doc/current/validation.html#getters so symfony will look for the getter `getQuantifiedAssociations` and ignore the internal property name.

PR #12465, opened 2020-07-16
PIM-9357: Make rules case-insensitive so it complies with family and attribute codes

anaisbaunelemaire:

**Before:** 

The rules were case sensitive. Meaning if you used a family code for your rule with no uppercase while your family code actually started with an uppercase then the rule didn't recognize it and so didn't find any product to apply the rule to.

**Solution:** 

In the FamilyFilter class, I added a way to replace the value by its familyCode.

`OK`: Done / Validated
`-`: Not needed


src/Akeneo/Pim/Enrichment/Bundle/Elasticsearch/Filter/Field/FamilyFilter.php
```diff
@@ -96,25 +96,31 @@ public function addFieldFilter(
...
      * @throws ObjectNotFoundException
      */
+
     protected function checkValue($field, $values)
     {
         FieldFilterHelper::checkArray($field, $values, static::class);
+        $familyCodes = [];
 
-        foreach ($values as $value) {
+        foreach ($values as $index => $value) {
```
mmetayer, line 112: not needed :wink:

PR #12480, opened 2020-07-20
pass bubble exception for attribute remove action.

joshiprashiddha:
Currently Attribute Remove Event Subscriber can't pass bubble exception message as exceptions are not handled in Controller for removeAction.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Akeneo/Pim/Structure/Bundle/Controller/InternalApi/AttributeController.php
+++ b/src/Akeneo/Pim/Structure/Bundle/Controller/InternalApi/AttributeController.php
@@ -381,7 +381,12 @@ class AttributeController
             );
         }
 
-        $this->remover->remove($attribute);
+        try{
+            $this->remover->remove($attribute);
+        }
+        catch (\Exception $e) {
+            return new JsonResponse(['message' => $e->getMessage()], Response::HTTP_BAD_REQUEST);
+        }
 
         return new JsonResponse(null, Response::HTTP_NO_CONTENT);
     }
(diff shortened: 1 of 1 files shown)


src/Akeneo/Pim/Structure/Bundle/Controller/InternalApi/AttributeController.php
```diff
@@ -381,7 +381,12 @@ public function removeAction(Request $request, string $code): JsonResponse
             );
         }
 
-        $this->remover->remove($attribute);
+        try{
+            $this->remover->remove($attribute);
+        }
+        catch (\Exception $e) {
```
ahocquard, line 387: It should be a dedicated exception for the removal of an attribute. It should not display any error message by default in production.

joshiprashiddha: Updated..

joshiprashiddha:
> Hello, I re-read the PR and I think it's not the good way to do what you want to achieve. Any validation error should be done in the validation step, before the removing. Unfortutanely, it's missing a validator on this usecase.
> 
> It should not catch an exception (\Exception or \LogicException) and displays the message. It's already too late. It can leak some data in a production environment.

We need to catch an exception thrown from PRE_REMOVE. Almost all controllers like FamilyController, AttributeOptionController handles the exception except AttributeController. So we need generic try/catch/exception pass exception messages from PRE_REMOVE to bubble message.

ahocquard:
Not very clean from our side to do it, but you're right, we already have this issue.

momoss (approved):

PR #12485, opened 2020-07-20
PIM-9361: Chunk large ES query

jmleroux:
Elasticsearch has a parametrized maximum request body size when indexing. (http.max_content_length, 100MB by default). If the request body is exceeded it will throw a 413 error (Request Entity Too Large), leading to an uncaught exception in the PIM. When indexing a large amount of products with a lot of data (e.g when reindexing products via the pim:product:index command), this body size can be exceeded.

This PR set a size limite for the query (100 MB) and split the query into multiple calls if it is exceeded.
The responses are then merged before returning.

### DIFF
--- a/src/Akeneo/Tool/Bundle/ElasticsearchBundle/Client.php
+++ b/src/Akeneo/Tool/Bundle/ElasticsearchBundle/Client.php
@@ -22,6 +22,9 @@ use Ramsey\Uuid\Uuid;
  */
 class Client
 {
+    /** @var int ElasticSearch max query size */
+    private const PARAMS_MAX_SIZE = 100000000;
+
     /** @var ClientBuilder */
     private $builder;
 
@@ -48,8 +51,13 @@ class Client
      * @param array         $hosts
      * @param string        $indexName
      */
-    public function __construct(ClientBuilder $builder, Loader $configurationLoader, array $hosts, $indexName, string $idPrefix = '')
-    {
+    public function __construct(
+        ClientBuilder $builder,
+        Loader $configurationLoader,
+        array $hosts,
+        $indexName,
+        string $idPrefix = ''
+    ) {
         $this->builder = $builder;
         $this->configurationLoader = $configurationLoader;
         $this->hosts = $hosts;
@@ -107,26 +115,49 @@ class Client
     public function bulkIndexes($documents, $keyAsId, Refresh $refresh = null)
     {
         $params = [];
+        $mergedResponse = [
+            'took' => 0,
+            'errors' => false,
+            'items' => [],
+        ];
 
         foreach ($documents as $document) {
             if (!isset($document[$keyAsId])) {
                 throw new MissingIdentifierException(sprintf('Missing "%s" key in document', $keyAsId));
             }
 
+            $paramsComputedSize = strlen(json_encode($params)) + strlen(json_encode($document));
+
+            if ($paramsComputedSize >= self::PARAMS_MAX_SIZE) {
+                $mergedResponse = $this->doBulkIndex($params, $mergedResponse);
+                $params = [];
+            }
+
             $params['body'][] = [
                 'index' => [
                     '_index' => $this->indexName,
-                    '_id' => $this->idPrefix.$document[$keyAsId],
+                    '_id' => $this->idPrefix . $document[$keyAsId],
                 ],
             ];
-
-            $params['body'][] = $document;
 
             if (null !== $refresh) {
                 $params['refresh'] = $refresh->getType();
             }
+
+            $params['body'][] = $document;
         }
 
+        $mergedResponse = $this->doBulkIndex($params, $mergedResponse);
+
+        if (isset($mergedResponse['errors']) && true === $mergedResponse['errors']) {
+            $this->throwIndexationExceptionFromReponse($mergedResponse);
+        }
+
+        return $mergedResponse;
+    }
+
+    private function doBulkIndex(array $params, array $mergedResponse): array
+    {
         try {
             $response = $this->client->bulk($params);
         } catch (\Exception $e) {
@@ -134,10 +165,16 @@ class Client
         }
 
         if (isset($response['errors']) && true === $response['errors']) {
-            $this->throwIndexationExceptionFromReponse($response);
+            $mergedResponse['errors'] = true;
         }
 
-        return $response;
+        $mergedResponse['items'] = array_merge($response['items'], $mergedResponse['items']);
+
+        if (isset($response['took'])) {
+            $mergedResponse['took'] += $response['took'];
+        }
+
+        return $mergedResponse;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -42,6 +42,7 @@
         "ext-gd": ">=2.0",
         "ext-imagick": "*",
         "ext-intl": "*",
+        "ext-json": "*",
```
BitOne, line 45: I would change that on a 4.0. Maybe on master, for the sake of completeness. But in fact, we are already using json_decode/json_encode a lot in our code :)

src/Akeneo/Tool/Bundle/ElasticsearchBundle/Client.php
```diff
@@ -41,15 +43,21 @@ class Client
 
     /**
      * Configure the PHP Elasticsearch client.
-     * To learn more, please see {@link https://www.elastic.co/guide/en/elasticsearch/client/php-api/current/_configuration.html}
+     * To learn more, please see
```
BitOne, line 46: Can you do a different commit for formatting only the code, please ? It drowns the important parts of your PR into noise...

ahocquard:
I think this pseudo code would work in a better way, because as @BitOne said it's too late the way you are doing it:

```
$numberBatch = ceil(strlen(json_encode($documents)) / self::PARAMS_MAX_SIZE);
$batchDocuments = array_chunk($documents, $numberBatch);
foreach ($batchDocuments as $documents) {
// call ES +
//logique merge des errors
}
```

src/Akeneo/Tool/Bundle/ElasticsearchBundle/Client.php
```diff
@@ -125,40 +140,60 @@ public function bulkIndexes($documents, $keyAsId, Refresh $refresh = null)
             if (null !== $refresh) {
                 $params['refresh'] = $refresh->getType();
             }
+
+            if (strlen(json_encode($params)) > self::PARAMS_MAX_SIZE) {
```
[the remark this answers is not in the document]
jmleroux: > It should be a parametrized size (from env var), with a default value.

As discussed IRL, we will not do it on 4.0

src/Akeneo/Tool/Bundle/ElasticsearchBundle/Client.php
```diff
@@ -107,37 +115,62 @@ public function index($id, array $body, Refresh $refresh = null)
...
 
         foreach ($documents as $document) {
             if (!isset($document[$keyAsId])) {
                 throw new MissingIdentifierException(sprintf('Missing "%s" key in document', $keyAsId));
             }
 
+            $paramsComputedSize = strlen(json_encode($params)) + strlen(json_encode($document));
+
+            if ($paramsComputedSize >= self::PARAMS_MAX_SIZE) {
+                $this->doBulkIndex($params, $mergedResponse);
```
ahocquard:
```suggestion
                $mergedResponse = $this->doBulkIndex($params, $mergedResponse);
```

src/Akeneo/Tool/Bundle/ElasticsearchBundle/Client.php
```diff
@@ -107,37 +115,62 @@ public function index($id, array $body, Refresh $refresh = null)
...
+            'errors' => false,
+            'items' => [],
+        ];
 
         foreach ($documents as $document) {
             if (!isset($document[$keyAsId])) {
                 throw new MissingIdentifierException(sprintf('Missing "%s" key in document', $keyAsId));
             }
 
+            $paramsComputedSize = strlen(json_encode($params)) + strlen(json_encode($document));
```
ahocquard, line 129: FYI, `json_encode($params)) + strlen(json_encode($document)` !== `strlen(json_encode($document + params)` (one character diff I think)

and missing in the params:
```
 $params['body'][] = [
            $params['body'][] = [

                'index' => [
                    '_index' => $this->indexName,
                    '_id' => $this->idPrefix . $document[$keyAsId],
                ],
                ],
            ];
```

src/Akeneo/Tool/Bundle/ElasticsearchBundle/Client.php
```diff
@@ -107,37 +115,62 @@ public function index($id, array $body, Refresh $refresh = null)
...
-            $params['body'][] = $document;
-
             if (null !== $refresh) {
                 $params['refresh'] = $refresh->getType();
             }
+
+            $params['body'][] = $document;
         }
 
+        $this->doBulkIndex($params, $mergedResponse);
```
ahocquard:
```suggestion
        $mergedResponse = $this->doBulkIndex($params, $mergedResponse);
```

src/Akeneo/Tool/Bundle/ElasticsearchBundle/Client.php
```diff
@@ -107,37 +115,66 @@ public function index($id, array $body, Refresh $refresh = null)
...
                 $params['refresh'] = $refresh->getType();
             }
+
+            $params['body'][] = $document;
         }
 
+        $mergedResponse = $this->doBulkIndex($params, $mergedResponse);
+
+        if (isset($mergedResponse['errors']) && true === $mergedResponse['errors']) {
+            $this->throwIndexationExceptionFromReponse($mergedResponse);
```
ahocquard, line 153: typo but not the purpose of this PR to fix it I suppose

src/Akeneo/Tool/Bundle/ElasticsearchBundle/Client.php
```diff
@@ -107,37 +115,66 @@ public function index($id, array $body, Refresh $refresh = null)
...
                 ],
             ];
 
-            $params['body'][] = $document;
-
             if (null !== $refresh) {
                 $params['refresh'] = $refresh->getType();
             }
+
+            $params['body'][] = $document;
```
ahocquard, line 147: please keept this line above refresh for lisibility (2 lines to forge 1 insert)

src/Akeneo/Tool/Bundle/ElasticsearchBundle/Client.php
jmleroux: Exactly, the specs detected it :)

src/Akeneo/Tool/Bundle/ElasticsearchBundle/Client.php
jmleroux: I fix it anyway. Good catch.

src/Akeneo/Tool/Bundle/ElasticsearchBundle/Client.php
jmleroux: Yep, I didn't want to complicate the process just for the _index_ params.

BTW, it's also for this reason that i didn't take your (very good) suggestion about splitting the document array before iterating, because it would have ignore a lot of these small additionnal parameters and that could have lead to size errors.

jmleroux:
> I think this pseudo code would work in a better way, because as @BitOne said it's too late the way you are doing it:
> 
> ```
> $numberBatch = ceil(strlen(json_encode($documents)) / self::PARAMS_MAX_SIZE);
> $batchDocuments = array_chunk($documents, $numberBatch);
> foreach ($batchDocuments as $documents) {
> // call ES +
> //logique merge des errors
> }
> ```

I don't take it for two reasons:
- it doesn't take the additional _index_ parameters into account, and if we have a lot of them, tis can be a problem
- it create additional arrays variables which can takes a non negligible amount of memory

PR #12496, opened 2020-07-22
Fix pdf renderer for image attributes

tseho:

In the pdf renderer of products, for attribute of type `pim_catalog_image`, the twig was trying to access `content.media`.

`content` is the attribute value and it's an object with the class `Akeneo\Pim\Enrichment\Component\Product\Value\MediaValue`:
https://github.com/akeneo/pim-community-dev/blob/3a6cd4edb7e3182cd79cbfe9c4ae2aa3d01e6a45/src/Akeneo/Pim/Enrichment/Component/Product/Value/MediaValue.php#L9-L18

The problem is, this class do not have a property `media`, it's called `data`.
see https://github.com/akeneo/pim-community-dev/blob/3a6cd4edb7e3182cd79cbfe9c4ae2aa3d01e6a45/src/Akeneo/Pim/Enrichment/Component/Product/Value/MediaValue.php#L34-L37

Since it was not accessible, the fallback to `__toString` was used, returning the key of the media.
see https://github.com/akeneo/pim-community-dev/blob/3a6cd4edb7e3182cd79cbfe9c4ae2aa3d01e6a45/src/Akeneo/Pim/Enrichment/Component/Product/Value/MediaValue.php#L42-L45

The code was clearly trying to display the original filename, not the storage path.

**Currently, it's rendering like this:**

Screenshot_2020-07-22_14-19-32

**With the fix, it's rendering like this:**

Screenshot_2020-07-22_14-20-17

| Micro Demo to the PO (Story only) -

### DIFF
--- a/CHANGELOG-4.0.md
+++ b/CHANGELOG-4.0.md
@@ -1,5 +1,9 @@
 # 4.0.x
 
+## Bug fixes
+
+- Fix pdf renderer for image attributes, it was displaying the path instead of the original filename
+
 # 4.0.42 (2020-07-22)
 
 # 4.0.41 (2020-07-20)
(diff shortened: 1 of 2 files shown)


adhocore:
hi, what is the issue with just using `content.data` that exists instead of `content.media is defined` which may not exist?

PR #12500, opened 2020-07-23
PIM-9358: Break down values & properties query with CTE to use less memory

ValentinMumble:

Trying to mitigate memory usage by breaking down the query with a CTE for the group codes, the combination of the JSON array aggregate & the JSON merge on the raw values was overflowing the `sort_buffer_size` which is set at 256Kb

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/CHANGELOG-4.0.md
+++ b/CHANGELOG-4.0.md
@@ -3,6 +3,7 @@
 ## Bug fixes
 
 - Fix pdf renderer for image attributes, it was displaying the path instead of the original filename
+- PIM-9358: Break down values & properties query with CTE to use less memory
 
 # 4.0.42 (2020-07-22)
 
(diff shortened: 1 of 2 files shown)


ahocquard:
After investigation:
- the size of the products in the issue is approximately 150ko
- this size is too big to be sorted in the buffer (even with 10 products)
- sorting is needed due to the group by

Here, it's better to do a CTE because it's not needed to sort raw values.

Imagine that we want to order by this way:
```
select p.identifier, p.updated, p.raw_values 
from pim_catalog_product p 
where p.identifier IN ("a", "b", "c", "d", "e", "f", "g", "h", "i", "j")
order by p.identifier, p.updated;
```

It is not possible to do a CTE in that case, as the order would not be guaranteed. So I'm wondering what is the best way to handle it. 

Also an article: https://www.percona.com/blog/2010/10/25/impact-of-the-sort-buffer-size-in-mysql/

@BitOne @greg-akeneo have maybe some opinions about it.

PR #12509, opened 2020-07-27
Make vertical list and column configurator take up all available vert…

aligent-lturner:
…ical space

 - This is in order to provide a larger area for drag and drop functionality

By default, the column selector for the product grid has some frustrating functionality with regards to drag and drop.
Since the "Selected Columns" list only takes up as much space as required by the current list, new column entries have to be first dragged inside the list before being dragged down if you want it to appear at the bottom.
The purpose of this PR is to make the list take up all available space, so that new entries can be dragged straight below the current entries.

Before change:
image

After change:
image

`OK`: Done / Validated
`-`: Not needed


aligent-lturner:
Screengrabs of behaviour before and after change:

Before:
columns_before
After:
columns_after

PR #12514, opened 2020-07-27
PIM-9288: refresh product values after attribute option removing

nmarniesse:

Fix https://akeneo.atlassian.net/browse/PIM-9288

When an attribte option is removed, the products/product models are not updated. So the value remains and the completeness could be wrong.  
This fix allows to:
- add a job to retrieve all the products and product models that had this option, and remove it form them
- add a subscriber to launch this job each time an option is removed

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Akeneo/Pim/Enrichment/Component/Product/Job/RemoveNonExistingProductValuesTasklet.php
+++ b/src/Akeneo/Pim/Enrichment/Component/Product/Job/RemoveNonExistingProductValuesTasklet.php
@@ -15,7 +15,6 @@ namespace Akeneo\Pim\Enrichment\Component\Product\Job;
 
 use Akeneo\Pim\Enrichment\Component\Product\Query\Filter\Operators;
 use Akeneo\Pim\Enrichment\Component\Product\Storage\GetProductAndProductModelIdentifiersWithValuesIgnoringLocaleAndScope;
-use Akeneo\Pim\Structure\Component\Model\Attribute;
 use Akeneo\Pim\Structure\Component\Repository\AttributeRepositoryInterface;
 use Akeneo\Tool\Component\Batch\Model\StepExecution;
 use Akeneo\Tool\Component\Connector\Step\TaskletInterface;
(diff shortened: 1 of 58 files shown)


src/Akeneo/Pim/Enrichment/Component/Product/Job/RemoveNonExistingProductValuesTasklet.php
```diff
@@ -0,0 +1,130 @@
...
+
+    /** @var EntityManagerClearerInterface */
+    private $entityManagerClearer;
+
+    /** @var int */
+    private $batchSize;
+
+    public function __construct(
+        GetProductAndProductModelIdentifiersWithValuesIgnoringLocaleAndScope $getProductAndProductModelIdentifiersWithValues,
+        AttributeRepositoryInterface $attributeRepository,
```
mmetayer, line 57: or PublicApi GetAttributes ?

src/Akeneo/Pim/Enrichment/Component/Product/Job/RemoveNonExistingProductValuesTasklet.php
```diff
@@ -0,0 +1,130 @@
...
+    private function checkFilters(array $filters): void
+    {
+        Assert::count($filters, 1);
+        $filter = current($filters);
+        Assert::keyExists($filter, 'field');
+        Assert::keyExists($filter, 'operator');
+        Assert::keyExists($filter, 'value');
+        Assert::same($filter['operator'], Operators::IN_LIST);
+        Assert::isArray($filter['value']);
+        Assert::allString($filter['value']);
```
mmetayer, line 128: seems really strict here, I don't understand why you don't directly pass attribute code + deleted attribute option codes to the job

nmarniesse: The values will work only with a list of string. Other cases are not handled in the ES query. As I cannot handle all cases (otherwise I would recreate a pqb) I prefer add this constraint.  

This format is the same format as the export or bulk actions jobs. This way we always create filter with same format, no matter if it's a manual query, a pqb, or anything else

src/Akeneo/Pim/Enrichment/Component/Product/Job/RemoveNonExistingProductValuesTasklet.php
nmarniesse: done

src/Akeneo/Pim/Enrichment/Component/Product/Job/RemoveNonExistingProductValuesTasklet.php
mmetayer: My point was: why even bother making a filter? you wouldn't have to do all these checks if your job parameters were
`$attributeCode` and `$optionCodes` instead of `$filter`... Unless I missed something

nmarniesse: I can do the change, for me it's the same

jmleroux:
check frozen

PR #12545, opened 2020-08-05
PIM-9112: Fix publish product if metric with null values

jmleroux:
I was fooled by the API return example and was trying with prices, but the problem is only with metric attributes.

I greped the provided dump and found that we have raw values like that: 

```
"weight": {"<all_channels>": {"<all_locales>": {"unit": null, "amount": null, "family": "Weight", "base_data": null, "base_unit": "KILOGRAM"}}}
```

The EmptyCleaner consider them as filled because of not null values in `family` and `base_unit` values.

This is a quick fix for invalid data coming from previous migrations and we must set a migration for 4.0 -> 5.0 to really clean the data and not have to handle it at product load.
There is a PR for that: https://github.com/akeneo/pim-community-dev/pull/11628

### DIFF
--- a/src/Akeneo/Pim/Enrichment/Component/Product/Factory/EmptyValuesCleaner.php
+++ b/src/Akeneo/Pim/Enrichment/Component/Product/Factory/EmptyValuesCleaner.php
@@ -56,6 +56,13 @@ final class EmptyValuesCleaner
         }
 
         if (is_array($data)) {
+            // @see PIM-9112: Fix publish product if metric with null values
+            // @todo to rework in 5.0:
+            // to clean with a migration as the data should not be null in database
+            // add the missing NonExistenMetricValueFilter to filter non existing metric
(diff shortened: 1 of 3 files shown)


src/Akeneo/Pim/Enrichment/Component/Product/Factory/EmptyValuesCleaner.php
```diff
@@ -56,6 +56,11 @@ private function isFilled($data): bool
         }
 
         if (is_array($data)) {
+            // @see PIM-9112: Fix publish product if metric with null values
+            // @todo remove in 5.0, the data must be cleaned at 4.0 -> 5.0 migration
```
ahocquard, line 60:
```suggestion
            // @todo to rework in 5.0:
            // to clean with a migration as the data should not be null in database 
            // add the missing NonExistenMetricValueFilter to filter non existing metric 
```

PR #12550, opened 2020-08-07
PIM-9391: Filter empty price and measurement values

tomglvng:

As of v4.0, empty values are not persisted anymore, and empty values are filtered. However the format of price and measurement attributes is a bit more complex than for other attributes (complex array), and the check performed to check if the value is empty did not take these formats into account.

In this PR we consider that a price or a measurement with a `null` amount is empty, and we filter these values.

`OK`: Done / Validated
`-`: Not needed


mmetayer:
The `circleci: pull_request_success` hook was not received, but the build passed: https://app.circleci.com/pipelines/github/akeneo/pim-enterprise-dev/22192/workflows/b28c3d12-030e-4510-8678-fb092ee0f541
so I'll merge anyway

PR #12567, opened 2020-08-12
PIM-9401: Fix ES filters with EMPTY operator

mmetayer:
In #11518, the usage of attribute filters' EMPTY operator was fixed, adding a should clause testing that the target attribute had to be part of either the 'attributes_for_this_level' list or in the 'attributes_of_ancestors' one. Unfortunately the fix was only applied in ProductAndProductModelWithSearchAggregator (i.e the product grid's query builder). 
This leads ES queries with an EMPTY filter returning wrong results in other PQBs, especially when there are models and variants involved (see examples in #11518) => this can cause some mass actions (including mass delete) to affect more entities than desired.
In this PR I apply the should clause directly in the attribute filters, the final query being:

- the ES attribute path must must not exist
AND
    - the 'attributes_for_this_level' must contain the attribute code
    OR
    - the 'attributes_of_ancestors' list must contain the attribute code

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Akeneo/Pim/Enrichment/Bundle/Elasticsearch/Filter/Attribute/DateFilter.php
+++ b/src/Akeneo/Pim/Enrichment/Bundle/Elasticsearch/Filter/Attribute/DateFilter.php
@@ -134,12 +134,28 @@ class DateFilter extends AbstractAttributeFilter implements AttributeFilterInter
                 ];
                 $this->searchQueryBuilder->addMustNot($existsClause);
 
-                $familyExistsClause = [
-                    'exists' => ['field' => 'family.code']
+                $attributeInEntityClauses = [
+                    [
+                        'terms' => [
+                            self::ATTRIBUTES_FOR_THIS_LEVEL_ES_ID => [$attribute->getCode()],
+                        ],
+                    ],
+                    [
+                        'terms' => [
+                            self::ATTRIBUTES_OF_ANCESTORS_ES_ID => [$attribute->getCode()],
+                        ],
+                    ]
                 ];
-                $this->searchQueryBuilder->addFilter($familyExistsClause);
-
+                $this->searchQueryBuilder->addFilter(
+                    [
+                        'bool' => [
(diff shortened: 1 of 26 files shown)


src/Akeneo/Pim/Enrichment/Bundle/Elasticsearch/Filter/Attribute/DateFilter.php
```diff
@@ -134,12 +134,28 @@ public function addAttributeFilter(
                 ];
                 $this->searchQueryBuilder->addMustNot($existsClause);
 
-                $familyExistsClause = [
```
mmetayer, line 137: the 'family exists' filter clause is already added in the `Akeneo\Pim\Enrichment\Component\Product\Query\AbstractEntityWithValuesQueryBuilder::addAttributeFilter()` method. I guess it could even be removed as the new condition is enough to filter products without family from the results

tests/back/Pim/Enrichment/Integration/PQB/Filter/Attribute/EmptyFilterWithProductsAndProductModelsIntegration.php
```diff
@@ -0,0 +1,244 @@
+<?php
+
+declare(strict_types=1);
+
+namespace back\Pim\Enrichment\Integration\PQB\Filter\Attribute;
```
mmetayer, line 5: wrong namespace

tests/back/Pim/Enrichment/Integration/PQB/Filter/Attribute/Media/LocalizableFilterIntegration.php
```diff
@@ -21,16 +21,9 @@ protected function setUp(): void
     {
         parent::setUp();
 
-        $this->createAttribute([
```
mmetayer, line 24: the 3 following tests were false positive

PR #12593, opened 2020-08-18
Fetch labels of options

ahocquard:

This is a POC to fetch efficiently labels of options in the API.
The format for simple select and multi select in the output is :
```
                    "collection": [
                        {
                            "locale": null,
                            "scope": null,
                            "data": {
                                "winter_2016": {
                                    "de_DE": "Winter 2016",
                                    "en_US": "Winter 2016",
                                    "fr_FR": "Hiver 2016"
                                }
                            }
                        }
                    ],
```
## Not done

Add a query parameters in the API to activate/deactivate this feature.

## Performance concern

This implementation fetches only the needed options. It adds only one request, so the IO complexity is reduce to the minimal.

This implementation iterates over all values of all the returned products to get the option values. Then, it executes the request and re-iterates over all values to replace with labels.

This is not ideal but simple. To improve that, the read value collection could be indexed by attribute type. It means we could get only the needed value of a given type, without iterating over the whole collection of values. However, it means that you have to index everything as soon as you create a read value collection, which is costly (index is good at reading, not at writing). Maybe it would not be so efficient.

Keep in mind that iterating over 100 products x 500 values (reference catalog) takes more than 30 ms. Iterating over it 2 times more than 60ms.
So, if you have this logic for asset or some other attributes, it can impact the performance. Benchmarking and documenting the impact is important in my opinion.

## Implementation choices

I decided to do a dedicated "Query\GetConnectorProducts" implementation. This is simple for now as it uses decoration pattern, but not future proof if you want to do it for asset (=EE). 

I have no time to do it, but this is a first draft in term of design.

When you have more than 2 layers (filter with permission, option codes as label, assets with labels, etc) in different repos, with activable/desactivable feature, registry pattern would be more adapted.
As it is easy to switch from decoration pattern to registry pattern, you can start with this implen and switch later on depending of the needs. 

## Known limit and bugs

The case sensitivity of the option code is important. You can have `Red` in the value but `red` in database. You should lowercase `value->getData()` when accessing the data in `$optionLabels` array built from the data in the database (if you test it you will understand).

I would test also the sql query to get option labels with 0 labels for an option.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Akeneo/Pim/Enrichment/Bundle/Storage/Sql/Connector/SqlGetConnectorProductsWithOptionLabels.php
+++ b/src/Akeneo/Pim/Enrichment/Bundle/Storage/Sql/Connector/SqlGetConnectorProductsWithOptionLabels.php
@@ -65,7 +65,8 @@ class SqlGetConnectorProductsWithOptionLabels implements Query\GetConnectorProdu
         }, $connectorProducts);
     }
 
-    /***
+    /**
+     * @param array $connectorProducts
      * @return array [['attribute_code', 'option_code']]
      */
     private function getOptionCodes(array $connectorProducts): array
@@ -133,7 +134,6 @@ class SqlGetConnectorProductsWithOptionLabels implements Query\GetConnectorProdu
             SELECT * FROM aggregated_attributes
             ;
         SQL;
-
 
         $row = $this->connection->executeQuery(
             sprintf($query, implode(',', $queryStringParams)),
--- a/src/Akeneo/Pim/Enrichment/Bundle/Storage/Sql/Connector/SqlGetConnectorProductsWithOptionLabels.php
+++ b/src/Akeneo/Pim/Enrichment/Bundle/Storage/Sql/Connector/SqlGetConnectorProductsWithOptionLabels.php
@@ -1,0 +1,146 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Akeneo\Pim\Enrichment\Bundle\Storage\Sql\Connector;
+
+use Akeneo\Pim\Enrichment\Component\Product\Connector\ReadModel\ConnectorProduct;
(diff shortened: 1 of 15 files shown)


src/Akeneo/Pim/Enrichment/Bundle/Storage/Sql/Connector/SqlGetConnectorProductsWithOptionLabels.php
```diff
@@ -0,0 +1,139 @@
...
+    private function getOptionWithLabels(array $optionCodes): array
+    {
+        if (empty($optionCodes)) {
+            return [];
+        }
+
+        $queryParams = array_fill(0, count($optionCodes), '(?, ?)');
+
+        $query = <<<SQL
+            WITH option_values AS (
```
ahocquard, line 103: if there is no label, this query does not return the option code

src/Akeneo/Pim/Enrichment/Component/Product/Connector/ReadModel/ConnectorProduct.php
```diff
@@ -178,6 +182,54 @@ public function addMetadata(string $key, $value): ConnectorProduct
...
+     * @param array $optionLabels array of all the labels of options, indexed by option code
+     *                            ['option_code' => ['en_US' => 'translation']
+     */
+    public function transformOptionWithLabels(array $optionLabels): ConnectorProduct
+    {
+        $values = $this->values->map(function (ValueInterface $value) use ($optionLabels) {
+            if ($value instanceof OptionValue) {
+                return new OptionValueWithLabels(
+                    $value->getAttributeCode(),
+                    [$value->getData() => $optionLabels[$value->getAttributeCode()][$value->getData()] ?? []],
```
ahocquard, line 195: if there is no label in the option code, as the query does not return the option in that case, there is `?? []` to handle that

PR #12717, opened 2020-09-11
RAC-226: Add Badge component

ValentinMumble:

Badge component, see: https://www.notion.so/akeneo/Badges-dbd7c5976558406c8d6384e3ee83bf25#4ff092b1f0234cdba6b9d57e017059a2

`OK`: Done / Validated
`-`: Not needed


akeneo-design-system/src/theme/theme.ts
```diff
@@ -66,12 +66,7 @@ type Theme = {
   color: Color;
 };
 
-enum Level {
-  Primary = 'primary',
-  Secondary = 'secondary',
-  Tertiary = 'tertiary',
-  Danger = 'danger',
-}
+type Level = 'primary' | 'secondary' | 'tertiary' | 'danger';
```
ValentinMumble, line 69: Using an `enum` will force an implicit cast from a string prop and causes weird behaviour when using it as an optional prop in `ArgsTable` in Stories. It also simplifies a lot the code, no need for the `getLevelColor` switch. My 2 cents, let me know what you think 😗

akeneo-design-system/src/components/Badge/Badge.stories.mdx
```diff
@@ -0,0 +1,64 @@
+import {Meta, Story, ArgsTable, Canvas} from '@storybook/addon-docs/blocks';
+import {Container} from 'storybook/container';
+import {Badge} from './Badge.tsx';
+
+<Meta
+  title="Components/Badge"
+  component={Badge}
+  argTypes={{
+    level: {type: {control: {type: 'select', options: ['primary', 'secondary', 'tertiary', 'danger']}}},
+    children: {table: {disable: true}},
```
ValentinMumble, line 10: To hide the `children` prop from the Story args table
image

akeneo-design-system/src/components/Checkbox/Checkbox.tsx
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
 import styled, {css} from 'styled-components';
-import {AkeneoThemedProps, getColor} from 'theme';
+import {AkeneoThemedProps, getColor} from '../../theme';
```
ValentinMumble, line 3: Had to update this for `Checkbox.unit.tsx` to pass 🤷

akeneo-design-system/tsconfig.json
```diff
@@ -26,6 +26,13 @@
...
     "target": "es6",
     "baseUrl": "./src",
-    "types": ["puppeteer", "jest-environment-puppeteer", "expect-puppeteer"]
+    "types": [
+      "puppeteer",
+      "jest-environment-puppeteer",
+      "expect-puppeteer",
+      "node",
+      "jest",
+      "@testing-library/jest-dom"
```
ValentinMumble, line 35: This fixes the `[ts] Property 'toHaveTextContent' does not exist on type 'Matchers<void>'.
any` issue we had on Github actions with React testing library.

akeneo-design-system/src/components/Badge/Badge.tsx
```diff
@@ -0,0 +1,37 @@
...
+  level?: Level;
+
+  /**
+   * Children of the Badge
+   */
+  children?: ReactNode;
+};
+
+const Badge = ({level = 'primary', children}: BadgeProps) => {
+  return <BadgeContainer level={level}>{children}</BadgeContainer>;
```
SamirBoulil, line 34: Maybe that's not part of this PR, but shouldn't this component accept a "level" and "text" props ? and not just any children ?

akeneo-design-system/src/components/Badge/Badge.stories.mdx
```diff
@@ -0,0 +1,64 @@
...
+<ArgsTable story="Standard" />
+
+## Variation on level
+
+<Canvas>
+  <Story name="Level">
+    {args => {
+      return (
+        <Container>
+          <Badge {...args} level="primary">
```
SamirBoulil, line 48: Having the badges inline could be clearer (do it only if you want to)

akeneo-design-system/src/theme/theme.ts
SamirBoulil: I like the fact that when a developper wants to set the level for the component he can rely on the Level enum Level.Primary etc.. instead of writing a string. On the other hand, TSC should be able to tell the developper if the level is not one of these too.
So I have mixed feelings about this.
Let's either way is fine by me.

SamirBoulil (approved):

akeneo-design-system/src/components/Badge/Badge.stories.mdx
```diff
@@ -0,0 +1,64 @@
+import {Meta, Story, ArgsTable, Canvas} from '@storybook/addon-docs/blocks';
```
SamirBoulil, line 1: 🔼 is it normal that the visual tests have been updated in this PR ? Isn't it the job of the github action to do this ?

SamirBoulil:
Are visual tests missing ?

akeneo-design-system/src/theme/theme.ts
ValentinMumble: The Level type is not string, it's only the predefined ones, so you can rely on autocompletion when you type the prop `level="`, I find the DX much nicer than having to import the `enum` then type `Level.xxx` in the prop

akeneo-design-system/src/components/Badge/Badge.stories.mdx
ValentinMumble: I merged the PR created by Github actions in this PR, I assumed this was the workflow?

akeneo-design-system/src/components/Badge/Badge.stories.mdx
ValentinMumble: What do you mean by inline? They look like this on the Story: 
image
If you talk about the code, Prettier did this

StevenVAIDIE (approved):

akeneo-design-system/src/components/Badge/Badge.tsx
ValentinMumble: Yes you're right, I don't think we can have another type of element inside a Badge, I'll make the change 👍

akeneo-design-system/src/components/Badge/Badge.stories.mdx
SamirBoulil: I was talking about the code, keep prettier's version then :)

akeneo-design-system/src/components/Badge/Badge.stories.mdx
SamirBoulil: ah yes! okay then!

akeneo-design-system/src/theme/theme.ts
ValentinMumble: That's how they do it in Material-UI https://github.com/mui-org/material-ui/blob/next/packages/material-ui/src/Button/Button.d.ts#L84

PR #12727, opened 2020-09-11
PIM-9443: Do not cache extensions.json

jmleroux:
We are loading extensions dynamically at page load and the `extensions.json` caused cache issues because it's "versioned" like other javascript assets.
By adding a random query string parameter, we will force it to be refreshed.
Note on perfs: with this query string, it is not cached anymore. But it's not a real problem because it's loaded only on HTML page refresh when the rest of the PIM is mainly a SPA. It will mainly be on login, so not very often for the same user.

### DIFF
--- a/CHANGELOG-4.0.md
+++ b/CHANGELOG-4.0.md
@@ -1,9 +1,5 @@
 # 4.0.x
 
-## Bug fixes
-
-- PIM-9443: Do not cache extensions.json
-
 # 4.0.54 (2020-08-28)
 
 ## Improvement
--- a/CHANGELOG-4.0.md
+++ b/CHANGELOG-4.0.md
@@ -1,5 +1,9 @@
 # 4.0.x
 
+## Bug fixes
+
+- PIM-9443: Do not cache extensions.json
+
 # 4.0.54 (2020-08-28)
 
 ## Improvement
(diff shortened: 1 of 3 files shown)


CHANGELOG-4.0.md
```diff
@@ -1,5 +1,9 @@
 # 4.0.x
 
+## Bug fixes
+
+- PIM-9443: Do not cache extensions.json
+
```
olivierpontier, line 6: Should we move it in the CHANGELOG.md file?

src/Akeneo/Platform/Bundle/UIBundle/Resources/public/js/form/config-provider.js
```diff
@@ -27,7 +27,7 @@ define(
         const loadConfig = function () {
             if (null === promise) {
                 promise = $.when(
-                    $.get('/js/extensions.json'),
+                    $.get('/js/extensions.json', {version: Math.random().toString(36).substring(7)}),
```
mmetayer, line 30: Can't we use the current version as query string? it would still allow caching the file, but it would be refreshed when a new version is out

jmleroux: That's exactly what i want, but dunno where to find it

jmleroux: I won't make an async call (with network exchange, Symfony bootstrap in memory, etc.) just to get the version and avoid to serve a little static file too often. If the version is already available in JS context, we will be able to imrove this PR later.

PR #12776, opened 2020-09-16
BH-46 Avoid polluting test migration output

BitOne:

The rollback migrations are not implemented in the application. So we test that they fail with an error. But the error is displayed in the output of the test, which can be very confusing.
And the CommandLauncher service is not able to redirect the error output. So the trait doesn't use it anymore. 
The PR allows as well the possibility to run the migration tests locally through the makefile 

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/upgrades/test_schema/ExecuteMigrationTrait.php
+++ b/upgrades/test_schema/ExecuteMigrationTrait.php
@@ -24,13 +24,13 @@ trait ExecuteMigrationTrait
         $pathFinder = new PhpExecutableFinder();
         $phpCommand = $pathFinder->find();
 
-        $rootDir = $this->getParameter('kernel.root_dir');
+        $rootDir = $this->getParameter('kernel.project_dir');
 
         $output = [];
         $status = null;
 
         exec(
-            sprintf('%s %s/../bin/console doctrine:migrations:execute %s --down -n 2>&1', $phpCommand, $rootDir, $migrationLabel),
+            sprintf('%s %s/bin/console doctrine:migrations:execute %s --down -n 2>&1', $phpCommand, $rootDir, $migrationLabel),
             $output,
             $status
         );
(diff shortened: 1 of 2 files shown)


upgrades/test_schema/ExecuteMigrationTrait.php
```diff
@@ -15,26 +15,33 @@
...
+        $pathFinder = new PhpExecutableFinder();
+        $phpCommand = $pathFinder->find();
+
+        $rootDir = $this->getParameter('kernel.root_dir');
+
+        $output = [];
+        $status = null;
+
+        exec(
+            sprintf('%s %s/../bin/console doctrine:migrations:execute %s --down -n 2>&1', $phpCommand, $rootDir, $migrationLabel),
```
tseho, line 33: Should the env be specified ?

The `CommandLauncher` class was explicitly using one,
see https://github.com/akeneo/pim-community-dev/blob/fd15a48460cca0a030ea1e65449a963763ae84d3/src/Akeneo/Tool/Component/Console/CommandLauncher.php#L63
https://github.com/akeneo/pim-community-dev/blob/fd15a48460cca0a030ea1e65449a963763ae84d3/config/services/test/test_services.yml#L22-L27

BitOne: Now the env is specified through env var. It was even planned at some point to remove the "--env" option from the bin/console command.

So the CommandLauncher should be fixed :)

franckbrouard (approved):

upgrades/test_schema/ExecuteMigrationTrait.php
```diff
@@ -15,26 +15,33 @@
...
+    abstract protected function getParameter(string $parameter);
 
     private function reExecuteMigration(string $migrationLabel): void
     {
-        $resultDown = $this->getCommandLauncher()->executeForeground(
-            sprintf('doctrine:migrations:execute %s --down -n', $migrationLabel)
+        $pathFinder = new PhpExecutableFinder();
+        $phpCommand = $pathFinder->find();
+
+        $rootDir = $this->getParameter('kernel.root_dir');
```
mmetayer, line 27: `kernel.root_dir` is deprecated, you should use `kernel.project_dir` instead

mmetayer (approved):

upgrades/test_schema/ExecuteMigrationTrait.php
BitOne: Thanks @mmetayer! I will fix it.

PR #12791, opened 2020-09-21
PM2020-19: Add Product Typology Filter

pierallard:

Adds a new filter in product grid

| Added Specs                       | :+1: 
| Added legacy Behats               | :-1: 
| Added acceptance tests            | :-1: 
| Added integration tests           | :-1: 
| Changelog updated                 | :-1: 
| Review and 2 GTM                  | :-1: 
| Micro Demo to the PO (Story only) | :-1: 

### DIFF
--- a/src/Oro/Bundle/PimFilterBundle/Filter/Product/ProductTypologyFilter.php
+++ b/src/Oro/Bundle/PimFilterBundle/Filter/Product/ProductTypologyFilter.php
@@ -1,0 +1,36 @@
+<?php
+
+namespace Oro\Bundle\PimFilterBundle\Filter\Product;
+
+use Akeneo\Pim\Enrichment\Component\Product\Query\Filter\Operators;
+use Oro\Bundle\FilterBundle\Datasource\FilterDatasourceAdapterInterface;
+use Oro\Bundle\FilterBundle\Filter\ChoiceFilter;
+
+class ProductTypologyFilter extends ChoiceFilter
+{
+    private const PRODUCT_TYPOLOGY_SIMPLE = 'simple';
+    private const PRODUCT_TYPOLOGY_VARIANT = 'variant';
+
+    /**
+     * {@inheritdoc}
+     */
+    public function apply(FilterDatasourceAdapterInterface $ds, $data)
+    {
+        $data = $this->parseData($data);
+        if (!$data) {
+            return false;
+        }
+
+        switch (current($data['value'])) {
+            case self::PRODUCT_TYPOLOGY_SIMPLE:
+                $this->util->applyFilter($ds, 'family_variant', Operators::IS_EMPTY, null);
+                break;
+            case self::PRODUCT_TYPOLOGY_VARIANT:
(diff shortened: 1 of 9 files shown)


src/Oro/Bundle/PimFilterBundle/Filter/Product/ProductTypologyFilter.php
```diff
@@ -0,0 +1,36 @@
...
+    public function apply(FilterDatasourceAdapterInterface $ds, $data)
+    {
+        $data = $this->parseData($data);
+        if (!$data) {
+            return false;
+        }
+
+        switch (current($data['value'])) {
+            case self::PRODUCT_TYPOLOGY_SIMPLE:
+                $this->util->applyFilter($ds, 'family_variant', Operators::IS_EMPTY, null);
```
nmarniesse, line 26: Should we also filter on product only? This filter removes product models?

nmarniesse (approved):
Added an integration test should be great IMO (see LabelOrIdentifierFilterIntegration.php for example)

src/Oro/Bundle/PimFilterBundle/Filter/Product/ProductTypologyFilter.php
pierallard: Having "no family variant" means you only have simple products, not variant nor product models.

nmarniesse: :+1:

nmarniesse: So having a family variant means variant or product model? It's for the `PRODUCT_TYPOLOGY_VARIANT` case just below

pierallard: Exaclty

nmarniesse: This is not a problem that the "variant" filter returns variant AND product model? We don't have to filter on product too?

mmetayer (approved):

src/Oro/Bundle/PimFilterBundle/Filter/Product/ProductTypologyFilter.php
pierallard: Variant filter returns variant AND product models ; but the "aggregator" removes the variant if we don't have conditions on the current level.
The "filter on product" is the "ungroup/group" feature, and it's another PR.

nmarniesse: Ok I understand now. If we filter on "variant", it does not return only variant products :roll_eyes: ... The naming is very confusing, but maybe it's just me. In fact it's a "simple product" filter with true/false value (I'm not totally convinced by this naming too)

mmetayer: probably smth we'll discuss tomorrow (ubiquitous language)

PR #12799, opened 2020-09-22
Mount inidb.d in mysql container instead of specific script

tseho:
**Description**

Following https://github.com/akeneo/pim-community-dev/pull/12392, there was an issue with the standard edition where `create_db_test.sh` was not copied and the mysql container could not boot properly.

The solution is to ensure the directory `./docker/initdb.d` exists on the host and mount this directory in the mysql entrypoint.
This way, it will mount an empty directory on standard and the mysql container can boot.

Also, following feedback from @ahocquard & @jmleroux, I'm removing the `make database-test` because it was confusing.

**Definition Of Done**

| CE Pull Request                   | -


std-build/install-required-files.sh
```diff
@@ -19,7 +19,8 @@ mkdir -p $STANDARD_DISTRIB_DIR/src \
          $STANDARD_DISTRIB_DIR/public \
          $STANDARD_DISTRIB_DIR/config/packages/dev \
          $STANDARD_DISTRIB_DIR/config/services \
-         $STANDARD_DISTRIB_DIR/docker
+         $STANDARD_DISTRIB_DIR/docker \
```
ahocquard, line 22: useless as it's `mkdir -p`

ahocquard (approved):

std-build/install-required-files.sh
BitOne: Technically useless, but I would keep it, as the intention is not the same ;)

BitOne (approved):

std-build/install-required-files.sh
ahocquard: You mean the intention is not the same as two lines above? :trollface:

SamirBoulil (approved):

jmleroux (approved):

PR #12880, opened 2020-09-29
GITHUB-APD-227: backport communication panel

ahocquard:
## Goal

The goal of this PR is to backport the PIM communication panel from master.

## BC break

It is not allowed to do BC breaks in a patch:
- migrations should be considered optional and not executed
- modification of composer.json in standard CE/EE is a BC
- modification of package.json in standard CE/EE is a BC
- adding a bundle is a BC

Unfortunately, I had to add 4 required dependencies in the `package.json` of CE standard (and 2 in standard EE).
To handle it, during the installation of the assets, it adds these dependencies in the package.json.

As this installation of the assets is triggered after a composer update and before a yarn update, it should work.

Note: in EE standard, the command to install the assets is the Symfony one, and not the PIM one. So, to add the missing dependencies during a composer update, I execute it in a function of "copyUpgradeFile". It's dirty, not related at all, but it's the only way I found to do it. 

### make it working  for 3.2:

- adapt to Symfony 3 directory tree (routing)
- add missing pim parameter for helpcenter url
- change `public` as `web` for the frontend
- add an installer part to copy the front in `web/bundles`, as we can't use workspace
- copy past common frontend components used it master directory in the communication panel bundle, because they don't exist in 3.2
- add resolver of namespace in `frontend/wepack/requirejs-utils.js`

```
            '@akeneo-pim-community/communication-channel': path.resolve(
                baseDir,
                'web/bundles/akeneocommunicationchannel-react/index.ts'
            ),
            '@akeneo-pim-community/legacy-bridge': path.resolve(
                baseDir,
                'web/bundles/akeneocommunicationchannel-react/legacy-bridge/src/index.ts'
            ),
            '@akeneo-pim-community/shared': path.resolve(
                baseDir,
                'web/bundles/akeneocommunicationchannel-react/shared/src/index.ts'
            ),
```

- add missing css property `@AknPanelWidth` not defined in 3.2

- I changed all imports due to this error:
```
ERROR in /srv/pim/web/bundles/akeneocommunicationchannel-react/components/panel/Panel.tsx
./web/bundles/akeneocommunicationchannel-react/components/panel/Panel.tsx
[tsl] ERROR in /srv/pim/web/bundles/akeneocommunicationchannel-react/components/panel/Panel.tsx(1,8)
      TS1192: Module '"/srv/pim/node_modules/@types/react/index"' has no default export.
```

From `import React from 'react';` to `import * as React from 'react';`

### Avoid BC break the most I can

It's not possible to declare a new bundle.
It's not possible to execute a migration in a patch.

Therefore:
- load dependency injection from UIBundle instead of CommunicationChannelBundle (no need to move PHP classes)
- move routing in UIBundle
- move translations, form extension and requirejs into UIBundle
- create the needed table on the fly if it does not exist

### Fix stuff that were not working properly

- Lint in front, as it adds a library (even if it's in required-dev, I prefer to avoid it)
- Coupling back because php_cs has an issue with a PSR 4 namespace: I removed the rule, not important in 3.2 

- Remove PSR-4 rule (so rename all namespaces...) due to weird mandatory declaration of this rule in standard CE (no explanation): it was causing a BC

- Use of file_get_content instead of Guzzle to fetch announcements, because Guzzle dependency is not required in standard CE...


SamirBoulil (approved):

mmetayer (approved):

ahocquard:
Green CI, I just pushed forced due to a changelog rebase.

PR #12891, opened 2020-09-30
CXP-412: Update the webhook event format

tomglvng:
TODO: add "**author**" and "**pim_source**".

```json
{
	"action": "product.created",
	"author": "julia",
	"event_id": "6ad821d8-2468-4632-89d2-92e442c72313",
	"event_date": "2020-01-01T00:00:00+00:00",
	"pim_source": "staging.akeneo.com",
	"data": {
		"resource": {
			"identifier": "blue_shirt"
		}
	}
}
```

### DIFF
--- a/src/Akeneo/Connectivity/Connection/back/Application/Webhook/WebhookEventBuilder.php
+++ b/src/Akeneo/Connectivity/Connection/back/Application/Webhook/WebhookEventBuilder.php
@@ -27,12 +27,17 @@ class WebhookEventBuilder
         $this->builders = $builders;
     }
 
-    public function build(BusinessEventInterface $businessEvent): WebhookEvent
+    /**
+     * @param array<mixed> $context
+     */
+    public function build(BusinessEventInterface $businessEvent, array $context = []): WebhookEvent
     {
         return new WebhookEvent(
             $businessEvent->name(),
             $businessEvent->uuid(),
             date(\DateTimeInterface::ATOM, $businessEvent->timestamp()),
+            $businessEvent->author(),
+            array_key_exists('pim_source', $context) ? $context['pim_source'] : '',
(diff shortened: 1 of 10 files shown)


src/Akeneo/Connectivity/Connection/back/Application/Webhook/WebhookEventBuilder.php
```diff
@@ -27,12 +27,17 @@ public function __construct(iterable $builders)
...
+     * @param array<mixed> $context
+     */
+    public function build(BusinessEventInterface $businessEvent, array $context = []): WebhookEvent
     {
         return new WebhookEvent(
             $businessEvent->name(),
             $businessEvent->uuid(),
             date(\DateTimeInterface::ATOM, $businessEvent->timestamp()),
+            $businessEvent->author(),
+            array_key_exists('pim_source', $context) ? $context['pim_source'] : '',
```
pchasle, line 40: You can use the null coalescing operator in this case :)
```suggestion
            $context['pim_source'] ?? ''
```
I wonder if we should not raise an error if 'pim_source' is not defined instead?!
Like `InvalidArgumentException('Context property 'pim_source' is mandatory');`
It's probably better to break early than having a "silent" error

pchasle (approved):

src/Akeneo/Connectivity/Connection/back/Application/Webhook/WebhookEventBuilder.php
tomglvng: After our discussion, I added the Exception. See https://github.com/akeneo/pim-community-dev/pull/12891/commits/f6a8023ef310b282b3af774edf0f19e4d21ee013

pierrejolly (approved):

PR #12944, opened 2020-10-06
PIM-9494: Fix the performances of attribute-select-filter on long lists of AttributeOptions

tseho:

When a filter of type "attribute select", that let you pick attribute options, is used in the datagrid, there was performances issues when there is a lot of attribute options (eg: 5000).

This is because the filter in front tries to be smart and only fetch the selected values:

https://github.com/akeneo/pim-community-dev/blob/162def2dabfa6f4b62ab5c9deee6a3a781c83bf1/src/Oro/Bundle/PimDataGridBundle/Resources/public/js/datafilter/filter/select2-choice-filter.js#L234-L245

For doing so, it launch a search limited to some `identifiers`.

It was working on most repositories, like `FamilySearchableRepository`:
https://github.com/akeneo/pim-community-dev/blob/162def2dabfa6f4b62ab5c9deee6a3a781c83bf1/src/Akeneo/Pim/Structure/Bundle/Doctrine/ORM/Repository/InternalApi/FamilySearchableRepository.php#L64-L67

But in `AttributeOptionRepository`, the key `ids` is expected, not `identifiers`.
https://github.com/akeneo/pim-community-dev/blob/162def2dabfa6f4b62ab5c9deee6a3a781c83bf1/src/Akeneo/Pim/Structure/Bundle/Doctrine/ORM/Repository/AttributeOptionRepository.php#L57-L63
By using the wrong key, the list was not limited and the controller was returning ALL the rows (5000 in our case) and select2 was not able to handle it.

To avoid any BC break on things I've not seen, I choose to have `AttributeOptionRepository` support both keys.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,7 @@
 - PIM-9478: Allow the modification of the identifier on a variant product
 - PIM-9481: Fix the list of product models when trying to get them by family variant
 - GITHUB-12899: Fix error shown when importing product models with the same code
+- PIM-9494: Fix the performances of attribute-select-filter on long lists of AttributeOptions
 
 ## New features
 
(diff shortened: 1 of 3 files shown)


jmleroux (approved):

nmarniesse (approved):

juliensnz (approved):

mmetayer (approved):
:+1: for the great PR description

PR #12973, opened 2020-10-08
DAPI-1283: Re-evaluate products impacted by attribute group activation updates

LaurentPetard:


AymericPlanche (approved):

rouliane (approved):
forUpdatesSince seems a bit weird, but I don't really have a better name :/

PR #12989, opened 2020-10-12
CXP-431: shorten the timeout while sending webhooks

pchasle:
- Added config parameters for the webhook client options
- Shortened the request timeout to 500ms

### DIFF
--- a/src/Akeneo/Connectivity/Connection/back/Infrastructure/Webhook/Client/GuzzleWebhookClient.php
+++ b/src/Akeneo/Connectivity/Connection/back/Infrastructure/Webhook/Client/GuzzleWebhookClient.php
@@ -32,11 +32,11 @@ class GuzzleWebhookClient implements WebhookClient
     /** @var LoggerInterface */
     private $logger;
 
-    /** @var array{concurrency: int, timeout: float} */
+    /** @var array{concurrency: ?int, timeout: ?float} */
     private $config;
 
     /**
-     * @param array{concurrency: int, timeout: float} $config
+     * @param array{concurrency: ?int, timeout: ?float} $config
      */
     public function __construct(
         ClientInterface $client,
@@ -76,9 +76,9 @@ class GuzzleWebhookClient implements WebhookClient
         };
 
         $pool = new Pool($this->client, $guzzleRequests(), [
-            'concurrency' => $this->config['concurrency'],
+            'concurrency' => $this->config['concurrency'] ?? null,
             'options' => [
-                'timeout' => $this->config['timeout']
+                'timeout' => $this->config['timeout'] ?? null
             ],
             'fulfilled' => function (Response $response, int $index) use (&$logContexts) {
                 $this->logger->info(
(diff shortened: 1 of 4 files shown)


src/Akeneo/Connectivity/Connection/back/Infrastructure/Webhook/Client/GuzzleWebhookClient.php
```diff
@@ -32,11 +32,22 @@ class GuzzleWebhookClient implements WebhookClient
...
+    private $config;
+
+    /**
+     * @param array{concurrency: int, timeout: float} $config
+     */
+    public function __construct(
+        ClientInterface $client,
+        EncoderInterface $encoder,
+        LoggerInterface $logger,
+        array $config
```
pierrejolly, line 45: IMO we should not inject the whole config array here, just the two parameters

pierrejolly (approved):
GTM after fix comment

src/Akeneo/Connectivity/Connection/back/Infrastructure/Webhook/Client/GuzzleWebhookClient.php
pchasle: Yes, it's what I do...

src/Akeneo/Connectivity/Connection/back/Infrastructure/Webhook/Client/GuzzleWebhookClient.php
```diff
@@ -65,9 +76,9 @@ public function bulkSend(iterable $webhookRequests): void
         };
 
         $pool = new Pool($this->client, $guzzleRequests(), [
-            'concurrency' => 5,
+            'concurrency' => $this->config['concurrency'],
```
willy-ahva, line 79: You must check that 'concurrency' and 'timeout' exist as they come from the DI.

src/Akeneo/Connectivity/Connection/back/Infrastructure/Webhook/Client/GuzzleWebhookClient.php
willy-ahva: Injecting an array as an option or config is a good way to avoid BC Breaks if we need other options later. I think we should keep it but add an option resolver.

src/Akeneo/Connectivity/Connection/back/tests/Unit/spec/Infrastructure/Webhook/Client/GuzzleWebhookClientSpec.php
```diff
@@ -43,7 +48,12 @@ public function it_sends_webhook_requests_in_bulk(): void
         $handlerStack = HandlerStack::create();
         $handlerStack->push($history);
 
-        $this->beConstructedWith(new Client(['handler' => $handlerStack]), new JsonEncoder(), new NullLogger());
+        $this->beConstructedWith(
+            new Client(['handler' => $handlerStack]),
+            new JsonEncoder(),
+            new NullLogger(),
+            ['concurrency' => 1, 'timeout' => 1]
```
willy-ahva, line 55: You should have a test with an empty array.

willy-ahva (approved):
GTM after fix

PR #12994, opened 2020-10-12
Removes translations files with wrong name

jmleroux:
Nelson pushed wrong translation files.

For the record, it's quite easy to remove them:

```bash
find src -regextype posix-extended -regex '.*messages\.[a-z]+\.yml$' | xargs rm
```


ValentinMumble (approved):

pierallard (approved):
BRULE

jmleroux:
Same for validators:

```bash
find src -regextype posix-extended -regex '.*validators\.[a-z]+\.yml$' | xargs rm
```

PR #12998, opened 2020-10-12
DAPI-1374: Index product key indicators in ES

rouliane:
**Description (for Contributor and Core Developer)**

**Definition Of Done (for Core Developer only)**

| Q                                 | A
| --------------------------------- | ---
| Added Specs                       | Todo
| Added legacy Behats               | Todo
| Added acceptance tests            | Todo
| Added integration tests           | Todo
| Changelog updated                 | Todo
| Review and 2 GTM                  | Todo
| Micro Demo to the PO (Story only) | Todo
| Migration script                  | -
| Tech Doc                          | -

`Todo`: Pending / Work in progress
`OK`: Done / Validated
`-`: Not needed

PR #13024, opened 2020-10-16
DAPI-1377: Transform evaluation results to numeric keys and values

LaurentPetard:
EE PR https://github.com/akeneo/pim-enterprise-dev/pull/9775

The goal is to persist only numeric keys and values for the JSON that represent the result of the evaluation of a criterion.
The codes of attributes, channels and locales are replaced by their ids. The properties and status are mapped with arbitrary numeric values.


AymericPlanche (approved):

olivierpontier (approved):

LaurentPetard:
Will need adaptations when https://github.com/akeneo/pim-community-dev/pull/12998 will be merged
Ideally we should be able to transform channels/locales, attributes and properties separately (for the key indicators, we don't need to transform the attributes)

PR #13050, opened 2020-10-20
DAPI-1377: Add foreign key to automatically delete product evaluations

LaurentPetard:
The migration script add the FK, and put all the evaluations to status "pending" to force the re-evaluation to save the results at the new format.

I used a temporary table instead of simply adding the FK on the existing tables. The new tables with the FK are populate with evaluations without result, and at status "pending". It allows to not have to delete the evaluations which the product doesn't exist.


upgrades/schema/Version_5_0_20201012085420_dqi_rework_evaluations_storage.php
```diff
@@ -0,0 +1,63 @@
...
+  INDEX status_index (status),
+  CONSTRAINT FK_dqi_product_model_criteria_evaluation FOREIGN KEY (product_id) REFERENCES pim_catalog_product_model (id) ON DELETE CASCADE
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_ci;
+SQL
+        );
+
+        $this->addSql(<<<SQL
+INSERT IGNORE INTO pim_data_quality_insights_product_criteria_evaluation (product_id, criterion_code, evaluated_at, status)
+SELECT evaluation_dep.product_id, evaluation_dep.criterion_code, evaluation_dep.evaluated_at, 'pending'
+FROM pim_data_quality_insights_product_criteria_evaluation_dep AS evaluation_dep;
```
AymericPlanche, line 46: where criterion_code != "title_formatting"

(find the real code that we deleted)
I'd this case for a test, I passed all the criterion of an existing customer to pending, there is still the criterion we removed in August, no need to keep it in the new table. 

Same for product models

AymericPlanche (approved):
Eager to have the test result of the migration on existing data ! :rocket:

upgrades/schema/Version_5_0_20201012085420_dqi_rework_evaluations_storage.php
AymericPlanche: criterion code = `consistency_text_title_formatting`

olivierpontier (approved):

PR #13068, opened 2020-10-21
Fix env conf for pubsub emulator

pchasle:


BitOne (approved):

AymericPlanche (approved):

.env
```diff
@@ -18,8 +18,8 @@ COMM_PANEL_API_URL=https://pim-comm-panel.akeneo.com
 FLAG_DATA_QUALITY_INSIGHTS_ENABLED=1
 SRNT_GOOGLE_APPLICATION_CREDENTIALS=
 GOOGLE_CLOUD_PROJECT=emulator-project
-PUBSUB_EMULATOR_HOST=pubsub-emulator:8085
-PUBSUB_TOPIC_BUSINESS_EVENT=dev-business-event
-PUBSUB_TOPIC_JOB_QUEUE=dev-job-queue
-PUBSUB_SUBSCRIPTION_WEBHOOK=dev-webhook
-PUBSUB_SUBSCRIPTION_JOB_QUEUE=dev-job-queue
+PUBSUB_EMULATOR_HOST=
```
mr-prud, line 21: I'm sorry, but PUBSUB_EMULATOR_HOST still set, with an empty string but still set, and as the documentation and the code is written :
https://github.com/googleapis/google-cloud-php-pubsub/blob/25e2b60c1c35b8417182d70ccbed06e788959b6b/src/PubSubClient.php#L150
and 
https://github.com/googleapis/google-cloud-php-pubsub/blob/25e2b60c1c35b8417182d70ccbed06e788959b6b/src/PubSubClient.php#L154

I'm not a PHP expert, yes I can't read the second line, empty, null true, false... could you check please.

mr-prud (commented):
please, answer my question

.env
mr-prud: Ok,
```
<?php

$foo="";
if((bool)$foo) {
    echo ("hello");
}
```

pchasle: Legitimate question; thank you for checking that :+1:

mmetayer (approved):

PR #13095, opened 2020-10-26
New attempt to fix a random failing legacy scenario

mmetayer:

Another attempt to fix a randomly failing legacy scenario

### DIFF
--- a/tests/legacy/features/user-management/role/create_edit_and_delete_role.feature
+++ b/tests/legacy/features/user-management/role/create_edit_and_delete_role.feature
@@ -20,12 +20,13 @@ Feature: Create and delete a user role
     When I save the role
     Then I should see the text "VeryDummyRole"
     When I am on the role index page
+    Then I should see roles "Administrator, VeryDummyRole"
     And I click on the "Delete" action of the row which contains "VeryDummyRole"
     Then I should see a confirm dialog with the following content:
       | title   | Confirm deletion                           |
       | content | Are you sure you want to delete this role? |
     When I confirm the deletion
-    Then I should not see the text "VeryDummyRole"
+    Then I should not see role "VeryDummyRole"
     When I click on the "Delete" action of the row which contains "Administrator"
     And I cancel the deletion
-    And I should see the text "Administrator"
+    And I should see role "Administrator"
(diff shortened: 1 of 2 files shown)


tests/legacy/features/user-management/role/create_edit_and_delete_role.feature
```diff
@@ -20,12 +20,13 @@ Feature: Create and delete a user role
...
     Then I should see the text "VeryDummyRole"
     When I am on the role index page
+    Then I should see roles "Administrator, VeryDummyRole"
     And I click on the "Delete" action of the row which contains "VeryDummyRole"
     Then I should see a confirm dialog with the following content:
       | title   | Confirm deletion                           |
       | content | Are you sure you want to delete this role? |
     When I confirm the deletion
-    Then I should not see the text "VeryDummyRole"
+    Then I should not see role "VeryDummyRole"
```
ahocquard, line 29: When deleting, it will refresh the page. As this assertion "I should not see" return true as soon it does not find the row (instead of spining at least 5 seconds), it will probably return true during the refresh of the page.

mmetayer: it internally calls the `\Context\Page\Base\Grid::getGrid()` method, which waits for the grid to be loaded and visible

ahocquard: But is the grid already displayed during deletion? Because it could assert on that previous display in that case.

mmetayer: nope, because after getting the grid (and waiting for it to be visible) it will store it into a NodeElement variable, and the `hasRow` method tires to get a `tbody > tr` from this variable (not from the browser anymore), and all that is nested in a `spin`, so for your scenario it would go like:

Browser: delete role
Test: I wait for the grid => previous grid displayed, stored into a variable
Test: the row is still present in the grid => the spin returns false
Browser: refresh the page
Test: I wait for the grid => the grid is not visible yet, the spin returns false
Browser: the page is refreshed, the new grid is displayed
Test: I wait for the grid => new grid stored into a variable
Test: the row is not present anymore => OK

mmetayer: I know local tests are not a proof, but the test passed 20 times in a row locally, while it failed 20 times in a row when I replaced `Then I should not see row "VeryDummyRole"` by `Then I should not see row "Administrator"`

pierallard (approved):

pierrejolly (approved):

PR #13102, opened 2020-10-26
PIM-9524: improve purge job command to avoid out of memory errors

nmarniesse:

Fix https://akeneo.atlassian.net/browse/PIM-9524

An out of memory error occurred when we have a lot of files to list. The error is thrown in the flysystem list function. Badly we cannot fix it efficiently because the `listContents` method builds and returns an array with all results: https://github.com/thephpleague/flysystem/blob/1.1.3/src/Adapter/Local.php#L288  
The flysystem v2 (not released yet) will return an iterator with a generator. So the problem will be fixed.  

For now, to try to fix the issue, instead of listing all files in the `.` directory recursively, I fetch all directories in level 1, then in level 2, finally in level 3 (no need to go further for this use case). And I use a generator to returns directories by group of 100.  
A memory error can still happen if there is too many files in a directory of level 2 so this solution is clearly not perfect.  

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Akeneo/Platform/Bundle/ImportExportBundle/Persistence/Filesystem/DeleteOrphanJobExecutionDirectories.php
+++ b/src/Akeneo/Platform/Bundle/ImportExportBundle/Persistence/Filesystem/DeleteOrphanJobExecutionDirectories.php
@@ -29,7 +29,7 @@ final class DeleteOrphanJobExecutionDirectories
 
     public function execute(): void
     {
-        $pathsByBatch = $this->getPathsByBatch();
+        $pathsByBatch = $this->getPathsAtLevel3ByBatch();
 
         foreach ($pathsByBatch as $paths) {
             $jobExecutionIds = $this->getJobExecutionIdsFromPaths($paths);
@@ -38,7 +38,7 @@ final class DeleteOrphanJobExecutionDirectories
         }
     }
 
-    private function getPathsByBatch(): \Iterator
+    private function getPathsAtLevel3ByBatch(): \Iterator
     {
         $paths = [];
         $firstLevelPaths = $this->archivistFilesystem->listPaths('.', false);
--- a/src/Akeneo/Platform/Bundle/ImportExportBundle/Persistence/Filesystem/DeleteOrphanJobExecutionDirectories.php
+++ b/src/Akeneo/Platform/Bundle/ImportExportBundle/Persistence/Filesystem/DeleteOrphanJobExecutionDirectories.php
@@ -54,14 +54,26 @@ final class DeleteOrphanJobExecutionDirectories
     private function getPathsAtLevel3ByBatch(): \Iterator
     {
         $paths = [];
-        $firstLevelPaths = $this->archivistFilesystem->listPaths('.', false);
-        foreach ($firstLevelPaths as $firstLevelPath) {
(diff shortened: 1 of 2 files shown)


jmleroux (approved):

StevenVAIDIE (approved):

src/Akeneo/Platform/Bundle/ImportExportBundle/Persistence/Filesystem/DeleteOrphanJobExecutionDirectories.php
```diff
@@ -27,18 +29,44 @@ public function __construct(Filesystem $archivistFilesystem, Connection $connect
 
     public function execute(): void
     {
-        $paths = $this->archivistFilesystem->listFiles('.', true);
+        $pathsByBatch = $this->getPathsAtLevel3ByBatch();
```
ahocquard, line 32: I would put the previous way in a private function, with a phpdoc comment to explain to use it once migrated to flysystem v2, eventually with a TIP ticket.

Otherwise, we will forget that it could be much simpler.

src/Akeneo/Platform/Bundle/ImportExportBundle/Persistence/Filesystem/DeleteOrphanJobExecutionDirectories.php
```diff
@@ -27,18 +29,44 @@ public function __construct(Filesystem $archivistFilesystem, Connection $connect
...
-        $existingJobExecutionIds = $this->getExistingJobExecutionIds($jobExecutionIds);
-        $this->deleteOrphanJobExecutionDirectories($paths, $existingJobExecutionIds);
+        foreach ($pathsByBatch as $paths) {
+            $jobExecutionIds = $this->getJobExecutionIdsFromPaths($paths);
+            $existingJobExecutionIds = $this->getExistingJobExecutionIds($jobExecutionIds);
+            $this->deleteOrphanJobExecutionDirectories($paths, $existingJobExecutionIds);
+        }
+    }
+
+    private function getPathsAtLevel3ByBatch(): \Iterator
```
ahocquard, line 41: I would add phpdoc to explain why

src/Akeneo/Platform/Bundle/ImportExportBundle/Persistence/Filesystem/DeleteOrphanJobExecutionDirectories.php
```diff
@@ -27,18 +29,44 @@ public function __construct(Filesystem $archivistFilesystem, Connection $connect
...
+            yield $paths;
+        }
     }
 
     private function getJobExecutionIdsFromPaths(array $paths): array
     {
         $jobExecutionIds = [];
         foreach ($paths as $path) {
-            $directories = explode(DIRECTORY_SEPARATOR, $path['dirname']);
+            $directories = explode(DIRECTORY_SEPARATOR, $path);
```
ahocquard, line 69: not possible to keep return directly it in the yield?

ahocquard (approved):

ahocquard:
Some comments for the future of this class.

nmarniesse:
unfreeze

PR #13225, opened 2020-11-10
PIM-9548: Mitigate deadlock issues on category API 

tseho:

There is a possible deadlock when creating several new categories at the same time.
Gedmo/Tree execute queries like `UPDATE pim_catalog_category SET lft = lft + 2 WHERE lft >= 2 AND root = ?`
producing GAP Lock issues.

The entity manager is not capable to recover from such error.
By locking with an external lock beforehand, we can mitigate this deadlock issue.

For the external lock, I've used https://symfony.com/doc/current/components/lock.html with a PDOStore (the lock is stored in the database, shared between instances)


ahocquard:
To add some details:

This strategy allows to have deadlock due to concurrent writing stuff on categories. There will be X locks (exclusive locks) only by one concurrent process, and so it avoids deadlocks (which imply 2 concurrent processes at least).

The limit of this solution is that S locks (shared locks) could be applied by a concurrent process, for reading stuff for example.
As an X lock cannot be granted on a row if a S lock holds the row.

What could happen then is:

X lock granted on category A  - transaction 1
S lock granted on category B - transaction 2
X lock pending for category B - transaction 1
S lock pending for category A - transaction 2

=> deadlock

It's not possible to avoid it from the application code. There are many other type of locks. The best solution is to retry the transaction, which is not possible easily due to ORM.

That's why  the provided solution mitigates the issue but does not prevent to occur if there is a mix a of reading/writing operations on categories.

src/Akeneo/Tool/Bundle/ClassificationBundle/Resources/config/savers.yml
```diff
@@ -0,0 +1,23 @@
+services:
+    pim_catalog.saver.category:
```
ahocquard, line 2: the declaration this service is part of the enrichment context

src/Akeneo/Pim/Enrichment/Bundle/Resources/config/savers.yml
```diff
@@ -52,13 +52,6 @@ services:
             - '@event_dispatcher'
             - '%pim_catalog.model.association_type.interface%'
 
-    pim_catalog.saver.category:
```
ahocquard, line 55: Keep it here, it's on purpose it is in the enrichment context (see the context map)

src/Akeneo/Tool/Bundle/ClassificationBundle/Resources/config/savers.yml
```diff
@@ -0,0 +1,23 @@
...
+        class: Akeneo\Tool\Bundle\ClassificationBundle\Storage\Orm\CategorySaver
+        arguments:
+            - '@doctrine.orm.entity_manager'
+            - '@event_dispatcher'
+            - '%pim_catalog.model.category.interface%'
+            - '@akeneo_classification.lock.factory'
+
+    akeneo_classification.lock.store:
```
ahocquard, line 10: These services should be part of the enrichment context (something like `akeneo.enrichment....`).
If a new feature uses classification (like the previous asset feature in 2.x), it should redeclare its own service for decoupling reason.

src/Akeneo/Tool/Bundle/ClassificationBundle/Storage/Orm/CategorySaver.php
```diff
@@ -0,0 +1,78 @@
...
+
+use Akeneo\Tool\Bundle\StorageUtilsBundle\Doctrine\Common\Saver\BaseSaver;
+use Akeneo\Tool\Component\Classification\Model\CategoryInterface;
+use Doctrine\Common\Persistence\ObjectManager;
+use Symfony\Component\EventDispatcher\EventDispatcherInterface;
+use Symfony\Component\Lock\Exception\LockConflictedException;
+use Symfony\Component\Lock\LockFactory;
+use Symfony\Component\Lock\LockInterface;
+
+class CategorySaver extends BaseSaver
```
ahocquard, line 14: Could avoid to extend basesaver and copy past the logic? This generic saver should not exist at first and it allows to decouple a little bit (and allows to have more flexibility for each entity).

Done here for example also: https://github.com/akeneo/pim-community-dev/blob/master/src/Akeneo/Channel/Bundle/Storage/Orm/ChannelSaver.php

src/Akeneo/Tool/Bundle/ClassificationBundle/Storage/Orm/CategorySaver.php
```diff
@@ -0,0 +1,78 @@
...
+            $lock->release();
+        }
+    }
+
+    /**
+     * There is a possible deadlock when creating several new categories at the same time.
+     * Gedmo/Tree execute queries like `UPDATE pim_catalog_category SET lft = lft + 2 WHERE lft >= 2 AND root = ?`
+     * producing GAP Lock issues.
+     *
+     * The entity manager is not capable to recover from such error.
```
ahocquard, line 49:
```suggestion
     * The entity manager is closed and a new one is created in the entity manager registry in that case. The issue is that the reference to the previous entity manager is kept in many places, as it does not get the entity manager from the registry. 
```

src/Akeneo/Tool/Bundle/ClassificationBundle/Storage/Orm/CategorySaver.php
```diff
@@ -0,0 +1,78 @@
...
+     * By locking with an external lock beforehand, we can mitigate this deadlock issue.
+     *
+     * This lock is restricted to the category root, since only rows belonging to this root are affected by the
```
ahocquard, line 52: :pray:

src/Akeneo/Tool/Bundle/ClassificationBundle/Storage/Orm/CategorySaver.php
```diff
@@ -0,0 +1,78 @@
...
+    {
+        $lockIdentifier = sprintf('create_category_in_root_%d', $object->getRoot());
+        $lock = $this->lockFactory->createLock($lockIdentifier, self::LOCK_TTL);
+
+        $errors = 0;
+        $acquired = false;
+
+        while (!$acquired && $errors < 3) {
+            try {
+                $acquired = $lock->acquire(true);
```
ahocquard, line 65: are you sure? 

It will call `$this->store->waitAndSave($this->key);` and in PDOStore `        @trigger_error(sprintf('%s() is deprecated since Symfony 4.4 and will be removed in Symfony 5.0.', __METHOD__), \E_USER_DEPRECATED);
`

They don't like OOP btw, `waitAndSave` function is not declared in the interface `PersistingStoreInterface` used by `Lock` class...

src/Akeneo/Tool/Bundle/ClassificationBundle/Storage/Orm/CategorySaver.php
```diff
@@ -0,0 +1,78 @@
...
+use Akeneo\Tool\Component\Classification\Model\CategoryInterface;
+use Doctrine\Common\Persistence\ObjectManager;
+use Symfony\Component\EventDispatcher\EventDispatcherInterface;
+use Symfony\Component\Lock\Exception\LockConflictedException;
+use Symfony\Component\Lock\LockFactory;
+use Symfony\Component\Lock\LockInterface;
+
+class CategorySaver extends BaseSaver
+{
+    private const LOCK_TTL = 10;
```
ahocquard, line 16:
```suggestion
    private const LOCK_TTL_IN_SECONDS = 10;
```

src/Akeneo/Tool/Bundle/ClassificationBundle/spec/Storage/Orm/CategorySaverSpec.php
```diff
@@ -0,0 +1,94 @@
...
+use Doctrine\Common\Persistence\ObjectManager;
+use PhpSpec\ObjectBehavior;
+use Prophecy\Argument;
+use Symfony\Component\EventDispatcher\EventDispatcherInterface;
+use Symfony\Component\EventDispatcher\GenericEvent;
+use Symfony\Component\Lock\Exception\LockConflictedException;
+use Symfony\Component\Lock\LockFactory;
+use Symfony\Component\Lock\LockInterface;
+
+class CategorySaverSpec extends ObjectBehavior
```
ahocquard, line 18: This spec does not provide you any confidence that a category is well saved: don't mock what you don't own. Integration test is better for persistence (integration tests on adapters in hexagonal architecture).

upgrades/schema/Version_5_0_20201112083948_add_lock_table.php
```diff
@@ -0,0 +1,26 @@
...
+namespace Pim\Upgrade\Schema;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+
+final class Version_5_0_20201112083948_add_lock_table extends AbstractMigration
+{
+    public function up(Schema $schema) : void
+    {
+        $sql = <<<SQL
```
ahocquard, line 12: where is is called when you install the PIM?

src/Akeneo/Tool/Bundle/ClassificationBundle/Storage/Orm/CategorySaver.php
tseho: :wink:

src/Akeneo/Tool/Bundle/ClassificationBundle/Storage/Orm/CategorySaver.php
tseho: not really. A new one is not automatically created, you have to ask for it. That's what I've tried by calling `ManagerRegistry::resetManager()`.
Something like:
```
The entity manager is not capable to recover from such error. If a new entity manager is created using the ManagerRegistry, there is a lot of issues because the reference to the previous entity manager is kept in many places.
```
WDYT?

upgrades/schema/Version_5_0_20201112083948_add_lock_table.php
tseho: That's a really good point.

tseho: `InstallerEvents::POST_DB_CREATE` ?

src/Akeneo/Tool/Bundle/ClassificationBundle/Storage/Orm/CategorySaver.php
tseho: > are you sure?

Yes, I'm aware that PDOStore does not implement `BlockingStoreInterface`, that's why it's decorated with `
Symfony\Component\Lock\Store\RetryTillSaveStore`

src/Akeneo/Tool/Bundle/ClassificationBundle/spec/Storage/Orm/CategorySaverSpec.php
tseho: You are right.
I should add an integration test (or check if it already exists)
Btw, the spec was still a good idea, because I found an issue with the initial version when writing it :man_shrugging:

src/Akeneo/Tool/Bundle/ClassificationBundle/Storage/Orm/CategorySaver.php
ahocquard:
```
The entity manager is closed and is not capable to recover from such error. A new entity manager should be created but if a new one is created using the ManagerRegistry, there is a lot of issues because the reference to the previous entity manager is kept in many places.
```

Just one thing to notice it closed the entity manager automatically (which is an important detail I think), but perfect otherwise :+1:

upgrades/schema/Version_5_0_20201112083948_add_lock_table.php
ahocquard: Yes, you should have a subscriber to call it I think :)

jmleroux (approved):

ahocquard (approved):
Thanks for taking into account my comments, GTM

PR #13226, opened 2020-11-10
PIM-9545: Update monolog config to avoid possible memory leak

mmetayer:

Our monolog config used to have the default value (0 =unlimited) for the `crossed_fingers` handlers `buffer_size` param, which means it stored every log message in a buffer until one had a level high enough to be flushed. 
This has almost always no consequence, except in large jobs, where the amount of logged messages can stack up and cause an out of memory error (e.g in product imports, an INFO message is logged each time a product is updated or created)

This PR limits the buffer_size to 1000, meaning that when a message is flushed, only the 1000 previous messages will be flushed alongside.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/config/packages/prod/monolog.yml
+++ b/config/packages/prod/monolog.yml
@@ -4,6 +4,7 @@ monolog:
             type:         fingers_crossed
             action_level: warning
             handler:      nested
+            buffer_size:  1000
         nested:
             type: stream
             path: '%kernel.logs_dir%/%kernel.environment%.log'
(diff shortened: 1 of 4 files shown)


config/packages/prod/monolog.yml
```diff
@@ -4,6 +4,7 @@ monolog:
             type:         fingers_crossed
             action_level: warning
             handler:      nested
+
```
jmleroux, line 7: Why not here?

jmleroux (approved):
GTM with question

nmarniesse (approved):
Same question as JM about the prod config

config/packages/prod/monolog.yml
mmetayer: that's a mistake :/

PR #13242, opened 2020-11-12
PIM-9550: Add attribute codes as an argument to the command 'pim:product:clean-removed-attributes'

tseho:

With this PR, the command `pim:product:clean-removed-attributes` take a list of attribute codes as an optional argument.
Example: `bin/console pim:product:clean-removed-attributes foo bar`

If used, the new service `CleanValuesOfRemovedAttributes` will handle the job.

The main optimization is that we are looking for the products & product models that have a value on one of those attributes with an elasticsearch query.

From there, the process is straightforward:
- Find the identifiers
- Load the products from the repository
- Save the products
- Clear the UnitOfWork
- Rinse and repeat

Screenshot_2020-11-13_16-01-51

Screenshot_2020-11-13_16-01-33

Screenshot_2020-11-13_16-02-08


src/Akeneo/Pim/Enrichment/Bundle/Command/CleanRemovedAttributesFromProductAndProductModelCommand.php
```diff
@@ -59,14 +66,25 @@ public function __construct(
     protected function configure()
     {
         $this
-            ->setDescription('Removes all values of deleted attributes on all products and product models');
+            ->setDescription('Removes all values of deleted attributes on all products and product models')
+            ->addArgument('attributes', InputArgument::OPTIONAL | InputArgument::IS_ARRAY);
```
ahocquard, line 70: BC break

src/Akeneo/Pim/Enrichment/Bundle/Command/CleanRemovedAttributesFromProductAndProductModelCommand.php
```diff
@@ -164,4 +182,39 @@ private function launchCleanTask(array $productIds, string $env, string $rootDir
...
+    private function cleanProductsWithAttributesCodes(
+        array $attributesCodes,
+        InputInterface $input,
+        OutputInterface $output
+    ): void {
+        $countAffected = $this->removeAttributesValuesFromProductAndProductModel->countAffectedProductAndProductModel($attributesCodes);
+
+        $io = new SymfonyStyle($input, $output);
+        $io->title('Clean removed attributes values');
+        $answer = $io->confirm(
```
ahocquard, line 195: extract variable + sprintf would make it more readable in my opinion, but not very important

src/Akeneo/Pim/Enrichment/Bundle/Product/RemoveAttributesValuesFromProductAndProductModel.php
```diff
@@ -0,0 +1,206 @@
...
+        $this->attributeRepository = $attributeRepository;
+        $this->elasticsearchClient = $elasticsearchClient;
+        $this->productModelRepository = $productModelRepository;
+        $this->productModelSaver = $productModelSaver;
+        $this->productRepository = $productRepository;
+        $this->productSaver = $productSaver;
+        $this->clearer = $clearer;
+    }
+
+    public function countAffectedProductAndProductModel(array $attributesCodes): int
```
ahocquard, line 54: I would split the count for product and product models, at least for the progress. But maybe it's complicated to have two progress bars and it does not worth it.

src/Akeneo/Pim/Enrichment/Bundle/Product/RemoveAttributesValuesFromProductAndProductModel.php
```diff
@@ -0,0 +1,206 @@
...
+                            }, $attributesCodes),
+                            'minimum_should_match' => 1,
+                        ],
+                    ],
+                ],
+            ],
+            'sort' => [
+                'identifier' => 'asc',
+            ],
+            'track_total_hits' => true,
```
ahocquard, line 192: you can remove this line

src/Akeneo/Pim/Enrichment/Bundle/Product/RemoveAttributesValuesFromProductAndProductModel.php
```diff
@@ -0,0 +1,206 @@
...
+        }
+
+        foreach ($attributesCodes as $attributeCode) {
+            $this->validateAttributeCode($attributeCode);
+        }
+    }
+
+    private function validateAttributeCode(string $attributeCode): void
+    {
+        $violations = $this->validator->validate($attributeCode, [
```
ahocquard, line 139: I think that this validation will never be maintained/synchronized with attributes :/ is the list `id,iD,etc...` required?

src/Akeneo/Pim/Enrichment/Bundle/Product/RemoveAttributesValuesFromProductAndProductModel.php
```diff
@@ -0,0 +1,206 @@
...
+use Akeneo\Pim\Enrichment\Component\Product\Repository\ProductModelRepositoryInterface;
+use Akeneo\Pim\Enrichment\Component\Product\Repository\ProductRepositoryInterface;
+use Akeneo\Pim\Structure\Bundle\Doctrine\ORM\Repository\AttributeRepository;
+use Akeneo\Tool\Bundle\ConnectorBundle\Doctrine\UnitOfWorkAndRepositoriesClearer;
+use Akeneo\Tool\Bundle\ElasticsearchBundle\Client;
+use Akeneo\Tool\Component\StorageUtils\Saver\BulkSaverInterface;
+use Symfony\Component\Validator\Constraints as Assert;
+use Symfony\Component\Validator\Validator\ValidatorInterface;
+
+class RemoveAttributesValuesFromProductAndProductModel
```
ahocquard, line 16: one implementation = final
If you want to mock it somewhere = several implementation (a mock is an implementation) = interface

But maybe the real question is:
- do you see it as a service used in a command, cleaning values = an interface is probably a good thing, as the implementation of a service can change.
- or do you see it as a command, and so: no infrastructure stuff (ES queries should be isolated in dedicated class, etc), and it's a final as there is only one implementation of a command.

This is important for how you will test also for future change.

src/Akeneo/Pim/Enrichment/Bundle/Product/RemoveAttributesValuesFromProductAndProductModel.php
```diff
@@ -0,0 +1,206 @@
...
+            new Assert\Regex('/^[a-zA-Z0-9_]+$/'),
+            new Assert\Regex('/^(?!(id|iD|Id|ID|associationTypes|categories|categoryId|completeness|enabled|(?i)\bfamily\b|groups|associations|products|scope|treeId|values|category|parent|label|(.)*_(products|groups)|entity_type|attributes)$)/'),
+            new Assert\Regex('/^[^\n]+$/D'),
+        ]);
+
+        if (count($violations) > 0) {
+            throw new \InvalidArgumentException(sprintf('The attribute code "%s" is not valid.', $attributeCode));
+        }
+
+        $attribute = $this->attributeRepository->findOneByIdentifier($attributeCode);
```
ahocquard, line 153: Avoid ORM, use queries for that. https://github.com/akeneo/pim-community-dev/blob/master/src/Akeneo/Pim/Structure/Component/Query/PublicApi/AttributeType/GetAttributes.php

ahocquard (commented):
There is no test yet, etc, I suppose it's still a draft, so I just comment.

I like the implementation:
- you can improve it if needed later on for a faster implementation
- you can use it in other part of the application
- it relies on attribute code, not attribute
- it avoids the PQB, which is much simpler

Maybe you could put the progress bar and the attributes in an object that is capable of validating itself. But the issue is that you have a query to fetch existing attributes.

src/Akeneo/Pim/Enrichment/Bundle/Command/CleanRemovedAttributesFromProductAndProductModelCommand.php
tseho: Is it ? It's an optional argument.

src/Akeneo/Pim/Enrichment/Bundle/Product/RemoveAttributesValuesFromProductAndProductModel.php
tseho: as of today, the progress bar is not 100% accurate.
Since I count both product_model & product with the attribute, when a product_model is cleaned, its variants are cleaned & indexed too, meaning we will probably clean less products than expected.

tseho: I could also put a filter on `attributes_for_this_level` for the count but then, it will count less than the amount of updated products.

src/Akeneo/Pim/Enrichment/Bundle/Product/RemoveAttributesValuesFromProductAndProductModel.php
tseho: I kinda copy/pasted the validation of `attribute::code`.
I don't mind being less strict there.

src/Akeneo/Pim/Enrichment/Bundle/Product/RemoveAttributesValuesFromProductAndProductModel.php
tseho: > do you see it as a service used in a command, cleaning values = an interface is probably a good thing, as the implementation of a service can change.
I think you are right, an interface would be better. Especially if we start using somewhere else.

tseho: I'm not sure anymore, will think about it.

src/Akeneo/Pim/Enrichment/Bundle/Product/RemoveAttributesValuesFromProductAndProductModel.php
tseho: done

src/Akeneo/Pim/Enrichment/Bundle/Product/RemoveAttributesValuesFromProductAndProductModel.php
tseho: I've done a refactor with new interfaces and separate queries.

src/Akeneo/Pim/Enrichment/Bundle/Product/RemoveAttributesValuesFromProductAndProductModel.php
tseho: done

src/Akeneo/Pim/Enrichment/Component/Product/ValuesRemover/CleanValuesOfRemovedAttributesInterface.php
```diff
@@ -0,0 +1,18 @@
+<?php
+
+namespace Akeneo\Pim\Enrichment\Component\Product\ValuesRemover;
+
+interface CleanValuesOfRemovedAttributesInterface
+{
+    public function validateRemovedAttributesCodes(array $attributesCodes): void;
```
tseho, line 7: I really hate having this exposed there.
2 alternatives:
Each public method validate the given attributes codes (costly, there is sql queries in the validation)
A DTO object, with the attributes codes and a validation on this DTO at creation. But since we need sql queries for the validation and sincethis should happen in the command `CleanRemovedAttributesFromProductAndProductModelCommand`, this will introduce a BC Break.

ahocquard: > costly, there is sql queries in the validation

GetAttributes uses a LRU cache, so it doesn't matter in term of perf

src/Akeneo/Pim/Enrichment/Bundle/Command/CleanRemovedAttributesFromProductAndProductModelCommand.php
ahocquard: Ah, I didn't notice that you kept the previous code. 
Maybe you could add a comment to make the argument mandatory on master. It would help to remove all the previous commands/sub-process code.

tseho: I'm not sure about making it required on master, If I remember well, sometimes, the users forget which attributes were removed and run the command on all products as a "catch-all".

jmleroux:
check frozen

jmleroux:
This command will probably be reworked soon by Raccoons.

PR #13262, opened 2020-11-17
alter completeness table

greg-akeneo:

It appears some users have reached the maximum value an identifier could reach in the completeness table. 
This PR alter the completeness table turning the `id` field from an int (32 bits) to bigint (64 bits). 

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,0 +1,1 @@
+ * turn completeness table `id` from int to bigint
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,1 +1,0 @@
- * turn completeness table `id` from int to bigint
--- a/upgrades/schema/Version_5_0_20201117114538_modify_auto_increment_integer.php
+++ b/upgrades/schema/Version_5_0_20201117114538_modify_auto_increment_integer.php
@@ -1,0 +1,20 @@
+<?php declare(strict_types=1);
+
+namespace Pim\Upgrade\Schema;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+
+final class Version_5_0_20201117114538_modify_auto_increment_integer.php extends AbstractMigration
+{
+    public function up(Schema $schema) : void
+    {
+        $sql = "ALTER TABLE pim_catalog_completeness MODIFY id bigint AUTO_INCREMENT";
+        $this->addSql($sql);
+    }
+
+    public function down(Schema $schema) : void
+    {
+        $this->throwIrreversibleMigrationException();
+    }
+}
--- a/upgrades/schema/Version_5_0_20201117114538_modify_auto_increment_integer.php
+++ b/upgrades/schema/Version_5_0_20201117114538_modify_auto_increment_integer.php
@@ -5,7 +5,7 @@ namespace Pim\Upgrade\Schema;
 use Doctrine\DBAL\Schema\Schema;
 use Doctrine\Migrations\AbstractMigration;
 
-final class Version_5_0_20201117114538_modify_auto_increment_integer.php extends AbstractMigration
+final class Version_5_0_20201117114538_modify_auto_increment_integer extends AbstractMigration
 {
     public function up(Schema $schema) : void
     {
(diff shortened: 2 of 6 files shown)


CHANGELOG
```diff
@@ -0,0 +1 @@
+ * turn completeness table `id` from int to bigint
```
jmleroux, line 1: Bad file. :wink:

jmleroux (changes requested):

CHANGELOG
greg-akeneo: ah…

jmleroux (approved):
GTM after remove of the empty CHANGELOG

BitOne:
Is this column really useful? Do we really need it?

upgrades/test_schema/Version_5_0_20201117114538_modify_auto_increment_integer_Integration.php
```diff
@@ -0,0 +1,41 @@
...
+
+    private const MIGRATION_LABEL = '_5_0_20201117114538_modify_auto_increment_integer';
+
+    public function test_completeness_id_has_changed_to_bigint(): void
+    {
+        $this->ensureCompletenessUsesNormalInt();
+        $this->reExecuteMigration(self::MIGRATION_LABEL);
+        $tableColumns = $this->schemaManager->listTableColumns('pim_catalog_completeness');
+
+        $found = false;
```
SamirBoulil, line 21: We could extract this bit into a private assertIdIsMigrated :)

SamirBoulil (approved):

nmarniesse (approved):

upgrades/schema/Version_5_0_20201117114538_modify_auto_increment_integer.php
```diff
@@ -0,0 +1,20 @@
+<?php declare(strict_types=1);
+
+namespace Pim\Upgrade\Schema;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+
+final class Version_5_0_20201117114538_modify_auto_increment_integer extends AbstractMigration
```
mmetayer, line 8: you also need to update the table creation script for new installs: `src/Akeneo/Pim/Enrichment/Bundle/EventSubscriber/Db/InitCompletenessDbSchemaSubscriber.php`

upgrades/test_schema/Version_5_0_20201117114538_modify_auto_increment_integer_Integration.php
```diff
@@ -0,0 +1,41 @@
...
+
+use Akeneo\Test\Integration\TestCase;
+
+final class Version_5_0_20201117114538_modify_auto_increment_integer_Integration extends TestCase
+{
+    use ExecuteMigrationTrait;
+
+    private const MIGRATION_LABEL = '_5_0_20201117114538_modify_auto_increment_integer';
+
+    public function test_completeness_id_has_changed_to_bigint(): void
```
mmetayer, line 15: There's no need to test the structure changes, (it should be handled by the `test_back_missing_structure_migrations`  step), these tests are for data migration.

Instead you could test that former ids were preserved, e.g
- insert data in the table with a given id
- execute migration
- check that the previously inserted id is the same (kind of a dumb test, not sure we really need it)

mmetayer (changes requested):
please update the table creation script

mmetayer:
Also, I agree with @BitOne, a better way to fix this would be to drop the id key, and to have a composite ` product_id`-` locale_id`-` channel_id`  primary key (I guess it would be easier now that completeness is not managed by Doctrine ORM anymore).

upgrades/test_schema/Version_5_0_20201117114538_modify_auto_increment_integer_Integration.php
greg-akeneo: ok, I do agree there are no much data to test, I will move this test in the appropriate section.

upgrades/schema/Version_5_0_20201117114538_modify_auto_increment_integer.php
greg-akeneo: done in f363545aef

mmetayer (approved):

PR #13335, opened 2020-11-24
PIM-9568: Fix performance issue when saving a big product group

LaurentPetard:
This fix a performance issue when saving a group with a lot of products (whatever the change)

The group updater was removing all the products before re-adding them all, and the group saver was saving all the products for the versioning.

Now, it will only add, remove and save the products that need it.

It's not an ideal solution, and the problem will remain if too many products are added/removed in the same request. But better solutions will cause BC-breaks.

### DIFF
--- a/CHANGELOG-4.0.md
+++ b/CHANGELOG-4.0.md
@@ -6,6 +6,7 @@
 
 - PIM-9565: Fix StandardToFlat boolean value converter
 - PIM-9550: Add attribute codes as an argument to the command "pim:product:clean-removed-attributes"
+- PIM-9568: Fix performance issue when saving a big product group
 
 # 4.0.72 (2020-11-16)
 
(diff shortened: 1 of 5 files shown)


jmleroux (approved):

CHANGELOG-4.0.md
```diff
@@ -6,6 +6,7 @@
 
 - PIM-9565: Fix StandardToFlat boolean value converter
 - PIM-9550: Add attribute codes as an argument to the command "pim:product:clean-removed-attributes"
+- PIM-9568: Fix performance issue when saving a big product group
```
Doodoune, line 9: Could you please move the changelog line? This fix is not in 4.0.73 ;)

LaurentPetard: Oops...

Doodoune (approved):

PR #13351, opened 2020-11-25
PIM-9573: Create remove_non_existing_product_values job instance

jmleroux:
Create remove_non_existing_product_values job instance at runtime

#12514 introduced a BC break and a new job instance created with a migration. If the migration is not run, the application will crash.

With this PR, we create the JobInstance at runtime if it doesn't exists so that the app can survive the missing migration :smiley:

### DIFF
--- a/src/Akeneo/Pim/Structure/Bundle/EventListener/RemoveNonExistingProductValuesSubscriber.php
+++ b/src/Akeneo/Pim/Structure/Bundle/EventListener/RemoveNonExistingProductValuesSubscriber.php
@@ -8,7 +8,7 @@ use Akeneo\Pim\Structure\Component\Model\AttributeOptionInterface;
 use Akeneo\Tool\Bundle\BatchBundle\Job\JobInstanceRepository;
 use Akeneo\Tool\Bundle\BatchBundle\Launcher\JobLauncherInterface;
 use Akeneo\Tool\Component\Batch\Model\JobInstance;
-use Akeneo\Tool\Component\Batch\Query\SqlCreateJobInstanceInterface;
+use Akeneo\Tool\Component\Batch\Query\CreateJobInstanceInterface;
 use Akeneo\Tool\Component\StorageUtils\StorageEvents;
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
 use Symfony\Component\EventDispatcher\GenericEvent;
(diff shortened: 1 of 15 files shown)


src/Akeneo/Pim/Structure/Bundle/EventListener/RemoveNonExistingProductValuesSubscriber.php
```diff
@@ -19,28 +21,24 @@
...
-
-    /** @var JobInstanceRepository */
-    private $jobInstanceRepository;
-
-    /** @var JobLauncherInterface */
-    private $jobLauncher;
-
-    /** @var string */
-    private $jobName;
+    private TokenStorageInterface $tokenStorage;
```
ValentinMumble, line 24: We are not on php7.4 on 4.0 are we? 🤔

ValentinMumble (approved):

src/Akeneo/Pim/Structure/Bundle/EventListener/RemoveNonExistingProductValuesSubscriber.php
jmleroux: woopsy

jmleroux: I first made it on master and forgot to revert this. CI stopped me :wink:

src/Akeneo/Tool/Component/Batch/Query/SqlCreateJobInstanceInterface.php
```diff
@@ -0,0 +1,15 @@
...
+declare(strict_types=1);
+
+namespace Akeneo\Tool\Component\Batch\Query;
+
+/**
+ * @author    JM Leroux <jean-marie.leroux@akeneo.com>
+ * @copyright 2020 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+interface SqlCreateJobInstanceInterface
```
nmarniesse, line 12:
```suggestion
interface CreateJobInstanceInterface
```

Otherwise the existence of the interface is questionable

nmarniesse (commented):
I don't get why the migration (the one that should create the job instance) is not played?  
It's not a migration problem that we try to fix by a workaround?

jmleroux:
> I don't get why the migration (the one that should create the job instance) is not played?
> It's not a migration problem that we try to fix by a workaround?

Yes it is, but we didn't take into account that the doc doesn't tell to play migration while upgrading from a patch, and that it shouldn't talk about it.

So yes, we're trying to fix a migration problem by a workaround, but on purpose.

nmarniesse (approved):
Thanks for your explaination

PR #13353, opened 2020-11-25
PIM-9559: Dispatch event when clean removed attributes command is over

ValentinMumble:

Setting methods as protected to be able to extend this command in EE

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/CHANGELOG-4.0.md
+++ b/CHANGELOG-4.0.md
@@ -3,7 +3,7 @@
 ## Bug fixes
 
 - PIM-9575: Prepend hash to navigate actions URLs
-- PIM-9559: Override clean removed attributes command in EE to also clean proposals
+- PIM-9559: Dispatch event when clean removed attributes command is over
 
 # 4.0.74 (2020-11-25)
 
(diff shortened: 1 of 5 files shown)


jmleroux (approved):

jmleroux:
check frozen

tseho:
Do we often extends `Command` ? I would think that a command should be `final`, with `private` methods if needed.
In EE, we have dedicated commands under the `pimee` prefix, for proposals and such, like `pimee:product-proposal:index`.

IMHO, it would be better to have either 2 commands (one for product/product models and one for proposals) or that the command call a service that would do the cleanup (service which can then be decorated in EE)

If it's too much work, the current solution should still be enough.

SamirBoulil (approved):

SamirBoulil:
I completely agree with @tseho 's remarks. But since we know this command might not be needed in the future since we are working on this topic. This solution is also fine in this context.

ahocquard (approved):

LaurentPetard (approved):

PR #13178, opened 2020-11-04
CXP-493: Do not save a product when it was not updated

mmetayer:
**Description (for Contributor and Core Developer)**

This PR aims at not saving a product which was not really updated when calling the product saver. This has several benefits:
- unnecessary PRE/POST_SAVE events are not dispatched anymore
- the heavy completeness calculation and ES indexation processes are not triggered anymore when they don't need to
- unnecessary webhooks are not called anymore

**How?**

A flag (`$wasUpdated`) is added to the Product and is set to `true` whenever the product is really updated. Practically, when calling a setter, the new value is now compared with the former one. The product saver will then only save the products with their flag set to `true`

This has a big consequence: it means that, as of now, the product **must be aware of every updated that is applied to it**. It doesn't change anything for simple or one-to-one properties, but it means that we don't want a one-to-many or many-to-many collection property to be updated "from outside" (e.g `$product->getvalues()->add($myvalue))

As a consequence, every getter for such a property (`values`, `groups`, `categories` and `associations`) *now returns a copy of the collection*. You will have to explicitely call the product's setters/adders/removers to update these collections: ` $product->addValue($myValue)`

**Choices / Remarks**

- Some services need to save the product anyway, even if it was not updated (technically speaking). For instance, this is the case for the `pim:product:clean-removed-attributes` console command, which just loads a product from the DB (this filters values linked to removed attributes, but does not update the `wasUpdated` flag) and immediately saves it. In order to achieve that I added a `force_save` option in the ProductSaver; when it's true, the product is saved regardless of his `wasUpdated` state

- comparing associations is a quite tedious and complex task. I decided to set the `wasUpdated` flag to true whenever a setter is called (even if the collection is not really updated). This, unfortunately, decreases this PR's efficiency; I believe the associations need to be reworked in order to be easier to use (both from a technical and from a functional POV)

- there are 2 product methods which do not  return copies: `getAssociationByType()` and `getAssociationByTypeCode()`. I couldn't find a way to make the 2-way associations work with it (there are duplicates because of the ORM's persistence with manytomany associations)

**Definition Of Done (for Core Developer only)**

| Q                                 | A
| --------------------------------- | ---
| Added Specs                       | Todo
| Added legacy Behats               | Todo
| Added acceptance tests            | Todo
| Added integration tests           | Todo
| Changelog updated                 | Todo
| Review and 2 GTM                  | Todo
| Micro Demo to the PO (Story only) | Todo
| Migration script                  | -
| Tech Doc                          | -

[...]

PR #13373, opened 2020-11-27
CXP-493: Clone ValueCollection instead of instantiating a new one

mmetayer:

Following #13178, the `AbstractProduct::getValues()` method now returns a copy of its WriteValueCollection. (this was already the case before for variant products, but not for simple products)
Problem is, this method is used quite a lot in the PIM, and instantiating a WriteValueCollection is quite costly, as [it performs a number of checks and computations for each value](https://github.com/akeneo/pim-community-dev/blob/master/src/Akeneo/Pim/Enrichment/Component/Product/Model/WriteValueCollection.php#L233). 
To avoid this, I now return a mere clone of the original value collection, as all these checks and computations were already performed on the original collection, so they don't need to be run on the new one. 

Here are the Blackfire results from a CPU time POV (sorry I couldn't find a way to compare the two views)

Before:

without_clone

After:

with_clone

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Akeneo/Pim/Enrichment/Component/Product/Model/AbstractProduct.php
+++ b/src/Akeneo/Pim/Enrichment/Component/Product/Model/AbstractProduct.php
@@ -281,11 +281,10 @@ abstract class AbstractProduct implements ProductInterface
      */
     public function getValues(): WriteValueCollection
     {
+        $values = clone $this->values;
         if (!$this->isVariant()) {
-            return WriteValueCollection::fromCollection($this->values);
+            return $values;
         }
-
-        $values = WriteValueCollection::fromCollection($this->values);
 
         return $this->getAllValues($this, $values);
     }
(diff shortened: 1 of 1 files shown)


pierallard (approved):

tseho (approved):

jmleroux (approved):
Add a CHANGELOG entry then GTM

ahocquard (approved):

SamirBoulil (approved):

ahocquard:
@jmleroux no need for changelog, as it's master no?

mmetayer:
@jmleroux this PR doesn't change any behavior, nor introduces new BC breaks. All behavior changes were already documented in #13178, see [this commit](https://github.com/akeneo/pim-community-dev/pull/13178/commits/6a21acae988f5a36fe56a11ff22f880c18002e07)

PR #13393, opened 2020-11-30
PIM-9543: Print PDF with Chinese characters

jmleroux:
We need to generate PDF with Chinese content. There may be a better font, but we didn't find it.
If anyone is aware of a better font, I'm ok to try it.

I limited the new font usage to string containing Asian characters, so that the PDF rendering should not be too much impacted.

To test with this content:

在愛活置交玉域護感内種口参顧大父回闘。

州都聞懲江面新法線去通求導謙無記骨運。

記波幸滑番覚最問図見体示禁均先本申 都続更覧探玉聞語介会米極身堀

並給駐世騒当地違初著図根申大成本創庫続家 欧家未所測堂者上抗今豊政 防等熊図連契言能属前注回止反日学覧土禁 旧更域維紙予写藤授食閉外信

繁停当懇井毎座策久提権今活

PS: I hope I'm not insulting someone or something with this string. I tested it on Google Translate and it seems to be ok. :laughing: 

Here is the resulting PDF:

[594877-2020-11-30_16-17-47.pdf](https://github.com/akeneo/pim-community-dev/files/5617157/594877-2020-11-30_16-17-47.pdf)


BitOne (approved):

src/Akeneo/Pim/Enrichment/Bundle/Twig/AttributeExtension.php
```diff
@@ -0,0 +1,40 @@
...
+    {
+        return [
+            new Twig_SimpleFunction('contains_asian_characters', [$this, 'containsAsianCharacters']),
+        ];
+    }
+
+    /**
+     * Test an attribute content for Asian characters
+     */
+    public function containsAsianCharacters(?string $text): bool
```
mmetayer, line 32:
```suggestion
    public function containsChineseCharacters(?string $text): bool
```
Asia has more than 2400 languages :slightly_smiling_face:

mmetayer (approved):
GTM after wording fix

src/Akeneo/Pim/Enrichment/Bundle/Twig/AttributeExtension.php
jmleroux: It's not only Chinese, but I can be more precise with smth like `containsHanCharacters`

PR #13398, opened 2020-11-30
CXP-503: Add the number of activated WH in the tracker

tomglvng:
In order to follow-up the Events API activation, we add the number of activated event subscriptions into the PIM tracker. 

`nb_active_event_subscription : integer
`


willy-ahva (approved):

pchasle (approved):
:rocket:

PR #13330, opened 2020-11-23
RAC-384: Fix fatal error when an attribute is removed then re-created with the same code but another type

tseho:
**Description (for Contributor and Core Developer)**

When a attribute is removed, attribute values in the products are kept.
When the attribute is re-created with the same code but a different type, the PIM tries to load the data into the new attribute structure.
When it goes from `text` to `texterea`, nothing happens.
When it goes from more complex attributes, like `price` to an incompatible one, it throws 500 errors.

The purpose of this PR is to catch those incompatible conversion errors and skip the failing attribute.
By doing so, the invalid value is not included in the responses of the API endpoints (both internal & external).

Any invalid value will be lost when the product is saved.

I've updated some specs accordingly.
I've added 2 integration tests on `ReadValueCollectionFactory` and `WriteValueCollectionFactory`.

**Definition Of Done (for Core Developer only)**

| Q                                 | A
| --------------------------------- | ---
| Added Specs                       | yes
| Added legacy Behats               | -
| Added acceptance tests            | -
| Added integration tests           | yes
| Changelog updated                 | yes
| Review and 2 GTM                  | Todo
| Micro Demo to the PO (Story only) | Todo
| Migration script                  | -
| Tech Doc                          | -

PR #13403, opened 2020-11-30
RAC-388: Fix fatal error when an attribute is removed then re-created with the same code but another type

tseho:

see https://github.com/akeneo/pim-community-dev/pull/13330 for the 3.2 version

Between 3.2 and 4.0, in `ReadValueCollectionFactory`, the implem of `createWithoutCheckingData` was added to improve the performances of the API.
Sadly, this allows fatal errors when an attribute is removed then re-created with the same code but another type.

After the benchmarks done by Steven and I, on a reference catalog (100 products with 120 attributes each), we found out that using `createByCheckingData` instead of `createWithoutCheckingData` effectively increase ReadValueCollectionFactory time by **+20%**.
This result in a time increase of **+8%** on the product API endpoint, when fetching 100 products.

Since most of those checks are done by calling a lot of php functions, like `is_string`, we also found out that prefixing those function by `\` slightly increase the perfs, and limit the overhead cost to **+5.5%**.

This is, in our opinion, an acceptable trade-off to resolve this critical error.
Also, **this fix will not be pulled on master**. On master, a proper fix, cleaning the data directly in the database, is currently developed.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Akeneo/Pim/Enrichment/Component/Product/Factory/NonExistentValuesFilter/NonExistentPriceCollectionValueFilter.php
+++ b/src/Akeneo/Pim/Enrichment/Component/Product/Factory/NonExistentValuesFilter/NonExistentPriceCollectionValueFilter.php
@@ -5,6 +5,7 @@ namespace Akeneo\Pim\Enrichment\Component\Product\Factory\NonExistentValuesFilte
 
 use Akeneo\Channel\Component\Query\FindActivatedCurrenciesInterface;
 use Akeneo\Pim\Structure\Component\AttributeTypes;
+use Akeneo\Tool\Component\StorageUtils\Exception\InvalidPropertyTypeException;
 
 /**
  * @author    Anael Chardan <anael.chardan@akeneo.com>
@@ -55,6 +56,14 @@ final class NonExistentPriceCollectionValueFilter implements NonExistentValuesFi
                     }
 
                     foreach ($valuesIndexedByLocale as $locale => $value) {
+                        if (!\is_array($value)) {
+                            throw InvalidPropertyTypeException::arrayExpected(
+                                $attributeCode,
+                                static::class,
+                                $value
+                            );
+                        }
+
                         $amountByCurrency = [];
                         foreach ($value as $price) {
                             if (isset($price['amount']) && isset($price['currency'])) {
--- a/src/Akeneo/Pim/Enrichment/Component/Product/Factory/NonExistentValuesFilter/NonExistentReferenceDataMultiSelectValuesFilter.php
+++ b/src/Akeneo/Pim/Enrichment/Component/Product/Factory/NonExistentValuesFilter/NonExistentReferenceDataMultiSelectValuesFilter.php
@@ -5,6 +5,7 @@ namespace Akeneo\Pim\Enrichment\Component\Product\Factory\NonExistentValuesFilte
 
 use Akeneo\Pim\Enrichment\Component\Product\Query\GetExistingReferenceDataCodes;
 use Akeneo\Pim\Structure\Component\AttributeTypes;
+use Akeneo\Tool\Component\StorageUtils\Exception\InvalidPropertyTypeException;
 
 /**
  * As assets are reference data also, we handle it in this filter.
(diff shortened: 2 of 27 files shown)


BitOne (approved):
Good job on the fix + the optimization and all the benchmarking.

Good to go from my point of view.

src/Akeneo/Pim/Enrichment/Component/Product/Factory/ReadValueCollectionFactory.php
```diff
@@ -83,12 +93,25 @@ private function createValues(array $rawValueCollections): array
...
+                        try {
+                            $values[] = $this->valueFactory->createByCheckingData(
+                                $attribute,
+                                $channelCode,
+                                $localeCode,
+                                $data
+                            );
+                        } catch (\TypeError | InvalidPropertyTypeException | InvalidPropertyException $ex) {
+                            if (null !== $this->logger) {
+                                $this->logger->warning(
```
ahocquard, line 106: Should it be a warning? It's a state that could happen in database. Nothing weird.

src/Akeneo/Pim/Enrichment/Component/Product/Factory/ReadValueCollectionFactory.php
```diff
@@ -22,14 +25,21 @@ class ReadValueCollectionFactory
     /** @var ChainedNonExistentValuesFilterInterface */
     private $chainedNonExistentValuesFilter;
 
+    // TODO merge master: property $logger should be nullable
+    /** @var LoggerInterface|null */
+    private $logger;
```
ahocquard, line 30: no need to remove null in master if we remove the try catch

src/Akeneo/Pim/Enrichment/Component/Product/Factory/NonExistentValuesFilter/NonExistentPriceCollectionValueFilter.php
```diff
@@ -55,6 +56,14 @@ private function getExistingPriceCollectionValues(array $priceCollectionValues):
                     }
 
                     foreach ($valuesIndexedByLocale as $locale => $value) {
+                        if (!\is_array($value)) {
```
ahocquard, line 59: to remove in master once you handle attribute deletion correctly

src/Akeneo/Pim/Enrichment/Component/Product/Factory/NonExistentValuesFilter/NonExistentReferenceDataMultiSelectValuesFilter.php
```diff
@@ -48,7 +49,7 @@ private function filterByType(OnGoingFilteredRawValues $onGoingFilteredRawValues
 
                 foreach ($productData['values'] as $channel => $valuesIndexedByLocale) {
                     foreach ($valuesIndexedByLocale as $locale => $values) {
-                        if (is_array($values)) {
+                        if (\is_array($values)) {
```
ahocquard, line 52: to remove in master once you handle attribute deletion correctly, as the try/catch will be removed and the data consistent

ahocquard (approved):
GTM after some fixes

ahocquard:
My advice is to pull-up yourself this PR. You have all the knowledge to do it at the opposite of the maintenance team, and it's critical as it concerns the product in my opinion.

src/Akeneo/Pim/Enrichment/Component/Product/Factory/ReadValueCollectionFactory.php
BitOne: Agree, more like a notice.

src/Akeneo/Pim/Enrichment/Component/Product/Factory/NonExistentValuesFilter/NonExistentPriceCollectionValueFilter.php
tseho: I don't think I will pull those additional checks. Should I ?

SamirBoulil: I agree with both of you, we should pull & revert the commits ^^ since we are creating a better solution for master.

SamirBoulil (approved):

src/Akeneo/Pim/Enrichment/Component/Product/Factory/NonExistentValuesFilter/NonExistentPriceCollectionValueFilter.php
ahocquard: It was to avoid to have this error on Serenity also until the implementation of the better solution, but as you wish as everything is in raccoon's paw.

PR #13426, opened 2020-12-02
DAPI-1392: add product quality scores in API

rouliane:

`OK`: Done / Validated
`-`: Not needed


src/Akeneo/Pim/Enrichment/Component/Product/Connector/ReadModel/ConnectorProduct.php
```diff
@@ -58,6 +59,8 @@ final class ConnectorProduct
     /** @var ReadValueCollection */
     private $values;
 
+    private ?ChannelLocaleRateCollection $qualityScores;
```
LaurentPetard, line 62: Wouldn't be better to set $qualityScores as `array` to avoid an external dependency on a domain object? 
(As a general rule it is preferable to not type with an array when it's possible, but this is a particular case)

LaurentPetard (approved):

src/Akeneo/Pim/Enrichment/Component/Product/Connector/ReadModel/ConnectorProduct.php
AymericPlanche: as discussed with Alex, he prefer a typed object instead of an array, I agree with him. We'll keep it ;)

rouliane: We will probably have to update the coupling detector for this bounded context. I forgot to check

AymericPlanche (approved):

PR #13428, opened 2020-12-03
CXP-596: Remove statefull cache from quantified associations queries

tseho:

In those 4 queries:
```
GetProductQuantifiedAssociationsByProductIdentifiers
GetProductModelQuantifiedAssociationsByProductIdentifiers
GetProductQuantifiedAssociationsByProductModelCodes
GetProductModelQuantifiedAssociationsByProductModelCodes
```
there is a validation step on association type codes because the quantified associations are stored in JSON format in the database, without foreign keys.

For each QuantifiedAssociation hydrated, we need to check that the corresponding QuantifiedAssociationTypeCode still exists in the database.
A simple cache was added, to avoid to run the query `SELECT code FROM pim_catalog_association_type WHERE is_quantified = true` hundred of times when validating.

Instead of having a statefull cache in the class, this PR move the sql query at the start of the hydratation of the batch.
This way, it will run one time for each batch of products.
And bonus, we also remove the risk of having an outdated cache between batches.


StevenVAIDIE (approved):
Perfect !

ValentinMumble (approved):

pchasle (approved):

willy-ahva (approved):

PR #13434, opened 2020-12-04
PIM-9513: Fix the use of an unexisting filter on the API so that it does not return an error 500

anaisbaunelemaire:

If I used this filter, I got a 500 (the "2" isn't in quote) :
`{{url}}/api/rest/v1/products?search={"updated": [{"operator": "SINCE LAST N HOURS","value": 2}]`

When I used this filter, I got a 422, that is expected (the "2" is in quote) :
({{url}}/api/rest/v1/products?search={"updated": [{"operator": "SINCE LAST N HOURS","value": "2"}]}

_NOTE: using "SINCE LAST N DAYS" filter works both with an int or a string (2 or "2")_

**Solution:**
The 500 error was raised because the function createFromFormat() expects a parameter to be string, int was given.
So, I converted $value to string.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/CHANGELOG-4.0.md
+++ b/CHANGELOG-4.0.md
@@ -1,5 +1,9 @@
 # 4.0.x
 
+## Bug fixes
+
+- PIM-9513: Fix the use of an unexisting filter on the API so that it does not return an error 500
+
 # 4.0.76 (2020-12-02)
 
 # 4.0.75 (2020-11-27)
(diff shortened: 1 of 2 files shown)


tseho (approved):
:+1:

jmleroux (approved):

PR #13435, opened 2020-12-04
RAC-403: Fix row behavior in job tracker behavior

tseho:

**problems:**
- the row was not clickable anymore (to view the job tracking page)
- the actions buttons were always visible (instead of only on hover)

Since the "Actions" react component is not able to tell its parent row to listen to the click, the trick is to restore the action `view` with `rowAction: true`. (thanks @ValentinMumble for the help)

Due to event propagation, when clicking on the stop button, the modal is open and the row open the view.
This is why I've updated `_listenToBodyEvents` to ensure that listened events are not already stopped.

For the hover, I've restored the css class handling this behavior.

Screenshot_2020-12-04_14-16-51

Screenshot_2020-12-04_14-17-07


SamirBoulil (approved):

StevenVAIDIE (approved):

src/Akeneo/Platform/Bundle/ImportExportBundle/Resources/config/datagrid/job_tracker.yml
```diff
@@ -49,6 +49,11 @@ datagrid:
                     isStoppable: boolean
                     showLink: string
                 data_name: id
+        actions:
```
StevenVAIDIE, line 52: You should do the same thing on the last execution (job instance view)

ValentinMumble (approved):

PR #13464, opened 2020-12-08
CPM-96: Set default values when creating a product or product model

mmetayer:

Given a boolean attribute with a `default_value`, this PR adds the default value at product / product model creation. Any existing value for this attribute will not be overridden.

I chose to implement it in a PRE_SAVE subscriber; this has pros and cons:

**pros**
- we are positive that the event is triggered when persisting a product or model, as the product(model)saver is the only mandatory service called (for instance we don't have to use the ProductUpdater, even if we do it 99% of the time)
- most entities related to the product (attributes, family...) are already loaded in the UnitOfWork at that point, so in a vast majority of the cases it should not trigger any additional SQL query

**cons**
- the product is updated after validation (and is not re-validated). I couldn't find a use case where adding a boolean value could break the validation
- the event subscriber must be executed before ComputeRawValuesEventSubscriber, therefore it must have a high priority. It might indicate that the computation of raw values should be directly executed in the saver instead of in an event subscriber.
- sometimes we don't want to add the default values. This is the case for product duplication, and for published products. I had to create an option that is passed to the subscriber in order to handle such cases (see the EE PR). I'm not very happy with that, but I wasn't able to find a cleaner way

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Akeneo/Pim/Enrichment/Bundle/EventSubscriber/EntityWithValues/AddDefaultValuesSubscriber.php
+++ b/src/Akeneo/Pim/Enrichment/Bundle/EventSubscriber/EntityWithValues/AddDefaultValuesSubscriber.php
@@ -4,7 +4,7 @@ declare(strict_types=1);
 
 namespace Akeneo\Pim\Enrichment\Bundle\EventSubscriber\EntityWithValues;
 
-use Akeneo\Channel\Component\Query\GetChannelCodeWithLocaleCodesInterface;
+use Akeneo\Channel\Component\Query\PublicApi\GetChannelCodeWithLocaleCodesInterface;
 use Akeneo\Pim\Enrichment\Component\Product\Factory\ValueFactory;
 use Akeneo\Pim\Enrichment\Component\Product\Model\EntityWithFamilyInterface;
 use Akeneo\Pim\Enrichment\Component\Product\Model\EntityWithFamilyVariantInterface;
@@ -94,9 +94,7 @@ class AddDefaultValuesSubscriber implements EventSubscriberInterface
             fn (AttributeInterface $attribute): string => $attribute->getCode()
         )->toArray();
 
-        return empty($attributeCodesWithDefaultValues) ?
-            [] :
-            $this->getAttributes->forCodes($attributeCodesWithDefaultValues);
+        return $this->getAttributes->forCodes($attributeCodesWithDefaultValues);
     }
 
     private function addValues(EntityWithFamilyVariantInterface $entity, Attribute $attribute): void
--- a/src/Akeneo/Pim/Enrichment/Bundle/EventSubscriber/EntityWithValues/AddDefaultValuesSubscriber.php
+++ b/src/Akeneo/Pim/Enrichment/Bundle/EventSubscriber/EntityWithValues/AddDefaultValuesSubscriber.php
@@ -1,0 +1,168 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Akeneo\Pim\Enrichment\Bundle\EventSubscriber\EntityWithValues;
+
+use Akeneo\Channel\Component\Query\GetChannelCodeWithLocaleCodesInterface;
+use Akeneo\Pim\Enrichment\Component\Product\Factory\ValueFactory;
+use Akeneo\Pim\Enrichment\Component\Product\Model\EntityWithFamilyInterface;
+use Akeneo\Pim\Enrichment\Component\Product\Model\EntityWithFamilyVariantInterface;
+use Akeneo\Pim\Structure\Component\Model\AttributeInterface;
+use Akeneo\Pim\Structure\Component\Query\PublicApi\AttributeType\Attribute;
+use Akeneo\Pim\Structure\Component\Query\PublicApi\AttributeType\GetAttributes;
+use Akeneo\Tool\Component\StorageUtils\StorageEvents;
+use Doctrine\Common\Collections\ArrayCollection;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\EventDispatcher\GenericEvent;
+
+/**
+ * @copyright 2020 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ *
+ * Adds default values (defined at attribute level) for products and product models
+ */
+class AddDefaultValuesSubscriber implements EventSubscriberInterface
+{
+    private GetAttributes $getAttributes;
+    private ValueFactory $valueFactory;
+    private GetChannelCodeWithLocaleCodesInterface $getChannelWithLocales;
+
+    private ?array $cachedChannelsAndLocales = null;
+
+    public function __construct(
+        GetAttributes $getAttributes,
+        ValueFactory $valueFactory,
+        GetChannelCodeWithLocaleCodesInterface $getChannelWithLocales
+    ) {
+        $this->getAttributes = $getAttributes;
+        $this->valueFactory = $valueFactory;
+        $this->getChannelWithLocales = $getChannelWithLocales;
+    }
+
+    public static function getSubscribedEvents()
+    {
+        return [
+            StorageEvents::PRE_SAVE => ['addDefaultValues', 100],
+        ];
+    }
+
+    public function addDefaultValues(GenericEvent $event)
+    {
+        $entity = $event->getSubject();
+
+        if (!$entity instanceof EntityWithFamilyInterface) {
+            return;
+        }
+        if (!$event->hasArgument('is_new') || true !== $event->getArgument('is_new')) {
+            return;
+        }
+        if ($event->hasArgument('add_default_values') && false === $event->getArgument('add_default_values')) {
+            return;
+        }
+
+        $attributesWithDefaultValue = $this->getAttributesWithDefaultValues($entity);
+        foreach ($attributesWithDefaultValue as $attribute) {
+            $this->addValues($entity, $attribute);
+        }
+    }
+
+    /**
+     * Returns all the attributes with a default_value property for a given entity:
+     * - for a simple product, the attribute must be part of the family
+     * - for a variant product or product model, the attribute must be part of the matching variant attribute set
+     *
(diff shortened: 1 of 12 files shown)


src/Akeneo/Pim/Enrichment/Bundle/EventSubscriber/EntityWithValues/AddDefaultValuesSubscriber.php
```diff
@@ -0,0 +1,168 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Akeneo\Pim\Enrichment\Bundle\EventSubscriber\EntityWithValues;
+
+use Akeneo\Channel\Component\Query\GetChannelCodeWithLocaleCodesInterface;
```
mmetayer, line 7: maybe we could move this in the `Akeneo\Channel\Component\Query\PublicApi` namespace

tests/back/Pim/Enrichment/Specification/Bundle/EventSubscriber/EntityWithValues/AddDefaultValuesSubscriberSpec.php
```diff
@@ -0,0 +1,430 @@
...
+
+        $this->addDefaultValues(
+            new GenericEvent($entity->getWrappedObject(), ['is_new' => true])
+        );
+    }
+
+    function it_adds_default_values_to_a_variant_product(
+        GetAttributes $getAttributes,
+        ValueFactory $valueFactory,
+        ProductModelInterface $product,
```
mmetayer, line 161: to fix

src/Akeneo/Pim/Enrichment/Bundle/EventSubscriber/EntityWithValues/AddDefaultValuesSubscriber.php
```diff
@@ -0,0 +1,168 @@
...
+    }
+
+    public function addDefaultValues(GenericEvent $event)
+    {
+        $entity = $event->getSubject();
+
+        if (!$entity instanceof EntityWithFamilyInterface) {
+            return;
+        }
+        if (!$event->hasArgument('is_new') || true !== $event->getArgument('is_new')) {
```
pierallard, line 57: I don't like random strings used as options. Maybe you could add a const in the ProductSaver, IS_NEW, or something like this... (same for add_default_values)

src/Akeneo/Pim/Enrichment/Bundle/EventSubscriber/EntityWithValues/AddDefaultValuesSubscriber.php
```diff
@@ -0,0 +1,168 @@
...
+            $attributes = $entity->getFamily()->getAttributes();
+        }
+
+        $attributeCodesWithDefaultValues = $attributes->filter(
+            fn (AttributeInterface $attribute): bool => null !== $attribute->getProperty('default_value')
+        )->map(
+            fn (AttributeInterface $attribute): string => $attribute->getCode()
+        )->toArray();
+
+        return empty($attributeCodesWithDefaultValues) ?
```
pierallard, line 97: empty() is not your responsibility, it's the one of the fetcher (which already early returns [])

pierallard: (I'm pinailling)

src/Akeneo/Pim/Enrichment/Bundle/EventSubscriber/EntityWithValues/AddDefaultValuesSubscriber.php
```diff
@@ -0,0 +1,168 @@
...
+    }
+
+    private function getChannelCodes(): array
+    {
+        $this->initializeChannelsAndLocales();
+
+        return array_keys($this->cachedChannelsAndLocales);
+    }
+
+    private function initializeChannelsAndLocales(): void
```
pierallard, line 158: I don't get the meaning of a cache here. As far as I see, you only call these methods one time ?

tests/back/Pim/Enrichment/Integration/Product/CreateEntityWithDefaultValuesIntegration.php
```diff
@@ -0,0 +1,243 @@
...
+    {
+        $product = $this->createProduct('new_product', ['family' => 'familyA']);
+        Assert::assertNull($product->getValue('a_yes_no_with_default_value'));
+
+        $product = $this->saveProduct($product);
+
+        $value = $product->getValue('a_yes_no_with_default_value');
+        Assert::assertInstanceOf(ValueInterface::class, $value);
+        Assert::assertSame(true, $value->getData());
+    }
```
pierallard, line 33: I don't see any tests for the case of a boolean with a "null" default value. I'm afraid of the createByCheckingData with "null"

pierallard (approved):

src/Akeneo/Pim/Enrichment/Bundle/EventSubscriber/EntityWithValues/AddDefaultValuesSubscriber.php
mmetayer: which fetcher are you talking about :thinking: ? I'm the one filtering the attributes here, the 'empty' **is** my responsibility

src/Akeneo/Pim/Enrichment/Bundle/EventSubscriber/EntityWithValues/AddDefaultValuesSubscriber.php
mmetayer: it's called at least once per new product (the subscriber is also triggered in bulk save)

tests/back/Pim/Enrichment/Integration/Product/CreateEntityWithDefaultValuesIntegration.php
mmetayer: attributes with a null default value are filtered out, see https://github.com/akeneo/pim-community-dev/pull/13464/files/f2ecaf6a26ec4c0cd7045267c6587d31ce70ca9e#diff-e94136833df468ac6c5263917132987de6778da5cca88a3d90d99bcb68abca83R92

src/Akeneo/Pim/Enrichment/Bundle/EventSubscriber/EntityWithValues/AddDefaultValuesSubscriber.php
mmetayer: I didn't add the "is_new" option, it's been there since 2017, as for "add_default_values" I think it's not the responsibility of the saver to define it.

tests/back/Pim/Enrichment/Integration/Product/CreateEntityWithDefaultValuesIntegration.php
mmetayer: Also, every boolean attribute of every catalog has a "null" default value, so if this doesn't work there will be at least 1000 tests failing :slightly_smiling_face:

src/Akeneo/Pim/Enrichment/Bundle/EventSubscriber/EntityWithValues/AddDefaultValuesSubscriber.php
pierallard: I'm talking of the getAttributes->forCodes.
If returns [] if empty.

mmetayer: arf, ok, I can change that

src/Akeneo/Pim/Enrichment/Bundle/EventSubscriber/EntityWithValues/AddDefaultValuesSubscriber.php
pierallard: I never saw this add_default_values and thought you introduced it.

mmetayer: I did introduce the "add_default_values", but not the "is_new" :slightly_smiling_face:

nmarniesse (approved):

src/Akeneo/Pim/Enrichment/Bundle/EventSubscriber/EntityWithValues/AddDefaultValuesSubscriber.php
```diff
@@ -0,0 +1,166 @@
...
+            $attributes = $entity->getFamily()->getAttributes();
+        }
+
+        $attributeCodesWithDefaultValues = $attributes->filter(
+            fn (AttributeInterface $attribute): bool => null !== $attribute->getProperty('default_value')
+        )->map(
+            fn (AttributeInterface $attribute): string => $attribute->getCode()
+        )->toArray();
+
+        return $this->getAttributes->forCodes($attributeCodesWithDefaultValues);
```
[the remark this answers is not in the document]
mmetayer: It's because the ValueFactory needs a PublicApi Attribute (`Akeneo\Pim\Structure\Component\Query\PublicApi\AttributeType`), and not an attribute entity (`Akeneo\Pim\Structure\Component\Model`)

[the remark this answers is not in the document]
nmarniesse: Ok! Idea for the future it could be great to convert a `Akeneo\Pim\Structure\Component\Model\Attribute` into a `Akeneo\Pim\Structure\Component\Query\PublicApi\AttributeType\Attribute` directly. Here we already have the data, we refetch them to have them in another object, maybe we can avoid that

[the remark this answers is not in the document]
mmetayer: yes, that's a good idea

PR #13472, opened 2020-12-08
DAPI-1458: Cleaning

LaurentPetard:
- Remove deprecated MySQL tables
- Remove metrics logs


olivierpontier (approved):

AymericPlanche (approved):
Wait Thuesday before merge  ;)

AymericPlanche:
I only think about it right now, we can also remove the step that try to delete evaluation of deleted products as we have a delete cascade constraint from now.

LaurentPetard (commented):
Refresh

PR #13476, opened 2020-12-09
PIM-9584 add dual indexation capabilities

nmarniesse:
To reindex documents in live (= without service interruption) we have to:
- keep the actual alias
- create a new alias and start the complete reindexing on it

Both indexes should be impacted when the PIM user wants to alter its document. This PR implements a dual indexation to solve this subject.  

The dual indexation can be activated by configuration. You just have to create a new client service for the new alias and add a `activate_dual_indexation_with_service` option to the "main" service:

```yaml
-- Before
akeneo_elasticsearch:
    hosts: "%index_hosts%"
    indexes:
        -
            service_name: "akeneo_elasticsearch.client.example"
            index_name: "%example%"
            configuration_files: "%example%"

-- After
akeneo_elasticsearch:
    hosts: "%index_hosts%"
    indexes:
        -
            service_name: "akeneo_elasticsearch.client.example"
            index_name: "%example%"
            configuration_files: "%example%"
            activate_dual_indexation_with_service: "akeneo_elasticsearch.client.example.new"
        -
            service_name: "akeneo_elasticsearch.client.example.new"
            index_name: "%example%_new"
            configuration_files: "%example%"
```

**Important**: The new alias is not automatically created It's your responsability to create the alias with the expected configuration (see for example this PR https://github.com/akeneo/pim-enterprise-dev/pull/10076 when we create the alias with a migration)


ahocquard (commented):
To see in review: I think we can hardcode some things.
Also, to check with the onboarder that is using this part of the code for multi-tenancy in ES index.

src/Akeneo/Tool/Bundle/ElasticsearchBundle/SwitchClientIndexer.php
```diff
@@ -0,0 +1,152 @@
...
+
+    private Client $client;
+    private VersionProviderInterface $versionProvider;
+    private string $mode;
+    private ?Client $clientForNextIndex = null;
+
+    public function __construct(
+        Client $client,
+        VersionProviderInterface $versionProvider,
+        string $mode = 'only_current_index'
```
AymericPlanche, line 42:
```suggestion
        string $mode = self::ONLY_CURRENT_INDEX_MODE
```

ahocquard:
@nmarniesse 

Is it not possible to do something like:

```
        -
            service_name: "akeneo_elasticsearch.client.example"
            index_name: "%example%"
            configuration_files: "%example%"
            with_dual_index: true
```

And create the correct services for that? It would ease the creation of an index and the alias could be hardcoded as `%example%_new` anyway. Convention over configuration.

nmarniesse:
> @nmarniesse
> 
> Is it not possible to do something like:
> 
> ```
>         -
>             service_name: "akeneo_elasticsearch.client.example"
>             index_name: "%example%"
>             configuration_files: "%example%"
>             with_dual_index: true
> ```
> 
> And create the correct services for that? It would ease the creation of an index and the alias could be hardcoded as `%example%_new` anyway. Convention over configuration.

I like your solution but the `akeneo_elasticsearch.client.example.new` service will not be explicitely defined. We can create the service automagicly but it's more clear like that IMHO.  

In our use case the "new" service is injected here https://github.com/akeneo/pim-enterprise-dev/pull/10076/files#diff-18e93a21dd62a2f19345eeb01a43d59dc15c0d39cc78025c7d3231d1975a9f62R32

ahocquard:
@nmarniesse yes, I thought about creating it in the DI. This way, just adding/remvoving the key would be enough. Maybe too much magic. On the other side less dangerous for the developer to not forget about the DI.

ahocquard (approved):

src/Akeneo/Tool/Bundle/ElasticsearchBundle/DependencyInjection/Configuration.php
```diff
@@ -56,6 +56,9 @@ public function getConfigTreeBuilder()
                                 ->defaultValue('')
                                 ->info('Prefix all document ids')
                             ->end()
+                            ->scalarNode('activate_dual_indexation_with_service')
+                                ->info('Activate dual indexation using the service specified as value')
```
jmleroux, line 60:
```suggestion
                                ->info('Activate dual indexation using the specified service as value')
```

jmleroux (approved):

PR #13488, opened 2020-12-10
installation doesn't create migration table when there are no migrations to play

rfauglas:

Tested with docker-compose run -u www-data --rm php php bin/console --env=test pim:installer:db and docker-compose run -u www-data --rm php php bin/console --pim:installer:db

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Akeneo/Platform/Bundle/InstallerBundle/Command/DatabaseCommand.php
+++ b/src/Akeneo/Platform/Bundle/InstallerBundle/Command/DatabaseCommand.php
@@ -31,6 +31,7 @@ use Symfony\Component\Process\Process;
  */
 class DatabaseCommand extends Command
 {
+    const NO_MIGRATION_OUTPUT = "0\n";
     protected static $defaultName = 'pim:installer:db';
 
     const LOAD_ALL = 'all';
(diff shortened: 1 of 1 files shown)


src/Akeneo/Platform/Bundle/InstallerBundle/Command/DatabaseCommand.php
```diff
@@ -356,4 +362,13 @@ protected function launchCommands()
 
         return $this;
     }
+
+    /**
```
nmarniesse, line 366: You can skip docblock as the argument and the return of the method are type hinted

nmarniesse (approved):

src/Akeneo/Platform/Bundle/InstallerBundle/Command/DatabaseCommand.php
```diff
@@ -316,6 +317,11 @@ private function setLatestKnownMigration(InputInterface $input): void
     {
         $latestMigration = $this->getLatestMigration($input);
 
+        if ($this->hasNoMigrationToExecute($latestMigration)) {
```
BitOne, line 320: It's not normal that there's no migration to execute...

src/Akeneo/Platform/Bundle/InstallerBundle/Command/DatabaseCommand.php
```diff
@@ -31,6 +31,7 @@
  */
 class DatabaseCommand extends Command
 {
+    const NO_MIGRATION_OUTPUT = "0\n";
```
jmleroux, line 34: Too hacky for me and mostly we don't explain why there is no migration

jmleroux (changes requested):
Maybe explain in the description why there can be no migration?

PR #13491, opened 2020-12-11
PIM-9585: Display the product models codes on the warnings

anaisbaunelemaire:

When importing a file, if warnings are raised with products, the identifier of each problematic product is given on the warnings 
With product models, the warning didn't display the code of the PM.

So, we added the product model code as a value to display in the process function of the ProductModelProcessor class.

**Explanation**

When `$jobParameters->get('enabledComparison') && null !== $productModel->getId()` is true, meaning that `enabledComparison` is enabled and a product model with this code already exists, the following line is executed:
https://github.com/akeneo/pim-community-dev/blob/07faafa9934bdc43c29f0cf49a9e983a772c03ae/src/Akeneo/Pim/Enrichment/Component/Product/Connector/Processor/Denormalizer/ProductModelProcessor.php#L119

Removing the code from `$standardProductModel`.
When a validation error happens, the warning cannot display the code since it's not there anymore.
The solution is to restore the code after this filter.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Akeneo/Pim/Enrichment/Component/Product/Connector/Processor/Denormalizer/ProductModelProcessor.php
+++ b/src/Akeneo/Pim/Enrichment/Component/Product/Connector/Processor/Denormalizer/ProductModelProcessor.php
@@ -102,7 +102,7 @@ class ProductModelProcessor extends AbstractProcessor implements ItemProcessorIn
             return null;
         }
 
-        if (!isset($standardProductModel['code'])) {
+        if (empty($code)) {
             $this->skipItemWithMessage($standardProductModel, 'The code must be filled');
         }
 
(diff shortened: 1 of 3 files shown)


src/Akeneo/Pim/Enrichment/Component/Product/Connector/Processor/Denormalizer/ProductModelProcessor.php
```diff
@@ -140,13 +141,15 @@ public function process($standardProductModel): ?ProductModelInterface
         } catch (PropertyException $exception) {
             $this->objectDetacher->detach($productModel);
             $message = sprintf('%s: %s', $exception->getPropertyName(), $exception->getMessage());
+            $standardProductModel['code'] = $code;
```
rouliane, line 144: I think you may have missed one use case Line 135, there is another skipItemWithMessage() call

anaisbaunelemaire:
Adding a spec is not possible because we expect an exception to be thrown. This exception includes an instance of FileInvalidItem which cannot be mocked. (cc @jmleroux )

src/Akeneo/Pim/Enrichment/Component/Product/Connector/Processor/Denormalizer/ProductModelProcessor.php
```diff
@@ -158,7 +168,7 @@ public function process($standardProductModel): ?ProductModelInterface
      *
      * @return ProductModelInterface
      */
-    private function findOrCreateProductModel(string $code): ProductModelInterface
+    public function findOrCreateProductModel(string $code): ProductModelInterface
```
jmleroux, line 171: Why changing that?

jmleroux (approved):
And what about an integration test?

src/Akeneo/Pim/Enrichment/Component/Product/Connector/Processor/Denormalizer/ProductModelProcessor.php
anaisbaunelemaire: It's a mistake

PR #13493, opened 2020-12-11
CXP-599: count the Events API requests sent

tomglvng:
Count the Events API requests sent by SendEventsToWebhookHandler

Add counter and increment after yield

Call service EventsApiRequestCounter(\DateTime $dateTime, int $count) before logger info
incrementCount(\DateTime $dateTime, int $count)
getCurrentCount(\DateTime $dateTime)

integration test (test that rolling hour is working, aka multiple count per minute on more than one minute)

e2e test (test only that the number of requests sent match the count added)


src/Akeneo/Connectivity/Connection/back/Application/Webhook/Command/SendBusinessEventToWebhooksHandler.php
```diff
@@ -83,6 +87,7 @@ public function handle(SendBusinessEventToWebhooksCommand $command): void
         $event = $command->event();
 
         $requests = function () use ($event, $webhooks) {
+            $apiEventsRequestCount = 0;
```
pierrejolly, line 90: What is the difference between `$apiEventsRequestCount`  and `$eventBuiltCount` ?

pchasle: Good point. That's the same thing. I think we can merge them and keep only `$apiEventsRequestCount`

src/Akeneo/Connectivity/Connection/back/Infrastructure/Persistence/Dbal/Repository/DbalEventsApiRequestCountRepository.php
```diff
@@ -0,0 +1,46 @@
...
+class DbalEventsApiRequestCountRepository implements EventsApiRequestCountRepository
+{
+    private DbalConnection $dbalConnection;
+
+    public function __construct(DbalConnection $dbalConnection)
+    {
+        $this->dbalConnection = $dbalConnection;
+    }
+
+    public function upsert(\DateTimeImmutable $dateTime, int $eventCount): int
```
pchasle, line 24:
```suggestion
    /**
     * Update the number of Events API requests for the current hour & minute.
     * `$dateTime` minute is used to determine the minute to update and to set the last `updated` date.
     * If the previous `updated` date is not from the current hour, then the count is reinitialized with the new count.
     */
    public function upsert(\DateTimeImmutable $dateTime, int $eventCount): int
```
Suggesting a little description to explain the query.

pchasle (approved):

pierrejolly (approved):

PR #13502, opened 2020-12-14
RAC-395: Validation error with direct link to job only on internal API

StevenVAIDIE:
In this PR, change the validation error to give the direct link to job only on internal API.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Akeneo/Pim/Enrichment/Component/Product/Normalizer/InternalApi/ConstraintViolationNormalizer.php
+++ b/src/Akeneo/Pim/Enrichment/Component/Product/Normalizer/InternalApi/ConstraintViolationNormalizer.php
@@ -27,11 +27,13 @@ class ConstraintViolationNormalizer implements NormalizerInterface, CacheableSup
         $path = $this->getStandardPath($violation);
         $translate = (bool)($context['translate'] ?? true);
 
+        $message = $this->getInternalApiMessage($violation);
+
         if (!$translate) {
             return [
                 'messageTemplate' => $violation->getMessageTemplate(),
                 'parameters' => $violation->getParameters(),
-                'message' => $violation->getMessage(),
+                'message' => $message,
                 'propertyPath' => $path,
                 'invalidValue' => $violation->getInvalidValue(),
             ];
@@ -39,14 +41,14 @@ class ConstraintViolationNormalizer implements NormalizerInterface, CacheableSup
 
         if (null === $path || '' === $path) {
             return [
-                'message' => $violation->getMessage(),
+                'message' => $message,
                 'global'  => true,
             ];
         }
 
         return [
             'path'    => $path,
-            'message' => $violation->getMessage(),
+            'message' => $message,
             'global'  => false,
         ];
     }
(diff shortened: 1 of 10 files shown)


tseho:
There is no context like `internal_api` in the constraint ?
It's something we do for the normalizers https://github.com/akeneo/pim-community-dev/pull/13502/files#diff-2327788c38e025215f97d7fb95d168aa9b5f59e8024d8204a20f2732079b75cdR19
The constraint would then return only one message, and the correct one by being aware of the context.
You wouldn't have to change the normalizer.

src/Akeneo/Pim/Enrichment/Component/Product/Normalizer/InternalApi/ConstraintViolationNormalizer.php
```diff
@@ -93,4 +95,17 @@ protected function getStandardPath(ConstraintViolation $violation)
 
         return Inflector::tableize($violation->getPropertyPath());
     }
+
+    private function getInternalApiMessage(ConstraintViolation $violation): string
+    {
+        if ( ! $violation->getConstraint()) {
```
ValentinMumble, line 101:
```suggestion
        if (!$violation->getConstraint()) {
```

src/Akeneo/Pim/Structure/Component/Validator/Constraints/BlacklistedAttributeCodeValidator.php
```diff
@@ -31,8 +46,22 @@ public function validate($attributeCode, Constraint $constraint): void
...
     }
+
+    private function addInternalViolation(string $attributeCode, BlacklistedAttributeCode $constraint): void
+    {
+        $jobExecutionId = $this->getBlacklistedAttributeJobExecutionId->forAttributeCode($attributeCode);
+        $internalApiMessage = $this->translator->trans(
+            $constraint->internalAPIMessage,
+            ['{{ link }}' => "#" . $this->router->generate(self::JOB_TRACKER_ROUTE, ['id' => $jobExecutionId])],
+            'validators'
+        );
```
ValentinMumble, line 63:
```suggestion
        $jobExecutionLink = sprintf('#%s', $this->router->generate(self::JOB_TRACKER_ROUTE, ['id' => $jobExecutionId]));
        $internalApiMessage = $this->translator->trans(
            $constraint->internalAPIMessage,
            ['{{ link }}' => $jobExecutionLink],
            'validators'
        );
```

ValentinMumble (approved):

juliensnz (approved):

StevenVAIDIE:
> There is no context like `internal_api` in the constraint ?
> It's something we do for the normalizers https://github.com/akeneo/pim-community-dev/pull/13502/files#diff-2327788c38e025215f97d7fb95d168aa9b5f59e8024d8204a20f2732079b75cdR19
> The constraint would then return only one message, and the correct one by being aware of the context.
> You wouldn't have to change the normalizer.

Unfortunately, it's not possible. The only thing that we can pass through the validator is groups. We cannot use them here because Attribute class have GroupSequenceProvider the order of group given by Controller are not keeped and the group in the context on Validator is always Attribute.

The right way is to don't translated message and keep the front doing the translation. But, it requires reworking the front to deal with this (common field on UiBundle)

PR #13513, opened 2020-12-14
CXP-600: The Events API should be able to rate limit the number of delivery

tomglvng:
We want to be able to implement a “rate limiting” on the Events API.
Rate limiting would be “per hour”.


config/services/pim_parameters.yml
```diff
@@ -78,3 +78,5 @@ parameters:
     webhook_timeout: 0.5
     webhook_max_bulk_size: 10
     webhook_active_event_subscriptions_limit: 3
+    webhook_requests_limit: 1000000
```
pchasle, line 81:
```suggestion
    webhook_requests_limit: 2500
```

pchasle (approved):

pierrejolly (approved):

PR #13524, opened 2020-12-16
Update config for Xdebug v3

mmetayer:

Update the config to handle Xdebug v3.

Xdebug can now be activated like this:

- for the webserver (UI/API):
```console
XDEBUG_MODE=debug APP_ENV=dev make up
```

- for a cli command:
```console
XDEBUG_MODE=debug APP_ENV=test docker-compose run --rm -u www-data php vendor/bin/phpunit path/to/my/Test.php
```

Note: the various values for xdebug.mode are listed [here](https://xdebug.org/docs/all_settings#mode)


jmleroux (approved):

make-file/dev.mk
```diff
@@ -40,11 +40,3 @@ phpunit:
 .PHONY: behat-legacy
 behat-legacy:
 	APP_ENV=behat $(PHP_RUN) vendor/bin/behat -p legacy -s all ${F}
-
-.PHONY: xdebug-on
-xdebug-on:
-	XDEBUG_ENABLED=1 $(MAKE) up
```
SamirBoulil, line 46: Why do we remove those targets ? We could just updated them with XDEBUG_MODE=debug ?

mmetayer: they're a frequent source of errors, as if you use them without specifying the APP_ENV environment variable they will start the containers in `prod` mode, which we almost never want.
Also, there are [several modes](https://xdebug.org/docs/all_settings#mode) to xdebug now, I think it's better to specifically tell which one(s) you wanna use, e.g `XDEBUG_MODE="develop,trace,gcstats" APP_ENV=dev make up`

nmarniesse: Samir it's related to this https://github.com/akeneo/pim-enterprise-dev/pull/10103#discussion_r543559000  

But I would keep it too, and adapt it to use `XDEBUG_MODE=debug APP_ENV=dev $(MAKE) up`

damien-carcel (approved):

jmleroux:
GTM with admin rights

make-file/dev.mk
jmleroux: Could we not think about a "default mode"? So we can override it. Still dunno how to make it...

PR #13531, opened 2020-12-16
Do not update products or models when quantified associations were not updated

mmetayer:

Last PR in the series of _**Do not save a product or product model if it was not actually updated**_ :slightly_smiling_face: 

Here we compare the quantified association collection before and after calling methods which potentially update it to compute the state of the product / model (aka `isDirty()`). In order to do so, I added the `QuantifiedAssociationCollection::equals()` public method

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Akeneo/Pim/Enrichment/Component/Product/Model/AbstractProduct.php
+++ b/src/Akeneo/Pim/Enrichment/Component/Product/Model/AbstractProduct.php
@@ -832,16 +832,27 @@ abstract class AbstractProduct implements ProductInterface
     /**
      * {@inheritdoc}
      */
+    public function getQuantifiedAssociations(): QuantifiedAssociationCollection
+    {
+        return clone $this->quantifiedAssociationCollection;
+    }
+
+    /**
+     * {@inheritdoc}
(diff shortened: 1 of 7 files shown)


src/Akeneo/Pim/Enrichment/Component/Product/Model/AbstractProduct.php
```diff
@@ -838,10 +846,13 @@ public function filterQuantifiedAssociations(array $productIdentifiersToKeep, ar
             return;
         }
 
+        $initialCollection = $this->getQuantifiedAssociations();
```
mmetayer, line 849: we could use `$initialCollection = $this->quantifiedAssociationCollection;` since the QuantifiedAssociationCollection class is immutable, but I feel that in the product scope, we don't "know" it's immutable, so I chose to use a clone

StevenVAIDIE (approved):

ahocquard (approved):

nmarniesse (approved):

willy-ahva (approved):

PR #13545, opened 2020-12-17
CXP-597 make event dispatching more resilient

pchasle:
Make event dispatching more resilient to infrastructure issues.
If there is an issue with Google PubSub (for example) then we should not necessarily break the PIM behavior and/or the UI. 
Instead the error is log as critical (still monitored by the SRE).


willy-ahva (approved):

pchasle (commented):
Unfroze PR  ...

tomglvng (approved):

PR #13551, opened 2020-12-17
CPM-101: remove doctrine cache bundle and update doctrine packages versions

nmarniesse:
**Doctrine cache bundle**

Doctrine cache bundle is no longer maintained: https://github.com/doctrine/DoctrineCacheBundle/issues/156

Since symfony 4.3 the best way to configure the cache is using the `symfony/cache` component

In this PR the doctrine-cache-bundle is removed and replaced by symfony/cache. The config is migrated to this new component.

**Update doctrine packages versions**

One of the initial goal was to remove the `doctrine/reflection` package (which is abandoned). But for now thee is some issues to do that: to remove it we need to update `doctrine/persistence` in 2.x and `doctrine/common` in 3.x. But there is some issues with the `friendsofsymfony/rest-bundle` that is the higher version available for now.  

Major version upgrade: `doctrine/doctrine-migrations-bundle` and `doctrine/migrations` are bumped from 1.x to 2.x. The config is adapted. Now the `migrate_version` table has a new `executed_at` column. The column is added automatically no need to do additional work. 
The BC break about `Doctrine\DBAL` namespace is fixed in this PR (see https://github.com/doctrine/migrations/releases/tag/v2.0.0 for further information)


config/packages/doctrine.yml
```diff
@@ -31,18 +31,40 @@ doctrine:
...
-doctrine_cache:
-    providers:
-        pim_apcu:
-            apcu: ~
-        pim_array:
-            array: ~
-        pim:
-            chain:
-                providers: [doctrine_cache.providers.pim_apcu, doctrine_cache.providers.pim_array]
+    column_length: 255
```
nmarniesse, line 70: Length fior the `version` column of the `migration_versions` table. The default for `doctrine/migration` v2 is 14. That means none of our igration would work....  
The length was 255 for the v1

nmarniesse:
Splitted

PR #13554, opened 2020-12-17
PIM-9601: Fix completeness mask memory usage

jmleroux:
Completeness can takes a lots of space in memory due to masks, locales, etc.
This PR is a first step to reduce the required memory by processing the completeness by chunks.

Next step will be to use iterable generators which should not need any arbitrary chunk size.


Doodoune:
Don't forget to add a changelog ;)

mmetayer (approved):

nmarniesse (approved):
GTM with a changelog

SamirBoulil (approved):

PR #13578, opened 2020-12-21
CPM-100: remove Symfony deprecations

nmarniesse:
Remove usage to Symfony deprecations. It would ease the upgrade to Symfony 5.

**TranlatorInterface**

Replace deprecated `Symfony\Component\Translation\TranslatorInterface` by `Symfony\Contracts\Translation\TranslatorInterface`

**Factory service definition**

Fix factory format in service for example `factory: 'doctrine.orm.entity_manager:getRepository'` is replaced by `factory: ['@doctrine.orm.entity_manager', 'getRepository']`

**Event dispatcher**

Fix arguments order in the event dispatcher. In SF5 the object is in first position, then event name.
In SF4 the order did not matter

**Inflector**

In SF5 the `Symfony\Component\Inflector\Inflector` class is deprecated. We have to use the doctrine Inflector instead

**Templating service**

The `@templating` service is deprecated. We have to use `@twig` instead

**GetResponseEvent**

`Symfony\Component\HttpKernel\Event\GetResponseEvent` is deprecated

**Dotenv constructor**

The constructor accepts one optional argument: `$usePutenv`. When true the env variables are also defined using `putenv()`. So the `getenv()` method can work in all our application. When `false` we can obtain the value by injecting the parameter via the DI (best way when possible) or use `$_ENV[]`.

In SF4 the `$usePutenv` default is `true`. In SF5 it's `false`. From symfony: `Beware that putenv() is not thread safe`  

To avoid the usage of `putenv()` we must replace all the usage of `getenv()`. I did the change in our code but unfortunately some dependencies use and need `getenv()` to work. For instance `google/cloud-core` and `google/cloud-pubsub` need it. So the solution is to allow `putenv()` for now.


mmetayer (approved):

src/Akeneo/Platform/Bundle/UIBundle/Translator/TranslatorDecorator.php
```diff
@@ -32,12 +34,15 @@ public function trans($id, array $parameters = [], $domain = null, $locale = nul
     }
 
     /**
-     * {@inheritdoc}
+     * Deprecated since Symfony 4.2, will be removed in Symfony 5.
+     * Use trans instead, see Symfony\Contracts\Translation\TranslatorInterface
```
nmarniesse, line 38: `transChoice()` is still used in some symfony 4.x components. Example: https://github.com/symfony/symfony/blob/4.4/src/Symfony/Component/Validator/Violation/ConstraintViolationBuilder.php#L174  
We cannot remove it for now.

BitOne (approved):

PR #13627, opened 2020-12-29
Update pim_parameters.yml

fitn:
This change allows sending all events for a maximum of customers increasing webhook requests limit.


ahocquard (approved):
GTM on technical part, but I trust for the consequences/impacts :)

pierrejolly (approved):

PR #13668, opened 2021-01-06
CI: Run Behat tests on EE only

jmleroux:
Behat tests will still be run every night, but we may not need them on every PR

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -648,15 +648,9 @@ workflows:
                     - test_front_static_acceptance_and_integration
                     - test_back_phpunit
 
-          - ready_to_build_only_ce?:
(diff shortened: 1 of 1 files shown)


nmarniesse (approved):

pierallard (approved):

ahocquard (commented):
I think that was the purpose of CE only context... Personally, I don't think that's a good idea, it adds even more coupling between CE and EE.

- How will you fix an issue on a nightly build in CE on behats?
- Does developer execute that often this workflow to try to optimize it?
- Creating exception in a build is a bad thing in my opinion. PIM builds are already too complex for my taste, a newcomer will have a lot of surprises with such behavior.

PR #13670, opened 2021-01-06
CPM-101: update doctrine versions to avoid deprecations

nmarniesse:
**Remove  `doctrine/reflection` dependency**

The initial goal was to remove the `doctrine/reflection` package (which is abandoned). We can remove it in composer.json but it is still installed because required by some others doctrine packages. To remove it completely we have to update `doctrine/persistence` in 2.x and `doctrine/common` in 3.x --> But there is a conflict with the `friendsofsymfony/rest-bundle` that is the higher version available for now. 
So for now we just remove it from composer.json. It is still installed and we still have a deprecation warning (in composer install for instance).  

**Remove some doctrine deprecations**

`Symfony\Bridge\Doctrine\RegistryInterface` is deprecated and replaced by `Doctrine\Persistence\ManagerRegistry`

**Update `doctrine/doctrine-bundle` dependency**

The version is updated from `1.12.0` to the last available: `2.2.2`. It will remove completely the DoctrineCacheBundle: https://github.com/doctrine/DoctrineBundle/blob/2.0.0/UPGRADE-2.0.md

As a consequence `doctrine/doctrine-migrations-bundle` and `doctrine/migrations` need to be bumped from 1.x to 2.x
The config is adapted, now the `migrate_version` table has a new `executed_at` column. The column is added automatically no need to do additional work. 

The BC break about `Doctrine\DBAL` namespace is fixed in this PR (see https://github.com/doctrine/migrations/releases/tag/v2.0.0 for further information)


nmarniesse:
PR in standby

PR #13673, opened 2021-01-07
DAPI-1296: add Pagination component

rouliane:

`OK`: Done / Validated
`-`: Not needed


akeneo-design-system/src/components/Pagination/Pagination.stories.mdx
```diff
@@ -0,0 +1,54 @@
...
+    <Story name="Standard">
+        {args => {
+            return <Pagination {...args}/>;
+        }}
+    </Story>
+</Canvas>
+
+<ArgsTable story="Standard" />
+
+## Example of render whith only a few pages
```
AymericPlanche, line 39:
```suggestion
## Example of render with only a few pages
```

akeneo-design-system/src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,181 @@
...
+import {getColor, AkeneoThemedProps} from '../../theme';
+
+type PaginationProps = {
+  /**
+   * The current page number (starting à 1).
+   */
+  currentPage: number;
+
+  /**
+   * The total number of pages.
```
AymericPlanche, line 12:
```suggestion
   * The total number of items.
```

AymericPlanche (approved):

akeneo-design-system/src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,181 @@
...
+  }
+
+  :hover {
+    background-color: ${getColor('grey', 20)};
+  }
+
+  ${props => (props.disabled ? disabledMixin : null)}
+`;
+
+function computePages(currentPage: number, numberOfPages: number) {
```
olivierpontier, line 131: May you can move this function in a dedicated script or a dedicated custom hook?

akeneo-design-system/src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,181 @@
...
+  :not(:last-child) {
+    margin-right: 10px;
+  }
+
+  :hover {
+    background-color: ${getColor('grey', 20)};
+  }
+
+  ${props => (props.disabled ? disabledMixin : null)}
+`;
```
olivierpontier, line 129: In order to simplify the reading of the code, what do you think about moving the code of the PaginationItem component in a dedicated file?

olivierpontier (approved):

akeneo-design-system/src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,181 @@
+import React, {FC} from 'react';
+import styled, {css} from 'styled-components';
+import {getColor, AkeneoThemedProps} from '../../theme';
+
+type PaginationProps = {
+  /**
+   * The current page number (starting à 1).
```
tseho, line 7:
```suggestion
   * The current page number (starting at 1).
```

akeneo-design-system/src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,181 @@
...
+
+  /**
+   * The total number of pages.
+   */
+  itemsTotal: number;
+
+  /**
+   * The number of items per page. Default to 25.
+   */
+  itemsPerPage?: number;
```
tseho, line 19: do we want a default value there ?

akeneo-design-system/src/components/Pagination/Pagination.tsx
tseho: I agree with Olivier, it would be better to isolate `PaginationItem`

akeneo-design-system/src/components/Pagination/Pagination.tsx
tseho: By isolating the `computePages` function, you can also easily unit test it and reduce the amount of tests on the component itself.

akeneo-design-system/src/components/Pagination/Pagination.stories.mdx
```diff
@@ -0,0 +1,54 @@
+import {Meta, Story, ArgsTable, Canvas} from '@storybook/addon-docs/blocks';
+import {Pagination} from './Pagination.tsx';
+
+<Meta
+    title="Components/Pagination"
+    component={Pagination}
+    argTypes={{
+    }}
```
juliensnz, line 8: Pro tips: If empty, this section can be removed :)

akeneo-design-system/src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,181 @@
...
+type PaginationProps = {
+  /**
+   * The current page number (starting à 1).
+   */
+  currentPage: number;
+
+  /**
+   * The total number of pages.
+   */
+  itemsTotal: number;
```
juliensnz, line 14:
```suggestion
  totalItems: number;
```
Seems more english to me but I'm maybe wrong

akeneo-design-system/src/components/Pagination/Pagination.tsx
juliensnz: yes could be. In the PIM it's usually 25

akeneo-design-system/src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,181 @@
...
+  itemsPerPage?: number;
+
+  /**
+   * Handler called when a pagination item is clicked.
+   */
+  onClick: (page: number) => void;
+};
+
+const MAX_PAGINATION_ITEMS_WITHOUT_SEPARATOR = 4;
+const PAGINATION_SEPARATOR = '...';
```
juliensnz, line 28:
```suggestion
const PAGINATION_SEPARATOR = '…';
```

I think the three dot sign is advised here. Context: https://musescore.org/en/node/280844

akeneo-design-system/src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,181 @@
...
+
+const Pagination: FC<PaginationProps> = ({currentPage, itemsTotal, itemsPerPage = 25, onClick}) => {
+  if (itemsPerPage <= 0) {
```
juliensnz, line 31: I think that in this case it's more a developper error and should throw an exception.
IMHO, having no item is something that can happen (if there is no content) but having pages of zero items is an error coming from the developer. What do you think?

akeneo-design-system/src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,181 @@
...
+    return <></>;
+  }
+
+  const numberOfPages = Math.ceil(itemsTotal / itemsPerPage);
+
+  if (numberOfPages <= 1 || currentPage > numberOfPages) {
```
juliensnz, line 37: Same here: having only one page is possible (if there is less than 25 items) but having a current page above the number of page in total means that there is something wrong and it should throw an error IMHO. What do you think?

akeneo-design-system/src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,181 @@
...
+    return <></>;
```
juliensnz, line 38: Pro tips: when you want to return an empty component, you should `return null;` :)

akeneo-design-system/src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,181 @@
...
+  }
+
+  const pages = computePages(currentPage, numberOfPages);
+
+  return (
+    <PaginationContainer>
+      {pages.map((page: number | string, index: number) => {
```
juliensnz, line 45: how a page could be a string?

akeneo-design-system/src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,181 @@
...
+        return (
+          <PaginationItem currentPage={page === currentPage} key={index} onClick={onClick}>
```
juliensnz, line 47: Here I think that the page would be a good candidate to be the key

akeneo-design-system/src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,181 @@
...
+  justify-content: center;
+`;
+
+type PaginationItemProps = {
+  currentPage: boolean;
+  onClick: (page: number) => void;
+};
+
+const PaginationItem: FC<PaginationItemProps> = ({currentPage, children, onClick}) => {
+  const title = `No. ${children as string}`;
```
juliensnz, line 70: I think that this string should be translatable and it feels weird to cast the children here without checking before hand. you can assert that it's a string with typeof if needed

akeneo-design-system/src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,181 @@
...
+
+  return (
+    <PaginationItemContainer
+      onClick={() => (children == PAGINATION_SEPARATOR ? null : onClick(children as number))}
```
juliensnz, line 74: In the DSM, we advise to always wrap callbacks in useCallback hook :)

akeneo-design-system/src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,181 @@
...
+      data-testid="paginationItem"
+      title={children != PAGINATION_SEPARATOR ? title : ''}
```
juliensnz, line 76:
```suggestion
      title={children !== PAGINATION_SEPARATOR ? title : ''}
```

akeneo-design-system/src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,181 @@
...
+      disabled={children == PAGINATION_SEPARATOR}
```
juliensnz, line 77:
```suggestion
      disabled={children === PAGINATION_SEPARATOR}
```

akeneo-design-system/src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,181 @@
...
+
+const disabledMixin = css`
+  cursor: default;
+  :hover {
+    background-color: ${getColor('white')};
+  }
+`;
+
+const PaginationItemContainer = styled.span<AkeneoThemedProps & PaginationItemProps & {disabled: boolean}>`
+  ${props => (props.currentPage ? currentPaginationItemMixin : otherPaginationItemMixin)}
```
juliensnz, line 103: pro tips: if you want, you can to this:
```suggestion
  ${({currentPage}) => (currentPage ? currentPaginationItemMixin : otherPaginationItemMixin)}
```

akeneo-design-system/src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,181 @@
...
+  line-height: 21px;
+  cursor: pointer;
+  font-family: inherit;
+  transition: background-color 0.1s ease 0s;
+  min-width: 40px;
+  text-align: center;
+  box-sizing: border-box;
+
+  :not(:last-child) {
+    margin-right: 10px;
```
juliensnz, line 121: Pro tips: as your parent is a `display: flex;`, you can use `gap: 10px;` to avoid having to manage the last case :)

akeneo-design-system/src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,181 @@
...
+    pages.push(THIRD_PAGE);
+  }
+
+  if (currentPage <= FOURTH_LAST) {
+    pages.push(PAGINATION_SEPARATOR);
+  }
+
+  pages.push(LAST_PAGE);
+
+  return pages;
```
juliensnz, line 178: What do you think about:
```suggestion
  const before = [
    1,
    2,
    3,
    4,
    currentPage - 1
  ];
  
  const after = [
    currentPage + 1
    numberOfPages - 3,
    numberOfPages - 2,
    numberOfPages - 1,
    numberOfPages
  ]
  
  const separator = currentPage >= 4 || currentPage <= numberOfPages - 3 ? 
    [PAGINATION_SEPARATOR] : 
    [currentPage];
  
  return arrayUnique([...before, ...separator, ...after])
}

const arrayUnique = (array: unknown[]) => {
  return [...new Set(array)]
```

juliensnz (approved):
Only small comments. Really good job guys!

akeneo-design-system/src/components/Pagination/Pagination.tsx
juliensnz: also always use `===`

akeneo-design-system/src/components/Pagination/Pagination.tsx
tseho: I've just tried yours and it doesn't pass the tests.

So I got curious and I tried to make a short one with reduce, but it's not as readable as I expected. Anyway,
```
return Array.from(new Array(total), (x, i) => i + 1)
  .reduce((pages, page) => {
    if (
      page === 1 ||
      page === total ||
      page === current ||
      page === current - 1 ||
      page === current + 1 ||
      (current === 1 && page === 3) ||
      (current === total && page === total - 2)
    ) {
      pages.push(page);
    }
  
    if (
      (page === current - 2 && page > 1 && current < total) ||
      (page === current - 3 && current === total) ||
      (page === current + 2 && page < total && total > 4)
    ) {
      pages.push('...');
    }
  
    return pages;
  }, []);
```

juliensnz: By viewing the screenshot, I realize that my implem is not right. But my point was more about the fact to avoid multiple ifs and use array instead. Maybe, I'm wrong and it's more complicated

akeneo-design-system/src/components/Pagination/Pagination.stories.mdx
rouliane: I thought so too, but if we don't specify this, the doc of the component parameters is not displayed, for some reason.

rouliane: I did a test again and I was wrong. It's just that I have to do a manuel hard refresh to make it display. So I'll remove it

akeneo-design-system/src/components/Pagination/Pagination.tsx
pchasle, line 74: From what I can see `PaginationItemContainer` are `span` and always clickable.
From an HTML semantic point-of-view they should probably be `button` instead, and non-clickable item should be `disabled` too.

pchasle: And the surrounding markup could use unordered list; take a look at the Bootstrap markup:
https://getbootstrap.com/docs/5.0/components/pagination/

akeneo-design-system/src/components/Pagination/Pagination.tsx
AymericPlanche: we're not proud of our `if` implementation, but it's the more "human readable" we've found.

akeneo-design-system/src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,116 @@
...
+   * Handler called when a pagination item is clicked.
+   */
+  onClick: (page: number) => void;
+};
+
+const MAX_PAGINATION_ITEMS_WITHOUT_SEPARATOR = 4;
+
+const Pagination: FC<PaginationProps> = ({currentPage, totalItems, itemsPerPage = 25, onClick}) => {
+  if (itemsPerPage <= 0) {
+    return <></>;
```
juliensnz, line 31: Why did you decide to not throw? and keep this return instead of null?

akeneo-design-system/src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,116 @@
...
+  }
+
+  const numberOfPages = Math.ceil(totalItems / itemsPerPage);
+
+  if (currentPage > numberOfPages) {
+    throw new Error('');
```
juliensnz, line 37: I'm not sure throwing an empty error is a good idea :/

akeneo-design-system/src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,116 @@
...
+  }
+
+  if (numberOfPages <= 1) {
+    return <></>;
```
juliensnz, line 41: Why did you decided to not throw null?

akeneo-design-system/src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,116 @@
...
+  }
+
+  const pages = computePages(currentPage, numberOfPages);
+
+  return (
+    <PaginationContainer>
+      {pages.map((page: number | string, index: number) => {
```
juliensnz, line 48:
```suggestion
      {pages.map((page: PageContent, index: number) => {
```

As you use this type multiple times, you could maybe create a dedicated type:
```
Type PageContent = string | number
```

akeneo-design-system/src/components/Pagination/Pagination.tsx
rouliane: Because I couldn't make it work. I'll try again.

akeneo-design-system/src/components/Pagination/Pagination.tsx
rouliane: It's a mistake, I'll fix it.

akeneo-design-system/src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+  /**
+   * The number of items per page. Default to 25.
+   */
+  itemsPerPage?: number;
+
+  /**
+   * Handler called when a pagination item is clicked.
+   */
+  onClick: (page: number) => void;
```
olivierpontier, line 24: What do you think about replacing `onClick` that is a very technical action by a more functional like bellow:
```suggestion
  followPage: (page: number) => void;
```

PR #13687, opened 2021-01-08
PIM-9620: Fix performance issue on API attributes partial update list

LaurentPetard:
There are a performance issue when updating attributes with the API end-point "PATCH attributes".

It's because an event POST_SAVE with the parameter `unitary` at `true` is dispatched for each attribute while it's a "batch" process. A subscriber of the DQI context listen to these events and launch computations that takes time.

The parameter `unitary` was designed for this kind of cases. It allows to distinct a single update from a multiple/batch update.

This kind of problem has already been spotted and fixed for the products. A subscriber has been added to put the parameter `unitary` to false and dispatch a POST_SAVE_ALL event.
dev/blob/master/src/Akeneo/Pim/Enrichment/Bundle/EventSubscriber/Product/OnSave/ApiAggregatorForProductPostSaveEventSubscriber.php

### DIFF
--- a/CHANGELOG-5.0.md
+++ b/CHANGELOG-5.0.md
@@ -15,6 +15,7 @@
 - DAPI-1470: Fix DateTime bad usage
 - DAPI-1469: Fix the size issue with the logo on login page
 - PIM-9622: Fix query that can generate a MySQL memory allocation error
+- PIM-9620: Fix performance issue on API attributes partial update list
 
 # 5.0.0 (2020-12-31)
 
--- a/src/Akeneo/Pim/Structure/.php_cd.php
+++ b/src/Akeneo/Pim/Structure/.php_cd.php
@@ -23,6 +23,7 @@ $rules = [
         'Oro\Bundle\FilterBundle\Datasource\FilterDatasourceAdapterInterface',
         'Oro\Bundle\FilterBundle\Filter\ChoiceFilter',
         'Oro\Bundle\PimFilterBundle\Datasource\FilterDatasourceAdapterInterface',
+        'Psr\Log\LoggerInterface',
 
         // TIP-906: Functional problem to query products before removing AttributeOption
         'Akeneo\Pim\Enrichment\Bundle\Elasticsearch\ProductAndProductModelQueryBuilderFactory',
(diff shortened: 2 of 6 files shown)


src/Akeneo/Pim/Structure/Bundle/EventSubscriber/ApiBatchAttributePostSaveEventSubscriber.php
```diff
@@ -0,0 +1,57 @@
...
+use Akeneo\Tool\Component\StorageUtils\StorageEvents;
+use Symfony\Component\EventDispatcher\EventDispatcherInterface;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\EventDispatcher\GenericEvent;
+
+/**
+ * @copyright 2021 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+final class ApiBatchAttributePostSaveEventSubscriber implements EventSubscriberInterface
```
LaurentPetard, line 17: I didn't find a better name. Tell if you do.

ahocquard: Let's keep it consistent: `ApiAggregatorForAttributePostSaveEventSubscriber `

AymericPlanche (approved):

src/Akeneo/Pim/Structure/Bundle/EventSubscriber/ApiBatchAttributePostSaveEventSubscriber.php
```diff
@@ -0,0 +1,57 @@
...
+    {
+        $this->isActivated = true;
+    }
+
+    public function deactivate(): void
+    {
+        $this->isActivated = false;
+    }
+
+    public function batchEvent(GenericEvent $event)
```
ahocquard, line 47: You have to dispatch the events at the end. This is part of the contract when updating attributes.

LaurentPetard: Their propagation are not stopped so they are still dispatched. I don't think that there's the need to dispatch a POST_SAVE_ALL in addition, as it has been done for the products

ahocquard (approved):
:+1:

PR #13736, opened 2021-01-15
CPM-128: Remove jstree from Product Export

pierallard:


src/Akeneo/Platform/Bundle/UIBundle/Resources/public/js/filter/product/category/Selector.tsx
```diff
@@ -0,0 +1,182 @@
...
+    return (categoryCodes || []).includes(category.code);
+  };
+
+  const handleChange = (value: string, checked: boolean) => {
+    const index = (categoryCodes || []).indexOf(value, 0);
+    if (checked) {
+      if (index <= -1) {
+        (categoryCodes || []).push(value);
+        setCategoryCodes([...(categoryCodes || [])]);
+        onChange(categoryCodes || []);
```
mmetayer, line 64: shouldn't we use a local variable here? (e.g ` const localCategoryCodes = categoryCodes || [];`)
if categoryCodes is not defined I think these 3 LOC will use an empty array (but not the same instance), so you will call setCategoryCodes and onChange with an empty array (not the one with the pushed value)

nmarniesse: I don't get why categoryCodes can be null

src/Akeneo/Platform/Bundle/UIBundle/Resources/public/js/filter/product/category/Selector.tsx
```diff
@@ -0,0 +1,182 @@
...
+`;
+
+type Category = {
+  id: number;
+  code: string;
+  parent: string | null;
+  labels: {[locale: string]: string};
+};
+
+type Channel = {
```
nmarniesse, line 27: Should be great to have this type in a generic file. But I don't really know where to put it

nmarniesse (approved):

mmetayer (approved):

PR #13742, opened 2021-01-18
PIM-9635: make the attribute option code insensitive in cache

nmarniesse:
When the product is updated with a simple option attribute, the validation uses the cache to validate the existence of the option code. There was an error when we used code with bad case.

For instance when the attribute code is `simple` and the option code is `yes`, and we update product with "YES" through API:

**Before this PR**

We make the query and put in cache:

```json
{"simple.YES": false, "simple.yes": true}
```
=> The validation is ok because there is ` "simple.yes": true`.

The second time, we don't make the query because we have the key `simple.YES` in cache, we use it so we have:

```json
{"simple.YES": false}
```
=> The validation fails. It misses the ` "simple.yes": true` part

**The fix**

To fix that I change the caching method. Now I cache:

- key: always in lower case (ex: `simple.yes`)
- value: the real code. It can be with lower and/or upper characters

This way by asking key cache in lower case we can retrieve the real codes.

### DIFF
--- a/src/Akeneo/Pim/Structure/Bundle/Query/PublicApi/AttributeOption/Cache/LRUCachedGetExistingAttributeOptions.php
+++ b/src/Akeneo/Pim/Structure/Bundle/Query/PublicApi/AttributeOption/Cache/LRUCachedGetExistingAttributeOptions.php
@@ -39,21 +39,37 @@ class LRUCachedGetExistingAttributeOptions implements GetExistingAttributeOption
                 return [];
             }
 
-            $results = array_fill_keys($nonCachedAttributeOptionKeys, false);
             $existingAttributeOptionCodes = $this->getExistingOptionCodes->fromOptionCodesByAttributeCode(
                 $this->fromCacheKeys($nonCachedAttributeOptionKeys)
             );
-            $existingKeys = array_fill_keys($this->toCacheKeys($existingAttributeOptionCodes), true);
 
-            return array_replace($results, $existingKeys);
+            $newCacheKeys = $this->toCacheKeys($existingAttributeOptionCodes);
+
+            // We build the cache as:
+            //  - the key represents the attribute option code that can be asked
+            //  - the value represents the real attribute option code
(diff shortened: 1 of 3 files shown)


src/Akeneo/Pim/Structure/Bundle/Query/PublicApi/AttributeOption/Cache/LRUCachedGetExistingAttributeOptions.php
```diff
@@ -39,21 +39,23 @@ public function fromOptionCodesByAttributeCode(array $optionCodesIndexedByAttrib
                 return [];
             }
 
-            $results = array_fill_keys($nonCachedAttributeOptionKeys, false);
+            $results = array_fill_keys($nonCachedAttributeOptionKeys, '');
             $existingAttributeOptionCodes = $this->getExistingOptionCodes->fromOptionCodesByAttributeCode(
                 $this->fromCacheKeys($nonCachedAttributeOptionKeys)
             );
-            $existingKeys = array_fill_keys($this->toCacheKeys($existingAttributeOptionCodes), true);
+            $newCacheKeys = $this->toCacheKeys($existingAttributeOptionCodes);
```
ahocquard, line 46: why the fix is not in toCacheKeys to index in lowercase?

nmarniesse: I need to have a method that lowercases the key and another that don't lowercase. I can do that it would be more clear

nmarniesse: Done in https://github.com/akeneo/pim-community-dev/pull/13742/commits/a57288062dffd0da3b12dac179b1fb75b88549f4

rouliane (approved):

jmleroux (approved):

PR #13743, opened 2021-01-18
PIM-9631: fix attribute groups not displayed in family due to JS error

rouliane:

This PR fixes a JS error that prevented attribute groups to be displayed in the family edit form where there was more that 20 attribute groups in the family. The endpoints to get attribute groups accepts a `limit` option that has a default value to 20 if not set. The option was indeed not set so the returned attribute groups were capped to 20, and this triggered a JS error because we tried to access a group that didn't exist.

We don't want a limit in this particular use case and the only to do it is to set a limit to a negative number.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - PIM-9595: Avoid 403 error when launching import with no view rights on import details
 - PIM-9622: Fix query that can generate a MySQL memory allocation error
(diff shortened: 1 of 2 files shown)


nmarniesse (approved):
This is not very scalable but I guess it's ok for attribute groups

anaisbaunelemaire (approved):

jmleroux (approved):

PR #13746, opened 2021-01-18
API-1401: Beta is still displayed on some screens

tomglvng:
5ee96fcd-754a-431f-90f4-e8a09b10781d

+ fix `event subscriptions` to `event subscription`


pchasle (commented):
You should not update en_AU, en_GB as they will be overwritten by Crowdin

pchasle (approved):
Did you validate the pluralisation changes with Mégane? Menu entry & page title.
If you did, GTM after fixes on en_AU, en_GB files

PR #13782, opened 2021-01-22
API-1415: add propagation time to send event log

pyaillet:
Add propagation time (min and max) to `send_api_event_request`.

```json
{
    "type": "event_api.send_api_event_request",
    [...],
    "max_propagation_seconds": 15,
    "min_propagation_seconds": 12,
}
```


src/Akeneo/Connectivity/Connection/back/tests/Unit/spec/Application/Webhook/Log/EventSubscriptionSendApiEventRequestLogSpec.php
```diff
@@ -115,9 +117,75 @@ public function it_returns_the_log_without_response()
...
-                    'timestamp' => 1577836800,
+                    'timestamp' => 1577836811,
+                ],
+            ],
+            'max_propagation_seconds' => 26098229,
+            'min_propagation_seconds' => 26098218,
+        ]);
+    }
+
+    public function it_returns_the_log_without_propagation_times()
```
pchasle, line 128: Should not happen, `eventDateTime` is always defined on an event.

pchasle (approved):

src/Akeneo/Connectivity/Connection/back/Application/Webhook/Log/EventSubscriptionSendApiEventRequestLog.php
```diff
@@ -94,9 +96,44 @@ public function toLog(): array
                     'timestamp' => $date ? $date->getTimestamp() : null,
                 ];
             }, $this->webhookRequest->apiEvents()),
-        ];
+        ] + $this->getPropagationSeconds();
```
tseho, line 99: I don't like to add 2 arrays together. Always feel hackish to me.
Regarding `getPropagationSeconds`, I would rather have 2 distincts methods like `getMaxPropagationSeconds(): int|null` and `getMinPropagationSeconds(): int|null`. 
IMHo, this way, the code is easier to read and it will be easier to disctincly add new metrics tomorrow.

src/Akeneo/Connectivity/Connection/back/tests/Unit/spec/Application/Webhook/Log/EventSubscriptionSendApiEventRequestLogSpec.php
pyaillet: We added this to handle the case when the event time couldn't be parsed.
Do you think wwe should ignore a null return from `\DateTimeImmutable::createFromFormat(\DateTime::ATOM, $event->eventDateTime());` ?

src/Akeneo/Connectivity/Connection/back/Application/Webhook/Log/EventSubscriptionSendApiEventRequestLog.php
pyaillet: We first tried the solution you describe.
However, it required going twice through the event array, and handling the null case for each array key in `toLog()` method, which was not that easy to read either.

No problem reverting back to this version if you think it's more readable though.

src/Akeneo/Connectivity/Connection/back/tests/Unit/spec/Application/Webhook/Log/EventSubscriptionSendApiEventRequestLogSpec.php
willy-ahva: It SHOULD not happen but the code says it could happen, so we handled the case.

src/Akeneo/Connectivity/Connection/back/Application/Webhook/Log/EventSubscriptionSendApiEventRequestLog.php
willy-ahva: @tseho as we tried the two solutions, I confirm that this solution is more readable and avoid doing twice the work. I prefer the code as it is now but we can discuss that.

willy-ahva (approved):

src/Akeneo/Connectivity/Connection/back/tests/Unit/spec/Application/Webhook/Log/EventSubscriptionSendApiEventRequestLogSpec.php
pchasle: Yes, this was just to highlight the expected behavior.
And it's probably better to handle this error properly, like that, than risking the application to break (by throwing an error).

src/Akeneo/Connectivity/Connection/back/Application/Webhook/Log/EventSubscriptionSendApiEventRequestLog.php
tseho: If there is a majority of opinion for your solution, let's go.

PR #13800, opened 2021-01-26
DAPI-1480: Add the ability to filter products on Quality Score through the API

LaurentPetard:
To make the filter easier to use for external users, I added a new "supported field" to have a simpler name, and I added the possibility to use the score letters as values.


src/Akeneo/Pim/Automation/DataQualityInsights/back/Infrastructure/Elasticsearch/Filter/QualityScoreFilter.php
```diff
@@ -33,6 +35,23 @@ public function addFieldFilter($field, $operator, $values, $locale = null, $chan
             throw InvalidPropertyTypeException::arrayExpected($field, static::class, $values);
         }
 
+        if (null === $channel) {
+            throw InvalidPropertyException::dataExpected($field, 'a channel', static::class);
+        }
+
+        if (null === $locale) {
```
AymericPlanche, line 42: null locale and null channel are default values in the signature method, aren't you adding negative side effects ? 

as this filter could be used in product grid and export builder (to come).

LaurentPetard: As discussed with Florence, the channel and the locale are mandatory. It could technically be possible to filter by score for any channel and locale, but it doesn't make any functional sense.
The signature comes from the Interface `FieldFilterInterface` so it can't be changed.
In any case, in master if `$locale` or `$channel` are `null`, the filter will be wrong and will return no products
(because of this line : `sprintf('data_quality_insights.scores.%s.%s', $channel, $locale) => $values`)

src/Akeneo/Pim/Automation/DataQualityInsights/back/Infrastructure/Elasticsearch/Filter/QualityScoreFilter.php
```diff
@@ -16,7 +18,7 @@ class QualityScoreFilter extends AbstractFieldFilter implements FieldFilterInter
 {
     public function __construct()
     {
-        $this->supportedFields = ['data_quality_insights_score'];
+        $this->supportedFields = ['data_quality_insights_score', 'quality_score'];
```
olivierpontier, line 21: I may be missing something, I'm confused with the naming. What is the difference between the `data_quality_insights_score` and` quality_score` fields?

olivierpontier (approved):

rouliane (approved):

src/Akeneo/Pim/Automation/DataQualityInsights/back/Infrastructure/Elasticsearch/Filter/QualityScoreFilter.php
LaurentPetard: It's the same filter but for two different fields (product-grid and API). I could have use only one field, but I think that `data_quality_insights_score` is not an enough clear name for external users.

AymericPlanche (approved):

PR #13801, opened 2021-01-26
PIM-9648: Mitigate DDoS risk on API auth endpoint by rejecting too large content

ValentinMumble:

This PR introduces a `kernel.request` event listener that is triggered before `FOSRestBundle`'s BodyListener and that watches the API auth route to throw if the content size is above 300 bytes, in order to mitigate DDoS attacks.
The request size varies on the `username`, `password` and `grant_type` content length.

Before:
image

After:
image

`OK`: Done / Validated
`-`: Not needed


src/Akeneo/Tool/Bundle/ApiBundle/Controller/TokenController.php
```diff
@@ -29,12 +29,19 @@ public function __construct(OAuth2 $oauthServer)
     /**
      * @param Request $request
      *
-     * @throws UnprocessableEntityHttpException
+     * @throws HttpException
```
jmleroux, line 32: I would keep the two Exceptions in phpdoc

jmleroux (approved):

StevenVAIDIE (approved):

pierrejolly (approved):

willy-ahva (approved):

src/Akeneo/Tool/Bundle/ApiBundle/Controller/TokenController.php
```diff
@@ -15,12 +16,11 @@
  */
 class TokenController
 {
-    /** @var OAuth2 */
-    protected $oauthServer;
+    // Maximum allowed request content size (in bytes)
+    protected const MAX_CONTENT_SIZE = 300;
```
tseho, line 20: Why 300 ?

ahocquard (commented):
It is not fixed. The json_decode occurs before reaching the controller.

src/Akeneo/Tool/Bundle/ApiBundle/Controller/TokenController.php
ahocquard: It's a good compromise as the username has not a fixed size. With a size of 300, you cannot do anything as an attacker, but it lets some flexibility if the size of the token change, the size of the username also etc.

nmarniesse:
> It is not fixed. The json_decode occurs before reaching the controller.

Ah

Can we block the request by adding an apache config?

ahocquard:
@nmarniesse that's a good idea indeed. But it does not prevent it for EE on-premise, CE, etc. @BitOne what do you think about it?

BitOne:
> @nmarniesse that's a good idea indeed. But it does not prevent it for EE on-premise, CE, etc. @BitOne what do you think about it?

I would prefer to have it fixed on the application side. It will be less fragile and have a more widespread effect.

ValentinMumble:
> It is not fixed. The json_decode occurs before reaching the controller.

@ahocquard yes indeed, we use FOSRestBundle's BodyListener to decode our requests' body...
https://github.com/FriendsOfSymfony/FOSRestBundle/blob/3.x/EventListener/BodyListener.php#L89
https://github.com/akeneo/pim-community-dev/blob/PIM-9648/config/packages/fos_rest.yml#L19

nmarniesse (commented):
IMHO it seems a bit complicated, specially the part where the decoder is overwritten.

What about a single event subscriber that is executed before the FOS Rest ones (so with higher priority) with something like:

```php
    public function onKernelRequest($event)
    {
        $request = $event->getRequest();
        if (API_AUTH_ROUTE !== $request->get('_route')) {
           return;
        }

        if (self::MAX_CONTENT_SIZE < strlen($request->getContent())) {
            throw new HttpException(
                Response::HTTP_REQUEST_ENTITY_TOO_LARGE,
                sprintf('Request content exceeded the maximum allowed size of %s bytes', self::MAX_CONTENT_SIZE)
            );
        }
    }
```

tseho:
I get it that you overrides the FOS BodyListener to let it skip the request on the concerned endpoint.

Have you tried to create an independent RequestListener, with the appropriate priority to be executed before the json_decode's one ?
In this one you could do `if url == ... and  strlen($content) > MAX_CONTENT_SIZE; then throw`
This way, you won't have to touch the fos_rest config nor the controller.

src/Akeneo/Tool/Bundle/ApiBundle/Listener/ApiBodyListener.php
```diff
@@ -0,0 +1,21 @@
...
+use Symfony\Component\HttpFoundation\Request;
+
+final class ApiBodyListener extends BodyListener
+{
+    protected const API_AUTH_ROUTE = 'fos_oauth_server_token';
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function isDecodeable(Request $request)
```
rybus, line 17: Is it not `decodable`?

ValentinMumble: Yes it is but I was overriding the method here so I had to keep the same name 😉 
Anyway, I'll apply the good suggestion above so this will be reverted

src/Akeneo/Tool/Bundle/ApiBundle/EventSubscriber/CheckRequestContentSizeSubscriber.php
```diff
@@ -0,0 +1,48 @@
...
+
+namespace Akeneo\Tool\Bundle\ApiBundle\EventSubscriber;
+
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\HttpKernel\Exception\HttpException;
+use Symfony\Component\HttpKernel\KernelEvents;
+
+/**
+ * This subscriber will watch the API auth route and throw a 413 if the request content size is too large.
```
jmleroux, line 13:
```suggestion
 * This subscriber will watch an auth route and throw a 413 if the request content size is too large.
```
Just in case we use it to protect other routes

src/Akeneo/Tool/Bundle/ApiBundle/EventSubscriber/CheckRequestContentSizeSubscriber.php
```diff
@@ -0,0 +1,48 @@
...
+ * This is to mitigate DDoS attacks on this unrestricted endpoint.
+ */
+final class CheckRequestContentSizeSubscriber implements EventSubscriberInterface
```
jmleroux, line 16: Ah. My previous comment is not good because this class is specialized thanks to the API_AUTH_ROUTE constant.

```suggestion
final class CheckApiRequestContentSizeSubscriber implements EventSubscriberInterface
```
In this case, could we not use a specialized naming?

jmleroux: Or make it more configurable...

src/Akeneo/Tool/Bundle/ApiBundle/EventSubscriber/CheckRequestContentSizeSubscriber.php
```diff
@@ -0,0 +1,48 @@
...
+
+namespace Akeneo\Tool\Bundle\ApiBundle\EventSubscriber;
+
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\HttpKernel\Exception\HttpException;
+use Symfony\Component\HttpKernel\KernelEvents;
+
+/**
+ * This subscriber will watch the API auth path and throw a 413 if the request content size is too large.
```
ahocquard, line 13:
```suggestion
 It is possible to create a lot of collisions in an PHP array by a malicious user. Doing it allows to DDOS easily and almost without any resource a server.
 
 To prevent it, it's important to protect the size sent in the json body, when the user is not authenticated.
 Do note that $_GET and $_POST variables are already protected by the PHP setting "max_input_vars". Therefore, "application/x-www-form-urlencoded" content-type is not concerned by this vulnerability.
 
 The goal of this subscriber si to prevent it.
 
 See:
 https://nikic.github.io/2011/12/28/Supercolliding-a-PHP-array.html
 https://www.securityweek.com/hash-table-collision-attacks-could-trigger-ddos-massive-scale
```

ahocquard (approved):
GTM if not reproducible.

tseho (approved):

nmarniesse (approved):
:+1:

PR #13872, opened 2021-02-04
API-1432: Replace StorageEvent observer by MessageBus observer

tseho:

Since those Events (ProductCreated, ProductUpdated) do belong to enrichment now, I'm suggesting to directly observe queued event in the message bus.
For legacy behats, we will adhere to a similar pattern.

This way, we can now assert like this:
```
$this->assertEventCount(1, ProductCreated::class);
```


pchasle (approved):

pierrejolly (approved):

willy-ahva (approved):

tests/back/Integration/IntegrationTestsBundle/Messenger/AssertEventCountTrait.php
```diff
@@ -0,0 +1,39 @@
...
+
+namespace Akeneo\Test\IntegrationTestsBundle\Messenger;
+
+use Akeneo\Platform\Component\EventQueue\BulkEventInterface;
+use Akeneo\Platform\Component\EventQueue\Event;
+use Symfony\Component\Messenger\TraceableMessageBus;
+
+trait AssertEventCountTrait
+{
+    public function assertEventCount(int $expectedCount, string $eventClassName): void
```
pchasle, line 13:
```suggestion
    public function assertEventCount(int $expectedCount, string $eventName): void
```
I'm not sure we should directly use the Events class here, to avoid being "linked" to the connectivity context.
Using the type directly, like "product.created" we would at least be decoupled properly.

or create an assert for each type of events
```
assertProductCreatedEventCount(int $expectedCount)
```

pchasle: Following a discussion with @willy-ahva . I take back this suggestion.
The code here is clear enough and there is probably no issue with the bounded context / coupling

PR #13875, opened 2021-02-05
API-1446 log events for products

pyaillet:
Log event for products just after they are sent in the queue

### DIFF
--- a/src/Akeneo/Pim/Enrichment/Bundle/EventSubscriber/BusinessEvent/DispatchProductCreatedAndUpdatedEventSubscriber.php
+++ b/src/Akeneo/Pim/Enrichment/Bundle/EventSubscriber/BusinessEvent/DispatchProductCreatedAndUpdatedEventSubscriber.php
@@ -97,7 +97,7 @@ final class DispatchProductCreatedAndUpdatedEventSubscriber implements EventSubs
                     [
                         'type' => 'business_event.dispatch',
                         'event_count' => count($this->events),
-                        'events' => array_map(function ($event) { 
+                        'events' => array_map(function ($event) {
                             return [
                                 'name' => $event->getName(),
                                 'uuid' => $event->getUuid(),
--- a/src/Akeneo/Pim/Enrichment/Bundle/EventSubscriber/BusinessEvent/DispatchProductCreatedAndUpdatedEventSubscriber.php
+++ b/src/Akeneo/Pim/Enrichment/Bundle/EventSubscriber/BusinessEvent/DispatchProductCreatedAndUpdatedEventSubscriber.php
@@ -89,6 +89,14 @@ final class DispatchProductCreatedAndUpdatedEventSubscriber implements EventSubs
 
         try {
             $this->messageBus->dispatch(new BulkEvent($this->events));
+            foreach ($this->events as $event) {
+                $this->logger->info(
+                    json_encode(
+                        $event->toLog(),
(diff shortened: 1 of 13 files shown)


src/Akeneo/Platform/Component/EventQueue/Event.php
```diff
@@ -50,4 +50,24 @@ public function getUuid(): string
...
+    public function toLog(): array
+    {
+        return [
+            'type' => $this->getName(),
+            'uuid' => $this->uuid,
+            'author' => $this->author->name(),
+            'author_type' => $this->author->type(),
+            'timestamp' => $this->timestamp,
+        ];
+    }
```
pchasle, line 72: I'm not a big fan of adding "log" methods to an object like that

pyaillet: Do you think we should handle the conversion logic in each dispatcher ? Or transform this into a dedicated object ?

pchasle: I'm looking at what we can do. Especially as we have a lot logging in the Events API codebase; something we are not used to on the rest of the PIM.

From my point of view, a log depend on a specific context in which it is triggered, as you might need additional informations or a custom structure.

Maybe a simple fix would be to call this method `normalize` instead, that just mean the object will be returned as a standardized associative array.

willy-ahva: I agree with @pchasle . A simple rename in `normalize` is ok for me.

willy-ahva (approved):
Maybe not in this PR but I see that the dispatchers are not spec. We should do it.

pyaillet:
D'autres remarques !

src/Akeneo/Pim/Enrichment/Bundle/EventSubscriber/BusinessEvent/DispatchProductCreatedAndUpdatedEventSubscriber.php
```diff
@@ -89,6 +89,20 @@ public function dispatchBufferedProductEvents(): void
...
+            foreach ($this->events as $event) {
+                $this->logger->info(
+                    json_encode(
+                        [
+                            'type' => $event->getName(),
+                            'uuid' => $event->getUuid(),
+                            'author' => $event->getAuthor()->name(),
+                            'author_type' => $event->getAuthor()->type(),
+                            'timestamp' => $event->getTimestamp(),
+                        ],
```
pchasle, line 101:
```suggestion
                    json_encode(
                        [
                          'type' => 'event_api.dispatch_event',
                          'event' => [
                                'type' => $event->getName(),
                                'uuid' => $event->getUuid(),
                                'author' => $event->getAuthor()->name(),
                                'author_type' => $event->getAuthor()->type(),
                                'timestamp' => $event->getTimestamp(),
                            ]
                        ],
```
Suggestion on the format

pchasle (approved):

tseho (approved):

pyaillet:
@pchasle @willy-ahva we introduced some changes since your reviews, is this still ok for you ?

pchasle (approved):
I would have used the `event_api` logger directly, instead of creating a new one.

pchasle:
If we follow the ubiquitous language, it should be `pim_event` instead of `business_event`

PR #13876, opened 2021-02-05
FIX: add +x on create_db_test.sh script

pyaillet:


pyaillet:
The first time the mysql container is started on MacOS it crashes on the execution of this script.
The second time it works because we bypass the init phase.
I and @pierrejolly tested this fix.

tseho (approved):

pierrejolly (approved):

PR #13908, opened 2021-02-11
API-1429 PoC Elasticsearch logging

pchasle:


composer.json
```diff
@@ -124,7 +124,8 @@
         "symfony/validator": "4.4.16",
         "symfony/yaml": "4.4.16",
         "twig/twig": "1.44.1",
-        "webmozart/assert": "1.9.1"
+        "webmozart/assert": "1.9.1",
+        "symfony/http-client": "4.4.18"
```
pchasle, line 128: https://symfony.com/blog/new-in-symfony-4-3-httpclient-component#comment-23297
> That's one of the reasons why the component itself does not implement the PSR-18 interface (we provide a bridge to PSR, but you will loose that benefit as the bridge will need to block on IO before returning the PSR-7 response).

PR #13914, opened 2021-02-12
PIM-9678: The time counter is still running despite the job failed

tomglvng:
Disable the ETA when job has failed.


src/Akeneo/Platform/Bundle/UIBundle/Resources/public/js/job/execution/Progress.tsx
```diff
@@ -53,14 +53,14 @@ const getStepExecutionTrackingTitle = (step: StepExecutionTracking): string => {
   return translate(key);
 };
 
-const getStepExecutionTrackingProgressLabel = (step: StepExecutionTracking): string => {
+const getStepExecutionTrackingProgressLabel = (jobStatus: string, step: StepExecutionTracking): string => {
   const translate = useTranslate();
```
SamirBoulil, line 57: shouldn't we add an early return here instead of going into the switch ?

SamirBoulil: Shouldn't we do what's at line 85?

olivierpontier (approved):

src/Akeneo/Platform/Bundle/UIBundle/Resources/public/js/job/execution/Progress.tsx
tomglvng: I don't think so: we don't want a duration for failing job I guess.

tomglvng: > shouldn't we add an early return here instead of going into the switch ?

I agree, but we already have the condition `!step.isTrackable` into the switch. Do you think an before-the-switch return is still necessary ?

tseho: I think @tomglvng is right. The current solution is good.
An early return would display "untrackable" for steps that are already completed and that's not what we want.
The only issue here is that we can have a situation where job is `failed` and the current step is `started`.

tseho (approved):

PR #13952, opened 2021-02-18
fixed jquery import in media.ts that makes slimbox plugin works

szymku:
fixed jquery import in media.ts that makes slimbox plugin works and fixes:
`Uncaught TypeError: o.slimbox is not a function`


nestorAkeneo:

PR #13963, opened 2021-02-22
Fix event log for dev

pyaillet:
Remove events_api and business_event logs from console for dev

### DIFF
--- a/config/packages/dev/monolog.yml
+++ b/config/packages/dev/monolog.yml
@@ -11,7 +11,5 @@ monolog:
         console:
             type:     console
             channels: ['!event', '!doctrine', '!event_api', '!business_event']
-        event_api:
-            type:     "null"
         business_event:
             type:     "null"
--- a/config/packages/dev/monolog.yml
+++ b/config/packages/dev/monolog.yml
@@ -11,3 +11,7 @@ monolog:
         console:
             type:     console
             channels: ['!event', '!doctrine', '!event_api', '!business_event']
+        event_api:
+            type:     "null"
+        business_event:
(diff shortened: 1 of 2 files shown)


pchasle (approved):

tseho:
Can you do it in EE too ?

config/packages/dev/monolog.yml
```diff
@@ -11,3 +11,7 @@ monolog:
         console:
             type:     console
             channels: ['!event', '!doctrine', '!event_api', '!business_event']
+        event_api:
+            type:     "null"
```
pchasle, line 15:
```suggestion
```
I would not hide the "event_api" logs, so that we can still see the logs in the consumer cli output

tseho:
In `test` & `behat` too, no ?

pyaillet:
@tseho I didn't notice any problem with `test` & `behat` setup.

tomglvng (approved):

config/packages/dev/monolog.yml
```diff
@@ -11,3 +11,5 @@ monolog:
         console:
             type:     console
             channels: ['!event', '!doctrine', '!event_api', '!business_event']
+        business_event:
+            type:     "null"
```
willy-ahva, line 15: please change for simple quote and don't align the "null" with the other line.
```suggestion
            type: 'null'
```

willy-ahva (approved):

config/packages/dev/monolog.yml
pyaillet: 👍 thanks for the review

PR #13964, opened 2021-02-22
API-1453 add events api debug log storage

pchasle:
- Configure Elasticsearch storage for Events API debug logs.

Events API architecture

- Add the "Event API requests limit reached" warning log


src/Akeneo/Connectivity/Connection/back/Application/Webhook/Service/EventsApiDebugLogger.php
```diff
@@ -0,0 +1,80 @@
...
+    private EventsApiDebugRepository $repository;
+
+    public function __construct(EventsApiDebugRepository $repository, int $bufferSize = 100)
+    {
+        $this->repository = $repository;
+        $this->bufferSize = $bufferSize;
+        $this->buffer = [];
+    }
+
+    public function logLimitOfEventApiRequestsReached(int $timestamp = null): void
```
tseho, line 42: why is `$timestamp` optional ?

pchasle: As the timestamp is generated when the log is created it's not really necessary to give it.
But it's very convenient (even mandatory) for the tests, to have reproducible output.

tseho: If it's for the tests, maybe we should inject something like https://github.com/akeneo/pim-enterprise-dev/blob/master/src/Akeneo/AssetManager/back/Domain/Query/ClockInterface.php

pchasle: Could be, but I like the simplicity of passing function arguments instead of relying on external dependencies. Plus it make sense to be able to give the time of the log when creating it.
But we can change it if you like.

tseho: I would rather not having this optional argument.
When we raise a log, the timestamp should be `now`, it's what we expect.
Having this as an optional argument, if it's only for the tests, we should use a facade instead.
We are already doing it for `sleep`, for example.
See https://github.com/akeneo/pim-community-dev/blob/master/src/Akeneo/Connectivity/Connection/back/Infrastructure/Webhook/Service/Sleep.php

src/Akeneo/Connectivity/Connection/back/tests/Unit/spec/Application/Webhook/Service/EventsApiDebugLoggerSpec.php
```diff
@@ -0,0 +1,86 @@
...
+    private function getClock(): Clock
+    {
+        return new class() implements Clock
+        {
+            public function now(): \DateTimeImmutable
+            {
+                return new \DateTimeImmutable('2021-01-01T00:00:00+00:00');
+            }
+        };
+    }
```
tseho, line 85: Can't we have a `FakeClock` class where you can set the datetime returned by `now` ?
Could be useful for futur tests.

pchasle: I added one if you want to check it

tseho (approved):

src/Akeneo/Connectivity/Connection/back/Application/Webhook/Service/EventsApiDebugLogger.php
```diff
@@ -0,0 +1,84 @@
...
+    {
+        $this->repository = $repository;
+        $this->clock = $clock;
+        $this->bufferSize = $bufferSize;
+        $this->buffer = [];
+    }
+
+    public function logLimitOfEventApiRequestsReached(): void
+    {
+        $this->addLog([
```
willy-ahva, line 48: Shouldn't we create a WriteModel for this array ?

willy-ahva (approved):

src/Akeneo/Connectivity/Connection/back/Application/Webhook/Service/EventsApiDebugLogger.php
pchasle: We could, it's the simplest version here but we can expand on that. TBD with the team.
(PHP Stan is still enforcing the data type)

PR #13970, opened 2021-02-23
API-1457 add download of events api logs

pchasle:
In the Event Subscription page, we added a "Download logs" button.
There is a new endpoint with a StreamedResponse, to stream the logs in the output. (with E2E test)
There is a new query on ES to scroll through the logs. (with integration tests)

Capture du 2021-02-26 11-16-00


tseho (approved):

src/Akeneo/Connectivity/Connection/back/Infrastructure/InternalApi/Controller/EventsApiDebugController.php
```diff
@@ -0,0 +1,82 @@
...
+                        '%s %s %s %s%s',
+                        \DateTime::createFromFormat(
+                            'U',
+                            (string)$log['timestamp'],
+                            new \DateTimeZone('UTC')
+                        )->format('Y/m/d H:i:s'),
+                        strtoupper($log['level']),
+                        $log['message'],
+                        json_encode($log['context']),
+                        PHP_EOL
```
pchasle, line 73: Don't you think that the query should return each line(s) already formatted? And not do this inside the controller.

pchasle: This seems to be because the controller use directly the `Domain\Webhook\Persistence\Query` while it should be a `Application\Webhook\Query` with a query handler using the persistence query and formatting the output.

src/Akeneo/Connectivity/Connection/back/tests/Integration/FakeClock.php
```diff
@@ -0,0 +1,26 @@
...
+
+namespace Akeneo\Connectivity\Connection\Tests\Integration;
+
+use Akeneo\Connectivity\Connection\Domain\Clock;
+
+/**
+ * @copyright 2021 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+class FakeClock implements Clock
```
pchasle, line 13: There is already a `FakeClock` available at `Akeneo\Connectivity\Connection\Infrastructure\Service\Clock`

tseho: There was not when we started working on it, since it was merged recently.
I'm going to fusion those.

src/Akeneo/Connectivity/Connection/back/Infrastructure/InternalApi/Controller/EventsApiDebugController.php
tseho: In this particular case, I don't think that having another layer on top of it will make it better. 
The role of the controller is too simply fetch the data (by calling the query) and create a response.

The thing we could extract is the formatting of the log, like with a service "LogFormatter" or something but I don't think that it's useful here.

I would rather keep it like this, if the rest of the team agrees.

src/Akeneo/Connectivity/Connection/back/tests/Integration/Persistence/Elasticsearch/Query/GetEventSubscriptionLogsQueryIntegration.php
```diff
@@ -0,0 +1,214 @@
...
+
+    public function test_it_returns_the_correct_amount_of_notice_and_info_logs()
+    {
+        $timestamp = $this->clock->now()->getTimestamp() - 10;
+
+        $this->generateLogs(
+            function ($index) use ($timestamp) {
+                return [
+                    'timestamp' => $timestamp,
+                    'level' => $index % 2 ? 'notice' : 'info',
```
pyaillet, line 44: Is there a reason for not using LogLevels constants from https://github.com/akeneo/pim-community-dev/blob/master/src/Akeneo/Connectivity/Connection/back/Domain/Webhook/Model/EventsApiDebugLogLevels.php ?

src/Akeneo/Connectivity/Connection/back/Application/Webhook/Query/GetEventSubscriptionLogsHandler.php
```diff
@@ -0,0 +1,72 @@
...
+
+use Akeneo\Connectivity\Connection\Domain\Webhook\Persistence\Query\GetEventSubscriptionLogsQueryInterface;
+use Symfony\Component\HttpFoundation\HeaderUtils;
+use Symfony\Component\HttpFoundation\StreamedResponse;
+
+/**
+ * @copyright 2021 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+class GetEventSubscriptionLogsHandler
```
tseho, line 15: As I said in my previous comment, I don't see the point in this query/handler.

Creating a Response is clearly the job of a Controller.
IMHO, It only add code without adding any value.
This is not a functional use case either, there is no side effect or whatever.
I really don't like the idea "we do it like this because we always do this way".

src/Akeneo/Connectivity/Connection/back/tests/Integration/Persistence/Elasticsearch/Query/GetEventSubscriptionLogsQueryIntegration.php
tseho: They did not exists yet, when we started the PR.
We need to rebase and fix it.

src/Akeneo/Connectivity/Connection/back/Application/Webhook/Query/GetEventSubscriptionLogsHandler.php
pchasle: Yes, there is an issue here, the response should be done in the controller, but the formating part should be done here.

For example, something like that:
```
$logs = $this->getEventSubscriptionLogsQuery->execute($query->connectionCode());

foreach ($logs as $log) {
      /** @var \Datetime $date */
      $date = \DateTime::createFromFormat(
          'U',
          (string)$log['timestamp'],
          new \DateTimeZone('UTC')
      );

      yield sprintf(
          '%s %s %s %s%s',
          $date->format('Y/m/d H:i:s'),
          strtoupper($log['level']),
          $log['message'],
          json_encode($log['context']),
          PHP_EOL
      );
 }
```

pchasle: The return type of `GetEventSubscriptionLogsHandler::handle` would be `iterable<string>` aka, une collection de string

pchasle: And in the Controller you can just do a
```
$response->setCallback(
      function () use ($logs) {
          foreach ($logs as $log) {
              echo $log;
              flush();
          }
      }
  );

```

src/Akeneo/Connectivity/Connection/back/Application/Webhook/Query/GetEventSubscriptionLogsHandler.php
```diff
@@ -0,0 +1,72 @@
...
+
+                    echo sprintf(
+                        '%s %s %s %s%s',
+                        $date->format('Y/m/d H:i:s'),
+                        strtoupper($log['level']),
+                        $log['message'],
+                        json_encode($log['context']),
+                        PHP_EOL
+                    );
+                    flush();
```
pchasle, line 65: Are you sure that we should flush after each line, and not buffer a little and flush every 100 lines or whatever ? I don't know if there is any performance impact here

src/Akeneo/Connectivity/Connection/back/Infrastructure/InternalApi/Controller/EventsApiDebugController.php
```diff
@@ -0,0 +1,45 @@
...
+
+    public function downloadEventSubscriptionLogs(Request $request): Response
+    {
+        if (true !== $this->securityFacade->isGranted('akeneo_connectivity_connection_manage_settings')) {
+            throw new AccessDeniedException();
+        }
+
+        $query = new GetEventSubscriptionLogsQuery($request->query->get('connection_code'));
+
+        return $this->getEventSubscriptionLogsHandler->handle($query);
```
pchasle, line 43: See previous comments, the Response should be created here.

src/Akeneo/Connectivity/Connection/back/tests/.php_cd.php
```diff
@@ -39,6 +39,10 @@
 
             //CXP-462
             'Akeneo\UserManagement\Component\Model\UserInterface',
+
+            'Symfony\Component\HttpFoundation\HeaderUtils',
+            'Symfony\Component\HttpFoundation\StreamedResponse',
+
```
pchasle, line 45: To remove, once the Response is added to the controller again

src/Akeneo/Connectivity/Connection/back/Application/Webhook/Query/GetEventSubscriptionLogsHandler.php
tseho: I agree with you, the response should have stayed in the controller context.
With your proposal, we add a layer, which increase complexicity and reduce lisibility for the sake of having a query handler.
Moreover, the handler is reduced to a formatter: here, take this values and output a string with the correct format.
It does not add more values than this.

src/Akeneo/Connectivity/Connection/back/Infrastructure/InternalApi/Controller/EventsApiDebugController.php
willy-ahva: I prefer keeping the CQRS structure as this case fits well with the pattern. The download log action is a user query. Moreover, it will be consistent with the current archi.

pierrejolly (approved):

PR #13973, opened 2021-02-23
CXP-623: Refactor log to use a dedicated log service

pierrejolly:

The goal is to create a dedicated logger service for the Events API with named methods to create each kind of log.

This will remove the need to instantiate the “log object” directly and call the json_encode function everywhere and it will improve the testing of the classes that log.


src/Akeneo/Connectivity/Connection/back/Application/Webhook/Service/EventSubscriptionLogInterface.php
```diff
@@ -0,0 +1,22 @@
...
+
+use Akeneo\Connectivity\Connection\Application\Webhook\Log\EventSubscriptionSendApiEventRequestLog;
+use Akeneo\Platform\Component\EventQueue\EventInterface;
+
+/**
+ * @author Pierre Jolly <pierre.jolly@akeneo.com>
+ * @copyright 2021 Akeneo SAS (http://www.akeneo.com)
+ * @license http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+interface EventSubscriptionLogInterface
```
pchasle, line 15:
```suggestion
interface EventsApiLogger
```
Naming suggestion, let's see what others think

src/Akeneo/Connectivity/Connection/back/Infrastructure/Symfony/Resources/config/services.yml
```diff
@@ -85,3 +85,8 @@ services:
 
     akeneo_connectivity.connection.webhook.sleep:
         class: Akeneo\Connectivity\Connection\Infrastructure\Webhook\Service\Sleep
+
+    akeneo_connectivity.connection.webhook.event_subscription_log:
```
pchasle, line 89:
```suggestion
    akeneo_connectivity.connection.webhook.events_api_logger:
```

src/Akeneo/Connectivity/Connection/back/Application/Webhook/Service/EventSubscriptionLogInterface.php
willy-ahva: Do you see in this interface that we are breaking the SRP? If you type-hint the service you inject in all the classes with this interface, it means you enforce to inject in a class that only needs to log a reach limit error, another class that MUST be able to log a reach limit error AND to log an event build error AND to log an event data builder error AND to ... etc. 
What I want to explain is: I'm ok to gather all the methods to log in to one class (even if I'm not convinced of the real benefit), but I'm not ok to create only one interface to gather them. You should create as many interfaces as you have business cases in order to respect the SRP.

src/Akeneo/Connectivity/Connection/back/Infrastructure/Webhook/Service/EventSubscriptionLog.php
```diff
@@ -0,0 +1,113 @@
...
+use Akeneo\Platform\Component\EventQueue\BulkEventInterface;
+use Akeneo\Platform\Component\EventQueue\EventInterface;
+use Psr\Log\LoggerInterface;
+
+/**
+ * @author Pierre Jolly <pierre.jolly@akeneo.com>
+ * @copyright 2021 Akeneo SAS (http://www.akeneo.com)
+ * @license http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+class EventSubscriptionLog implements EventSubscriptionLogInterface
```
willy-ahva, line 18: This class should implement 5 interfaces.

src/Akeneo/Connectivity/Connection/back/Application/Webhook/Service/EventSubscriptionLogInterface.php
pierrejolly: Yes you're right I'll fix this

src/Akeneo/Connectivity/Connection/back/Infrastructure/Webhook/Service/EventSubscriptionLog.php
pierrejolly: Yes you're right I'll fix this

PR #13987, opened 2021-02-25
PIM-9701: Fix role deletion when a user do not have any role

nmarniesse:

When removing a role, we checked first if a user will not have any role. This behaviour assumes a user have always a role. But this is not the case, we can go to a role edit page and remove its users -> a user can have 0 role.  
When a user have no role in the PIM, we cannot remove a role anymore because the "check that a user will not have a role" is always true.  

To fix that this PR changes the check. Now we check if there is users that have only this role -> if yes we cannot remove the role.

`OK`: Done / Validated
`-`: Not needed


src/Akeneo/UserManagement/Bundle/Storage/Query/SqlGetUserCount.php
```diff
@@ -0,0 +1,45 @@
...
+),
+user_with_only_role AS (
+    SELECT uwr.username
+    FROM user_with_role uwr
+        JOIN oro_user_access_role r ON uwr.id = r.user_id
+    GROUP BY uwr.username
+    HAVING COUNT(r.role_id) = 1
+)
+SELECT count(username) FROM user_with_only_role
+SQL;
```
jmleroux, line 41:
```suggestion
            WITH
            user_with_role AS (
                SELECT u.id, u.username
                FROM oro_user u
                    JOIN oro_user_access_role ur ON u.id = ur.user_id
                    JOIN oro_access_role r ON ur.role_id = r.id AND r.role = :role
            ),
            user_with_only_role AS (
                SELECT uwr.username
                FROM user_with_role uwr
                    JOIN oro_user_access_role r ON uwr.id = r.user_id
                GROUP BY uwr.username
                HAVING COUNT(r.role_id) = 1
            )
            SELECT count(username) FROM user_with_only_role
            SQL;
```
Just a detail: PHP 7.4 allow for HEREDOC indentation. I find it more readable

jmleroux (approved):
GTM with a tiny comment

StevenVAIDIE (approved):

PR #14000, opened 2021-02-26
CXP-623: Refactor log to use a dedicated log service

pierrejolly:

The goal is to create a dedicated logger service for each kind of Events API log.

This will remove the need to instantiate the “log object” directly and call the json_encode function everywhere and it will improve the testing of the classes that log.


pyaillet (approved):

src/Akeneo/Connectivity/Connection/back/tests/Unit/spec/Application/Webhook/Service/Logger/EventBuildLoggerSpec.php
```diff
@@ -2,20 +2,32 @@
...
+    }
+
+    public function it_is_initializable(): void
+    {
+        $this->shouldBeAnInstanceOf(EventBuildLogger::class);
+    }
+
+    public function it_logs_event_build(LoggerInterface $logger): void
+    {
+        $bulkEvent = new BulkEvent([
```
willy-ahva, line 30: You should put at least two events in the BulkEvent to check that it is well-formed.

willy-ahva (approved):

pchasle (approved):

PR #14007, opened 2021-03-01
API-1471 move log buffer to repository

pchasle:
We buffer the log to optimise the indexation in elasticsearch.
This PR aime to decouple the buffering from the logger and add it to the persistence layer.


pierrejolly (approved):

pyaillet (approved):
GTM

tomglvng (approved):

PR #14010, opened 2021-03-02
API-1471 add interfaces to events api debug logger

pchasle:
Add two interfaces to the Events API logger:
- limit of events api request reached
- event subscription skipped own event


pierrejolly (approved):

pyaillet (approved):
GTM

PR #14032, opened 2021-03-05
API-1491 add infinite scroll

pchasle:


src/Akeneo/Connectivity/Connection/front/src/webhook/components/EventLogList.tsx
```diff
@@ -1,4 +1,121 @@
...
+                }
+            ),
+            search_after: '_encrypted_',
+            total,
+        };
+    }
+}
+
+type Log = {
+    id: number;
```
tseho, line 31: There is no id on the log. Do we need one ?
I'm aware it's useful for `key={id}` but can the index be enough in our case ?

pchasle: Using the index of the map function for example, is really not recommended. I generally prefer to define an index at the source of the data directly (fetcher) if none is provided by the backend.
In this case, we might need to return an index for each entry (maybe the elasticsearch one?) if we don't want to have an issue with the rerender of the list once the search function is implemented.

PR #14048, opened 2021-03-09
CPM-155: Add symfony messenger job queues

nmarniesse:
### TL;DR

Today we have a homemade queue to launch jobs. This queue uses doctrine to store the job list. The goal of this PR is to replace this queue and use the Symfony Messenger component in order to be able to use any queue system we want (Google Pub/Sub for instance).  

Please note this PR introduce the new system **but it's not used for now**. The switch to the new system will be done in a sperate PR.  

#### Consumer's blacklist

The actual queue has a blacklist system: we can consume all jobs except ones with *csv_category_import* code. This way we can have multipe consumers that execute different types of job. With Messenger we can't configure a blacklist, but there is a system of priority based on transports.

We choose to have 3 transports using 3 queues to handle jobs:
- one for UI jobs (mass edit, mass delete)
- one for import/export jobs (included quick exports)
- one for other jobs, called data maintenance jobs

Having 3 queues allow to:
- handle only jobs of given type(s) in a consumer, for example I want to launch only UI jobs, or only UI + import/export jobs
- handle a priority in jobs (when we launch a consumer with 3 transports, the consumer will handle the first transport in priority, then the second, and finally the third)  

To work with 3 queues we need 3 job message PHP classes. In fact we have 4 classes: `UIJobExecutionMessage`, `ImportJobExecutionMessage`, `ExportJobExecutionMessage` and `DataMaintenanceJobExecutionMessage`. Import and export messages use the same queue but we keep like that because we don't know yet if we need to separate them in 2 different queues.  

Note: we could have only one queue and have 3 consumers using the new [filter feature](https://cloud.google.com/pubsub/docs/filtering). But there is some limits: the filter can be changed once the consumer is created, the filter does not work on pubsub emulator (we receive all messages!). Finally we think it's clearer to have 3 different queues.  

#### Acknowledge queue's message

In the messenger workflow the worker pulls the message, then it gives the message to the handler(s), then at the end  of the process the message is acknowledged. If the message is not acknowledged in time, the message can be pull again by another worker/consumer.  
The duration can be configured with a `redeliver_timeout` option when using a doctrine transport. With Google PubSub the time can be configured with the `ackDeadlineSeconds` option (see [doc here](https://cloud.google.com/pubsub/docs/reference/rest/v1/projects.subscriptions/create)). Sadly the maximum time is 10 minutes. That means if the job executiion is not finished after 10 minutes, another worker will receive the message again and try to launch the job again. To avoid this situation we choose to ack the message immediately after receiving it. That is the purpose of the `ack_message_right_after_pull` option (not mandatory, false by default)

#### Transports config

The job messages are send/receive to/from the queues via transports. Thanks to symfony messenger we can configure different queues depending on the env:

| envs | queue
| dev / prod  | doctrine queues (sql table)
| behat / test | Google Pubsub
| test_fake | in_memory

### How to add a job in the new queue?

As before just use the `akeneo_batch_queue.launcher.queue_job_launcher` service. When the switch to the new queue will be done, the service will use the new queue.  

### How to launch the new consumer worker?

When the job queues will be in service, the new way to consume messages will be:

```bash
bin/console  messenger:consume ui_job import_export_job data_maintenance_job [--limit=...]
```

With `--limit=n` the worker will stop after n received messages

### Q/A

`OK`: Done / Validated
`-`: Not needed


src/Akeneo/Tool/Bundle/BatchQueueBundle/MessageHandler/JobExecutionMessageHandler.php
```diff
@@ -0,0 +1,149 @@
...
+    {
+        $jobInstanceCode = $this->executionMessageRepository->getJobInstanceCode($jobExecutionMessage);
+
+        $arguments = [
+            $jobInstanceCode,
+            $jobExecutionMessage->getJobExecutionId(),
+        ];
+
+        foreach ($jobExecutionMessage->getOptions() as $optionsName => $optionValue) {
+            // todo: check this code, it's weird
```
nmarniesse, line 129: todo

src/Akeneo/Tool/Bundle/BatchQueueBundle/MessageHandler/JobExecutionMessageHandler.php
```diff
@@ -0,0 +1,145 @@
...
+
+        try {
+            $arguments = array_merge(
+                [$pathFinder->find(), $console, 'akeneo:batch:job'],
+                $this->extractArgumentsFromMessage($jobExecutionMessage)
+            );
+            $process = new Process($arguments);
+            $process->setTimeout(null);
+
+            $this->logger->notice(sprintf('Launching job execution "%s".', $jobExecutionMessage->getJobExecutionId()));
```
nmarniesse, line 71: Add time execution in seconds in log

src/Akeneo/Tool/Bundle/BatchQueueBundle/Queue/MessengerJobExecutionQueue.php
```diff
@@ -0,0 +1,34 @@
...
+    {
+        $this->bus = $bus;
+    }
+
+    public function publish(JobExecutionMessageInterface $jobExecutionMessage): void
+    {
+        $this->bus->dispatch($jobExecutionMessage);
+    }
+
+    public function consume(string $consumer, JobQueueConsumerConfiguration $configuration): ?JobExecutionMessageInterface
```
nmarniesse, line 30: add a todo to remove in CPM-152

src/Akeneo/Tool/Bundle/BatchQueueBundle/spec/Queue/MessengerJobExecutionQueueSpec.php
```diff
@@ -0,0 +1,35 @@
...
+    {
+        $this->shouldImplement(JobExecutionQueueInterface::class);
+        $this->shouldBeAnInstanceOf(MessengerJobExecutionQueue::class);
+    }
+
+    function it_publishes_job_execution_in_the_queue(MessageBusInterface $bus)
+    {
+        $jobExecutionMessage = DataMaintenanceJobExecutionMessage::createJobExecutionMessage(1, []);
+        $enveloppe = new Envelope($jobExecutionMessage);
+        $bus->dispatch($jobExecutionMessage)->willReturn($enveloppe);
```
nmarniesse, line 30:
```suggestion
        $bus->dispatch($jobExecutionMessage)->willReturn($envelope);
```

tests/back/Integration/IntegrationTestsBundle/Launcher/JobLauncher.php
```diff
@@ -418,4 +473,12 @@ public function launchAuthenticatedSubProcessImport(
 
         self::launchSubProcessImport($jobCode, $content, $username, $fixturePaths, $config);
     }
+
+    public function flushMessengerJobQueue(): void
+    {
+        do {
+            $enveloppes = $this->jobQueueTransport->get();
```
nmarniesse, line 480:
```suggestion
            $envelopes = $this->jobQueueTransport->get();
```

pchasle:
I'm reading the PR description, and about this part:
> ... we choose to ack the message immediately after receiving it.
> That is the purpose of the ack_message_right_after_pull option ...

We asked ourselves the same question (and I think that Florian had the issue too, on the Exboarder).

A common solution to this issue is to periodically extends the _ackDeadline_ by calling _modifyAckDeadline_.
So you will not need to acknowledge immediately the message, and deny you the possibility to reject the job later on, in case of failure.

As most job have some kind of progress, it should be possible to periodically notify PubSub that your are still processing the message.
Did you looked into something like that ? What do you think of it ?

src/Akeneo/Tool/Bundle/MessengerBundle/Transport/GooglePubSub/Client.php
```diff
@@ -44,22 +44,27 @@ public static function fromDsn(
...
+        Assert::string($options['project_id']);
+        Assert::string($options['topic_name']);
 
         foreach (['project_id', 'topic_name'] as $key) {
-            if (!is_string($options[$key])) {
-                throw new \InvalidArgumentException(
-                    sprintf('Option "%s" is missing or invalid.', $key)
-                );
-            }
+            Assert::string($options[$key], sprintf('Option "%s" is missing or invalid.', $key));
```
pchasle, line 59: You can remove the foreach.

And you can use the Symfony OptionResolver if you like. I should probably have used it in the first place.

nmarniesse:
> I'm reading the PR description, and about this part:
> 
> > ... we choose to ack the message immediately after receiving it.
> > That is the purpose of the ack_message_right_after_pull option ...
> 
> We asked ourselves the same question (and I think that Florian had the issue too, on the Exboarder).
> 
> A common solution to this issue is to periodically extends the _ackDeadline_ by calling _modifyAckDeadline_.
> So you will not need to acknowledge immediately the message, and deny you the possibility to reject the job later on, in case of failure.
> 
> As most job have some kind of progress, it should be possible to periodically notify PubSub that your are still processing the message.
> Did you looked into something like that ? What do you think of it ?

Yes I think about this option but this is not easy to implement. And I don't get if there is a time limit with _modifyAckDeadline_ the doc is not clear about that. Anyway we could update the ack deadline each x minutes but I don't know how to implement that. The handler don't have any information about the native message nor the transport.  

But before to implement it, the real question is: do we want to retry a job in failure? Today we don't do that and I'm not sure we want it. If we implement a retry mode I guess more work are needed: how to inform the user about the retry, etc...  
Could be an interesting topic but not for today IMO

src/Akeneo/Tool/Bundle/BatchQueueBundle/README.md
```diff
@@ -0,0 +1,24 @@
...
+## Transports
+
+Symfony Messenger works with transports that are defined in the configuration. A transport defines where the message will be sent (be careful
+if no transport is defined the job is executed synchronously). Different transport can be defined per environment. We can use
+doctrine (= a mysql table) or Google Pub/Sub for instance.
+
+## Jobs priority
+
+Some job types can slow down other jobs, either because of their duration or because too jobs are created too often. To improve that
+we create 4 job queues:
```
nmarniesse, line 18: 3

nmarniesse:
Adding make target to consume job messages will added in a separate PR

src/Akeneo/Tool/Bundle/BatchQueueBundle/MessageHandler/JobExecutionMessageHandler.php
```diff
@@ -0,0 +1,150 @@
...
+        }
+
+        $executionTimeInSec = time() - $startTime;
+        $this->logger->notice('Job execution "{job_execution_id}" is finished in {execution_time_in_sec} seconds.', [
+            'job_execution_id' => $jobExecutionMessage->getJobExecutionId(),
+            'execution_time_in_sec' => $executionTimeInSec,
+        ]);
+    }
+
+    private function executeProcess(Process $process, JobExecutionMessageInterface $jobExecutionMessage)
```
mmetayer, line 96:
```suggestion
    private function executeProcess(Process $process, JobExecutionMessageInterface $jobExecutionMessage): void
```

src/Akeneo/Tool/Bundle/BatchQueueBundle/README.md
```diff
@@ -0,0 +1,87 @@
...
+## Presentation
+
+The Symfony Messenger component helps applications send and receive messages to/from other applications or via message queues. In our case 
+we send the message that tells a job should be launched. A worker receives the message from the queue asynchronously and
+executes the job.  
+
+## Transports
+
+Symfony Messenger works with transports that are defined in the configuration. A transport defines where the message will be sent (be careful
+if no transport is defined the job is executed synchronously). Different transport can be defined per environment. We can use
```
mmetayer, line 12:
```suggestion
if no transport is defined then the job is executed synchronously). Various transports can be defined per environment. We can use
```

src/Akeneo/Tool/Bundle/BatchQueueBundle/README.md
```diff
@@ -0,0 +1,87 @@
...
+doctrine (= a mysql table) or Google Pub/Sub for instance.
+
+## Jobs priority
+
+Some job types can slow down other jobs, either because of their duration or because too jobs are created too often. To improve that
```
mmetayer, line 17:
```suggestion
Some job types can slow down other jobs, either because of their duration or because too many jobs are created too often. To improve that
```

src/Akeneo/Tool/Bundle/BatchQueueBundle/README.md
```diff
@@ -0,0 +1,87 @@
...
+| Consume 1 message       |                           |                      |                        |
+| ----------------------> |                           |                      |                        |
+|                         |                           |                      |                        |
+| ...                     |                           |                      |                        |
+```
+### Why do we ackowledge the message differently with Google PubSub?
+
+Google PubSub has a maximum time to acknowledge a message. By default it's 10 seconds and it can be configured
+with the `ackDeadlineSeconds` option (https://cloud.google.com/pubsub/docs/reference/rest/v1/projects.subscriptions/create).
+The maximum we can configured is 10 minutes. If the the message is not acknowledged in time, the message can be delivered again.
```
mmetayer, line 81:
```suggestion
The maximum that can be configured is 10 minutes. If the message is not acknowledged in that interval, it can be delivered again.  
```

src/Akeneo/Tool/Bundle/BatchQueueBundle/README.md
```diff
@@ -0,0 +1,87 @@
...
+| ----------------------> |                           |                      |                        |
+|                         |                           |                      |                        |
+| ...                     |                           |                      |                        |
+```
+### Why do we ackowledge the message differently with Google PubSub?
+
+Google PubSub has a maximum time to acknowledge a message. By default it's 10 seconds and it can be configured
+with the `ackDeadlineSeconds` option (https://cloud.google.com/pubsub/docs/reference/rest/v1/projects.subscriptions/create).
+The maximum we can configured is 10 minutes. If the the message is not acknowledged in time, the message can be delivered again.  
+But it's not rare a job execution exceeds 10 minutes, and we don't want to receipt twice the same message. We thought about 2 solutions:
```
mmetayer, line 82:
```suggestion
But it's not rare a job execution exceeds 10 minutes, and we don't want to receive the same message twice. We thought about 2 solutions:
```

src/Akeneo/Tool/Bundle/BatchQueueBundle/README.md
```diff
@@ -0,0 +1,87 @@
...
+- When the job is too long we can extend the acknowledge time by a doing a request (https://cloud.google.com/pubsub/docs/reference/rest/v1/projects.subscriptions/modifyAckDeadline)
+- As soon the message is receipted, we acknowledge the message without waiting the handler
```
mmetayer, line 84:
```suggestion
- As soon the message is received, we acknowledge the message without waiting for the handler
```

src/Akeneo/Tool/Bundle/MessengerBundle/README.md
```diff
@@ -82,6 +82,20 @@ From the Symfony Messenger point of view, these are three independent queues. Bu
   Default to `false`, but can be enabled to make the transport create the topic and subscription for you.
   This is useful when using the in-memory [Pub/Sub emulator](https://cloud.google.com/pubsub/docs/emulator) (enabled when the environment variable `PUBSUB_EMULATOR_HOST` is defined).
 
+- `ack_message_right_after_pull: ?bool`
+
+  Default to `false`, it allows to ack the message right after pulling it.  
+  By default Google Pub/Sub waits for an acknowledgement during the next 10 seconds after the message is pulled. After that the message is available once again for another subscriber using the same subscription.  
+  It can be a problem because in Symfony Messenger the message is acknowledged after the message is handled. For long processes the message will not be acknowledged in time.  
+  The maximum custom deadline you can specify is 600 seconds (10 minutes). If this limit is not high enough, consider set this option to `true`.
```
mmetayer, line 90:
```suggestion
  The maximum custom deadline you can specify is 600 seconds (10 minutes). If this limit is not high enough, consider setting this option to `true`.  
```

mmetayer (approved):
awesome job (queue :grin:)

src/Akeneo/Tool/Bundle/BatchQueueBundle/README.md
```diff
@@ -0,0 +1,87 @@
...
+|                         | <------------------------ |                      |                        |
+|                         |                           |                      |                        |
+|                         | Acknowledge message       |                      |                        |
+|                         | (GooglePubSub only)       |                      |                        |
+|                         | ------------------------> |                      |                        |
+|                         |                           |                      |                        |
+|                         | Spread message to handler |                      |                        |
+|                         | --------------------------+--------------------> |                        |
+|                         |                           |                      |                        |
+|                         |                           |                      | Run job (new process)  |
```
ahocquard, line 49:
```suggestion
|                         |                           |                      | Run job (in  sub-process)  |
```

src/Akeneo/Tool/Bundle/BatchQueueBundle/README.md
```diff
@@ -0,0 +1,87 @@
...
+
+Google PubSub has a maximum time to acknowledge a message. By default it's 10 seconds and it can be configured
+with the `ackDeadlineSeconds` option (https://cloud.google.com/pubsub/docs/reference/rest/v1/projects.subscriptions/create).
+The maximum we can configured is 10 minutes. If the the message is not acknowledged in time, the message can be delivered again.  
+But it's not rare a job execution exceeds 10 minutes, and we don't want to receipt twice the same message. We thought about 2 solutions:
+- When the job is too long we can extend the acknowledge time by a doing a request (https://cloud.google.com/pubsub/docs/reference/rest/v1/projects.subscriptions/modifyAckDeadline)
+- As soon the message is receipted, we acknowledge the message without waiting the handler
+
+The first solution is hard to implement, the only advantage is we can retry to execute a failing job. We are not sure we
+want this feature, at least for the moment, so we took the second solution.
```
ahocquard, line 87: ### Why do we have an healthcheck date updated regularly?

## History
Initially, the queue was an homemade queue implemented with Mysql.
A daemon (never-ending PHP command) consumed one message at a time. A message corresponds to an Akeneo job to launch. This job is executed in a sub-process. It increase the stability of the daemon because all it pushed all the complexity of a job (ORM, ES, potential memory leaks, etc) inside a sub-process.

- If the daemon fail, it's not an issue: the job updates its status by itself (error, stopped, etc).
- If the job fail, it's not an issue: the daemon can detect that and change the status of the job. This way, it does not stay as started, even if there was an unexpected and unrecoverable issue (such as memory leak).

But there is one missing case. Imagine the daemon is failing for any reason, and after that the job too. There is no safeguard to change the job state.

And here come the healthcheck datetime. Updated every X seconds by the daemon, the health check a job allows to say "the job is still running, everything is fine".

If a job has a status at STARTED and the healthcheck is not updated, it means that:
- the daemon failed and can't change the healthcheck
- the job itself can't change its own status: it probably failed also

So, in this case, despite the status "STARTED" in database, it's considered as failed in the UI for the final user.
 
## Why is it useless when running with docker?
Without docker, f the daemon is killed or failed, the jobs (in sub-processes) are **not** killed.

With docker, if the daemon is killed or failed, as it's the main process, it will stop the contained (and so the sub-process).
In that case (pretty common case), the healthcheck is useless.

ahocquard (approved):
End of a nice life. I will miss youuuuu, my queuuuue :heart: :musical_note:

src/Akeneo/Tool/Bundle/BatchQueueBundle/README.md
nmarniesse: @ahocquard Could I add this comment in the readme? Maybe it was the original purpose when you wrote it.. Thanks

ahocquard: yep, It was the purpose of my comment, to keep a trace of this thing :)

src/Akeneo/Tool/Bundle/BatchQueueBundle/README.md
```diff
@@ -85,3 +85,30 @@ But it's not rare a job execution exceeds 10 minutes, and we don't want to recei
 
 The first solution is hard to implement, the only advantage is we can retry to execute a failing job. We are not sure we
 want this feature, at least for the moment, so we took the second solution.
+
+### Why do we have an healthcheck date updated regularly?
+
+#### History
+
+Initially, the queue was a homemade queue implemented with Mysql.  
+A daemon (never-ending PHP command) consumed one message at a time. A message corresponds to an Akeneo job to launch. This job is executed in a sub-process. It increase the stability of the daemon because all it pushed all the complexity of a job (ORM, ES, potential memory leaks, etc) inside a sub-process.
```
ahocquard, line 94:
```suggestion
A daemon (never-ending PHP command) consumed one message at a time. A message corresponds to an Akeneo job to launch. This job is executed in a sub-process. It increases the stability of the daemon because it pushed all the complexity of a job (ORM, ES, potential memory leaks, etc) inside a sub-process.
```

src/Akeneo/Tool/Bundle/BatchQueueBundle/README.md
```diff
@@ -85,3 +85,30 @@ But it's not rare a job execution exceeds 10 minutes, and we don't want to recei
...
+
+If a job has a status at STARTED and the healthcheck is not updated, it means that:
+- the daemon failed and can't change the healthcheck
+- the job itself can't change its own status: it probably failed also
+
+So, in this case, despite the status "STARTED" in database, it's considered as failed in the UI for the final user.
+
+#### Why is it useless when running with docker?
+
+Without docker, if the daemon is killed or fails, the jobs (in sub-processes) is **not** killed.
```
ahocquard, line 111:
```suggestion
Without docker, if the daemon is killed or failed, the jobs (in sub-processes) is **not** killed.
```

src/Akeneo/Tool/Bundle/BatchQueueBundle/README.md
```diff
@@ -85,3 +85,30 @@ But it's not rare a job execution exceeds 10 minutes, and we don't want to recei
...
+
+With docker, if the daemon is killed or fails, as it's the main process, it will stop the contained (and so the sub-process).
```
ahocquard, line 113:
```suggestion
With docker, if the daemon is killed or failed, as it's the main process, it will stop the container (and so the sub-process).
```

PR #14066, opened 2021-03-12
API-1488: Fetch Event Subscription logs with infinite scrolling

tseho:

Screenshot_2021-03-12_01-13-38

**Todo**
- [ ] Tests


src/Akeneo/Connectivity/Connection/front/src/webhook/scroll/hooks/useScrollPosition.ts
```diff
@@ -0,0 +1,48 @@
...
+    callback: (position: ScrollPosition) => void,
+    deps: DependencyList = [],
+    delay = 300
+): void => {
+    const handleScroll = useCallback(() => {
+        if (null === ref.current) {
+            return;
+        }
+
+        const scrollParent = findScrollParent(ref.current);
```
willy-ahva, line 22: I don't understand why you are looking for the parent.

tseho: Good point, I'm going to write a quick PoC why.
(Maybe I will discover that I've f***ed up)

willy-ahva (approved):

src/Akeneo/Connectivity/Connection/front/src/webhook/scroll/hooks/useScrollPosition.ts
tseho: Given a div with a fixed height, a scrollable overflow and an `ul` with 100 lines inside.

Screenshot_2021-03-15_16-06-42

I've done a quick PoC with/without the find parent:
https://github.com/tseho/poc-infinite-scroll/blob/master/src/App.tsx#L10-L16
```
  useScrollPosition(ref, (position: ScrollPosition) => {
    console.log('with parent', position);
  }, [], 300);

  useScrollPositionWithoutParent(ref, (position: ScrollPosition) => {
    console.log('without parent', position);
  }, [], 300);
```

This is the console log produced:

Screenshot_2021-03-15_16-08-37

Basically, when I request the scrolling position of my ref (the `ul` in this case), I always have `{ scrollTop: 0, clientHeight: 4200, scrollHeight: 4200 }` because yes, this element has effectively a height of 4200px and it is not scrolling.
The scrolling is happening in the first parent with a fixed height & `overflow-y: auto`.

This is where the magic of `findScrollParent` is, it will find the real div being scrolled, and check its scroll position.

src/Akeneo/Connectivity/Connection/front/src/webhook/scroll/hooks/useInfiniteScroll.ts
```diff
@@ -0,0 +1,109 @@
...
+
+            setState(state => ({
+                ...state,
+                lastPage: page,
+                isStopped: null === page,
+                isLoading: false,
+                shouldFetch: false,
+            }));
+        })();
+    }, [shouldFetch]);
```
pchasle, line 68: :warning: **IGNORE THIS COMMENT**

I'm keeping it to have a track record

Issue with the proposed change is that the cleanup effect will trigger when `shouldFetch` is updated and will cancel any fetch in progress

But I think that this expose the underlying issue with the  coupling between scrolling and fetching in this hook.

---

I'm not a big fan of the `isMounted` and would prefer to use the `useEffect` [cleanup](https://reactjs.org/docs/hooks-effect.html#example-using-hooks-1) mechanism like suggested [here](https://github.com/facebook/react/issues/14369#issuecomment-468267798) 

```suggestion
    useEffect(() => {
        let cancelled = false;
        (async () => {
            if (!shouldFetch || isLoading || isStopped) {
                return;
            }

            setState(state => ({
                ...state,
                isLoading: true,
            }));

            const page = await loadNextPage(lastPage);
            
            if (cancelled) {
                return;
            }
  
            setState(state => ({
                ...state,
                lastPage: page,
                isStopped: null === page,
                isLoading: false,
                shouldFetch: false,
            }));
        })();
        
        return () => {
            cancelled = true;
        };
    }, [shouldFetch]);
```

pchasle (approved):

src/Akeneo/Connectivity/Connection/front/src/webhook/scroll/hooks/useInfiniteScroll.ts
tseho: I like the way of returning the function to useEffect + `cancelled` to handle this issue.
This is something that could be interesting to dig.

PR #14070, opened 2021-03-12
PIM-9715: Fix issue when deleting an attribute used as label for a family

anaisbaunelemaire:

When an attribute is used as a label for a family, you can delete it. 
Then, on the family parameters, the identifier is displayed as the family label, but, before, only for the display.
In reality, the label was empty (on the family listing, and on the database, the attribute as label was null). 

So, I added 2 methods that update (on all the families using it) the "attribute as label" by the sku if the attribute used as the label is deleted.

`OK`: Done / Validated
`-`: Not needed


src/Akeneo/Pim/Structure/Bundle/EventSubscriber/UpdateFamilyAttributeAsLabel.php
```diff
@@ -0,0 +1,79 @@
...
+            $attributes,
+            fn ($attr): bool => $attr instanceof AttributeInterface
+        );
+        if ([] === $attributes) {
+            return;
+        }
+        $this->updateFamiliesWithoutAttributeAsLabel();
+    }
+
+    private function updateFamiliesWithoutAttributeAsLabel(): void
```
ahocquard, line 63: You should not modify the content of an entity like that. All changes should be centralized through a repository instead, which will be responsible of modifying the entity and trigger the events.

In that case, for example, it is not in the versioning, which is a bug.

Also, modifying the label_as_attribute should trigger the whole re-indexation of the products inside the family, because the label inside the products is not consistent anymore in Elasticsearch.

ahocquard (commented):
> When an attribute is used as a label for a family, you can delete it.


Is it new? How is it possible? Attribute as label is required (check the UI) all along the life of a family entity. It's not normal to be able to delete it. It's not even possible to delete through the UI.

It was not functionally not possible before and this is a behavior change :thinking:

ahocquard:
@anaisbaunelemaire I re-read the thread mentioning this fix. I missed the case where it's the attribute which is deleted (with the consequence that it removes attribute as label).

As it's validated by a PM, it's ok to do that BUT it's an improvement. As any improvement, the PM should ask for the technical consequences because it's not that trivial.

Here, it's not easy because:
- when deleting an attribute a job is triggered to re-index everything
- when changing an attribute as label (or family requirements, etc), I think a job is triggered to re-index everything (+completeness calculation)

The challenge here is to modify the family from the repository (for the versioning, family update event, etc), but without triggering two times the indexation.

PR #14096, opened 2021-03-16
CPM-99: update dependencies to be PHP8 compatible

nmarniesse:
Here is a first PR to prepare the migration to PHP8. Here we update some dependencies to be php8 compatible:

- `dompdf/dompdf`
- `elasticsearch/elasticsearch`
- `friendsofsymfony/jsrouting-bundle`
- `friendsofsymfony/rest-bundle`
- `gedmo/doctrine-extensions` (need `doctrine/common` to be bumped)
- `imagine/imagine`
- `liip/imagine-bundle`
- `ocramius/proxy-manager`
- `oneup/flysystem-bundle`
- `friends-of-behat/symfony-extension`
- `liuggio/fastest`
- `mcustiel/phiremock`

JS routing bundle: in the new version, the new way to use PHP routes in JS is to dump a JSON file instead of a JS file. And now we had to describe explicitely which routes we want to expose. The `.*` exposes all routes, as we do until now.  

Fix codebase:
- Deprecate required parameters after optional parameters in function/method signatures
- usort(): Returning bool from comparison function is deprecated, return an integer less than, equal to, or greater than zero

Dependencies to migrate next:

- `doctrine/*`: **will be done in a separate PR**, a lot of things to update in our codebase because we need to use new major versions

- [`friendsofsymfony/oauth-server-bundle`](https://packagist.org/packages/friendsofsymfony/oauth-server-bundle) **not php8 compatible yet**


config/packages/fos_js_routing.yml
```diff
@@ -1,2 +1,2 @@
 fos_js_routing:
-    routes_to_expose: ~
+    routes_to_expose: [ '.*' ]
```
nmarniesse, line 2: ` routes_to_expose: ~` is not a valid config anymore. The equivalent is to use a reg exp to match all routes

composer.json
```diff
@@ -66,24 +66,24 @@
...
+        "friendsofsymfony/jsrouting-bundle": "2.7.0",
         "friendsofsymfony/oauth-server-bundle": "1.6.2",
-        "friendsofsymfony/rest-bundle": "2.8.0",
-        "gedmo/doctrine-extensions":"v2.4.37",
+        "friendsofsymfony/rest-bundle": "2.8.6",
+        "gedmo/doctrine-extensions":"v3.0.3",
         "google/cloud-pubsub": "1.29.1",
         "guzzlehttp/guzzle": "6.5.5",
-        "imagine/imagine": "0.7.1",
+        "imagine/imagine": "1.2.4",
```
mmetayer, line 79: no breaking change which impacts us in this one?

mmetayer (approved):

composer.json
nmarniesse: No, 2 BC are documented and we don't use them https://github.com/avalanche123/Imagine/releases/tag/1.0.0

BitOne (approved):

willy-ahva (approved):

PR #14111, opened 2021-03-18
PIM-9748: Upgrade JQuery for security reasons

tomglvng:
This PR upgrades jQuery version following a XSS vulnerability that was fixed in version 3.5.0.
As this version introduces a breaking change on a specific method (see https://jquery.com/upgrade-guide/3.5/), this PR also updates all self-closing tags in HTML templates and in JS files using jQuery.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -6163,10 +6163,10 @@ jest@^26.4.2:
     import-local "^3.0.2"
     jest-cli "^26.6.3"
 
-jquery@3.5.0:
-  version "3.5.0"
-  resolved "https://registry.yarnpkg.com/jquery/-/jquery-3.5.0.tgz#9980b97d9e4194611c36530e7dc46a58d7340fc9"
-  integrity sha512-Xb7SVYMvygPxbFMpTFQiHh1J7HClEaThguL15N/Gg37Lri/qKyhRGZYzHRyLH8Stq3Aow0LsHO2O2ci86fCrNQ==
+jquery@3.6.0:
+  version "3.6.0"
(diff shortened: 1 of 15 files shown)


BitOne (approved):

jmleroux (approved):
Good catch!!

yarn.lock
```diff
@@ -3809,19 +3809,19 @@ domutils@^2.4.2, domutils@^2.4.3, domutils@^2.4.4:
     domelementtype "^2.0.1"
     domhandler "^4.0.0"
 
-draft-js@^0.11.7:
-  version "0.11.7"
-  resolved "https://registry.yarnpkg.com/draft-js/-/draft-js-0.11.7.tgz#be293aaa255c46d8a6647f3860aa4c178484a206"
-  integrity sha512-ne7yFfN4sEL82QPQEn80xnADR8/Q6ALVworbC5UOSzOvjffmYfFsr3xSZtxbIirti14R7Y33EZC5rivpLgIbsg==
+draft-js@^0.10.5:
```
olivierpontier, line 3812: Is normal that the version of draft-js is downgraded?

ValentinMumble: It's now actually the correct version that we require in the `package.json` https://github.com/akeneo/pim-community-dev/blob/master/akeneo-design-system/package.json#L25
But not sure why it's changed now, there may have been a mistake when we added the draft-js library here https://github.com/akeneo/pim-community-dev/commit/e34cd294e0977e85a683a67503fc3dbda4309b18
I remember we first added the latest version (0.11.7) but needed to downgrade for compatibility issues with the Asset Manager IIRC, we probably forgot to regenerate the `yarn.lock` when we downgraded the version to 0.10.5

PR #14112, opened 2021-03-18
BH-467: Allow pubsubClient to use native Google Service account mechanism

mr-prud:
…nism

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Akeneo/Tool/Bundle/MessengerBundle/Transport/GooglePubSub/PubSubClientFactory.php
+++ b/src/Akeneo/Tool/Bundle/MessengerBundle/Transport/GooglePubSub/PubSubClientFactory.php
@@ -26,8 +26,11 @@ class PubSubClientFactory
 
     public function createPubSubClient(array $config): PubSubClient
     {
-        return new PubSubClient(array_merge([
-            'keyFilePath' => $this->keyFilePath
-        ], $config));
+        if (!empty($keyFilePath)) {
+            return new PubSubClient(array_merge([
+                'keyFilePath' => $this->keyFilePath
+            ], $config));
+        }
+        return new PubSubClient($config);
     }
 }
--- a/src/Akeneo/Tool/Bundle/MessengerBundle/Transport/GooglePubSub/PubSubClientFactory.php
+++ b/src/Akeneo/Tool/Bundle/MessengerBundle/Transport/GooglePubSub/PubSubClientFactory.php
@@ -26,7 +26,7 @@ class PubSubClientFactory
 
     public function createPubSubClient(array $config): PubSubClient
     {
-        if (!empty($keyFilePath)) {
+        if (!empty($this->keyFilePath)) {
             return new PubSubClient(array_merge([
                 'keyFilePath' => $this->keyFilePath
             ], $config));


src/Akeneo/Tool/Bundle/MessengerBundle/Transport/GooglePubSub/PubSubClientFactory.php
```diff
@@ -26,8 +26,11 @@ public function __construct(string $keyFilePath)
 
     public function createPubSubClient(array $config): PubSubClient
     {
-        return new PubSubClient(array_merge([
-            'keyFilePath' => $this->keyFilePath
-        ], $config));
+        if (!empty($keyFilePath)) {
```
pyaillet, line 29:
```suggestion
        if (!empty($this->keyFilePath)) {
```

pyaillet (changes requested):

src/Akeneo/Tool/Bundle/MessengerBundle/Transport/GooglePubSub/PubSubClientFactory.php
pchasle: I don't get what this is fixing.

The `$config` here, should not include the `keyFilePath` currently.

So if the value was empty when the `PubSubFactory` was instantiated that means the google `PubSubClient` is constructed with `'keyFilePath' => null`.

And if that is the case, from what I can see in the google library `vendor/google/cloud-core/src/ClientTrait.php`  in the method `getKeyFile` it should just try to load it from the env var `GOOGLE_APPLICATION_CREDENTIALS`

pchasle: If you did your test with this, it means the pim ignored the `keyFilePath` even if it was defined. And that wont be the case anymore.

Maybe it explain the underlying issue that the `keyFilePath` provided is wrong on flex ? or there should not be any.
And the env var GOOGLE_APPLICATION_CREDENTIALS as the right value.

pchasle: The `keyFilePath` come from `SRNT_GOOGLE_APPLICATION_CREDENTIALS`

src/Akeneo/Tool/Bundle/MessengerBundle/Transport/GooglePubSub/PubSubClientFactory.php
```diff
@@ -26,8 +26,11 @@ public function __construct(string $keyFilePath)
...
     public function createPubSubClient(array $config): PubSubClient
     {
-        return new PubSubClient(array_merge([
-            'keyFilePath' => $this->keyFilePath
-        ], $config));
+        if (!empty($this->keyFilePath)) {
+            return new PubSubClient(array_merge([
+                'keyFilePath' => $this->keyFilePath
+            ], $config));
+        }
```
tseho, line 33:
```suggestion
        if (null !== $this->keyFilePath) {
            $config = array_merge([
                'keyFilePath' => $this->keyFilePath
            ], $config));
        }
```

it's a non-blocking suggestion.

tseho (approved):

willy-ahva (approved):

mr-prud:
we will fix in the EE

PR #14122, opened 2021-03-18
PIM-9740: Prevent to delete a channel used in a product export job

LaurentPetard:
When a channel is removed, the products export jobs that are configured with this channel display wrong information and cannot be modified.

The functional solution chosen is to prevent to remove a channel that is used in a product export job.


src/Akeneo/Channel/Bundle/Query/Sql/IsChannelUsedInProductProductExportJob.php
```diff
@@ -0,0 +1,45 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Akeneo\Channel\Bundle\Query\Sql;
```
LaurentPetard, line 5: I'm not sure that it's the right place (bundle) for this query, but I think that it's to specific to be placed in the batch bundle...
If you have a better place, I'm interested.

BitOne (approved):

fitn (approved):

PR #14123, opened 2021-03-18
API-1520: Move download logs button

tseho:

The download logs button was on the event subscription configuration page.
This PR move it to the dedicated new screen for the logs.

Screenshot_2021-03-18_18-07-35

### DIFF
--- a/src/Akeneo/Connectivity/Connection/front/src/shared/router/use-router.ts
+++ b/src/Akeneo/Connectivity/Connection/front/src/shared/router/use-router.ts
@@ -1,0 +1,8 @@
+import {useContext} from 'react';
+import {RouterContext} from './router-context';
+
+export const useRouter = () => {
+    const {generate} = useContext(RouterContext);
+
+    return generate;
+};
--- a/src/Akeneo/Connectivity/Connection/front/src/webhook/pages/EventLogs.tsx
+++ b/src/Akeneo/Connectivity/Connection/front/src/webhook/pages/EventLogs.tsx
@@ -27,9 +27,7 @@ export const EventLogs: FC = () => {
                 <PageHeader
                     breadcrumb={<EventLogsBreadcrumb />}
                     userButtons={<UserButtons />}
-                    buttons={[
-                        <DownloadLogsButton key={0} disabled={true}/>,
-                    ]}
+                    buttons={[<DownloadLogsButton key={0} disabled={true} />]}
                 />
                 <PageContent>
                     <Loading />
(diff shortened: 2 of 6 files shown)


src/Akeneo/Connectivity/Connection/front/src/shared/router/use-router.ts
```diff
@@ -0,0 +1,8 @@
+import {useContext} from 'react';
+import {RouterContext} from './router-context';
+
+export const useRouter = () => {
+    const {generate} = useContext(RouterContext);
```
pchasle, line 5: There is already `useRoute` doing almost the same thing ?!

src/Akeneo/Connectivity/Connection/front/src/webhook/pages/EventLogs.tsx
```diff
@@ -50,19 +81,15 @@ const EventLogsBreadcrumb: FC = () => {
...
-    if (false === eventSubscription.enabled) {
-        return <EventSubscriptionDisabled connectionCode={connectionCode} />;
-    }
+type DownloadLogsButtonProps = {
+    disabled?: boolean;
+    href?: string;
+};
 
-    return <>{children}</>;
+const DownloadLogsButton: FC<DownloadLogsButtonProps> = props => {
```
pchasle, line 89: You can put this in its separate file, un let him generate the url too. With just the connection code as a props

src/Akeneo/Connectivity/Connection/front/src/shared/router/use-router.ts
tseho: yes and no.
The problem with `useRoute`, it's a hook.
With hooks, you cannot have conditions around it.

```
// This code will crash
if (...) {
  const url = useRoute();
}
```

Since you must preserve, between renders, the order and amount of hooks called, you must always call `useRoute` even if you don't have the parameters yet and are not using the url.

By exposing it with this new hook, you can do:
```
const generateUrl = useRouter();
```
And React will not complain.

And that's why this hook already exists in `legacy-bridge`, see https://github.com/akeneo/pim-community-dev/blob/master/src/Akeneo/Platform/Bundle/UIBundle/Resources/workspaces/legacy-bridge/src/hooks/useRouter.ts

src/Akeneo/Connectivity/Connection/front/src/webhook/pages/EventLogs.tsx
tseho: I will try something better.

pchasle (approved):

willy-ahva (approved):

PR #14156, opened 2021-03-24
PIM-9767: Fix minimum & maximum user password validation

ValentinMumble:

Setting new validation constraints on user password length, minimum of 8 characters and maximum of 4096 as it's the Symfony's encoder limit https://github.com/symfony/symfony/blob/5.x/src/Symfony/Component/Security/Core/Encoder/BasePasswordEncoder.php#L27.

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Akeneo/UserManagement/Bundle/Controller/Rest/UserController.php
+++ b/src/Akeneo/UserManagement/Bundle/Controller/Rest/UserController.php
@@ -1,5 +1,7 @@
 <?php
 
+declare(strict_types=1);
+
 namespace Akeneo\UserManagement\Bundle\Controller\Rest;
 
 use Akeneo\Tool\Component\Localization\Factory\NumberFactory;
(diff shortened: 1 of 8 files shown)


jmleroux (approved):

src/Akeneo/UserManagement/Bundle/Controller/Rest/UserController.php
```diff
@@ -41,53 +43,25 @@
...
-    /** @var RemoverInterface */
-    private $remover;
-
-    /** @var TranslatorInterface */
-    private $translator;
-
-    /** @var SecurityFacade */
-    private $securityFacade;
+    private const PASSWORD_MINIMUM_LENGTH = 8;
+    private const PASSWORD_MAXIMUM_LENGTH = 64;
```
BitOne, line 47: I don't think we should limit the length of the password. In the ticket, it's mentioned that we should check that passwords up to 64 in length are working properly, not that we should set a limit.

BitOne (commented):
Can you add an e2e test that makes sure that people with passwords smaller than 8 characters can still log in?
Otherwise, we might lock a lot of users out of their PIM...

src/Akeneo/UserManagement/Bundle/Controller/Rest/UserController.php
ValentinMumble: @BitOne There is this behat that tests the login via the UI, as Mary (with password `mary`) https://github.com/akeneo/pim-community-dev/blob/ee0b8fae656148b80f2cb92e11d37ffa94332135/tests/legacy/features/platform/security/login.feature#L6
Do we need more?

olivierpontier (approved):

jmleroux (approved):
GTM with the integration tests

src/Akeneo/UserManagement/Bundle/Controller/Rest/UserController.php
```diff
@@ -412,79 +402,92 @@ private function updateUser(UserInterface $user, array $data): JsonResponse
...
-            );
+
+        return $violations;
+    }
+
+    private function validatePasswordLength(string $password, string $propertyPath): ConstraintViolationListInterface
+    {
+        $violations = new ConstraintViolationList();
+
+        if (self::PASSWORD_MINIMUM_LENGTH > strlen($password)) {
```
BitOne, line 465: Do not use strlen as this will not work with a UTF-8 characters. mb_strlen is to be used.

src/Akeneo/UserManagement/Bundle/Controller/Rest/UserController.php
```diff
@@ -412,79 +402,92 @@ private function updateUser(UserInterface $user, array $data): JsonResponse
...
+            $violations->add(new ConstraintViolation(
+                $this->translator->trans('pim_user.user.fields_errors.new_password.minimum_length'),
+                '',
+                [],
+                '',
+                $propertyPath,
+                ''
+            ));
+        } elseif (self::PASSWORD_MAXIMUM_LENGTH < strlen($password)) {
```
BitOne, line 474: Same here...

PR #14157, opened 2021-03-24
Fix duplicate user error

nmarniesse:
- Error message was not displayed when username contains space
- Add front validation: when username contain space or is too short
- Show validation message on blur
- Disable submit button when a validation error occurred

### DIFF
--- a/src/Akeneo/UserManagement/Bundle/Resources/workspaces/user-ui/src/components/CreateUserForm.tsx
+++ b/src/Akeneo/UserManagement/Bundle/Resources/workspaces/user-ui/src/components/CreateUserForm.tsx
@@ -22,7 +22,7 @@ type CreateUserFormProps = {
 };
 
 const CreateUserForm = ({userId, onCancel, onSuccess, onError}: CreateUserFormProps) => {
-  const {register, handleSubmit, errors, watch} = useForm();
+  const {register, handleSubmit, errors, watch} = useForm({mode: 'onBlur'});
   const router = useRouter();
   const translate = useTranslate();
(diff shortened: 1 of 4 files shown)


mmetayer (approved):

src/Akeneo/UserManagement/Bundle/Resources/workspaces/user-ui/src/components/CreateUserForm.tsx
```diff
@@ -78,6 +78,22 @@ const CreateUserForm = ({userId, onCancel, onSuccess, onError}: CreateUserFormPr
...
+    '' === value ? translate('pim_user_management.form.error.required') : true;
+  const shouldNotContainSpace = (value: string): true | string =>
+    /\s/.test(value) ? translate('pim_user_management.form.error.should_not_contain_space') : true;
+  const shouldContain3Characters = (value: string): true | string =>
+    value.length < 3 ? translate('pim_user_management.form.error.too_short_value', {count: 3}) : true;
+
+  const getRegisterParameters = (fieldName: string): {[key: string]: any} => {
+    switch (fieldName) {
+      case 'username':
+        return {validate: {shouldNotContainSpace, shouldNotBeBlank, shouldContain3Characters}};
```
pierallard, line 91: If I remember well, we use react hook form here, and if I remember well (2), there is already a validator for empty values (not sure)
```
  ref={
    register({
      required: 'error message' // JS only: <p>error message</p> TS only support string
    })
  }
  ```

pierallard (approved):

PR #14187, opened 2021-03-29
API-1538: Fix Search events logs query when filtering

willy-ahva:

The query scores the MUST higher than the SHOULD, so when you add a MUST at the same level as a SHOULD, the should is almost ignored. 

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Akeneo/Connectivity/Connection/back/tests/Integration/Persistence/Elasticsearch/Query/SearchEventSubscriptionDebugLogsQueryIntegration.php
+++ b/src/Akeneo/Connectivity/Connection/back/tests/Integration/Persistence/Elasticsearch/Query/SearchEventSubscriptionDebugLogsQueryIntegration.php
@@ -89,6 +89,38 @@ class SearchEventSubscriptionDebugLogsQueryIntegration extends TestCase
         Assert::assertEquals(100, $result['total']);
     }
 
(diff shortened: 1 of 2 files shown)


src/Akeneo/Connectivity/Connection/back/tests/Integration/Persistence/Elasticsearch/Query/SearchEventSubscriptionDebugLogsQueryIntegration.php
```diff
@@ -279,7 +311,7 @@ public function test_it_filters_on_log_level()
             [
```
tseho, line 311: You are right, this test shouldn't test more than expected.
The real mixed up here, is the `'connection_code' => 'whatever',`.
The test seems to have been writed like this "write the test first, dump the result, add the assert".

It would be better to cleanup the case,
Insert one log for each level, without connection code, assert that there is 2 results and the levels are the requested ones.

tseho: (same for the other tests you found)

tseho (approved):

pchasle (approved):

PR #14189, opened 2021-03-29
PIM-9774: Fix variant axis 'metric' validation on import

jmleroux:
The MetricValue data type is not enforced by any type hinting.

I could have fix by a cast here: https://github.com/akeneo/pim-community-dev/blob/4.0/src/Akeneo/Pim/Enrichment/Component/Product/Model/AbstractMetric.php#L38 but the impacts could be important, so I chose to only fix the comparison in 4.0

### DIFF
--- a/src/Akeneo/Pim/Enrichment/Component/Product/Model/AbstractMetric.php
+++ b/src/Akeneo/Pim/Enrichment/Component/Product/Model/AbstractMetric.php
@@ -97,10 +97,7 @@ abstract class AbstractMetric implements MetricInterface
             return $metric->getData() === $this->data;
         }
 
-        return 0 === strcmp(
-                preg_replace('/\.0*$/', '', $metric->getData()),
-                preg_replace('/\.0*$/', '', $this->data)
-            );
+        return 0 === bccomp($metric->getData(), $this->data, 100);
     }
 
     /**
--- a/src/Akeneo/Pim/Enrichment/Component/Product/Model/AbstractMetric.php
+++ b/src/Akeneo/Pim/Enrichment/Component/Product/Model/AbstractMetric.php
@@ -31,7 +31,7 @@ abstract class AbstractMetric implements MetricInterface
      *
      * @param string $family
      * @param string $unit
-     * @param float  $data
+     * @param float  $data (TODO: should not be float becaause of precision issues)
      * @param string $baseUnit
      * @param float  $baseData
      */
--- a/src/Akeneo/Pim/Enrichment/Component/Product/Model/AbstractMetric.php
+++ b/src/Akeneo/Pim/Enrichment/Component/Product/Model/AbstractMetric.php
@@ -27,6 +27,8 @@ abstract class AbstractMetric implements MetricInterface
     protected $family;
 
     /**
+     * TODO master: add type hiting for more strict data
+     *
      * @param string $family
      * @param string $unit
      * @param float  $data
--- a/src/Akeneo/Pim/Enrichment/Component/Product/Model/AbstractMetric.php
+++ b/src/Akeneo/Pim/Enrichment/Component/Product/Model/AbstractMetric.php
@@ -87,7 +87,7 @@ abstract class AbstractMetric implements MetricInterface
      */
     public function isEqual(MetricInterface $metric)
     {
-        return $metric->getData() === $this->data && $metric->getUnit() === $this->unit;
+        return (float) $metric->getData() === (float) $this->data && $metric->getUnit() === $this->unit;
     }
 
     /**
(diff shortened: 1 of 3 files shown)


src/Akeneo/Pim/Enrichment/Component/Product/Model/AbstractMetric.php
```diff
@@ -87,7 +89,7 @@ public function getFamily()
      */
     public function isEqual(MetricInterface $metric)
     {
-        return $metric->getData() === $this->data && $metric->getUnit() === $this->unit;
+        return (float) $metric->getData() === (float) $this->data && $metric->getUnit() === $this->unit;
```
BitOne, line 92: Are we sure it's a float?

Float must be avoided as much as possible, as it's an approximation.

BitOne (commented):
Please don't use float...

src/Akeneo/Pim/Enrichment/Component/Product/Model/AbstractMetric.php
jmleroux: > Are we sure it's a float?

It's not very clear what we can have in this data. 
The object is commented with data as float.
The factory is commented with data as double here https://github.com/akeneo/pim-community-dev/blob/4.0/src/Akeneo/Pim/Enrichment/Component/Product/Factory/MetricFactory.php#L55

And I had string in my testes on local examples. :no_mouth:

jmleroux: The issue for this fix is to compare '30' with '30.0000' that should be considered as equals.

src/Akeneo/Pim/Enrichment/Component/Product/Model/AbstractMetric.php
```diff
@@ -87,7 +89,18 @@ public function getFamily()
...
-        return $metric->getData() === $this->data && $metric->getUnit() === $this->unit;
+        if ($metric->getUnit() !== $this->unit) {
+            return false;
+        }
+
+        if (!is_string($metric->getData()) || !is_string($this->data)) {
+            return $metric->getData() === $this->data;
+        }
+
+        return 0 === strcmp(
```
BitOne, line 100: https://www.php.net/manual/en/function.bccomp.php would be less hackish (and yes BCMath is already a PIM requirements ;) )

BitOne: The scale would be the same one as in DB.

src/Akeneo/Pim/Enrichment/Component/Product/Model/AbstractMetric.php
```diff
@@ -27,9 +27,11 @@ abstract class AbstractMetric implements MetricInterface
     protected $family;
 
     /**
+     * TODO master: add string type hiting for more strict data type and to handle big numbers and/or decimals
+     *
      * @param string $family
      * @param string $unit
-     * @param float  $data
+     * @param float  $data (TODO: should not be float becaause of precision issues)
```
olivierpontier, line 34:
```suggestion
     * @param float  $data (TODO: should not be float because of precision issues)
```

olivierpontier (approved):

src/Akeneo/Pim/Enrichment/Component/Product/Model/AbstractMetric.php
jmleroux: We don't use DECIMAL to store them in DB

src/Akeneo/Pim/Enrichment/Component/Product/Model/AbstractMetric.php
```diff
@@ -87,7 +89,15 @@ public function getFamily()
      */
     public function isEqual(MetricInterface $metric)
     {
-        return $metric->getData() === $this->data && $metric->getUnit() === $this->unit;
+        if ($metric->getUnit() !== $this->unit) {
+            return false;
+        }
+
+        if (!is_string($metric->getData()) || !is_string($this->data)) {
```
nmarniesse, line 96: Is it possible to have one data in string and another in integer/float? For example `50` and `"50.0000"`  
The result will be always false currently. Is it a case that we must handle?

nmarniesse (approved):
Just one question to be sure

src/Akeneo/Pim/Enrichment/Component/Product/Model/AbstractMetric.php
jmleroux: In my tests, there were all strings. The SLA ticket was also about strings and I prefer to not break other cases.

But to answer your question, I didn't see where/if we were enforcing the use of strings... :cry:

nmarniesse: :+1: Let's keep as is

PR #14191, opened 2021-03-29
PIM-9783: decrease batch when calculate completeness

nmarniesse:
Fix https://akeneo.atlassian.net/browse/PIM-9783

The query to calculate completeness fails with 8.084 product identifiers. The batch was 10.000, it is now 1.000.  

Tested on https://cora-prod-clone-sds-17727.dev.cloud.akeneo.com
FYI it works with a batch of 5.000. But as the query depends on the catalog structure (number of product models + number of product values), a  margin of error is considered and the batch is set to 1.000.


olivierpontier (approved):

ValentinMumble (approved):
Changelog entry maybe? 😉

PR #14192, opened 2021-03-29
API-1497 add date range filter

pchasle:
In this PR:
- [x] add date range filter

To be done in others PRs:
- [ ] add tests
- [ ] add datetime field to the dsm
- [ ] (to be defined) add datetime library or adapter to the DSM
- [ ] (to be defined) add date picker to the DSM

Events API architecture (1)

Preview:
Peek 2021-03-30 01-05


src/Akeneo/Connectivity/Connection/front/src/shared/user/use-user.ts
```diff
@@ -0,0 +1,11 @@
+import {useContext} from 'react';
+import {UserContext} from './user-context';
+
+export const useUser = () => {
```
pchasle, line 4: To have an easy access (and type hinted) to the locale & user timezone

src/Akeneo/Connectivity/Connection/front/src/webhook/model/EventSubscriptionLogFilters.ts
```diff
@@ -1,21 +1,41 @@
...
+export type FiltersConfig = {
+    dateTime: {
+        min: number;
+        max: number;
+    };
+};
+
+export const getFiltersConfig: () => FiltersConfig = () => ({
+    dateTime: {
+        min: DateTime.utc().minus({hours: 72}).toSeconds(),
```
pchasle, line 38: Ideally, this should come from the backend to avoid having the config in two different place. Maybe later one.

src/Akeneo/Connectivity/Connection/front/src/webhook/components/DateRangeFilter/DateTimeInput.tsx
```diff
@@ -0,0 +1,80 @@
...
+                type='date'
+                value={(value && timestampToZonedString(value, DATE_INPUT_FORMAT)) || ''}
+                min={min && timestampToZonedString(min, DATE_INPUT_FORMAT)}
+                max={max && timestampToZonedString(max, DATE_INPUT_FORMAT)}
+                onChange={handleDateChange}
+                invalid={invalid}
+            />
+            <TextInput
+                type='time'
+                value={(value && timestampToZonedString(value, TIME_INPUT_FORMAT)) || ''}
```
pchasle, line 67: The input should be changed to "unmanaged" with 'defaultValue' instead.
Because the managed input doesn't allow to edit the value on firefox, or any other browser that doesn't support the date & time inputs and fallback to text input.

src/Akeneo/Connectivity/Connection/front/src/webhook/components/DateRangeFilter/DateTimeInput.tsx
```diff
@@ -0,0 +1,80 @@
...
+            day: newDate.day,
+        });
+        onChange(newZonedDateTime.toSeconds());
+    };
+
+    const handleTimeChange = (timeString: string) => {
+        const newTime = DateTime.fromFormat(timeString, TIME_INPUT_FORMAT, {zone});
+        if (undefined === value) {
+            return;
+        }
```
tseho, line 45: why an early return here ?
Where does handleTimeChange is called when value is undefined ?

src/Akeneo/Connectivity/Connection/front/src/webhook/hooks/api/use-infinite-event-subscription-logs.ts
```diff
@@ -54,7 +54,12 @@ const useInfiniteEventSubscriptionLogs = (
     }
 
     if (!isSameAsDefaultFiltersValues(filters)) {
-        parameters.filters = JSON.stringify(filters);
+        parameters.filters = JSON.stringify({
+            levels: filters.levels,
+            text: filters.text,
+            timestamp_from: filters.dateTime.start || null,
+            timestamp_to: filters.dateTime.end || null,
+        });
```
tseho, line 62: Why does EventSubscriptionLogFilters type is not following the same structure ?

src/Akeneo/Connectivity/Connection/front/src/webhook/components/EventLogList.tsx
```diff
@@ -42,7 +44,11 @@ const Arrow: FC = () => {
 export const EventLogList: FC<{connectionCode: string}> = ({connectionCode}) => {
     const translate = useTranslate();
     const scrollContainer = useRef(null);
-    const [filters, setFilters] = useState<EventSubscriptionLogFilters>(DEFAULT_EVENT_SUBSCRIPTION_LOG_FILTERS);
+
+    const [{filters, config}, setFilters] = useState<{
+        filters: EventSubscriptionLogFilters;
+        config: FiltersConfig;
+    }>({filters: getDefaultFilters(), config: getFiltersConfig()});
```
tseho, line 51: The filters config has nothing to do in this useState, it's not something that change.

src/Akeneo/Connectivity/Connection/front/src/webhook/components/EventLogListFilters.tsx
```diff
@@ -12,9 +13,10 @@ const StyledSectionTitle = styled(SectionTitle)`
 
 export const EventLogListFilters: FC<{
     filters: EventSubscriptionLogFilters;
+    config: FiltersConfig;
```
tseho, line 16: IMHO, config should not be an exposed prop to the component `EventLogListFilters`. The only need we currently have is for validation a range of date.
We already are in `EventLogListFilters`, it should be internal.

src/Akeneo/Connectivity/Connection/front/src/webhook/hooks/api/use-infinite-event-subscription-logs.ts
pchasle: In this case the date range I keep in the state doesn't have the same type has the one expected by the query.

Also, I always take care to separate the format of my request payload from any internal application state, because if I don't I'm just creating an unnecessary coupling between them, or worse, I would likely try to feat a square in a circle with two data structure that don't share the same objective.

src/Akeneo/Connectivity/Connection/front/src/webhook/components/DateRangeFilter/DateTimeInput.tsx
pchasle: The goal is just to ignore the time input change if the date is not selected first.

I guess I can remove it as this is enforced by the "disabled" state of the time field and the handler should not be called with an undefined value.

akeneo-design-system/src/components/Dropdown/Overlay/Overlay.tsx
```diff
@@ -20,7 +20,7 @@ const Container = styled.div<
   background: ${getColor('white')};
   box-shadow: 0 0 4px 0 rgba(0, 0, 0, 0.3);
   padding: 0 0 10px 0;
-  max-width: 320px;
+  max-width: 400px;
```
pchasle, line 23: Stéphane told me that there should not be any `max-width`, but from my tests it's necessary to avoid an item with a long text to expand indefinitely.

There is event a demo for that in the DSM documentation https://akeneo.github.io/akeneo-design-system/?path=/docs/components-dropdown--standard#image-item

So I just extended a little the `max-width`, but maybe the real fix will be to define it on the `Dropdown.Item` element instead of the `Overlay`

src/Akeneo/Connectivity/Connection/front/src/webhook/components/DateRangeFilter/DateTimeInput.tsx
pchasle: I can't removed or else the type checking will be angry as value can be undefined

willy-ahva (approved):

src/Akeneo/Connectivity/Connection/front/src/webhook/model/EventSubscriptionLogFilters.ts
```diff
@@ -1,21 +1,41 @@
...
     levels: [
         EventSubscriptionLogLevel.INFO,
         EventSubscriptionLogLevel.NOTICE,
         EventSubscriptionLogLevel.WARNING,
         EventSubscriptionLogLevel.ERROR,
     ],
     text: '',
-};
+    dateTime: {},
+});
```
tseho, line 23: Why do you replace it by a function ?

src/Akeneo/Connectivity/Connection/front/src/webhook/components/EventLogList.tsx
```diff
@@ -42,7 +42,10 @@ const Arrow: FC = () => {
 export const EventLogList: FC<{connectionCode: string}> = ({connectionCode}) => {
     const translate = useTranslate();
     const scrollContainer = useRef(null);
-    const [filters, setFilters] = useState<EventSubscriptionLogFilters>(DEFAULT_EVENT_SUBSCRIPTION_LOG_FILTERS);
+
+    const [{filters}, setFilters] = useState<{
+        filters: EventSubscriptionLogFilters;
+    }>({filters: getDefaultFilters()});
```
tseho, line 48: why do you change the struct if there is only `filters` inside ?

src/Akeneo/Connectivity/Connection/front/src/webhook/components/EventLogList.tsx
```diff
@@ -51,6 +54,9 @@ export const EventLogList: FC<{connectionCode: string}> = ({connectionCode}) =>
         return <NoEventLogs />;
     }
 
+    const handleFiltersChange = (filters: EventSubscriptionLogFilters) =>
+        setFilters(state => ({...state, filters, isDefaultFilters: false}));
```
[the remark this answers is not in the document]
pchasle: Yes, good catch. Forgot to cleanup that

src/Akeneo/Connectivity/Connection/front/src/webhook/components/EventLogList.tsx
pchasle: Good catch too, forgot to simplify the struct after removing the config

src/Akeneo/Connectivity/Connection/front/src/webhook/hooks/api/use-infinite-event-subscription-logs.ts
tseho: I understand that it's a bad idea to try to feat a square a square in a circle and there is always a time when you have to diverge.

But there, was it useful ?
In `src/Akeneo/Connectivity/Connection/front/src/webhook/components/EventLogListFilters.tsx`, you do:
```
onChange={(start, end) =>
   onChange({
        ...filters,
        dateTime: {start, end},
   })
}
```
It was clear for me than `timestamp_from` and `timestamp_to` should be at the upper level of the filters state, adding a sub structure in this state only add more boilerplate, IMHO.

tseho (approved):
Tests are missing and some changes were unnecessary, IMHO.

PR #14194, opened 2021-03-30
PIM-9792: Use mb_strlen to handle special chars

ValentinMumble:

Uses `mb_strlen` to handle special characters correctly in password length computation. However, we still need to use `strlen` on the max length check because Symfony's BasePasswordEncore will compare its strict length when trying to encode it with salt.
See: https://github.com/symfony/symfony/blob/5.x/src/Symfony/Component/Security/Core/Encoder/BasePasswordEncoder.php#L100

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Akeneo/UserManagement/Bundle/Controller/Rest/UserController.php
+++ b/src/Akeneo/UserManagement/Bundle/Controller/Rest/UserController.php
@@ -471,9 +471,9 @@ class UserController
                 $propertyPath,
                 ''
             ));
-            // We have to use `strlen` here because Symfony's BasePasswordEncoder will check
-            // the actual character count when trying to encode it with salt.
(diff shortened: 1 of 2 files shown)


src/Akeneo/UserManagement/Bundle/Controller/Rest/UserController.php
```diff
@@ -471,6 +471,9 @@ private function validatePasswordLength(string $password, string $propertyPath):
                 $propertyPath,
                 ''
             ));
+            // We have to use `strlen` here because Symfony's BasePasswordEncoder will check
+            // the actual character count when trying to encode it with salt.
```
BitOne, line 475:
```suggestion
            // the actual byte count when trying to encode it with salt.
```

BitOne (approved):
GTM After fix

olivierpontier (approved):

PR #14204, opened 2021-03-31
PIM-9726: mitigate deadlock on categories (backport PIM-9548)

willy-ahva:

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Akeneo/Pim/Enrichment/Component/Lock/Query/EnsureLockTableExistsInterface.php
+++ b/src/Akeneo/Pim/Enrichment/Component/Lock/Query/EnsureLockTableExistsInterface.php
@@ -1,0 +1,17 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Akeneo\Pim\Enrichment\Component\Lock\Query;
+
+/**
+ * TODO pull up remove this interface
+ *
+ * @author    Willy Mesnage <willy.mesnage@akeneo.com>
+ * @copyright 2021 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+interface EnsureLockTableExistsInterface
+{
+    public function execute(): void;
+}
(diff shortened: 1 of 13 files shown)


BitOne (changes requested):
Why do we apply a different solution than usual with deadlock?

upgrades/schema/Version_4_0_20210331092434_add_lock_table.php
```diff
@@ -0,0 +1,26 @@
...
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+
+final class Version_4_0_20210331092434_add_lock_table extends AbstractMigration
+{
+    public function up(Schema $schema) : void
+    {
+        $sql = <<<SQL
+CREATE TABLE lock_keys (
```
ahocquard, line 13: To create on the fly:
```suggestion
CREATE TABLE IF NOT EXISTS lock_keys (
```

src/Akeneo/Pim/Enrichment/Component/Lock/Query/EnsureLockTableExistsInterface.php
```diff
@@ -0,0 +1,17 @@
...
+namespace Akeneo\Pim\Enrichment\Component\Lock\Query;
+
+/**
+ * TODO pull up remove this interface
+ *
+ * @author    Willy Mesnage <willy.mesnage@akeneo.com>
+ * @copyright 2021 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+interface EnsureLockTableExistsInterface
```
ahocquard, line 14: useless interface but not a big deal as it should not be pulled up

ahocquard (approved):
I would put directly the creation of the table inside the CategorySaver instead of wrapping it in a class + interface.

But as it will not be pull up it's ok.

BitOne (approved):

PR #14206, opened 2021-03-31
PIM-9779: fix ACE orders after finding ACLs

nmarniesse:
Fix https://akeneo.atlassian.net/browse/PIM-9779  

On a very specific environment, we cannot add a specific ACL for a specific role. This behaviour is very strange, but we are not the only one to observe the problem. It's in the `symfony/security-acl` bundle, it's exactly the same as:  
https://github.com/symfony/security-acl/issues/5 
https://github.com/symfony/security-acl/issues/23 
https://github.com/symfony/security-acl/issues/24 
https://github.com/symfony/security-acl/issues/28 

To fix it a PR was opened https://github.com/symfony/security-acl/pull/29, approved by one person, but never merged.  

We had 2 choices:
- fix the problem by code by applying the patch described by the PR (or equivalent)
- fix by SQL query

We think the fix by SQL query is too risky to be executed on all customers. Because we don't understand all the usecases/features of the security-acl bundle, and it's difficult to go back in case of errors/side effects/whatever
Plus there is a internal cache for ACLs that complicates the operation.

**The fix**

The fix is an adaptation of the PR: after fetching the ACLs we re-order the ACEs. My first idea was to applied strictly the PR modification, but the `hydrateObjectIdentities` method is private and can't be override easily (too many private things + too many inheritence).  
So we fix the ACE order after loading the ACLs.  

**Bonus**

For information here is a query to return the "bad" entries in the database. The `ace_order` should begin by 0 and should be incremental by `class_id`

```sql
WITH acl_entries_with_expected_ace_order AS (
    SELECT id,
           class_id,
           security_identity_id,
           ace_order,
           ROW_NUMBER() over (PARTITION BY class_id ORDER BY ace_order) - 1 AS expected_ace_order
    FROM acl_entries
    WHERE object_identity_id IS NULL -- if we remove this condition, we have a lot of additional entries. Is it normal?
    ORDER BY class_id, ace_order
)
SELECT * FROM acl_entries_with_expected_ace_order WHERE ace_order != expected_ace_order;
```

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -61,5 +61,8 @@
             <directory suffix="EndToEnd.php">src</directory>
         </testsuite>
 
+        <testsuite name="OroSecurityBundle">
+            <file>src/Oro/Bundle/SecurityBundle/Tests/Unit/Acl/Dbal/MutableAclProviderTest.php</file>
+        </testsuite>
     </testsuites>
 </phpunit>
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,7 @@
 - PIM-9783: Optimize batch query when compute completeness
 - PIM-9715: Prevent the deletion of an attribute used as a label by a family
 - PIM-9781: Fix Category tree not refreshing when switching locale
+- PIM-9779: Fix ACE order when loading ACLs
 - PIM-9739: Fix connection users, users, channels having a link to a sub-category
 
 ## New features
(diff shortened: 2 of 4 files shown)


mmetayer (approved):
:+1:

olivierpontier (approved):

phpunit.xml.dist
```diff
@@ -61,5 +61,8 @@
             <directory suffix="EndToEnd.php">src</directory>
         </testsuite>
 
+        <testsuite name="OroSecurityBundle">
+            <file>src/Oro/Bundle/SecurityBundle/Tests/Unit/Acl/Dbal/MutableAclProviderTest.php</file>
```
nmarniesse, line 65: The Oro tests were deactivated. Here I re-activate one test file only

BitOne (approved):
Good job Nini!

PR #14210, opened 2021-03-31
PIM-9783: optimize query for completeness

nmarniesse:
See https://akeneo.atlassian.net/browse/PIM-9783

With the old version, MySQL don't use index range scan if a lot of product identifiers are given (+6000 in our tests).  
Old version: 1min12 for the query with 8084 idnetifiers
New version: less than 2s

Thanks @ahocquard

### DIFF
--- a/src/Akeneo/Pim/Enrichment/Bundle/Storage/Sql/Completeness/SqlGetCompletenessProductMasks.php
+++ b/src/Akeneo/Pim/Enrichment/Bundle/Storage/Sql/Completeness/SqlGetCompletenessProductMasks.php
@@ -51,6 +51,10 @@ final class SqlGetCompletenessProductMasks implements GetCompletenessProductMask
     {
         // TODO - TIP-1212: Replace the first LEFT JOIN (to pim_catalog_family) by an INNER JOIN
         $sql = <<<SQL
+WITH
+filtered_product AS (
+    SELECT id FROM pim_catalog_product WHERE identifier IN (:productIdentifiers)
+)
 SELECT
     product.id AS id,
     product.identifier AS identifier,
@@ -60,11 +64,11 @@ SELECT
            COALESCE(pm2.raw_values, '{}'),
            product.raw_values
     ) AS rawValues
-FROM pim_catalog_product product
+FROM filtered_product
+    INNER JOIN pim_catalog_product product ON filtered_product.id = product.id
     LEFT JOIN pim_catalog_family family ON product.family_id = family.id
     LEFT JOIN pim_catalog_product_model pm1 ON product.product_model_id = pm1.id
     LEFT JOIN pim_catalog_product_model pm2 ON pm1.parent_id = pm2.id
-WHERE product.identifier IN (:productIdentifiers)
 GROUP BY product.identifier
 SQL;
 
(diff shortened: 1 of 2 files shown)


jmleroux (approved):

src/Akeneo/Pim/Enrichment/Bundle/Storage/Sql/Completeness/SqlGetCompletenessProductMasks.php
```diff
@@ -51,6 +51,10 @@ public function fromProductIdentifiers(array $productIdentifiers): array
     {
         // TODO - TIP-1212: Replace the first LEFT JOIN (to pim_catalog_family) by an INNER JOIN
```
BitOne, line 52: Maybe time to remove this comment?

BitOne (approved):
GTM once fixed

StevenVAIDIE (approved):

ahocquard (approved):

ahocquard:
I would add a comment on the why, good practice when optimizing a query. It can avoid a regression.

src/Akeneo/Pim/Enrichment/Bundle/Storage/Sql/Completeness/SqlGetCompletenessProductMasks.php
nmarniesse: So do you want I replace the left join by an inner join? The family is mandatory now?  
I don't have the history, don't really know what the side effects could be.

BitOne: Do you know if it still a valid point? I prefer to remove a misleading comment if possible.

BitOne: It's a bit outside of this PR (more like the boy scout rules), but TIP-1212 is about deciding functionally to make family mandatory for all products. Anyway, forget about my comment, the subject is larger than this PR :)

BitOne (approved):

src/Akeneo/Pim/Enrichment/Bundle/Storage/Sql/Completeness/SqlGetCompletenessProductMasks.php
nmarniesse: Yes still relevant. I remove the comment only

src/Akeneo/Pim/Enrichment/Bundle/Storage/Sql/Completeness/SqlGetCompletenessProductMasks.php
```diff
@@ -110,9 +117,9 @@ private function buildProductMasks(array $rows): array
     {
         $attributeCodes = [];
         foreach ($rows as $row) {
-            $attributeCodes = array_merge($attributeCodes, array_keys($row['cleanedRawValues']));
+            $attributeCodes = array_unique(array_merge($attributeCodes, array_keys($row['cleanedRawValues'])));
```
ahocquard, line 120: why?

nmarniesse: I just add a comment to explain. In the case of the customer, updating 100 product models has for consequence to compute the masks of 8.000 products. In this case the array_merge alone takes more than 1 minute (max execution time is reached I don't have the real time). With array_unique it takes less than 1 second.  
It's because doing array_merge on a big array is very slow, and as the array grows more and more, the cost is exponential.  

Yes, adding `array_unique` could be slowest for small catalog (we talk about milliseconds) but a lot faster for big catalog

PR #14219, opened 2021-04-01
API-1534: Fix log context width

tseho:

There is no `max-width: 100%` on the `<pre>` element, the behavior with max-width in % does not work as expected.
Without a width, `overflow: scroll` does not work.
Without wrapping of the content, the `<pre>` will take the width on its content, overflowing the table.
Hence the following:
Screenshot 2021-03-29 at 10 33 57
(you can see that the row is wider than the table header)

The only solution I found, which will work on all browsers and most screen is to set a width to it.
The width is not random, it's the exact same one as the initial design.

Screenshot_2021-04-01_19-55-59

`OK`: Done / Validated
`-`: Not needed


pierrejolly (approved):

tseho:
Approved by Sophie & Mégane.

willy-ahva (approved):

PR #14234, opened 2021-04-06
PIM-9798: Refresh completeness on product grid after family import

nmarniesse:

When import some families, we can change only required attributes of the family. That means the product values don't change but completeness does.  
In another hand we don't save products when they don't change, so the completeness was never updated.  

I thought about 2 possible solutions:
- force save the products to go back to the past behaviour -> the completeness calculation is done automatically after save
- add the ComputeCompletenessOfFamilyProductsTasklet step to the family import job

The second solution is not perfect as the completeness would be calculate twice if the products change, so I do the first solution.  

`OK`: Done / Validated
`-`: Not needed


jmleroux (approved):

src/Akeneo/Pim/Enrichment/Component/Product/Connector/Job/ComputeDataRelatedToFamilyProductsTasklet.php
```diff
@@ -36,35 +36,16 @@
...
-
-    /** @var JobRepositoryInterface */
-    private $jobRepository;
-
-    /** @var EntityManagerClearerInterface */
-    private $cacheClearer;
-
-    /** @var int */
-    private $batchSize;
+    private ?StepExecution $stepExecution;
```
mmetayer, line 39: = null ?

mmetayer (approved):

PR #14257, opened 2021-04-13
Allows ctrl-c for stopping the php process in docker-compose

tseho:

Do you hate when you start a consumer with docker-compose and you cannot stop it ?

Screenshot_2021-04-13_17-19-41

~Have you tried turning your computer off and on again?~

---

Now, with this PR, it will work.

Screenshot_2021-04-13_17-18-43

**How**

I'm simply proposing to use the option `init: true` see https://docs.docker.com/compose/compose-file/compose-file-v3/#init
It's only possible since https://github.com/akeneo/pim-community-dev/pull/14303, thanks to our dear @jmleroux 

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -16,6 +16,7 @@ services:
       - '${HOST_COMPOSER_HOME:-~/.composer}:/var/www/.composer'
     working_dir: '/srv/pim'
     command: 'php'
+    init: true
     networks:
       - 'pim'
 


jmleroux:
> We could use the option init: true see https://docs.docker.com/compose/compose-file/compose-file-v3/#init
But it's only working since docker-compose format 3.7.

And why is it not possible to use the 3.7 format?

tseho:
> And why is it not possible to use the 3.7 format?
That's a great question.

It's easy to change it in the docker-compose file buuuuuut there is a small issue.
docker-compose format 3.7 is supported since docker-compose v1.22.0, see https://docs.docker.com/compose/release-notes/#1220

And it's where the fun begins.
For example, in circleci, we are using the image `ubuntu-1604:201903-01` which was built at the time with https://github.com/CircleCI-Public/circleci-dockerfiles/blob/8e8eb0462471a6693fa8d1aee36bc5c9784d0793/buildpack-deps/images/xenial/Dockerfile#L70-L74
(Meaning the "latest" version at the time was included, not updated since)
I've tried, it does not work with our current image https://app.circleci.com/pipelines/github/akeneo/pim-community-dev/21452/workflows/7fe11d47-f836-4b77-b4df-d00eb2575e31/jobs/88219
So, we would have to change the image used by our CI.
And who knows what else ?

BitOne:
Clearly not really fan of changing the init for this reason, as it's not really blocking (a kill command would exit the process without issue).

pchasle:
Since @jmleroux migrated docker-compose to 3.7, does this means we can add the option `init: true` ?

tseho:
@pchasle yes, I will update the pr accordingly asap.

tseho:
I've updated the PR, and went back to the initial solution with `init: true`

BitOne (approved):

nmarniesse (approved):
:+1:

jmleroux (approved):

PR #14269, opened 2021-04-15
CPM-158: Fix MediaFileInput

pierallard:


akeneo-design-system/src/components/Input/MediaFileInput/MediaFileInput.tsx
```diff
@@ -135,6 +135,11 @@ type MediaFileInputProps = Override<
      */
     clearTitle: string;
 
+    /**
+     * Defines if the input can be cleared.
+     */
+    clearable?: boolean;
```
nmarniesse, line 141: What is the use case(s) for clearable? Looks like a desguised readOnly, or a mix between readOnly and required

nmarniesse: Seen IRL. Don't totally agree with the usage but it was approved by the front guys

nmarniesse (approved):

mmetayer (approved):

PR #14276, opened 2021-04-15
API-1564: Add tests & reduce amount of re-render in useInfiniteScroll

tseho:

2 parts:
- new generic test on `use-infinite-event-subscription-logs.ts`
- new test about amount of re-renders of `useInfiniteScroll.ts`

### DIFF
--- a/src/Akeneo/Connectivity/Connection/front/tests/src/webhook/scroll/hooks/useInfiniteScroll.test.ts
+++ b/src/Akeneo/Connectivity/Connection/front/tests/src/webhook/scroll/hooks/useInfiniteScroll.test.ts
@@ -88,3 +88,26 @@ test('Reset the scroll should fetch the first page', async () => {
 
     unmount();
 });
+
+test('The hook does not render more times than necessary', async () => {
+    const ref = {
+        current: document.getElementById('content'),
+    };
+
+    const loadNextPage = jest.fn().mockImplementation(() => {
+        return Promise.resolve(null);
+    });
+
+    let renderCount = 0;
+
+    const {waitForNextUpdate, unmount} = renderHook(() => {
+        renderCount++;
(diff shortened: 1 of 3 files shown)


tomglvng (approved):

src/Akeneo/Connectivity/Connection/front/tests/src/webhook/scroll/hooks/useInfiniteScroll.test.ts
```diff
@@ -88,3 +88,26 @@ test('Reset the scroll should fetch the first page', async () => {
 
     unmount();
 });
+
+test('The hook does not render more times than necessary', async () => {
```
pchasle, line 92: That seems weird to test that, and I'm wondering if the test is viable since you don't call `waitForNextUpdate` more, you don't really know if subsequent async executions will occur?

pchasle (approved):

RiffVentura (approved):

src/Akeneo/Connectivity/Connection/front/tests/src/webhook/scroll/hooks/useInfiniteScroll.test.ts
tseho: When I added the test, it was failing because it was rendering 3 times. That's why I've slightly changed the hook.
So it was not useless, like I was afraid it could be.

But those renders are happenings during the same cycle from I what I could see, the `await waitForNextUpdate` only wait for the fetch promise resolving.

You cannot wait for one more `waitForNextUpdate` because if there is none, and that's the case here, it will fail on the timeout.

pchasle: Alright, I understand. Thanks

PR #14284, opened 2021-04-16
Api 1341 - Updated API error output format

RiffVentura:
API concatenates error messages together without leaving a space character in between
This pull request threats only errors done by search filters validation for the following endpoints:
- family (is already fixed on master)
- category
- attribute
- attributeGroup

~~Updated namespaces from DateTime to DateTimeInterface as the former was removed since PHP 7.2~~

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Akeneo/Pim/Enrichment/Bundle/Doctrine/ORM/Repository/ExternalApi/CategoryRepository.php
+++ b/src/Akeneo/Pim/Enrichment/Bundle/Doctrine/ORM/Repository/ExternalApi/CategoryRepository.php
@@ -178,7 +178,7 @@ class CategoryRepository extends EntityRepository implements ApiResourceReposito
         ];
         $availableSearchFilters = array_keys($constraints);
 
-        $exceptionMessage = '';
+        $exceptionMessages = [];
         foreach ($searchFilters as $property => $searchFilter) {
             if (!in_array($property, $availableSearchFilters)) {
                 throw new \InvalidArgumentException(sprintf(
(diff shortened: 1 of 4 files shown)


pchasle (approved):
:ok_hand:
You might have some tests to update

tseho:
> Updated namespaces from DateTime to DateTimeInterface as the former was removed since PHP 7.2
Can you elaborate ?

I'm not aware of `\DateTime` or  `\DateTimeInterface` to be removed from PHP.

RiffVentura:
> > Updated namespaces from DateTime to DateTimeInterface as the former was removed since PHP 7.2
> 
> Can you elaborate ?
> 
> I'm not aware of `\DateTime` or `\DateTimeInterface` to be removed from PHP.

My bad, I meant DateTime constants
Look at the changelog here:
https://www.php.net/manual/en/class.datetime.php
Basically constants are moved from DateTime to DateTimeInterface

tomglvng (approved):

tseho:
I edited my message:
> Ok, you are talking about the constants. And that's not correct, those constants were not removed, they were moved from the class to the interface but does not change anything because DateTime extends DateTimeInterface.
So, IMHO, this is an unnecessary change since both versions are valid.

RiffVentura:
> So, IMHO, this is an unnecessary change since both versions are valid.

I thought it can be dangerous to leave it be, since it can be removed later at any point
Do you want to me to revert it ?

tseho:
> I thought it can be dangerous to leave it be, since it can be removed later at any point
This is not a deprecation, they removed nothing, they just moved constants from a class to its interface.
They will not remove it, it's not dangerous.

tseho:
> Do you want to me to revert it ?

Would be better, I honestly think it was unnecessary.

tseho (approved):

PR #14300, opened 2021-04-20
PIM-9697: Move GoogleCloudStorageAdapter override to CE repository

mmetayer:

This class is needed for the growth edition

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -87,6 +87,7 @@
         "psr/event-dispatcher": "^1.0.0",
         "ramsey/uuid": "^3.7",
         "sensio/framework-extra-bundle": "^5.4",
+        "superbalist/flysystem-google-storage": "7.2.2",
         "symfony/acl-bundle": " 1.0.1",
         "symfony/asset": "4.4.18",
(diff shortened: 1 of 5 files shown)


composer.json
```diff
@@ -152,6 +152,7 @@
         "phpstan/phpstan": "^0.12.25",
         "phpstan/phpstan-symfony": "^0.12.6",
         "league/flysystem-aws-s3-v3": "^1.0",
+        "superbalist/flysystem-google-storage": "7.2.2",
```
mmetayer, line 155: I have a doubt here: I added this entry to the `require-dev`  section, as there is no service definition which relates to it in CE, and it is only used for running the specification.
However the class is now part of our codebase, so maybe it should be added add this to the `require`  section instead ...?

jmleroux: +1 for moving to 'require'

jmleroux (approved):
GTM after fix

AymericPlanche (approved):

Doodoune (approved):

PR #14301, opened 2021-04-20
CPM-159: Fix username

pierallard:
Before, the fullname returned " Mary Smith " instead of "Mary Smith"
This is only used by proposals, and it badly displayed username in this page.

### DIFF
--- a/src/Akeneo/UserManagement/Component/Model/User.php
+++ b/src/Akeneo/UserManagement/Component/Model/User.php
@@ -280,7 +280,7 @@ class User implements UserInterface
      */
     public function getFullName()
     {
-        return join(' ', array_filter([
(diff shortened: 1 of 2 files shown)


nmarniesse (approved):

src/Akeneo/UserManagement/Component/Model/User.php
```diff
@@ -280,13 +280,13 @@ public function getNameSuffix()
      */
     public function getFullName()
     {
-        return implode(' ', [
+        return join(' ', array_filter([
```
mmetayer, line 283:
```suggestion
        return \implode(' ', array_filter([
```

mmetayer (approved):

PR #14349, opened 2021-04-26
PIM-9820: Fix the Error 500 on the product grid with the date filter

anaisbaunelemaire:

Here is the description of the error:
On the product grid, on the filter "Created at", select "less than", 
then set a date and save the filter.
Reopen the filter, select "more than" > the date "disappears".
If you don't input a new date and save the filter > error 500 

I tested this case, too:
On the product grid, on the filter "Created at", select "more than", 
then set a date and save the filter.
Reopen the filter, select "less than" > the date "disappears".
If you don't input a new date and save the filter > error 500, also.

So, I fixed the issue returning false in isValidData() function when 
-> $data['type'] is "more than" and if $data['value']['start'] is null
-> $data['type'] is "less than" and if $data['value']['end'] is null

`OK`: Done / Validated
`-`: Not needed

### DIFF
--- a/src/Oro/Bundle/PimFilterBundle/spec/Filter/ProductValue/DateRangeFilterSpec.php
+++ b/src/Oro/Bundle/PimFilterBundle/spec/Filter/ProductValue/DateRangeFilterSpec.php
@@ -256,12 +256,12 @@ class DateRangeFilterSpec extends ObjectBehavior
 
     function it_does_not_parse_array_with_more_than_filter_and_no_start_date()
     {
-        $this->parseData(['value' => ['start' => null], 'type' => 3])->shouldReturn(false);
+        $this->parseData(['value' => ['start' => null], 'type' => DateRangeFilterType::TYPE_MORE_THAN])->shouldReturn(false);
     }
 
     function it_does_not_parse_array_with_less_than_filter_and_no_end_date()
     {
-        $this->parseData(['value' => ['end' => null], 'type' => 4])->shouldReturn(false);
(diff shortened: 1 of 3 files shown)


src/Oro/Bundle/PimFilterBundle/spec/Filter/ProductValue/DateRangeFilterSpec.php
```diff
@@ -254,6 +254,16 @@ function it_does_not_parse_array_with_not_datetime_type_end()
         $this->parseData(['value' => ['end' => 'tomorrow'], 'type' => 1])->shouldReturn(false);
     }
 
+    function it_does_not_parse_array_with_more_than_filter_and_no_start_date()
+    {
+        $this->parseData(['value' => ['start' => null], 'type' => 3])->shouldReturn(false);
```
tomglvng, line 259: Instead of 3, add the constant's name: it is more readable.

src/Oro/Bundle/PimFilterBundle/spec/Filter/ProductValue/DateRangeFilterSpec.php
```diff
@@ -254,6 +254,16 @@ function it_does_not_parse_array_with_not_datetime_type_end()
...
+    }
+
+    function it_does_not_parse_array_with_less_than_filter_and_no_end_date()
+    {
+        $this->parseData(['value' => ['end' => null], 'type' => 4])->shouldReturn(false);
```
tomglvng, line 264: Same here

tomglvng (approved):
GTM after fixes

StevenVAIDIE (approved):

PR #14370, opened 2021-04-29
Cxp 650 bis

willy-ahva:

- [x] Tests
- [ ] Migration & Installer
- [x] PM Validation (Story)
- [ ] Tech Doc


pchasle (approved):

src/Akeneo/Connectivity/Connection/front/src/connect/pages/Marketplace.tsx
```diff
@@ -69,10 +70,11 @@ export const Marketplace: FC = () => {
     const translate = useTranslate();
     const fetchMarketplaceUrl = useFetchMarketingUrl();
     const [marketplaceUrl, setMarketplaceUrl] = useState<string>('');
+    const dashboardHref = `#${useRoute('akeneo_connectivity_connection_audit_index')}`;
```
RiffVentura, line 73: For consistency reasons and because it's the preferred way: 
```javascript
import {useRouter} from '../../shared/router/use-router';
//...
const generateUrl = useRouter();
const dashboardHref = `#${generateUrl('akeneo_connectivity_connection_audit_index')}`
```

willy-ahva: We have both, I don't think it really matters. @pchasle any comment about it?

RiffVentura (approved):

src/Akeneo/Connectivity/Connection/front/src/connect/pages/Marketplace.tsx
pchasle: Good point @RiffVentura , but I agree that it doesn't matter much.
Going forward maybe we should use generateUrl as the default, but it's not really worth to relaunch a full CI for that.

PR #14380, opened 2021-04-30
PIM-9837: introduce a retry with increasing sleep delay to reduce back pressure

rfauglas:
Introduce a retry with increasing sleep delay to reduce back pressure because of overheat  message received  from ES.
This was not enough, indicating back pressure is also primarily caused by batch size.
Therefore I introduced a decreasing batch size. 
It introduces a much higher complexity: shall I write unit tests?

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/src/Akeneo/Pim/Enrichment/Bundle/Command/BackoffElasticSearchStateHandler.php
+++ b/src/Akeneo/Pim/Enrichment/Bundle/Command/BackoffElasticSearchStateHandler.php
@@ -9,7 +9,7 @@ use Symfony\Component\Console\Helper\ProgressBar;
 class BackoffElasticSearchStateHandler
 {
     public const RETRY_COUNTER = 10;
-    public const INTIAL_WAIT_DELAY = 10;
+    public const INITIAL_WAIT_DELAY = 10;
     public const BACKOFF_LOGARITHMIC_INCREMENT = 2;
 
     public function doIndex(iterable $chunkedCodes, ProgressBar $progressBar, \Closure $codesEsHandler): int
@@ -18,10 +18,11 @@ class BackoffElasticSearchStateHandler
 
         $progressBar->start();
         foreach ($chunkedCodes as $productModelCodes) {
+            $treatedBachSize= sizeof($productModelCodes);
             $batchSize = sizeof($productModelCodes);
             $backOverheat = false;
             $retryCounter = self::RETRY_COUNTER;
-            $waitDelay = self::INTIAL_WAIT_DELAY;
+            $waitDelay = self::INITIAL_WAIT_DELAY;
             $batchEsCodes = $productModelCodes;
             do {
                 if ($backOverheat) {
@@ -34,12 +35,12 @@ class BackoffElasticSearchStateHandler
                     array_splice($productModelCodes, 0, $batchSize);
                     $backOverheat = false;
                     $retryCounter = self::RETRY_COUNTER;
-                    $waitDelay = self::INTIAL_WAIT_DELAY;
+                    $waitDelay = self::INITIAL_WAIT_DELAY;
                     $indexedCount += count($batchEsCodes);
                 } catch (BadRequest400Exception $e) {
                     if ($e->getCode() == 429) {
                         $backOverheat = true;
-                        $waitDelay = $waitDelay + self::INTIAL_WAIT_DELAY; //Heuristic: linear increment
+                        $waitDelay = $waitDelay + self::INITIAL_WAIT_DELAY; //Heuristic: linear increment
                         $retryCounter--;
                         $batchSize = intdiv($batchSize, self::BACKOFF_LOGARITHMIC_INCREMENT); //Heuristic: logarithmics decrement
                     }
@@ -50,7 +51,7 @@ class BackoffElasticSearchStateHandler
                 throw $e;
             }
 
-            $progressBar->advance(count($productModelCodes));
+            $progressBar->advance($treatedBachSize);
         }
         $progressBar->finish();
 
--- a/src/Akeneo/Pim/Enrichment/Bundle/Command/BackoffElasticSearchStateHandler.php
+++ b/src/Akeneo/Pim/Enrichment/Bundle/Command/BackoffElasticSearchStateHandler.php
@@ -5,6 +5,8 @@ namespace Akeneo\Pim\Enrichment\Bundle\Command;
 
 use Elasticsearch\Common\Exceptions\BadRequest400Exception;
 use Symfony\Component\Console\Helper\ProgressBar;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\HttpFoundation\Response;
 
 /**
  * @copyright 2021 Akeneo SAS (http://www.akeneo.com)
@@ -16,49 +18,63 @@ class BackoffElasticSearchStateHandler
     public const INITIAL_WAIT_DELAY = 10;
     public const BACKOFF_LOGARITHMIC_INCREMENT = 2;
 
-    public function doIndex(iterable $chunkedCodes, ProgressBar $progressBar, \Closure $codesEsHandler): int
+    public function doIndex(iterable $chunkedCodes, ProgressBar $progressBar, \Closure $codesEsHandler, OutputInterface $output): int
     {
         $indexedCount = 0;
 
         $progressBar->start();
         foreach ($chunkedCodes as $codes) {
             $treatedBachSize= count($codes);
-            $batchSize = sizeof($codes);
-            $backOverheat = false;
-            $retryCounter = self::RETRY_COUNTER;
-            $waitDelay = self::INITIAL_WAIT_DELAY;
-            $batchEsCodes = $codes;
-            do {
-                if ($backOverheat) {
-                    echo("Sleeping before retry due to back pressure {$waitDelay} seconds, with batch size of {$batchSize} \n");
-                    sleep($waitDelay);
-                    $batchEsCodes = array_slice($productModelCodes, 0, $batchSize);
-                }
-                try {
-                    $codesEsHandler($batchEsCodes);
-                    array_splice($productModelCodes, 0, $batchSize);
-                    $backOverheat = false;
-                    $retryCounter = self::RETRY_COUNTER;
-                    $waitDelay = self::INITIAL_WAIT_DELAY;
-                    $indexedCount += count($batchEsCodes);
-                } catch (BadRequest400Exception $e) {
-                    if ($e->getCode() == 429) {
-                        $backOverheat = true;
-                        $waitDelay = $waitDelay + self::INITIAL_WAIT_DELAY; //Heuristic: linear increment
-                        $retryCounter--;
-                        $batchSize = intdiv($batchSize, self::BACKOFF_LOGARITHMIC_INCREMENT); //Heuristic: logarithmics decrement
-                    }
-                }
-            } while (($backOverheat && $retryCounter) || 0 !== count($productModelCodes));
 
-            if ($backOverheat && isset($e)) {
-                throw $e;
-            }
+            $this->bulkExecute($codes, $output, $codesEsHandler);
 
             $progressBar->advance($treatedBachSize);
         }
         $progressBar->finish();
 
         return $indexedCount;
+    }
+
+    protected function resetState(): array
+    {
+        $backOverheat = false;
+        $retryCounter = self::RETRY_COUNTER;
+        $waitDelay = self::INITIAL_WAIT_DELAY;
+        return array($backOverheat, $retryCounter, $waitDelay);
+    }
+
+    public  function bulkIndex($codes, OutputInterface $output, \Closure $codesEsHandler):int
+    {
+        $batchSize = count($codes);
+        $indexed = 0;
+        list($backOverheat, $retryCounter, $waitDelay) = $this->resetState();
+
+        do {
+            $batchEsCodes = $codes;
+            if ($backOverheat) {
+                $output->writeln("Sleeping before retry due to back pressure {$waitDelay} seconds, with batch size of {$batchSize} \n");
+                sleep($waitDelay);
+                $batchEsCodes = array_slice($codes, 0, $batchSize);
+            }
+
+            try {
+                $codesEsHandler($batchEsCodes);
+                array_splice($codes, 0, $batchSize);
+                list($backOverheat, $retryCounter, $waitDelay) = $this->resetState();
+                $indexed += count($batchEsCodes);
+            } catch (BadRequest400Exception $e) {
+                if ($e->getCode() == Response::HTTP_TOO_MANY_REQUESTS) {
+                    $backOverheat = true;
+                    $retryCounter--;
+                    $waitDelay = $waitDelay + self::INITIAL_WAIT_DELAY; //Heuristic: linear increment
+                    $batchSize = intdiv($batchSize, self::BACKOFF_LOGARITHMIC_INCREMENT); //Heuristic: logarithmics decrement
+                }
+            }
+        } while (($backOverheat && $retryCounter) || count($codes));
+
+        if ($backOverheat && isset($e)) {
+            throw $e;
+        }
+        return $indexed;
     }
 }
--- a/src/Akeneo/Pim/Enrichment/Bundle/Command/BackoffElasticSearchStateHandler.php
+++ b/src/Akeneo/Pim/Enrichment/Bundle/Command/BackoffElasticSearchStateHandler.php
@@ -17,7 +17,7 @@ class BackoffElasticSearchStateHandler
         $indexedCount = 0;
 
         $progressBar->start();
-        foreach ($chunkedCodes as $productModelCodes) {
+        foreach ($chunkedCodes as $codes) {
             $treatedBachSize= sizeof($productModelCodes);
             $batchSize = sizeof($productModelCodes);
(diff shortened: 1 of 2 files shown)


src/Akeneo/Pim/Enrichment/Bundle/Command/IndexProductModelCommand.php
```diff
@@ -126,7 +130,35 @@ private function doIndex(iterable $chunkedProductModelCodes, ProgressBar $progre
 
         $progressBar->start();
         foreach ($chunkedProductModelCodes as $productModelCodes) {
-            $this->productModelDescendantAndAncestorsIndexer->indexFromProductModelCodes($productModelCodes);
+            $codes=join($productModelCodes, " ");
+            echo("Product models codes: {$codes}\n");
```
fitn, line 134:
```suggestion
```

src/Akeneo/Pim/Enrichment/Bundle/Command/IndexProductModelCommand.php
```diff
@@ -126,7 +130,35 @@ private function doIndex(iterable $chunkedProductModelCodes, ProgressBar $progre
...
+
+            $backOverheat=false;
+            $retryCounter = self::RETRY_COUNTER;
+            $waitDelay = self::INTIAL_WAIT_DELAY;
+            do {
+                if ($backOverheat) {
+                    echo("Sleeping before retry due to back pressure {$waitDelay} seconds \n");
```
fitn, line 141:
```suggestion
```

src/Akeneo/Pim/Enrichment/Bundle/Command/IndexProductModelCommand.php
```diff
@@ -126,7 +130,35 @@ private function doIndex(iterable $chunkedProductModelCodes, ProgressBar $progre
...
+                    sleep($waitDelay);
+                    echo("Waking up for retry...\n");
```
fitn, line 143:
```suggestion
```

src/Akeneo/Pim/Enrichment/Bundle/Command/IndexProductModelCommand.php
```diff
@@ -126,7 +130,35 @@ private function doIndex(iterable $chunkedProductModelCodes, ProgressBar $progre
...
+                }
+                try {
+                    $this->productModelDescendantAndAncestorsIndexer->indexFromProductModelCodes($productModelCodes);
+                    $backOverheat=false;
+                } catch (BadRequest400Exception $e) {
+                    if ($e->getCode() == 429) {
+                        $backOverheat = true;
+                        $waitDelay = $waitDelay * self::RETRY_LOGARITHMIC_INCREMENT;
+                        $retryCounter--;
+                        echo("Back pressure exception received, {$retryCounter} retries remaining...\n");
```
fitn, line 153:
```suggestion
```

fitn (approved):

src/Akeneo/Pim/Enrichment/Bundle/Command/IndexProductModelCommand.php
```diff
@@ -24,6 +25,9 @@
  */
 class IndexProductModelCommand extends Command
 {
+    const RETRY_COUNTER = 3;
+    const INTIAL_WAIT_DELAY = 10;
```
fitn, line 29:
```suggestion
    const INITIAL_WAIT_DELAY = 10;
```

src/Akeneo/Pim/Enrichment/Bundle/Command/IndexProductModelCommand.php
```diff
@@ -126,7 +130,35 @@ private function doIndex(iterable $chunkedProductModelCodes, ProgressBar $progre
 
         $progressBar->start();
         foreach ($chunkedProductModelCodes as $productModelCodes) {
-            $this->productModelDescendantAndAncestorsIndexer->indexFromProductModelCodes($productModelCodes);
+            $codes=join($productModelCodes, " ");
+            echo("Product models codes: {$codes}\n");
+
+            $backOverheat=false;
+            $retryCounter = self::RETRY_COUNTER;
+            $waitDelay = self::INTIAL_WAIT_DELAY;
```
fitn, line 138:
```suggestion
            $waitDelay = self::INITIAL_WAIT_DELAY;
```

fitn (approved):

src/Akeneo/Pim/Enrichment/Bundle/Command/BackoffElasticSearchStateHandler.php
```diff
@@ -0,0 +1,60 @@
+<?php
+
```
nmarniesse, line 2:
```suggestion
declare(strict_types=1);
```

src/Akeneo/Pim/Enrichment/Bundle/Command/BackoffElasticSearchStateHandler.php
```diff
@@ -0,0 +1,60 @@
...
+
+namespace Akeneo\Pim\Enrichment\Bundle\Command;
+
+use Elasticsearch\Common\Exceptions\BadRequest400Exception;
+use Symfony\Component\Console\Helper\ProgressBar;
+
+class BackoffElasticSearchStateHandler
```
nmarniesse, line 9:
```suggestion
/**
 * @copyright 2021 Akeneo SAS (http://www.akeneo.com)
 * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
 */
class BackoffElasticSearchStateHandler
```

src/Akeneo/Pim/Enrichment/Bundle/Command/BackoffElasticSearchStateHandler.php
```diff
@@ -0,0 +1,60 @@
...
+    public const RETRY_COUNTER = 10;
+    public const INITIAL_WAIT_DELAY = 10;
+    public const BACKOFF_LOGARITHMIC_INCREMENT = 2;
+
+    public function doIndex(iterable $chunkedCodes, ProgressBar $progressBar, \Closure $codesEsHandler): int
+    {
+        $indexedCount = 0;
+
+        $progressBar->start();
+        foreach ($chunkedCodes as $productModelCodes) {
```
nmarniesse, line 20:
```suggestion
        foreach ($chunkedCodes as $codes) {
```

Could be productModel agnostic

src/Akeneo/Pim/Enrichment/Bundle/Command/IndexProductModelCommand.php
```diff
@@ -48,6 +50,15 @@ public function __construct(
         $this->productAndProductModelClient = $productAndProductModelClient;
         $this->productModelDescendantAndAncestorsIndexer = $productModelDescendantAndAncestorsIndexer;
         $this->connection = $connection;
+        $this->batchEsStateHandler = new BackoffElasticSearchStateHandler();
```
nmarniesse, line 53: Prefer create a service and inject it

nmarniesse (commented):
The `doIndex` method is very complex and not easy to understand.  
Small question: is it not enough to just handle the waitDelay, and not modify the batch size?  

Otherwise maybe cut this method in small private methods would break the high complexity of this one (but not improve it for testing...)

src/Akeneo/Pim/Enrichment/Bundle/Command/IndexProductModelCommand.php
rfauglas: I used the extract class refactoring of PHPStorm which was safer refactoring to me...

BitOne: I don't see the point of putting the doIndex inside another class. That's not exactly where the complexity is.

src/Akeneo/Pim/Enrichment/Bundle/Command/BackoffElasticSearchStateHandler.php
```diff
@@ -0,0 +1,64 @@
...
+                }
+                try {
+                    $codesEsHandler($batchEsCodes);
+                    array_splice($productModelCodes, 0, $batchSize);
+                    $backOverheat = false;
+                    $retryCounter = self::RETRY_COUNTER;
+                    $waitDelay = self::INITIAL_WAIT_DELAY;
+                    $indexedCount += count($batchEsCodes);
+                } catch (BadRequest400Exception $e) {
+                    if ($e->getCode() == 429) {
```
BitOne, line 45: You could use the constant from Symfony so it would be more explicit: https://github.com/symfony/http-foundation/blob/4.4/Response.php#L78

src/Akeneo/Pim/Enrichment/Bundle/Command/BackoffElasticSearchStateHandler.php
```diff
@@ -0,0 +1,64 @@
...
+        foreach ($chunkedCodes as $codes) {
+            $treatedBachSize= sizeof($codes);
+            $batchSize = sizeof($codes);
+            $backOverheat = false;
+            $retryCounter = self::RETRY_COUNTER;
+            $waitDelay = self::INITIAL_WAIT_DELAY;
+            $batchEsCodes = $codes;
+            do {
+                if ($backOverheat) {
+                    echo("Sleeping before retry due to back pressure {$waitDelay} seconds, with batch size of {$batchSize} \n");
```
BitOne, line 33: Output->writeln as it"s a command

src/Akeneo/Pim/Enrichment/Bundle/Command/BackoffElasticSearchStateHandler.php
```diff
@@ -0,0 +1,64 @@
...
+                    sleep($waitDelay);
+                    $batchEsCodes = array_slice($productModelCodes, 0, $batchSize);
+                }
+                try {
+                    $codesEsHandler($batchEsCodes);
+                    array_splice($productModelCodes, 0, $batchSize);
+                    $backOverheat = false;
+                    $retryCounter = self::RETRY_COUNTER;
+                    $waitDelay = self::INITIAL_WAIT_DELAY;
```
BitOne, line 42: It is changed a bit later in the catch, so I think it gets reset in each loop execution... And same for the retryCounter above. Am I missing something?

src/Akeneo/Pim/Enrichment/Bundle/Command/BackoffElasticSearchStateHandler.php
```diff
@@ -0,0 +1,64 @@
...
+                    $indexedCount += count($batchEsCodes);
+                } catch (BadRequest400Exception $e) {
+                    if ($e->getCode() == 429) {
+                        $backOverheat = true;
+                        $waitDelay = $waitDelay + self::INITIAL_WAIT_DELAY; //Heuristic: linear increment
+                        $retryCounter--;
+                        $batchSize = intdiv($batchSize, self::BACKOFF_LOGARITHMIC_INCREMENT); //Heuristic: logarithmics decrement
+                    }
+                }
+            } while (($backOverheat && $retryCounter) || sizeof($productModelCodes));
```
BitOne, line 52: sizeof($productModelCodes)
```suggestion
            } while (($backOverheat && $retryCounter) || 0 !== count($productModelCodes));
```
Still a bit dangerous as iterable doesn't necessarily mean countable.

src/Akeneo/Pim/Enrichment/Bundle/Command/BackoffElasticSearchStateHandler.php
```diff
@@ -0,0 +1,64 @@
...
+            $batchSize = sizeof($codes);
+            $backOverheat = false;
+            $retryCounter = self::RETRY_COUNTER;
+            $waitDelay = self::INITIAL_WAIT_DELAY;
+            $batchEsCodes = $codes;
+            do {
+                if ($backOverheat) {
+                    echo("Sleeping before retry due to back pressure {$waitDelay} seconds, with batch size of {$batchSize} \n");
+                    sleep($waitDelay);
+                    $batchEsCodes = array_slice($productModelCodes, 0, $batchSize);
```
BitOne, line 35: From where this `productModelCodes` variable is coming?
We only have `chunkedCodes` & `codes` above.

Unit testing is required :)

src/Akeneo/Pim/Enrichment/Bundle/Command/BackoffElasticSearchStateHandler.php
```diff
@@ -0,0 +1,64 @@
...
+    public const INITIAL_WAIT_DELAY = 10;
+    public const BACKOFF_LOGARITHMIC_INCREMENT = 2;
+
+    public function doIndex(iterable $chunkedCodes, ProgressBar $progressBar, \Closure $codesEsHandler): int
+    {
+        $indexedCount = 0;
+
+        $progressBar->start();
+        foreach ($chunkedCodes as $codes) {
+            $treatedBachSize= sizeof($codes);
```
BitOne, line 25:
```suggestion
            $treatedBachSize= count($codes);
```

src/Akeneo/Pim/Enrichment/Bundle/Command/IndexProductModelCommand.php
rfauglas: Well concerning  @nmarniesse  says it doIndex is complex, @BitOne says there is no complexity, i'd bend towards @nmarniesse... doIndex shows  a cyclomatic complexity of 8 and a  npath  over 8 as well...
The  reasons I did extract it are
- I did extract it to isolate this  logic of "backing off", from the rest of the logic of the command which makes already to many things (sql calls,  symfony cmd integration, etc...).
- It is easier to test, than within the command (i have to change it's visibility into  the command  for testing which is not very neat...).
(- additionaly, it makes sense to put this logic for  product indexing aswell....)

src/Akeneo/Pim/Enrichment/Bundle/Command/BackoffElasticSearchStateHandler.php
rfauglas: Yes, I'm afraid...

src/Akeneo/Pim/Enrichment/Bundle/Command/BackoffElasticSearchStateHandler.php
rfauglas: Done

src/Akeneo/Pim/Enrichment/Bundle/Command/BackoffElasticSearchStateHandler.php
rfauglas: I changed it list($backOverheat, $retryCounter, $waitDelay) = $this->resetState(); to reveal the intention behind, it is the same as LOC 28...
when a request succeed, we can reset backoff State...( I hesitated to introduce a State object...). That is also a heuristic choice...

src/Akeneo/Pim/Enrichment/Bundle/Command/BackoffElasticSearchStateHandler.php
rfauglas: Done.

src/Akeneo/Pim/Enrichment/Bundle/Command/BackoffElasticSearchStateHandler.php
rfauglas: I changed it to a resetState method to better reveal the intention... factored with LOC 30.

src/Akeneo/Pim/Enrichment/Bundle/Command/IndexProductModelCommand.php
rfauglas: To further lower complexity we could introduce a state pattern holding the # states... 
But we have only 2 states and a termination condition...not sure it is worth it...

BitOne: You're just moving complexity into another class, which create at the end of the day more complexity :)

The complexity of the function could be reduced by dividing the function in smaller parts. Moving the function around won't change much and adds complexity because of the closure.

And by the way, do we really need to have an increase in the delay, or only reducing the size of the batches works as well? If if works without the delay part, that would reduce the complexity of the function.

src/Akeneo/Pim/Enrichment/Bundle/Command/BackoffElasticSearchStateHandler.php
BitOne: Can you push your changes please?

src/Akeneo/Pim/Enrichment/Bundle/Command/IndexProductModelCommand.php
rfauglas: Concerning the "extract class" and complexity, my argumentation is not around complexity, I agree with you indirection doesn't really lowers complexity...  But concerning the testability, as far as I know Phpspec  doesn't support partial mock nor testing of private methods... so "extract method" does improve lisibility, but doesn't really reduce complexity from a testing purpose. 
Concerning the sleep delay, it was a heuristic, it made sense in a back pressure context, but here it seems ES response rather seems related to request size, so I agree we can remove it, this one lowers complexity seriously ;-)....

PR #14384, opened 2021-05-03
CXP-667: Update the Connection settings info panel

tomglvng:
Capture du 2021-05-03 16-30-31
Capture du 2021-05-04 09-45-41
Capture du 2021-05-04 09-44-15

### DIFF
--- a/src/Akeneo/Connectivity/Connection/back/Infrastructure/Symfony/Resources/translations/jsmessages.en_US.yml
+++ b/src/Akeneo/Connectivity/Connection/back/Infrastructure/Symfony/Resources/translations/jsmessages.en_US.yml
@@ -156,9 +156,9 @@ akeneo_connectivity.connection:
         flash:
             success: Connection successfully deleted.
     helper:
-        title: 👋 Welcome to the feature that will properly connect your PIM!
+        title: '{0} For now, you have no connection to link your PIM to third-party applications.|{1} You have 1 connection linking your PIM to third-party applications. Well done, this way your PIM is well integrated into its ecosystem.|]1,Inf] You have {{ count }} connections linking your PIM to third-party applications. Well done, this way your PIM is well integrated into its ecosystem.'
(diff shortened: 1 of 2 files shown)


tseho (approved):

src/Akeneo/Connectivity/Connection/back/Infrastructure/Symfony/Resources/translations/jsmessages.en_US.yml
```diff
@@ -156,9 +156,9 @@ akeneo_connectivity.connection:
         flash:
             success: Connection successfully deleted.
     helper:
-        title: 👋 Welcome to the feature that will properly connect your PIM!
+        title: '{0} For now, you have no connection to link your PIM to third-party applications.|{1} You have 1 connection linking your PIM to third-party applications. Well done, this way your PIM is well integrated into its ecosystem.|]1,Inf] You have {{ count }} connections linking your PIM to third-party applications. Well done, this way your PIM is well integrated into its ecosystem.'
```
RiffVentura, line 159: I think it would be better to not inline pluralization as it becomes hard to read and update

RiffVentura (approved):

src/Akeneo/Connectivity/Connection/back/Infrastructure/Symfony/Resources/translations/jsmessages.en_US.yml
tomglvng: You're right, but I think it's ok for this one. If we have another case like that, we must find another way.

src/Akeneo/Connectivity/Connection/back/Infrastructure/Symfony/Resources/translations/jsmessages.en_US.yml
```diff
@@ -156,9 +156,9 @@ akeneo_connectivity.connection:
...
         description: A PIM without any third-party integrations is kinda useless, isn't it? Thanks to connections, you can declare all the data flows that will be pushed inside and pulled outside your PIM.
```
pchasle, line 160: Since the description have been removed from the page it would have been better to remove the translation as well :)

PR #14391, opened 2021-05-04
Remove default exposed ports for MySQL, ES and Selenium

jmleroux:
These ports are only useful for debug and can be easily made available in a docker-compose override and they are causing crashes on CI when running in parallel multi-container environment.

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -53,8 +53,6 @@ services:
     volumes:
       - './:/srv/pim:ro'
       - '/dev/shm:/dev/shm'
-    ports:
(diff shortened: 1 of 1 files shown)


BitOne:
I would prefer a nice simple sed to remove the port exposition on the CI more than changing it for everybody. I know that devs can add their own override, but it's a pain to do so...

adesaegher (approved):

PR #14402, opened 2021-05-04
API-1581: Remove our informative logs in Events API feature

pchasle:
Remove unused Events API logs:

Removed:
- Skip own PIM Events
- Event data build error
- Log resource version
- API Events build time

Kept:
- Events API limit reached
- Events API Request sent (success, failure and timeout)


tseho (approved):
As discussed, we need to check with PY before merging, to verify that we broke no metrics with this.

tomglvng (approved):

pchasle:
> As discussed, we need to check with PY before merging, to verify that we broke no metrics with this.

Seen with PY, it will break some charts/average on the build time, build errors ... but it's not an issue since we don't need them anymore.

PR #14440, opened 2021-05-07
PIM-9835 PIM-9834: Fix "MySQL server has gone away" error when importing a lot of attribute options

nmarniesse:

https://akeneo.atlassian.net/browse/PIM-9834

When importing new options to an attribute that already have a lot of options, a "MySQL server has gone away" could happen. 
When we save and flush the attribute options in the database:
- Doctrine's unit of work detect the attribute has changed, save the new attribute and send events.
- A listener catches this event to create a new version of the attribute
-  To build the new version we normalized the attribute and its options. The `getOptions()` method on attribute model fetches all options in a single query without limit => MySQL crashes when too many options are in DB.  

Now a new query is added to fetch the option codes with a search after method. It avoid MySQL to crash.

- [x] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc


src/Akeneo/Pim/Structure/Component/Normalizer/Versioning/AttributeNormalizer.php
```diff
@@ -84,19 +78,17 @@ public function hasCacheableSupportsMethod(): bool
...
-        if ($options->isEmpty()) {
-            return null;
-        }
-
-        $data = [];
-        foreach ($options as $option) {
-            $data[] = 'Code:' . $option->getCode();
+        $attributeOptionCodes = $this->getAttributeOptionCodes->forAttributeCode($attribute->getCode());
+        $normalizedOption = null;
+        foreach ($attributeOptionCodes as $attributeOptionCode) {
```
nmarniesse, line 83: It should be better to add a limit? Maybe do some additional tests to check how many options it's possible to handle

romaindak (approved):

jmleroux (approved):

pchasle (approved):

PR #14516, opened 2021-05-18
API 1579 - Be able to filter categories on their updated date

RiffVentura:
Added ability to filter categories on their updated date for  GET categories REST endpoint:
New search filter on updated date with `>` as its only operator that accepts date values in ISO-8601 format

- [x] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc


tests/back/Pim/Enrichment/Integration/Doctrine/ORM/Repository/ExternalApi/CategoryRepositoryApiResourceIntegration.php
```diff
@@ -149,6 +149,50 @@ public function test_to_search_non_root_categories_only(): void
...
+            0
+        );
+    }
+
+    public function test_to_search_categories_after_update_date(): void
+    {
+        $this->initFixtures();
+
+        $connection = $this->get('database_connection');
+        $connection->exec('UPDATE pim_catalog_category SET updated="2019-05-15 16:27:00"');
```
tomglvng, line 181: For category updating, you can use the category updater. See https://github.com/akeneo/pim-community-dev/blob/master/tests/back/Pim/Enrichment/Integration/Doctrine/Common/Saver/CategorySaverIntegration.php#L220

RiffVentura: I normally would but category updater is not ready for a Datetime field moreover `updated` field is supposed to be read-only

tseho: Like Rifat said, the `updated` is read only, you should not be able to update it with the category updater.
This is the cleanest way to do it right now.

tests/back/Pim/Enrichment/Integration/Doctrine/ORM/Repository/ExternalApi/CategoryRepositoryApiResourceIntegration.php
```diff
@@ -149,6 +149,50 @@ public function test_to_search_non_root_categories_only(): void
...
+    {
+        $this->expectException(\InvalidArgumentException::class);
+
+        $this->getRepository()->searchAfterOffset(
+            ['updated' => [['operator' => '>', 'value' => '2019-06-09 12:00:00']]],
+            ['code' => 'ASC'],
+            10,
+            0
+        );
+    }
```
tseho, line 174: The validation of operator/value should not be in the integration test, because for each integration test, we set up the database and you don't need it. It should be an unit test instead.

There is no spec on this repo yet, if you want to add those tests, you can easily add a new spec file with only those.

tseho (approved):

willy-ahva (approved):

PR #14526, opened 2021-05-19
PIM-9869: Fix download log in job tracker is only available when log is located in the fpm server

StevenVAIDIE:
The download log is not present in cloud version because we called file_exist function. 

The file does not exist because the file generated is present in the pim-daemon-server not in the fpm-server. Now we use the filesystem dedicated to the log and check if the file exist.

We only check if there is a logFile related to the JobExecution because when the job is not started the jobFile is empty.
I remove the `is_file($jobExecution->getLogFile()` statement because it required that the file is located in the same server. The \Akeneo\Tool\Component\Connector\LogArchiver already check this.

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -69,6 +69,7 @@
 - PIM-9741: Fix choice filter mask not closing when selecting with keyboard
 - PIM-9806: Enable authentication temporary lock to protect against brute force attack
 - PIM-9864: Fix 500 error when using DateTime filter with invalid value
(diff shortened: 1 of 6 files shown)


ValentinMumble (approved):

rfauglas (approved):
Maybe some additional test? 
Could the controler be unit tested?

PR #14549, opened 2021-05-21
PIM-9863 Improve ES backoff pressure handling algorithm.

rfauglas:
When ES undergoes heavy BO office due to node synchronisation. it sends a 429 error answer.
I first try to introduce increasing retry temporisation. Which was of no effect.
I tried then to iterativel reduce batch size, which solved the problem on the customer side.
It was then decided to remove the temporisation since it is of no effect (although it is what is hinted by a 429 response).

- [x] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/src/Akeneo/Pim/Enrichment/Bundle/Command/BackoffElasticSearchStateHandler.php
+++ b/src/Akeneo/Pim/Enrichment/Bundle/Command/BackoffElasticSearchStateHandler.php
@@ -9,7 +9,7 @@ use Symfony\Component\Console\Helper\ProgressBar;
 class BackoffElasticSearchStateHandler
 {
     public const RETRY_COUNTER = 10;
-    public const INTIAL_WAIT_DELAY = 10;
+    public const INITIAL_WAIT_DELAY = 10;
     public const BACKOFF_LOGARITHMIC_INCREMENT = 2;
 
     public function doIndex(iterable $chunkedCodes, ProgressBar $progressBar, \Closure $codesEsHandler): int
@@ -18,10 +18,11 @@ class BackoffElasticSearchStateHandler
 
         $progressBar->start();
         foreach ($chunkedCodes as $productModelCodes) {
+            $treatedBachSize= sizeof($productModelCodes);
             $batchSize = sizeof($productModelCodes);
             $backOverheat = false;
             $retryCounter = self::RETRY_COUNTER;
-            $waitDelay = self::INTIAL_WAIT_DELAY;
+            $waitDelay = self::INITIAL_WAIT_DELAY;
             $batchEsCodes = $productModelCodes;
             do {
                 if ($backOverheat) {
@@ -34,12 +35,12 @@ class BackoffElasticSearchStateHandler
                     array_splice($productModelCodes, 0, $batchSize);
                     $backOverheat = false;
                     $retryCounter = self::RETRY_COUNTER;
-                    $waitDelay = self::INTIAL_WAIT_DELAY;
+                    $waitDelay = self::INITIAL_WAIT_DELAY;
                     $indexedCount += count($batchEsCodes);
                 } catch (BadRequest400Exception $e) {
                     if ($e->getCode() == 429) {
                         $backOverheat = true;
-                        $waitDelay = $waitDelay + self::INTIAL_WAIT_DELAY; //Heuristic: linear increment
+                        $waitDelay = $waitDelay + self::INITIAL_WAIT_DELAY; //Heuristic: linear increment
                         $retryCounter--;
                         $batchSize = intdiv($batchSize, self::BACKOFF_LOGARITHMIC_INCREMENT); //Heuristic: logarithmics decrement
                     }
@@ -50,7 +51,7 @@ class BackoffElasticSearchStateHandler
                 throw $e;
             }
 
-            $progressBar->advance(count($productModelCodes));
+            $progressBar->advance($treatedBachSize);
         }
         $progressBar->finish();
 
--- a/src/Akeneo/Pim/Enrichment/Bundle/Command/BackoffElasticSearchStateHandler.php
+++ b/src/Akeneo/Pim/Enrichment/Bundle/Command/BackoffElasticSearchStateHandler.php
@@ -5,6 +5,8 @@ namespace Akeneo\Pim\Enrichment\Bundle\Command;
 
 use Elasticsearch\Common\Exceptions\BadRequest400Exception;
 use Symfony\Component\Console\Helper\ProgressBar;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\HttpFoundation\Response;
 
 /**
  * @copyright 2021 Akeneo SAS (http://www.akeneo.com)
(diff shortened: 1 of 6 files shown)


src/Akeneo/Pim/Enrichment/Bundle/Command/BackoffElasticSearchStateHandler.php
```diff
@@ -1,60 +1,76 @@
...
-                    }
-                }
-            } while (($backOverheat && $retryCounter) || sizeof($productModelCodes));
+        $this->retryCounter = $retryCounter;
+        $this->backoffLogarithmicIncrement = $backoffLogarithmicIncrement;
+    }
+
+    protected function resetState(): array
+    {
+        return array(false, $this->retryCounter);
```
ahocquard, line 37:
```suggestion
        return [false, $this->retryCounter];
```

`array` is almost not used anymore

src/Akeneo/Pim/Enrichment/Bundle/Command/BackoffElasticSearchStateHandler.php
```diff
@@ -1,60 +1,76 @@
...
-    public const BACKOFF_LOGARITHMIC_INCREMENT = 2;
+    const RETRY_COUNTER = 10;
+    const BACKOFF_LOGARITHMIC_INCREMENT = 2;
+
+    private int $retryCounter;
+    private int $backoffLogarithmicIncrement;
 
-    public function doIndex(iterable $chunkedCodes, ProgressBar $progressBar, \Closure $codesEsHandler): int
+    /**
+     * BackoffElasticSearchStateHandler constructor.
```
ahocquard, line 25: to remove as typehint + function name is already explicit in the signature

src/Akeneo/Pim/Enrichment/Bundle/Command/BackoffElasticSearchStateHandler.php
```diff
@@ -1,60 +1,76 @@
...
+    }
 
-            if ($backOverheat && isset($e)) {
-                throw $e;
+    public function bulkExecute(array $codes, BulkEsHandlerInterface $codesEsHandler):int
+    {
+        $initialBatchSize = count($codes);
+        $batchSize = $initialBatchSize;
+        $indexed = 0;
+        list($backOverheat, $retryCounter) = $this->resetState();
```
ahocquard, line 45:
```suggestion
        [$backOverheat, $retryCounter] = $this->resetState();
```

(same as `array`, not so used anymore)

src/Akeneo/Pim/Enrichment/Bundle/Command/IndexProductModelCommand.php
```diff
@@ -123,17 +122,41 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
     }
 
+    private function doIndex(iterable $chunkedCodes, ProgressBar $progressBar, BulkEsHandlerInterface $codesEsHandler, OutputInterface $output): int
+    {
+        $indexedCount = 0;
+
+        $progressBar->start();
+        foreach ($chunkedCodes as $codes) {
+            $treatedBachSize = $this->batchEsStateHandler->bulkExecute($codes, $codesEsHandler);
+            $indexedCount+=$indexedCount;
```
ahocquard, line 152: If I read correctly, this variable represents the number of batch and not the number of index entities. Therefore, the printed output should be wrong.

src/Akeneo/Pim/Enrichment/Bundle/Command/BackoffElasticSearchStateHandler.php
```diff
@@ -1,60 +1,76 @@
...
+ */
 class BackoffElasticSearchStateHandler
 {
-    public const RETRY_COUNTER = 10;
-    public const INITIAL_WAIT_DELAY = 10;
-    public const BACKOFF_LOGARITHMIC_INCREMENT = 2;
+    const RETRY_COUNTER = 10;
+    const BACKOFF_LOGARITHMIC_INCREMENT = 2;
+
+    private int $retryCounter;
```
ahocquard, line 21: These two stateful variables could be deleted. You could rely only on const, as you never modify it.

ahocquard (commented):
Not used for asset/ref entities indexation? No issue with that?

src/Akeneo/Pim/Enrichment/Bundle/Command/BackoffElasticSearchStateHandler.php
rfauglas: Const  are default value. I  initialize instance variables in constructor and allow overriding...

rfauglas:
> Not used for asset/ref entities indexation? No issue with that?
You're right. I wrote it to be reusable... But I'm not aware of existing problems...

src/Akeneo/Pim/Enrichment/Bundle/Command/BackoffElasticSearchStateHandler.php
rfauglas: IDE generated crap... I would be great to share common IDE setup :-(   ce+ee(+es)*  (4.0 + 5.0 + master...)

src/Akeneo/Pim/Enrichment/Bundle/Command/BackoffElasticSearchStateHandler.php
ahocquard: Ah ok, for testability I suppose :+1:

src/Akeneo/Pim/Enrichment/Bundle/Command/BackoffElasticSearchStateHandler.php
```diff
@@ -1,60 +1,71 @@
...
-            if ($backOverheat && isset($e)) {
-                throw $e;
+    public function bulkExecute(array $codes, BulkEsHandlerInterface $codesEsHandler):int
+    {
+        $initialBatchSize = count($codes);
+        $batchSize = $initialBatchSize;
+        $indexed = 0;
+        [$backOverheat, $retryCounter] = $this->resetState();
+
+        do {
```
ahocquard, line 42: I could not read it and safely assert "it does not lose code when running if an error occurs" or "I can't have an infinite loop".
It's pretty complicated, as it reset a state that is also modified if some conditions occurred (exception).

Also, you get the exception from the outer scope of the exception. This relies on a bad side effect of PHP, that is only cleaning internal scope at the end of the function.

As the problem is "divide and conquer", I would use recursivity to have cleaner scope and less side effects.

```
public function bulkExecute(array $codes, BulkEsHandlerInterface $codesEsHandler):int {
    $numberRetry = 0;
    $this->executeAttempt([$codes], $codesEsHandler, $numberRetry);
}

/**
* batchOfCodes is an array of array. Each sub-array represents a batch of codes to index.
* [
*     ['code_1', 'code_2'],
*     ['code_3', 'code_4'],
* ]
**/
public function executeAttempt(array $batchOfCodes, BulkEsHandlerInterface $codesEsHandler, int $numberRetry): int {
    
    foreach ($batchOfCodes as $codes) {
	    try {
	    	$codesEsHandler->bulkExecute($codes);
	    } catch (BadRequest400Exception $e) {
	        if ($e->getCode() == Response::HTTP_TOO_MANY_REQUESTS && $numberRetry < $this->maxNumberRetry) {
	        	$batchSize = intdiv(count($codes), $this->backoffLogarithmicIncrement);
	        	$smallerBatchOfCodes = array_chunk($codes, $batchSize);
	        	$this->executeAttempt($batchOfCodes, $codesEsHandler, $numberRetry++);
	        } else {
	        	throw $e;
	        }
	    }
	    
    }

   
}

File: src/Akeneo/Pim/Enrichment/Bundle/Command/BackoffElasticSearchStateHandler.php (hunk 5 above)
Comment (rfauglas) [reply src/Akeneo/Pim/Enrichment/Bundle/Command/BackoffElasticSearchStateHandler.php:21]: Yes you're right!

File: tests/back/Pim/Enrichment/Specification/Bundle/Command/BackoffElasticSearchStateHandlerSpec.php [hunk 7]
```diff
@@ -1,14 +1,56 @@
...
+    public function it_will_stop_after_a_403_response(BulkEsHandlerInterface $bulkEsHandler)
+    {
+        $codes = range(1, 17);
+        $bulkEsHandler->bulkExecute(Argument::any())->willThrow(new BadRequest400Exception("", Response::HTTP_FORBIDDEN));
+        $this->shouldThrow(BadRequest400Exception::class)->during('bulkExecute', [$codes,$bulkEsHandler]);
+        $bulkEsHandler->bulkExecute(Argument::any())->shouldHaveBeenCalledOnce();
+    }
+
+    public function it_will_make_several_attempts_reducing_batch_size(BulkEsHandlerInterface $bulkEsHandler) {
+        $codes = range(1, 17);
```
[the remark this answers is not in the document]
rfauglas: It is mostly for ease of writing/testing...

src/Akeneo/Pim/Enrichment/Bundle/Command/IndexProductModelCommand.php
```diff
@@ -123,17 +122,41 @@ protected function execute(InputInterface $input, OutputInterface $output)
             return self::ERROR_CODE_USAGE;
         }
 
-        $numberOfIndexedProducts = $this->batchEsStateHandler->doIndex($chunkedProductModelCodes, new ProgressBar($output, $productModelCount),
-            function ($codes) {
-                $this->productModelDescendantAndAncestorsIndexer->indexFromProductModelCodes($codes);
-            });
+        $batchESHandler = new class($this->productModelDescendantAndAncestorsIndexer) implements BulkEsHandlerInterface {
```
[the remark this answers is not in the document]
rfauglas: Done

src/Akeneo/Pim/Enrichment/Bundle/Command/BackoffElasticSearchStateHandler.php
rfauglas: You solution improves definitely readability.
Note I had to adapt spec: the intention is not the same. I restart at "full" batch size after backend recovery, whereas recursion means handling consecutive reduced sizes batch. But since this algorithm is a heuristic it is better to favor readability to some "premature optimisation...". I was aware of this readability limitations that I could improve  using some state patterns, but i usually do it using inner class... which is as far as I known exotic idiom and probably not so testable using phpspec... So recursivity is the best... ;-)

ahocquard (approved):
Thanks for taking in account my suggestion :+1:

src/Akeneo/Pim/Enrichment/Bundle/Command/BulkEsHandlerInterface.php
```diff
@@ -0,0 +1,13 @@
...
+
+namespace Akeneo\Pim\Enrichment\Bundle\Command;
+
+/**
+ * @copyright 2021 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+interface BulkEsHandlerInterface
+{
+    public function bulkExecute(array $codes);
```
tseho, line 12: What's the type of the returned value ? void ? int ?
You seems to use the returned value here:
https://github.com/akeneo/pim-community-dev/pull/14549/files#diff-7d221962eba1189c6e46e87b09812ea5152cfe9cc82dcae3a6c3b04a3de0d112R52
but nothing is returned :
https://github.com/akeneo/pim-community-dev/pull/14549/files#diff-3edd925a2062ef54ff37260546dcf6b9a3aff3411e449b724a2487e0b1fe3cdcR134

src/Akeneo/Pim/Enrichment/Bundle/Command/BackoffElasticSearchStateHandler.php
```diff
@@ -1,60 +1,65 @@
...
-                }
-            } while (($backOverheat && $retryCounter) || sizeof($productModelCodes));
+        $this->maxNumberRetry = $retryCounter;
+        $this->backoffLogarithmicIncrement = $backoffLogarithmicIncrement;
+    }
 
-            if ($backOverheat && isset($e)) {
-                throw $e;
-            }
+    protected function resetState(): array
```
ahocquard, line 30: unused now

src/Akeneo/Pim/Enrichment/Bundle/Command/BulkEsHandlerInterface.php
rfauglas: done!

tseho (approved):

PR #14560, opened 2021-05-24
CXP-679: add profile column to the user

willy-ahva:

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc


tseho (approved):

RiffVentura (approved):

src/Akeneo/UserManagement/Bundle/Resources/public/js/form/fields/select-profile.ts
```diff
@@ -0,0 +1,44 @@
+import * as $ from 'jquery';
+import * as _ from 'underscore';
+const __ = require('oro/translator');
+const BaseSelect = require('pim/form/common/fields/select');
+const FetcherRegistry = require('pim/fetcher-registry');
+const containerTemplate = require('pim/templates/user/form/fields/select-profile-container');
+
+type InterfaceNormalizedProfile = {
```
pchasle, line 8:
```suggestion
type NormalizedProfile = {
```

src/Akeneo/UserManagement/Bundle/Resources/public/js/form/fields/select-profile.ts
```diff
@@ -0,0 +1,44 @@
...
+  code: string;
+  label: string;
+};
+
+class SelectProfile extends BaseSelect {
+  // @ts-ignore
```
pchasle, line 14: Why the `@ts-ignore`?

If this is just an issue with the typing you can use an `object`, `unknown` or `any`

src/Akeneo/UserManagement/Bundle/Resources/translations/jsmessages.en_US.yml
```diff
@@ -35,6 +35,16 @@ pim_enrich:
                 delete:
                     confirm: Are you sure you want to delete this user?
 
+pim_user.profile:
```
pchasle, line 38: Shouldn't that be under `pim_user_management` instead?

src/Akeneo/UserManagement/Bundle/Resources/translations/jsmessages.en_US.yml
```diff
@@ -35,6 +35,16 @@ pim_enrich:
...
+    why_is_it_needed: Why do we need this information?
+    product_manager: I manage product catalogs.
+    redactor: I enrich product data.
+    pim_integrator: I integrate Akeneo PIM into our IT ecosystem.
+    pim_administrator: I administrate Akeneo PIM.
+    asset_manager: I manage assets in the PIM.
+    translator: I translate product, asset, and/or reference entity data.
+    third_party_developer: I develop solutions connected with Akeneo PIM.
```
pchasle, line 46: I would put the profile options under a distinct key like `profile_labels` or something like that.

pchasle (approved):

src/Akeneo/UserManagement/Bundle/Resources/public/js/form/fields/select-profile.ts
willy-ahva: because the containerTemplate is never user, it comes from gran parent and I just needed to override it

PR #14570, opened 2021-05-26
PIM-9876: Purge outdated data before consolidating dashboard data (master)

rouliane:
This PR fixes 2 issues in DQI :
* It puts back the purge of old product's scores. When we switched from 2 axis rates to a single product score we removed by mistake the purge of product scores.
* It switches the order of the purge and the consolidation. This will avoid a bad situation when the consolidation fails and the purge is never executed, so the table keeps growing. It may seems weird to do a purge before a consolidation, but it's OK because in the consolidation process we only take the most recent product score (it's a bit more complicated than that because it depends of the date that is passed to the consolidation step of the job, but basically it's the same thing).

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/src/Akeneo/Pim/Automation/DataQualityInsights/back/Application/Consolidation/PurgeOutdatedData.php
+++ b/src/Akeneo/Pim/Automation/DataQualityInsights/back/Application/Consolidation/PurgeOutdatedData.php
@@ -6,6 +6,7 @@ namespace Akeneo\Pim\Automation\DataQualityInsights\Application\Consolidation;
 
 use Akeneo\Pim\Automation\DataQualityInsights\Domain\Model\Write\DashboardPurgeDateCollection;
 use Akeneo\Pim\Automation\DataQualityInsights\Domain\Repository\DashboardScoresProjectionRepositoryInterface;
+use Akeneo\Pim\Automation\DataQualityInsights\Domain\Repository\ProductScoreRepositoryInterface;
 use Akeneo\Pim\Automation\DataQualityInsights\Domain\ValueObject\ConsolidationDate;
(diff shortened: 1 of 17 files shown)


src/Akeneo/Pim/Automation/DataQualityInsights/back/Application/Consolidation/PurgeOutdatedData.php
```diff
@@ -51,4 +63,10 @@ public function purgeDashboardProjectionRatesFrom(\DateTimeImmutable $date): voi
 
         $this->dashboardScoresProjectionRepository->purgeRates($purgeDates);
     }
+
+    private function purgeProductScoresFrom(\DateTimeImmutable $date): void
+    {
+        $purgeDate = $date->modify(sprintf('-%d DAY', self::EVALUATIONS_RETENTION_DAYS));
```
olivierpontier, line 69: During the periodic task job we already define the date to `NOW -1 DAY`: https://github.com/akeneo/pim-community-dev/blob/master/src/Akeneo/Pim/Automation/DataQualityInsights/back/Infrastructure/Symfony/Command/SchedulePeriodicTasksCommand.php#L46
Does it means we will keep 2 days?

rouliane: Yes, because we need to keep the day before in case the consolidation job does not finish during the day it has been launched. So we will always keep the current ("unfinished") day, and the previous (complete) day

olivierpontier (approved):

olivierpontier (approved):

PR #14606, opened 2021-05-31
POC: ack job message after pull in all cases

nmarniesse:

Previously we set a machanism to ack the message after pulling it only if the transport was pubsub (please see the ADR to know the reason).  
Now we ack the message in all cases thanks to an event subscriber.

Issue:
The event subscriber allows to ack the message before handling it. But after handling it, a new ack is requested. I don't figure out how to disable this second ack: it's hardcoded in the Worker https://github.com/symfony/messenger/blob/4.4/Worker.php#L150  
It seems there is no problem to ack the same message twice, but it would be better to ack once.  

- [x] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/src/Akeneo/Tool/Bundle/BatchQueueBundle/EventListener/AckMessageEventListener.php
+++ b/src/Akeneo/Tool/Bundle/BatchQueueBundle/EventListener/AckMessageEventListener.php
@@ -1,0 +1,53 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Akeneo\Tool\Bundle\BatchQueueBundle\EventListener;
+
+use Akeneo\Tool\Component\BatchQueue\Queue\JobExecutionMessageInterface;
+use Psr\Container\ContainerInterface;
+use Psr\Container\NotFoundExceptionInterface;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\Messenger\Event\WorkerMessageReceivedEvent;
+
+/**
+ * Using Google Pub/Sub we should ack the message within the next 10 seconds after pulling it (it can be configured
+ * to 10 minutes maximum). After that the message is re-dispatched. As the job execution can be longer, we
+ * take the decision to ack the message just after pulling it. The aim to this subscriber is to ack all job messages
+ * before the job execution.
+ *
+ * @copyright 2021 Akeneo SAS (http://www.akeneo.com)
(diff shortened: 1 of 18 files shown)


docteurklein:
Indeed, I don't think it's a problem to ACK twice. Unfortunately it's not possible to change the Worker class behavior without rewriting it and the `messenger:consume` command.

src/Akeneo/Tool/Bundle/BatchQueueBundle/EventListener/AckMessageEventListener.php
```diff
@@ -0,0 +1,53 @@
...
+    {
+        return [
+            WorkerMessageReceivedEvent::class => 'ackMessage',
+        ];
+    }
+
+    public function ackMessage(WorkerMessageReceivedEvent $event): void
+    {
+        $envelope = $event->getEnvelope();
+        if (!$envelope->getMessage() instanceof JobExecutionMessageInterface) {
```
docteurklein, line 41: so this is the essence of this PR. Are we sure it's future-proof? 
Could a future message risk to pass this test even tho we don't want to ack?

src/Akeneo/Tool/Bundle/BatchQueueBundle/EventListener/AckMessageEventListener.php
```diff
@@ -0,0 +1,53 @@
...
+            return;
+        }
+
+        try {
+            $receiver = $this->receiverLocator->get($event->getReceiverName());
+        } catch (NotFoundExceptionInterface $e) {
+            return;
```
docteurklein, line 48: do we want to silently fail for something that shouldn't happen?

nmarniesse: you're right I remove the silent exception

src/Akeneo/Tool/Bundle/BatchQueueBundle/EventListener/AckMessageEventListener.php
nmarniesse: Only job messages implement JobExecutionMessageInterface. There is no interest to create a new object message with this interface if it's not a job (and it would be a serious problem of design)

pierallard (approved):

mmetayer (approved):

PR #14607, opened 2021-05-31
PLG-263: auto apply permissions on upper and lower levels

rouliane:

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/src/Akeneo/Pim/Enrichment/Bundle/Controller/Ui/CategoryTreeController.php
+++ b/src/Akeneo/Pim/Enrichment/Bundle/Controller/Ui/CategoryTreeController.php
@@ -404,7 +404,7 @@ class CategoryTreeController extends Controller
     private function formatPermissionField(FormView $formView): array
     {
         return [
-            'value' => $formView->vars['value'],
+            'value' => array_values($formView->vars['value']),
             'fullName' => $formView->vars['full_name'],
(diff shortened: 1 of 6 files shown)


src/Akeneo/Pim/Structure/Bundle/Resources/workspaces/settings-ui/src/components/categories/EditPermissionsForm.tsx
```diff
@@ -30,7 +30,7 @@ const EditPermissionsForm = ({formData, onChangePermissions, onChangeApplyPermis
       <Field label={translate('category.permissions.view.label')}>
         <MultiSelectInput
           readOnly={false}
-          value={formData.permissions.view.value}
+          value={Object.values(formData.permissions.view.value).map(permission => permission)}
```
rouliane, line 33: dirty fix to be able to handle both array or JSON data. I can't explain why yet (but I have a clue what it could be), but we GET data, the `value` is an array, but when we POST data, the `value` returned by the backend is a JSON. I didn't find a way to force it to an array so I handle it on the front side.

olivierpontier (approved):

LaurentPetard (approved):

PR #14613, opened 2021-05-31
API-1632: custom DateTimeImmutable Doctrine Type

pchasle:
Fix an issue when a date is stored with the UTC format, then selected from the database and hydrated into a `\DateTimeImmutable` via doctrine type conversion utility.
Doctrine will use the default server timezone directly to instantiate it, instead of doing the proper conversion:
1. instantiation in UTC...
2. ...then conversion to server timezone.

The fix is to add a custom DateTimeImmutable Doctrine Type to handle properly UTC conversion from the database value.
We already have one for the DateTime, see [UTCDateTimeType](https://github.com/akeneo/pim-community-dev/blob/API-1632/src/Akeneo/Tool/Bundle/StorageUtilsBundle/Doctrine/DBAL/Types/UTCDateTimeType.php)

Doctrine documentation on this topic:
https://www.doctrine-project.org/projects/doctrine-orm/en/2.6/cookbook/working-with-datetime.html#handling-different-time


tseho (approved):

src/Akeneo/Pim/Enrichment/Bundle/Storage/Sql/ElasticsearchProjection/GetElasticsearchProductProjection.php
```diff
@@ -79,7 +79,6 @@ public function fromProductIdentifiers(array $productIdentifiers): array
 
             $values = $this->valuesNormalizer->normalize($row['values'], self::INDEXING_FORMAT_PRODUCT_AND_MODEL_INDEX);
 
-            // use Type::DATETIME and not Type::DATETIME_IMMUTABLE as it's overridden in the PIM (UTCDateTimeType) to handle UTC correctly
```
ahocquard, line 82: why don't you use it now?

```
Type::getType(Type::DATETIME_IMMUTABLE)->convertToPhpValue($row['created_date'], $platform)
```

ahocquard (approved):

src/Akeneo/Pim/Enrichment/Bundle/Storage/Sql/ElasticsearchProjection/GetElasticsearchProductProjection.php
pchasle: Good point :D

PR #14629, opened 2021-06-02
OB-759: Fix missing measurements in cache

ValentinMerlet:

After creating a measurement, the cache contains only one measurement so the `FindAllMeasurementFamilies` query returns only one measurements instead of all. 
I also found some forgotten copy/paste from asset context and fixed them.

### DIFF
--- a/src/Akeneo/Tool/Bundle/MeasureBundle/Persistence/MeasurementFamilyRepository.php
+++ b/src/Akeneo/Tool/Bundle/MeasureBundle/Persistence/MeasurementFamilyRepository.php
@@ -37,7 +37,7 @@ class MeasurementFamilyRepository implements MeasurementFamilyRepositoryInterfac
     public function all(): array
     {
         if (empty($this->allMeasurementFamiliesCache)) {
-            $this->allMeasurementFamiliesCache = $this->loadAssetFamiliesIndexByCodes();
+            $this->allMeasurementFamiliesCache = $this->loadMeasurementFamiliesIndexByCodes();
         }
 
         return array_values($this->allMeasurementFamiliesCache);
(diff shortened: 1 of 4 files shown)


CaillaudPA (approved):

src/Akeneo/Tool/Bundle/MeasureBundle/Persistence/MeasurementFamilyRepository.php
```diff
@@ -84,6 +84,7 @@ public function save(MeasurementFamily $measurementFamily)
             );
         }
 
+        $this->allMeasurementFamiliesCache = $this->all();
         $this->allMeasurementFamiliesCache[$normalizedMeasurementFamily['code']] = $measurementFamily;
```
PierrickMartos, line 88: I'm not sure but line 88 is not needed anymore no? The cache is already loaded with `$this->all();` and since you insert first, should already be here.

kpAkeneo: i think is for secure value of $normalizedMeasurementFamily['code'] but make a control before is maybe better

StevenVAIDIE: In case of creating an measurement, then creating another measurement the second measurement will not be there. At the second call the all will not load all assets ($this->allMeasurementFamiliesCache is not empty) so the second measurement will not be present

StevenVAIDIE (approved):

StevenVAIDIE (commented):
Can you add an integration test please ?

src/Akeneo/Tool/Bundle/MeasureBundle/Persistence/MeasurementFamilyRepository.php
ValentinMerlet: > I'm not sure but line 88 is not needed anymore no? The cache is already loaded with $this->all(); and since you insert first, should already be here.

I think you're right, waiting for Raccoons feedback before making the change

SamirBoulil (approved):
THANK YOU :)

ValentinMumble (approved):

PierrickMartos (approved):

PR #14632, opened 2021-06-03
PIM-9885: Associating a product to itself with a 2-way association returns an error 500

tomglvng:
Add a check that avoids associating a product with itself for two-way association.
Handle this new exception for import, api and ui.
Peek 03-06-2021 10-32


src/Akeneo/Pim/Enrichment/Component/Product/Updater/Setter/AssociationFieldSetter.php
```diff
@@ -132,6 +136,13 @@ private function updateAssociatedProducts(
                     $productIdentifier
                 );
             }
+
+            if ($associationType->isTwoWay()
+                && $owner instanceof Product
+                && $owner->getIdentifier() === $associatedProduct->getIdentifier()) {
+                throw new TwoWayAssociationWithTheSameProductException();
+            }
```
StevenVAIDIE, line 144: Please move this into src/Akeneo/Pim/Enrichment/Bundle/Doctrine/ORM/Updater/TwoWayAssociationUpdater.php.
Maybe do the same thing for product model

StevenVAIDIE (commented):
Functionnal expectation is really "having an error when created an two-way association with same product" ?

By adding the following line in src/Akeneo/Pim/Enrichment/Bundle/Doctrine/ORM/Updater/TwoWayAssociationUpdater.php we can just ignore it
```
            if ($associatedEntity instanceof ProductInterface && $owner->getIdentifier() === $associatedEntity->getIdentifier()) {
                return;
            }
```

If the expectation is throw an error, we should not display the current product in the association grid

tomglvng:
> Functionnal expectation is really "having an error when created an two-way association with same product" ?
> 
> By adding the following line in src/Akeneo/Pim/Enrichment/Bundle/Doctrine/ORM/Updater/TwoWayAssociationUpdater.php we can just ignore it
> 
> ```
>             if ($associatedEntity instanceof ProductInterface && $owner->getIdentifier() === $associatedEntity->getIdentifier()) {
>                 return;
>             }
> ```
> 
> If the expectation is throw an error, we should not display the current product in the association grid

I'll ask to PMs ;)

tomglvng:
> > Functionnal expectation is really "having an error when created an two-way association with same product" ?
> > By adding the following line in src/Akeneo/Pim/Enrichment/Bundle/Doctrine/ORM/Updater/TwoWayAssociationUpdater.php we can just ignore it
> > ```
> >             if ($associatedEntity instanceof ProductInterface && $owner->getIdentifier() === $associatedEntity->getIdentifier()) {
> >                 return;
> >             }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > If the expectation is throw an error, we should not display the current product in the association grid
> 
> I'll ask to PMs ;)

For the moment, we keep the exception system.

src/Akeneo/Pim/Enrichment/Bundle/Controller/ExternalApi/ProductController.php
```diff
@@ -22,6 +22,7 @@
 use Akeneo\Pim\Enrichment\Component\Product\Event\ProductDomainErrorEvent;
 use Akeneo\Pim\Enrichment\Component\Product\Exception\InvalidArgumentException as ProductInvalidArgumentException;
 use Akeneo\Pim\Enrichment\Component\Product\Exception\ObjectNotFoundException;
+use Akeneo\Pim\Enrichment\Component\Product\Exception\TwoWayAssociationWithTheSameProductException;
```
rfauglas, line 25: Do you need a dedicated exception? Could a  ProductDomainErrorEvent with proper translation key be used?

src/Akeneo/Pim/Enrichment/Bundle/Controller/InternalApi/ProductController.php
```diff
@@ -266,7 +267,14 @@ public function postAction(Request $request, $id)
...
             throw new BadRequestHttpException();
         }
-        $this->updateProduct($product, $data);
+        try {
+            $this->updateProduct($product, $data);
+        } catch (TwoWayAssociationWithTheSameProductException $e) {
+            return new JsonResponse([
+                'message' => $e->getMessage(),
+                'global' => true],
+                Response::HTTP_UNPROCESSABLE_ENTITY);
```
rfauglas, line 276: Can't we use an exception that will be mapped automatically in a controller filter?

src/Akeneo/Pim/Enrichment/Bundle/Doctrine/ORM/Updater/TwoWayAssociationUpdater.php
```diff
@@ -40,6 +41,12 @@ public function createInversedAssociation(
         string $associationTypeCode,
         EntityWithAssociationsInterface $associatedEntity
     ): void {
+        if ($owner instanceof ProductInterface
+            && $associatedEntity instanceof ProductInterface
+            && $owner->getIdentifier() === $associatedEntity->getIdentifier()) {
+            throw new TwoWayAssociationWithTheSameProductException();
```
rfauglas, line 47: Can't you use an InvalidPropertyException? Maybe with a new constant?

rfauglas (approved):
Maybe some additional unit tests for the new logic?
And if InvalidPropertyException/InvalidAttributeException can't be used, then you can merge ;-)

LaurentPetard (approved):

src/Akeneo/Pim/Enrichment/Bundle/Controller/InternalApi/ProductController.php
tomglvng: We need to have such a format in the front.

src/Akeneo/Pim/Enrichment/Bundle/Doctrine/ORM/Updater/TwoWayAssociationUpdater.php
tomglvng: See https://github.com/akeneo/pim-community-dev/blob/a54e0a9607868042ea720196d3a03c9c54a4a379/src/Akeneo/Pim/Enrichment/Bundle/Controller/ExternalApi/ProductController.php#L556 : we already catch PropertyException and throw the suitable DocumentedHttpException. In our case, the DocumentedHttpException is different (message and doc ref): that's why I used a dedicated Exception.

src/Akeneo/Pim/Enrichment/Bundle/Controller/ExternalApi/ProductController.php
tomglvng: See below for the reason why I used a dedicated Exception. About translation key, I don't think we need (see for instance https://github.com/akeneo/pim-community-dev/blob/a54e0a9607868042ea720196d3a03c9c54a4a379/src/Akeneo/Pim/Enrichment/Bundle/Controller/ExternalApi/ProductController.php#L556 : no translation key)

PR #14697, opened 2021-06-10
PIM-9902: fix specific local attribute export in model and product model

rfauglas:

- [x] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/src/Akeneo/Pim/Enrichment/Component/Product/ValuesFiller/FillMissingProductModelValues.php
+++ b/src/Akeneo/Pim/Enrichment/Component/Product/ValuesFiller/FillMissingProductModelValues.php
@@ -123,8 +123,14 @@ final class FillMissingProductModelValues implements FillMissingValuesInterface
                     $nullValues[$attribute->getCode()][$channel->getCode()]['<all_locales>'] = $nullValue;
                 }
             } elseif (!$attribute->isScopable() && $attribute->isLocalizable()) {
-                foreach ($this->getLocales() as $locale) {
-                    $nullValues[$attribute->getCode()]['<all_channels>'][$locale->getCode()] = $nullValue;
+                if ($attribute->isLocaleSpecific()) {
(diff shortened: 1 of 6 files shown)


src/Akeneo/Pim/Enrichment/Component/Product/ValuesFiller/FillMissingProductModelValues.php
```diff
@@ -123,8 +123,14 @@ function (AttributeInterface $attribute): bool {
                     $nullValues[$attribute->getCode()][$channel->getCode()]['<all_locales>'] = $nullValue;
                 }
             } elseif (!$attribute->isScopable() && $attribute->isLocalizable()) {
-                foreach ($this->getLocales() as $locale) {
-                    $nullValues[$attribute->getCode()]['<all_channels>'][$locale->getCode()] = $nullValue;
+                if ($attribute->isLocaleSpecific()) {
```
StevenVAIDIE, line 126: This filter should not be done here because:
- This service is in charge to create the matrix of exported header with value equal to null and merge with product values. 
- An attribute locale specific can be edited after creation. By doing here the product value will not be filtered at quick export.
For example, if Julia create an attribute tax localisable and local specific to Canadian and French. When Julia enrich this attribute in French and Canadian into an product. And set the attribute locale specific only to Canadian. Then the column tax-fr_FR should not be exported. 

Here it will be exported this is why we do it after in the ProductExportProcessor: https://github.com/akeneo/pim-community-dev/blob/master/src/Akeneo/Pim/Enrichment/Component/Product/Connector/Processor/Normalization/ProductProcessor.php#L74

StevenVAIDIE (changes requested):

StevenVAIDIE (approved):
Seen IRL => it will be fixed by maintenance if problem occured.

Do not forget to add CHANGELOG

anaisbaunelemaire (approved):

PR #14720, opened 2021-06-14
API-1659 Setup RequestContext with missing host

RiffVentura:
Added  host configuration to the RequestContext of the router 

Routing  defaults  to localhost as host when generating a route within the consumer. To mitigate this issue, a setup for the RequestContext was added
Setup required a creation of new bundle for configuration of this nature

- [x] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc


src/Akeneo/Platform/Bundle/FrameworkBundle/Service/PimUrl.php
```diff
@@ -0,0 +1,24 @@
...
+
+declare(strict_types=1);
+
+namespace Akeneo\Platform\Bundle\FrameworkBundle\Service;
+
+/**
+ * @copyright 2021 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+class PimUrl
```
pchasle, line 11: I think that it will be better to have the service expose the host, scheme, port directly.
Easier to test the parsing of the URL and easier to use.

tseho: Additionally to the url ?
I don't know if that makes sense, most of the time, when you use it, you want the url, you don't want to check if there is a port in it for example.

To me, the signature of the service make sense as it is.
In the `boot`, yes, we have to explode it because it's in the `boot` method that we bridge between the full url and the router context who want the individual parts.

tseho:
Don't forget to open a PR in EE too, you need to declares it in EE bundles.php too

src/Akeneo/Platform/Bundle/FrameworkBundle/Service/PimUrl.php
pchasle: It's better than having to explode the url each time to get exactly what we need.
We can expose the whole url too or let the user of this service reconstruct the url by itself

pchasle (approved):

tomglvng (approved):

willy-ahva (approved):

src/Akeneo/Platform/Bundle/FrameworkBundle/Service/PimUrl.php
tseho: > It's better than having to explode the url each time to get exactly what we need.
There is no such thing. AFAIK, this is the only place in the PIM where the AKENEO_PIM_URL is exploded.
All the other places I've seen, even in the Events API, the url is used without manipulation.
Since we are thinking on the long term, in a soon future, this environment variable will disappear and every service will have to call a service like this one to access the URL of the PIM.
I don't want to force all the other use cases to glue the url because there was one place where we wanted the parts.

config/bundles.php
```diff
@@ -53,4 +53,5 @@
     Symfony\Bundle\WebProfilerBundle\WebProfilerBundle::class => ['dev' => true, 'test' => true, 'behat' => true],
     Symfony\Bundle\WebServerBundle\WebServerBundle::class => ['dev' => true, 'test' => true, 'behat' => true],
     Akeneo\FreeTrial\Infrastructure\Symfony\AkeneoFreeTrialBundle::class =>  ['all' => true],
+    Akeneo\Platform\Bundle\FrameworkBundle\PimFrameworkBundle::class => ['all' => true]
```
tseho, line 56: You broke growth edition by leaving-out the trailing `,`

When merging the bundles.php in GE, it look like this:
```
Akeneo\FreeTrial\Infrastructure\Symfony\AkeneoFreeTrialBundle::class =>  ['all' => true],
Akeneo\Platform\Bundle\FrameworkBundle\PimFrameworkBundle::class => ['all' => true]
Akeneo\Platform\Bundle\MonitoringBundle\AkeneoMonitoringBundle::class => ['all' => true],
```
Which is not valid php, breaking the build of GE.

tseho (changes requested):

config/bundles.php
RiffVentura: facepalm

tseho: > facepalm

No, no 😅
It's not your fault, 
1) we should have better tools to enforce code style
2) we should build GE differently

tseho (approved):

PR #14731, opened 2021-06-15
CXP-342-343: update attribute localizable and not localizable error message

willy-ahva:

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc


src/Akeneo/Pim/Enrichment/Component/Error/DocumentationBuilder/LocalizableAttribute.php
```diff
@@ -0,0 +1,60 @@
...
+
+use Akeneo\Pim\Enrichment\Component\Error\Documentation\Documentation;
+use Akeneo\Pim\Enrichment\Component\Error\Documentation\DocumentationCollection;
+use Akeneo\Pim\Enrichment\Component\Error\Documentation\HrefMessageParameter;
+use Akeneo\Pim\Enrichment\Component\Error\Documentation\RouteMessageParameter;
+use Akeneo\Pim\Enrichment\Component\Error\DocumentationBuilderInterface;
+use Akeneo\Pim\Enrichment\Component\Product\Exception\LocalizableAttributeException;
+use Akeneo\Pim\Enrichment\Component\Product\Exception\NotLocalizableAttributeException;
+
+class LocalizableAttribute implements DocumentationBuilderInterface
```
willy-ahva, line 15: As this is for localizable AND not localizable attribute, someone may have a better name?

tomglvng: First, I was thinking `LocalizableAndNonLocalizableAttribute`, but why don't we keep it simple ? `Attribute` ?

tomglvng (approved):

src/Akeneo/Pim/Enrichment/Component/Error/DocumentationBuilder/LocalizableAttribute.php
willy-ahva: Because it's about localization, and we have a lot of exceptions about attributes, we need to specify something. I think I'll let it like this.

tseho (approved):
I strongly advise against the temptation to refactor code around your PR (like removing phpdoc & type-hinting properties).
It make the PR slightly harder to read (why are those lines changed ? Does the new type is correct ? Was the type defined in the phpdoc the correct one ?)
And it can also make them harder to merge due to more surface for conflicts.

This is a non-blocking comment, GTM.

RiffVentura (approved):

pchasle (approved):

willy-ahva:
> I strongly advise against the temptation to refactor code around your PR (like removing phpdoc & type-hinting properties).
> It make the PR slightly harder to read (why are those lines changed ? Does the new type is correct ? Was the type defined in the phpdoc the correct one ?)
> And it can also make them harder to merge due to more surface for conflicts.
> 
> This is a non-blocking comment, GTM.
> 
> EDIT: I'm not saying that we shouldn't refactor, I think we should do those in separate & dedicated PRs. I don't mind if you add a dedicated ticket in the sprint. If an error is introduced, the git history will tell us why this line was changed and it can be reverted without touching the feature.

Could be done in a dedicated commit next time if you're ok? Making another PR to clean code is a heavy process.

PR #14750, opened 2021-06-17
PIM-9914: Fix backslash incorrectly considered as an escape character in exports

pierrejolly:

* The backslash \ should not be considered as a an escape character where today the PIM says it is an escape character
* The quote ” should always be doubled when it is set as a value (which is today’s behavior except when there are backslashes)

- [x] Tests

### DIFF
--- a/src/Akeneo/Tool/Component/Connector/Writer/File/FlatItemBufferFlusher.php
+++ b/src/Akeneo/Tool/Component/Connector/Writer/File/FlatItemBufferFlusher.php
@@ -4,8 +4,8 @@ namespace Akeneo\Tool\Component\Connector\Writer\File;
 
 use Akeneo\Tool\Component\Batch\Model\StepExecution;
 use Akeneo\Tool\Component\Batch\Step\StepExecutionAwareInterface;
+use Akeneo\Tool\Component\Connector\Writer\WriterFactory;
 use Box\Spout\Common\Exception\UnsupportedTypeException;
-use Box\Spout\Writer\WriterFactory;
 use Box\Spout\Writer\WriterInterface;
 
 /**
--- a/src/Akeneo/Tool/Component/Connector/Writer/GlobalFunctionsHelper.php
+++ b/src/Akeneo/Tool/Component/Connector/Writer/GlobalFunctionsHelper.php
@@ -7,9 +7,7 @@ namespace Akeneo\Tool\Component\Connector\Writer;
 use Box\Spout\Common\Helper\GlobalFunctionsHelper as SpoutGlobalFunctionsHelper;
 
 /**
- * @author    Pierre Jolly <pierre.jolly@akeneo.com>
- * @copyright 2021 Akeneo SAS (http://www.akeneo.com)
- * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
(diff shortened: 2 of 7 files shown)


src/Akeneo/Tool/Component/Connector/Writer/WriterFactory.php
```diff
@@ -0,0 +1,25 @@
...
+
+
+use Box\Spout\Writer\WriterFactory as SpoutWriterFactory;
+use Box\Spout\Writer\WriterInterface;
+
+/**
+ * @author    Pierre Jolly <pierre.jolly@akeneo.com>
+ * @copyright 2021 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
```
nmarniesse, line 15: Can you add a comment and create a TIP ticket? Something like:

```suggestion
  * TIP-xxxx: update Spout library on master in 3.x to avoid the override of the SpoutWriterFactory class
  */
```

nmarniesse (approved):

src/Akeneo/Tool/Component/Connector/Writer/WriterFactory.php
nmarniesse: If you're working on it, no need I guess. Just maybe a `TODO pullup: skip this file because...`

src/Akeneo/Tool/Component/Connector/Writer/GlobalFunctionsHelper.php
```diff
@@ -0,0 +1,37 @@
...
+declare(strict_types=1);
+
+namespace Akeneo\Tool\Component\Connector\Writer;
+
+use Box\Spout\Common\Helper\GlobalFunctionsHelper as SpoutGlobalFunctionsHelper;
+
+/**
+ * @author    Pierre Jolly <pierre.jolly@akeneo.com>
+ * @copyright 2021 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
```
mmetayer, line 12: You're not really the author :) 
I would put a link to the original file (https://github.com/box/spout/blob/v3.0.0/src/Spout/Common/Helper/GlobalFunctionsHelper.php), and explain why we needed to extend the class instead of bumping the version requirement. Also, I would drop the license line (as spout is under Apache license). And finally, please add a todo merge master/6.0: drop this class and bump box/spout to v3

mmetayer (approved):

src/Akeneo/Tool/Component/Connector/Writer/File/FlatItemBufferFlusher.php
```diff
@@ -4,8 +4,8 @@
 
 use Akeneo\Tool\Component\Batch\Model\StepExecution;
 use Akeneo\Tool\Component\Batch\Step\StepExecutionAwareInterface;
+use Akeneo\Tool\Component\Connector\Writer\WriterFactory;
```
mmetayer, line 7: you can also fix the import in `Pim\Behat\Context\Domain\Collect\ImportProfilesContext`

PR #14809, opened 2021-06-22
CXP-342: Update error messages

willy-ahva:

screenshot-localhost_8080-2021 06 23-11_54_52

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc


pchasle (approved):

RiffVentura:
Shouldn't there be a change within translations files as well ?

pchasle:
> Shouldn't there be a change within translations files as well ?

If I'm not mistaken we don't need to translate those errors has the UI will not allow to trigger them in the first place.
And the Rest API errors are not translated.

RiffVentura (approved):

PR #14811, opened 2021-06-23
CXP-359 & CXP-367: Update error messages - Range

tomglvng:
EE --> https://github.com/akeneo/pim-enterprise-dev/pull/11126
Messages : 
 - The %attribute% attribute requires an equal or greater than %min_value% value.
 - The %attribute% attribute requires an equal or lesser than %max_value% value.

PR #14814, opened 2021-06-23
CXP-367: Update error message - The %attribute% attribute requires an equal or lower than %min_value% value.

tomglvng:


tomglvng:
See https://github.com/akeneo/pim-community-dev/pull/14811

PR #10867, opened 2019-10-09
TIP-1314: Batch API events

nmarniesse:
**Description (for Contributor and Core Developer)**

Add a subscriber to batch POST_SAVE events for product/product model in the update list API

**Definition Of Done (for Core Developer only)**

| Q                                 | A
| --------------------------------- | ---
| Added Specs                       | OK
| Added legacy Behats               | -
| Added acceptance tests            | -
| Added integration tests           | OK
| Changelog updated                 | OK
| Review and 2 GTM                  | Todo
| Micro Demo to the PO (Story only) | -
| Migration script                  | -
| Tech Doc                          | -

`Todo`: Pending / Work in progress
`OK`: Done / Validated
`-`: Not needed

PR #14815, opened 2021-06-23
PIM-9929: Fix performance issue on API attributes options partial update list

rouliane:

The goal is to unplug the synchronous DQI evaluation of options during an API mass update to increase the performances. The options will be evaluated with a dedicated asynchronous DQI job.

Same PR as the attributes (https://github.com/akeneo/pim-community-dev/pull/13687, see description), but for the options. This is also the same behavior as the products (https://github.com/akeneo/pim-community-dev/pull/10867).

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc


olivierpontier (approved):
:+1:

src/Akeneo/Pim/Structure/Bundle/Controller/ExternalApi/AttributeOptionController.php
```diff
@@ -289,7 +284,19 @@ public function partialUpdateAction(Request $request, $attributeCode, $code)
...
         $resource = $request->getContent(true);
-        $response = $this->partialUpdateStreamResource->streamResponse($resource, ['attributeCode' => $attributeCode]);
+
+        $this->apiAggregatorForAttributeOptionPostSave->activate();
+
+        $response = $this->partialUpdateStreamResource->streamResponse($resource, ['attributeCode' => $attributeCode], function () {
+            try {
+                $this->apiAggregatorForAttributeOptionPostSave->dispatchAllEvents();
+            } catch (\Throwable $exception) {
+                $this->logger->critical('An exception has been thrown in the post-save events', [
```
BitOne, line 294: Do we really want to continue? If we want to continue, then it's a warning, not a critical.
If we don't want to continue, then shouldn't catch the Exception and let Symfony do its work.

BitOne (changes requested):

src/Akeneo/Pim/Structure/Bundle/Controller/ExternalApi/AttributeOptionController.php
rouliane: I guess we want to continue in this case because for example, an error in ONE option shouldn't send an error for the whole API call, don't you think ? So yes I agree it should indeed be a warning instead of a critical.

By the way it's the exact same behavior as the product (or attributes) you validated last year : https://github.com/akeneo/pim-community-dev/pull/11972/files#diff-796532e4e92366b2c5f9e7843f71c11cb8781681f80fd6db0adecae82c7485c5R433 . Is something different this time ?

BitOne: Yeap, it should have been a warning too in the code you linked....

rouliane: I'll change it to a warning in both cases then :+1:

rouliane: Fixed, in 3 controllers (attribute, attribute options and product)

LaurentPetard (approved):

src/Akeneo/Pim/Structure/Bundle/Controller/ExternalApi/AttributeOptionController.php
BitOne: Super, thanks Julian!

BitOne (approved):

PR #14818, opened 2021-06-24
Symfony codestyle

mmetayer:

**Note:** ~The 2 last commits are simple examples of what the new rules are, and are not meant to be merged in this PR~ the commits were removed

#### Description
This PR is an attempt to homogenize the code style in the pim's codebase. The goal is to apply the `@Symfony` php-cs-fixer ruleset on updated / new files, but not change the CI's config (with more permissive rules) in order not to have a failing build if some files were not updated according to the new rules. This will allow migrating the codebase progressively, and avoid a big PR with all the code base updated at once, which would cause a lot of file conflicts, and tedious PR rebases.

In order to do so, I added a new makefile target, which can be used at will or in the following pre-commit hook:

pre-commit hook:
```bash
#!/usr/bin/env bash
  
FILES=`git status --porcelain | grep -P '^[AM] .+\.php$' | cut -c 4- | tr '\n' ' '`

if [[ -z "$FILES" ]]
then
    exit 0
fi

PIM_CONTEXT=dev make lint-fix-back O=${FILES}
res=$?

if [[ $res -ne 0 ]]
then
    exit $res
fi

git add $FILES
exit 0
```
To install the hook, just copy/paste it into the `.git/hooks/pre-commit` file in your project, and give it execution rights (`chmod +x .git/hooks/pre-commit`). The hook will fix the codestyle for every added/updated PHP file tracked in the git index, and will display a diff of the fixes.

If you don't want to use the hook for a specific commit, you can do so by passing the `-n` (aka `--no-verify`) option to your `git commit` command.

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/.php_cs_symfony.dist
+++ b/.php_cs_symfony.dist
@@ -1,0 +1,21 @@
+<?php
+
+declare(strict_types=1);
+
+$finder = PhpCsFixer\Finder::create()
+    ->name('*.php')
+    ->notName('*Spec.php')
+    ->in(__DIR__ . '/tests')
+    ->in(__DIR__ . '/src');
+
+$rules = [
+    '@Symfony' => true,
+];
+
+$config = new PhpCsFixer\Config();
+
+return $config
+    ->setRules($rules)
+    ->setFinder($finder)
+    ->setCacheFile('var/.php_cs_symfony.cache')
+    ->setRiskyAllowed(true);
--- a/make-file/dev.mk
+++ b/make-file/dev.mk
@@ -49,3 +49,6 @@ xdebug-on:
 cypress-interactive:
 	docker-compose -f docker-compose-cypress.yml run --rm -u 1000:1000 -e DISPLAY -v /tmp/.X11-unix:/tmp/.X11-unix --entrypoint cypress cypress open --project .
 
+.PHONY: lint-fix-back
+lint-fix-back: #Doc: run php-cs-fixer with symfony codestyle
+	${PHP_RUN} vendor/bin/php-cs-fixer fix --config=.php_cs_symfony.dist --diff --path-mode=intersection ${O}


.php_cs_symfony.dist
```diff
@@ -0,0 +1,19 @@
...
+    '@Symfony' => true,
+];
+
+$config = new PhpCsFixer\Config();
+
+return $config
+    ->setRules($rules)
+    ->setFinder($finder)
+    ->setCacheFile('var/.php_cs_symfony.cache')
+    ->setRiskyAllowed(true);
```
mmetayer, line 19: to discuss. This allows applying some risky rules which _might_ change the behavior

make-file/dev.mk
```diff
@@ -49,3 +49,6 @@ xdebug-on:
 cypress-interactive:
 	docker-compose -f docker-compose-cypress.yml run --rm -u 1000:1000 -e DISPLAY -v /tmp/.X11-unix:/tmp/.X11-unix --entrypoint cypress cypress open --project .
 
+.PHONY: lint-fix-back
```
mmetayer, line 52: add a description?

pchasle (commented):
Thanks Mathias :v: 

Would it be possible to add the pre-commit hook and the instruction inside the internal_doc for easy retrieval?

Also, small question on the example commits:
There is a lot of multi-line code samples transformed to be a single long line and that make the code a little less readable.
Do you now from what rule this come from?

pchasle:
> Also, small question on the example commits:
> There is a lot of multi-line code samples transformed to be a single long line and that make the code a little less readable.
> Do you now from what rule this come from?

I found `method_argument_space` https://cs.symfony.com/doc/rules/function_notation/method_argument_space.html
Defined as `['on_multiline' => 'ignore']` for `@Symfony`, so it does seem strange that the lines where updated in your example commit.

mmetayer:
@pchasle 
> Would it be possible to add the pre-commit hook and the instruction inside the internal_doc for easy retrieval?

good idea, will do :ok_hand: 

> Defined as ['on_multiline' => 'ignore'] for @Symfony, so it does seem strange that the lines where updated in your example commit.

yep, it bothered me a bit as well; I'll try to dig a little on this specific issue before merging

mmetayer:
@pchasle 
multi-line transformed into single line are due to 2 rules which are part of the `@Symfony` ruleset:
- [single_line_throw](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/3.0/doc/rules/function_notation/single_line_throw.rst) which implies that every `throw` statement should be on a single line
- [class_definition: single_line](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/3.0/doc/rules/class_notation/class_definition.rst#single_line) which implies that the class definition should be on a songle line, regardless of the count of classes/interfaces it extrends/implements

I personnally am not a big fan of these 2 rules as I find it makes the code less readable, but we agreed on using the Symfony ruleset in the backend guild; besides this ruleset is already used in [shared catalogs](https://github.com/akeneo/exboarder/blob/master/back/config/tests/php-cs-fixer.dist) and the goal behind this PR is to homgeneize the codestyle in all Akeneo products. This can be further discussed/debated, but I'll let it as is for now

rfauglas:
Did you perform so rebase on this PR? Can you tell us how painful it is? This will give us a hint on how impactful this PR might be for PR and rebase...

mmetayer:
@rfauglas I actually did not rebase the 2 last commits (I only re-ran the command after rebasing the first 2 commits), as they are not meant to be merged in this PR, and are just here to provide an overview of the new code style, so I can't really answer your question
Anyway, the hook only handles the files added/updated in a PR, so I guess a rebase would not be too painful, however pull ups from 5.0 might have a few more conflicts than they used to (but hopefully quite easy to resolve most of the time).

willy-ahva (approved):

nmarniesse (approved):
About the pre commit hook, I think it would be great to version this hook in the repo.  
In a second step maybe add a `install_pre_commit_hook.sh` script would be great too (or a make target but maybe overkill)

make-file/dev.mk
greg-akeneo: yup, that would allow the target to show up when `make help`

PR #14852, opened 2021-06-28
CXP-766: Specs for the Apps endpoint

tseho:

If you want to check live the spec, go https://editor.swagger.io/, `File` -> `Import url` and paste `https://github.com/akeneo/pim-community-dev/blob/CXP-766/src/Akeneo/Connectivity/Connection/docs/marketplace/api.yaml`

Screenshot_2021-07-12_11-51-38

Screenshot_2021-07-12_12-04-47

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/src/Akeneo/Connectivity/Connection/docs/marketplace/api.yaml
+++ b/src/Akeneo/Connectivity/Connection/docs/marketplace/api.yaml
@@ -1,7 +1,7 @@
 swagger: "2.0"
 info:
     description: ""
-    version: "1.0.2"
+    version: "1.0.3"
     title: "Marketplace"
 host: "marketplace.akeneo.com"
 basePath: "/api/1.0"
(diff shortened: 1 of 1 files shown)


src/Akeneo/Connectivity/Connection/docs/marketplace/api.yaml
```diff
@@ -81,6 +81,79 @@ paths:
...
+                  enum:
+                      - serenity
+                      - growth-edition
+                      - enterprise-edition
+                      - community-edition
+                - name: version
+                  in: query
+                  type: string
+                  enum:
+                      - "5.0"
```
RiffVentura, line 107: Shouldn't it be 5 or higher ?

tseho: what do you mean ?

tseho: For the context, in the marketplace, PIM versions are created manually in the backoffice (it's a CMS) and are then available dynamically in the filters.

The problem with the current filter used by the marketplace frontend, it's that they use the version id to filter, not the version-version. 
See https://marketplace.akeneo.com/search?edition=all&version=40&api_use=all&sort=date

We want predictable codes to use the filter, not an unknown id.
So, with this spec, we specify, for the known versions, that we want to use the version number, not this unknown id.

Yes, when 6.0 will be released, it will be available in the filters, but not documented in the spec.

pchasle (approved):

RiffVentura (approved):

src/Akeneo/Connectivity/Connection/docs/marketplace/api.yaml
RiffVentura: What i meant : Would there be a an app for the previous version of PIM since the App concept  do not exist in previous  versions (prior to 5.0) ? but  I think I understood.

tseho: oh, I get it.
We cannot have Apps that are compatible with 3.2 since the concept of Apps did not exists at that time.
So, you can use `version=3.2` but it will never return any result.
Even then, only 6.0 will be pertinent.
This is a good point and I think we can keep it like this.
We can update the spec accordingly when 6.0 is released.

romaindak (approved):

PR #14857, opened 2021-06-29
CPM-193: Add Boolean and Select for table input

pierallard:


front-packages/akeneo-design-system/src/components/Input/TableInput/TableInputBoolean/TableInputNumber.unit.tsx
```diff
@@ -0,0 +1 @@
+// TODO
```
nmarniesse, line 1: Rename in TableInputBoolean.unit.tsx and todo

PR #14882, opened 2021-07-01
PLG-278: Activate user when he logs in for the first time in the Free Trial

rouliane:

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/src/Akeneo/FreeTrial/back/Infrastructure/MySql/MySqlInvitedUserRepository.php
+++ b/src/Akeneo/FreeTrial/back/Infrastructure/MySql/MySqlInvitedUserRepository.php
@@ -24,7 +24,8 @@ final class MySqlInvitedUserRepository implements InvitedUserRepository
     public function save(InvitedUser $invitedUser): void
     {
         $query = <<<SQL
-INSERT INTO akeneo_free_trial_invited_user (email, status, created_at) VALUES (:email, :status, NOW());
+INSERT INTO akeneo_free_trial_invited_user (email, status, created_at) VALUES (:email, :status, NOW())
+ON DUPLICATE KEY UPDATE status = :status;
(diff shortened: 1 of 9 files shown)


LaurentPetard (approved):

src/Akeneo/FreeTrial/back/Infrastructure/MySql/MySqlInvitedUserRepository.php
```diff
@@ -24,7 +24,8 @@ public function __construct(Connection $dbalConnection)
     public function save(InvitedUser $invitedUser): void
     {
         $query = <<<SQL
-INSERT INTO akeneo_free_trial_invited_user (email, status, created_at) VALUES (:email, :status, NOW());
+INSERT INTO akeneo_free_trial_invited_user (email, status, created_at) VALUES (:email, :status, NOW())
+ON DUPLICATE KEY UPDATE status = :status;
```
olivierpontier, line 28: We should talk about this use case. IMHO there are several ways to handle it. In that case, in the ideal scenario, the API should return an error saying that the invitation is already sent before saving it.

olivierpontier (approved):
GTM after talking about the save of invited users

src/Akeneo/FreeTrial/back/Infrastructure/MySql/MySqlInvitedUserRepository.php
rouliane: I think you misunderstood : this is only a trick to handle both create and update in one query (create = invite user via the form, update = update its status in the db). Let's talk about it tomorrow!

BryanS29-akeneo (approved):

PR #14890, opened 2021-07-02
CXP-806: symfony/cache

tseho:

Explicitly requires `symfony/cache`, add a cache pool named `cache.data` alongside the default `cache.app` & `cache.system` with PDO as an adapter.

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/config/packages/framework.yml
+++ b/config/packages/framework.yml
@@ -24,6 +24,7 @@ framework:
             frontend:
                 version_strategy: pim_enrich.version_strategy
     cache:
+        app: cache.adapter.pdo
         pools:
             name:
                 adapter: cache.adapter.apcu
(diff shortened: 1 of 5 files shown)


romaindelouard (approved):

romaindak (approved):

config/packages/framework.yml
```diff
@@ -27,3 +27,6 @@ framework:
         pools:
             name:
                 adapter: cache.adapter.apcu
+            cache.data:
+                adapter: cache.adapter.pdo
+                public: true
```
pchasle, line 32: We should specify the configuration for the db_table (default: cache_items) and probably add the migration to create the table too (since if we don't it will be created at runtime and this might be an issue for the prod environnement).

https://github.com/symfony/symfony/blob/5.3/src/Symfony/Component/Cache/Adapter/PdoAdapter.php#L53

PR #14891, opened 2021-07-02
Fix coupling between Connectivity and UserManagement

pchasle:
The goal of this PR is to remove the dependency on the *UserManagement* bounded context from the "EventDataBuilder" interfaces used by the *Enrichment* bounded context to build its own events data sent by the Events API.

From the
```php
namespace Akeneo\Platform\Component\Webhook;

interface EventDataBuilderInterface
```

Before, with the coupling to UserManagement
```php
public function build(BulkEventInterface $event, Akeneo\UserManagement\Component\Model\UserInterface $user): EventDataCollection;
```

After, the interface is not coupled to UserManagement but use a simple DTO with only its required data (user id & username)
```php
public function build(BulkEventInterface $event, Context $context): EventDataCollection;
```

### DIFF
--- a/src/Akeneo/Platform/Component/Webhook/Context.php
+++ b/src/Akeneo/Platform/Component/Webhook/Context.php
@@ -1,0 +1,31 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Akeneo\Platform\Component\Webhook;
+
+/**
+ * @copyright 2021 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+final class Context
+{
+    private string $username;
+    private int $userId;
+
+    public function __construct(string $username, int $userId)
+    {
+        $this->username = $username;
+        $this->userId = $userId;
+    }
+
+    public function getUsername(): string
+    {
+        return $this->username;
+    }
+
+    public function getUserId(): int
+    {
+        return $this->userId;
+    }
+}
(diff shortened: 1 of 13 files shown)


src/Akeneo/Platform/Component/Webhook/Context.php
```diff
@@ -0,0 +1,31 @@
...
+
+namespace Akeneo\Platform\Component\Webhook;
+
+/**
+ * @copyright 2021 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+class Context
+{
+    private string $username;
```
willy-ahva, line 13: you could use public properties instead of getters for a DTO

src/Akeneo/Platform/Component/Webhook/Context.php
```diff
@@ -0,0 +1,31 @@
...
+
+declare(strict_types=1);
+
+namespace Akeneo\Platform\Component\Webhook;
+
+/**
+ * @copyright 2021 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+class Context
```
willy-ahva, line 11: what about put it as final?

willy-ahva (approved):

src/Akeneo/Platform/Component/Webhook/Context.php
pchasle: I agree with you (even if we are missing some kind of "read only" type in PHP). But this will not be consistent with the rest of the PIM DTO :/

tseho (approved):

PR #14895, opened 2021-07-02
CXP-788 : Fetch data from marketplace API (mock implementation)

tomglvng:

- [x] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc


src/Akeneo/Connectivity/Connection/back/Domain/Marketplace/GetAllExtensionsQueryInterface.php
```diff
@@ -0,0 +1,16 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Akeneo\Connectivity\Connection\Domain\Marketplace;
```
pchasle, line 5: What do you think about having:

Common DTO for an extension, maybe it's fine for now, or maybe it should be specific to and live with the query.
```php
namespace Akeneo\Connectivity\Connection\Domain\Marketplace\DTO;

class Extension
{
}
```

Folder `GetAllExtensionsQuery` that host the query interface and the response object:

```php
namespace Akeneo\Connectivity\Connection\Domain\Marketplace\Query\GetAllExtensions;

use Akeneo\Connectivity\Connection\Domain\Marketplace\DTO\Extension;

class ExtensionList
{
    /** @var array<Extension> */
    private array $extensions;
}
```
```php
namespace Akeneo\Connectivity\Connection\Domain\Marketplace\Query\GetAllExtensions;

interface GetAllExtensionsQueryInterface
{
    public function execute(): ExtensionList;
}
```

src/Akeneo/Connectivity/Connection/back/Infrastructure/Symfony/Resources/config/marketplace.yml
```diff
@@ -1,16 +1,25 @@
...
+      - '@akeneo_connectivity.connection.marketplace.query.get_user_profile'
 
-    akeneo_connectivity.connection.marketplace.query.get_user_profile:
-        class: Akeneo\Connectivity\Connection\Infrastructure\Persistence\Dbal\Query\DbalGetUserProfileQuery
-        arguments:
-            - '@database_connection'
+  akeneo_connectivity.connection.marketplace.query.get_user_profile:
+    class: Akeneo\Connectivity\Connection\Infrastructure\Persistence\Dbal\Query\DbalGetUserProfileQuery
+    arguments:
+      - '@database_connection'
```
pchasle, line 16: Please, do not change the indentation. It's 4 spaces for yaml files.

src/Akeneo/Connectivity/Connection/back/Domain/Marketplace/DTO/Extension.php
```diff
@@ -0,0 +1,172 @@
+<?php
+declare(strict_types=1);
+
+namespace Akeneo\Connectivity\Connection\Domain\Marketplace\DTO;
```
tseho, line 4: In the present case, `Extension` is not a DTO, it's our model.
IHMO, It should be in `Domain\Marketplace\Model`, like the other Connectivity models.

src/Akeneo/Connectivity/Connection/back/Domain/Marketplace/DTO/Extension.php
```diff
@@ -0,0 +1,172 @@
...
+
+use Ramsey\Uuid\Uuid;
+use Ramsey\Uuid\UuidInterface;
```
tseho, line 7: What's the value of using Ramsey\Uuid here ?

src/Akeneo/Connectivity/Connection/back/Domain/Marketplace/DTO/Extension.php
```diff
@@ -0,0 +1,172 @@
...
+            Uuid::fromString($values['id']),
+            $values['name'],
+            $values['logo'],
+            $values['author'],
+            $values['partner'],
+            $values['description'],
+            $values['url'],
+            $values['categories'],
+            $values['certified']
+        );
```
tseho, line 94: You want a neat trick to avoid declaring twice your properties, both in named constructor `create(...)` and in the default constructor ?

Since you are inside the class, you can skip the constructor and do it like this:
```
class Extension
{
    private string $name;
    
    public static function create(array $values): self
    {
        $self = new self();
        $self->name = $values['name'];
        
        return $self;
    }
```

src/Akeneo/Connectivity/Connection/back/Domain/Marketplace/GetAllExtensionsQueryInterface.php
tseho: This is an interesting proposition but I'm not seeing the immediate value, and more than anything, it will change **again** how we declare stuff in the domain.

If we want it, we need to think & discuss seriously about it and refactor accordingly all of Connectivity domains at once for consistency.

This is out of scope for the current PR.

src/Akeneo/Connectivity/Connection/back/Infrastructure/Symfony/DependencyInjection/AkeneoConnectivityConnectionExtension.php
```diff
@@ -17,6 +17,8 @@ class AkeneoConnectivityConnectionExtension extends Extension
 {
     public function load(array $configs, ContainerBuilder $container): void
     {
+        $container->setParameter('connectivity.marketplace_fixtures_directory', __DIR__ . '/../Resources/fixtures/');
```
tseho, line 20: why there ?
Can't you use `__DIR__`directly in WebMarketplaceApi to locate the file ?

I'm not seeing the benefit of creating a parameter for it.

tseho (approved):

src/Akeneo/Connectivity/Connection/back/Domain/Marketplace/DTO/Extension.php
tomglvng: Consistency with returned data: the 'id' is an UUID.

tseho: Here https://github.com/akeneo/pim-community-dev/blob/53fc71fa21e13b860ee57baf2ffcfbe5f505ea91/src/Akeneo/Platform/Component/EventQueue/EventInterface.php#L21
It's also an UUID, an UUID can be represented by a string.

What's our deal with UUID in `Extension` ? are we generating UUIDs ? do we need to validate UUIDs ?
Or this ID is simply a string we don't event have to type ?

tseho: Moreover, `Ramsey\Uuid\UuidInterface` is not a domain VO, it's an external interface.
By doing so, you are coupling your domain to an external library, which against our decoupling principles.

src/Akeneo/Connectivity/Connection/back/Infrastructure/Symfony/DependencyInjection/AkeneoConnectivityConnectionExtension.php
tomglvng: In case we want to redefine the service but need another fixture path directory (tests for example), this way of doing is enough flexible.

tseho: The fixtures are a temporary thing for the marketplace extensions.
You can merge anyway, but when adding stuff like "maybe we will need to use it one day in tests even if we don't use today", you pile up technical debt. 
IMHO, it's easier to keep things simple (without sacrificing quality) and wait for the day where we need to customize this, if it happens.

tseho:
Can you also add the missing rules in https://github.com/akeneo/pim-community-dev/blob/master/src/Akeneo/Connectivity/Connection/back/tests/.php_cd.php
regarding this new Domain ?

src/Akeneo/Connectivity/Connection/back/Infrastructure/Symfony/Resources/fixtures/marketplace-data-extensions.json
```diff
@@ -1,5 +1,5 @@
 {
-    "total": 120,
+    "count": 120,
```
tseho, line 2:
```suggestion
    "total": 120,
```

The marketplace spec explicitely specify `total`

tomglvng: Sometimes it was `count` in our code, other `total`. I will harmonize with 'total' so

tseho: yes, but here, it's the mock for the marketplace endpoint, if we don't follow the format of the marketplace API, we will hit a wall :sweat_smile:

src/Akeneo/Connectivity/Connection/back/Domain/Marketplace/DTO/Extension.php
pchasle: Yes, I agree with Quentin, keep it decoupled and simple (string is fine). We generally only add VO when needed.

src/Akeneo/Connectivity/Connection/back/Domain/Marketplace/Model/ExtensionList.php
```diff
@@ -0,0 +1,63 @@
+<?php
+declare(strict_types=1);
+
+namespace Akeneo\Connectivity\Connection\Domain\Marketplace\Model;
```
willy-ahva, line 4: Should it really be in the Model namespace?

tseho (changes requested):
After discussion, plz remove `Ramsey\Uuid` from our domain.
A simple `string` is enough, there is no need for a Value Object for now.

src/Akeneo/Connectivity/Connection/back/Domain/Marketplace/Model/Extension.php
```diff
@@ -0,0 +1,153 @@
...
+     *     logo: string,
+     *     author: string,
+     *     partner: string,
+     *     description: string,
+     *     url: string,
+     *     categories: array<string>,
+     *     certified: bool,
+     * } $values
+     */
+    public static function create(array $values): self
```
willy-ahva, line 57: A named constructor is a very good idea, but the name should reflect the intent. If you use `create` you describe nothing more than a public constructor. Here I would prefer a name like `fromWebMarketplaceValues`

src/Akeneo/Connectivity/Connection/back/Domain/Marketplace/Model/Extension.php
```diff
@@ -0,0 +1,153 @@
...
+        $self->partner = $values['partner'];
+        $self->description = $values['description'];
+        $self->url = $values['url'];
+        $self->categories = $values['categories'];
+        $self->certified = $values['certified'];
+
+        return $self;
+    }
+
+    public function id(): UuidInterface
```
willy-ahva, line 78: If you don't need all those getters, please don't add them. (And use the template `getSomething` for getters).

src/Akeneo/Connectivity/Connection/back/Domain/Marketplace/Model/Extension.php
```diff
@@ -0,0 +1,153 @@
...
+     *  logo: string,
+     *  author: string,
+     *  partner: string,
+     *  description: string,
+     *  url: string,
+     *  categories: array<string>,
+     *  certified: bool,
+     * }
+     */
+    public function normalize(): array
```
willy-ahva, line 139: just an idea, but shouldn't we use `toArray()` instead of `normalize`?

src/Akeneo/Connectivity/Connection/back/Domain/Marketplace/Model/Extension.php
```diff
@@ -0,0 +1,153 @@
...
+     *     description: string,
+     *     url: string,
+     *     categories: array<string>,
+     *     certified: bool,
+     * } $values
+     */
+    public static function create(array $values): self
+    {
+        foreach (self::REQUIRED_KEYS as $key) {
+            Assert::keyExists($values, $key);
```
willy-ahva, line 60: Use your own Domain Exception, please.

src/Akeneo/Connectivity/Connection/back/tests/Unit/spec/Infrastructure/Marketplace/GetAllExtensionsQuerySpec.php
```diff
@@ -0,0 +1,60 @@
...
+/**
+ * @copyright 2021 Akeneo SAS (http://www.akeneo.com)
+ * @license http://opensource.org/licenses/osl-3.0.php Open Software License (OSL 3.0)
+ */
+class GetAllExtensionsQuerySpec extends ObjectBehavior
+{
+    public function let(
+        WebMarketplaceApi $webMarketplaceApi,
+        VersionProviderInterface $versionProvider
+    )
```
willy-ahva, line 23: on 1 line please

src/Akeneo/Connectivity/Connection/back/tests/Unit/spec/Infrastructure/Marketplace/GetAllExtensionsQuerySpec.php
```diff
@@ -0,0 +1,60 @@
...
+    public function it_is_initializable(): void
+    {
+        $this->shouldHaveType(GetAllExtensionsQuery::class);
+    }
+
+    public function it_execute_and_returns_extension_result(
+        WebMarketplaceApi $webMarketplaceApi,
+        VersionProviderInterface $versionProvider
+    )
+    {
```
willy-ahva, line 37: 1 line

src/Akeneo/Connectivity/Connection/back/tests/Unit/spec/Infrastructure/Marketplace/GetAllExtensionsQuerySpec.php
```diff
@@ -0,0 +1,60 @@
...
+        $this->beConstructedWith($webMarketplaceApi, $versionProvider);
+    }
+
+    public function it_is_initializable(): void
+    {
+        $this->shouldHaveType(GetAllExtensionsQuery::class);
+    }
+
+    public function it_execute_and_returns_extension_result(
```
willy-ahva, line 33:
```suggestion
    public function it_executes_and_returns_extension_result(
```

src/Akeneo/Connectivity/Connection/back/Domain/Marketplace/Model/ExtensionList.php
tseho: What about an typed array for `ExtensionList` and keeping the model for `Extension` ?

src/Akeneo/Connectivity/Connection/back/Domain/Marketplace/Model/Extension.php
tseho: It's where this model is falling apart, we don't even use its getters.
We construct an object, then normalize before serializing it for the API Response.

src/Akeneo/Connectivity/Connection/back/Domain/Marketplace/Model/Extension.php
tseho: I've thrown a simple InvalidArgument instead.

tseho (approved):

src/Akeneo/Connectivity/Connection/back/Domain/Marketplace/Model/Extension.php
```diff
@@ -0,0 +1,102 @@
...
+
+declare(strict_types=1);
+
+namespace Akeneo\Connectivity\Connection\Domain\Marketplace\Model;
+
+/**
+ * @copyright 2021 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+class Extension
```
willy-ahva, line 11: final

willy-ahva (approved):

src/Akeneo/Connectivity/Connection/back/Domain/Marketplace/Model/Extension.php
tseho: I can't, it break the specs.

PR #14918, opened 2021-07-06
PIM-9946: forbid import with identical identifier

rfauglas:

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/src/Akeneo/Pim/Enrichment/Component/Product/Validator/UniqueValuesSet.php
+++ b/src/Akeneo/Pim/Enrichment/Component/Product/Validator/UniqueValuesSet.php
@@ -45,15 +45,15 @@ class UniqueValuesSet
         $data = $value->__toString();
         $attributeCode = $value->getAttributeCode();
 
-        if (isset($this->uniqueValues[$attributeCode][$data])) {
-            $storedIdentifier = $this->uniqueValues[$attributeCode][$data];
-            if ($storedIdentifier !== $identifier) {
-                return false;
-            }
-        }
-
         if (!isset($this->uniqueValues[$attributeCode])) {
             $this->uniqueValues[$attributeCode] = [];
+        }
+
+        foreach ($this->uniqueValues[$attributeCode] as $keyData => $storedIdentifier) {
+            if ((strcasecmp($keyData, $data) == 0)
+            && ($storedIdentifier !== $identifier)) {
+                return false;
+            }
         }
 
         if (!isset($this->uniqueValues[$attributeCode][$data])) {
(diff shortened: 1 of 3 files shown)


src/Akeneo/Pim/Enrichment/Component/Product/Validator/UniqueValuesSet.php
```diff
@@ -45,17 +45,17 @@ public function addValue(ValueInterface $value, EntityWithValuesInterface $entit
...
-                return false;
-            }
-        }
-
         if (!isset($this->uniqueValues[$attributeCode])) {
             $this->uniqueValues[$attributeCode] = [];
         }
 
+        foreach ($this->uniqueValues[$attributeCode] as $keyData => $storedIdentifier) {
+            if ((strcasecmp($keyData, $data) == 0)
```
pchasle, line 53: It might not be safe to use `strcasecmp` for a case insensitive comparison.
See https://stackoverflow.com/questions/5473542/case-insensitive-string-comparison

Also, if I look at the issue I was under the impression that it's the identifier that should be checked with a case-insensitive comparison during an import and I'm not sure that this is fixing the correct issue?

src/Akeneo/Pim/Enrichment/Component/Product/Validator/UniqueValuesSet.php
```diff
@@ -45,17 +45,17 @@ public function addValue(ValueInterface $value, EntityWithValuesInterface $entit
...
-            if ($storedIdentifier !== $identifier) {
-                return false;
-            }
-        }
-
         if (!isset($this->uniqueValues[$attributeCode])) {
             $this->uniqueValues[$attributeCode] = [];
         }
 
+        foreach ($this->uniqueValues[$attributeCode] as $keyData => $storedIdentifier) {
```
ahocquard, line 52: why not just add the value as lowercase? It would avoid to perform a foreach on the whole list and benefit an access in O(1) instead of O(n)

rfauglas: Good point.

src/Akeneo/Pim/Enrichment/Component/Product/Validator/UniqueValuesSet.php
rfauglas: Implementation changed ;-)

src/Akeneo/Pim/Enrichment/Component/Product/Validator/UniqueValuesSet.php
rfauglas: I changed it even simpler! ;-)

pchasle (approved):

ahocquard (approved):

LaurentPetard (approved):

romaindak (approved):

ValentinMumble:
check

PR #14923, opened 2021-07-07
Prevent from dropping a production database

gplanchat:
Add counter-measures in order to prevent from dropping a production database with `make prod`

The `DROP_DATABASE=1` variable has to be set in order to run the database installation, replacing the existing one

### DIFF
--- a/std-build/Makefile
+++ b/std-build/Makefile
@@ -53,7 +53,12 @@ front: assets css front-packages javascript-dev
 
 .PHONY: database
 database:
+ifeq (${DROP_DATABASE},1)
 	$(PHP_RUN) bin/console pim:installer:db ${O}
+else
+	@echo "Beware: this action will reset your current database and replace it with a freshly installed (empty) one."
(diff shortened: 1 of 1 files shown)


std-build/Makefile
```diff
@@ -53,7 +53,12 @@ front: assets css front-packages javascript-dev
 
 .PHONY: database
 database:
+ifeq (${DROP_DATABASE},1)
 	$(PHP_RUN) bin/console pim:installer:db ${O}
+else
+	@echo "Beware: this action will reset your current database and replace it with a freshly installed (empty) one."
+	@echo "Please run \x1b[32mDROP_DATABASE=1 make database O=\"${O}\"\x1b[0m if it is your intention to install a fresh database."
```
BitOne, line 60: Thanks @gplanchat for your PR. Can you remove the escaping sequence, as it make things harder on the logs and may not be compatible with all terms.

BitOne (approved):
GTM after fix.

std-build/Makefile
gplanchat: Sure, I've changed a bit the messages to make them clearer without having colors

PR #14934, opened 2021-07-08
CXP-785: add skeleton

willy-ahva:
skeleton

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/src/Akeneo/Connectivity/Connection/front/src/connect/components/MarketplaceIsLoading.tsx
+++ b/src/Akeneo/Connectivity/Connection/front/src/connect/components/MarketplaceIsLoading.tsx
@@ -1,0 +1,83 @@
+import React, {FC} from 'react';
+import {SectionTitle} from 'akeneo-design-system';
+import {useTranslate} from '../../shared/translate';
+import styled, {keyframes} from 'styled-components';
+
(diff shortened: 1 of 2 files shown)


tseho (approved):

romaindak (approved):

RiffVentura:
Shouldn't skeletons of helper, extensions list and app list  be independent ? 
Extensions list and app list are fetched via two different endpoint calls, so helper skeleton should be there until both calls respond
IMO we should have two skeleton components : one for helper, one for a list (we can then reuse it on apps later)

tseho:
> Shouldn't skeletons of helper, extensions list and app list be independent ?
Are we not waiting for the response of both endpoints before displaying the content ?

RiffVentura:
> > Shouldn't skeletons of helper, extensions list and app list be independent ?
> 
> Are we not waiting for the response of both endpoints before displaying the content ?

I was not aware that it was the end goal. Nevermind :)

tseho:
> I was not aware that it was the end goal. Nevermind :)

This is easier, especially for the helper part where we add both counts.

PR #14984, opened 2021-07-16
PLG-107: In product grid, improve the page title when a view is selected

BryanS29-akeneo:

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc


src/Akeneo/Platform/Bundle/UIBundle/Resources/public/js/grid/view-selector.js
```diff
@@ -446,12 +460,14 @@ define([
     selectView: function (view) {
       DatagridState.set(this.gridAlias, {
         view: view.id,
+        text: view.text,
```
olivierpontier, line 463: As we do not use directly the `DatagridState` service to display the `text` parameter. We should remove this line.

src/Akeneo/Platform/Bundle/UIBundle/Resources/public/js/form/common/product/ProductGridContext.tsx
```diff
@@ -0,0 +1,37 @@
...
+  children: string;
+};
+
+const ProductGridContext = ({type, children}: Props) => {
+  const translate = useTranslate();
+
+  return (
+    <ContextContainer>
+      <ViewNameContainer>{children}</ViewNameContainer>{' '}
+      {type === 'public' || type === 'view' ? `(${translate('pim_common.public_view')})` : null}
```
olivierpontier, line 16: In order to ease the readability of the code, we could manage the separating space directly in the template string like below:

```suggestion
      <ViewNameContainer>{children}</ViewNameContainer>
      {type === 'public' || type === 'view' ? ` (${translate('pim_common.public_view')})` : null}
```

olivierpontier:
The `form extensions` allows to define multiple modules to display in a same `dropZone`. So, the `pim-product-index-grid-context` code seems to not be appropriate if we had to display others modules in the `context` drop zone.
Furthermore, we try to name our modules and components in a way to identify what they do.

We should rename the different codes, modules and components in something like `ProductGridViewTitle`. @BryanS29-akeneo @rouliane @LaurentPetard what do you think about it? Do we have suggestion? :pray:

src/Akeneo/Platform/Bundle/UIBundle/Resources/public/js/form/common/product/ProductGridViewTitle.tsx
```diff
@@ -0,0 +1,46 @@
...
+  type: string;
+  projectDetails?: ProjectDetails;
+  children: string;
+};
+
+type ProjectDetails = {
+  dueDateLabel: string;
+  dueDate: string;
+  completionRatio: number;
+  badgeClass: string;
```
olivierpontier, line 15: As we using React component, and the Badge component is existing in the DSM, the badgeClass property does not seem to be useful

olivierpontier (approved):

src/Akeneo/Platform/Bundle/UIBundle/Resources/public/js/form/common/product/product-grid-view-title-context.ts
```diff
@@ -0,0 +1,59 @@
+import {ProductGridViewTitle} from '../../../grid/ProductGridViewTitle';
+
+const BaseView = require('pimui/js/view/base');
+const mediator = require('oro/mediator');
+
+type CurrentView = {
```
LaurentPetard, line 6: As the type is contextual (not exported) I think that it's better to define only the attributes you need (so only `type` and `text`)

LaurentPetard (approved):

PR #15012, opened 2021-07-23
RAC-731: do not return anymore count but category codes from list of category codes and add entrypoint to filter category without view permission

StevenVAIDIE:

We decided to rename and change the intent of CountCategoriesPerTree by GetCategoryChildrenCodesPerTreeInterface because the previous PublicApi is for only for our use case, by returning child code it could be used by another context.

Now we call a null filter in the PublicAPI, in EE we overide this service to apply permission

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc


SamirBoulil (commented):
I don't think this scales well, I'd still keep the counting part done by the SQL query. You need to check this with Alex possibly, but returning children category codes looks too optimistic to me.

I don't think the argument of this publicAPI is too specific for our usecase, I'm sure there will be other cases in the PIM where one will need to get the count of children selected.

tests/back/Pim/Enrichment/Integration/Category/GetCategoryChildrenCodesPerTreeIntegration.php
```diff
@@ -106,48 +106,55 @@ public function invalidCategoryCodes(): array
...
     public function categoryCodesToSelectWithChildren(): array
     {
         return [
-            'Select No categories' => [[], ['master' => 0, 'season' => 0]],
-            'Select root category' => [['master'], ['master' => 7, 'season' => 0]],
+            'Select No categories' => [[], ['master' => [], 'season' => []]],
+            'Select root category' => [
+                ['master'],
+                [
+                    'master' => ['master', 'furniture', 'desk', 'library', 'clothes', 'shoes', 'tshirt'],
```
SamirBoulil, line 149: yes on a simple catalog like this, it makes sense but on a catalog with 1000 of categories, we might have issues.

ValentinMumble (approved):

PR #15036, opened 2021-07-27
CXP-820 Added user and client to AuthCode

RiffVentura:

We cannot use Authorization Code grant type in the PIM because the AuthCode table is incomplete.

If you take a look at the documentation of the OAuth library that we are using (https://github.com/FriendsOfSymfony/FOSOAuthServerBundle/blob/master/Resources/doc/index.md)
You can see that the doctrine entity for AuthCode is defined like this:

Screenshot_2021-08-11_13-46-55

The problem is, when this bundle was installed and configured in the PIM, since it was not used, the 2 columns (client & user) were not implemented.

With the upcoming new feature "Marketplace App Activation", we need those for the new authentication of oauth clients.

This PR adds the missing doctrine column definitions & the migration.

- [ ] Tests
- [x] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc


tseho (approved):

BitOne (approved):

upgrades/test_schema/Version_6_0_20210727080505_add_client_id_and_user_id_Integration.php
```diff
@@ -0,0 +1,102 @@
...
+use Akeneo\Test\Integration\Configuration;
+use Akeneo\Test\Integration\TestCase;
+use Doctrine\DBAL\Connection;
+use PHPUnit\Framework\Assert;
+
+/**
+ * @copyright 2021 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+class Version_6_0_20210727080505_add_client_id_and_user_id_Integration extends TestCase
```
mmetayer, line 16: this test is not really needed, structure migrations are already tested in the test_back_missing_structure_migrations job. Integration tests are only useful if your migration involves data :)

mmetayer (approved):

upgrades/test_schema/Version_6_0_20210727080505_add_client_id_and_user_id_Integration.php
tseho: @mmetayer Damn, I forgot about this. thanks !
Do you think we should trash it before merging or we can keep it like this, since it's already there ?

tseho: We're going to drop it.

PR #14978, opened 2021-07-15
CXP-811 Add new product web API ACL

romaindak:
**Description (for Contributor and Core Developer)**

image

**Definition Of Done (for Core Developer only)**

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Changelog (maintenance bug fixes)
- [ ] Tech Doc

PR #15082, opened 2021-08-10
CXP-811: add a migration for product acls

tseho:

In the PR https://github.com/akeneo/pim-community-dev/pull/14978 that was introducing new ACLs, the lack of migration created inconsistencies between roles in production.

complete technical analysis on notion: https://www.notion.so/akeneo/Technical-analysis-a2a4188bac60430b8e77af655f64ad58

This PR introduce the missing migration.
My only regret is to not be able to write a pure sql migration, I've tried and some OroAcl behavior is completely unexpected.
For example, I was not able to properly manage the `ace_order` value in the `acl_entries` table.
Any mistake in the inserted data will completely corrupt the ACLs when you try to save the same role from the UI with updated permissions, which is unacceptable.

**Edit**

Following @mmetayer's suggestion, I've tried to add more assertions.
And all hell broke loose.
By adding more assertions (which are read-only, I've inspected the SQL logs for any unexpected query from OroAcl), some existing ones are broken.
Worst, the assertions are randomly failing depending on the order they are executed.

Why ?
I went digging with xdebug and the problem is, the ACL services (in `symfony/security-acl`) are completely stateful.
https://github.com/symfony/security-acl/blob/5855e094f7e4d8319d479b08e48decb503ab2aca/Dbal/AclProvider.php#L493-L494
https://github.com/symfony/security-acl/blob/5855e094f7e4d8319d479b08e48decb503ab2aca/Dbal/AclProvider.php#L559-L560

Innocently, a solution would be clear those reference after the migration.
This cannot be done, they explicitly check for it and throw if it happened:
https://github.com/symfony/security-acl/blob/5855e094f7e4d8319d479b08e48decb503ab2aca/Dbal/MutableAclProvider.php#L317

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/upgrades/test_schema/Version_6_0_20210715082931_add_product_web_api_acl_Integration.php
+++ b/upgrades/test_schema/Version_6_0_20210715082931_add_product_web_api_acl_Integration.php
@@ -12,7 +12,6 @@ use Akeneo\Tool\Component\StorageUtils\Saver\SaverInterface;
 use Akeneo\UserManagement\Bundle\Doctrine\ORM\Repository\RoleWithPermissionsRepository;
 use Akeneo\UserManagement\Component\Model\Role;
 use Akeneo\UserManagement\Component\Storage\Saver\RoleWithPermissionsSaver;
-use Akeneo\UserManagement\Component\Updater\RoleWithPermissionsUpdater;
 use Oro\Bundle\SecurityBundle\Acl\Persistence\AclManager;
 use Symfony\Component\Security\Acl\Domain\ObjectIdentity;
 use Symfony\Component\Security\Core\Authentication\Token\UsernamePasswordToken;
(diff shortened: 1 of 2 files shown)


RiffVentura (approved):

upgrades/test_schema/Version_6_0_20210715082931_add_product_web_api_acl_Integration.php
```diff
@@ -0,0 +1,96 @@
...
+
+    public function test_the_migration_adds_the_new_permissions_to_existing_roles()
+    {
+        /** @var SimpleFactoryInterface $roleFactory */
+        $roleFactory = $this->get('pim_user.factory.role');
+        /** @var SaverInterface $roleSaver */
+        $roleSaver = $this->get('pim_user.saver.role');
+
+        /** @var Role $role */
+        $role = $roleFactory->create();
```
mmetayer, line 50: can you also create a role that has some permissions (but not everyone of them) and test that, after the migration, they still have the same permissions plus the 3 new ones?

mmetayer (approved):
a little suggestion for the test, but GTM

upgrades/test_schema/Version_6_0_20210715082931_add_product_web_api_acl_Integration.php
tseho: that's a great idea

tseho: WDYT of the new assertions ?

mmetayer (approved):
seems my suggestion brought you a lot of work, sorry about that :joy:

BitOne:
The main problem I see is the use of "external" services (meaning the migration is not self-sufficient). This means that if these services change in the future, the migration may not work anymore.
It happened in the past big time...

 So do you think it would be possible to replace all the calls to these services with direct SQL calls, so the migration would be self-contained and far more reliable?

I know those ACL tables are a pain in terms of structure, but maybe by looking at the generated SQL queries from the ORM?

BitOne:
As seen with @tseho in Slack, we will use the services here. The business and technical logic around the ACL is very complex (more than 1000 queries to get the data to build the proper change query in a very simple case), and the services involved pretty stable.
So the compromise here is to allow the usage of services for this particular migration.

BitOne (approved):
As seen with @tseho in Slack, we will use the services here. The business and technical logic around the ACL is very complex (more than 1000 queries to get the data to build the proper change query in a very simple case), and the services involved pretty stable.
So the compromise here is to allow the usage of services for this particular migration.

PR #10764, opened 2019-09-23
PIM-6424: update exports when a locale is removed from a channel

rouliane:
**Description (for Contributor and Core Developer)**

Product exports are now updated when a locale is removed of a channel : if the removed locale was used as a filter in an export, the locale filter is removed.

**Definition Of Done (for Core Developer only)**

| Q                                 | A
| --------------------------------- | ---
| Added Specs                       | Todo
| Added legacy Behats               | Todo
| Added acceptance tests            | Todo
| Added integration tests           | Todo
| Changelog updated                 | Todo
| Review and 2 GTM                  | Todo
| Micro Demo to the PO (Story only) | Todo
| Migration script                  | -
| Tech Doc                          | -

`Todo`: Pending / Work in progress
`OK`: Done / Validated
`-`: Not needed

PR #15112, opened 2021-08-13
CXP-831: Remove the hack making all services public in test

tseho:

We had a hack, `MakeServicesPublicForTestEnv.php`, like its name says, which was making all services to `public: true` in test.
I'm going to remove it.

**Why ?**
This is dangerous. 
Last week, we merged a PR where the code was calling a service, directly through the container, without Dependency Injection. In `test`, it was working and the tests were green because it was public (and accessible).
In `prod`, the service was not accessible because it was private. The code was failing. 
Fortunately, it was detected during our upgrade process.
To avoid this issue to happen again, I'm removing this difference between `test` and `prod`.

**What's changing ?**
In integration tests, we still call services through the container. And it will work because since symfony 4.1, there is a special service, only for unit tests, where the services are still available.
But, services that are clearly not used at all and are private are still removed during the container build.
And the tests calling those removed services are now failing.

It's happening in 2 cases:
- the service in CE is only used by EE (so, the CE build will remove it unless it's public, we have to make it public)
- the service is not used at all, and it's burning time :fire: 

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc


tseho (commented):
.

src/Akeneo/Pim/Enrichment/Bundle/Resources/config/queries.yml
```diff
@@ -310,11 +310,6 @@ services:
         arguments:
             - '@database_connection'
 
-    pim_catalog.query.get_channel_active_locale_codes:
-        class: 'Akeneo\Channel\Bundle\Doctrine\Query\GetChannelActiveLocaleCodes'
-        arguments:
-            - '@database_connection'
-
```
tseho, line 317: This query was introduced in 2.3 for a fix https://github.com/akeneo/pim-community-dev/pull/10764/files and it's not used at all in master.

src/Akeneo/Tool/Bundle/ApiBundle/Resources/config/repositories.yml
```diff
@@ -11,6 +11,7 @@ services:
 
     pim_api.repository.product_model:
         class: 'Akeneo\Tool\Bundle\ApiBundle\Doctrine\ORM\Repository\ProductModelRepository'
+        public: true
```
tseho, line 14: It's only used in EE.

src/Oro/Bundle/PimDataGridBundle/Resources/config/queries.yml
```diff
@@ -1,6 +1,7 @@
 services:
     pim_datagrid.product_grid.query.list_attributes:
         class: Oro\Bundle\PimDataGridBundle\Query\Sql\ListAttributesUseableInProductGrid
+        public: true
```
tseho, line 4: this query is used in CE but not in EE, it's removed without `public: true`.

PR #15120, opened 2021-08-16
CXP-821: Authorize endpoint for Apps

tseho:

https://user-images.githubusercontent.com/1421130/129590053-c1fc0740-bd79-4807-ad18-a06da358ea3c.mp4

---

When  Julia click on "Connect Yell to the PIM", the App request an authorization code by calling the url `/connect/apps/v1/authorize` with the following parameters:
- `client_id`
- `response_type`
- `state`
- `scope`
- `redirect_uri`

When the controller is called, we validate the values.
If a validation error occurred, Julia is redirected to the Activation Wizard with the error code in the url.
The Activation Wizard will display correctly the error in another PR.

If the validation succeeded, the parameters are stored in the session and Julia is redirected to the Activation Wizard with the client_id in the url.
The next steps of the Activation Wizard will be done in other PRs.

*Why stored in session ?*

The Activation Wizard is done in several steps and it's only at the last step, when Julia confirmed the whole process, that we will create the connection in the database.

- [x] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc


src/Akeneo/Connectivity/Connection/back/Application/Apps/AppAuthorizationSessionInterface.php
```diff
@@ -0,0 +1,17 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Akeneo\Connectivity\Connection\Application\Apps;
+
+use Akeneo\Connectivity\Connection\Application\Apps\Command\RequestAppAuthorizationCommand;
```
RiffVentura, line 7: not used

src/Akeneo/Connectivity/Connection/back/Infrastructure/Symfony/Resources/translations/jsmessages.en_US.yml
```diff
@@ -65,6 +65,18 @@ akeneo_connectivity.connection:
...
                     error: Sorry, an error occurred while connecting the App.
+            constraint:
+                client_id:
+                    not_blank: The client_id is required.
+                    must_be_valid: The client_id does not match any known client.
+                response_type:
+                    not_blank: The response_type is required.
+                    must_be_code: The response_type must be "code".
+                scope:
+                    must_be_valid: The scopes are not valid.
```
RiffVentura, line 76: Shouldn't we pointed out which scope is at fault here ?

RiffVentura (approved):

src/Akeneo/Connectivity/Connection/back/Infrastructure/Symfony/Resources/translations/jsmessages.en_US.yml
tseho: With the current solution, I'm only sending one error code to the front.
The error is then translated in the Wizard.

If I want to uses more complex translations (like passing placeholders to it), we have to rethink it all.
Do you pass the translated message in the url ?
Do you pass both the error code & the placeholders in the url ?
Do you respond with the page instead of a redirect ?

Ideally, you would have a real response where the url would reply with a http response.
If the controller was using a twig template (for example), we could build the response with a better message, more context (like the google example in CXP-813).

Problem is, we don't use twig anymore, we are really separating the front/back, it's painful for such cases.

tseho: Why are we insisting on processing the request in a backend controller, not in the front like all the others ?

Because we are talking about security, and there is nothing more insecure than the front to do this.

PR #15144, opened 2021-08-19
PIM-10026: fix session being stored during stateless calls

tseho:

During API calls, the session is initialized because we access it on UserContext and LocaleSubscriber.
We don't have to create and save sessions during API calls.

The proposed solution is to check if the current firewall is stateless.
Since the API is of course stateless, it doesn't start a session anymore.
see https://github.com/akeneo/pim-community-dev/blob/master/config/packages/security.yml#L43-L48

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/src/Akeneo/UserManagement/Bundle/Context/UserContext.php
+++ b/src/Akeneo/UserManagement/Bundle/Context/UserContext.php
@@ -371,7 +371,7 @@ class UserContext
     protected function getSessionLocale()
     {
         $request = $this->getCurrentRequest();
-        if (null !== $request && $request->hasSession() && $request->getSession()->isStarted()) {
+        if ($this->hasActiveSession($request)) {
             $localeCode = $request->getSession()->get('dataLocale');
             if (null !== $localeCode) {
                 $locale = $this->localeRepository->findOneByIdentifier($localeCode);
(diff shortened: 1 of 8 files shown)


rfauglas (approved):

jmleroux (approved):

tseho:
@jmleroux too soon, it was not working as expected
With the new version, I think I broke a few unit tests.

jmleroux (approved):

src/Akeneo/UserManagement/Bundle/Context/UserContext.php
```diff
@@ -351,7 +371,7 @@ protected function getRequestLocale()
     protected function getSessionLocale()
     {
         $request = $this->getCurrentRequest();
-        if (null !== $request && $request->hasSession()) {
+        if (null !== $request && $request->hasSession() && $request->getSession()->isStarted()) {
```
BitOne, line 374: In the case of the UI, the session is maybe not yet started. The previous behavior was to create the session. So here there's a risk it doesn't work anymore as it would never reach this part?

tseho: I forgot to change this part back. Yes, `isStarted` was not the way to go.

src/Akeneo/UserManagement/Bundle/Context/UserContext.php
```diff
@@ -112,7 +112,10 @@ public function getCurrentLocale(): LocaleInterface
             throw new \LogicException('There are no activated locales');
         }
 
-        if (null !== $this->getCurrentRequest() && $this->getCurrentRequest()->hasSession()) {
+        if (null !== $this->getCurrentRequest()
+            && $this->getCurrentRequest()->hasSession()
+            && $this->getCurrentRequest()->getSession()->isStarted()
```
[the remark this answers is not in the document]
tseho: You were right and that was the case.

BitOne (approved):

PR #15149, opened 2021-08-19
PIM-10004: Optimize counting job execution warnings

LaurentPetard:
The loading of the process tracker data-grid can be very low when there are a huge number of warnings  (stored in `akeneo_batch_warning`)
It's because this number of warnings is calculate with a `COUNT` on the table `akeneo_batch_warning` that need to be joined to the table `akeneo_batch_step_execution`

As the query is built and executed inside the Oro datagrid stack, and because we want to keep the possibility to sort the job executions by number of warning, the query can't be simply optimized.

This solution is to store the number of warnings for each step, and calculate the number of warnings by job execution with a `SUM`. This will avoid to have to join on the table `akeneo_batch_warning`. 
And thanks to the purges, the size of the table `akeneo_batch_step_execution` remains reasonable enough to keep good performance.

The migration of the data (calculate the number of warnings for every step) could be too long to be done with the schema migration. So it's done in a dedicated command launched from a cron. 
As it would also be costly to determine if the migration is completed, I introduced a new MySQL table to store the status of any "one time" task like that.


src/Akeneo/Tool/Bundle/BatchBundle/Command/MigrateJobExecutionWarningCountCommand.php
```diff
@@ -0,0 +1,149 @@
...
+            ]
+        )->fetch(\PDO::FETCH_ASSOC);
+
+        return isset($bulkResult['first_id']) && isset($bulkResult['last_id']) ? $bulkResult : null;
+    }
+
+    private function updateStepExecutionWarningCount(int $firstId, int $lastId): void
+    {
+        $query = <<<SQL
+UPDATE akeneo_batch_step_execution AS step_execution
```
BitOne, line 126: This kind of request can be quite long and it will lock the table during its execution, which will block all job execution. I would prefer a batch approach in PHP. The total execution time will be longer, but this would avoid locking the whole table.

tomglvng (approved):

upgrades/test_schema/Version_6_0_20210819080024_add_warning_count_in_step_execution_Integration.php
```diff
@@ -0,0 +1,39 @@
...
+    use ExecuteMigrationTrait;
+
+    private const MIGRATION_LABEL = '_6_0_20210819080024_add_warning_count_in_step_execution';
+
+    protected function getConfiguration()
+    {
+        return $this->catalog->useMinimalCatalog();
+    }
+
+    public function test_it_adds_warning_count_column_in_step_execution_table_and_create_one_time_task_table()
```
pchasle, line 20: I think that testing structure changes only is not necessary as it's already taken care of by `.circleci/detect_structure_changes.sh` in the `test_back_missing_structure_migrations` step.
Maybe to confirm with blackhawks

LaurentPetard: Yes I agree

pchasle (approved):

BitOne (approved):

PR #15165, opened 2021-08-23
PLG-56: Free-Trial catalog installation

LaurentPetard:
Define the value of the option "catalog" in all the catalog installation events. 
It's needed to detect that it's the Free-Trial catalog that is being installed


olivierpontier (approved):

BryanS29-akeneo (approved):

rouliane (approved):

LaurentPetard (commented):
Wake-up master

PR #15167, opened 2021-08-23
BH-681: Use PHP CS Fixer v3

jmleroux:
CS Fixer v2 is deprecated and not supported anymore


composer.json
```diff
@@ -54,7 +54,7 @@
         "ext-openssl": "*",
         "ass/xmlsecurity": "1.1.1",
         "box/spout": "2.7.3",
-        "doctrine/annotations": "1.8.0",
+        "doctrine/annotations": "1.13.2",
```
jmleroux, line 57: Needed by CS Fixer v3

composer.json
```diff
@@ -130,7 +130,7 @@
         "maennchen/zipstream-php": "^2.1"
     },
     "require-dev": {
-        "akeneo/php-coupling-detector": "0.3.*",
+        "akeneo/php-coupling-detector": "^0.7.0",
```
jmleroux, line 133: To allow the use of CS Fixer v3

PR #15181, opened 2021-08-24
BH-681: Bundle notation is deprecated

jmleroux:
https://symfony.com/blog/new-in-symfony-4-1-deprecated-the-bundle-notation

https://symfony.com/doc/current/templates.html#bundle-templates

**Note about `@@` notation**:

https://symfony.com/doc/current/service_container.html#service-parameters
```
# if the value of a string argument starts with '@', you need to escape
# it by adding another '@' so Symfony doesn't consider it a service
# the following example would be parsed as the string '@securepassword'
# - '@@securepassword'
```

### DIFF
--- a/src/Oro/Bundle/DataGridBundle/Provider/SystemAwareResolver.php
+++ b/src/Oro/Bundle/DataGridBundle/Provider/SystemAwareResolver.php
@@ -12,6 +12,7 @@ class SystemAwareResolver implements ContainerAwareInterface
     const STATIC_METHOD_CLEAN_REGEX = '#([^\'"%:\s]+)::([\w\._]+)#';
     const SERVICE_METHOD = '#@([\w\._]+)->([\w\._]+)(\((.*)\))*#';
     const SERVICE = '#@([\w\._]+)#';
+    const TWIG_TEMPLATE = '#^@.+\.twig$#';
 
     /**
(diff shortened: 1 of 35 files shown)


BitOne (approved):

ValentinMerlet (approved):

src/Oro/Bundle/DataGridBundle/Provider/SystemAwareResolver.php
```diff
@@ -87,6 +88,9 @@ protected function resolveSystemCall($datagridName, $key, $val)
         }
 
         switch (true) {
+            case preg_match(static::TWIG_TEMPLATE, $val, $match):
```
jmleroux, line 91: I have to add this (dirty) check because this class would not be compatible with the modern Twig template notation

ValentinMerlet: There is no instructions in your `case` statement, so the behavior is the same as the `default`, I'm not sure to understand why you're adding this case.

jmleroux: Oh yes, sorry for my poor explanation: if I don't add this, the template is identified as a service because of the `static::SERVICE` regex.

PR #15183, opened 2021-08-24
Cpm 340 - add a focus on label or search component

ayasakeneo:
Add a focus on label or search component

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc


src/Akeneo/Pim/Structure/Bundle/Resources/workspaces/settings-ui/src/pages/CreateAttributeModal.tsx
```diff
@@ -41,13 +41,23 @@ const CreateAttributeModal: React.FC<CreateAttributeModalProps> = ({
...
     onStepConfirm({code, label, ...extraFieldsData});
   };
 
+  const focus = (labRef: React.RefObject<HTMLInputElement>) => {
+    labRef?.current?.focus();
+  };
+
+  React.useEffect(() => {
+    focus(labelRef);
+  }, []);
```
StevenVAIDIE, line 59: You could use the hook `useAutoFocus`hook from the DSM => an example of usage => https://github.com/akeneo/pim-enterprise-dev/blob/master/src/Akeneo/AssetManager/front/application/component/library/MassDelete/MassDeleteModal.tsx#L24.

It will not work on the other file because Search is not rendered at first render but here it will be more readable

PR #15218, opened 2021-08-30
BH-681: Update Doctrine dependencies

jmleroux:


src/Akeneo/Pim/Enrichment/Component/Product/Repository/ReferenceDataRepositoryResolverInterface.php
```diff
@@ -14,7 +16,7 @@ interface ReferenceDataRepositoryResolverInterface
     /**
      * @param string $referenceDataType
      *
-     * @return \Doctrine\Common\Persistence\ObjectRepository
+     * @return ObjectRepository
```
rfauglas, line 19: move to  method signature?

rfauglas (approved):

config/packages/doctrine.yml
```diff
@@ -59,3 +59,5 @@ doctrine_migrations:
     dir_name: '%doctrine_migrations_dir%'
     namespace: Pim\Upgrade\Schema
     name: PIM Migrations
+    column_name: 'version'
+    column_length: 191
```
docteurklein, line 63: if I'm not mistaken, this won't be taken into account if the table already exists https://github.com/doctrine/migrations/pull/970

jmleroux: Indeed. It should not be an issue but I had some metadata issues lately. I still have some work on this PR and will fix them in the meantime.

ahocquard (approved):

composer.json
```diff
@@ -55,24 +55,23 @@
...
+        "doctrine/event-manager": "^1.1.1",
+        "doctrine/instantiator": "^1.4.0",
+        "doctrine/migrations": "^2.3.4",
+        "doctrine/orm": "^2.9.5",
+        "doctrine/persistence": "^2.2.2",
         "dompdf/dompdf" : "1.0.2",
         "elasticsearch/elasticsearch": "7.11.0",
         "friendsofsymfony/jsrouting-bundle": "2.7.0",
-        "friendsofsymfony/oauth-server-bundle": "1.6.2",
+        "cutwise/oauth-server-bundle": "2.0.4",
```
nmarniesse, line 74: This package is not very used https://packagist.org/packages/cutwise/oauth-server-bundle
At least the version is fixed but be aware using this package could be risky?

BitOne (approved):

composer.json
BitOne: Agree with Nini.

It would be less risky to fork it in the Akeneo organisation.

config/packages/doctrine.yml
jmleroux: The column is the same than the one used for previous PIM versions and I changed the length to use 190 max which is the actual length value.

PR #15224, opened 2021-08-30
PIM-10037: Fix family variant query to return correct number of results

ValentinMumble:

Fix the get families with variant query to select distinct results in order for the limit parameter to work as intended.
Before, with a limit of 20, the query would return multiple times the same family, as the select was not distinct:
image
After the DISTINCT, we have the correct number of results:
image

- [x] PM Validation (Story)


rfauglas (approved):
Pas de test d'intégration?

anaisbaunelemaire (approved):

PR #15238, opened 2021-09-01
BH-681: Fix various deprecations

jmleroux:


composer.json
```diff
@@ -150,8 +151,8 @@
         "squizlabs/php_codesniffer": "2.*",
         "friends-of-behat/symfony-extension": "^2.2",
         "liuggio/fastest": "1.8.*",
-        "phpstan/phpstan": "^0.12.25",
-        "phpstan/phpstan-symfony": "^0.12.6",
+        "phpstan/phpstan": "^0.12.97",
+        "phpstan/phpstan-symfony": "0.12.42",
```
jmleroux, line 155: `phpstan/phpstan-symfony` was updated to 0.12.43 a few hours ago resulting in a lot of "false negatives" and blocking the CI

composer.json
```diff
@@ -47,6 +47,7 @@
         "ext-gd": ">=2.0",
         "ext-imagick": "*",
         "ext-intl": "*",
+        "ext-json": "*",
```
ahocquard, line 50: to remove for 8.0 I think (https://php.watch/versions/8.0/ext-json)

ahocquard (approved):

composer.json
jmleroux: Good catch. I will then not add it in the PR because the final purpose is to migrate to PHP8

ValentinMerlet (approved):

PR #15250, opened 2021-09-03
BH-681: Remove old migration scripts

jmleroux:
These migrations scripts are useless for v6 and later


SamirBoulil (approved):
Can you check there no tests left associated with those migrations ?

olivierpontier (approved):

jmleroux:
> Can you check there no tests left associated with those migrations ?

CI should tell :wink:

PR #15256, opened 2021-09-06
CXP-812: fix migration

jmleroux:
This migration is failing when testing for 4.0 upgrade to master
https://app.circleci.com/pipelines/gh/akeneo/pim-community-standard/3898/workflows/f26b907f-11c2-4708-90b2-d72972052742/jobs/5516

We try to fetch the Roles with Doctrine ORM and the Role entity is not compatible with a 4.0 database because of the `type` column not existing in 4.0: https://github.com/akeneo/pim-community-dev/blob/master/src/Akeneo/UserManagement/Bundle/Resources/config/model/doctrine/Role.orm.yml#L20

By fetching only the role column, we are sure this migration will be compatible with both databases.

### DIFF
--- a/upgrades/schema/Version_4_1_20200127171059_unauthorized_view_all_jobs_permission.php
+++ b/upgrades/schema/Version_4_1_20200127171059_unauthorized_view_all_jobs_permission.php
@@ -3,6 +3,7 @@
 namespace Pim\Upgrade\Schema;
 
 use Doctrine\Common\Collections\ArrayCollection;
+use Doctrine\DBAL\FetchMode;
 use Doctrine\DBAL\Schema\Schema;
 use Doctrine\Migrations\AbstractMigration;
 use Oro\Bundle\SecurityBundle\Model\AclPermission;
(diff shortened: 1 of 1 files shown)


ahocquard:
If you migrate from 4.0 to master, you should not migrate to 5.0 first? I mean, it's impossible to code something that is forward compatible, as you don't know the future (#madameirma).

upgrades/schema/Version_4_1_20200127171059_unauthorized_view_all_jobs_permission.php
```diff
@@ -12,17 +13,18 @@
...
 
     /** @var ContainerInterface */
     private $container;
 
-    public function up(Schema $schema) : void
+    public function up(Schema $schema): void
     {
         $aclManager = $this->container->get('oro_security.acl.manager');
-        $roles = $this->container->get('pim_user.repository.role')->findAll();
+        $roles = $this->fetchRoles();
```
ahocquard, line 27: it expects a doctrine entity inside the line below, no?

```
$aclManager->getPrivilegeRepository()
->savePrivileges(new RoleSecurityIdentity($role), new ArrayCollection([$privilege]));
```

jmleroux: Nope, it works with both entity and string. And the master version will work only with strings.

jmleroux:
> If you migrate from 4.0 to master, you should not migrate to 5.0 first? I mean, it's impossible to code something that is forward compatible, as you don't know the future (#madameirma).

The issue will be the same, the database change was introduced in master (6.0)

nmarniesse (approved):

ahocquard (approved):

tseho (approved):

BitOne (approved):

PR #15264, opened 2021-09-07
BH-681: Use FriendsOfBehat libraries

jmleroux:
behat/Symfony2Extension is deprecated and we should use the most up-to-date FoB libraries

One of the main change is that all contextes are real services included in Symfony DI, not just in specific `behat.yml` config

Some cleaning about return types was also necessary (and will still be after merge :wink: )


tests/legacy/features/Behat/Context/HookContext.php
```diff
@@ -177,26 +177,6 @@ public function collectErrors()
         }
     }
 
-    /**
-     * @BeforeScenario
-     */
-    public function maximize()
```
jmleroux, line 183: Handled by Behat config

momoss (approved):

tests/legacy/features/Context/FeatureContext.php
```diff
@@ -217,18 +153,23 @@ public function wait($condition = null)
...
-                "$('#page').css('display') != 'none'",         // Page is displayed (no progress bar)
+            $conditions = [
+                "document.readyState == 'complete'",
+                // Page is ready
+                "typeof $ != 'undefined'",
+                // jQuery is loaded
+                "!$.active",
+                // No ajax request is active
+                "$('#page').css('display') != 'none'",
+                // Page is displayed (no progress bar)
```
ahocquard, line 168: a little bit messy to have the comment after the line, except for this line... I would keep the previous indentation

ahocquard (approved):

franckbrouard (approved):
LGTM ! nice work JM :)

jmleroux:
merging without EE build because I have a dedicated green PR for EE and i'd like to unlock the builds for other mates.

PR #15266, opened 2021-09-07
PIM-10063: make query builder case insensitive for attribute filter

nmarniesse:

Before the query builder was case sensitive . That means if the attribute code is `test` we cannot filter on `TEST`.  
It was like that because we can create some attribute with code like `CATEGORIES`, `FAMILIES`, ... So when a filter is added on `categories` we cannot know if this refers to the system filter or the attribute filter.  
Having the PQB case sensitive provides some bad UX, as the PIM is case insensitive almost everywhere, and not here.  

Now the PQB is case insensitive. To do that:
- now we forbid the creation of attribute with `CATEGORIES` code. So the conflict could not happen
- Our SAAS environments does not have attribute like that, so there is no conflict with the existent attribute
- the PQB is now simplified, and we can even save a query if we add a system filter

- [x] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/src/Akeneo/Pim/Enrichment/Component/Product/Query/AbstractEntityWithValuesQueryBuilder.php
+++ b/src/Akeneo/Pim/Enrichment/Component/Product/Query/AbstractEntityWithValuesQueryBuilder.php
@@ -21,29 +21,16 @@ class AbstractEntityWithValuesQueryBuilder implements ProductQueryBuilderInterfa
     public const DOCUMENT_TYPE_FACET_NAME = 'document_type_facet';
     private const DOCUMENT_TYPE_FIELD = 'document_type';
 
-    /** @var AttributeRepositoryInterface */
-    protected $attributeRepository;
-
     /** @var mixed */
     protected $qb;
 
-    /** @var FilterRegistryInterface */
-    protected $filterRegistry;
-
-    /** @var SorterRegistryInterface */
-    protected $sorterRegistry;
-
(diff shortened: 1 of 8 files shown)


src/Akeneo/Pim/Enrichment/Component/Product/Query/AbstractEntityWithValuesQueryBuilder.php
```diff
@@ -110,22 +97,16 @@ public function getQueryBuilder()
      */
     public function addFilter($field, $operator, $value, array $context = [])
     {
-        $code = FieldFilterHelper::getCode($field);
-        $attribute = $this->attributeRepository->findOneByIdentifier($code);
-
-        // In case of non case sensitive database configuration you can have attributes with code matching a field.
-        // For example "id" would match an attribute named "ID" so we double check here that we are adding the desired
-        // filter (ref PIM-6064)
-        if (null !== $attribute && $attribute->getCode() !== $code) {
```
nmarniesse, line 119: The new attribute validation prevents to have attributes with code like `ID`, `CATEGORIES`, ... And we check our SAAS clients does not have attributes like that. So now we can remove this part with confidence

olivierpontier (approved):

RiffVentura (approved):

rfauglas (approved):

PR #15270, opened 2021-09-08
CXP-811: Add API ACLs to existing roles in pure SQL

tseho:

The original migration had mixed results observed in production.
Some environments did not have the new ACLs on all roles.
This was slightly expected, due to unpredictability of the legacy ACL system & its cache.

Worst, when the original migration was run a second time on those broken environments, the ACLs were fixed.

Since it's unthinkable to go "let's execute the migration until it's working", I'm going for a new migration, SQL-only this time.

To compare the results between the migrations, I did this small script to compare the after state of both migrations.
Since I know the original is working locally, with a fresh PIM without cache, I'm quit confident in the results.

```
#/bin/bash

APP_ENV=test make database
APP_ENV=test docker-compose run -u www-data --rm -T php php vendor/bin/phpunit -c . --testsuite PIM_Integration_Test upgrades/test_schema/Version_6_0_20210715082931_add_product_web_api_acl_Integration.php
docker exec -i $(docker-compose ps -q mysql) mysqldump --skip-extended-insert -u root -proot akeneo_pim_test > dump_old.sql

APP_ENV=test make database
APP_ENV=test docker-compose run -u www-data --rm -T php php vendor/bin/phpunit -c . --testsuite PIM_Integration_Test upgrades/test_schema/Version_6_0_20210908142400_add_product_web_api_acl_Integration.php
docker exec -i $(docker-compose ps -q mysql) mysqldump --skip-extended-insert -u root -proot akeneo_pim_test > dump_new.sql

diff dump_old.sql dump_new.sql
```

I've verified that there is no diff between those two AND that the ACLs are correctly persisted in the database.

Sorry if the migration is painful to read :(

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/upgrades/schema/Version_6_0_20210908142400_add_product_web_api_acl.php
+++ b/upgrades/schema/Version_6_0_20210908142400_add_product_web_api_acl.php
@@ -55,7 +55,7 @@ final class Version_6_0_20210908142400_add_product_web_api_acl extends AbstractM
 
     private function aclIsRegistered(string $acl): bool
     {
-        return (bool) $this->connection->fetchOne(
+        return (bool) $this->connection->fetchColumn(
             <<<SQL
 SELECT COUNT(*)
 FROM acl_classes
@@ -79,7 +79,7 @@ SQL,
             ]
         );
 
-        $aclClassId = (int) $this->connection->fetchOne(
+        $aclClassId = (int) $this->connection->fetchColumn(
             <<<SQL
 SELECT id
 FROM acl_classes
@@ -110,7 +110,7 @@ SQL,
             ]
         );
 
-        $aclObjectIdentityId = (int) $this->connection->fetchOne(
+        $aclObjectIdentityId = (int) $this->connection->fetchColumn(
             <<<SQL
 SELECT id
 FROM acl_object_identities
@@ -140,7 +140,7 @@ SQL,
 
     private function aclIsAlreadyDefinedForRole(string $role, string $acl): bool
     {
-        return (bool) $this->connection->fetchOne(
+        return (bool) $this->connection->fetchColumn(
             <<<SQL
 SELECT COUNT(*)
 FROM acl_entries
@@ -158,7 +158,7 @@ SQL,
 
     private function addAclToRole(string $role, string $acl, bool $granted): void
     {
-        $classId = (int) $this->connection->fetchOne(
+        $classId = (int) $this->connection->fetchColumn(
             <<<SQL
 SELECT id
 FROM acl_classes
@@ -169,7 +169,7 @@ SQL,
             ]
         );
 
-        $securityEntityId = (int) $this->connection->fetchOne(
+        $securityEntityId = (int) $this->connection->fetchColumn(
             <<<SQL
 SELECT id
 FROM acl_security_identities
@@ -250,7 +250,7 @@ SQL,
 
     private function roleIsGrantedEverythingByDefault(string $role): bool
     {
-        return (bool) $this->connection->fetchOne(
+        return (bool) $this->connection->fetchColumn(
             <<<SQL
 SELECT COUNT(*)
 FROM acl_entries
@@ -273,12 +273,14 @@ SQL,
      */
     private function getRoles(): array
     {
-        return $this->connection->fetchFirstColumn(
+        return array_map(function($row) {
+            return $row['identifier'];
+        }, $this->connection->fetchAll(
             <<<SQL
 SELECT identifier
 FROM acl_security_identities
 SQL
-        );
+        ));
     }
 
     /**
--- a/upgrades/schema/Version_6_0_20210908142400_add_product_web_api_acl.php
+++ b/upgrades/schema/Version_6_0_20210908142400_add_product_web_api_acl.php
@@ -1,0 +1,305 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Pim\Upgrade\Schema;
+
+use Doctrine\DBAL\Schema\Schema;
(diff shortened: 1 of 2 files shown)


upgrades/schema/Version_6_0_20210908142400_add_product_web_api_acl.php
```diff
@@ -0,0 +1,305 @@
...
+        );
+    }
+
+    private function registerAcl(string $acl): void
+    {
+        $this->connection->executeQuery(
+            <<<SQL
+INSERT INTO acl_classes (class_type)
+VALUES (:acl)
+SQL,
```
jmleroux, line 76:
```suggestion
            'INSERT INTO acl_classes (class_type) VALUES (:acl)',
```
Just to say we can also use simple strings...

BitOne (approved):

upgrades/schema/Version_6_0_20210908142400_add_product_web_api_acl.php
```diff
@@ -0,0 +1,305 @@
...
+                'acl' => $acl,
+            ]
+        );
+
+        $aclClassId = (int) $this->connection->fetchOne(
+            <<<SQL
+SELECT id
+FROM acl_classes
+WHERE class_type = :acl
+SQL,
```
jmleroux, line 87:
```suggestion
            <<<SQL
                SELECT id
                FROM acl_classes
               WHERE class_type = :acl
           SQL,
```

jmleroux: Is it not more readable ?

jmleroux (approved):
Only small style suggestions, great work Quentin!

upgrades/schema/Version_6_0_20210908142400_add_product_web_api_acl.php
```diff
@@ -0,0 +1,305 @@
...
+    acl_entries.object_identity_id = (
+        SELECT aoi.id
+        FROM acl_object_identities aoi
+        JOIN acl_classes ac on aoi.class_id = ac.id
+        WHERE object_identifier = "action"
+        AND ac.class_type = "(root)"
+        LIMIT 1
+    )
+)
+ORDER BY ace_order DESC
```
ahocquard, line 209: useless order by as it's an update

ahocquard (approved):
It looks ok. I crosscheck with SQL statement I did in https://github.com/akeneo/pim-community-dev/blob/master/tests/back/Integration/catalog/technical/acl.sql

Let's :crossed_fingers: on the `acl_entries` table.

FYI, you can use `select` in an update `query (see link above).

upgrades/schema/Version_6_0_20210908142400_add_product_web_api_acl.php
tseho: are you sure ?
There is an unique index on `class_id, `object_identity_id, field_name, ace_order`.
By updating from the higher to the lowest one, I will never violate the constraint.
Maybe it's useless and I've done so for years for nothing.

upgrades/schema/Version_6_0_20210908142400_add_product_web_api_acl.php
```diff
@@ -0,0 +1,305 @@
...
+        );
+    }
+
+    private function aclIsAlreadyDefinedForRole(string $role, string $acl): bool
+    {
+        return (bool) $this->connection->fetchOne(
+            <<<SQL
+SELECT COUNT(*)
+FROM acl_entries
+LEFT JOIN acl_security_identities ON acl_security_identities.id = acl_entries.security_identity_id
```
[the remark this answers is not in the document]
tseho: you are right

upgrades/schema/Version_6_0_20210908142400_add_product_web_api_acl.php
ahocquard: Ah you're right, I didn't know there was such index. Mysql does not respect deferrable checking (which is the standard), so it's mandatory.

upgrades/schema/Version_6_0_20210908142400_add_product_web_api_acl.php
```diff
@@ -0,0 +1,305 @@
...
+    :object_identity_id
+)
+SQL,
+            [
+                'object_identity_id' => $aclObjectIdentityId,
+            ]
+        );
+    }
+
+    private function aclIsAlreadyDefinedForRole(string $role, string $acl): bool
```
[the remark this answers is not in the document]
tseho: If the ACL is already defined in the database, either true/false, it means that the users manually changed it from the UI.
I don't want to override that.
I'm only inserting the new ACL for Roles that were never updated with this ACL.

> Are we sure this function does not return always true on a live environment?

I was able to check on the sql dump of one of the live environment where the previous migration was not executed properly, this line was simply missing.

upgrades/schema/Version_6_0_20210908142400_add_product_web_api_acl.php
tseho: As I said, I prefer to write the SQL queries like this.
This is what we do usually in the PIM, since a long time.

If we decide to add a different rule about it, I will comply without issue.
For the current PR, I don't feel like rewriting it.

Also, regarding the fact that there is a lot of queries in this migration and it can be visually annoying, I don't want to hide the fact that the current tables for the ACLs are much harded to migrate than necessary.

I hope you don't mind that I resolve your comments about CS.

PR #15280, opened 2021-09-10
BH-681: Use doctrine/migrations v3

jmleroux:
Use v3 library version to be up to date and PHP8 / Symfony5 ready.

The main change is that we now must use the FQDN of migrations


src/Akeneo/Platform/Bundle/InstallerBundle/Command/DatabaseCommand.php
```diff
@@ -71,7 +57,6 @@ public function __construct(
         $this->connection = $connection;
         $this->fixtureJobLoader = $fixtureJobLoader;
         $this->eventDispatcher = $eventDispatcher;
-        $this->env = $env;
```
jmleroux, line 74: Not used in the command

src/Akeneo/Platform/Bundle/InstallerBundle/Command/DatabaseCommand.php
```diff
@@ -346,6 +314,8 @@ private function setLatestKnownMigration(InputInterface $input): void
     {
         $latestMigration = $this->getLatestMigration($input);
 
+        $this->commandExecutor->runCommand('doctrine:migrations:sync-metadata-storage', ['-q' => true]);
```
jmleroux, line 317: To sunchronize migration table structure. Idempotent, can be played even if not needed.

config/packages/doctrine.yml
```diff
@@ -56,6 +56,11 @@ framework:
...
 doctrine_migrations:
-    dir_name: '%doctrine_migrations_dir%'
-    namespace: Pim\Upgrade\Schema
-    name: PIM Migrations
+    migrations_paths:
+        'Pim\Upgrade\Schema': '%doctrine_migrations_dir%'
+    storage:
+        table_storage:
+            table_name: 'migration_versions'
+            version_column_name: 'version'
```
StevenVAIDIE, line 64: Should we really need to change it ? cf:https://github.com/doctrine/migrations/blob/3.2.x/UPGRADE.md they recommand to no configure it they are the default value
Screenshot from 2021-09-13 14-39-40

StevenVAIDIE (approved):
Onboarder use some class of the old namespace deleted in 3.1 Doctrine\DBAL\Migrations like \Doctrine\DBAL\Migrations\IrreversibleMigrationException or Doctrine\DBAL\Migrations\AbstractMigration maybe we should also made a PR in this repo

config/packages/doctrine.yml
jmleroux: Yes we need, because the default name is now `doctrine_migrations_version`.
The version_column_name is not mandatory, but I kept the same than previous versions

jmleroux:
> Onboarder use some class of the old namespace deleted in 3.1 Doctrine\DBAL\Migrations like \Doctrine\DBAL\Migrations\IrreversibleMigrationException or Doctrine\DBAL\Migrations\AbstractMigration maybe we should also made a PR in this repo

Yes we should, and in fact it should have been detected in the CI on EE side. :thinking:

jmleroux:
> Onboarder use some class of the old namespace deleted in 3.1 Doctrine\DBAL\Migrations like \Doctrine\DBAL\Migrations\IrreversibleMigrationException or Doctrine\DBAL\Migrations\AbstractMigration maybe we should also made a PR in this repo

@StevenVAIDIE , I think we already fixed the deprecations in another PR and that's why CI is green.
An example of current Onboarder migrations: https://github.com/akeneo/pim-onboarder/blob/master/upgrades/schema/Version20201026100039_catch_up_various_tables_structure_with_4_3.php#L63

BitOne (approved):

PR #15293, opened 2021-09-14
Update migrations

jmleroux:
Version_6_0_20210715082931_add_product_web_api_acl migration with Version_6_0_20210817135301_add_role_type_column

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -842,7 +842,7 @@ workflows:
           - test_back_data_migrations:
                 name: test_back_data_migrations_ce
                 requires:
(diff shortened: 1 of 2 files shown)


tseho (approved):

tseho:
by the way, since this migration is now duplicated/made useless by the latest one which is sql only,
Is it an option to make it optional ?
That would prevent future issues.

We could either wrap the full migration with a try/catch or simply remove its content.
WDYT ?

jmleroux:
Closed in favor of https://github.com/akeneo/pim-community-dev/pull/15297

PR #15300, opened 2021-09-15
CXP-868: read only mode on form if not granted

willy-ahva:

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/src/Akeneo/Connectivity/Connection/workspaces/permission-form/src/component/PermissionFormWidget.tsx
+++ b/src/Akeneo/Connectivity/Connection/workspaces/permission-form/src/component/PermissionFormWidget.tsx
@@ -20,6 +20,7 @@ type Props = {
     onAdd: (identifier: string) => void;
     onRemove: (identifier: string) => void;
     disabled: boolean;
+    readOnly: boolean;
     allByDefaultIsSelected: boolean;
     onSelectAllByDefault: () => void;
     onDeselectAllByDefault: () => void;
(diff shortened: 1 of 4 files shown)


tseho (approved):

src/Akeneo/Connectivity/Connection/workspaces/permission-form/src/component/PermissionFormWidget.tsx
```diff
@@ -51,13 +53,14 @@ export const PermissionFormWidget: FC<Props> = ({
                 value={selection}
                 onAdd={onAdd}
                 onRemove={onRemove}
-                disabled={disabled}
+                disabled={disabled || readOnly}
```
RiffVentura, line 56: `readOnly` seems redundant, We can just use ` disabled` it achieves the same, and semantically its accurate

willy-ahva: disabled flag is used to disabled the select but not the checkbox. Read-only is to put the whole field and the checkbox disabled.

RiffVentura: In which case would you have one and not the other ?
If you can't set permissions one by one why would you be able to set all of them ?

willy-ahva: If you clicked on the checkbox to turn it to "All" you want to have the select disabled but not the checkbox because you want to be able to uncheck it. If you don't have the permission, you want to have the entire field disabled.

RiffVentura: oh I get it now

RiffVentura (approved):

PR #15301, opened 2021-09-15
CXP-842: add support for select2 QueryParamsBuilder

tseho:

This is needed for the EE PR : https://github.com/akeneo/pim-enterprise-dev/pull/11764

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/src/Akeneo/Connectivity/Connection/workspaces/permission-form/src/component/MultiSelectInputWithDynamicOptions.tsx
+++ b/src/Akeneo/Connectivity/Connection/workspaces/permission-form/src/component/MultiSelectInputWithDynamicOptions.tsx
@@ -43,8 +43,6 @@ type Props = {
     onRemove?: (value: string) => void;
 };
 
-const defaultBuildQueryParams = (term: string) => ({q: term});
-
 export const MultiSelectInputWithDynamicOptions = ({
     url,
     fetchByIdentifiers,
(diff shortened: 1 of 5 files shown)


RiffVentura (approved):

src/Akeneo/Connectivity/Connection/workspaces/permission-form/src/component/MultiSelectInputWithDynamicOptions.tsx
```diff
@@ -42,13 +47,16 @@ export const MultiSelectInputWithDynamicOptions = ({
...
+    buildQueryParams,
     disabled,
     value,
     onChange,
     onAdd,
     onRemove,
 }: Props) => {
     const ref = useRef<HTMLInputElement>(null);
+
+    /* istanbul ignore next */
```
masacc, line 59: for my personal culture, when do you choose to add these comment lines ?

tseho: For the workspace `permission-form`, I'm targeting 100% coverage.
In a few rares cases, like here where I interact with select2, you cannot test some lines.
So instead of lowering the coverage score, I will admit that I'm failing to test this and it should bother newer developments.

Also, why can't we test this particular `handleInitSelection` function ?
This function is given to select2 configuration as an option. It's only called by select2, deep into the library, when it is instancied with a list of identifiers.
I don't want to test select2 directly and I don't know how to check that is method is properly executed by select2 with the correct context.

Until now, we had a coverage of 90% because I was waiting for a solution to this, I have not.

willy-ahva (approved):

PR #15316, opened 2021-09-17
CXP-840 Confirm App Wizard with steps

RiffVentura:

Added confirm handling for App Wizard with steps
Sends confirm App request then sequentially saves each form provider

### DIFF
--- a/src/Akeneo/Connectivity/Connection/back/Infrastructure/Symfony/Resources/translations/jsmessages.en_US.yml
+++ b/src/Akeneo/Connectivity/Connection/back/Infrastructure/Symfony/Resources/translations/jsmessages.en_US.yml
@@ -63,6 +63,9 @@ akeneo_connectivity.connection:
             unreachable: We can't reach the marketplace, please come back later.
             scroll_to_top: Scroll to the top
         apps:
+            flash:
+                permissions_error.title: 'Permissions on {{ entity }} could not be saved.'
+                permissions_error.description: Please check the App permission settings.
             wizard:
                 title: Connect
                 action:
@@ -84,8 +87,6 @@ akeneo_connectivity.connection:
                 flash:
                     success: App successfully configured.
                     error: Sorry, an error occurred while connecting the App.
-                    permissions_error.title: 'Permissions on {{ entity }} could not be saved.'
-                    permissions_error.description: Please check the App permission settings.
                 permission:
                     helper: If you need help please check our article
                     helper_link: How to set-up app permissions?
--- a/src/Akeneo/Connectivity/Connection/back/Infrastructure/Symfony/Resources/translations/jsmessages.en_US.yml
+++ b/src/Akeneo/Connectivity/Connection/back/Infrastructure/Symfony/Resources/translations/jsmessages.en_US.yml
@@ -84,6 +84,8 @@ akeneo_connectivity.connection:
                 flash:
                     success: App successfully configured.
                     error: Sorry, an error occurred while connecting the App.
+                    permissions_error.title: 'Permissions on {{ entity }} could not be saved.'
+                    permissions_error.description: Please check the App permission settings.
                 permission:
                     helper: If you need help please check our article
                     helper_link: How to set-up app permissions?
--- a/src/Akeneo/Connectivity/Connection/front/src/connect/components/AppWizardWithSteps/AppWizardWithSteps.tsx
+++ b/src/Akeneo/Connectivity/Connection/front/src/connect/components/AppWizardWithSteps/AppWizardWithSteps.tsx
@@ -9,6 +9,8 @@ import {AppWizardData} from '../../../model/Apps/wizard-data';
 import {useFetchAppWizardData} from '../../hooks/use-fetch-app-wizard-data';
 import {useTranslate} from '../../../shared/translate';
 import {PermissionFormProvider, usePermissionFormRegistry} from '../../../shared/permission-form-registry';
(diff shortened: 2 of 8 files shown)


src/Akeneo/Connectivity/Connection/back/Infrastructure/Symfony/Resources/translations/jsmessages.en_US.yml
```diff
@@ -84,6 +84,8 @@ akeneo_connectivity.connection:
                 flash:
                     success: App successfully configured.
                     error: Sorry, an error occurred while connecting the App.
+                    permissions_error.title: 'Permissions on {{ entity }} could not be saved.'
+                    permissions_error.description: Please check the App permission settings.
```
RiffVentura, line 88: it might be useful to move these messages out of wizard namespace, as we could use them inside App settings

src/Akeneo/Connectivity/Connection/front/src/connect/components/AppWizardWithSteps/AppWizardWithSteps.tsx
```diff
@@ -73,6 +78,43 @@ export const AppWizardWithSteps: FC<Props> = ({clientId}) => {
         history.push('/connect/marketplace');
     };
 
+    const handleConfirm = async () => {
+        try {
```
tseho, line 82: Why do you catch everything ?
Wouldn't that be better if you catch each part individually and react depending on it ?

willy-ahva (approved):

willy-ahva (commented):
Waiting for test but it's ok for now

src/Akeneo/Connectivity/Connection/front/src/connect/components/AppWizardWithSteps/AppWizardWithSteps.tsx
```diff
@@ -73,6 +78,43 @@ export const AppWizardWithSteps: FC<Props> = ({clientId}) => {
...
+            const {userGroup} = await confirmAuthorization();
+
+            for (const provider of providers) {
+                const success = await provider.save(userGroup, permissions[provider.key]);
```
tseho, line 86: What happen if the `save` method throw an error instead of reject a promise ?
Does it stop the other providers ? will it display the error `akeneo_connectivity.connection.connect.apps.wizard.flash.error` instead of the explicit error we defined ?

src/Akeneo/Connectivity/Connection/front/src/connect/components/AppWizardWithSteps/AppWizardWithSteps.tsx
```diff
@@ -73,6 +78,43 @@ export const AppWizardWithSteps: FC<Props> = ({clientId}) => {
...
+                        translate(
+                            'akeneo_connectivity.connection.connect.apps.wizard.flash.permissions_error.description'
+                        ),
+                        {
+                            titleMessage: translate(
+                                'akeneo_connectivity.connection.connect.apps.wizard.flash.permissions_error.title',
+                                {entity: provider.label}
+                            ),
+                        }
+                    );
```
tseho, line 99: What do you thing of extracting this in a separate function, something like
`notifyPermissionProviderError(provider.label)`

It would make the `handleConfirm()` content more easier to read.

tseho (commented):
You forgot the frontend tests on the handleConfirm.

src/Akeneo/Connectivity/Connection/front/src/connect/components/AppWizardWithSteps/AppWizardWithSteps.tsx
RiffVentura: Would you like having this instead :
Save would throw an error if anything goes south and returns a resolved void Promise if everything is fine
```javascript
    const handleConfirm = async () => {
        let userGroup = null;
        try {
            const confirmResult = await confirmAuthorization();
            userGroup = confirmResult.userGroup;
        } catch (e) {
            notify(
                NotificationLevel.ERROR,
                translate('akeneo_connectivity.connection.connect.apps.wizard.flash.error')
            );
            return;
        }

        for (const provider of providers) {
            try {
                await provider.save(userGroup, permissions[provider.key]);
            } catch {
                notifyPermissionProviderError(provider.label);
            }
        }

        notify(
            NotificationLevel.SUCCESS,
            translate('akeneo_connectivity.connection.connect.apps.wizard.flash.success')
        );

        redirectToMarketplace();
    };
```

tseho: That's better.

And for this part, what do you think of:
```
        try {
            const {userGroup} = await confirmAuthorization();
        } catch (e) {
            notify(
                NotificationLevel.ERROR,
                translate('akeneo_connectivity.connection.connect.apps.wizard.flash.error')
            );
            return;
        }
```
?

RiffVentura: won't it have issues with variable scoping (My IDE is not happy either) ?

tseho: That's sad, but I understand.

tseho (approved):

PR #15331, opened 2021-09-20
CXP-872: Add a default_permissions column to the user group table

tseho:

For the Activate App feature, we want to be able to configure default permissions to an user group (since each app will have a dedicated user group).

Basically, we add a new JSON column, `{[key: string]: boolean}` where different features in EE are able to add dynamically setup default values for their permissions.
Example:
```
{"attribute_group_view": true, "asset_familly_view": true}
```

See the PR in EE for examples: https://github.com/akeneo/pim-enterprise-dev/pull/11815/files

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/src/Akeneo/UserManagement/Component/Model/Group.php
+++ b/src/Akeneo/UserManagement/Component/Model/Group.php
@@ -21,6 +21,8 @@ class Group implements GroupInterface
     /** @var ArrayCollection */
     protected $roles;
 
+    protected ?array $defaultPermissions = null;
+
     /**
      * @param string $name [optional] Group name
      */
@@ -156,6 +158,16 @@ class Group implements GroupInterface
                 '$roles must be an instance of Doctrine\Common\Collections\Collection or an array'
             );
         }
+    }
+
+    public function getDefaultPermissions(): ?array
+    {
+        return $this->defaultPermissions;
+    }
+
+    public function setDefaultPermissions(array $defaultPermissions): void
+    {
+        $this->defaultPermissions = $defaultPermissions;
     }
 
     /**
--- a/src/Akeneo/UserManagement/Component/Model/Group.php
+++ b/src/Akeneo/UserManagement/Component/Model/Group.php
@@ -160,11 +160,17 @@ class Group implements GroupInterface
         }
     }
 
+    /**
+     * {@inheritdoc}
+     */
     public function getDefaultPermissions(): ?array
     {
         return $this->defaultPermissions;
     }
 
+    /**
+     * {@inheritdoc}
+     */
     public function setDefaultPermissions(array $defaultPermissions): void
     {
         $this->defaultPermissions = $defaultPermissions;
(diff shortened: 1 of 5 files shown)


src/Akeneo/UserManagement/Component/Model/Group.php
```diff
@@ -158,6 +160,16 @@ public function setRoles($roles): void
         }
     }
 
+    public function getDefaultPermissions(): ?array
+    {
+        return $this->defaultPermissions;
+    }
+
+    public function setDefaultPermissions(array $defaultPermissions): void
```
masacc, line 168: can't be set to null ?

src/Akeneo/UserManagement/Component/Model/GroupInterface.php
```diff
@@ -80,4 +80,8 @@ public function removeRole($role): void;
      * @throws \InvalidArgumentException
      */
     public function setRoles($roles): void;
+
+    public function getDefaultPermissions(): ?array;
+
+    public function setDefaultPermissions(array $defaultPermissions): void;
```
masacc, line 86: can you type-hint the array in phpdoc in these 2 functions ?

masacc (approved):

src/Akeneo/UserManagement/Component/Model/Group.php
tseho: Why would I want to set it to null from the code ?

masacc: I don't know why but as it can be nullable, I asked if you made a mistake

tseho: Oh ok, yes, I did on purpose.
The column has to be nullable, you cannot enforce a `NOT NULL` on a JSON column. (performances)
To reflect that, when the entity is hydrated from the sql result, it can have `null` or a deserialized json array, I have to mark the property as `array|null`.
This is it for the nullable.

Regarding the setter, I thought it would be better to enforce an array, I don't think of a case where you would want to destroy the default permissions by passing null.
I'd rather be safe than sorry :)

tests/back/UserManagement/Integration/Bundle/Export/ExportUserGroupIntegration.php
```diff
@@ -46,10 +46,10 @@ public function it_exports_user_groups_in_csv(): void
     {
         $expectedCsv = <<<CSV
 name
-All
 "IT support"
 Manager
 Redactor
+All
```
tseho, line 52: I'm not sure why this test suddenly broke, I mean, the roles are inserted like this in the fixtures: `src/Akeneo/Platform/Bundle/InstallerBundle/Resources/fixtures/minimal/user_groups.csv`

RiffVentura (approved):

PR #15356, opened 2021-09-22
CPM-371: Add context in GetAdditionalProperties* contracts

mmetayer:

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc


src/Akeneo/Pim/Enrichment/Bundle/Resources/config/queries.yml
```diff
@@ -410,3 +410,8 @@ services:
     Akeneo\Pim\Enrichment\Bundle\Storage\Sql\Category\CountTreesChildren:
         arguments:
             - '@database_connection'
+
+    Akeneo\Pim\Enrichment\Component\Product\Query\PublicApi\GetRawValues:
```
ahocquard, line 414: This is not part of the public API: it's only for internal usage in the Enrichment context. Persistence format should not be exposed outside.

nmarniesse: We have a few places where we need raw values, for indexation for instance (for DQI and now for the table attribute in GE).  
We plug the logic with the `GetAdditionalPropertiesForProductProjectionInterface::fromProductIdentifiers(array $productIdentifiers): array`

We can load the entire products and build our projection using the `getValues()` method, but it's a very long process. For performance we fetch only the raw values and filter them to have only table attribute values. It's way faster.

If we want to keep performance, another solution is to replaced this public API by something like: `GetReadValueCollections::forProductIdentifiersFilteredByAttributeTypes(array $productIdentifiers, array $attributeTypes): array;` that return ReadValueCollection indexed by product identifiers.  

What do you think? Do you have a better idea?

ahocquard: Exposing an internal storage format is at the opposite of the philosophy of the hexagonal architecture, and exposing it in PublicAPI is not recommended at all (DQI have some coupling issues so I would not design necessary with that in mind).

We can discuss it all together, the double JSON fetching when indexing bother me much more in term of performance.

nmarniesse: I push a new version where we put the value collections in a context. This way we don't need to fetch again

pierallard (approved):

pierallard (approved):

mmetayer (commented):
GTM

nmarniesse (approved):

PR #15372, opened 2021-09-24
Run cypress in prod mode

jmleroux:
Running Cypress over a PIM in dev has two drawbacks:
- the app is slower
- we can have issues with selector because of the debug bar when targeting buttons for example

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -301,11 +301,11 @@ jobs:
           name: Start containers
           command: |
             docker load -i php-pim-image.tar
-            APP_ENV=dev C='fpm mysql elasticsearch httpd object-storage pubsub-emulator' make up
+            APP_ENV=prod C='fpm mysql elasticsearch httpd object-storage pubsub-emulator' make up
             docker/wait_docker_up.sh
       - run:
           name: Install database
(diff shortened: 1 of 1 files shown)


SamirBoulil (approved):

tseho:
Are you sure that starting the application is prod mode is a good idea in the CI ?
I think it's working in CE but not at all in EE, since `prod` in EE means Serenity.

Regarding the debug bar, I think we should fix this config file:
https://github.com/akeneo/pim-community-dev/blob/master/config/packages/dev/framework.yml
like this:
```
framework:
    profiler:
        enabled: '%env(bool:APP_DEBUG)%'
```

jmleroux:
Thiis is why I made the PR on CE only. We already have solutions to run Cypress on Serenity, but I dont't knowyet if we should use them. The speed difference between prod and dev mode can be significant on the CI for the Cypress tests stability and avoid timeouts.

BitOne:
The real subject is "how to run in prod mode outside of prod", if that makes sense.
The main issue we have is Minio and the obligation to use a S3 adapter for Flysystem. If we can use a GCS emulator, then it would be possible to use prod mode locally and on the CI.

olivierpontier (approved):

PR #15390, opened 2021-09-28
CXP-891 Updated confirm authorization endpoint

RiffVentura:
Updated confirm authorization endpoint

The App Wizard requires a user group name to save permissions via form providers
So these changes updates Confirm authorization endpoint to return an id as well as user group name (that was created for the app)

### DIFF
--- a/src/Akeneo/Connectivity/Connection/back/Domain/Apps/Persistence/Query/GetConnectedAppIdAndUserGroupQueryInterface.php
+++ b/src/Akeneo/Connectivity/Connection/back/Domain/Apps/Persistence/Query/GetConnectedAppIdAndUserGroupQueryInterface.php
@@ -1,0 +1,17 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Akeneo\Connectivity\Connection\Domain\Apps\Persistence\Query;
+
+/**
+ * @copyright 2021 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+interface GetConnectedAppIdAndUserGroupQueryInterface
+{
+    /**
+     * @return array{appId:string, userGroup:string}|null $appIdAndUserGroup
+     */
+    public function execute(string $appPublicId): ?array;
+}
--- a/src/Akeneo/Connectivity/Connection/back/Domain/Apps/Persistence/Query/GetConnectedAppIdAndUserGroupQueryInterface.php
+++ b/src/Akeneo/Connectivity/Connection/back/Domain/Apps/Persistence/Query/GetConnectedAppIdAndUserGroupQueryInterface.php
@@ -11,7 +11,7 @@ namespace Akeneo\Connectivity\Connection\Domain\Apps\Persistence\Query;
 interface GetConnectedAppIdAndUserGroupQueryInterface
 {
     /**
-     * @return array{appId:string, userGroup:string}|null $appIdAndUserGroup
+     * @return array{appId:string, userGroup:string}|null $marketplaceAppId
      */
-    public function execute(string $appPublicId): ?array;
+    public function execute(string $marketplaceAppId): ?array;
 }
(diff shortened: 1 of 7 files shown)


src/Akeneo/Connectivity/Connection/back/Domain/Apps/Persistence/Query/GetConnectedAppIdAndUserGroupQueryInterface.php
```diff
@@ -0,0 +1,17 @@
...
+/**
+ * @copyright 2021 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+interface GetConnectedAppIdAndUserGroupQueryInterface
+{
+    /**
+     * @return array{appId:string, userGroup:string}|null $appIdAndUserGroup
+     */
+    public function execute(string $appPublicId): ?array;
```
masacc, line 16: rename `$appPublicId` in `$marketplaceAppId` ?

src/Akeneo/Connectivity/Connection/back/Infrastructure/Apps/Controller/ConfirmAuthorizationAction.php
```diff
@@ -63,9 +62,9 @@ public function __invoke(Request $request, string $clientId): Response
             return new RedirectResponse('/');
         }
 
-        $app = $this->appRepository->findOneById($clientId);
-        if (null !== $app) {
-            return new JsonResponse(['appId' => $app->getId()]);
+        $connectedAppIdAndUserGroup = $this->getConnectedAppIdAndUserGroupQuery->execute($clientId);
```
masacc, line 65: rename `$clientId` in `$marketplaceAppId` ?

masacc (approved):

src/Akeneo/Connectivity/Connection/back/Infrastructure/Apps/Persistence/Query/GetConnectedAppIdAndUserGroupQuery.php
```diff
@@ -0,0 +1,41 @@
...
+    {
+        $selectQuery = <<<SQL
+SELECT acca.id, oag.name
+FROM pim_api_client pac
+JOIN akeneo_connectivity_connection acc on pac.id = acc.client_id
+JOIN akeneo_connectivity_connected_app acca on acc.code = acca.connection_code
+JOIN oro_user_access_group ouag on acc.user_id = ouag.user_id
+JOIN oro_access_group oag on ouag.group_id = oag.id
+WHERE pac.marketplace_public_app_id = :id AND oag.name != 'All'
+SQL;
```
tseho, line 33: Can you remove the aliases ? I think it's harder to read with it.

src/Akeneo/Connectivity/Connection/back/Infrastructure/Apps/Persistence/Query/GetConnectedAppIdAndUserGroupQuery.php
```diff
@@ -0,0 +1,41 @@
...
+        $resultRow = $this->connection->executeQuery($selectQuery, ['id' => $appPublicId])->fetch() ?: null;
```
tseho, line 34:
```suggestion
        $stmt = $this->connection->executeQuery($selectQuery, ['id' => $appPublicId]);

        if (false === $row = $stmt->fetch()) {
          return null;
        }

        return $row;
```
What do you think ?
Also, instead of mapping the keys, I usually rename the columns directly in my query.
Like `SELECT id AS appId, ...`
It's not required, only a suggestion.

src/Akeneo/Connectivity/Connection/back/Infrastructure/Apps/Controller/ConfirmAuthorizationAction.php
RiffVentura: on second thought I would gladly refactor it but in another pull request

tseho: I love refactoring but be careful not to do it too much or too early, we have so much to do still :)
(of course, if it's blocking for new developments, it's always smart to start there)

RiffVentura: I think renaming here is a bit bigger that we expect : renaming here implies changing routing parameter and also front end as well

src/Akeneo/Connectivity/Connection/back/Domain/Apps/Persistence/Query/GetConnectedAppIdAndUserGroupQueryInterface.php
RiffVentura: Good question, I was unsure what is the proper naming here
Since we save it in the client table as `marketplace_public_app_id` shouldn't we stick with it through out the rest of the code ?

tseho: I'm good with `$marketplaceAppId`

tseho (approved):

PR #15392, opened 2021-09-28
CXP-886: Display the Permissions tab

masacc:

In EE, if the user has the role `Manage apps`, after an App has been activated, he can access to a tab displaying permissions of the App. In order to display the tab only in EE, we base the condition onto whether or not the count of registry providers is greater than 0.

image

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/src/Akeneo/Connectivity/Connection/workspaces/permission-form/src/registry/PermissionFormRegistry.ts
+++ b/src/Akeneo/Connectivity/Connection/workspaces/permission-form/src/registry/PermissionFormRegistry.ts
@@ -40,6 +40,11 @@ const PermissionFormRegistry = {
             })
         );
     },
+    countProviders: (): number => {
+        const providers = _config.providers || [];
+
+        return Object.keys(providers).length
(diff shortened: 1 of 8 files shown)


src/Akeneo/Connectivity/Connection/workspaces/permission-form/src/registry/PermissionFormRegistry.ts
```diff
@@ -41,6 +41,11 @@ const PermissionFormRegistry = {
             })
         );
     },
+    countProviders: (): number => {
```
tseho, line 44:
```suggestion
    count: (): number => {
```
I would go for a simpler method name.

masacc: why ? I'd rather be explicit when I name something to avoid confusion.

PR #15414, opened 2021-10-04
CXP-887: In Connected App table, add a FK to its dedicated user group

masacc:

Connected App model will need to perform some request to fetch its permissions which depend on its user group. For that purpose, add a direct dependency between the connected app and its user group.

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc


masacc:
> Can you update the query `GetConnectedAppIdAndUserGroupQuery` to use this FK ?

Done

PR #15442, opened 2021-10-13
RAC-870: Extract microfrontend to a workspace

StevenVAIDIE:

In this PR :
- Move front-packages into workspace in order to mutualize node_modules. To do this we needed to share the devDependencies of each front-packages
- Launch prettier with the new version of prettier
- Fix tests that use userEvent with option {allAtOnce: true} this option is not supported anymore to be more closer to real usage
- Remove @testing-library/dom because @testing-library/user-event didn't use @testing-library/react by default if @testing-library/dom is installed, we didn't need to use act now (@testing-library/user-event wrap click with act by default) 
- Remove some act not needed anymore since we removed @testing-library/dom (other will be removed in another PR)

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/std-build/package.json
+++ b/std-build/package.json
@@ -18,6 +18,8 @@
   },
   "workspaces": [
     "vendor/akeneo/pim-community-dev",
+    "vendor/akeneo/pim-community-dev/front-packages/shared",
+    "vendor/akeneo/pim-community-dev/front-packages/measurement",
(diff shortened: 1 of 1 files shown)


tseho:
What happened with those jest tests in Connectivity ?
the upgrade of `@testing-library/user-event` broke them ?

PR #15446, opened 2021-10-13
CXP-811: enforce Product/ProductModel ACL on the API

tseho:

Our production monitoring tells us that there is no active connections trying to fetch products without the correct ACL.
We can safely apply the ACL.
I've kept the logging part for now, to be able to easily monitor if customers encounters the problem in the next few weeks.

- [x] Tests
- [ ] PM Validation (Story)
- [x] Tech Doc https://github.com/akeneo/pim-api-docs/pull/527


willy-ahva:
Did we write some documentation about this new ACL? If no, we should

tseho:
> Did we write some documentation about this new ACL? If no, we should
good point :+1: 
I will create the api documentation PR right now.

PR #15451, opened 2021-10-14
BH-794: fix delete_pr_environment job in CI

Airmanbzh:

fix delete_pr_environment job in CI

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc


jmleroux (changes requested):
I think we must have the same options everywhere or at least identify and explain the side effects. Adapting these specific tests to make them green feels like a hack to me.

PR #15573, opened 2021-10-28
CPM-400: fix notification check for behat

nmarniesse:

Sometimes on behat test the page is not correctly displayed. For instance the right side is truncated and we can't access the notification button/menu on the right.
It's because the text is too long:

image

Now with the fix the text is truncated:

image

- [x] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc


pierallard (approved):
:viandecss:

ahocquard (approved):
:smile: :tada:

PR #15593, opened 2021-10-29
CXP-938: Fix issues with Activate OAuth

tseho:
**Issues:**

- With the previous version, an expiration date was attributed to the token with `time() + null`, which mean that the token was expiring immediately ...
- When using fake apps, don't check the code challenge in the marketplace
- When validating a token request, don't call the marketplace if any of `code_challenge` & `code_identifier` are empty, it will fail anyway on the `NotBlank`.
- the endpoint for requesting a token should be `POST`, not `GET`

**Bonus:**

- I've implemented the last step of the behat scenario
- When an app is already authorized, skip the wizard and redirect directly with the Authorization Code. *This is a workaround waiting for the proper re-activation feature*.

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/src/Akeneo/Connectivity/Connection/back/Infrastructure/Apps/Validation/CodeChallengeMustBeValidValidator.php
+++ b/src/Akeneo/Connectivity/Connection/back/Infrastructure/Apps/Validation/CodeChallengeMustBeValidValidator.php
@@ -6,6 +6,7 @@ namespace Akeneo\Connectivity\Connection\Infrastructure\Apps\Validation;
 
 use Akeneo\Connectivity\Connection\Domain\Apps\DTO\AccessTokenRequest;
 use Akeneo\Connectivity\Connection\Infrastructure\Marketplace\WebMarketplaceApiInterface;
+use Akeneo\Platform\Bundle\FeatureFlagBundle\FeatureFlag;
 use Symfony\Component\Validator\Constraint;
 use Symfony\Component\Validator\ConstraintValidator;
 use Symfony\Component\Validator\Exception\UnexpectedTypeException;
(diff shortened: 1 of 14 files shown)


src/Akeneo/Connectivity/Connection/back/tests/Unit/spec/Infrastructure/Apps/Validation/CodeChallengeMustBeValidValidatorSpec.php
```diff
@@ -67,6 +70,23 @@ public function it_validates_that_the_code_challenge_is_valid(
         $this->validate($value, $constraint);
     }
 
+    public function it_skip_the_validator_if_a_value_is_empty(
```
willy-ahva, line 73:
```suggestion
    public function it_skips_the_validator_if_a_value_is_empty(
```

willy-ahva (approved):

src/Akeneo/Connectivity/Connection/back/Infrastructure/Apps/Validation/CodeChallengeMustBeValidValidator.php
```diff
@@ -39,6 +44,17 @@ public function validate($value, Constraint $constraint)
             );
         }
 
+        if (true === $this->fakeAppsFeatureFlag->isEnabled()) {
+            return;
+        }
+
+        if (empty($value->getClientId())
```
RiffVentura, line 51: I am failing to grasp the meaning behind this condition
Can you elaborate please ?  :)

RiffVentura (approved):

src/Akeneo/Connectivity/Connection/back/Infrastructure/Apps/Validation/CodeChallengeMustBeValidValidator.php
tseho: When the DTO is validated, all Constraints are executed.
If you call the endpoint without any parameter, all Constraints are called, the ones with `NotBlank` will fail on missing parameters but this one was calling the marketplace, even when the parameters are missing.

It was requesting `marketplace/api/1.0/app//challenge?code_identifier=&code_challenge=` which is, IMHO, totally useless.

By skiping the validator when one of the value is empty, the request still fail because one parameter is missing but we didn't call the marketplace API unnecessary.

RiffVentura: :goat:

PR #15609, opened 2021-11-02
PIM-10133 - PIM-10134: Prevent deletion of a channel used in a published product/shared catalog export job

nmarniesse:

- Backport of https://github.com/akeneo/pim-community-dev/pull/15472  
- prevent deletion of a channel used in a published product export

- [x] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/tests/back/Channel/Integration/Channel/Query/Sql/IsChannelUsedInProductProductExportJobIntegration.php
+++ b/tests/back/Channel/Integration/Channel/Query/Sql/IsChannelUsedInProductProductExportJobIntegration.php
@@ -29,13 +29,29 @@ class IsChannelUsedInProductProductExportJobIntegration extends TestCase
         $this->assertTrue($isChannelUsedInProductProductExportJob->execute('mobile'));
     }
 
-    private function givenAProductImportJob(): void
+    public function test_it_determines_if_a_channel_is_used_in_published_product_export_jobs(): void
+    {
+        $isChannelUsedInProductProductExportJob = $this->get(IsChannelUsedInProductProductExportJob::class);
+
+        $this->assertFalse($isChannelUsedInProductProductExportJob->execute('mobile'));
+
+        $this->givenAProductImportJob('csv_published_product_import');
+        $this->assertFalse($isChannelUsedInProductProductExportJob->execute('mobile'));
+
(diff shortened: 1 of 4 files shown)


tests/back/Channel/Integration/Channel/Query/Sql/IsChannelUsedInProductProductExportJobIntegration.php
```diff
@@ -29,13 +29,29 @@ public function test_it_determines_if_a_channel_is_used_in_product_export_jobs()
         $this->assertTrue($isChannelUsedInProductProductExportJob->execute('mobile'));
     }
 
-    private function givenAProductImportJob(): void
+    public function test_it_determines_if_a_channel_is_used_in_published_product_export_jobs(): void
```
nmarniesse, line 32: Maybe strange to have this test in CE as published product is a EE feature. I can move it in EE (with the service declaration), but I think it's simpler like that

mmetayer: no strong opinion on this. It does feel weird to have an EE job name here, but at the same time it allows us not to override the service definition in EE, thus preventing us from any mistake in the future (e.g adding a restricted job name in CE and forgetting to add it in the EE service definition override as well)

mmetayer (approved):

tests/back/Channel/Integration/Channel/Query/Sql/IsChannelUsedInProductProductExportJobIntegration.php
ahocquard: The most problematic issue is the coupling it creates. It should prevent it at validation step, through a registry asking "can my channel be deleted"?

Published is coupled to channel, not the opposite. And the test would be logically inside published product.

ahocquard:
Could we create a card for this technical debt as it creates more and more coupling? Thanks!

nmarniesse:
> Could we create a card for this technical debt as it creates more and more coupling? Thanks!

Done https://akeneo.atlassian.net/browse/TIP-1551

LaurentPetard (approved):

RiffVentura (approved):

PR #15613, opened 2021-11-02
RAC-945: Add endpoint to search job execution

StevenVAIDIE:

In this PR : 
- We added a controller/handler/query that search job executions
- We remove the fact that IntegrationTest load by default Fixture, now tests need to call it explicitly to avoid confusion when reading the test.

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc


ValentinMumble (approved):

src/Akeneo/Platform/Job/back/tests/Acceptance/Application/SearchJobExecution/SearchJobExecutionHandlerTest.php
```diff
@@ -0,0 +1,35 @@
...
+{
+    protected function setUp(): void
+    {
+        static::bootKernel(['debug' => false]);
+    }
+
+    /**
+     * @test
+     */
+    public function it_returns_the_search_result()
```
SamirBoulil, line 22: Missing case where it returns actual non empty job execution table ?

src/Akeneo/Platform/Job/back/tests/Acceptance/FakeServices/InMemorySearchJobExecution.php
```diff
@@ -0,0 +1,41 @@
...
+class InMemorySearchJobExecution implements SearchJobExecutionInterface
+{
+    private InMemoryJobExecutionRepository $jobExecutionRepository;
+
+    public function __construct(InMemoryJobExecutionRepository $jobExecutionRepository)
+    {
+        $this->jobExecutionRepository = $jobExecutionRepository;
+    }
+
+    public function search(SearchJobExecutionQuery $query): array
```
SamirBoulil, line 20: That works, but re-implementing an in memory version of the search that will never ever be used in production it a bit overkill to me.
i'd rather stub this class or use a dummy to keep the tests really simple.

(The risk here is that we add more and more complexity to handle different sort/usecases where it does not add any business value and is complex in the end, i don't know if i'm clear ^^)

SamirBoulil (approved):

src/Akeneo/Platform/Job/back/tests/Acceptance/FakeServices/InMemorySearchJobExecution.php
StevenVAIDIE: Should we implemented something like that ? 
src/Akeneo/AssetManager/tests/back/Common/Fake/InMemoryFindAssetIdentifiersForQuery.php
src/Akeneo/AssetManager/tests/front/integration/responses/Asset/Search/code_filtered_with_not_in_operator.json

PR #15620, opened 2021-11-03
Assert that the value is correctly set in the field

tseho:

With a test like this:
```
Given I change the "Weather conditions" to "Hot, Cold, Dry, Wet"
And I save the product
```
The select2 component is updated by triggering js events.
Sadly, the save button is then clicked before all the changes are propagated to the form state stored in backbone.
That's why only a part of the selected values are saved and we randomly have the following error: `Expected product field "Weather conditions" to contain "cold, dry, hot, wet", but got "dry, hot, wet".`

I'm proposing that each time we fill a field in the PEF, we spin until the value is as expected.
It should prevent this race condition.

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/tests/legacy/features/Context/Page/Base/ProductEditForm.php
+++ b/tests/legacy/features/Context/Page/Base/ProductEditForm.php
@@ -553,6 +553,86 @@ class ProductEditForm extends Form
     }
 
     /**
+     * Checks if the specified field is set to the expected value, raises an exception if not
+     *
+     * @throws ExpectationException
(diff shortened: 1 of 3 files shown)


mmetayer (approved):

willy-ahva (approved):

ahocquard (approved):
Thanks for the fix :+1:

PR #15626, opened 2021-11-03
BH-859: Use Symfony cache for Oro ACL cache

jmleroux:
Doctrine\Common\Cache\FilesystemCache is deprecated (as the whole doctrine/cache package) and is not available with PHP 8

### DIFF
--- a/config/packages/framework.yml
+++ b/config/packages/framework.yml
@@ -23,5 +23,7 @@ framework:
                 version_strategy: pim_enrich.version_strategy
     cache:
         pools:
-            name:
+            pim_apcu_cache:
                 adapter: cache.adapter.apcu
+            oro_acl_cache:
+                adapter: cache.adapter.filesystem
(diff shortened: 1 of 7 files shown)


config/packages/framework.yml
```diff
@@ -23,5 +23,7 @@ framework:
                 version_strategy: pim_enrich.version_strategy
     cache:
         pools:
-            name:
+            pim_apcu_cache:
```
jmleroux, line 26: This pool name is not used, but I think it's better to have something more explicit

config/packages/framework.yml
```diff
@@ -23,5 +23,7 @@ framework:
...
                 adapter: cache.adapter.apcu
+            oro_acl_cache:
+                adapter: cache.adapter.filesystem
```
jmleroux, line 29: This new pool will be the one used by ACLs

nmarniesse (approved):

jmleroux:
check frozen

mmetayer (approved):

PR #15655, opened 2021-11-05
BH-859: Use lazy commands

jmleroux:
There are several avantages to use lazy commands. From a performance point of vue, I've traced **a lot** of services instanciated just to regsiter commands.

I create this PR because I've run into an issue when warming up the application without a database container up. One of the instantiated commands triggered a listener trying to make a database call and then resulting in an error when I had no use of this call (only seen in APP_DEBUG=1 mode).

From https://symfony.com/blog/new-in-symfony-3-4-lazy-commands:

> the Symfony Console suffered a shortcoming since day one: you must instantiate all commands to register them in the console application. The reason is that even the command name itself is defined inside a method called configure(), so you must instantiate the command class to configure it and get the command name.
>
> In practice, this shortcoming is not a real problem for lots of applications. However, if you have lots of commands or they are very complex (because they make use of a lot of services) this can hurt performance. Moreover, if there is a single error when instantiating any of the commands, you can't run the application console, even if you are not executing the failing command.
>
> In Symfony 3.4, we were finally able to fix this issue and console commands can now be lazy. This means that commands no longer need to be instantiated when running the console application, so executing any command will be a bit faster in Symfony 3.4.

Then in https://symfony.com/blog/new-in-symfony-5-3-lazy-command-description:

> Symfony commands aren’t fully lazy. If you run bin/console without arguments (or the equivalent bin/console list command) you see the entire list of commands registered in the application and their descriptions. This requires instantiating all commands, because their description is not lazy.
>
> You might think that this is not important, because you rarely list commands and it’s OK if it takes a few milliseconds to run. That’s true, but this makes it harder/impossible to implement another feature that we want to include in future Symfony versions: console autocompletion.

https://symfony.com/doc/current/service_container.html#service-container-services-load-example

### DIFF
--- a/src/Akeneo/Pim/Automation/DataQualityInsights/back/Infrastructure/Symfony/Command/ConsolidateDashboardRatesCommand.php
+++ b/src/Akeneo/Pim/Automation/DataQualityInsights/back/Infrastructure/Symfony/Command/ConsolidateDashboardRatesCommand.php
@@ -17,8 +17,10 @@ use Symfony\Component\Console\Output\OutputInterface;
  */
 final class ConsolidateDashboardRatesCommand extends Command
 {
-    /** @var ConsolidateDashboardRates */
-    private $consolidateDashboardRates;
+    protected static $defaultName = 'pim:data-quality-insights:consolidate-dashboard-rates';
+    protected static $defaultDescription = 'Consolidate the Data-Quality-Insights dashboard rates.';
+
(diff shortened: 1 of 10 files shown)


ValentinMumble (approved):

StevenVAIDIE (approved):
Same here but i think it desired

jmleroux:
> Same here but i think it desired

??

PR #15661, opened 2021-11-07
BH-859: Fix Doctrine deprecations for migrations

jmleroux:


upgrades/test_schema/Version_4_0_20200117145507_remove_compute_model_descendant_jobs_Integration.php
```diff
@@ -124,7 +124,7 @@ protected function getComputeProductModelDescendantJobId(): ?int
     {
         $sql = "SELECT id FROM akeneo_batch_job_instance WHERE code = 'compute_product_models_descendants'";
 
-        $jobId = $this->getConnection()->executeQuery($sql)->fetchAll(\PDO::FETCH_COLUMN, 0)[0] ?? null;
+        $jobId = $this->getConnection()->executeQuery($sql)->fetchFirstColumn()[0] ?? null;
```
StevenVAIDIE, line 127: No sure but it's not :

```suggestion
        $jobId = $this->getConnection()->executeQuery($sql)->fetchOne() ?? null;
```

StevenVAIDIE (approved):

upgrades/test_schema/Version_4_0_20200117145507_remove_compute_model_descendant_jobs_Integration.php
jmleroux: Nope. fetchOne would replace a fetch(Mode::Column). Here it is a fetchALL :wink: 

At least it shows you're making a real review, thank you! :pray:

jmleroux: And I didn't see you removed the [0] indexed access just after. In this case, you may be totally right!

jmleroux: Both should work, but your way is better. I will not change it because don't want to run the CI and there are already so much improvement/cleaning to do in this corner of the PIM....

Airmanbzh (approved):

PR #15695, opened 2021-11-10
RAC-975: Backend service to query job executions should support sorting on each columns

ghoussard:

Handle sort column and directions in query to be able to sort all columns of job execution table.

- [x] Tests

### DIFF
--- a/src/Akeneo/Platform/Job/back/Infrastructure/Query/SearchJobExecution.php
+++ b/src/Akeneo/Platform/Job/back/Infrastructure/Query/SearchJobExecution.php
@@ -11,7 +11,6 @@ use Akeneo\Tool\Component\Batch\Job\BatchStatus;
 use Doctrine\DBAL\Connection;
 use Doctrine\DBAL\Types\Type;
 use Doctrine\DBAL\Types\Types;
-use InvalidArgumentException;
 
 /**
  * @author Grégoire Houssard <gregoire.houssard@akeneo.com>
@@ -136,7 +135,7 @@ SQL;
         $sortDirection = $query->sortDirection;
 
         if (!in_array($sortDirection, ['ASC', 'DESC'])) {
-            throw new InvalidArgumentException(sprintf('Sort direction "%s" is not supported', $query->sortDirection));
+            throw new \InvalidArgumentException(sprintf('Sort direction "%s" is not supported', $query->sortDirection));
         }
 
         switch ($query->sortColumn) {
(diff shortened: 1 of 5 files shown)


src/Akeneo/Platform/Job/back/Infrastructure/Query/SearchJobExecution.php
```diff
@@ -44,12 +45,13 @@ public function search(SearchJobExecutionQuery $query): array
     LEFT JOIN akeneo_batch_step_execution se on je.id = se.job_execution_id
     %s
     GROUP BY je.id, ji.label, ji.type, je.start_time, je.user, je.status
-    ORDER BY ISNULL(je.start_time) DESC, je.start_time DESC
+    %s
```
ahocquard, line 48: This query will count all job execution steps for all job instances, before returning the X elements. Indeed, the limit offset is executed only at the very end.

I would do something like (not tested, pseudo code):
```
with job_executions as (
    select ....
    from akeneo_batch_job_execution je  
          JOIN akeneo_batch_job_instance ji on je.job_instance_id = ji.id
     %s -- where
     %s -- order by  
     limit :offset, :limit; 
)

  select 
       job_executions.* , 
       count(se.job_execution_id) as total_warning,
       json_merge(
            json_arrayagg(ifnull(se.failure_exceptions, 'a:0:{}')), 
            json_arrayagg(ifnull(se.errors, 'a:0:{}'))
      ) as errors
  from job_executions 
       left join akeneo_batch_step_execution se on je.id = se.job_execution_id
 group by job_executions.id
```

ahocquard (approved):

src/Akeneo/Platform/Job/back/Infrastructure/Query/SearchJobExecution.php
ghoussard: Yep, it can be more efficient to do that but we won't be able to sort by warning count because the order by clause cannot includes the aggregate warning count of step_executions ?

src/Akeneo/Platform/Job/back/Infrastructure/Query/SearchJobExecution.php
```diff
@@ -129,6 +131,40 @@ private function buildQueryParamsTypes(): array
         ];
     }
 
+    private function buildSqlOrderByPart(SearchJobExecutionQuery $query): string
+    {
+        $sortDirection = $query->sortDirection;
+
+        if (!in_array($sortDirection, ['ASC', 'DESC'])) {
+            throw new InvalidArgumentException(sprintf('Sort direction "%s" is not supported', $query->sortDirection));
```
ValentinMumble, line 139:
```suggestion
            throw new \InvalidArgumentException(sprintf('Sort direction "%s" is not supported', $query->sortDirection));
```
And remove it from phpcd

ValentinMumble (approved):

src/Akeneo/Platform/Job/back/Infrastructure/Query/SearchJobExecution.php
ahocquard: Yes indeed :+1:  On the other side, there are more than 1 million job executions for some customers, so it will not work. Better to discuss about it with the team and the PM to explain the (existing) issue.

PR #15724, opened 2021-11-17
RAC-973: Process tracker filters should be persisted in the local storage

ghoussard:

Filter state are now persisted in browser local storage.
The page is not keep cause search bar debounce useEffect but it better because it will avoid failure when job are purged between filter appliance and page reload.


juliensnz (approved):
Wow such nice hooks 👀

pierrejolly (approved):

StevenVAIDIE (approved):

PR #15789, opened 2021-11-25
PIM-10098: Introduce a command/handler to remove a product model

nmarniesse:

With the former way, some subscribers listen the `PRE_SAVE` event in order to check if the product model can be removed or not. If not, an exception is thrown. This exception cannot be caught easily, especially when the listener is in EE. So the HttpKernel catch the exception and format a response. There are some problems with this way to do:
- where do we translate the message? See [PIM-10098](https://akeneo.atlassian.net/browse/PIM-10098)
- only one error message can be returned, it's the first exception encounters

With a command/handler the 2 problems are solved because the we validate the command. That means:
- The validations can be added in any bundle
- Translations are done by the validation mechanism
- All validation messages can be returned to the user at once

- [x] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/src/Akeneo/Pim/Enrichment/Bundle/Controller/ExternalApi/ProductModelController.php
+++ b/src/Akeneo/Pim/Enrichment/Bundle/Controller/ExternalApi/ProductModelController.php
@@ -260,7 +260,7 @@ class ProductModelController
 
         $command = new RemoveProductModelCommand($productModel->getCode());
         $violations = $this->validator->validate($command);
-        if (0 < count($violations)) {
+        if (0 < \count($violations)) {
             return new JsonResponse([
                 'code' => Response::HTTP_UNPROCESSABLE_ENTITY,
                 'messages' => \array_map(
--- a/src/Akeneo/Pim/Enrichment/Bundle/Controller/ExternalApi/ProductModelController.php
+++ b/src/Akeneo/Pim/Enrichment/Bundle/Controller/ExternalApi/ProductModelController.php
@@ -5,6 +5,8 @@ declare(strict_types=1);
 namespace Akeneo\Pim\Enrichment\Bundle\Controller\ExternalApi;
 
 use Akeneo\Pim\Enrichment\Bundle\EventSubscriber\ProductModel\OnSave\ApiAggregatorForProductModelPostSaveEventSubscriber;
+use Akeneo\Pim\Enrichment\Component\Command\ProductModel\RemoveProductModelCommand;
+use Akeneo\Pim\Enrichment\Component\Command\ProductModel\RemoveProductModelHandler;
 use Akeneo\Pim\Enrichment\Component\Product\Connector\ReadModel\ConnectorProductModelList;
 use Akeneo\Pim\Enrichment\Component\Product\Connector\UseCase\ListProductModelsQuery;
 use Akeneo\Pim\Enrichment\Component\Product\Connector\UseCase\ListProductModelsQueryHandler;
@@ -26,7 +28,6 @@ use Akeneo\Tool\Component\Api\Pagination\PaginatorInterface;
 use Akeneo\Tool\Component\Api\Security\PrimaryKeyEncrypter;
 use Akeneo\Tool\Component\StorageUtils\Exception\PropertyException;
 use Akeneo\Tool\Component\StorageUtils\Factory\SimpleFactoryInterface;
-use Akeneo\Tool\Component\StorageUtils\Remover\RemoverInterface;
 use Akeneo\Tool\Component\StorageUtils\Repository\IdentifiableObjectRepositoryInterface;
 use Akeneo\Tool\Component\StorageUtils\Saver\SaverInterface;
 use Akeneo\Tool\Component\StorageUtils\Updater\ObjectUpdaterInterface;
@@ -47,6 +48,7 @@ use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
 use Symfony\Component\Security\Core\Authentication\Token\Storage\TokenStorageInterface;
 use Symfony\Component\Security\Core\Exception\InvalidArgumentException;
 use Symfony\Component\Serializer\Normalizer\NormalizerInterface;
+use Symfony\Component\Validator\ConstraintViolationInterface;
 use Symfony\Component\Validator\Validator\ValidatorInterface;
 use Webmozart\Assert\Assert;
 
@@ -83,7 +85,8 @@ class ProductModelController
     private LoggerInterface $logger;
     private LoggerInterface $apiProductAclLogger;
     private SecurityFacade $security;
-    private RemoverInterface $productModelRemover;
+    private RemoveProductModelHandler $removeProductModelHandler;
+    private ValidatorInterface $validator;
 
     public function __construct(
         ProductQueryBuilderFactoryInterface $pqbFactory,
@@ -112,7 +115,8 @@ class ProductModelController
         array $apiConfiguration,
         LoggerInterface $apiProductAclLogger,
         SecurityFacade $security,
-        RemoverInterface $productModelRemover
+        RemoveProductModelHandler $removeProductModelHandler,
+        ValidatorInterface $validator
     ) {
         $this->pqbFactory = $pqbFactory;
         $this->pqbSearchAfterFactory = $pqbSearchAfterFactory;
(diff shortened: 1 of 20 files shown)


src/Akeneo/Pim/Enrichment/Component/Product/Command/ProductModel/RemoveProductModelHandler.php
```diff
@@ -0,0 +1,44 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of the Akeneo PIM Enterprise Edition.
```
LaurentPetard, line 6: It's not the good header, it's the EE one (same for the other added files)

LaurentPetard (approved):

src/Akeneo/Pim/Enrichment/Bundle/Controller/ExternalApi/ProductModelController.php
```diff
@@ -253,7 +258,20 @@ public function deleteAction(string $code): JsonResponse
...
         }
 
-        $this->productModelRemover->remove($productModel);
+        $command = new RemoveProductModelCommand($productModel->getCode());
+        $violations = $this->validator->validate($command);
+        if (0 < \count($violations)) {
+            return new JsonResponse([
+                'code' => Response::HTTP_UNPROCESSABLE_ENTITY,
+                'messages' => \array_map(
+                    fn (ConstraintViolationInterface $violation): string => $violation->getMessage(),
```
rfauglas, line 267: you mention internationalization in pr, but here it is not the case... IMO we could factor this code with adequate Exception handling, instead of replicating in all controllers......

src/Akeneo/Pim/Enrichment/Component/Product/Command/ProductModel/RemoveProductModelCommand.php
```diff
@@ -0,0 +1,24 @@
...
+
+/**
+ * @copyright 2021 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+final class RemoveProductModelCommand
+{
+    private string $productModelCode;
+
+    public function __construct(string $productModelCode)
```
rfauglas, line 15: I'm not sure if we should use functional keys in commands and API...

src/Akeneo/Pim/Enrichment/Component/Product/Job/DeleteProductsAndProductModelsTasklet.php
```diff
@@ -183,19 +190,52 @@ protected function doDelete(array $entities): void
...
+            } else {
+                $skippedProductModelCount++;
+                foreach ($violations as $violation) {
+                    $this->stepExecution->addWarning($violation->getMessage(), [], new DataInvalidItem($productModel));
+                }
+            }
+        }
+
+        $deletedProductModelsCount = $this->countProductModelsToDelete($productModelsToRemove);
+        foreach ($commandsToExecute as $command) {
```
rfauglas, line 227: whatch out if you have an error, you leave treatment....

src/Akeneo/Pim/Enrichment/Component/Product/Job/DeleteProductsAndProductModelsTasklet.php
```diff
@@ -183,19 +190,52 @@ protected function doDelete(array $entities): void
...
+            ($this->removeProductModelHandler)($command);
```
rfauglas, line 228: Here  if you have an Exception or an Error you're fully dependant on the catch of the command...
If it is not intercepted, you will have the first n-1 commands committed, n raise an Error that stops treatment and the remaining p elements are not treated.... And you don't know which  commands have been skipped (which should not be the case)  and should be replayed... 
This kind of behavior is  global, offered by framework...

src/Akeneo/Pim/Enrichment/Bundle/Controller/ExternalApi/ProductModelController.php
nmarniesse: The validation mechanism translates the message. So the i18n problems is solved "silently".  
Each controller has its own way to format messages. I don't say it's perfect but today it's the case and the aim of this PR is not to improve that

src/Akeneo/Pim/Enrichment/Component/Product/Command/ProductModel/RemoveProductModelCommand.php
nmarniesse: We _always_ should use functional key in command :)  
The intention of the client is to remove the product model that have particular code. And I think we should reflect that. In the recent parts of the application the id is never exposed (UI, API, ...).  
Plus if you put `int $id` here, you make a coupling between your Application layer and your DB implementation.

rfauglas (approved):
Of course, my comments are  far beyond current frame... it is great work!

src/Akeneo/Pim/Enrichment/Component/Product/Command/ProductModel/RemoveProductModelCommand.php
rfauglas: I know it is the practice at Akeneo, I understand to some extent, but still  I'm not convinced, it doesn't echo to my practice/knownledge of API... But I'll have to work on arguments, based on bibliography ...

nmarniesse: I don't wanted to say it's the practice in Akeneo so let's do the same. I wanted to say it's the good practice IMHO.

To explain a little bit, in DDD and Clean Architecture the commands are in the Application layer. This layer knows only about himself and the Domain layer. The Domain layer is driven by ubiquitous language = a common language between functional people and tech people. That's why in the Domain layer, we don't talk about id that is a technical choice made in the Infrastructure layer. Objects are identified by their code.  
In the enrichment BC we mixed a little the layers, so it's not a big deal if we use the code or the id but I prefer to use the code

src/Akeneo/Pim/Enrichment/Bundle/Controller/InternalApi/ProductModelController.php
```diff
@@ -358,25 +321,27 @@ public function listFamilyVariantProductModels(Request $request)
...
         if (!$request->isXmlHttpRequest()) {
             return new RedirectResponse('/');
         }
 
         $productModel = $this->findProductModelOr404($id);
-        $this->productModelRemover->remove($productModel);
+        $command = new RemoveProductModelCommand($productModel->getCode());
+        $violations = $this->validator->validate($command);
+        if (0 < \count($violations)) {
+            // Actually the UI expects only one error message in order to display it as a flash message.
```
mmetayer, line 336:
```suggestion
            // Currently the UI expects only one error message in order to display it as a flash message.
```

src/Akeneo/Pim/Enrichment/Component/Product/Command/ProductModel/RemoveProductModelHandler.php
```diff
@@ -0,0 +1,39 @@
...
+        $this->productModelRemover = $productModelRemover;
+        $this->productAndProductModelClient = $productAndProductModelClient;
+    }
+
+    public function __invoke(RemoveProductModelCommand $command): void
+    {
+        $productModel = $this->productModelRepository->findOneByIdentifier($command->productModelCode());
+        Assert::notNull($productModel);
+        $this->productModelRemover->remove($productModel);
+        $this->productAndProductModelClient->refreshIndex();
```
mmetayer, line 37: I'd let that in the controller, as it's purely technical (we need to refresh the index to have the next information with the subsequent request to display the grid), plus it will slow down the process a lot when deleting multiple models at the same time

src/Akeneo/Pim/Enrichment/Component/Product/Job/DeleteProductsAndProductModelsTasklet.php
mmetayer, line 228: this may slow down the process a lot, as you will reindex each product model (and its ancestors and descendants) one at a time, and possibly reindex it if one of its ancestors/descendants is deleted afterward.
Maybe we could think of a handler able to delete multiple models at a time?

mmetayer (approved):

PR #15819, opened 2021-11-30
TIP-1547: activate dynamic config

ahocquard:
Dynamic config is a Circle CI feature where the CI config file can be generated in a pre-compilation "step".
We will use this feature to Circle CI config in several files:
- in the pre-step (`config.yml` file), it merge all files with yq into a file `merged_config.yml`
- then it reads this file

Do note that with this behavior, anchor should be located in the same file as where they are used. Otherwise, `yq` can't merge the file as some are not Yaml compliant.  That's why I did a single "build_and_deploy" file for now.

Note: I think it's not a big deal because we should limit our usage of YAML feature such as anchor to keep things simple. It's not even correctly parsed in every YAML parser and our anchor/alias usage does respect the new YAML specification. 

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc


tseho (approved):

BitOne (approved):

.circleci/jobs/features/other.yml
```diff
@@ -0,0 +1,428 @@
...
+            - store_test_results:
+                  path: ~/project/pim/var/tests
+            - store_artifacts:
+                  path: ~/project/pim/var/tests
+            - store_artifacts:
+                  path: ~/project/pim/var/logs
+
+    test_database:
+        machine:
+            image: ubuntu-1604:201903-01
```
vgelot, line 409:
```suggestion
            image: ubuntu-2004:202010-01
```

vgelot (approved):

juliensnz (approved):
GTM for the intent. I'm not an expert on circle CI conf so don't hesitate to seek for more than two GTM :)

jmleroux (approved):
Great work!

PR #15831, opened 2021-12-01
CXP-1013: Add an Open App button to Connected Apps

tseho:

Screenshot_2021-12-01_13-47-20

Here is the strategy.

- First, we fetch connected apps from the internal API.
- we render the view (the open app button is disabled)
- Then, we immediatly fetch all marketplace Apps from the marketplace API.
- If the marketplace API fails, nothing happens, the Open App stays disabled.
- If we successfully retrieved the marketplace Apps, we merge both responses and re-render.

Benefits:
- The Connected Apps UI is never slowed down by the marketplace
- The Connected Apps UI still works if the marketplace is down.
- Only 2 API calls are done, even with a lot of Connected Apps.

- [x] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/src/Akeneo/Connectivity/Connection/front/src/connect/hooks/use-connected-apps.ts
+++ b/src/Akeneo/Connectivity/Connection/front/src/connect/hooks/use-connected-apps.ts
@@ -1,0 +1,62 @@
+import {useEffect, useState} from 'react';
+import {NotificationLevel, useNotify} from '../../shared/notify';
+import {ConnectedApp} from '../../model/Apps/connected-app';
+import {useFeatureFlags} from '../../shared/feature-flags';
+import {useFetchConnectedApps} from './use-fetch-connected-apps';
+import {useFetchApps} from './use-fetch-apps';
+import {useTranslate} from '../../shared/translate';
+
+export const useConnectedApps = (): ConnectedApp[] | null | false => {
+    const featureFlag = useFeatureFlags();
+    const notify = useNotify();
+    const translate = useTranslate();
+    const fetchConnectedApps = useFetchConnectedApps();
+    const fetchApps = useFetchApps();
+    const [connectedApps, setConnectedApps] = useState<ConnectedApp[] | null | false>(null);
+
+    useEffect(() => {
+        if (!featureFlag.isEnabled('marketplace_activate')) {
+            setConnectedApps([]);
+            return;
+        }
+
+        (async () => {
+            let connectedApps: ConnectedApp[] | null | false;
+
+            try {
+                connectedApps = await fetchConnectedApps();
+                setConnectedApps(connectedApps);
+            } catch (e) {
+                setConnectedApps(false);
+                notify(
+                    NotificationLevel.ERROR,
+                    translate('akeneo_connectivity.connection.connect.connected_apps.list.flash.error')
+                );
+                return;
+            }
+
+            if (!connectedApps || connectedApps.length === 0) {
+                return;
+            }
+
+            try {
+                const apps = await fetchApps();
+                setConnectedApps(connectedApps =>
+                    !connectedApps
+                        ? connectedApps
+                        : connectedApps.map(connectedApp => {
+                              const app = apps.apps.find(app => app.id === connectedApp.id);
+                              return {
+                                  ...connectedApp,
+                                  activate_url: app?.activate_url || undefined,
+                              };
+                          })
+                );
+            } catch (e) {
+                return;
+            }
+        })();
+    }, [fetchConnectedApps]);
+
+    return connectedApps;
+};
--- a/src/Akeneo/Connectivity/Connection/front/src/connect/hooks/use-connected-apps.ts
+++ b/src/Akeneo/Connectivity/Connection/front/src/connect/hooks/use-connected-apps.ts
@@ -41,17 +41,19 @@ export const useConnectedApps = (): ConnectedApp[] | null | false => {
 
             try {
                 const apps = await fetchApps();
-                setConnectedApps(connectedApps =>
-                    !connectedApps
-                        ? connectedApps
-                        : connectedApps.map(connectedApp => {
-                              const app = apps.apps.find(app => app.id === connectedApp.id);
-                              return {
-                                  ...connectedApp,
-                                  activate_url: app?.activate_url || undefined,
-                              };
-                          })
-                );
+                setConnectedApps(state => {
+                    if (state === null || state === false) {
+                        return state;
+                    }
+
+                    return state.map(connectedApp => {
+                        const app = apps.apps.find(app => app.id === connectedApp.id);
+                        return {
+                            ...connectedApp,
+                            activate_url: app?.activate_url || undefined,
+                        };
+                    });
+                });
             } catch (e) {
                 return;
             }
--- a/src/Akeneo/Connectivity/Connection/front/src/model/Apps/connected-app.ts
+++ b/src/Akeneo/Connectivity/Connection/front/src/model/Apps/connected-app.ts
@@ -9,4 +9,5 @@ export type ConnectedApp = {
     categories: string[];
     certified: boolean;
     partner: string | null;
+    activate_url?: string;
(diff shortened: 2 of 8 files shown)


AymericPlanche (approved):
:pray:

pchasle (approved):

src/Akeneo/Connectivity/Connection/front/tests/src/connect/hooks/use-connected-apps.test.ts
```diff
@@ -0,0 +1,135 @@
...
+});
+
+test('it returns an empty list if the feature flag is disabled', () => {
+    (useFeatureFlags as jest.Mock).mockImplementation(() => ({isEnabled: () => false}));
+
+    const {result} = renderHook(() => useConnectedApps());
+    expect(result.current).toEqual([]);
+});
+
+test('it notify if there it cannot retrieve connected apps', async () => {
```
masacc, line 24:
```suggestion
test('it notifies if it cannot retrieve connected apps', async () => {
```

src/Akeneo/Connectivity/Connection/front/tests/src/connect/hooks/use-connected-apps.test.ts
```diff
@@ -0,0 +1,135 @@
...
+        },
+    });
+
+    const {result, waitForNextUpdate} = renderHook(() => useConnectedApps());
+    expect(result.current).toEqual(null);
+    await waitForNextUpdate();
+    expect(result.current).toEqual([]);
+});
+
+test('it does not fails if it cannot retrieve marketplace apps', async () => {
```
masacc, line 52:
```suggestion
test('it does not fail if it cannot retrieve marketplace apps', async () => {
```

src/Akeneo/Connectivity/Connection/front/tests/src/connect/hooks/use-connected-apps.test.ts
```diff
@@ -0,0 +1,135 @@
...
+        },
+    });
+
+    const {result, waitForNextUpdate} = renderHook(() => useConnectedApps());
+    expect(result.current).toEqual(null);
+    await waitForNextUpdate();
+    expect(result.current).toEqual([connectedApp]);
+});
+
+test('it fetch connected apps', async () => {
```
masacc, line 84:
```suggestion
test('it fetches connected apps', async () => {
```

src/Akeneo/Connectivity/Connection/front/src/connect/hooks/use-connected-apps.ts
```diff
@@ -0,0 +1,62 @@
...
+            }
+
+            if (!connectedApps || connectedApps.length === 0) {
+                return;
+            }
+
+            try {
+                const apps = await fetchApps();
+                setConnectedApps(connectedApps =>
+                    !connectedApps
```
masacc, line 45:
```suggestion
                    !apps
```

src/Akeneo/Connectivity/Connection/front/src/connect/hooks/use-connected-apps.ts
```diff
@@ -0,0 +1,62 @@
...
+                        ? connectedApps
+                        : connectedApps.map(connectedApp => {
+                              const app = apps.apps.find(app => app.id === connectedApp.id);
```
masacc, line 48: what happens if there is no app in the marketplace with the same id ?

src/Akeneo/Connectivity/Connection/front/src/model/Apps/connected-app.ts
```diff
@@ -9,4 +9,5 @@ export type ConnectedApp = {
     categories: string[];
     certified: boolean;
     partner: string | null;
+    activate_url?: string;
```
masacc, line 12: both lines are not consistent for the same type hinting

src/Akeneo/Connectivity/Connection/front/src/connect/hooks/use-connected-apps.ts
tseho: nope, that's on purpose, it's not a mistake.

src/Akeneo/Connectivity/Connection/front/src/model/Apps/connected-app.ts
tseho: Which both lines ?
These 2 ?
```
partner: string | null;
activate_url?: string;
```
That's normal, it's doesn't mean the same thing.
the `activate_url?: string;` tells that the property is not always there.
But when it's there, there is always a string in it.
`partner: string | null;` tells that the property is always here, but it contains either null or string.

src/Akeneo/Connectivity/Connection/front/src/connect/hooks/use-connected-apps.ts
tseho: In the following lines, I do this:
`activate_url: app?.activate_url || undefined,`
`app?.activate_url` means that if app is null, it does not throw and returns null instead.
Then the `|| undefined` fallback in that case to undefined, like if `activate_url` was never defined.

src/Akeneo/Connectivity/Connection/front/src/model/Apps/connected-app.ts
masacc: OK, I had not well understood the meaning of the `?`... thank you!

src/Akeneo/Connectivity/Connection/front/src/connect/hooks/use-connected-apps.ts
tseho: `setConnectedApps()` is a setter for useState.
It take in argument the following signature: `(currentState) => newState`
So, the trick 
```
setConnectedApps(connectedApps =>
                    !connectedApps
                    ? connectedApp
                    : connectedApps
```

Can be translated like this:

```
setConnectedApps((currentState) => {
  if(currentState === null || currentState === false) {
    return currentState;
  }
  return currentState.map(...);
});

masacc: OK, could you explain to me why? because it seems a bit weird to me to write "if no connectedApps, return connectedApps"

tseho: If you think that it's not readable enough, I can rewrite it.

masacc: no that's ok thank you for the explanation

tseho: > OK, could you explain to me why? because it seems a bit weird to me to write "if no connectedApps, return connectedApps"

Because the variable `connectedApps` in the state can be either an array of apps, `false` or `null`.
So what I'm writing is "If connectedApps is either false or null, return accordingly false or null, otherwise, do your stuff"

tseho: I really think we should manage API responses in a better way, like react-query does.

instead of having `useState<ConnectedApp[] | null | false>(null);`
We should have something like `useState<{loading: boolean; failed: boolean; data: any}>();

tseho: I've rewrited it to make it more readable.
Thanks for the feedback.

pchasle: You have a `useQuery` hook in the shared folder if you want

tseho: I remember but we would like to phase it out and use a decent library instead.

PR #15865, opened 2021-12-06
PIM-10187: Fix impossible to configure the columns of the product grid if attribute labels are too long

StevenVAIDIE:
In this PR I fixed the drag and drop in the column datagrid selector. 

Why we have the problem ?
Because the column does not have the same width depending on the length of label. When the attribute group label is too long the column does not take 30% the other take less. When we try to drag and drop we cannot drop in the right column because the width of the column is more than the column so the pointer is blocked.

Now the column are not anymore a flex-basic => it's a width so we need to handle the overflow

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/src/Akeneo/Platform/Bundle/UIBundle/Resources/public/less/components/VerticalList.less
+++ b/src/Akeneo/Platform/Bundle/UIBundle/Resources/public/less/components/VerticalList.less
@@ -23,6 +23,12 @@
     white-space: nowrap;
     overflow: hidden;
 
+    &-itemLabel {
+      overflow: hidden;
+      text-overflow: ellipsis;
+      white-space: nowrap;
+    }
+
     &--selectable {
       cursor: pointer;
     }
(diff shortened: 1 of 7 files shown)


src/Akeneo/Platform/Bundle/UIBundle/Resources/public/less/components/VerticalList.less
```diff
@@ -23,6 +23,12 @@
     white-space: nowrap;
     overflow: hidden;
 
+    &-itemLabel {
+      overflow: hidden;
+      text-overflow: ellipsis;
```
mmetayer, line 28: don't you have to add a "title" somewhere in case the text is truncated?

ValentinMumble (approved):

mmetayer (approved):

PR #15880, opened 2021-12-09
PIM-10194: Fix pagination for list products/product models endpoints with search_after pagination type

mmetayer:

Following #15825 the pagination's "search_after" in the products and product models is now the identifier/code, instead of the encrypted MySQL id. However this id is still needed to perform the ES query, because every PQB inserts a `['sort' => ['_id' => 'asc']]` clause in the ES query, and ES needs every item of the `sort` clause to be also present in the `search_after` clause.
Also, the search_after `identifier` property expects a lowercase value (the identifiers are normalized in lowercase in the indexes, and ES does not seem to transform the values of "search_after" when performing the query)

- [x] Tests
- [ ] Migration & Installer
- [x] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/src/Akeneo/Pim/Enrichment/Bundle/Storage/Sql/Product/SqlGetProductId.php
+++ b/src/Akeneo/Pim/Enrichment/Bundle/Storage/Sql/Product/SqlGetProductId.php
@@ -1,0 +1,29 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Akeneo\Pim\Enrichment\Bundle\Storage\Sql\Product;
+
+use Akeneo\Pim\Enrichment\Component\Product\Query\GetId;
+use Doctrine\DBAL\Connection;
+
+/**
+ * @copyright 2021 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+final class SqlGetProductId implements GetId
+{
+    public function __construct(private Connection $connection)
+    {
+    }
+
+    public function fromIdentifier(string $identifier): null|string
+    {
+        $id = $this->connection->executeQuery(
+            'SELECT id FROM pim_catalog_product WHERE identifier = :identifier',
+            ['identifier' => $identifier]
+        )->fetchOne();
+
+        return false === $id ? null : (string)$id;
+    }
+}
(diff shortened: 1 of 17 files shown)


ahocquard (approved):

src/Akeneo/Pim/Enrichment/Component/Product/Query/GetId.php
```diff
@@ -0,0 +1,14 @@
...
+
+declare(strict_types=1);
+
+namespace Akeneo\Pim\Enrichment\Component\Product\Query;
+
+/**
+ * @copyright 2021 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+interface GetId
```
mmetayer, line 11: I'm not very pleased with the naming, feel free to suggest something better :slightly_smiling_face:

src/Akeneo/Pim/Enrichment/Bundle/Storage/Sql/Product/SqlGetProductId.php
```diff
@@ -0,0 +1,29 @@
...
+namespace Akeneo\Pim\Enrichment\Bundle\Storage\Sql\Product;
+
+use Akeneo\Pim\Enrichment\Component\Product\Query\GetId;
+use Doctrine\DBAL\Connection;
+
+/**
+ * @copyright 2021 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+final class SqlGetProductId implements GetId
```
ahocquard, line 14: `Get` generally throw an exception because you expect a result (and Find return null)

src/Akeneo/Pim/Enrichment/Component/Product/Query/GetId.php
```diff
@@ -0,0 +1,14 @@
...
+{
+    public function fromIdentifier(string $identifier): null|string;
```
jmleroux, line 13:
```suggestion
    public function fromIdentifier(string $identifier): ?string;
```

jmleroux (approved):

src/Akeneo/Pim/Enrichment/Component/Product/Query/GetId.php
mmetayer: Now that union types are available, I prefer using this notation, I find it more readable :slightly_smiling_face:

PR #15945, opened 2021-12-15
PIM-10192: Retry mechanism in  delete  action of ES documents

rfauglas:
https://akeneo.atlassian.net/browse/PIM-10192
After analysis it is an internal Error caused by ElasticSearch due to concurrent access on ressources protected by optimistic lock.
I chose to implement in the Client Adaptation layer.

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc


src/Akeneo/Tool/Bundle/ElasticsearchBundle/Client.php
```diff
@@ -378,13 +386,27 @@ public function getConfigurationLoader(): Loader
...
+            $attempts++;
+            try {
+                $this->client->deleteByQuery([
+                    'index' => $this->indexName,
+                    'body' => $body,
+                ]);
+                return;
+            } catch (Conflict409Exception $e) {
+                $exception = $e;
+                sleep($this->maxExpectedIndexationLatency);
```
jmleroux, line 405:
```suggestion
                usleep($this->maxExpectedIndexationLatency);
```
Sleep parameter is in seconds. 100 seconds is huge. I guess you wanted 100 microsecs.

rfauglas: I convert  it in constructor... I wanted milliseconds...

RiffVentura (approved):

nmarniesse (approved):

src/Akeneo/Tool/Bundle/ElasticsearchBundle/Client.php
```diff
@@ -32,6 +33,9 @@ class Client
     private NativeClient $client;
     private string $idPrefix;
     private int $maxChunkSize;
+    private int $maxExpectedIndexationLatency;
```
ahocquard, line 36:
```suggestion
    private int $maxExpectedIndexationLatencyInMs;
```

Always better to specify the unit in the name, it often avoids many issues.

ahocquard (approved):
Pretty clean to do that in the client :+1:

src/Akeneo/Tool/Bundle/ElasticsearchBundle/Client.php
nmarniesse: I hesitate to add this comment too :)  
But `Ms` is usually milliseconds, not microseconds

ahocquard: I thought it was milliseconds, that's why. hence the importance to specify it.
`maxExpectedIndexationLatencyInMicroseconds`. Long but explicit.

rfauglas: i agree with both of you I'd add  we miss type alias to express units, which can be got around with naming conventions...
configuration in units can be precised in configuration. See my last proposal in last commit.

ahocquard: I would change the key configuration also: `max_expected_indexation_latency`

PR #15969, opened 2021-12-16
PIM-10208: Fix currency settings page crashing when label is not found

ValentinMumble:
Fixing the currency settings crashing when the currency label for an unknown currency code was not found (because we are using Symfony\Component\Intl to retrieve currency labels)

### DIFF
--- a/tests/back/Channel/Specification/Bundle/Twig/LocaleExtensionSpec.php
+++ b/tests/back/Channel/Specification/Bundle/Twig/LocaleExtensionSpec.php
@@ -66,6 +66,14 @@ class LocaleExtensionSpec extends ObjectBehavior
 
     }
 
+    function it_returns_null_when_the_currency_label_is_not_found($userContext, LocaleInterface $en)
+    {
+        $userContext->getCurrentLocale()->willReturn($en);
+        $this->currencyLabel('XSU')->shouldReturn(null);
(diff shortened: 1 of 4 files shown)


jmleroux (approved):

tests/back/Channel/Specification/Bundle/Twig/LocaleExtensionSpec.php
```diff
@@ -66,6 +66,14 @@ function it_provides_a_currency_label_translated_in_the_specified_locale($userCo
 
     }
 
+    function it_returns_null_when_the_currency_label_is_not_found($userContext, LocaleInterface $en)
+    {
+        $userContext->getCurrentLocale()->willReturn($en);
+        $this->currencyLabel('XSU')->shouldReturn(null);
+        $this->currencyLabel('XSU', 'fr_FR')->shouldReturn(null);
+
```
mmetayer, line 74:
```suggestion
```

RiffVentura (approved):

tests/back/Channel/Specification/Bundle/Twig/LocaleExtensionSpec.php
ValentinMumble: :ah:

nmarniesse (approved):

PR #15996, opened 2021-12-17
CXP-1016 Removed non default user group and role from system count

RiffVentura:
Updated count numbers on system panel for user group and roles
Now count ignores  user groups and roles that are not default


willy-ahva (approved):

tseho (approved):

RiffVentura:
warm master

PR #16050, opened 2021-12-22
CXP-928 Limit connections and apps to 50 through UI

RiffVentura:
Added a proper limit to the number of connections and connected Apps through user interface
- Added endpoint to fetch if max limit is reached
- Created a front hook to fetch the max limit
- Applied limit on  connections 'Create' button
- Applied limit on apps 'Connect' button

Apps
image

Connections
image


src/Akeneo/Connectivity/Connection/back/Application/Settings/Query/IsConnectionsNumberLimitReachedHandler.php
```diff
@@ -0,0 +1,23 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Akeneo\Connectivity\Connection\Application\Settings\Query;
+
+use Akeneo\Connectivity\Connection\Domain\Settings\Persistence\Query\CountAllConnectionsQueryInterface;
+
+class IsConnectionsNumberLimitReachedHandler
```
tseho, line 9: Why a handler ? It doesn't handle a command.
Instead of having a Handler & a Query, why don't you have only one query ?

```
class IsMaximumNumberOfConnectionsReachedQuery {
 public function execute(): bool;
}
```
What do you think ?

src/Akeneo/Connectivity/Connection/front/src/shared/hooks/use-connections-limit-reached.ts
```diff
@@ -0,0 +1,35 @@
...
+
+        return response.json().then((content: MaxLimitReached) => content.limitReached);
+    }, [url]);
+
+    useEffect(() => {
+        fetchCallback()
+            .then(setLimitReached)
+            .catch(() => {
+                setLimitReached(true);
+            });
```
tseho, line 31: Why do you put true when there is an error ?

src/Akeneo/Connectivity/Connection/front/src/shared/hooks/use-connections-limit-reached.ts
```diff
@@ -0,0 +1,35 @@
...
+    }, [fetchCallback, setLimitReached]);
```
tseho, line 32: In that case, I think you shouldn't have dependencies in your useEffect.
You only want to execute it once.

src/Akeneo/Connectivity/Connection/back/Application/Settings/Query/IsConnectionsNumberLimitReachedHandler.php
RiffVentura: Initially I wanted to have only a query that returns boolean result
After some thought, I found it ugly that a query dealing with database do the "validation" within the same class
and violates SRP in my eyes

Why command ? To be more consistent with other files within the same namespace : there are "query" objects that act as DTO parameter for handlers. (I've renamed to Handler after noticing it). Do you think it should be renamed to query instead ?

I am not sure about overall naming I've chosen though...

tseho: Regarding the SRP, do you think that having a query returning true/false to the question "is the limit reached ?" violate the SRP ?

src/Akeneo/Connectivity/Connection/front/src/shared/hooks/use-connections-limit-reached.ts
RiffVentura: I've had some issues with my test: settings null  does not rerenders the component making it hard to test

From usability it doesn't change much IMO (I would prefer boolean only without null at all)

src/Akeneo/Connectivity/Connection/back/Application/Settings/Query/IsConnectionsNumberLimitReachedHandler.php
RiffVentura: Well now that you rephrase it like that, no. I'm probably abusing Command pattern :sweat_smile:

tseho: Yes, it's not easy to find the correct rule of thumb for this.

Anyway, remember:
Command/Handler => do something, mutate data.
Query => returns data without mutating anything.

guillaumefer-akeneo (approved):

src/Akeneo/Connectivity/Connection/back/Infrastructure/InternalApi/Controller/IsConnectionsNumberLimitReachedAction.php
```diff
@@ -0,0 +1,31 @@
...
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+
+/**
+ * @copyright 2021 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+class IsConnectionsNumberLimitReachedAction
+{
+    public function __construct(private IsConnectionsNumberLimitReachedQueryInterface $isConnectionsNumberLimitReachedHandler)
```
tseho, line 19:
```suggestion
    public function __construct(private IsConnectionsNumberLimitReachedQueryInterface $isConnectionsNumberLimitReachedQuery)
```

src/Akeneo/Connectivity/Connection/back/Infrastructure/InternalApi/Controller/IsConnectionsNumberLimitReachedAction.php
```diff
@@ -0,0 +1,31 @@
...
+    {
+    }
+
+    public function __invoke(Request $request): Response
+    {
+        if (!$request->isXmlHttpRequest()) {
+            return new RedirectResponse('/');
+        }
+
+        return new JsonResponse(['limitReached' => $this->isConnectionsNumberLimitReachedHandler->execute()]);
```
tseho, line 29:
```suggestion
        return new JsonResponse(['limitReached' => $this->isConnectionsNumberLimitReachedQuery->execute()]);
```

src/Akeneo/Connectivity/Connection/back/tests/.php_cd.php
```diff
@@ -237,6 +237,8 @@
             'Akeneo\Connectivity\Connection\Infrastructure\Service\Encrypter',
 
             'Symfony\Component\OptionsResolver\OptionsResolver',
+
+            'Akeneo\Connectivity\Connection\Infrastructure\Connections\GetConnectionsNumberLimit'
```
tseho, line 241:
```suggestion
            'Akeneo\Connectivity\Connection\Infrastructure\Connections\GetConnectionsNumberLimit',
```

src/Akeneo/Connectivity/Connection/back/Infrastructure/Symfony/Resources/config/validators.yml
```diff
@@ -1,6 +1,6 @@
 parameters:
     env(ALLOWED_NETWORK_WHITELIST): ''
-
+    akeneo_connectivity.connection.max_limit: 50
```
tseho, line 3:
```suggestion
    akeneo_connectivity.connection.max_limit: 50

```

tseho: no need to remove the blank line between parameters & services ;)

src/Akeneo/Connectivity/Connection/back/tests/EndToEnd/Connection/IsConnectionsNumberLimitReachedActionEndToEnd.php
```diff
@@ -0,0 +1,45 @@
...
+/**
+ * @copyright 2021 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+class IsConnectionsNumberLimitReachedActionEndToEnd extends WebTestCase
+{
+    protected function getConfiguration(): Configuration
+    {
+        return $this->catalog->useMinimalCatalog();
+    }
```
tseho, line 22:
```suggestion
    protected function getConfiguration(): ?Configuration
    {
        return null;
    }
```

Trick: If you don't need the fixtures (like default product families etc), you can return null here. Your test will be faster ;)

tseho (approved):

src/Akeneo/Connectivity/Connection/back/tests/EndToEnd/Connection/IsConnectionsNumberLimitReachedActionEndToEnd.php
RiffVentura: I've tried but null configuration fails so I leave it as it is

tseho: That's sad. It's not a big deal.

src/Akeneo/Connectivity/Connection/back/Infrastructure/Connections/GetConnectionsNumberLimit.php
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * The following class is introduced to ease testability of classes that might require it.
+ * As the set limit is relatively high, without it integration tests would create a great amount
+ * of connections, dampening tests performance
+ *
+ * @copyright 2021 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+class GetConnectionsNumberLimit
```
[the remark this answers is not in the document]
RiffVentura: Initially it was done this way but in order to test `IsConnectionsNumberLimitReachedQuery` I would need to create enough connections to breach this limit. As it would impact integration test time, this object was created to overwrite the limit to a lower number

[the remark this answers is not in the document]
willy-ahva: Ok, so you just had to override the existing service in the DI dedicated to the test env and change the limit :)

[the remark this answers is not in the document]
tseho: This is also a solution but the problem is that if you do that, you really change the behavior of `IsConnectionsNumberLimitReachedQuery` to all integration tests.
Both solutions are not ideal, and when we discussed about it with Rifat, no other solution appeared.

PR #16064, opened 2021-12-23
PIM-10217: Fix cannot quick export product model when id is not present in grid context

ghoussard:

When a user launch a quick export of a product model with a grid context without id (code), the job fails because the `ProductModelWriter` does not have the code to retrieve media file. 
This PR uses the same way as the product quick export and set a `identifier` (which will not be added into exported file) property that will be use to fallback if code is not exported.

I'm not thrilled about this solution, feel free to suggest me a cleaner one.

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/src/Akeneo/Pim/Enrichment/Component/Product/Connector/Processor/QuickExport/ProductAndProductModelProcessor.php
+++ b/src/Akeneo/Pim/Enrichment/Component/Product/Connector/Processor/QuickExport/ProductAndProductModelProcessor.php
@@ -140,6 +140,8 @@ class ProductAndProductModelProcessor extends AbstractProcessor
                 );
             } elseif (in_array($codeProperty, $selectedProperties) || 'identifier' === $codeProperty) {
                 $propertiesToExport[$codeProperty] = $property;
+            } elseif ('code' === $codeProperty) {
+                $propertiesToExport['identifier'] = $property;
             }
         }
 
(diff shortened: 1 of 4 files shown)


RiffVentura (approved):

src/Akeneo/Pim/Enrichment/Component/Product/Connector/Processor/QuickExport/ProductAndProductModelProcessor.php
```diff
@@ -140,6 +140,8 @@ function ($attributeCode) use ($selectedProperties) {
                 );
             } elseif (in_array($codeProperty, $selectedProperties) || 'identifier' === $codeProperty) {
                 $propertiesToExport[$codeProperty] = $property;
+            } elseif ('code' === $codeProperty) {
+                $propertiesToExport['identifier'] = $property;
```
StevenVAIDIE, line 144: For me you just need this, the writer doesn't need to know that something the identifier is in identifier
```suggestion
            } elseif (in_array($codeProperty, $selectedProperties) || 'identifier' === $codeProperty || 'code' === $codeProperty) {
                $propertiesToExport[$codeProperty] = $property;
            }
```

StevenVAIDIE (approved):

src/Akeneo/Pim/Enrichment/Component/Product/Connector/Processor/QuickExport/ProductAndProductModelProcessor.php
ghoussard: If we do that, the code column and value will be present in the exported csv/xlsx file even if it is not in the grid context.

PR #16082, opened 2022-01-03
CXP-1047: Add test app table

tseho:

I've left `user_id` nullable with `ON DELETE SET NULL` because if we have a `DELETE CASCADE` on it, and the related user is deleted, it would delete his test apps, not the related connected apps.

I think it's better that users are required to use our Command+Handler through the UI to delete a test app, it will take care of side effects instead of leaving the database in a unexpected state.

It's easy to display in the UI that the creator of the test app was deleted, if the value is null.

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc


RiffVentura (approved):

guillaumefer-akeneo (approved):

tseho:
check frozen

jmleroux (approved):

PR #16088, opened 2022-01-03
Cpm 444 ignore mocks

cemorin:
We discovered that the folders __mocks__ were exported during the webpack build. This PR removes all the "__mocks__" folders from wepback build.

build before the update: 
main.min.js   8.69 MiB    main  [emitted]              main
               
build after: 
main.min.js   8.64 MiB    main  [emitted]              main


juliensnz (approved):

tseho (approved):

pchasle (approved):

cemorin:
Pim is frozen on loading screen with this config

PR #16094, opened 2022-01-04
CXP-1041 Display test apps

RiffVentura:
image

PR #16102, opened 2022-01-06
CXP-1048: Be able to create a test app

willy-ahva:

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc


src/Akeneo/Connectivity/Connection/back/Application/TestApps/Command/CreateTestAppCommandHandler.php
```diff
@@ -0,0 +1,45 @@
...
+    public function __construct(
+        private RandomCodeGeneratorInterface $randomCodeGenerator,
+        private CreateTestAppQueryInterface $createTestAppQuery,
+    ) {
+    }
+
+    /**
+     * @return array{'clientId': string, 'clientSecret': string}
+     */
+    public function handle(CreateTestAppCommand $command): array
```
tseho, line 26: I have a question here, should a CommandHandler return nothing ?
I totally get why it's bothering in the current.
Maybe that the clientId should an UUID created before, included in the Command object.
This way, you can do:

```
$clientId = UUID::create();
$command = new CreateTestAppCommand(
            $clientId,
            $request->get('name', ''),
            $request->get('activateUrl', ''),
            $request->get('callbackUrl', ''),
            $user->getId(),
        );
$handler->handle($command);

$credentials = $this->query->execute($clientId);
```

Thoughts ?

willy-ahva: I asked myself the same question! I think you are right.

src/Akeneo/Connectivity/Connection/back/Infrastructure/Symfony/Resources/config/validation/CreateTestAppCommand.yml
```diff
@@ -0,0 +1,16 @@
...
+            - NotBlank:
+                  message: 'akeneo_connectivity.connection.test_app.errors.creation.name.not_blank'
+
+        activateUrl:
+            - Url:
+                  message: 'akeneo_connectivity.connection.test_app.errors.creation.activate_url.must_be_url'
+
+        callbackUrl:
+            - Url:
+                  message: 'akeneo_connectivity.connection.test_app.errors.creation.callback_url.must_be_url'
```
tseho, line 16: Length too ?
If their urls are longer than 255, it will fail during the insert.

src/Akeneo/Connectivity/Connection/back/tests/Unit/spec/Infrastructure/TestApps/Controller/InternalApi/CreateTestAppsActionSpec.php
```diff
@@ -0,0 +1,172 @@
...
+        FeatureFlag $featureFlag,
+        Request $request,
+    ): void {
+        $featureFlag->isEnabled()->willReturn(false);
+        $this
+            ->shouldThrow(new NotFoundHttpException())
+            ->during('__invoke', [$request]);
+    }
+
+    public function it_redirects_to_the_root_if_the_request_does_not_come_from_a(
```
tseho, line 86:
```suggestion
    public function it_redirects_to_the_root_if_the_request_does_not_come_from_ajax(
```

src/Akeneo/Connectivity/Connection/back/Application/TestApps/Command/CreateTestAppCommandHandler.php
RiffVentura: on the other hand it requires an additional query to database when we already have the information

src/Akeneo/Connectivity/Connection/back/tests/Integration/TestApps/Persistence/CreateTestAppQueryIntegration.php
```diff
@@ -0,0 +1,68 @@
...
+
+class CreateTestAppQueryIntegration extends TestCase
+{
+    private Connection $connection;
+    private CreateTestAppQuery $createTestAppQuery;
+    private RandomCodeGeneratorInterface $randomCodeGenerator;
+
+    public function test_it_creates_a_test_app(): void
+    {
+        $this->createAdminUser();
```
RiffVentura, line 24: returns `UserInterface` so there is no need to fetch Id through `getAdminId`

src/Akeneo/Connectivity/Connection/back/tests/Integration/TestApps/Persistence/GetTestAppSecretQueryIntegration.php
```diff
@@ -0,0 +1,75 @@
...
+
+class GetTestAppSecretQueryIntegration extends TestCase
+{
+    private Connection $connection;
+    private GetTestAppSecretQuery $getTestAppSecretQuery;
+    private RandomCodeGeneratorInterface $randomCodeGenerator;
+
+    public function test_it_gets_the_secret_associated_to_a_test_app_client_id(): void
+    {
+        $this->createAdminUser();
```
RiffVentura, line 24: same here

RiffVentura (approved):

guillaumefer-akeneo (approved):

src/Akeneo/Connectivity/Connection/back/Infrastructure/Symfony/Resources/translations/jsmessages.en_US.yml
```diff
@@ -206,6 +206,21 @@ akeneo_connectivity.connection:
                 redirect_uri:
                     not_blank: The redirect_uri is required.
                     must_be_url: The redirect_uri is not a valid url.
+    test_app:
```
masacc, line 209: there is a pending question for this key here : https://github.com/akeneo/pim-community-dev/pull/16094#discussion_r780098427

src/Akeneo/Connectivity/Connection/back/Application/TestApps/Command/CreateTestAppCommandHandler.php
tseho: Yes it does, but more about having clear purpose and roles between our commands and queries.

Command + Handler: do this specific action/mutation I'm asking, I don't care how. (From the outside, It's hidden in the implem of the handler)
Query: Give me this piece of information. (A query should always be idempotent)

If you start mixing both, you always end up with trouble.

What if later on, we want to return the credentials in another use case ? shit, I can only retrieve them during the creation.
Ok, I can add a new query. But then, you have two way to retrieve the credentials, one way during creation, one way for other use cases. And you have to maintain both and be careful about them staying consistent.

PR #16104, opened 2022-01-06
Add e2e for authentication feature

tomglvng:


tseho (approved):

tseho:
> Failed asserting that actual size 3 matches expected size 2.

Damn, that's right.
It didn't close the tab from the previous test so you now have 3 tabs instead of 2.

My best guest would be, flag the context when a new tab is added.
Add a hook on post-scenario, check if the flag is enabled, meaning a new tab was open, then try those methods:
https://mink.behat.org/en/latest/guides/session.html#resetting-the-session
`$session->restart()` will restart the browser, it's quit slow for the current test, it's not a big deal.

tomglvng:
> Error from the CI:
> 
> > Failed asserting that actual size 3 matches expected size 2.
> 
> Damn, that's right. It didn't close the tab from the previous test so you now have 3 tabs instead of 2.
> 
> My best guest would be, flag the context when a new tab is added. Add a hook on post-scenario, check if the flag is enabled, meaning a new tab was open, then try those methods: https://mink.behat.org/en/latest/guides/session.html#resetting-the-session `$session->restart()` will restart the browser, it's quit slow but for those tests, it's not a big deal.

Thx for helping! Done in https://github.com/akeneo/pim-community-dev/pull/16104/commits/a7ea51cdb3a6783e9a18976a845ba1b48f6ea376

masacc (approved):

RiffVentura (approved):

AymericPlanche (approved):

PR #16132, opened 2022-01-12
Use symfony's `lint:container` command to check services definitions

mmetayer:

This PR aims at replacing the `check-services-instantiability` script with Symfony's native [`lint:container`](https://symfony.com/blog/new-in-symfony-4-4-service-container-linter) cli command, which analyses the dependency injection graph to ensure injected arguments match type declarations

****Pros:****
- one less script to maintain
- the script uses reflection, so it can analyze every service definition (no need to exclude a service from the analysis because it performs an SQL query when instantiated - cf `one_login_auth` service)
- it can be performed in every environment (it analyses `prod` services in CI build)

****Cons:****
- it does not really instantiate services, so we could have some surprises if constructors or factories perform operations more complex than just assigning injected arguments to class properties

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc


src/Oro/Bundle/PimDataGridBundle/Resources/config/form_types.yml
```diff
@@ -1,6 +1,5 @@
 parameters:
     pim_datagrid.form.type.datagrid_view.class:          Oro\Bundle\PimDataGridBundle\Form\Type\DatagridViewType
-    pim_datagrid.form.type.datagrid_filter_choice.class: Oro\Bundle\PimDataGridBundle\Form\Type\DatagridFilterChoiceType
```
mmetayer, line 3: not used since v3, and it was not even instantiable in EE

jmleroux (approved):

BitOne (approved):

nmarniesse (approved):

ValentinMumble (approved):

PR #16134, opened 2022-01-12
CXP 1040 Handle no app logo for the PageHeader

RiffVentura:
Test  apps have no logo image,  we should display AppIllustration  instead. 
Since `PageHeader` is used to display connected (Test) App settings, it should be able to display Illustration if logo path is not provided

image

### DIFF
--- a/src/Akeneo/Connectivity/Connection/front/src/common/components/PageHeader.tsx
+++ b/src/Akeneo/Connectivity/Connection/front/src/common/components/PageHeader.tsx
@@ -21,6 +21,35 @@ const AknTitleContainerBreadcrumbs = styled.div.attrs(() => ({className: 'AknTit
     min-height: 32px;
 `;
 
+const Header = styled.header`
+    position: sticky;
+    top: 0;
+    padding: 40px 40px 20px;
+    background: white;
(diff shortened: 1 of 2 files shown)


tseho (approved):

src/Akeneo/Connectivity/Connection/front/src/common/components/PageHeader.tsx
```diff
@@ -66,3 +82,20 @@ const Header = styled.header`
...
+    border-radius: 4px;
+    display: flex;
+    overflow: hidden;
+    flex-basis: 142px;
+    flex-shrink: 0;
+
+    & > * {
+        width: 100%;
+    }
+`;
```
masacc, line 101: why this styled component is at the end of the file?

masacc (approved):

src/Akeneo/Connectivity/Connection/front/src/common/components/PageHeader.tsx
tseho: There was another component at the end of the file, it does not make sense, I've moved both up there, like everywhere else.

PR #14009, opened 2021-03-02
PIM-9698: refresh ES index after creating a product from the UI

willy-ahva:
**Description (for Contributor and Core Developer)**

After creating at least 2 or 3 products from the UI we saw that the ProductCreated event was not sent to the event subscriptions.

**Definition Of Done (for Core Developer only)**

| Q                                 | A
| --------------------------------- | ---
| Added Specs                       | -
| Added legacy Behats               | -
| Added acceptance tests            | -
| Added integration tests           | -
| Changelog updated                 | OK
| Review and 2 GTM                  | Todo
| Micro Demo to the PO (Story only) | Todo
| Migration script                  | -
| Tech Doc                          | -

`Todo`: Pending / Work in progress
`OK`: Done / Validated
`-`: Not needed

PR #16145, opened 2022-01-13
PIM-10233: [Backport] Refresh ES index after creating a product from the UI in order to well send product created event to event subscriptions

ValentinMumble:
Backporting https://github.com/akeneo/pim-community-dev/pull/14009

### DIFF
--- a/CHANGELOG-5.0.md
+++ b/CHANGELOG-5.0.md
@@ -1,5 +1,9 @@
 # 5.0.x
 
+## Bug fixes
+
+- PIM-10233: [Backport] Refresh ES index after creating a product from the UI in order to well send product created event to event subscriptions
+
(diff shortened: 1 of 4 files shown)


nmarniesse (approved):

CHANGELOG-5.0.md
```diff
@@ -1,5 +1,9 @@
 # 5.0.x
 
+## Bug fixes
```
ahocquard, line 3: I don't understand the original fix. It seems to be a workaround. 

In which way does it solve the fact we emit event?

ghoussard: The product is not fetched when we build the event. We think that the event builder try to fetch a product before that one will be indexed.

PR #16168, opened 2022-01-18
AOB-1445-implemente-findby : implemente findby in memory and change spec

kpAkeneo:
implemente findby in memory to correct integration test and change spec


tests/back/Acceptance/Product/InMemoryProductRepository.php
```diff
@@ -77,7 +77,21 @@ public function findAll()
 
     public function findBy(array $criteria, array $orderBy = null, $limit = null, $offset = null)
     {
-        throw new NotImplementedException(__METHOD__);
+        $products = [];
+        foreach ($this->products as $product) {
+            $keepThisProduct= true;
+            foreach ($criteria as $key => $value) {
+                $getter = sprintf('get%s', ucfirst($key));
```
ValentinMerlet, line 84: Maybe we should use the Symfony PropertyAccess component instead. 
This will avoid issues like if we try to use this method by passing the `enabled` property as criteria. As `getEnabled()` method does not exist in the product, it will fail.

PR #16170, opened 2022-01-19
AOB-1445 : Implement findBy in memory and change spec

kpAkeneo:
In pim-onboarder we use productRepository->findBy but him is not implemented in findby so some integration-test of pim was on error so i implemente in memory to correct integration test and change spec

### DIFF
--- a/tests/back/Acceptance/Product/InMemoryProductRepository.php
+++ b/tests/back/Acceptance/Product/InMemoryProductRepository.php
@@ -80,20 +80,7 @@ class InMemoryProductRepository implements
      */
     public function findBy(array $criteria, array $orderBy = null, $limit = null, $offset = null)
     {
-        $products = [];
-        foreach ($this->products as $product) {
-            $keepThisProduct= true;
-            foreach ($criteria as $value) {
-                if ($product->getIdentifier() !== $value) {
-                    $keepThisProduct = false;
-                    break;
-                }
-            }
-            if ($keepThisProduct) {
-                $products[] = $product;
-            }
-        }
-        return $products;
+        return $this->products->filter(fn (ProductInterface $product) => $product->getIdentifier() === $criteria['identifier'])->toArray();
     }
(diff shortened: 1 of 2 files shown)


tests/back/Acceptance/Product/InMemoryProductRepository.php
```diff
@@ -76,7 +77,23 @@ public function findAll()
...
-        throw new NotImplementedException(__METHOD__);
+        $products = [];
+        $propertyAccessor = PropertyAccess::createPropertyAccessorBuilder()
+            ->getPropertyAccessor();
+
+        foreach ($this->products as $product) {
+            $keepThisProduct= true;
+            foreach ($criteria as $key => $value) {
+                if ($propertyAccessor->getValue($product, $key) !== $value) {
+                    $keepThisProduct = false;
```
mmetayer, line 88:
```suggestion
                    $keepThisProduct = false;
                    break;
```

mmetayer (approved):

tests/back/Acceptance/Product/InMemoryProductRepository.php
```diff
@@ -76,7 +77,23 @@ public function findAll()
 
     public function findBy(array $criteria, array $orderBy = null, $limit = null, $offset = null)
     {
-        throw new NotImplementedException(__METHOD__);
+        $products = [];
```
ahocquard, line 80: could we hardcode the criteria "identifier" instead and document it if you only need that? I mean, this method only partially implement the findBy anyway (order by, limit, offset, criterias are not fully implemented).

This method should be deprectated and I would prefer to keep it simple. Don't hesitate to say it's used by onboarder v1 btw.

kpAkeneo: yes i can make that if you want ;)

ahocquard (approved):

PR #16189, opened 2022-01-21
PIM-10232: Quick fix on the "Error a new entity was found in the relationship..." error during jobs

nmarniesse:

https://akeneo.atlassian.net/browse/PIM-10232

**What's happening?**

Given a job that needs to iterate on products to make some product updates, it generally uses a PQB that hydrates products per batch of 100, and returns them one by one.
When 100 products are hydrated with doctrine ORM, they are also in the doctrine's Unit of Work.
The job treat them also per batch of 100, and clears the Doctrine UoW after each batch of 100.

So now what's happening if there is a desynchro between ES and MySQL, for example a product is returned by ES but unknown in MySQL?
The ES query returns 100 products, but MySQL returns 99 products. The cursor have 99 products in memory. So when the job ask for the 100th product, the cursor do a new ES query to ask the next 100 products, and begin to return the first one.
The job has now exactly 100 products, so it treats them and clear the UoW. Now the UoW is cleared BUT the cursor still have products in memory (99 in this case). The next time this product will be treated, the UoW will tell it does not know them and throws the "Error a new entity was found in the relationship..." error.

**The fix**

The goal is to fix the cursors to always returns 100 entities (except there is no more results of indeed). That will fix the bugs in several jobs.

Before:
- ES returns 100 identifiers
- We hydrate them but only 99 entities can be found in the DB
- Error later in the jobs because of UOW clear

After:
- ES returns 100 identifiers
- We hydrate them but only 99 entities can be found in the DB
- We ask ES one more identifier
- We hydrate it and now the cursor have 100 entities
- No error in the jobs

The situation is still fragile and we plan to work on a better and cleaner fix. But it fixes quickly the bug and give us some time to work on it. 

The cursors that use "search_after" method are fixed. Other that use "from/to" are not fixed (and we cannot fix it with the same method) but this is not a problem: they are used in API/datagrid use cases, where the bug is not relevant.

- [x] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/tests/back/Pim/Enrichment/Integration/PQB/UpdateProductsWhenElasticsearchIsDesynchronisedIntegration.php
+++ b/tests/back/Pim/Enrichment/Integration/PQB/UpdateProductsWhenElasticsearchIsDesynchronisedIntegration.php
@@ -1,0 +1,84 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of the Akeneo PIM Enterprise Edition.
+ *
+ * (c) 2022 Akeneo SAS (https://www.akeneo.com)
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace AkeneoTest\Pim\Enrichment\Integration\PQB;
+
+use Akeneo\Pim\Enrichment\Component\Product\Model\ProductInterface;
+use Akeneo\Pim\Enrichment\Component\Product\Query\Filter\Operators;
+use Akeneo\Test\Integration\Configuration;
+
+final class UpdateProductsWhenElasticsearchIsDesynchronisedIntegration extends AbstractProductQueryBuilderTestCase
+{
+    protected function getConfiguration(): Configuration
+    {
+        return $this->catalog->useTechnicalCatalog();
+    }
+
+    /**
+     * @test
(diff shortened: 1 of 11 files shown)


tests/back/Pim/Enrichment/Integration/PQB/UpdateProductsWhenElasticsearchIsDesynchronisedIntegration.php
```diff
@@ -0,0 +1,84 @@
...
+            $product->setEnabled(true);
+            $batchedProducts[] = $product;
+
+            if ($batchSize <= \count($batchedProducts)) {
+                $this->get('pim_catalog.saver.product')->saveAll($batchedProducts);
+                $batchedProducts = [];
+                $this->get('pim_connector.doctrine.cache_clearer')->clear();
+            }
+        }
+        if (0 > \count($batchedProducts)) {
```
amigaire, line 61:
```suggestion
        if (0 < \count($batchedProducts)) {
```

nmarniesse: Thanks, it's fixed

amigaire (approved):

RiffVentura (approved):

PR #16195, opened 2022-01-24
Apply association sorting in the versioned normalizer for better compare

Thijzer:
Hi,

We have a case with a small catalog, when importing full file data, the version manager reports that changes have been made every time a product was imported, even if the product is the same. Creating unwanted version history changes. 

image

This change helps to mitigate that problem although maybe more a elegant solution might be possible.

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc


nestorAkeneo:

PR #16199, opened 2022-01-25
PIM-10241: fix user account disabled can connect regression

rfauglas:

An emergency  fix... Please note added code is injected in CustomDaoAuthenticationProvider which is based on a deprecated class since Symfony 5.3

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/src/Akeneo/UserManagement/Bundle/Resources/config/services.yml
+++ b/src/Akeneo/UserManagement/Bundle/Resources/config/services.yml
@@ -44,3 +44,7 @@ services:
             - '%pim_user.account_lock.duration%'
             - '%pim_user.account_lock.max_consecutive_failures%'
             - '%security.authentication.hide_user_not_found%'
+
+    security.user_checker:
+        class: Akeneo\UserManagement\Bundle\Provider\CustomUserChecker
+        abstract: false
--- a/tests/back/UserManagement/Specification/Bundle/Provider/CustomUserCheckerSpec.php
+++ b/tests/back/UserManagement/Specification/Bundle/Provider/CustomUserCheckerSpec.php
@@ -22,11 +22,11 @@ class CustomUserCheckerSpec extends ObjectBehavior
         $this->shouldThrow(DisabledException::class)->duringCheckPreAuth($user);
     }
 
-    function it_throws_exception_if_user_account_is_locked()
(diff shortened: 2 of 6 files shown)


RiffVentura (approved):

tests/back/UserManagement/Specification/Bundle/Provider/CustomUserCheckerSpec.php
```diff
@@ -0,0 +1,32 @@
...
+    }
+
+    function it_throws_exception_if_user_is_disabled()
+    {
+        $user = new User();
+        $user->setEnabled(false);
+        $this->shouldThrow(DisabledException::class)->duringCheckPreAuth($user);
+    }
+
+    function it_throws_exception_if_user_account_is_locked()
```
nmarniesse, line 25: duplicated test?
And maybe add a test to check no exception is thrown for an enabled user

nmarniesse (approved):

tests/back/UserManagement/Specification/Bundle/Provider/CustomUserCheckerSpec.php
rfauglas: Taken into account.

jmleroux (approved):

src/Akeneo/UserManagement/Bundle/Resources/config/services.yml
```diff
@@ -38,3 +38,7 @@ services:
             - '%pim_user.account_lock.duration%'
             - '%pim_user.account_lock.max_consecutive_failures%'
             - '%security.authentication.hide_user_not_found%'
+
+    security.user_checker:
```
StevenVAIDIE, line 42: The best solution is not to make Akeneo\UserManagement\Component\Model\User extends \Symfony\Component\Security\Core\User\InMemoryUser instead of duplicate the https://github.com/symfony/security-core/blob/5.4/User/InMemoryUserChecker.php logic ?

rfauglas: Well I tried to do as little duplication as possible. It seemed to me "less intrusive"... 
But I just checked and indeed InMemoryUser  is a  final class...

PR #16207, opened 2022-01-26
PIM-10240: Fix 500 error to a proper 422 error message

nmarniesse:

https://akeneo.atlassian.net/browse/PIM-10240

When giving null for a price value data, we have a 500 error, because we expect an array. In this PR we now return a 422 error and a good message. 

In our API we don't validate the payload directly. We catch exceptions to return a 422 error when:
- we check duplicate values
- we filter values
- we try to update the product with the values
- we validate the product before saving it

So the validation of the payload format is checked in different places, and some validations are clearly missing.

To handle that I start a PayloadFormatValidator that is responsible to validate the format.
Using a json-schema was a possibility but we cannot specify the error message we want for each case.

- [x] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/src/Akeneo/Pim/Enrichment/Bundle/Controller/ExternalApi/ProductController.php
+++ b/src/Akeneo/Pim/Enrichment/Bundle/Controller/ExternalApi/ProductController.php
@@ -29,6 +29,7 @@ use Akeneo\Pim\Enrichment\Component\Product\Normalizer\ExternalApi\ConnectorProd
 use Akeneo\Pim\Enrichment\Component\Product\ProductModel\Filter\AttributeFilterInterface;
 use Akeneo\Pim\Enrichment\Component\Product\Query\GetConnectorProducts;
 use Akeneo\Pim\Enrichment\Component\Product\Query\ProductQueryBuilderFactoryInterface;
+use Akeneo\Pim\Enrichment\Component\Product\Validator\ExternalApi\PayloadFormat;
 use Akeneo\Pim\Structure\Component\Repository\ExternalApi\AttributeRepositoryInterface;
 use Akeneo\Tool\Bundle\ApiBundle\Cache\WarmupQueryCache;
 use Akeneo\Tool\Bundle\ApiBundle\Checker\DuplicateValueChecker;
@@ -208,7 +209,8 @@ class ProductController
         RemoveParentInterface $removeParent,
         GetProductsWithCompletenessesInterface $getProductsWithCompletenesses,
         LoggerInterface $apiProductAclLogger,
-        SecurityFacade $security
+        SecurityFacade $security,
+        private ValidatorInterface $validator
     ) {
         $this->normalizer = $normalizer;
         $this->channelRepository = $channelRepository;
@@ -386,6 +388,15 @@ class ProductController
         $this->denyAccessUnlessAclIsGranted('pim_api_product_edit');
 
         $data = $this->getDecodedContent($request->getContent());
+        $violations = $this->validator->validate($data, new PayloadFormat());
+        if (0 < $violations->count()) {
+            $firstViolation = $violations->get(0);
+            throw new DocumentedHttpException(
+                Documentation::URL . 'patch_products__code_',
+                sprintf('%s Check the expected format on the API documentation.', $firstViolation->getMessage()),
+                new \LogicException($firstViolation->getMessage())
+            );
+        }
 
         try {
             $this->duplicateValueChecker->check($data);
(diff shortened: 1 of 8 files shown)


tests/back/Pim/Enrichment/Integration/Validation/PayloadFormatValidatorIntegration.php
```diff
@@ -0,0 +1,133 @@
...
+            (string) $violations,
+            \sprintf('Violation message is not found, have: %s', $violations)
+        );
+    }
+
+    /** @test */
+    public function it_adds_violations_when_validate_a_wrong_price_value_format(): void
+    {
+        $payloads = [];
+        $payloads[] = [
```
RiffVentura, line 82: phpUnit dataprovider fits really well here

src/Akeneo/Pim/Enrichment/Bundle/Controller/ExternalApi/ProductController.php
```diff
@@ -386,6 +388,15 @@ public function createAction(Request $request): Response
         $this->denyAccessUnlessAclIsGranted('pim_api_product_edit');
 
         $data = $this->getDecodedContent($request->getContent());
+        $violations = $this->validator->validate($data, new PayloadFormat());
+        if (0 < $violations->count()) {
+            $firstViolation = $violations->get(0);
+            throw new DocumentedHttpException(
+                Documentation::URL . 'patch_products__code_',
```
RiffVentura, line 395: Shouldn't this be :
```suggestion
                Documentation::URL . 'post_products',
```
?

RiffVentura (approved):

tests/back/Pim/Enrichment/Integration/Validation/PayloadFormatValidatorIntegration.php
nmarniesse: I'm not a big fan of dataProvider. Especially because in this case the DB is reset between each test. So the test is 4 times longer

RiffVentura: You are right ! I didn't think it through

src/Akeneo/Pim/Enrichment/Bundle/Controller/ExternalApi/ProductController.php
nmarniesse: Good catch, it's fixed

src/Akeneo/Pim/Enrichment/Component/Product/Validator/ExternalApi/PayloadFormatValidator.php
```diff
@@ -0,0 +1,129 @@
...
+ */
+final class PayloadFormatValidator extends ConstraintValidator
+{
+    private const WRONG_LOCALE_FORMAT = 'Property "%s" expects an array with the key "locale" as string. Check the expected format on the API documentation.';
+    private const WRONG_SCOPE_FORMAT = 'Property "%s" expects an array with the key "scope" as string. Check the expected format on the API documentation.';
+
+    /** @var array<string, string> */
+    private array $attributeTypeByCodes = [];
+
+    public function __construct(private AttributeRepositoryInterface $attributeRepository)
```
rfauglas, line 32: What is the purpose of a constructor that does nothing?

rfauglas (approved):
Difficult to perform a  "significant" rewiew without a deep knowledge of implementation details...  A contextual validator doesn't mean much to me...

src/Akeneo/Pim/Enrichment/Component/Product/Validator/ExternalApi/PayloadFormatValidator.php
nmarniesse: It's the new php 8 capability https://stitcher.io/blog/constructor-promotion-in-php-8

So this code

```php
public function __construct(private AttributeRepositoryInterface $attributeRepository)
{
}
```

is the same than 

```php
private AttributeRepositoryInterface $attributeRepository;

public function __construct(AttributeRepositoryInterface $attributeRepository)
{
    $this->attributeRepository = $attributeRepository;
}
```

rfauglas: Oh yes, i read about it, I had not seen implementation. Nice!

PR #16208, opened 2022-01-26
OCT-4: Do not call the marketplace API when validating a test app

tseho:

When connecting a test App, the code challenge should be handled PIM-side, we shoudn't call the marketplace API.

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/src/Akeneo/Connectivity/Connection/back/Infrastructure/Apps/Validation/CodeChallengeMustBeValidValidator.php
+++ b/src/Akeneo/Connectivity/Connection/back/Infrastructure/Apps/Validation/CodeChallengeMustBeValidValidator.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace Akeneo\Connectivity\Connection\Infrastructure\Apps\Validation;
 
 use Akeneo\Connectivity\Connection\Domain\Apps\DTO\AccessTokenRequest;
+use Akeneo\Connectivity\Connection\Domain\Marketplace\TestApps\Persistence\GetTestAppQueryInterface;
 use Akeneo\Connectivity\Connection\Infrastructure\Marketplace\WebMarketplaceApiInterface;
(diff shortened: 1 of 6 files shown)


guillaumefer-akeneo (approved):

src/Akeneo/Connectivity/Connection/back/Infrastructure/Apps/Validation/CodeChallengeMustBeValidValidator.php
```diff
@@ -44,27 +41,44 @@ public function validate($value, Constraint $constraint)
...
+    }
+
+    private function isCodeChallengeValid(AccessTokenRequest $value): bool
+    {
+        $testApp = $this->getTestAppQuery->execute($value->getClientId());
+
+        if (null !== $testApp) {
+            $secret = $testApp['secret'];
+            $codeIdentifier = $value->getCodeIdentifier();
+            $expectedCodeChallenge = \hash('sha256', $codeIdentifier . $secret);
```
RiffVentura, line 73: Is code challenge is issued the same way for test app ? (I don't quite get why would it the same value)

RiffVentura (approved):

src/Akeneo/Connectivity/Connection/back/Infrastructure/Apps/Validation/CodeChallengeMustBeValidValidator.php
tseho: the formula for a code challenge is `sha256($codeIdentifier . $secret)`.

For real Apps, the PIM doesn't know the secret, so it ask to the Marketplace "Are those client_id, code_identifier, code_challenge valid ?", and the marketplace retrieve the client secret using the client id, thjen does the same encoding and check if the value match.

Here, for test apps, we know the secret, so we can check if code challenge is valid directly without calling the marketplace.

PR #16250, opened 2022-02-04
BH-986: Prepare v6.0.0 release

jmleroux:


ValentinMumble (approved):

vgelot (approved):

Airmanbzh (approved):

tests/back/Platform/Specification/CommunityVersionSpec.php
```diff
@@ -1,22 +0,0 @@
-<?php
-
-namespace Specification\Akeneo\Platform;
-
-use Akeneo\Platform\CommunityVersion;
-use PhpSpec\ObjectBehavior;
-use PHPUnit\Framework\Assert;
-
-class CommunityVersionSpec extends ObjectBehavior
```
StevenVAIDIE, line 9: Why do you remove this file ? For me you should keep it by asserting 6.0.0 instead. When we are going to pull up 5.0 we will have conflict no ?

jmleroux: I can revert it to test the version name and a version pattern, but not a fixed version because it will move with tags

jmleroux: Like here: https://github.com/akeneo/pim-enterprise-dev/pull/12858/commits/bdf2e95691c179b560aaffab1e9547ccec083163

PR #16264, opened 2022-02-07
PIM-10264: Optimize variant product ratio query

ghoussard:

Description of the issue and the fix here : https://akeneo.atlassian.net/browse/PIM-10264?atlOrigin=eyJpIjoiYjdlNzVkOWJmMjU4NGJmNTgwMzE0N2M3ZDMwYWM0YTYiLCJwIjoiaiJ9

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/src/Akeneo/Pim/Enrichment/Bundle/Storage/Sql/Completeness/VariantProductRatio.php
+++ b/src/Akeneo/Pim/Enrichment/Bundle/Storage/Sql/Completeness/VariantProductRatio.php
@@ -53,7 +53,7 @@ final class VariantProductRatio implements VariantProductRatioInterface
         return <<<SQL
                 FROM pim_catalog_product_model AS root_product_model
                 INNER JOIN pim_catalog_product_model as sub_product_model ON sub_product_model.parent_id = root_product_model.id
-                INNER JOIN pim_catalog_product product ON product.product_model_id = sub_product_model.id
+                INNER JOIN pim_catalog_product product ON product.product_model_id = sub_product_model.id AND product.product_model_id IS NOT NULL
 SQL;
     }
 
@@ -61,7 +61,7 @@ SQL;
     {
         return <<<SQL
                 FROM pim_catalog_product_model AS root_product_model
-                INNER JOIN pim_catalog_product product ON product.product_model_id = root_product_model.id
+                INNER JOIN pim_catalog_product product ON product.product_model_id = root_product_model.id AND product.product_model_id IS NOT NULL
 SQL;
     }
 
@@ -82,7 +82,7 @@ SQL;
                 %s
                 INNER JOIN pim_catalog_completeness completeness ON product.id = completeness.product_id
                 WHERE
-                    root_product_model.id IS NOT NULL AND root_product_model.id = :root_product_model_id
+                    root_product_model.id = :root_product_model_id
               ) AS product
             INNER JOIN pim_catalog_locale locale ON locale.id = product.locale_id
             INNER JOIN pim_catalog_channel channel ON channel.id = product.channel_id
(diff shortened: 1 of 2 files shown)


src/Akeneo/Pim/Enrichment/Bundle/Storage/Sql/Completeness/VariantProductRatio.php
```diff
@@ -53,15 +53,15 @@ private function joinToProductWithTwoLevels(): string
...
                 INNER JOIN pim_catalog_product_model as sub_product_model ON sub_product_model.parent_id = root_product_model.id
-                INNER JOIN pim_catalog_product product ON product.product_model_id = sub_product_model.id
+                INNER JOIN pim_catalog_product product ON product.product_model_id = sub_product_model.id AND product.product_model_id IS NOT NULL
 SQL;
     }
 
     private function joinToProductWithOneLevel(): string
     {
         return <<<SQL
                 FROM pim_catalog_product_model AS root_product_model
```
ahocquard, line 63: useless for this function, as Mysql can directly inferred that product.product_model_id is not null due to equals condition with the table "pim_catalog_product_model" where you apply a filter `=`.

src/Akeneo/Pim/Enrichment/Bundle/Storage/Sql/Completeness/VariantProductRatio.php
```diff
@@ -53,15 +53,15 @@ private function joinToProductWithTwoLevels(): string
...
                 INNER JOIN pim_catalog_product_model as sub_product_model ON sub_product_model.parent_id = root_product_model.id
-                INNER JOIN pim_catalog_product product ON product.product_model_id = sub_product_model.id
+                INNER JOIN pim_catalog_product product ON product.product_model_id = sub_product_model.id AND product.product_model_id IS NOT NULL
```
ahocquard, line 56: It's good to add a comment to explain the why when you have a tricky fix.
In Phpdoc for example:

```
At first, the `product_model_id` filtering on to check that is not null seems to be irrelevant. Actually, it is useful because Mysql has not enough information to determine itself that `product_model_id` is not null.

Therefore, if there are many empty values for `product_model_id` in the database (= many simple products), it does not use the index on parent id when looking for a variant product. In such case, Mysql think that filtering on `product_model_id` is not a good filter and prefer to perform a very expensive Full Table Scan on the product table.

See https://dev.mysql.com/doc/refman/8.0/en/index-statistics.html
Another option would be to use `innodb_stats_method=nulls_unequal` but it can have many side effects.
```

SamirBoulil (approved):

ahocquard (approved):
GTM after fix

src/Akeneo/Pim/Enrichment/Bundle/Storage/Sql/Completeness/VariantProductRatio.php
ghoussard: I don't know why but an error 500 occurs later in php if I remove this condition... 😕 
`│ pim-web-779db468d8-zsgfc:fpm {"message":"An error occured while trying to fetch the measurement family of a metric value to present it.","context":{"akeneo_context":"Enrichment","path_in │
│ fo":"https://insight-sandbox-sds-22101.dev.cloud.akeneo.com/enrich/product-model/rest/10"},"level":300,"level_name":"WARNING","channel":"app","datetime":{"date":"2022-02-09 13:00:36.9011 │
│ 53","timezone_type":3,"timezone":"Etc/UTC"},"extra":[]} `

PR #16289, opened 2022-02-11
RAB: Fix service API does not return ContraintValidation when attribute regex is not satisfied

StevenVAIDIE:

We tried to set an regex validation on an attribute, then enrich it with the new service API => we didn't have exception (no violation).

We tried to resave the product without changing any data and we have an error. There is a product validator, with weird Guesser that check those things so we used this validator. (same than the used in the ProductController)
 
- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc


nmarniesse (approved):
:+1:

PR #16302, opened 2022-02-14
CPM-503-BH: define  root connection

rfauglas:
@pierallard,  this the  option i see: define a root connection.
I'm not sure if it is acceptable to leave a root connection in  terms of security... This might be managed by the cron that will run the command... If this command is run as a cron job, it is possible to inject specific environment variables... 

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc


config/packages/doctrine.yml
```diff
@@ -18,6 +18,21 @@ doctrine:
                 server_version: '8.0'
                 mapping_types:
                     json: string
+            root_connection:
```
nmarniesse, line 21: I'm not a big fan of adding a root connection that can be used by any part of the code. Only migrations should be able to use it. In term of security we should do better IMHO

Is it possible to always use one connection, and inject different values only for the container that run the migration? In order to inject root values in migration jobs (job not exposed to the public) and default values for the web (exposed)

rfauglas: Indeed, this root password  should be injected in migration jobs only. 
I pinged again @Airmanbzh yesterday to remind this requirement, which is expressed here: https://www.notion.so/akeneo/Black-hawks-ADR-uuid-8f3f308cffd04ed7ae5d8fac9a4fe3d8#c611cffa7f4943339b9164387a2cc6c3

PR #16314, opened 2022-02-16
PIM-10281: Update Circle CI machine executors

jmleroux:
CircleCI’s Ubuntu 14.04 and 16.04 images will be permanently unavailable as of May 31, 2022. Migrate to a newer image today.

https://circleci.com/blog/ubuntu-14-16-image-deprecation/

--- EDIT ---

After review from teammates (thank you @StevenVAIDIE ), I rewrote the Circle CI config to change the machine executor to make it easier to change in the following PR.
I'd love to have only one occurence of the alias array to not repeat it in several files, but it is not supported by the `yq` merge operation

### DIFF
--- a/.circleci/jobs/build_and_deploy.yml
+++ b/.circleci/jobs/build_and_deploy.yml
@@ -12,7 +12,7 @@ aliases:
                   username: _json_key  # default username when using a JSON key file to authenticate
                   password: $GCLOUD_SERVICE_KEY_DEV  # JSON service account you created, do not encode to base64
 
-    - &executor-machine: ubuntu-2004:202111-02
+executor-machine: &executor-machine 'ubuntu-2004:202111-02'
 
 commands:
     set_gcloud_config_dev:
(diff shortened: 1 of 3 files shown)


StevenVAIDIE (approved):
What about occurrence in build_and_deploy.yml and connectivity.yml ? There is some ubuntu-2004:202101-01 it could be nice to have always the same image

jmleroux:
> What about occurrence in build_and_deploy.yml and connectivity.yml ? There is some ubuntu-2004:202101-01 it could be nice to have always the same image

I'll try to mutualize the executor machine.I think i remember seeing something like that in another Akeneo repo (Onbarder)

ahocquard (approved):
Could we avoid to use alias? I will block me when splitting the CI (alias is not supported across different yaml files)

StevenVAIDIE (approved):

jmleroux:
> EDIT: I didn't read the description with enough attention, it's ok

> but it is not supported by the yq merge operation

It's not supported out of the box, but I found some scripting workarounds. They are a bit tricky and not really appropriate for this hotfix PR, but it could be done in your work to split workflows.

PR #16315, opened 2022-02-16
CXP-953: remove acceptance tests

guillaumefer-akeneo:

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc


tseho (approved):

tseho:
Using the command `make connectivity-connection-coverage`, can you compare the total code coverage generated in the directory `coverage/Connectivity/Back/Global/` BEFORE and AFTER your PR ?

It would be interesting to know if we lost coverage, and where.

RiffVentura (approved):
Could you execute a full CI cycle just to be safe
Aren't there, InMemory doubles that should be deleted as a result?

PR #16335, opened 2022-02-18
PIM-10288: Fix product associations came duplicated in API

pierrejolly:
When you add an association on a product model and its variant the associations come up twice in the API. This PR attempt to fix this behavior


SamirBoulil (dismissed):
I think we have some tests for those query. Did you try adding the duplicated case to check if that works ?

pierrejolly:
Yes @SamirBoulil I have tried these duplicated case and everything seems ok now

RiffVentura (approved):

StevenVAIDIE (approved):

PR #16353, opened 2022-02-21
PIM-10268: Fix identifier filter should not be displayed if user doesn't choose to display it

StevenVAIDIE:

In this PR: 
- I removed the fact that the identifier is always displayed
- I added a listener in order to add the filter into the default list of filters (as the identifier can be other than sku we cannot add it manually in the grid config) 
 
- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc


StevenVAIDIE:
refresh

ghoussard (approved):

SamirBoulil (approved):

PR #16364, opened 2022-02-22
PIM-10293 add batch-size option to pim completness command

MadeWilson:


RiffVentura (approved):
I'm not sure if it's really addressing the OutOfMemory issue.
This batches identifiers retrieval only
Shouldn't it batch completeness and indexation as well?

StevenVAIDIE (approved):

PR #16366, opened 2022-02-22
CPM-499: add acceptance tests for Upser Product service API

nmarniesse:

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/src/Akeneo/Channel/back/tests/Acceptance/InMemory/InMemoryIsLocaleEditable.php
+++ b/src/Akeneo/Channel/back/tests/Acceptance/InMemory/InMemoryIsLocaleEditable.php
@@ -26,7 +26,7 @@ final class InMemoryIsLocaleEditable implements IsLocaleEditable
      */
     public function forUserId(string $localeCode, int $userId): bool
     {
-        /** @var UserInterface $user */
+        /** @var UserInterface|null $user */
         $user = $this->userRepository->findOneBy(['id' => $userId]);
         if (null === $user) {
             return false;
--- a/src/Akeneo/Channel/back/tests/phpstan.neon.dist
+++ b/src/Akeneo/Channel/back/tests/phpstan.neon.dist
@@ -4,4 +4,5 @@ parameters:
     - %rootDir%/../../../src/Akeneo/Channel
   excludePaths:
     - %rootDir%/../../../src/Akeneo/Channel/back/tests/Specification
+    - %rootDir%/../../../src/Akeneo/Channel/back/tests/Acceptance/Specification
     - %rootDir%/../../../src/Akeneo/Channel/back/Infrastructure


src/Akeneo/Channel/Locale/back/Test/Acceptance/InMemory/InMemoryGetEditableLocaleCodes.php
```diff
@@ -0,0 +1,51 @@
...
+            $ownedCategoryCodes = \array_merge(
+                $ownedCategoryCodes,
+                $this->editableLocalesPerUserGroupName[$groupName] ?? []
+            );
+        }
+
+        return \array_unique($ownedCategoryCodes);
+    }
+
+    public function addOwnedCategoryCode(string $groupName, string $localeCode): void
```
nmarniesse, line 47: Usually we use a repository to store the information, and the in memory query use the repository. But in this case the repository is in EE (LocaleAccessRepository) and the methods of this repository don't help us for this query.  
So I decided to do simpler and add this method

tests/back/Acceptance/User/InMemoryGroupRepository.php
```diff
@@ -18,8 +18,8 @@
  */
 class InMemoryGroupRepository implements GroupRepositoryInterface, SaverInterface, ObjectRepository
```
SamirBoulil, line 19: Are you using those in your tests ? I don't think you should rely on those global inMemory implementations but rather keep a copy in your new BC?

nmarniesse: We discussed about that this yesterday with @ahocquard . And we need to discuss with other people to find a proper way to write tests. To be explain our thoughts shortly:

- copy/paste in memories in our BC is better to isolate the BC, and launch dedicated CI. But with this, the business is duplicated in several BC. For example, if the InMemoryGroupRepository is duplicated in Enrichment, Asset, ... It means all the teams need to know and maintain the business of user/group. If the user/group business changes, all teams need to fix their own in memory implementation!
- another solution can be: the team that owns the user/group BC provides the real implementation (public API) **and** the in memory. This way only one team maintain the user/group business. But in another hand, the tests of the BC are not isolated.

src/Akeneo/Pim/Enrichment/Product/back/Test/Feature/upsert.feature
```diff
@@ -0,0 +1,57 @@
...
+  Scenario: Create an empty product with unknown user
+    When an unknown user tries to upsert the "foo" product
+    Then there is a violation saying the user is unknown
+
+  Scenario: Create an empty product with empty string identifier
+    When the "julia" user upserts the "" product
+    Then there is a violation with message: The product identifier requires a non empty string
+
+  Scenario: Create a product with text attribute value
+    Given a set text value intent on the "a_text-null-null" attribute with the "test" text value
```
pierallard, line 33: I think this -null-null a little bit weird in a feature file.
But I saw the function parsing this, and there is no beautiful way to know when you write a_text-toto if toto is a channel or locale, and you way of coding this prevent it.
If you have a good idea to remove the -null items, feel free, if not, TANPI.

src/Akeneo/Pim/Enrichment/Product/back/Test/Acceptance/Context/ProductContext.php
```diff
@@ -0,0 +1,122 @@
...
+    public function thereIsAMessageSayingTheUserIsUnknown(): void
+    {
+        $this->constraintViolationsContext->thereIsAViolationWithMessage('The "-10" user does not exist');
+    }
+
+    private function upsertProduct(UpsertProductCommand $command): void
+    {
+        $violations = $this->validator->validate($command);
+        $this->constraintViolationsContext->add($violations);
+        // @TODO later: call the handler (we cannot do that now because legacy validation is not in memory)
```
pierallard, line 108: Can you add a card in the backlog to not forget this plz ?

pierallard (approved):

src/Akeneo/Pim/Enrichment/Product/back/Test/Feature/upsert.feature
nmarniesse: it's fixed

cemorin (approved):

PR #16405, opened 2022-02-25
RAC-1215: fix error when attributeOption doesn't have label

MadeWilson:


src/Akeneo/Pim/Structure/Bundle/Query/PublicApi/AttributeOption/Sql/SqlSearchAttributeOptions.php
```diff
@@ -95,7 +95,7 @@ private function getAttributeOptions(
         return array_map(
             static fn (array $attributeOption) => new AttributeOption(
                 $attributeOption['code'],
-                json_decode($attributeOption['labels'], true),
+                null !== $attributeOption['labels'] ? json_decode($attributeOption['labels'], true) : [],
```
pierrejolly, line 98: Shouldn't we return the code between brackets when the label is empty?

MadeWilson: no the code is already given no matter what. In our component we have a method getLabel() that return the code if no label exist

ghoussard (approved):
Could you add a test case in tests/back/Pim/Structure/Integration/AttributeOption/SqlSearchAttributeOptionsIntegration.php to cover this case please ?

MadeWilson:
> Could you add a test case in tests/back/Pim/Structure/Integration/AttributeOption/SqlSearchAttributeOptionsIntegration.php to cover this case please ?

sure

StevenVAIDIE (approved):

PR #16404, opened 2022-02-25
CPM-510: Add AttributeValueShouldBeEditable constraint and validator

cemorin:
**Description (for Contributor and Core Developer)**
Add validation on attribute group permission for UpserProductCommand userValueIntents (always authorized on CE)

**Definition Of Done (for Core Developer only)**

- [x] Tests
- [ ] PM Validation (Story)

PR #16406, opened 2022-02-25
CPM-527: Fixes

nmarniesse:

This PR fixes:

- one path with valueUserIntents in the UpsertProductHandler. Others are fixed in this [PR](https://github.com/akeneo/pim-community-dev/pull/16404/files)
- A LegacyViolationException is turned into ViolationException in the handler
- Replace the GetEditableLocaleCodes interface by IsLocaleEditable. The advantage in CE is to avoid a SQL query: the implem always returns true instead of doing a query to return all the activated locales.


cemorin (approved):

src/Akeneo/Pim/Enrichment/Product/back/Application/UpsertProductHandler.php
```diff
@@ -139,12 +139,12 @@ private function updateProduct(ProductInterface $product, UpsertProductCommand $
                         $e->getMessage(),
                         [],
                         $command,
-                        "valueUserIntent[$index]",
+                        "valueUserIntents[$index]",
```
pierrejolly, line 142: I think this property path is incorrect

pierrejolly (approved):

PR #16418, opened 2022-02-28
CPM-539: use update instead of "insert on duplicate key update"

nmarniesse:
Before we did a select query, then a "insert on duplicate key update". The problem is some items can be removed between this 2 statements. In this case the 2nd query will insert bad rows. Even if the case would probably never happen, we could not ensure it will not happen.
To prevent this, we now use update

### DIFF
--- a/src/Akeneo/Pim/Enrichment/Bundle/Command/MigrateToUuid/MigrateToUuidFillJson.php
+++ b/src/Akeneo/Pim/Enrichment/Bundle/Command/MigrateToUuid/MigrateToUuidFillJson.php
@@ -86,7 +86,7 @@ class MigrateToUuidFillJson implements MigrateToUuidStep
         return $allItemsMigrated;
     }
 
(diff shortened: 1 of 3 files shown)


pierallard (approved):
Perfect, but 2 notices
- the updateProductModelAssociations and updateProductAssociations are now the same but 1 line. Is it a good idea to re-merge them as they were before, to have only 1 method ?
- If it's not relou for you, can you do the same for the Add Uuid subscriber ? If not, no problem.

ahocquard (commented):
How does it fix the issue? The associations are still passed as function parameters :thinking:

nmarniesse:
> How does it fix the issue? The associations are still passed as function parameters thinking

The issue concerns the insert in the product table, not the json associations.  

In the json associations, the products can still be removed just before the save query. This is not a big deal because product associations are never cleaned up when a product/product model is removed

pierallard (approved):

PR #16472, opened 2022-03-04
CPM-530: Add SetSimpleSelectValue and SetMultiSelectValue user intents

Balamung:

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/src/Akeneo/Pim/Enrichment/Product/back/API/Command/UserIntent/SetSimpleSelectValue.php
+++ b/src/Akeneo/Pim/Enrichment/Product/back/API/Command/UserIntent/SetSimpleSelectValue.php
@@ -1,0 +1,44 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of the Akeneo PIM Enterprise Edition.
+ *
+ * (c) 2022 Akeneo SAS (https://www.akeneo.com)
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Akeneo\Pim\Enrichment\Product\API\Command\UserIntent;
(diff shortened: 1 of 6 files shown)


src/Akeneo/Pim/Enrichment/Product/back/API/Command/UserIntent/SetSimpleSelectValue.php
```diff
@@ -0,0 +1,40 @@
...
+
+declare(strict_types=1);
+
+/**
+ * @copyright 2022 Akeneo SAS (https://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+namespace Akeneo\Pim\Enrichment\Product\API\Command\UserIntent;
+
+class SetSimpleSelectValue implements ValueUserIntent
```
nmarniesse, line 11: What happen if the value is en empty string? The value is cleared or there is an error saying the option '' does not exist?
I would add a validation to check the value is not empty (symfony validation)

Same for SetMultiSelectValue

nmarniesse (approved):

src/Akeneo/Pim/Enrichment/Product/back/API/Command/UserIntent/SetSimpleSelectValue.php
Balamung: I expected the "option does not exist", but it does indeed clear the value. Shouldn't this be let as a possibility ? The SetTextValue has the same behavior for instance ?

cemorin (approved):

PR #16475, opened 2022-03-04
CPM-532: add Family user intent

nmarniesse:

Add user intents about families:

- SetFamily
- RemoveFamily


src/Akeneo/Pim/Enrichment/Product/back/Application/UpsertProductHandler.php
```diff
@@ -153,5 +151,30 @@ private function updateProduct(ProductInterface $product, UpsertProductCommand $
                 throw new \InvalidArgumentException(\sprintf('The "%s" intent cannot be handled.', get_class($valueUserIntent)));
             }
         }
+
+        $familyUserIntent = $command->familyUserIntent();
+        if (null !== $familyUserIntent) {
+            if ($familyUserIntent instanceof SetFamily || $familyUserIntent instanceof RemoveFamily) {
```
pierallard, line 157: Why this condition is not done in the validator part ?
For me, it complexify a little bit this part, to throw InvalidArgumentException or ViolationException in the same "bout de code".

pierallard (approved):

src/Akeneo/Pim/Enrichment/Product/back/Application/UpsertProductHandler.php
nmarniesse: All family user intents should implement the FamilyUserIntent interface. And all classes that implements this interface must be processed by the handler. Otherwise it's an error (= we forgot to implement the handler).  
Adding a validator does not guarantee the handler could process it

PR #16551, opened 2022-03-14
CXP-1143: hasOutdatedScopes migration

RiffVentura:
Added `hasOutdatedScopes` to `akeneo_connectivity_connected_app` table as TINYINT boolean that default to false

### DIFF
--- a/upgrades/schema/Version_7_0_20220314113925_add_has_outdated_scopes_to_connected_app_table.php
+++ b/upgrades/schema/Version_7_0_20220314113925_add_has_outdated_scopes_to_connected_app_table.php
@@ -11,11 +11,26 @@ final class Version_7_0_20220314113925_add_has_outdated_scopes_to_connected_app_
 {
     public function up(Schema $schema): void
     {
+        $table = $schema->getTable('akeneo_connectivity_connected_app');
+        if ($table->hasColumn('has_outdated_scopes')) {
+            $this->disableMigrationWarning();
+            return;
+        }
+
         $this->addSql("ALTER TABLE akeneo_connectivity_connected_app ADD has_outdated_scopes TINYINT DEFAULT 0 NOT NULL");
     }
 
     public function down(Schema $schema): void
     {
         $this->throwIrreversibleMigrationException();
+    }
+
+    /**
+     * Function that does a non altering operation on the DB using SQL to hide the doctrine warning stating that no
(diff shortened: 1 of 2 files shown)


willy-ahva (approved):

upgrades/schema/Version_7_0_20220314113925_add_has_outdated_scopes_to_connected_app_table.php
```diff
@@ -0,0 +1,21 @@
...
+{
+    public function up(Schema $schema): void
+    {
+        $this->addSql("ALTER TABLE akeneo_connectivity_connected_app ADD has_outdated_scopes TINYINT DEFAULT 0 NOT NULL");
+    }
+
+    public function down(Schema $schema): void
+    {
+        $this->throwIrreversibleMigrationException();
+    }
```
masacc, line 20:
```suggestion

    public function up(Schema $schema): void
    {
        $table = $schema->getTable('akeneo_connectivity_connected_app');
        if ($table->hasColumn('has_outdated_scopes')) {
            $this->disableMigrationWarning();

            return;
        }

        $this->addSql("ALTER TABLE akeneo_connectivity_connected_app ADD has_outdated_scopes TINYINT DEFAULT 0 NOT NULL");
    }

    public function down(Schema $schema): void
    {
        $this->throwIrreversibleMigrationException();
    }

    /**
     * Function that does a non altering operation on the DB using SQL to hide the doctrine warning stating that no
     * sql query has been made to the db during the migration process.
     */
    private function disableMigrationWarning(): void
    {
        $this->addSql('SELECT 1');
    }
```

we use this type of migration once before to not fail when executed multiple times (I don't remember when it could happen)

masacc (approved):

upgrades/schema/Version_7_0_20220314113925_add_has_outdated_scopes_to_connected_app_table.php
RiffVentura: Probably related with CI  and/or tests executed multiple times

masacc: No I think it's for some kind of installation/migration process on some client instances...

masacc (approved):

PR #16581, opened 2022-03-16
RAC-1214: [Shared Catalogs] Displayed logo not resized

MadeWilson:
In this PR we changed how the image fit in the div 

**Before**
Screenshot from 2022-03-21 15-10-17

**After**
Screenshot from 2022-03-21 15-09-33

### DIFF
--- a/front-packages/akeneo-design-system/src/components/Input/MediaFileInput/MediaFileInput.tsx
+++ b/front-packages/akeneo-design-system/src/components/Input/MediaFileInput/MediaFileInput.tsx
@@ -89,6 +89,12 @@ const UploadProgress = styled(ProgressBar)`
 `;
 
 const MediaFileImage = styled(Image)`
+  max-width: 95%;
+  max-height: 95%;
+  position: absolute;
+  top: 50%;
+  left: 50%;
+  transform: translate(-50%, -50%);
   border: none;
 `;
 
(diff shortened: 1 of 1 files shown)


ghoussard (approved):
Don't forget the CHANGELOG 👍🏻 
Edit: My bad, for the DSM, no CHANGELOG needed

front-packages/akeneo-design-system/src/components/Input/MediaFileInput/MediaFileInput.tsx
```diff
@@ -272,9 +282,7 @@ const MediaFileInput = React.forwardRef<HTMLInputElement, MediaFileInputProps>(
           </>
         ) : null !== value ? (
           <>
-            <MediaFileImage
-              height={isCompact ? 47 : 120}
-              width={isCompact ? 47 : 120}
+            <PreviewImage
```
StevenVAIDIE, line 285: The following code should do the work no ? Doing some tricky translation and using percentage seem to be overkill and dangerous
```suggestion
            <MediaFileImage
              height={isCompact ? 47 : 120}
              width={isCompact ? 47 : 120}
              fit="contain"
```

MadeWilson: with that solution it will fit in a square of 120*120, not sure that's what Felix had in mind. Dont see the danger in % since MediaFileInputContainer wich is the parent had it size define in absolute value.

StevenVAIDIE: I think it's was Stephane want (mockup of Stephane https://www.notion.so/akeneo/File-3d6b2dfaee734e7cb1247aab485efd98).

By doing the %, now the file name is not correctly placed
Before:

Screenshot from 2022-03-17 11-22-32

Now:
Screenshot from 2022-03-17 11-22-18

PR #16596, opened 2022-03-17
CXP-1120 Added update scope endpoint to flag connected app

RiffVentura:
Added endpoint to flag connected app with scope changes

### DIFF
--- a/config/packages/security.yml
+++ b/config/packages/security.yml
@@ -44,6 +44,13 @@ security:
             pattern:                        ^/connect/apps/v1/openid/public-key
             security:                       false
 
+        app_scopes_update:
+            pattern:                        ^/connect/apps/v1/scopes/update
+            fos_oauth:                      true
+            stateless:                      true
+            access_denied_handler:          pim_api.security.access_denied_handler
+            provider:                       pim_user_api
+
         api_index:
             pattern:                        ^/api/rest/v1$
             security:                       false
--- a/src/Akeneo/Connectivity/Connection/back/Infrastructure/Apps/Controller/Public/ScopesUpdateAction.php
+++ b/src/Akeneo/Connectivity/Connection/back/Infrastructure/Apps/Controller/Public/ScopesUpdateAction.php
@@ -7,6 +7,7 @@ namespace Akeneo\Connectivity\Connection\Infrastructure\Apps\Controller\Public;
 use Akeneo\Connectivity\Connection\Application\Apps\Command\FlagAppContainingOutdatedScopesCommand;
 use Akeneo\Connectivity\Connection\Application\Apps\Command\FlagAppContainingOutdatedScopesHandler;
 use Akeneo\Connectivity\Connection\Domain\Apps\Persistence\FindOneConnectedAppByUserIdQueryInterface;
+use Akeneo\UserManagement\Component\Model\UserInterface;
 use Symfony\Component\HttpFoundation\JsonResponse;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
@@ -17,7 +18,7 @@ use Symfony\Component\Security\Core\Authentication\Token\Storage\TokenStorageInt
  * @copyright 2022 Akeneo SAS (http://www.akeneo.com)
  * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
  */
-class ScopesUpdateAction
+final class ScopesUpdateAction
 {
     public function __construct(
         private TokenStorageInterface $tokenStorage,
@@ -28,6 +29,7 @@ class ScopesUpdateAction
 
     public function __invoke(Request $request): Response
     {
+        /** @var UserInterface|null $user */
         $user = $this->tokenStorage->getToken()?->getUser();
         if (null === $user) {
             throw new AccessDeniedHttpException('Not an authenticated App');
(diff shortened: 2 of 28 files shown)


src/Akeneo/Connectivity/Connection/back/Infrastructure/Apps/Controller/Public/ScopesUpdateAction.php
```diff
@@ -0,0 +1,50 @@
...
+        if (null === $connectedApp) {
+            throw new AccessDeniedHttpException('Not an authenticated App');
+        }
+
+        $this->flagAppContainingOutdatedScopesHandler->handle(new FlagAppContainingOutdatedScopesCommand(
+            $connectedApp,
+            $request->get('scopes', ''),
+        ));
+
+        return new JsonResponse('Ok');
```
masacc, line 48: I think we could return a 204 with an empty body, what do you think ?

src/Akeneo/Connectivity/Connection/back/Infrastructure/Apps/Controller/Public/ScopesUpdateAction.php
```diff
@@ -0,0 +1,50 @@
...
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\HttpKernel\Exception\AccessDeniedHttpException;
+use Symfony\Component\Security\Core\Authentication\Token\Storage\TokenStorageInterface;
+
+/**
+ * @copyright 2022 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+final class ScopesUpdateAction
```
masacc, line 21: I think we should name the Actions with an imperative verb first, something like `UpdateScopesAction` or `FlagConnectedAppWithOutdatedScopesAction`

masacc (approved):

src/Akeneo/Connectivity/Connection/back/Infrastructure/Apps/Controller/Public/ScopesUpdateAction.php
```diff
@@ -0,0 +1,50 @@
...
+    public function __construct(
+        private TokenStorageInterface $tokenStorage,
+        private FindOneConnectedAppByUserIdQueryInterface $findOneConnectedAppByUserIdQuery,
+        private FlagAppContainingOutdatedScopesHandler $flagAppContainingOutdatedScopesHandler,
+    ) {
+    }
+
+    public function __invoke(Request $request): Response
+    {
+        /** @var UserInterface|null $user */
```
tseho, line 32: You shouldn't change the static type of the variable.
As you were already aware, $user is not always `Akeneo\UserManagement\Component\Model\UserInterface`.
In reality, the result is an object with the interface `Symfony\Component\Security\Core\User\UserInterface`.

In your code here, in the future, what happen if the logged-in user is using another class implementing `Symfony\Component\Security\Core\User\UserInterface` but not `Akeneo\UserManagement\Component\Model\UserInterface` ?
You will have a bug hard to debug because it will throw later for unexpected reasons.

That's why we should not change the type returned by a method, but instead check that you receive what you expect.
see https://github.com/akeneo/pim-community-dev/blob/master/src/Akeneo/Connectivity/Connection/back/Infrastructure/Apps/OAuth/CreateAccessToken.php#L98-L100

src/Akeneo/Connectivity/Connection/back/Infrastructure/Apps/Controller/Public/ScopesUpdateAction.php
tseho: I think the 200 response is ok.

src/Akeneo/Connectivity/Connection/back/Infrastructure/Apps/Controller/Public/ScopesUpdateAction.php
```diff
@@ -0,0 +1,50 @@
...
+        }
+
+        $connectedApp = $this->findOneConnectedAppByUserIdQuery->execute($user->getId());
+        if (null === $connectedApp) {
+            throw new AccessDeniedHttpException('Not an authenticated App');
+        }
+
+        $this->flagAppContainingOutdatedScopesHandler->handle(new FlagAppContainingOutdatedScopesCommand(
+            $connectedApp,
+            $request->get('scopes', ''),
```
tseho, line 45:
```suggestion
            $request->query->get('scopes', ''),
```

Since symfony 5.4, it's deprecated to use the magic `$request->get()` method, it's better to use the specific InputBag.

tseho (approved):

src/Akeneo/Connectivity/Connection/back/Infrastructure/Apps/Controller/Public/ScopesUpdateAction.php
RiffVentura: I was unsure what exactly should be returned here. Either way is fine by me

config/packages/security.yml
```diff
@@ -44,6 +44,13 @@ security:
             pattern:                        ^/connect/apps/v1/openid/public-key
             security:                       false
 
+        app_scopes_update:
+            pattern:                        ^/connect/apps/v1/scopes/update
+            fos_oauth:                      true
+            stateless:                      true
+            access_denied_handler:          pim_api.security.access_denied_handler
+            provider:                       pim_user_api
```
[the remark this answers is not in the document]
RiffVentura: [Done](https://github.com/akeneo/pim-enterprise-dev/pull/13243)

PR #16607, opened 2022-03-18
PIM-10345: Having a full numerical attribute code cause an error on product model import

MadeWilson:
In this PR: 

I fixed the following error : When user import a product with a numeric only attribute name, we have a fatal error with the following message: `preg_match(): #2 ($subject) must be of type string, int given.`

**Why we didn't have this error message in product import ?** 
Because src/Akeneo/Pim/Enrichment/Component/Product/Connector/ArrayConverter/FlatToStandard/Product.php doesn't have `declare(strict_types=1);` so the cast is automatically done by PHP.

Current PHPStan level is 2 on enrichment, only level 5 can catch this error. 

Instead of removing the `declare(strict_types=1);` i manually casted the variable

### DIFF
--- a/tests/back/Pim/Enrichment/Specification/Component/Product/Connector/ArrayConverter/FlatToStandard/ProductModelSpec.php
+++ b/tests/back/Pim/Enrichment/Specification/Component/Product/Connector/ArrayConverter/FlatToStandard/ProductModelSpec.php
@@ -2,6 +2,7 @@
 
 namespace Specification\Akeneo\Pim\Enrichment\Component\Product\Connector\ArrayConverter\FlatToStandard;
 
+use Akeneo\Tool\Component\Connector\Exception\ArrayConversionException;
 use PhpSpec\ObjectBehavior;
 use Akeneo\Tool\Component\Connector\ArrayConverter\ArrayConverterInterface;
 use Akeneo\Tool\Component\Connector\ArrayConverter\FieldsRequirementChecker;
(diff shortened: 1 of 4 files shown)


SamirBoulil (approved):

tests/back/Pim/Enrichment/Specification/Component/Product/Connector/ArrayConverter/FlatToStandard/ProductModelSpec.php
```diff
@@ -2,6 +2,7 @@
 
 namespace Specification\Akeneo\Pim\Enrichment\Component\Product\Connector\ArrayConverter\FlatToStandard;
 
+use Akeneo\Tool\Component\Connector\Exception\ArrayConversionException;
```
ghoussard, line 5: Unused import

ghoussard (approved):
Don't forget the CHANGELOG

BitOne:
Your description is missing.

It would be necessary to have more insight on when this variable became an integer. Here you're treating the symptoms (checking if it's an integer and cast it to string). We must fix the root cause before anything. And only resorting on treating the symptoms when we don't have the choice.

So an analysis of the why is required here in your description.

Moreover, can you add more tests please, in order to cover a wider range of risks, please?

For example:
 - float numbers
 - string beginning with a number
 - real string but with only numbers in it (here you have an int)
 - real string with a float in them
 
 But the tests should really be done where the root issue is really.

BitOne (commented):
See comment above.

PR #16626, opened 2022-03-21
PLG-819 fix(Data quality insights) PEF quality score for product model 'A' misaligned with label

FSAkeneo:

Product edition form :  quality score for product model 'A' was misaligned with label 'Quality Score:'

- [ ] PM Validation (Story)


etiennekerbastard (approved):

BryanS29-akeneo:
There is no way to do the condition on the first element of the score bar instead of directly the value 'A'?

LaurentPetard (approved):

PR #16643, opened 2022-03-22
CPM-574: Add step to set not nullable uuid columns, and add constraints and indexes

pierallard:
This PR:
- migrate every column to set it "not nullable" once they are all filled
- migrate every foreign table to add a constraint for each former constraint, using _uuid instead of _id
- migrate every foreign table to add a Foreign Key for each foreign column
- migrate some foreign tables to add indexes on _uuid when there was a _id
- migrate every foreign table PRIMARY KEY when the previous one uses _id


src/Akeneo/Pim/Enrichment/Bundle/Command/MigrateToUuid/MigrateToUuidSetNotNullableUuidColumns.php
```diff
@@ -0,0 +1,112 @@
...
+
+        return $result !== 'NO';
+    }
+
+    public function addMissing(Context $context): bool
+    {
+        $logContext = $context->logContext;
+
+        $updatedItems = 0;
+        foreach (self::TABLES as $tableName => $columnNames) {
```
nmarniesse, line 76: The column resource_uuid for the pim_versioning_version table should not be not null.  
To remove also in the `getMissingCount` method

mmetayer (approved):

cemorin (approved):

PR #16657, opened 2022-03-23
TIP-1556: allow to fake the feature flags service itself

ahocquard:
# Problem

The current solution for the feature flags allows to use the DI to determine how to fetch the status of a feature flag. Therefore, when declaring a new feature flag, a new service should be created. As you probably want to test the behavior when a feature flag is deactivated, you have to create several implementation for `prod`, `test`, `test_fake` and `behat` environment.

Also:
- you create services that are always similar in the different environments
- It makes the DX about feature flags quite painful, as you redeclare always the same thing
- it's not possible to mock a feature flag just by knowing its name.

You have to do something like:
```
        static::getContainer()->get('akeneo_asset.feature_flag')->enabled('asset');
```

I propose something like:
```
        static::getContainer()->get(feature_flags')->enabled('asset');
```

It allows to do use variable, to make activation of a feature flag generic. Useful for behat or acceptance test, where you don't want to inject all feature flags services to activate them or not. Now it's possible to do:

```
        foreach ($featureFlags as $featureFlag) {
            $this->featureFlags->enable($featureFlag);
        }
```

More generally, I think we should not expose the feature flag service name `akeneo_asset.feature_flag` but just being coupled to the registry `feature_flags` and the feature flag name `asset_manager` (as for a command bus actually). The consequence is that the service name could change over time for a same feature flag and is reusable across feature flags.

For example, we could have reusable services `feature_flag.activated_only_in_serenity` or another one `feature_flag.activated_only_in_trial_edition`, and switch their use depending of our business needs.

## General rules for testing

- Testing will be done depending of **feature flag**, not **edition**. The edition is just packaging. Also, doing it by edition would be similar to do it by feature, but with one additional level of indirection.
It would hide the coupling between the features, and be less flexible.

- Feature flags are **disabled** by default. A test suite/test class/single test should explicitly activate feature flag if it needs it to work. This way, it's much more explicit if a coupling exist.

- All the codebase should use the `feature_flags` registry instead of directly injecting the feature flag. It eases drastically the DX in term of testing.

## Integration tests

All tests will use the `feature_flags` service now. It's an in memory service, as there is no need of persistence for these tests **for now** (it could be the case for tests with sub processes).

It's easily done in setUp like this:
```
        static::getContainer()->get(feature_flags')->enabled('asset');
```

## Acceptance tests

Same as phpunit: in memory implementation of `feature_flags`. I used behat tag to activate a feature "@asset-manager-feature-enabled".

Note: we could also use Gherkin itself, but behat tag allows to create dedicated suite for a feature, and it's easy to search in the codebase (quite useful for the DX when I developed) and does not pollute the test itself.

## Behat tests

Use tag as for acceptance tests.

It's not possible to use an in memory `feature_flags` services, as a same test triggers several distinct PHP processes (HTTP requests) that should share the same state of feature flags.
That's why persist the enabled feature flags into a file.

 
 :warning: for very complex reason (container boot time DI decision to know which service to use, override of onboarder, etc), onboarder is failing. To fix it, the onboarder needs this PR merged... it's the snake eating its own tail #ouroboros.
 
I could not find a satisfying solution to make the onboarder work.  I will deactivate its requirement in the CI, then fix the onboarder, then reactivate it. Validated with @docteurklein and @phaseinducer . 

 

- [X] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc


src/Akeneo/Platform/Bundle/FeatureFlagBundle/Internal/InMemoryFeatureFlags.php
```diff
@@ -0,0 +1,55 @@
...
+ * It can be changed after Symfony container boot. It is mainly used for testing purpose. Its configuration shares the same lifecycle as the Symofny container. Therefore, it cannot be used to configure feature flags in sub processes, or for end to end testing with multiple HTTP requests.
+ *
+ * @copyright 2022 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+class InMemoryFeatureFlags implements FeatureFlags
+{
+    private array $flags = [];
+
+    public function __construct(private Registry $registry)
```
docteurklein, line 21: not necessary

tests/back/Platform/Integration/FeatureFlag/FilePersistedFeatureFlagsIntegration.php
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    private function initializeFeatureFlag(): FilePersistedFeatureFlags
+    {
+        $registry = new Registry();
+        $registry->add('foo', new Enabled());
+        $registry->add('bar', new Disabled());
+
+        $featureFlags = new FilePersistedFeatureFlags($registry, sys_get_temp_dir() . '/');
+        $featureFlags->deleteFile();
```
StevenVAIDIE, line 42: We should not test without delete the file ?

src/Akeneo/Platform/Bundle/FeatureFlagBundle/Internal/ImmutableFeatureFlags.php
```diff
@@ -0,0 +1,39 @@
...
+ */
+class ImmutableFeatureFlags implements FeatureFlags
+{
+    /** @var Registry */
+    private $registry;
+
+    public function __construct(Registry $registry)
+    {
+        $this->registry = $registry;
+    }
```
StevenVAIDIE, line 21: TMTC
```suggestion
    public function __construct(private Registry $registry)
    {
    }
```

src/Akeneo/Platform/Bundle/FeatureFlagBundle/Internal/ImmutableFeatureFlags.php
```diff
@@ -0,0 +1,39 @@
...
+
+    /**
+     * @param string $feature
+     *
+     * @return bool
+     */
```
StevenVAIDIE, line 27:
```suggestion
```

StevenVAIDIE (approved):

src/Akeneo/Platform/Bundle/FeatureFlagBundle/Internal/InMemoryFeatureFlags.php
```diff
@@ -0,0 +1,63 @@
...
+    public function enable(string $feature): void
+    {
+        $this->throwExceptionIfFlagDoesNotExist($feature);
+
+        $this->flags[$feature] = true;
+    }
+
+    public function disable(string $feature): void
+    {
+        $this->throwExceptionIfFlagDoesNotExist($feature);
```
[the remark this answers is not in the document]
ahocquard: :+1: I will remove it if it's useless in the future.

tests/back/Platform/Integration/FeatureFlag/FilePersistedFeatureFlagsIntegration.php
ahocquard: It resets the state between each file. So no, it's mandatory for test isolation.

src/Akeneo/Platform/Bundle/FeatureFlagBundle/Internal/FilePersistedFeatureFlags.php
```diff
@@ -0,0 +1,87 @@
...
+            unlink($this->filepath);
+        }
+    }
+
+    /**
+     * @param string $feature
+     */
+    private function throwExceptionIfFlagDoesNotExist(string $feature): void
+    {
+        $this->registry->get($feature);
```
phaseinducer, line 85: This feels weird. But I understand that you want to keep the exception throwing at the lowest level.

BitOne (approved):

tests/back/Platform/Integration/FeatureFlag/FilePersistedFeatureFlagsIntegration.php
ahocquard: It's a bad practice to perform IO at instantiation time.

tests/back/Acceptance/FeatureFlag/FeatureFlagContext.php
```diff
@@ -0,0 +1,40 @@
...
+{
+    public function __construct(
+        private InMemoryFeatureFlags $featureFlags
+    ) {
+    }
+
+    /**
+     * @BeforeScenario
+     */
+    public function enabledFeatureFlags(BeforeScenarioScope $scope)
```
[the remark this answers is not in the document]
ahocquard: I will do it later on, when I will discover if there are some limits

src/Akeneo/Platform/Bundle/FeatureFlagBundle/Internal/FilePersistedFeatureFlags.php
ahocquard: As throw exception it's part of the registry API, it's totally fine. But it's something unusual in PHP as exception is not guaranteed at compile time to cascade.

Also, it seems weird because it's `void` so you think it returns nothing. But in PHP, `void` is not really `void`, as error handling is through exception. In another language with monadic error handling instead of exception (rust for example), it would be less "weird" as it would return something.

tests/back/Acceptance/FeatureFlag/FeatureFlagContext.php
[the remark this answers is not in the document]
ahocquard: https://akeneo.atlassian.net/browse/TIP-1571

PR #16669, opened 2022-03-24
CXP-1124: Refresh the name & logo of a Connected App when different

tseho:

When cached information from the App Store changes (name, logo), we need to update it in the ConnectedApp for consistency.

This is triggered when both versions are reconciled in the front.

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/src/Akeneo/Connectivity/Connection/back/Domain/Apps/Model/ConnectedApp.php
+++ b/src/Akeneo/Connectivity/Connection/back/Domain/Apps/Model/ConnectedApp.php
@@ -132,14 +132,18 @@ final class ConnectedApp
         bool $certified = false,
         ?string $partner = null,
     ): self {
-        $self = clone $this;
-        $self->name = $name;
-        $self->logo = $logo;
-        $self->author = $author;
-        $self->categories = $categories;
-        $self->certified = $certified;
-        $self->partner = $partner;
-
-        return $self;
+        return new self(
+            $this->id,
+            $name,
+            $this->scopes,
+            $this->connectionCode,
+            $logo,
+            $author,
+            $this->userGroupName,
+            $categories,
+            $certified,
+            $partner,
+            $this->isTestApp,
+        );
     }
 }
--- a/src/Akeneo/Connectivity/Connection/back/Domain/Apps/Model/ConnectedApp.php
+++ b/src/Akeneo/Connectivity/Connection/back/Domain/Apps/Model/ConnectedApp.php
@@ -121,6 +121,9 @@ final class ConnectedApp
         ];
     }
 
+    /**
+     * @param array<string> $categories
+     */
     public function withUpdatedDescription(
         string $name,
         ?string $logo,
--- a/src/Akeneo/Connectivity/Connection/back/Domain/Apps/Model/ConnectedApp.php
+++ b/src/Akeneo/Connectivity/Connection/back/Domain/Apps/Model/ConnectedApp.php
@@ -120,4 +120,23 @@ final class ConnectedApp
             'is_test_app' => $this->isTestApp,
         ];
     }
+
+    public function withUpdatedDescription(
+        string $name,
+        ?string $logo,
+        ?string $author,
+        array $categories = [],
+        bool $certified = false,
+        ?string $partner = null,
+    ): self {
+        $self = clone $this;
+        $self->name = $name;
+        $self->logo = $logo;
+        $self->author = $author;
+        $self->categories = $categories;
+        $self->certified = $certified;
+        $self->partner = $partner;
+
+        return $self;
+    }
 }
--- a/src/Akeneo/Connectivity/Connection/back/tests/Unit/spec/Domain/Apps/Model/ConnectedAppSpec.php
+++ b/src/Akeneo/Connectivity/Connection/back/tests/Unit/spec/Domain/Apps/Model/ConnectedAppSpec.php
@@ -132,7 +132,7 @@ class ConnectedAppSpec extends ObjectBehavior
         ]);
     }
 
-    public function is_updates_description_properties(): void
+    public function it_updates_description_properties(): void
     {
         $updated = $this->withUpdatedDescription(
             'New Name',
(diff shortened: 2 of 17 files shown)


guillaumefer-akeneo (approved):

src/Akeneo/Connectivity/Connection/back/Domain/Apps/Model/ConnectedApp.php
```diff
@@ -120,4 +120,26 @@ public function normalize(): array
...
+     */
+    public function withUpdatedDescription(
+        string $name,
+        ?string $logo,
+        ?string $author,
+        array $categories = [],
+        bool $certified = false,
+        ?string $partner = null,
+    ): self {
+        $self = clone $this;
```
willy-ahva, line 135: I don't think it's a good way to do it. I'd rather prefer to use the constructor doing $newApp = new self();. If you don't use the constructor you could bypass some business hydration rules. Here is one for example https://github.com/akeneo/pim-community-dev/pull/16609/files#diff-e8d6eedb28436cc1c33ad993eee15a0a5c6fb8da0f7b5fcbd2c4090f3fe1c350R63
Moreover, the name of the method doesn't really reflect what it does. What do you think about instantiating a new ConnectedApp in the handler with the data from the previous one instead of adding a new method?

src/Akeneo/Connectivity/Connection/back/tests/Unit/spec/Domain/Apps/Model/ConnectedAppSpec.php
```diff
@@ -131,4 +131,30 @@ public function it_is_not_a_test_app_by_default(): void
             'is_test_app' => false,
         ]);
     }
+
+    public function is_updates_description_properties(): void
```
willy-ahva, line 135:
```suggestion
    public function it_updates_description_properties(): void
```
If you keep this method

src/Akeneo/Connectivity/Connection/front/src/connect/hooks/use-connected-apps.ts
```diff
@@ -63,6 +81,16 @@ export const useConnectedApps = (): ConnectedApp[] | null | false => {
                         };
                     });
                 });
+
+                // trigger a refresh when there is an inconsistency between the data in a connected app and the data
+                // in the app store
+                connectedApps.forEach(connectedApp => {
+                    const app = apps.apps.find(app => app.id === connectedApp.id);
+
+                    if (undefined !== app && !isAppSubsetIdentical(app, connectedApp)) {
```
willy-ahva, line 90: Reviewing your PR makes me realize that we shouldn't do it in front but in the backend with an event after fetching Apps from the Akeneo App store. Doing a comparison in front is non-sense. I understand that we do it for a performance purpose, but I think we have a problem with the architecture here.

src/Akeneo/Connectivity/Connection/front/tests/src/connect/hooks/use-connected-apps.test.ts
```diff
@@ -231,3 +232,87 @@ test('it fetches connected test apps', async () => {
     await waitForNextUpdate();
     expect(result.current).toEqual([expectedApp]);
 });
+
+test('it trigger a connnected app update if there is inconsistency between it and the app store data', async () => {
```
willy-ahva, line 236:
```suggestion
test('it triggers a connected app update if there is an inconsistency between it and the app store data', async () => {
```

willy-ahva (approved):

src/Akeneo/Connectivity/Connection/front/src/connect/hooks/use-connected-apps.ts
tseho: I agree with you, I hate doing it there.
IHMO, a better solution would be to add a cronjob instead of slowing down the queries in the back.

src/Akeneo/Connectivity/Connection/back/Domain/Apps/Model/ConnectedApp.php
tseho: > If you don't use the constructor you could bypass some business hydration rules.
Ok, here my thoughts on this.
Having an Anemic model with setters is always bad because you can't guarantee that your object is still valid at any moment, some code could set contradictory values in it.
Using methods like this, on a "rich" model guarantee that the mutation only happens inside the class, in methods that are able to validate if the change is valid.

> I don't think it's a good way to do it.
However, I will fix it and call the constructor inside.

> Moreover, the name of the method doesn't really reflect what it does. What do you think about instantiating a new ConnectedApp

I want to avoid to create a new instance of the same object externally. For the code POV, you are creating a new instance, not updating the existing one.
Example:
Let's say my class is like this:
```
class Foo {
public function __construct(
  public readonly string $id,
  public readonly string $name,
)
}

$foo = new Foo('1', 'Initial name');
$repository->save($foo);

// I update it
$foo = new Foo($foo->id, 'New name');
$repository->save($foo);

// Now, my class changes
class Foo {
public function __construct(
  public readonly string $id,
  public readonly string $name,
  public readonly bool $published = true,
)
}

$foo = new Foo('1', 'Initial name', false);
$repository->save($foo);

// The code to update the name in my handler is still working but I will loose the $published value
$foo = new Foo($foo->id, 'New name');
$repository->save($foo);
```

src/Akeneo/Connectivity/Connection/front/src/connect/hooks/use-connected-apps.ts
tseho: Do you think we should abandon the PR as it is and restart with a cronjob & the probably related ADR ?

RiffVentura: Moreover, that's  a lot of overhead since it's done one by one

RiffVentura (approved):

src/Akeneo/Connectivity/Connection/front/src/connect/hooks/use-connected-apps.ts
tseho: I can't argue with Willy's, I did a dirty version of it but I'm curious, where is the overhead ?

masacc (approved):

PR #16688, opened 2022-03-28
PIM-10372: Fix letter case issue when importing channels

LaurentPetard:
Locale codes of channel labels translations could be persisted with a different letter case that the (good) one persisted in the locales table.

This fix prevent that problem by replacing the locale code with the locale entity one during the data conversion from flat to standard.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@
 - PIM-10346: Fix spellcheck badge not displayed on attribute options
 - PIM-10336: Fix product Export edition in error if no locale selected
 - PIM-10345: Fix issue when importing product model with an attribute constituted of only digits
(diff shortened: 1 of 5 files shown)


FSAkeneo (approved):

ahocquard (changes requested):
It's still possible to do it with the API, so the fix is not valid.

- in updater layer
- or why don't we fix the translation issue itself? As the table is case insensitive, the issue should be on PHP layer, doing a case sensitive comparison somewhere.

docteurklein:
or we could use mysql collations? https://dev.mysql.com/doc/refman/8.0/en/case-sensitivity.html

FSAkeneo:
I'm closing in favor of https://github.com/akeneo/pim-community-dev/pull/16707

PR #16694, opened 2022-03-28
OB-1033: Update doctrine/persistence dependency

ValentinMerlet:
Composer update with a small update of doctrine/persistence package in order to be able to type class properties.

Note: Do not pull-up.


ValentinMerlet:
I'm closing it as it's outdated

PR #16696, opened 2022-03-28
RAC-1208: Hide "reference data" attributes from SaaS users

ghoussard:

In this PR, I introduced a new feature flag configuration `OnlyFlexibilityOnPremiseFeatureFlag`. Its purpose is to have the possibility to enabled some feature only on Flexibility and OnPremise environment. It will be very useful to hide any feature from SaaS.
Also I added a reference_data feature and I added a check into the SelectAttributeType.tsx to remove reference data attribute type from the create options. That will hide reference data creation from the SaaS UI.

**/!\ This PR just hides reference data, don't disable it.**
I created a card to completly disable it. At this time, it is a quickwin to just hide it and it is ok for our PM.

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc


ghoussard:
Refresh

StevenVAIDIE (approved):

ValentinMumble (approved):

PR #16718, opened 2022-03-31
RAB-584: Handle array object in fake generateRoute

StevenVAIDIE:

Microfrontend router does not handle array query param. In this PR, I fixed it and moved it into a dedicated model in order to test it

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc


front-packages/shared/src/microfrontend/model/routing.ts
```diff
@@ -0,0 +1,15 @@
+const createQueryParam = (parameters?: object) => {
+  if (!parameters) return '';
+
+  const queryParameters = Object.entries(parameters).map(([key, val]) => {
+    if (Array.isArray(val)) {
+      return val.map(value => `${key}[]=${decodeURIComponent(value)}`).join('&');
```
ghoussard, line 6:
```suggestion
      return val.map(value => `${key}[]=${encodeURIComponent(value)}`).join('&');
```

front-packages/shared/src/microfrontend/model/routing.ts
```diff
@@ -0,0 +1,15 @@
...
+    }
+
+    return `${key}=${decodeURIComponent(val)}`;
```
ghoussard, line 9:
```suggestion
    return `${key}=${encodeURIComponent(val)}`;
```

ghoussard (approved):

ValentinMumble (approved):

PR #16775, opened 2022-04-08
CXP-1154: rewrite phpcd rules

masacc:

New PHP CD rules for Connectivity :

- Domain of {x} should use only classes from the same Domain
- Application of {x} should use only classes from the same Domain or same Application
- Infrastructure can use anything

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/src/Akeneo/Connectivity/Connection/back/tests/.php_cd.php
+++ b/src/Akeneo/Connectivity/Connection/back/tests/.php_cd.php
@@ -11,6 +11,31 @@ $finder->notPath('tests');
 $builder = new RuleBuilder();
 
 $rules = [
+    // Domain should only use classes from Domain
+    $builder->only(
+        [
+            'Akeneo\Connectivity\Connection\Domain',
+
+            // Exceptions
+            'Symfony\Component\Validator\Context\ExecutionContextInterface',
+            'Symfony\Component\Validator\ConstraintViolationInterface',
+            'Symfony\Component\Validator\ConstraintViolationListInterface',
+
+            'Akeneo\Platform\Component\EventQueue\EventInterface',
+            'Akeneo\Platform\Component\EventQueue\Author',
+            'Akeneo\Platform\Component\Webhook\EventBuildingExceptionInterface',
+
+            'Akeneo\Pim\Enrichment\Component\Product\Message\ProductModelCreated',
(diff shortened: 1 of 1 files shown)


tseho:
Why did you skip the rules at the root level ?

```
    Domain should use only classes from Domain (there will be exceptions)
    Application should use only classes from Domain & Application
    Infrastructure can use anything
```

I think it would be great to have rules like

```
$builder->only(
        [
            'Akeneo\Connectivity\Connection\Domain',
        ]
    )->in('Akeneo\Connectivity\Connection\Domain'),
```
Because you may not know, but when we added new subdomains, like Marketplace & Apps, we forgot to add the specific rules in phpcd for this new domains and the rules were violated, infrastructure classes were used in the domain.

masacc:
> Because you may not know, but when we added new subdomains, like Marketplace & Apps, we forgot to add the specific rules in phpcd for this new domains and the rules were violated, infrastructure classes were used in the domain.

Ok, I was wondering if this is useful because it will duplicate a lot of lines. I'll add them.

tseho:
Do you think that all the specific rules are useful ? honest question, since you saw them all

tseho:
Do we want to have a context map and apply this context map to our rules ?
by example, "it's normal if Application/Apps is using Domain/Connections due to the inheritance of the features but Application/Connections should not use Domain/Apps"

masacc:
There are indeed some related "context" : connection / apps / marketplace is an good example. I honestly don't know what is the best solution. If I may have to guess, I would say that everything in Connectivity can use something else in Connectivity without any potential issue. On the opposite, we should explicitly show coupling with other contexts (like userManagement) in PHP CD.

masacc:
I just pushed the changes with the global rules for Domain and Application to have a complete view of coupling.

tseho:
I didn't know it would make so many rules.
What do you think of the result ?

masacc:
Those in Domain really surprised me but I did not check for what reason they exist. In Application, I think there is a couple of them that can be avoidable (those related to Infrastructure), but others make sense IMHO (related to other contexts or Symfony/vendor tools like Logger).

guillaumefer-akeneo (approved):

PR #16806, opened 2022-04-12
OCT-16: fix Marketplace page error 400

guillaumefer-akeneo:
image
image

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/src/Akeneo/Connectivity/Connection/back/Infrastructure/Marketplace/Controller/Internal/GetAllAppsAction.php
+++ b/src/Akeneo/Connectivity/Connection/back/Infrastructure/Marketplace/Controller/Internal/GetAllAppsAction.php
@@ -47,7 +47,11 @@ final class GetAllAppsAction
         } catch (\Exception $e) {
             $this->logger->error(\sprintf('unable to retrieve the list of apps, got error "%s"', $e->getMessage()));
 
-            return new JsonResponse(GetAllAppsResult::create(0, [])->normalize());
+            if (Response::HTTP_BAD_REQUEST === $e->getCode()) {
+                return new JsonResponse(GetAllAppsResult::create(0, [])->normalize());
+            }
+
+            return new Response(null, Response::HTTP_NO_CONTENT);
         }
 
         $username = $this->userContext->getUser()->getUsername();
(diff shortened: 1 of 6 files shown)


tseho:
What's the behavior if the marketplace is down with your new version ?

guillaumefer-akeneo:
> What's the behavior if the marketplace is down with your new version ?

image

The UnreachableMarketplace component is no longer rendered for the moment.

tseho:
That's not good :D

When we catch the exception
```
} catch (\Exception $e) {
    $this->logger->error(\sprintf('unable to retrieve the list of apps, got error "%s"', $e->getMessage()));

    return new Response(null, Response::HTTP_NO_CONTENT);
}
```

Is there a way to catch a different exception for 400 errors or check the given exception if it's a 400 error and only then, return an empty list but keep the default behavior of sending `new Response(null, Response::HTTP_NO_CONTENT);` for all other errors ?

guillaumefer-akeneo:
> That's not good :D
> 
> When we catch the exception
> 
> ```
> } catch (\Exception $e) {
>     $this->logger->error(\sprintf('unable to retrieve the list of apps, got error "%s"', $e->getMessage()));
> 
>     return new Response(null, Response::HTTP_NO_CONTENT);
> }
> ```
> 
> Is there a way to catch a different exception for 400 errors or check the given exception if it's a 400 error and only then, return an empty list but keep the default behavior of sending `new Response(null, Response::HTTP_NO_CONTENT);` for all other errors ?

It's fixed :D

guillaumefer-akeneo:
> That's not good :D
> 
> When we catch the exception
> 
> ```
> } catch (\Exception $e) {
>     $this->logger->error(\sprintf('unable to retrieve the list of apps, got error "%s"', $e->getMessage()));
> 
>     return new Response(null, Response::HTTP_NO_CONTENT);
> }
> ```
> 
> Is there a way to catch a different exception for 400 errors or check the given exception if it's a 400 error and only then, return an empty list but keep the default behavior of sending `new Response(null, Response::HTTP_NO_CONTENT);` for all other errors ?

it's fixed

src/Akeneo/Connectivity/Connection/back/Infrastructure/Marketplace/Controller/Internal/GetAllAppsAction.php
```diff
@@ -47,6 +47,10 @@ public function __invoke(Request $request): Response
         } catch (\Exception $e) {
             $this->logger->error(\sprintf('unable to retrieve the list of apps, got error "%s"', $e->getMessage()));
 
+            if (Response::HTTP_BAD_REQUEST === $e->getCode()) {
+                return new JsonResponse(GetAllAppsResult::create(0, [])->normalize());
+            }
+
```
masacc, line 53: Are there some tests to create/update for this use case?

willy-ahva (approved):

RiffVentura (approved):

masacc (approved):

masacc (approved):

PR #16844, opened 2022-04-15
PLG-825: Adaptations to persist double scores in ES

LaurentPetard:
For ES indexation, we need both scores, but for displaying in the PEF and the product-grid we need the scores enabled according to the feature flag. Although they use the same MySQL queries.

As we need either the enabled score or both scores depending of the usage context, I changed the type of return of the MySQL queries that fetch scores. It will now return a read-model with both scores, and a dedicated service will be able to determine the enabled ones if needed.

### DIFF
--- a/src/Akeneo/Pim/Automation/DataQualityInsights/back/Application/GetEnabledScoresStrategy.php
+++ b/src/Akeneo/Pim/Automation/DataQualityInsights/back/Application/GetEnabledScoresStrategy.php
@@ -1,0 +1,29 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Akeneo\Pim\Automation\DataQualityInsights\Application;
+
+use Akeneo\Pim\Automation\DataQualityInsights\Domain\Model\ChannelLocaleRateCollection;
+use Akeneo\Pim\Automation\DataQualityInsights\Domain\Model\Read;
+use Akeneo\Platform\Bundle\FeatureFlagBundle\FeatureFlag;
+
+/**
+ * @copyright 2022 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+class GetEnabledScoresStrategy
+{
+    public function __construct(
+        private FeatureFlag $allCriteriaFeature,
+    ) {
+    }
+
+    /**
+     * Determines the scores to use, according to the current enabled features
+     */
+    public function __invoke(Read\Scores $scores): ChannelLocaleRateCollection
+    {
+        return $this->allCriteriaFeature->isEnabled() ? $scores->allCriteria() : $scores->partialCriteria();
+    }
+}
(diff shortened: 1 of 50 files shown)


src/Akeneo/Pim/Automation/DataQualityInsights/back/Application/GetEnabledScoresStrategy.php
```diff
@@ -0,0 +1,29 @@
...
+
+use Akeneo\Pim\Automation\DataQualityInsights\Domain\Model\ChannelLocaleRateCollection;
+use Akeneo\Pim\Automation\DataQualityInsights\Domain\Model\Read;
+use Akeneo\Platform\Bundle\FeatureFlagBundle\FeatureFlag;
+
+/**
+ * @copyright 2022 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+class GetEnabledScoresStrategy
```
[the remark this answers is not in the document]
antoinetrouve: simply : 
```suggestion
class GetScores
```
?

antoinetrouve (approved):

src/Akeneo/Pim/Automation/DataQualityInsights/back/Infrastructure/Elasticsearch/BulkUpdateProductQualityScoresIndex.php
```diff
@@ -60,7 +60,8 @@ public function __invoke(ProductIdCollection $productIdCollection): void
                 'script' => [
                     'inline' => "ctx._source.data_quality_insights = params;",
                     'params' => [
-                        'scores' => $qualityScores->toArrayIntRank(),
+                        'scores' => $qualityScores->allCriteria()->toArrayIntRank(),
+                        'scores_partial_criteria' => $qualityScores->partialCriteria()->toArrayIntRank(),
```
FSAkeneo, line 64: Why do we need to have scores_partial_criteria as well, isn't it a sort of 'subset' of 'scores' ?
If it is for the front, then maybe the front could switch behavior according to dqi_full feature flag ?

FSAkeneo (approved):

src/Akeneo/Pim/Automation/DataQualityInsights/back/Infrastructure/Elasticsearch/BulkUpdateProductQualityScoresIndex.php
LaurentPetard: Searching products from ES is used for the product-grid, but also for the exports and the public API

FSAkeneo: Thanks @LaurentPetard It makes sense for filtering, but I imagine we can modify the filtering code backendside so that it's  different when full_dqi is on. That would eliminate the redundancy (if there is redundancy, I'm unsure). Maybe I miss something else ;)

src/Akeneo/Pim/Automation/DataQualityInsights/back/Application/GetEnabledScoresStrategy.php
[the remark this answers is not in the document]
LaurentPetard: GetScoresByCriteriaStrategy

PR #16851, opened 2022-04-19
CXP-1174: Faster CXP CI

tseho:
**Before**
Screenshot_2022-04-28_15-03-53

_From start to finish: 16-17min_

**After**
Screenshot_2022-04-28_14-26-16

_From start to finish: 9-10min_

---

**The trick**

By installing the dependencies before building the PIM, we can already execute a lot of tests (unit, integration) while compiling the JS for example.
The `build_ce` is still the same and still tries to install the dependencies, but since they are already there or were cached, it has no impact.

**What about persistence of the files?**

Persisting the same workspace in parallelized jobs is possible.
https://circleci.com/blog/deep-diving-into-circleci-workspaces/
Thanks @ahocquard for the tips

By persisting only subdirectories when installing the dependencies:
```
-   persist_to_workspace:
        root: ~/
        paths:
            - project/vendor
```
```
-   persist_to_workspace:
        root: ~/
        paths:
            - project/node_modules
            - project/front-packages/akeneo-design-system/lib
            - project/front-packages/shared/lib
```

Circle CI is able to merge the layers at no cost
Screenshot_2022-04-28_14-27-54

**Faster CI, early feedback**

The CI, while still running the same amount of tests, is obviously faster without having to rely on heavy parallelization. 
And this way, we have feedback earlier, for example, PHP unit tests are termined 2min after the CI starts. (before, it was around 8-10min after the CI started)

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/.circleci/jobs/build_and_deploy.yml
+++ b/.circleci/jobs/build_and_deploy.yml
@@ -107,10 +107,105 @@ jobs:
             - run:
                   name: Remove MySQL port translation (see BH-664)
                   command: yq delete --inplace docker-compose.yml services.mysql.ports
+            - run:
+                  name: Copy docker-compose.override.yml.dist
+                  command: cp .circleci/docker-compose.override.yml.dist docker-compose.override.yml
+            -   run:
+                    name: Setup tests results folder and log folder
+                    command: mkdir -p var/tests/phpspec var/tests/csfixer var/logs var/tests/screenshots ~/.composer
+            -   run:
+                    name: Change owner on project dir (docker needs uid 1000, circleci can be another uid)
+                    command: sudo chown -R 1000:1000 ../project
+            -   run:
+                    name: Change permissions on project dir
+                    command: sudo chmod -R 777 ../project
             - persist_to_workspace:
                   root: ~/
                   paths:
                       - project
+
+    install_ce_back_dependencies:
+        machine:
+            image: *executor-machine
+        steps:
+            -   attach_workspace:
+                    at: ~/
+            -   run:
+                    name: Creating cache key for back dependencies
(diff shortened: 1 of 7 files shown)


willy-ahva (approved):

.circleci/jobs/build_and_deploy.yml
```diff
@@ -107,11 +107,103 @@ jobs:
...
+                    command: sudo chown -R 1000:1000 ../project
+            -   run:
+                    name: Change permissions on project dir
+                    command: sudo chmod -R 777 ../project
             - persist_to_workspace:
                   root: ~/
                   paths:
                       - project
 
+    install_back_dependencies:
```
ahocquard, line 124:
```suggestion
    install_ce_back_dependencies:
```

Other step build_dev supports both CE and EE. I think we should precise it's just CE here.

.circleci/jobs/build_and_deploy.yml
```diff
@@ -107,11 +107,103 @@ jobs:
...
+                    command: docker-compose pull php
+            -   run:
+                    name: Change owner on project dir (docker needs uid 1000, circleci can be another uid)
+                    command: sudo chown -R 1000:1000 ../project
+            -   run:
+                    name: Change permissions on project dir
+                    command: sudo chmod -R 777 ../project
+            -   run:
+                    name: Install back dependencies
+                    command: make vendor
```
ahocquard, line 148: - run:
```suggestion
            - run:
                  name: Copy docker-compose.override.yml.dist
command: cp .circleci/docker-compose.override.yml.dist docker-compose.override.yml
```
            for inmemory database

PR #16865, opened 2022-04-21
Cpm 609 update pim_comment_comment table creation and fix code to use uuid

amigaire:

- [x] Tests

### DIFF
--- a/src/Akeneo/Pim/Enrichment/Bundle/Doctrine/ORM/Repository/CommentRepository.php
+++ b/src/Akeneo/Pim/Enrichment/Bundle/Doctrine/ORM/Repository/CommentRepository.php
@@ -2,8 +2,10 @@
 
 namespace Akeneo\Pim\Enrichment\Bundle\Doctrine\ORM\Repository;
 
+use Akeneo\Pim\Enrichment\Component\Comment\Model\CommentInterface;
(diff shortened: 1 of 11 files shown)


src/Akeneo/Pim/Enrichment/Bundle/Doctrine/ORM/Repository/CommentRepository.php
```diff
@@ -20,9 +23,22 @@ class CommentRepository extends EntityRepository implements CommentRepositoryInt
...
         return $this->findBy(
             ['resourceId' => $resourceId, 'resourceName' => $resourceName, 'parent' => null],
             ['createdAt'  => 'desc']
         );
     }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getCommentsByUuid(string $resourceName, UuidInterface $resourceUuid): array
```
nmarniesse, line 37: TODO: the former `getComments` is not used anymore because today we can only do comments on a Product. Do we keep the former method? Or do we delete it?  

The `resourceId` column in DB is not used neither.

nmarniesse, line 37: TODO: the former `getComments` is not used anymore because today we can only do comments on a Product. Do we keep the former method? Or do we delete it?  

The `resourceId` column in DB is not used neither.

PR #16873, opened 2022-04-25
Migration cron

mmetayer:
Some final fixes for the feature branch:

- Add (and keep) temporary indexes which are only useful for the migration (they will be removed once every client is migrated)
- Clean pim_catalog_completeness table
- Prioritize the migration command over DQI-related jobs

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/src/Akeneo/Pim/Enrichment/Bundle/Command/MigrateToUuid/MigrateToUuidCleanCompletenessTable.php
+++ b/src/Akeneo/Pim/Enrichment/Bundle/Command/MigrateToUuid/MigrateToUuidCleanCompletenessTable.php
@@ -1,0 +1,107 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Akeneo\Pim\Enrichment\Bundle\Command\MigrateToUuid;
+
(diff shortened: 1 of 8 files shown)


cemorin (approved):

src/Akeneo/Pim/Enrichment/Bundle/Command/MigrateToUuid/MigrateToUuidCleanCompletenessTable.php
```diff
@@ -0,0 +1,107 @@
...
+                SQL,
+                [
+                    'limit' => self::BATCH_SIZE,
+                ],
+                [
+                    'limit' => \PDO::PARAM_INT
+                ]
+            );
+
+            $cleanedRows += self::BATCH_SIZE;
```
nmarniesse, line 74: We can have the real deleted rows with `SELECT ROW_COUNT();`  
But this solution is fine to me

nmarniesse (approved):

PR #16880, opened 2022-04-25
RAB-561: Add default value when retrieving summary info on StepExecution

grena:

When retrieving a summary info on a StepExecution, we give the ability to set a default value.
This way we can enforce some integer instead of current string, for example.

- [x] Tests


ghoussard (approved):
You forgot to modify the call in Tailored Import side ?

StevenVAIDIE (approved):

PR #16884, opened 2022-04-26
PIM-10418 WIP

nmarniesse:

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc


src/Akeneo/Pim/Enrichment/Bundle/Versioning/CaseInsensitiveValueComparator.php
```diff
@@ -0,0 +1,65 @@
...
+    {
+        return [Product::class, ProductModel::class];
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function supportsField(string $field): bool
+    {
+        if (null === $this->indexedSupportedFields) {
```
ahocquard, line 40: too greedy to do that on each value

src/Akeneo/Pim/Structure/Bundle/Query/InternalApi/Attribute/SqlGetAttributeCodes.php
```diff
@@ -0,0 +1,38 @@
...
+final class SqlGetAttributeCodes implements GetAttributeCodes
+{
+    public function __construct(private Connection $connection)
+    {
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function forAttributeTypes(array $attributeTypes): array
```
ahocquard, line 24: no cache, no batching: it can fail and will slow down drastically the performance and the memory footprint.

ahocquard (commented):
It's very complicated/greedy to guess the attribute type in the versioning. It's an agnostic layer. Also, you are introducing notion of product values inside `src/Akeneo/Tool/Component`, which is a generic component.

ahocquard:
Why don't we fix the frontend?

PR #16885, opened 2022-04-26
fix(api-1586): update violation error message

tfehringer:

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/src/Akeneo/Pim/Enrichment/Bundle/Doctrine/ORM/Repository/ExternalApi/CategoryRepository.php
+++ b/src/Akeneo/Pim/Enrichment/Bundle/Doctrine/ORM/Repository/ExternalApi/CategoryRepository.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Akeneo\Pim\Enrichment\Bundle\Doctrine\ORM\Repository\ExternalApi;
 
+use Akeneo\Pim\Enrichment\Component\Validator\Constraints\Type;
 use Akeneo\Tool\Component\Api\Repository\ApiResourceRepositoryInterface;
 use Akeneo\Tool\Component\Classification\Repository\CategoryRepositoryInterface;
 use Doctrine\ORM\EntityManager;
@@ -11,7 +12,7 @@ use Doctrine\ORM\EntityRepository;
 use Doctrine\ORM\QueryBuilder;
 use Doctrine\ORM\UnexpectedResultException;
 use Symfony\Component\Validator\Constraints as Assert;
-use Symfony\Component\Validator\Validation;
+use Symfony\Component\Validator\Validator\ValidatorInterface;
 
 /**
  * @author    Willy Mesnage <willy.mesnage@akeneo.com>
@@ -20,22 +21,13 @@ use Symfony\Component\Validator\Validation;
  */
 class CategoryRepository extends EntityRepository implements ApiResourceRepositoryInterface
 {
-    /** @var CategoryRepositoryInterface */
-    protected $categoryRepository;
-
-    /**
-     * @param EntityManager $entityManager
-     * @param string $className
-     * @param CategoryRepositoryInterface $categoryRepository
-     */
     public function __construct(
-        EntityManager $entityManager,
-        $className,
-        CategoryRepositoryInterface $categoryRepository
+        protected EntityManager $entityManager,
+        protected string $className,
+        protected CategoryRepositoryInterface $categoryRepository,
+        private ValidatorInterface $validator
     ) {
         parent::__construct($entityManager, $entityManager->getClassMetadata($className));
-
-        $this->categoryRepository = $categoryRepository;
     }
 
     public function getIdentifierProperties()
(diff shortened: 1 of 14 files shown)


nestorAkeneo:

dFayet (approved):

tests/back/Pim/Enrichment/EndToEnd/Category/ExternalApi/GetCategoryEndToEnd.php
```diff
@@ -91,6 +91,24 @@ public function testGetACategoryWithPosition()
...
+
+        $client->request('GET', 'api/rest/v1/categories?search={"code":[{"operator":"IN","value":1234}]}');
+
+        $response = $client->getResponse();
+
+        $this->assertSame(Response::HTTP_BAD_REQUEST, $response->getStatusCode());
+        $this->assertSame(
+            [
+                'code' => 400,
+                'message' => 'In order to search on category codes you must send an array of category codes as value, {{ integer }} given. This value should be of type iterable.'
```
benjaminleneve, line 106: Normally, the type should not be surrounded by the double {{ integer }}

tfehringer: Are you sure ? it was like this prior changes i just replicate the existing message.

benjaminleneve: I tested on the previous version which returned array instead of string to check:

{
    "code": 400,
    "message": "In order to search on category codes you must send an array of category codes as value, array given. This value should be of type iterable."
}

However, I don't see why double {{ }} are not interpreted.

tfehringer: i'll remove the brackets then !

tfehringer: WDYT @pchasle ?

src/Akeneo/Pim/Enrichment/Bundle/Doctrine/ORM/Repository/ExternalApi/CategoryRepository.php
```diff
@@ -4,14 +4,15 @@
 
 namespace Akeneo\Pim\Enrichment\Bundle\Doctrine\ORM\Repository\ExternalApi;
 
+use Akeneo\Pim\Structure\Component\Validator\Constraints\Type;
```
pchasle, line 7: You should probably not use the validator defined in another bounded context. Adding an entry in the coupling detector file `.php_cd.php` is a generally a red flag.

You could duplicate the validator in the Enrichment/Category bounded context. Maybe to validate with the owner squad (Weasels?)

tests/back/Pim/Enrichment/EndToEnd/Category/ExternalApi/GetCategoryEndToEnd.php
pchasle: Yes, brackets should not be visible if the string is correctly interpolated :thinking:

src/Akeneo/Pim/Enrichment/Bundle/Doctrine/ORM/Repository/ExternalApi/CategoryRepository.php
tfehringer: Duplicate code that is not bounded to a context it is not great. It is purely technical like a Symfony's validator (cross context). Maybe we could add a `Core` context above other contexts ?

tfehringer: cc @olivierpontier

ahocquard: I agree with Paul. Category context should be autonomous and not linked to such class. Having a "Core" is generally a red flag:
- either it's a platform for a defined purpose (API framework for pagination/HAL format, etc is one for example)
- either it's owned by the context

In both cases, a clear ownership should be defined. Here, in the current state, I don't see the point to create a platform for that kind of stuff.

Last but not least, category is a context that will grow in a near-future, with its own stack and own choices. We could imagine the validation of the API (body + request parameters) being done by JSON Schema instead for example. 

So, I would copy this simple class to avoid to introduce more coupling. A big effort should be done by the team that will handle category for the decoupling (weasel), let's not add more work for them.

tests/back/Pim/Enrichment/EndToEnd/Category/ExternalApi/GetCategoryEndToEnd.php
benjaminleneve: I still have this problem of visible brackets in local

dFayet (approved):

olivierpontier (approved):

benjaminleneve (approved):

PR #16888, opened 2022-04-26
OBS-78: Add support for rows with warnings in Table component

ValentinMerlet:
Add support for WarningRows in Table component for Onboarder needs.

**Screenshot:**
warnable_rows


mpoissonnier (approved):

front-packages/akeneo-design-system/src/components/Table/Table.stories.mdx
```diff
@@ -475,3 +475,56 @@ If you want your `Table.Header` to stay on top when scrolling in the Table, just
     }}
   </Story>
 </Canvas>
+
+## Warnable rows
```
juliensnz, line 479: Are we sure about the wording? WarningRow seems more english no?

front-packages/akeneo-design-system/src/components/Table/Table.stories.mdx
```diff
@@ -475,3 +475,56 @@ If you want your `Table.Header` to stay on top when scrolling in the Table, just
...
+
+<Canvas>
+    <Story name="Warnable rows">
+        {args => {
+            return (
+                <Scrollable height={250}>
+                    <Table hasWarnedRows={true}>
```
juliensnz, line 486: Don't you think this could be computed?

front-packages/akeneo-design-system/src/components/Table/Table.stories.mdx
```diff
@@ -475,3 +475,56 @@ If you want your `Table.Header` to stay on top when scrolling in the Table, just
...
+                        <Table.Header sticky={0}>
+                            <Table.HeaderCell>Name</Table.HeaderCell>
+                            <Table.HeaderCell>Family</Table.HeaderCell>
+                            <Table.HeaderCell>Order</Table.HeaderCell>
+                            <Table.HeaderCell>Genus</Table.HeaderCell>
+                            <Table.HeaderCell>Conservation status</Table.HeaderCell>
+                            <Table.HeaderCell>Actions</Table.HeaderCell>
+                        </Table.Header>
+                        <Table.Body>
+                            <Table.Row hasWarning={true}>
```
juliensnz, line 496: Maybe we could use "level" as elsewhere in the DSM?

juliensnz (commented):
I think that we should discuss the naming of those props as it feels like it's not following our conventions

front-packages/akeneo-design-system/src/components/Table/Table.stories.mdx
ValentinMerlet: I thought too that the name was bad, ok for `WarningRow`, I'm doing the change.

olivierpontier (approved):

front-packages/akeneo-design-system/src/components/Table/Table.stories.mdx
juliensnz: even if there is no need, I find it more consistent with the rest of the DSM. Aside from that, if a warning row exists, I think that an error row could exist too :)

rouliane: for now we have only the warning state, there is no plan for a "error" state or something on the designer side as far as I know. Wouldn't it be confusing if we support a level prop with only "warning" as value ?
But now I think about it, we can do it like this for consistency purposes, I don't really mind.

ValentinMerlet: I guess we should find another naming for our component then, any suggestions?

front-packages/akeneo-design-system/src/components/Table/Table.stories.mdx
rouliane: > It could be

I checked in the code and it could be more complicated that I thought. Since all rows are in TableBody and we need to set the "has warning property" to the root Table component, I don't see an easy way to do it :/

PR #16890, opened 2022-04-26
TIP-1563: add feature flag on product read only attribute

ahocquard:
Read only attribute prevent any edit of an attribute in the UI.
It's still editable in import, API, etc (no security on backend side).

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/src/Akeneo/Pim/Structure/Component/Normalizer/Standard/AttributeNormalizer.php
+++ b/src/Akeneo/Pim/Structure/Component/Normalizer/Standard/AttributeNormalizer.php
@@ -3,6 +3,7 @@
 namespace Akeneo\Pim\Structure\Component\Normalizer\Standard;
 
 use Akeneo\Pim\Structure\Component\Model\AttributeInterface;
+use Akeneo\Platform\Bundle\FeatureFlagBundle\FeatureFlags;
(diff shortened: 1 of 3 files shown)


src/Akeneo/Pim/Structure/Component/Normalizer/Standard/AttributeNormalizer.php
```diff
@@ -103,4 +88,13 @@ public function hasCacheableSupportsMethod(): bool
     {
         return true;
     }
+
+    private function isReadOnlyFeatureAvailableAndEnabled(): bool
+    {
+        try {
+            return $this->featureFlags->isEnabled('read_only_product_attribute');
+        } catch (\InvalidArgumentException $notAvailableFeatureFlagException) {
```
ahocquard, line 96: Either :
- I declare a feature flag always at false (confusing)
- I decorate the service in EE, but it's already decorated... in the table attribute bundle (and I would like to triple Symfony decoration)
- or catch the exception if the feature flag is missing in CE


I chose the last one:
- for simplicity
- because read only property exists in CE (just not used)
- this feature is only used in UI and pretty simple, has never been changed, so no need to complexify things

jmleroux (approved):

StevenVAIDIE (approved):

PR #16891, opened 2022-04-26
CXP-1166: Adapt the level of permissions to the connected app scopes

willy-ahva:

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/src/Akeneo/Connectivity/Connection/back/Infrastructure/Symfony/Resources/translations/jsmessages.en_US.yml
+++ b/src/Akeneo/Connectivity/Connection/back/Infrastructure/Symfony/Resources/translations/jsmessages.en_US.yml
@@ -155,6 +155,7 @@ akeneo_connectivity.connection:
                     authorizations:
                         title: authorizations
                         information: 'To know more about app authorizations, check out our {{ link }}.'
+                        no_access_to_product_information: 'This connected app doesn’t access your product information.'
                         information_link_anchor: Help Center article
                         no_scope: No specific authorizations have been requested.
                     authentication:
--- a/src/Akeneo/Connectivity/Connection/front/src/connect/components/ConnectedApp/ConnectedAppAuthorizations.tsx
+++ b/src/Akeneo/Connectivity/Connection/front/src/connect/components/ConnectedApp/ConnectedAppAuthorizations.tsx
@@ -8,6 +8,7 @@ import ScopeMessage from '../../../model/Apps/scope-message';
 import {ConnectedAppScopeListIsLoading} from './ConnectedAppScopeListIsLoading';
 import {ScopeList} from '../ScopeList';
 import styled from 'styled-components';
+import isGrantedOnProduct from '../../is-granted-on-product';
 
 const ScopeListContainer = styled.div`
     margin: 10px 20px;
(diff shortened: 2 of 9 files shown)


src/Akeneo/Connectivity/Connection/front/src/connect/components/ConnectedApp/connectedAppScopesExplorer.ts
```diff
@@ -0,0 +1,15 @@
+import {ConnectedApp} from '../../../model/Apps/connected-app';
```
willy-ahva, line 1: Please take attention to the name of the file. WDYT ?

RiffVentura: What is the reasoning behind its name?

willy-ahva: Functions in the file was about providing some info about the scope of a connected App. If I was in back I could have done a rich model but, I'm not. So I decided to do a "tooling" file.  found Connect App Scope Explorer pretty clear. Is it a good idea ? Should I split it in two files? Should I use the kebab case? (I love this name, I know you too @RiffVentura)

src/Akeneo/Connectivity/Connection/front/src/connect/components/ConnectedApp/connectedAppScopesExplorer.ts
```diff
@@ -0,0 +1,15 @@
...
+
+const containsProductScope: (connectedApp: ConnectedApp) => boolean = (connectedApp: ConnectedApp): boolean => {
+    const productsRegexp = new RegExp('.*products.*');
+
+    return connectedApp.scopes.find(scope => productsRegexp.test(scope)) !== undefined;
```
RiffVentura, line 6: not a fan of regex TBH I would rather have an explicit list
```suggestion
    const productScopes = ['read_products', 'write_products', 'delete_products');

    return connectedApp.scopes.find(scope => productScopes.inlcudes(scope)) !== undefined;
```
WDYT ?

src/Akeneo/Connectivity/Connection/front/src/connect/components/ConnectedApp/connectedAppScopesExplorer.ts
willy-ahva: Finally, I think we should split it. Files will have the name of the function. It will be clearer. WDYT ?

RiffVentura (approved):

masacc (approved):

tseho (approved):
Great PR

Tip when you write unit tests on small functions like `is-granted-on-product`, you can write the equivalent of phpunit data providers with jest.

See https://github.com/akeneo/pim-community-dev/blob/e0595bd531fe5be8e6f215d360445411b01dca80/src/Akeneo/Connectivity/Connection/front/tests/src/connect/components/AppWizard/getStepConfirmationLabel.test.ts

PR #16896, opened 2022-04-27
PLG-841: POC Experiment CQRS pattern to decouple Category from the use of Doctrine

olivierpontier:

Replace the dependency on Doctrine by experimenting with the CQRS pattern. 

- [x] Define a command to create a category from the External API in order to define a facade around the existing services for creating a category
- [x] Define a query to read a category by its code from the External API to define a facade around the existing services for reading a category
- [ ] Replace the implementation of the create category command by using SQL implementation
- [x] Replace the implementation of the get category query by using SQL implementation
- [ ] EXternal API Get Category: handle the `with_position=1` argument and provide the position of the category 

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc


src/Akeneo/Category/back/API/Command/CreateCategoryCommand.php
```diff
@@ -0,0 +1,53 @@
...
+        $labels = $data['labels'] ?? [];
+        Assert::isArray($labels);
+
+        Assert::allString(\array_keys($labels));
+        Assert::allString(\array_values($labels));
+
+        return new self($data['code'], $parent, $labels);
+    }
+
+    public function getCode(): string
```
ahocquard, line 39: better to use psalm immutable class , to not rely on getter/setter

src/Akeneo/Category/back/API/CommandBus.php
```diff
@@ -0,0 +1,47 @@
...
+
+namespace Akeneo\Category\API;
+
+use Akeneo\Category\API\UnknownCommandException;
+use Symfony\Component\Messenger\Envelope;
+use Symfony\Component\Messenger\MessageBusInterface;
+use Webmozart\Assert\Assert;
+
+/**
+ * This bus is a simpler implementation, it will be replaced by a more generic solution.
```
ahocquard, line 13: :+1:

src/Akeneo/Category/back/Infrastructure/ExternalApi/CategoryController.php
```diff
@@ -35,63 +40,22 @@
...
-        SimpleFactoryInterface $factory,
-        ObjectUpdaterInterface $updater,
-        ValidatorInterface $validator,
-        SaverInterface $saver,
-        RouterInterface $router,
-        PaginatorInterface $paginator,
-        ParameterValidatorInterface $parameterValidator,
-        StreamResourceResponse $partialUpdateStreamResource,
-        array $apiConfiguration
+        protected ApiResourceRepositoryInterface $repository,
```
ahocquard, line 44: private is totally fine. It was an old rule to put "protected everywhere"

src/Akeneo/Category/back/Infrastructure/ExternalApi/CategoryNormalizer.php
```diff
@@ -0,0 +1,30 @@
...
+        private DateTimeNormalizer $dateTimeNormalizer,
+    ) {
+    }
+
+    public function normalize(Category $category): array
+    {
+        return [
+            'code' => $category->getCode(),
+            'parent' => $category->getParentCode(),
+            'updated' => $this->dateTimeNormalizer->normalize($category->getUpdated(), 'standard'),
```
ahocquard, line 26: Here, you are doing some logic inside a normalizer (get the good format of the date). You should use a read model for Public API instead, with all the correct data.

Then, normalizer is almost uselss (`toJson` function in the Public API read model is enough). Read model is easier to test, typed, and is a real contract.

https://github.com/akeneo/pim-community-dev/pull/16819

src/Akeneo/Category/back/Infrastructure/Storage/Query/SqlGetCategory.php
```diff
@@ -0,0 +1,52 @@
...
+        private Connection $connection
+    ) {
+    }
+
+    public function byCode(string $code): ?Category
+    {
+        $sql = <<<SQL
+SELECT category.code, category.updated, parentCategory.code as parentCode
+    ,(
+        SELECT JSON_OBJECTAGG(labels.locale, labels.label)
```
ahocquard, line 27: No query inside a `SELECT`.  Use CTE and JOIN.

src/Akeneo/Category/back/API/Command/CreateCategoryCommand.php
```diff
@@ -0,0 +1,53 @@
...
+class CreateCategoryCommand
+{
+    private function __construct(
+        private string  $code,
+        private ?string $parent,
+        private array   $labels,
+    ) {
+    }
+
+    public static function fromArray(array $data): self
```
ahocquard, line 22: In my opinion, it's not useful anymore with constructor + named parameters.

ahocquard (commented):
I would split it in several things:
- introduce the command, to use everywhere then inside the PIM 
- then, every part of the code should use the command (mostly impact on all tests, + import, + category API)

=> second point can be done in parallel

- the read part. I would take inspiration of https://github.com/akeneo/pim-community-dev/pull/16819/files instead.
Indeed, I prefer to use DTO and a dedicated search API for the API (not implemented in your API). It allowd more resuability.

Normalizer should not contain any logic (it's a `toJson`). We abuse it far too much at akeneo (our normalizers format are kind of API, but with an array instead of an object). Also, if something is doing an IO in a normalizer, ou will have n+1 issue.
I personally think that ReadOnly models for different usecases + `toJson` method inside the model is fine/simple for now.

=> once search builder is implemented, you can parallelize the introduction of the different read models for the different usecases. I don't know if there are too many use cases (API and CSV export could use the same read model as it's the same usecase, read model on category already exist, etc).

PR #16921, opened 2022-04-29
PIM-10420: Handle status computed by health check time on process tracker

pierrejolly:

**What's going wrong ?**
We forgot to handle the recalculation of the status based on the health time check : this will cause some bugs when jobs fails due to external down like a mysql pod crash. Now, users can seen that their jobs are failed.

- [x] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@
 - PIM-10421: Add missing translation key for delete button
 - PIM-10418: Simple and multi select values not showing if not imported with the correct letter case
 - PIM-10426: Fix empty array should be normalized as empty JSON object in Measurement Family API
+- PIM-10420: Handle status resolving when job crashes due to external issue (mysql crashes for example)
 - PIM-10416: Fix letter case issue when importing families
 - PIM-10427: Fix display of boolean value in variant axis
 
--- a/src/Akeneo/Platform/Job/back/Application/SearchJobExecution/Model/JobExecutionHealthCheck.php
+++ b/src/Akeneo/Platform/Job/back/Application/SearchJobExecution/Model/JobExecutionHealthCheck.php
@@ -1,0 +1,38 @@
+<?php
+
+namespace Akeneo\Platform\Job\Application\SearchJobExecution\Model;
+
+use Akeneo\Platform\Job\Domain\Model\Status;
+
+final class JobExecutionHealthCheck
+{
+    private const MAX_TIME_TO_UPDATE_HEALTH_CHECK = 5;
+    private const HEALTH_CHECK_INTERVAL = 5;
+
+    public function __construct(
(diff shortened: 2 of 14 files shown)


ghoussard (approved):

masacc (approved):

LaurentPetard (approved):

src/Akeneo/Platform/Job/back/Application/SearchJobExecution/SearchJobExecutionHandler.php
```diff
@@ -42,4 +48,25 @@ private function validateQuery(SearchJobExecutionQuery $query): void
...
     }
+
+    /**
+     * @param JobExecutionRow[] $jobExecutionRows
+     */
+    private function computeAndUpdateStatuses(array $jobExecutionRows): array
+    {
+        foreach ($jobExecutionRows as $jobExecutionRow) {
+            $currentStatus = $jobExecutionRow->getStatus();
+            $computedStatus = $this->computeStatus->fromCurrentStatusAndHealthCheckedAt(
```
ahocquard, line 59: why don't you just handle it this logic in the model itself, instead of a service? Why do you need to update the database?

src/Akeneo/Platform/Job/back/Application/SearchJobExecution/SearchJobExecutionHandler.php
```diff
@@ -42,4 +48,25 @@ private function validateQuery(SearchJobExecutionQuery $query): void
             throw new \InvalidArgumentException('Page can not be greater than 50 when no filter are set');
         }
     }
+
+    /**
+     * @param JobExecutionRow[] $jobExecutionRows
+     */
+    private function computeAndUpdateStatuses(array $jobExecutionRows): array
```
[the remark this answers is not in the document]
ghoussard: Yes not a great design indeed. The reason I do an update is that the status in db is not correct. So yes, we might just apply the transform on the read model and send it to frontend but the incorrect status in db will remains and we have to keep status in db correct because of filtering.
Our second idea with Pierre is to add a cron to do thoses updates but I don't know if it can be in the scope of the SLA.

src/Akeneo/Platform/Job/back/Application/SearchJobExecution/SearchJobExecutionHandler.php
ghoussard: It is in a service because of test. It is easier to test the service with a spec because we can mock ClockInterface and we cannot mock a `new DateTime()`.
For the update, see my previous reply.

And I don't know which field you are talking about

We can discuss about it IRL if you want

PR #16928, opened 2022-05-02
TIP-1567: be able to flag feature flags for ACL

ahocquard:

This PR allows to configure and display in the UI the feature flags from configuration. 

Quentin gave me a solution on how to do it in connectivity. I analyzed some other ways to do (filtering in `findAcl` method, filter php `annotation`, etc). Each method does not work as expected. Also, it's dangerous to try a new thing as it's ACLs that are saved in memcache on the fly (nothing is done at compile time...).

So, at the end, I took code introduced by @tseho for connectivity and generalized it for all feature flags. I didn't modify the ACL object itself, because there are 4 models to modify to pass the data... (ACL, PrivilegeACL, etc).

It impacts connectivity tests, as feature flag registry is the fake, with all features deactivated by default. That's why in EE, I didn't test if the ACL are hidden or not (it should be unit tests).
Connectivity context put a lot of effort at covering it, and I judged it's enough as coverage, according to the complexity of the ACL.

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc


tseho:
Do you want to highlight in https://github.com/akeneo/pim-community-dev/blob/51fd446ee4c6407e646a0055dfc060bf5055748e/src/Oro/Bundle/SecurityBundle/Acl/Persistence/AclPrivilegeRepository.php#L148 that the event is mutable ?
We talked about `$event = $this->dispatch...`.

ahocquard:
> it could be a good idea to improve it slightly.

Mutable event is not the biggest issue (at all) and I understand your pain at trying to find a solution (that helped me a lot). To be honest, the event is not even useful (we could filter directly in the method triggering the event). It was done for decoupling purpose I suppose. 

The plan is to remove ACL implementation completely in the coming quarters. So, not an issue in my opinion. 

As feature flags deadline is next week, I don't plan to modify it for now.

tseho (approved):

PR #16955, opened 2022-05-05
CPM: clean product's getId() and setId()

nmarniesse:

Remove all remaining usages of getId/setId on Product. 

- setId is removed
- getId is kept because the shared interfaces (EntityWithValuesInterface, ...) but it throws an exception now
- a new `isNew` method is added (before we checked the id is null or not)

### DIFF
--- a/src/Akeneo/Pim/Enrichment/Bundle/Controller/InternalApi/ProductController.php
+++ b/src/Akeneo/Pim/Enrichment/Bundle/Controller/InternalApi/ProductController.php
@@ -468,7 +468,7 @@ class ProductController
 
         // don't filter during creation, because identifier is needed
         // but not sent by the frontend during creation (it sends the sku in the values)
-        if (null !== $product->getCreated() && $product->isVariant()) {
+        if (!$product->isNew() && $product->isVariant()) {
             $data = $this->productAttributeFilter->filter($data);
         }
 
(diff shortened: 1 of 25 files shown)


src/Akeneo/Pim/Enrichment/Component/Product/Model/AbstractProduct.php
```diff
@@ -82,18 +82,24 @@ public function __construct()
     }
 
     /**
+     * @TODO: remove this method?
```
nmarniesse, line 85: TODO. Remove or keep (and return uuid as string)

src/Akeneo/Pim/Enrichment/Bundle/Controller/InternalApi/ProductController.php
```diff
@@ -468,7 +468,7 @@ protected function updateProduct(ProductInterface $product, array $data)
 
         // don't filter during creation, because identifier is needed
         // but not sent by the frontend during creation (it sends the sku in the values)
-        if (null !== $product->getId() && $product->isVariant()) {
+        if (null !== $product->getCreated() && $product->isVariant()) {
```
cemorin, line 471: $product->isNew() can be used?

cemorin (approved):

pierallard (approved):

PR #16963, opened 2022-05-06
CXP-1187: Add catalog scopes

tseho:

It was requested that the ACLs behind the new scopes are not visible from the UI, to do so, we added a way of hiding ACLs in this commit https://github.com/akeneo/pim-community-dev/pull/16963/commits/9c8e16c1aaebe91b6024f769b85b577480c44d65

Screenshot_2022-05-06_18-29-01

Screenshot_2022-05-06_18-28-35

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc


guillaumefer-akeneo (approved):

ahocquard:
Why not the same solution as for feature flags?

tseho:
> Why not the same solution as for feature flags?
Because I want the ACL to still exist.

RiffVentura (approved):

components/catalogs/back/tests/Unit/Infrastructure/Security/CatalogScopeMapperTest.php
```diff
@@ -0,0 +1,138 @@
...
+                'scope' => 'unknown',
+                'result' => [],
+            ],
+        ];
+    }
+
+    /**
+     * @dataProvider messages
+     */
+    public function testItReturnsMessagesForOneScope(string $scope, $expected): void
```
masacc, line 65: I wonder if that kind of test is really useful. I mean it's almost like testing constant and array access.

masacc (approved):

willy-ahva (approved):

components/catalogs/back/tests/Unit/Infrastructure/Security/CatalogScopeMapperTest.php
tseho: Agree, but the ScopeRegistry will call the methods not the constants, it's only there to guarantee that the internal logic is handled correctly.

It's not the most valuable test but it's a low cost one.

PR #16992, opened 2022-05-10
[adr-uuid] Store success on db

pierallard:
Before this PR, we have to check if every state was already done
The issue is that our biggest client takes a very long time to compute.
With this solution, we store the success of the migration, and only check for it when we run the cmd.

### DIFF
--- a/src/Akeneo/Pim/Enrichment/Bundle/Command/MigrateToUuid/MigrateToUuidCommand.php
+++ b/src/Akeneo/Pim/Enrichment/Bundle/Command/MigrateToUuid/MigrateToUuidCommand.php
@@ -64,7 +64,7 @@ class MigrateToUuidCommand extends Command
         $withStats = $input->getOption('with-stats');
         $context = new Context($input->getOption('dry-run'), $withStats);
 
-        if (!$this->shouldBeExecuted()) {
+        if ($this->isAlreadySuccessfull()) {
             $this->logger->notice('No step should be executed. Skip the migration.');
 
             return self::SUCCESS;
@@ -122,9 +122,13 @@ class MigrateToUuidCommand extends Command
                 );
             }
 
+            $this->success();
             $this->logger->notice('Migration done!', ['migration_duration_in_second' => time() - $startMigrationTime]);
-        } finally {
-            $this->stop();
+        } catch (\Throwable $e) {
+            $this->logger->error('Migration failed!', ['message' => $e->getMessage(), 'exception' => $e]);
+            $this->fail();
+
+            return Command::FAILURE;
         }
 
         return Command::SUCCESS;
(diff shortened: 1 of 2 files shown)


src/Akeneo/Pim/Enrichment/Bundle/Command/MigrateToUuid/MigrateToUuidCommand.php
```diff
@@ -122,9 +122,13 @@ protected function execute(InputInterface $input, OutputInterface $output): int
                 );
             }
 
+            $this->success();
             $this->logger->notice('Migration done!', ['migration_duration_in_second' => time() - $startMigrationTime]);
-        } finally {
-            $this->stop();
+        } catch (\Throwable $e) {
+            $this->logger->error('Migration failed!', ['message' => $e->getMessage()]);
```
mmetayer, line 128:
```suggestion
            $this->logger->error('Migration failed!', ['message' => $e->getMessage(), 'exception' => $e]);
```
(this way we'll have the exception stack trace)

src/Akeneo/Pim/Enrichment/Bundle/Command/MigrateToUuid/MigrateToUuidCommand.php
```diff
@@ -143,7 +147,19 @@ private function start()
         ]);
     }
 
-    private function stop()
+    private function success()
```
mmetayer, line 150:
```suggestion
    private function success(): void
```

src/Akeneo/Pim/Enrichment/Bundle/Command/MigrateToUuid/MigrateToUuidCommand.php
```diff
@@ -143,7 +147,19 @@ private function start()
...
+            UPDATE `pim_one_time_task`
+            SET status=:status, end_time=NOW()
+            WHERE code=:code
+        SQL, [
+            'code' => self::$defaultName,
+            'status' => 'finished',
+        ]);
+    }
+
+    private function fail()
```
mmetayer, line 162:
```suggestion
    private function fail(): void
```

cemorin (approved):

src/Akeneo/Pim/Enrichment/Bundle/Command/MigrateToUuid/MigrateToUuidCommand.php
pierallard: It will not be an issue for the logger to give exception like that ?

mmetayer: nope, it will be normalized as well, see https://github.com/Seldaek/monolog/blob/60ad5183b5e5d6c9d4047e9f3072d36071dcc161/src/Monolog/Formatter/JsonFormatter.php#L170

nmarniesse (approved):

PR #17023, opened 2022-05-13
CPM-643: Add service API for product PQB

nmarniesse:
A new query is created to return product uuids based on a query.
As we discuss the query is an array for now. The format is the same as the one used in the public API. It can be a proper object later.

- Add a query bus (same as CPM-625)
- Create a new simpler ProductQueryBuilderInterface in the Enrichment\Product namespace. But use the older one in order to take advantage of the existing filters, sorters, ...
- Inject a fetcher in the cursor so the cursor don't make the query itself -> better responsibility
- handle permissions


src/Akeneo/Pim/Enrichment/Product/back/Application/PQB/GetProductUuidsHandler.php
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    public function __invoke(GetProductUuids $getProductUuids): ProductUuidCursor
+    {
+        $violations = $this->validator->validate($getProductUuids);
+        if (0 < $violations->count()) {
+            throw new ViolationsException($violations);
+        }
+
+        $this->applyProductSearchQueryParametersToPQB->apply($this->pqb, $getProductUuids->searchFilters(), null, null, null);
```
nmarniesse, line 35: applyProductSearchQueryParametersToPQB is in the former Enrichment namespace. Do we re-implement it in the new one?

src/Akeneo/Pim/Enrichment/Product/back/Infrastructure/Query/Elasticsearch/ProductQueryBuilderAdapter.php
```diff
@@ -0,0 +1,94 @@
...
+use Akeneo\UserManagement\Component\Repository\UserRepositoryInterface;
+use Webmozart\Assert\Assert;
+
+/**
+ * This class is an adapter between the former implementation of PQB and the new one.
+ *
+ * @copyright 2022 Akeneo SAS (https://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+final class ProductQueryBuilderAdapter extends AbstractEntityWithValuesQueryBuilder implements ProductQueryBuilderInterface
```
nmarniesse, line 28: See if we can remove the extends Abstract... and recopy only the methods that are useful for us.

pierallard (approved):

cemorin (approved):

src/Akeneo/Pim/Enrichment/Product/back/Infrastructure/Query/Elasticsearch/ProductQueryBuilderAdapter.php
mmetayer: Hard to do bc ApplyProductSearchQueryParametersToPQB::apply() requires a "legacy" productquerybuilderinterface (despite only using the `addFilter()` method). We could try "cutting" the legcy interface into smaller interfaces, but it's clearly not the scope of this PR. I'll add a ticket in our backlog for that

PR #17076, opened 2022-05-19
CXP-1182: use HAL paginator

masacc:

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/components/catalogs/back/src/Infrastructure/Controller/Public/GetCatalogsAction.php
+++ b/components/catalogs/back/src/Infrastructure/Controller/Public/GetCatalogsAction.php
@@ -7,8 +7,10 @@ namespace Akeneo\Catalogs\Infrastructure\Controller\Public;
 use Akeneo\Catalogs\Infrastructure\Security\DenyAccessUnlessGrantedTrait;
 use Akeneo\Catalogs\Infrastructure\Security\GetCurrentUserIdTrait;
 use Akeneo\Catalogs\ServiceAPI\Messenger\QueryBus;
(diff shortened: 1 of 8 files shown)


components/catalogs/back/src/Infrastructure/Controller/Public/GetCatalogsAction.php
```diff
@@ -31,24 +33,45 @@ public function __construct(
...
+        return new JsonResponse($this->paginate($catalogs, $page, $limit), Response::HTTP_OK);
+    }
+
+    /**
+     * @param array<Catalog> $catalogs
+     * @return array<array-key, mixed>
+     */
+    private function paginate(array $catalogs, int $page, int $limit): array
+    {
+        $items = \array_map(fn (Catalog $catalog) => $this->normalizer->normalize($catalog, 'external_api'), $catalogs);
```
tseho, line 64: No need for array map, you can directly normalize the array. It's a mistake I made.

components/catalogs/back/src/Infrastructure/Symfony/Resources/config/services.yml
```diff
@@ -2,6 +2,8 @@ services:
     _defaults:
         autowire: true
         autoconfigure: true
+        bind:
```
tseho, line 5: I don't like bind, I'd rather have an alias and inject explicitly the offset hal paginator in the controller

tseho (approved):

masacc:
update

PR #17078, opened 2022-05-20
BH-1159: Rework BatchCommand to use execution ID without batch code

jmleroux:
Rework the BatchCommand to make it ignore the batch `code` parameter when an execution ID (`execution` parameter`) is already provided. Indeed, the batch code is part of the execution parameter.

This would allow us to launch some batches by knowing only the execution ID.

If the execution ID is not provided, the batch code is still used as before.

Also added a `getJobRepository` in the interface because it was internally required and was using polymorphic calls.

### DIFF
--- a/src/Akeneo/Tool/Bundle/BatchBundle/Command/BatchCommand.php
+++ b/src/Akeneo/Tool/Bundle/BatchBundle/Command/BatchCommand.php
@@ -146,6 +146,7 @@ class BatchCommand extends Command
                 $jobExecution->setExecutionContext(new ExecutionContext());
             }
             $jobInstance = $jobExecution->getJobInstance();
+            $job = $this->jobRegistry->get($jobInstance->getJobName());
         } else {
             $code = $input->getArgument('code');
             $jobInstance = $this->getJobManager()->getRepository($this->jobInstanceClass)->findOneBy(['code' => $code]);
--- a/src/Akeneo/Tool/Bundle/BatchBundle/Command/BatchCommand.php
+++ b/src/Akeneo/Tool/Bundle/BatchBundle/Command/BatchCommand.php
@@ -15,10 +15,12 @@ use Akeneo\Tool\Component\Batch\Job\JobParametersFactory;
 use Akeneo\Tool\Component\Batch\Job\JobParametersValidator;
 use Akeneo\Tool\Component\Batch\Job\JobRegistry;
 use Akeneo\Tool\Component\Batch\Job\JobRepositoryInterface;
+use Akeneo\Tool\Component\Batch\Model\JobExecution;
 use Akeneo\Tool\Component\Batch\Model\JobInstance;
 use Akeneo\Tool\Component\Batch\Model\StepExecution;
 use Doctrine\ORM\EntityManagerInterface;
-use Psr\Log\LoggerInterface;
+use Doctrine\Persistence\ObjectManager;
+use Monolog\Logger;
 use Symfony\Bridge\Doctrine\ManagerRegistry;
 use Symfony\Bridge\Monolog\Handler\ConsoleHandler;
 use Symfony\Component\Console\Command\Command;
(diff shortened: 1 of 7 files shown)


src/Akeneo/Tool/Bundle/BatchBundle/Command/BatchCommand.php
```diff
@@ -40,70 +42,26 @@
...
-        JobRepositoryInterface $jobRepository,
-        ManagerRegistry $doctrine,
-        ValidatorInterface $validator,
-        Notifier $notifier,
-        JobRegistry $jobRegistry,
-        JobParametersFactory $jobParametersFactory,
-        JobParametersValidator $jobParametersValidator,
-        string $jobInstanceClass,
-        string $jobExecutionClass
+        private Logger $logger,
```
jmleroux, line 52: We need a real `Logger`. Using the interface is incorrect and creates polymorphic calls.

src/Akeneo/Tool/Bundle/BatchBundle/Command/BatchCommand.php
```diff
@@ -184,18 +131,8 @@ protected function execute(InputInterface $input, OutputInterface $output): int
...
-            $this->validateJobParameters($job, $jobInstance, $jobParameters, $code);
-            $jobExecution = $job->getJobRepository()->createJobExecution($job, $jobInstance, $jobParameters);
-
-            $username = $input->getOption('username');
-            if (null !== $username) {
-                $jobExecution->setUser($username);
-                $job->getJobRepository()->updateJobExecution($jobExecution);
-            }
-        } else {
+        if (null !== $executionId) {
```
jmleroux, line 134: If the execution ID is provided, I use it instead of the batch code.

src/Akeneo/Tool/Bundle/BatchBundle/Command/BatchCommand.php
```diff
@@ -304,52 +259,35 @@ protected function writeExceptions(OutputInterface $output, array $exceptions, $
...
-     * @param JobParameters $jobParameters
-     * @param string        $code
-     *
      * @throws \RuntimeException
      */
     protected function validateJobParameters(JobInterface $job, JobInstance $jobInstance, JobParameters $jobParameters, string $code) : void
     {
         // We merge the JobInstance from the JobManager EntityManager to the DefaultEntityManager
         // in order to be able to have a working UniqueEntity validation
-        $defaultJobInstance = $this->getDefaultEntityManager()->merge($jobInstance);
```
jmleroux, line 352: unused vraiable

src/Akeneo/Tool/Bundle/BatchBundle/Command/BatchCommand.php
jmleroux: Because of this method not present in the interface: https://github.com/akeneo/pim-community-dev/pull/17078/files#diff-5d49f0ca86c6c290b1e379f7243e9626d729203151640bea1146181e95f2ebceR110

mmetayer, line 52: I would let the interface here, but test if it's an instance of Monolog/Logger before pushing the handler

src/Akeneo/Tool/Bundle/BatchBundle/Command/BatchCommand.php
```diff
@@ -382,17 +316,13 @@ private function getErrorMessages(ConstraintViolationList $errors): string
...
-     * @return array
      */
-    private function decodeConfiguration($data): array
+    private function decodeConfiguration(string $data): array
     {
-        $config = json_decode($data, true);
+        $config = \json_decode($data, true);
 
-        switch (json_last_error()) {
+        switch (\json_last_error()) {
```
mmetayer, line 325: maybe we could replace the whole block by 
```php
$config = \json_decode($data, true, 512, JSON_THROW_ON_ERROR);
```
?

mmetayer (approved):

src/Akeneo/Tool/Bundle/BatchBundle/Command/BatchCommand.php
jmleroux: :+1: 
Much simpler and we won't have to check for every case.

PR #17096, opened 2022-05-24
RAB-623: Decouple SqlFindRequiredValueKeyCollectionForChannelAndLocales

pierrejolly:

In this PR I added the ability to : 
* retrieve Channel by codes
* retrieve Locales by codes

### DIFF
--- a/src/Akeneo/Channel/back/Infrastructure/Query/Cache/CachedFindChannels.php
+++ b/src/Akeneo/Channel/back/Infrastructure/Query/Cache/CachedFindChannels.php
@@ -12,10 +12,22 @@ use Akeneo\Tool\Component\StorageUtils\Cache\CachedQueryInterface;
 final class CachedFindChannels implements FindChannels, CachedQueryInterface
 {
     private ?array $cache = null;
+    private array $cacheByCodes = [];
 
     public function __construct(
         private FindChannels $findChannels
     ) {
+    }
+
+    public function findByCodes(array $codes): array
+    {
+        $cacheKey = implode('', $codes);
+
+        if (empty($this->cacheByCodes[$cacheKey])) {
+            $this->cacheByCodes[$cacheKey] = $this->findChannels->findByCodes($codes);
+        }
+
+        return $this->cacheByCodes[$cacheKey];
     }
 
     public function findAll(): array
--- a/src/Akeneo/Channel/back/Infrastructure/Query/Cache/CachedFindLocales.php
+++ b/src/Akeneo/Channel/back/Infrastructure/Query/Cache/CachedFindLocales.php
@@ -14,6 +14,7 @@ final class CachedFindLocales implements FindLocales, CachedQueryInterface
 {
     private ?array $indexedCache = null;
     private ?array $cache = null;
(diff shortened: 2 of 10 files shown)


src/Akeneo/Channel/back/Infrastructure/Query/Cache/CachedFindLocales.php
```diff
@@ -14,6 +14,7 @@ final class CachedFindLocales implements FindLocales, CachedQueryInterface
 {
     private ?array $indexedCache = null;
     private ?array $cache = null;
+    private array $cacheByCodes = [];
```
ghoussard, line 17: Why you don't use an single cache property ? An array index by code should do the coffee for the 3 methods

src/Akeneo/Channel/back/Infrastructure/Query/Cache/CachedFindChannels.php
```diff
@@ -12,12 +12,24 @@
...
+    private array $cacheByCodes = [];
 
     public function __construct(
         private FindChannels $findChannels
     ) {
     }
 
+    public function findByCodes(array $codes): array
+    {
+        $cacheKey = implode('', $codes);
```
ghoussard, line 24: This a bit wtf to concatenate codes to build an index (as these codes does not have any relation)

src/Akeneo/Channel/back/Infrastructure/Query/Cache/CachedFindChannels.php
```diff
@@ -12,12 +12,24 @@
 final class CachedFindChannels implements FindChannels, CachedQueryInterface
 {
     private ?array $cache = null;
+    private array $cacheByCodes = [];
```
ghoussard, line 15: Here you can use an single cache property indexed by code

ghoussard (approved):

src/Akeneo/Channel/back/Infrastructure/Query/Cache/CachedFindChannels.php
ghoussard: For instance, image you fetch first `print` and `ecommerce` channels so you cached them with printecommerce as index. Then you would fetch `ecommerce` and `mobile` channels, your cache will not works because `ecommerce` is concatenate with `print`. If you have a single array with a code as index, you will be able to fetch `ecommerce` from the cache and `mobile` from the query

pierrejolly: Yes, it's just a "quick win", I'm looking for a better solution but I would like to avoid to loop over all the results to index the array by code

src/Akeneo/Channel/back/Infrastructure/Query/Cache/CachedFindChannels.php
pierrejolly: I would like to avoid to loop over all the results to index the array by code :/

pierrejolly:
After a discussion with the team we decided to avoid multiplying the number of available query in the public API

PR #17098, opened 2022-05-24
CXP-1209: Add state column to the catalog table

willy-ahva:

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/components/catalogs/back/src/Infrastructure/Installer/InstallerSubscriber.php
+++ b/components/catalogs/back/src/Infrastructure/Installer/InstallerSubscriber.php
@@ -38,6 +38,7 @@ class InstallerSubscriber implements EventSubscriberInterface
                 id BINARY(16) NOT NULL PRIMARY KEY,
                 name VARCHAR(255) NOT NULL,
                 owner_id INT NOT NULL,
+                is_enabled TINYINT NOT NULL DEFAULT 0,
                 created DATETIME NOT NULL DEFAULT CURRENT_TIMESTAMP,
                 updated DATETIME NOT NULL DEFAULT CURRENT_TIMESTAMP,
                 INDEX idx_owner (owner_id),
--- a/components/catalogs/back/src/Infrastructure/Normalizer/CatalogNormalizer.php
+++ b/components/catalogs/back/src/Infrastructure/Normalizer/CatalogNormalizer.php
@@ -35,7 +35,7 @@ class CatalogNormalizer implements NormalizerInterface, CacheableSupportsMethodI
         return [
             'id' => $object->getId(),
             'name' => $object->getName(),
-            'enabled' => false,
+            'enabled' => $object->isEnabled(),
         ];
     }
 }
--- a/components/catalogs/back/src/ServiceAPI/Model/Catalog.php
+++ b/components/catalogs/back/src/ServiceAPI/Model/Catalog.php
@@ -14,6 +14,7 @@ final class Catalog
         private string $id,
         private string $name,
         private int $ownerId,
+        private bool $enabled,
     ) {
     }
 
@@ -32,12 +33,18 @@ final class Catalog
         return $this->ownerId;
     }
 
+    public function isEnabled(): bool
+    {
+        return $this->enabled;
+    }
+
     public function withNewName(string $name): self
     {
         return new self(
             $this->id,
             $name,
             $this->ownerId,
+            $this->enabled,
         );
     }
 }
--- a/components/catalogs/back/tests/Integration/Infrastructure/Controller/Public/DeleteCatalogActionTest.php
+++ b/components/catalogs/back/tests/Integration/Infrastructure/Controller/Public/DeleteCatalogActionTest.php
@@ -41,7 +41,7 @@ class DeleteCatalogActionTest extends IntegrationTestCase
         $this->commandBus->execute(new CreateCatalogCommand(
             'db1079b6-f397-4a6a-bae4-8658e64ad47c',
             'Store US',
-            $this->tokenStorage->getToken()->getUser()->getId()
+            $this->tokenStorage->getToken()->getUser()->getId(),
         ));
 
         $this->client->request(
@@ -59,7 +59,7 @@ class DeleteCatalogActionTest extends IntegrationTestCase
         $this->commandBus->execute(new CreateCatalogCommand(
             'db1079b6-f397-4a6a-bae4-8658e64ad47c',
             'Store US',
-            $this->tokenStorage->getToken()->getUser()->getId()
+            $this->tokenStorage->getToken()->getUser()->getId(),
         ));
 
         $this->client->request(
--- a/components/catalogs/back/tests/Integration/Infrastructure/Controller/Public/GetCatalogActionTest.php
+++ b/components/catalogs/back/tests/Integration/Infrastructure/Controller/Public/GetCatalogActionTest.php
@@ -38,7 +38,7 @@ class GetCatalogActionTest extends IntegrationTestCase
         $this->commandBus->execute(new CreateCatalogCommand(
             'db1079b6-f397-4a6a-bae4-8658e64ad47c',
             'Store US',
-            $this->tokenStorage->getToken()->getUser()->getId()
+            $this->tokenStorage->getToken()->getUser()->getId(),
         ));
 
         $this->client->request(
--- a/upgrades/schema/Version_7_0_20220524134005_add_is_enabled_column_to_catalog.php
+++ b/upgrades/schema/Version_7_0_20220524134005_add_is_enabled_column_to_catalog.php
@@ -1,0 +1,36 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Pim\Upgrade\Schema;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+
+/**
+ * @author    Willy Mesnage <willy.mesnage@akeneo.com>
+ * @copyright 2022 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+final class Version_7_0_20220524134005_add_is_enabled_column_to_catalog extends AbstractMigration
+{
+    public function up(Schema $schema): void
+    {
+        $this->skipIf(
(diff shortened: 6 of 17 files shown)


components/catalogs/back/src/ServiceAPI/Command/CreateCatalogCommand.php
```diff
@@ -20,6 +21,8 @@ public function __construct(
         #[Assert\Length(min: 1, max: 255)]
         private string $name,
         private int $ownerId,
+        #[Assert\Choice(choices: Catalog::STATES)]
+        private string $state,
```
tseho, line 25: Why can you specify the state ? shouldn't be internal to the feature ?

components/catalogs/back/src/Infrastructure/Normalizer/CatalogNormalizer.php
```diff
@@ -35,7 +35,7 @@ public function normalize(mixed $object, ?string $format = null, array $context
         return [
             'id' => $object->getId(),
             'name' => $object->getName(),
-            'enabled' => false,
+            'state' => $object->getState(),
```
tseho, line 38:
```suggestion
            'enabled' => $object->isEnabled(),
```

I don't think we should leak the state

components/catalogs/back/src/ServiceAPI/Model/Catalog.php
```diff
@@ -10,11 +10,21 @@
  */
 final class Catalog
 {
+    public const ENABLED = 'enabled';
+    public const DISABLED = 'disabled';
+    public const STATES = [self::DISABLED, self::ENABLED];
```
tseho, line 15:
```suggestion
    public const ENABLED = 'enabled';
    public const DISABLED = 'disabled';
    public const STATES = [self::DISABLED, self::ENABLED];
```

IHMO, should be private, it shouldn't leak outside your model.

upgrades/schema/Version_7_0_20220524134005_add_state_column_to_catalog.php
```diff
@@ -0,0 +1,30 @@
...
+ * @copyright 2022 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+final class Version_7_0_20220524134005_add_state_column_to_catalog extends AbstractMigration
+{
+    public function up(Schema $schema): void
+    {
+        $this->addSql(<<<SQL
+        ALTER TABLE akeneo_catalog
+        ADD state VARCHAR(20) NOT NULL;
```
tseho, line 21:
```suggestion
        ADD state VARCHAR(20) NOT NULL DEFAULT "disabled";
```

components/catalogs/back/tests/Integration/Infrastructure/Controller/Public/DeleteCatalogActionTest.php
```diff
@@ -41,7 +41,7 @@ public function testItDeletesTheCatalog(): void
         $this->commandBus->execute(new CreateCatalogCommand(
             'db1079b6-f397-4a6a-bae4-8658e64ad47c',
             'Store US',
-            $this->tokenStorage->getToken()->getUser()->getId()
+            $this->tokenStorage->getToken()->getUser()->getId(),
```
tseho, line 44: I agree.
It is something that we can enforce with phpstan ?

components/catalogs/back/tests/Integration/Infrastructure/Controller/Public/GetCatalogActionTest.php
```diff
@@ -58,6 +58,7 @@ public function testItFindsTheCatalog(): void
         Assert::assertSame('db1079b6-f397-4a6a-bae4-8658e64ad47c', $payload['id']);
         Assert::assertSame('Store US', $payload['name']);
         Assert::assertSame(false, $payload['enabled']);
+        Assert::assertSame(false, $payload['enabled']);
```
tseho, line 61:
```suggestion
```
We were already testing it, the line just before.

upgrades/schema/Version_7_0_20220524134005_add_is_enabled_column_to_catalog.php
```diff
@@ -0,0 +1,30 @@
...
+/**
+ * @author    Willy Mesnage <willy.mesnage@akeneo.com>
+ * @copyright 2022 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+final class Version_7_0_20220524134005_add_is_enabled_column_to_catalog extends AbstractMigration
+{
+    public function up(Schema $schema): void
+    {
+        $this->addSql(<<<SQL
```
tseho, line 19: You need to check in $schema if the column already exists.

```
$this->skipIf(
            $schema->getTable('akeneo_catalog')->hasColumn('is_enabled'),
            'is_enabled column already exists in akeneo_catalog'
        );
```

upgrades/test_schema/Version_7_0_20220524134005_add_is_enabled_column_to_catalog_Integration.php
```diff
@@ -0,0 +1,52 @@
...
+    use ExecuteMigrationTrait;
+
+    private const MIGRATION_LABEL = '_7_0_20220524134005_add_is_enabled_column_to_catalog';
+
+    protected function getConfiguration()
+    {
+        return null;
+    }
+
+    public function test_id_adds_is_enabled_column(): void
```
tseho, line 20:
```suggestion
    public function test_it_adds_is_enabled_column(): void
```
cc @masacc

tseho (approved):

components/catalogs/back/tests/Integration/Infrastructure/Controller/Public/DeleteCatalogActionTest.php
willy-ahva: Yes, this rule, and the one that removes the not needed imports.

upgrades/schema/Version_7_0_20220524134005_add_is_enabled_column_to_catalog.php
willy-ahva: done

guillaumefer-akeneo (approved):

willy-ahva:
warmup

upgrades/schema/Version_7_0_20220524134005_add_is_enabled_column_to_catalog.php
```diff
@@ -0,0 +1,35 @@
...
+    public function up(Schema $schema): void
+    {
+        $this->skipIf(
+            $schema->getTable('akeneo_catalog')->hasColumn('is_enabled'),
+            'is_enabled column already exists in akeneo_catalog'
+        );
+
+        $this->addSql(<<<SQL
+        ALTER TABLE akeneo_catalog
+        ADD is_enabled TINYINT NOT NULL DEFAULT 0;
```
pierallard, line 26: I don't know this table.
Do you have any idea about the max size of this table ? Don't hesitate to ask Popeye or Sophie. If this table is big on some clients, you will be limited by the 20 min max of migration during deployment
Second remark: ALTER a TABLE freeze the table. It means that you will lock any query on this table during the migration.
A good practice is to add
`, ALGORITHM=INPLACE, LOCK=NONE` at the end of the ALTER TABLE. This is longer, but it does not lock the table.

pierallard (commented):
Some comments before GTM plz

upgrades/schema/Version_7_0_20220524134005_add_is_enabled_column_to_catalog.php
willy-ahva: Dear Pierallard, thanks for your review. As the feature is hidden, this table is not used for the moment. So, there is no risk to exceed 20 min long. Moreover, this table will have 1250 long at the maximum in use. Thanks for the good practice I'll do it <3

components/catalogs/back/src/Infrastructure/Installer/InstallerSubscriber.php
```diff
@@ -38,6 +38,7 @@ public function updateSchema(): void
                 id BINARY(16) NOT NULL PRIMARY KEY,
                 name VARCHAR(255) NOT NULL,
                 owner_id INT NOT NULL,
+                is_enabled TINYINT NOT NULL DEFAULT 0,
```
mmetayer, line 41: this doesn't comply with your migration (I'm pretty sure the order of the columns is checked in the `test_database` job), you should either add it at the end (after `updated`, or update the migration to insert it after `owner_id`)

upgrades/schema/Version_7_0_20220524134005_add_is_enabled_column_to_catalog.php
pierallard: Btw, when the Dodo's team will prepare the migration, they will test it to check if it's ZDD, so if your query takes more than 20 minutes, deployment will be cancelled, that's the reason we take a big look at the migrations.
Change the ALGORITHM and LOCK in your query, and that's GTM for me.

willy-ahva: already done

upgrades/schema/Version_7_0_20220524134005_add_is_enabled_column_to_catalog.php
```diff
@@ -0,0 +1,36 @@
...
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+final class Version_7_0_20220524134005_add_is_enabled_column_to_catalog extends AbstractMigration
+{
+    public function up(Schema $schema): void
+    {
+        $this->skipIf(
+            $schema->getTable('akeneo_catalog')->hasColumn('is_enabled'),
+            'is_enabled column already exists in akeneo_catalog'
+        );
```
nmarniesse, line 22: Note for the future the skipIf method is not great unfortunately. Because it does not mark the migration as done, so at each deployment the process will try to migrate this class. It's micro-optimization, it can be merged like that.

nmarniesse (approved):

PR #17101, opened 2022-05-24
BH-1056: Upgrade phpseclib to v3.0

jmleroux:
Upgrade phpseclib to be less prone to security breach and allow us to configure the .rnd file path ([here](https://github.com/akeneo/pim-community-dev/pull/16149/files#diff-67cbf1d31bcf9bb04169cf1e1736d9c0f99be6d1bb16a5f0777e94465d87926bR3))

Putting it in the /tmp directory avoid to create it at the project root and add it to the repo if we forget to ignore it.

See also the official for v3 improvements: https://phpseclib.com/docs/why#phpseclib-30-vs-phspeclib-10--20


ghoussard (approved):

ValentinMumble (approved):

ValentinMumble:
Hello @jmleroux, any news on this PR?

jmleroux:
I think it's ready, but need to be tested with real SSO access to ensure backward compatibility

src/Akeneo/Connectivity/Connection/back/Infrastructure/Apps/AsymmetricKeysGenerator.php
```diff
@@ -5,38 +5,44 @@
...
+    {
+        /*
+         * Following algorithm is the implementation documented by the phpseclib library
+         * in order to generate self-signed public key and private key.
+         * see http://phpseclib.sourceforge.net/x509/guide.html#selfsigned
+         */
+        RSA::setOpenSSLConfigPath($this->openSSLConfigPath);
+        /** @var RSA\PrivateKey $privateKey */
+        $privateKey = RSA::createKey();
+        $publicKey = $privateKey->getPublicKey();
```
masacc, line 31: To be fully compatible with how the public key is used on the app side, we need to add this line of configuration. It seems that in phpseclib3 they moved to a more secure one called PSS. All of this is a bit obscure to me, so don't hesitate to challenge my proposal.
```suggestion
        RSA::setOpenSSLConfigPath($this->openSSLConfigPath);
        /** @var RSA\PrivateKey $privateKey */
        $privateKey = RSA::createKey();
        $privateKey = $privateKey->withPadding(RSA::SIGNATURE_PKCS1);
        $publicKey = $privateKey->getPublicKey();
```

To see a working example, take a look at the demo app that is using `lcobucci/jwt` lib (4.1.5). The error appears in `src/Signer/OpenSSL.php` when trying to do `openssl_pkey_get_details($key)` on the public key. It returns a type `-1` meaning unknown. Adding this line of config makes it return a type `OPENSSL_KEYTYPE_RSA`.

tseho:
Hello,
Can you wait a little more before merging this PR ?
We confirmed that it will probably break User Authentication for Apps.
We will assess which apps are affected in production and Mégane will test manually each one.

tseho (approved):

olivierpontier (approved):

PR #17103, opened 2022-05-25
BH-1075-1:Add a check between MySQL and Elasticsearch

niarcotyk:

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/src/Akeneo/Pim/Enrichment/Bundle/Command/IndexProductCommand.php
+++ b/src/Akeneo/Pim/Enrichment/Bundle/Command/IndexProductCommand.php
@@ -136,12 +136,12 @@ class IndexProductCommand extends Command
 
     private function getAllProductIdentifiers(int $batchSize): iterable
     {
-        $formerId = 0;
+        $formerId = NULL;
         $sql = <<< SQL
-SELECT id, identifier
+SELECT CONCAT('product_',BIN_TO_UUID(uuid)) AS _id, BIN_TO_UUID(uuid) AS uuid, identifier
 FROM pim_catalog_product
-WHERE id > :formerId
-ORDER BY id ASC
+WHERE (CASE WHEN :formerId IS NULL THEN TRUE ELSE uuid > :formerId END)
+ORDER BY uuid ASC
 LIMIT :limit
 SQL;
         while (true) {
@@ -152,7 +152,7 @@ SQL;
                     'limit' => $batchSize,
                 ],
                 [
-                    'formerId' => \PDO::PARAM_INT,
+                    'formerId' => \PDO::PARAM_STR,
                     'limit' => \PDO::PARAM_INT,
                 ]
             )->fetchAllAssociative();
@@ -161,8 +161,33 @@ SQL;
                 return;
             }
 
-            $formerId = (int)end($rows)['id'];
-            yield array_column($rows, 'identifier');
+            $formerId = end($rows)['uuid'];
+
+            $existingMysqlIdentifiers = array_column($rows, '_id');
+
+            $results = $this->productAndProductModelClient->search([
+                'query' => [
+                    'ids' => [
+                        'values' =>
+                            $existingMysqlIdentifiers
+                    ]
+                ],
+                '_source' => false,
+                'size' => $batchSize
+            ]);
+
+            $esIdentifiers = array_map(function($doc) {
+                return $doc['_id'];
+            }, $results["hits"]["hits"]);
+
+            $diff = array_reduce($rows,function($carry,$item) use($esIdentifiers){
+                if(!in_array($item['_id'],$esIdentifiers)){
+                    $carry[] = $item['identifier'];
+                }
+                return $carry;
+            },[]);
+
+            yield $diff;
         }
     }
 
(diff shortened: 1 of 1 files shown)


LudovicTOURMAN (commented):
:heart:

src/Akeneo/Pim/Enrichment/Bundle/Command/IndexProductCommand.php
```diff
@@ -67,6 +67,12 @@ protected function configure()
                 InputOption::VALUE_NONE,
                 'Index all existing products into Elasticsearch'
             )
+            ->addOption(
+                'diff',
+                'd',
+                InputOption::VALUE_NONE,
+                'Display diff between MySQL and Elasticsearch'
```
docteurklein, line 74: not display, but actually "ignore" products that are already there

src/Akeneo/Pim/Enrichment/Bundle/Command/IndexProductCommand.php
```diff
@@ -89,6 +95,16 @@ protected function execute(InputInterface $input, OutputInterface $output): int
         if (true === $input->getOption('all')) {
             $chunkedProductIdentifiers = $this->getAllProductIdentifiers($batchSize);
             $productCount = 0;
+        } elseif(true === $input->getOption('diff')) {
+            /*foreach($this->getDiffProductIdentifiers($batchSize) as $identifiers => $arrayIds){
```
docteurklein, line 99: either keep it or remove it, but don't keep comments.
Also (if uncommented), this would execute the traversal twice: once for displaying, one for actual indexing.
Better do the display in the same loop

src/Akeneo/Pim/Enrichment/Bundle/Command/IndexProductCommand.php
```diff
@@ -199,4 +215,70 @@ private function checkIndexExists(): void
...
+
+            $formerId = end($rows)['uuid'];
+
+            $existingMysqlIdentifiers = array_column($rows, '_id');
+            $existingMysqlUpdated = array_column($rows, 'updated');
+
+            $results = $this->productAndProductModelClient->search([
+                'query' => [
+                    'bool' => [
+                        'must' => [
```
docteurklein, line 254: not sure why we're having both "bool must" and filter. Is that something we thought about and decided it was good?

docteurklein (approved):
GTM once we comments are resolved

src/Akeneo/Pim/Enrichment/Bundle/Command/IndexProductCommand.php
```diff
@@ -199,4 +208,70 @@ private function checkIndexExists(): void
...
+            }, $results["hits"]["hits"]);
+
+            $diff = array_reduce($rows,function($carry,$item) use($esIdentifiers){
+                if(!in_array($item['_id'],$esIdentifiers)){
+                    $carry[] = $item['identifier'];
+                }
+                return $carry;
+            },[]);
+
+            yield $diff;
```
nmarniesse, line 274: Idea of improvement: here you can yield a few of identifiers, or maybe 0 identifier. The indexer process is more efficient if we give it hundreds of identifiers at once. 
The idea behind that is to stack the identifiers in an array and each time the array size is more than X elements, we yield it and reset it.

nmarniesse (approved):

src/Akeneo/Pim/Enrichment/Bundle/Command/IndexProductCommand.php
docteurklein: that's a great idea if the differences are spread out between pages. 
We could make `$diff` a `$this->diff` and wrap the `yield` like:

```
            $this->diff = array_reduce($rows,function($carry,$item) use($esIdentifiers){
                if(!in_array($item['_id'],$esIdentifiers)){
                    $carry[] = $item['identifier'];
                }
                return $carry;
            }, $this->diff);

            if (count($this->diff) > 100) {
                yield $this->diff;
                $this->diff = [];
            }
```

not sure, to be tested;

docteurklein: we can also ship a first version and optimize later

ahocquard (approved):

src/Akeneo/Pim/Enrichment/Bundle/Command/IndexProductCommand.php
```diff
@@ -67,6 +67,12 @@ protected function configure()
                 InputOption::VALUE_NONE,
                 'Index all existing products into Elasticsearch'
             )
+            ->addOption(
+                'diff',
+                'd',
+                InputOption::VALUE_NONE,
+                'Resolve differences between MySQL and Elasticsearch'
```
ahocquard, line 74: In another PR:

```suggestion
                'Index both missing products present in Mysql and not in ES and outdated product document in ES'
```

tseho:
What's going on with this PR ? :thinking:

niarcotyk:
> What's going on with this PR ? thinking

Like every time I rebase. I try to fix the problem

PR #17158, opened 2022-06-03
CXP-1186: catalogs/{id}/product-uuids

tseho:

Screenshot_2022-06-03_13-00-28

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/components/catalogs/back/src/Application/Handler/GetProductUuidsHandler.php
+++ b/components/catalogs/back/src/Application/Handler/GetProductUuidsHandler.php
@@ -27,8 +27,8 @@ final class GetProductUuidsHandler
             $query->getCatalogId(),
             $query->getSearchAfter(),
             $query->getLimit(),
-            $query->getUpdatedBefore(),
             $query->getUpdatedAfter(),
+            $query->getUpdatedBefore(),
         );
     }
 }
--- a/components/catalogs/back/src/Application/Handler/GetProductUuidsHandler.php
+++ b/components/catalogs/back/src/Application/Handler/GetProductUuidsHandler.php
@@ -27,6 +27,8 @@ final class GetProductUuidsHandler
             $query->getCatalogId(),
             $query->getSearchAfter(),
             $query->getLimit(),
+            $query->getUpdatedBefore(),
+            $query->getUpdatedAfter(),
         );
     }
 }
--- a/components/catalogs/back/src/Application/Handler/GetProductUuidsHandler.php
+++ b/components/catalogs/back/src/Application/Handler/GetProductUuidsHandler.php
@@ -1,0 +1,32 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Akeneo\Catalogs\Application\Handler;
+
+use Akeneo\Catalogs\Application\Persistence\GetProductUuidsQueryInterface;
+use Akeneo\Catalogs\ServiceAPI\Query\GetProductUuidsQuery;
+
+/**
+ * @copyright 2022 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+final class GetProductUuidsHandler
+{
+    public function __construct(
+        private GetProductUuidsQueryInterface $query,
+    ) {
+    }
+
+    /**
+     * @return array<string>
+     */
+    public function __invoke(GetProductUuidsQuery $query): array
+    {
+        return $this->query->execute(
+            $query->getCatalogId(),
+            $query->getSearchAfter(),
+            $query->getLimit(),
+        );
+    }
+}
--- a/components/catalogs/back/src/Infrastructure/Controller/Public/GetProductUuidsAction.php
+++ b/components/catalogs/back/src/Infrastructure/Controller/Public/GetProductUuidsAction.php
@@ -1,0 +1,146 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Akeneo\Catalogs\Infrastructure\Controller\Public;
+
+use Akeneo\Catalogs\Infrastructure\Security\DenyAccessUnlessGrantedTrait;
+use Akeneo\Catalogs\Infrastructure\Security\GetCurrentUserIdTrait;
+use Akeneo\Catalogs\ServiceAPI\Messenger\QueryBus;
+use Akeneo\Catalogs\ServiceAPI\Model\Catalog;
+use Akeneo\Catalogs\ServiceAPI\Query\GetCatalogQuery;
+use Akeneo\Catalogs\ServiceAPI\Query\GetProductUuidsQuery;
+use Akeneo\Platform\Bundle\FrameworkBundle\Security\SecurityFacadeInterface;
+use Symfony\Component\HttpFoundation\JsonResponse;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\HttpKernel\Exception\BadRequestHttpException;
+use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
+use Symfony\Component\Messenger\Exception\ValidationFailedException;
+use Symfony\Component\Routing\RouterInterface;
+use Symfony\Component\Security\Core\Authentication\Token\Storage\TokenStorageInterface;
+
+/**
+ * @copyright 2022 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+class GetProductUuidsAction
+{
+    use GetCurrentUserIdTrait;
+    use DenyAccessUnlessGrantedTrait;
+
+    public function __construct(
+        private QueryBus $queryBus,
+        private TokenStorageInterface $tokenStorage,
+        private SecurityFacadeInterface $security,
+        private RouterInterface $router,
+    ) {
+    }
+
+    public function __invoke(Request $request, string $id): Response
+    {
+        $this->denyAccessUnlessGrantedToListCatalogs();
+        $this->denyAccessUnlessGrantedToListProducts();
+
+        $catalog = $this->getCatalog($id);
+
+        $this->denyAccessUnlessOwnerOfCatalog($catalog, $this->getCurrentUserId());
+
+        [$searchAfter, $limit] = $this->getParameters($request);
+        $uuids = $this->getProductUuids($catalog, $searchAfter, $limit);
+
+        return new JsonResponse($this->paginate($catalog, $uuids, $searchAfter, $limit), Response::HTTP_OK);
+    }
+
+    private function getCatalog(string $id): Catalog
+    {
+        try {
+            $catalog = $this->queryBus->execute(new GetCatalogQuery($id));
+        } catch (ValidationFailedException $e) {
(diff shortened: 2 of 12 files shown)


components/catalogs/back/src/Infrastructure/Controller/Public/GetProductUuidsAction.php
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    public function __invoke(Request $request, string $id): Response
+    {
+        $this->denyAccessUnlessGrantedToListCatalogs();
+        $this->denyAccessUnlessGrantedToListProducts();
+
+        $catalog = $this->getCatalog($id);
+
+        $this->denyAccessUnlessOwnerOfCatalog($catalog, $this->getCurrentUserId());
```
masacc, line 47: I think this function never throw because in previous line, the query will already throw an exception `'Catalog "%s" does not exist or you can\'t access it.'`. You want to be ceinture bretelles ?

components/catalogs/back/src/Infrastructure/Controller/Public/GetProductUuidsAction.php
```diff
@@ -0,0 +1,146 @@
...
+                        'limit' => $limit,
+                    ]),
+                ],
+            ],
+            '_embedded' => [
+                'items' => $uuids,
+            ],
+        ];
+
+        if (\count($uuids) >= $limit) {
```
masacc, line 134: I'm not sure about this condition : 
- I don't know how `count($uuids)` can be greater than `$limit`
- I think if `count($uuids) === $limit`, we don't know if there is a next page

components/catalogs/back/src/Infrastructure/TemporaryEnrichmentBridge/SearchAfterSizeUuidResultCursorFactory.php
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     * @param mixed $queryBuilder
+     * @param array<array-key, mixed> $options
+     */
+    public function createCursor($queryBuilder, array $options = [])
+    {
+        /** @var array{_source: array<string>, sort?: array<string>, size: int} $queryBuilder */
```
masacc, line 40: why do you place this line here and not in the method phpdoc or just before first call of `$queryBuilder`?

components/catalogs/back/src/Infrastructure/Controller/Public/GetProductUuidsAction.php
tseho: $this->getCatalog() returns `Catalog "%s" does not exist or you can\'t access it.` when it can't find it.
The message is generic for obfuscation purposes, but neither que the command query/query bus does check the ownership.

The ownership of the catalog is only done inside `denyAccessUnlessOwnerOfCatalog`

components/catalogs/back/src/Infrastructure/Controller/Public/GetProductUuidsAction.php
tseho: It's ISO with the default PIM paginator.

To improve this, we need to add an additional query to count to total of entries.

components/catalogs/back/src/Infrastructure/TemporaryEnrichmentBridge/SearchAfterSizeUuidResultCursorFactory.php
tseho: If my memory serve me well, I tried but phpstan was complaining that is was not compatible with the phpdoc declared in the parent class, this is a workaround.
Maybe I'm wrong but I want to trash this class anyway, it should be handled by the futur enrichment ServiceAPI that was promised.

masacc (approved):

components/catalogs/back/src/Infrastructure/Persistence/GetProductUuidsQuery.php
```diff
@@ -103,6 +112,57 @@ private function findProductSelectionCriteria(string $catalogId): array
...
+        $updatedAfterDateTime = \DateTimeImmutable::createFromFormat(\DateTimeInterface::ATOM, (string) $updatedAfter);
+
+        if (false !== $updatedBeforeDateTime) {
+            $updatedBeforeDateTime = $updatedBeforeDateTime->setTimezone(new \DateTimeZone('UTC'));
+        }
+        if (false !== $updatedAfterDateTime) {
+            $updatedAfterDateTime = $updatedAfterDateTime->setTimezone(new \DateTimeZone('UTC'));
+        }
+
+        if (null !== $updatedBefore && null !== $updatedAfter) {
```
tseho, line 133: If I were you, I would have moved all this logic to another function, something like `[$operator, $value] = $this->parseUpdatedParameters($updatedBefore, $updateAfter);` 
with early returns at each step.
I think the reading would be easier.

(embedding multiple `if` is a red flag)

components/catalogs/back/src/Application/Handler/GetProductUuidsHandler.php
```diff
@@ -0,0 +1,34 @@
...
+     * @return array<string>
+     */
+    public function __invoke(GetProductUuidsQuery $query): array
+    {
+        return $this->query->execute(
+            $query->getCatalogId(),
+            $query->getSearchAfter(),
+            $query->getLimit(),
+            $query->getUpdatedBefore(),
+            $query->getUpdatedAfter(),
```
tseho, line 31: It's funny, why `before` after `after` ? It's like having `max` then `min`. I'm more used to see the other way around.
When you want products between an interval, you have to do `$this->query->execute('...', null, 100, '2022-09-01T17:45:00+02:00', '2022-09-01T17:35:00+02:00');` where the second parameter has to be lower than the first.

Not blocking, just confusing :)

tseho (commented):
I can't approve my own PR but it's a GTM from me.

components/catalogs/back/src/Infrastructure/Controller/Public/GetProductUuidsAction.php
```diff
@@ -0,0 +1,167 @@
...
+
+        return [$searchAfter, $limit, $updatedBefore, $updatedAfter];
+    }
+
+    /**
+     * @return array<string>
+     */
+    private function getProductUuids(Catalog $catalog, ?string $searchAfter, int $limit, ?string $updatedBefore, ?string $updatedAfter): array
+    {
+        if (!$catalog->isEnabled()) {
```
willy-ahva, line 101: Shouldn't it be in the handler? IMO it's part of the business that if the catalog is disabled it does not return products.

components/catalogs/back/src/Infrastructure/Controller/Public/GetProductUuidsAction.php
```diff
@@ -0,0 +1,167 @@
...
+            throw new ViolationHttpException($e->getViolations());
+        }
+    }
+
+    /**
+     * @param array<string> $uuids
+     *
+     * @return array{_links: array{self: array{href: string}, first: array{href: string}, next?: array{href: string}}, _embedded: array{items: string[]}}
+     */
+    private function paginate(Catalog $catalog, array $uuids, ?string $searchAfter, int $limit, ?string $updatedBefore, ?string $updatedAfter): array
```
willy-ahva: It could have been extracted in a dedicated service. But ok

willy-ahva: And maybe it could be renamed formatToHAL

masacc (approved):

components/catalogs/back/src/Infrastructure/Persistence/GetProductUuidsQuery.php
```diff
@@ -103,6 +112,57 @@ private function findProductSelectionCriteria(string $catalogId): array
         return $criteria;
     }
 
+    /**
+     * @return array<mixed>
+     */
+    private function getUpdatedFilters(?string $updatedBefore, ?string $updatedAfter): array
```
willy-ahva: Agree with Quentin + I wonder if this part should be extracted. Here you are translating string into a filter format. I don't think it should in the same class as the query. The query should receive it already translated. It will be easier to unit test it with a dedicated class.

components/catalogs/back/src/Infrastructure/Controller/Public/GetProductUuidsAction.php
masacc: I think we already have this discussion in another PR, don't we? Although I agree, I could not remember what was the reason to do it here...

tseho: We did the same in `GetProductIdentifiersAction`.

components/catalogs/back/src/Infrastructure/Controller/Public/GetProductUuidsAction.php
tseho: What's worst, it's that this service already exists in the API bundle.
But we can't use without changing it a lot, it has been designed for a specific usecase with generic urls and so on, and does not match what we need.
That's why I resolved myself to embed the expected behavior here.

components/catalogs/back/src/Infrastructure/Persistence/GetProductUuidsQuery.php
masacc: Where would you do the transformation? In the handler? I think we can transmit DateTime objects to the query and the query only takes care of transforming them to pqb format. WDYT?

tseho: Given the signature of the query, for me, it's fine that you convert it in here.
Maybe Willy was talking about injecting a "helper" class in the query whose job is two take 2 datetimes and return one PQB filter.

components/catalogs/back/src/Infrastructure/Controller/Public/GetProductUuidsAction.php
```diff
@@ -0,0 +1,166 @@
...
+
+        return new JsonResponse($this->paginate($catalog, $uuids, $searchAfter, $limit, $updatedAfter, $updatedBefore), Response::HTTP_OK);
+    }
+
+    private function getCatalog(string $id): Catalog
+    {
+        try {
+            $catalog = $this->queryBus->execute(new GetCatalogQuery($id));
+        } catch (ValidationFailedException $e) {
+            throw new NotFoundHttpException(\sprintf('Catalog "%s" does not exist or you can\'t access it.', $id), $e);
```
RiffVentura, line 61: IMO we shouldn't imply that the resource might exist

masacc: why ?

RiffVentura (approved):

components/catalogs/back/src/Infrastructure/Controller/Public/GetProductUuidsAction.php
RiffVentura: From security POV it gives away that resource exist (imagine if it were username instead: now I know that this username is valid)

masacc: seen IRL, this comment is enough not precise to be ok on a security point of view

PR #17196, opened 2022-06-10
RAB-761: Add id field on CategoryTree

grena:

Add id field on `CategoryTree`, in order to avoid another fetch to retrieve it.
`id` is mainly used to retrieve children categories.

- [x] Tests


ValentinMumble (approved):

ghoussard (approved):
Thank you for your contibution!

PR #17220, opened 2022-06-15
[POC] Using workload identity instead of gcp key

vgelot:

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc


docteurklein:
`GOOGLE_APPLICATION_CREDENTIALS`
is still used in some places: https://github.com/akeneo/pim-enterprise-dev/search?q=GOOGLE_APPLICATION_CREDENTIALS

This could mess with the app because the SDK client will implicitely use it if available.

vgelot:
Yes, of course, that why it is also removed in the corresponding ee branch :)

PR #17230, opened 2022-06-16
CXP-1227: Add default criteria when a catalog is created

tseho:

I've also added some annotations to have phpunit coverage to correctly reflect what is tested.

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/components/catalogs/back/src/Application/Handler/CreateCatalogHandler.php
+++ b/components/catalogs/back/src/Application/Handler/CreateCatalogHandler.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Akeneo\Catalogs\Application\Handler;
 
+use Akeneo\Catalogs\Application\Persistence\UpdateCatalogProductSelectionCriteriaQueryInterface;
 use Akeneo\Catalogs\Application\Persistence\UpsertCatalogQueryInterface;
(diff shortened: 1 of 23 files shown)


masacc (approved):

components/catalogs/back/src/Application/Handler/CreateCatalogHandler.php
```diff
@@ -26,5 +28,16 @@ public function __invoke(CreateCatalogCommand $command): void
             $command->getOwnerUsername(),
             false,
         );
+
+        $this->updateCatalogProductSelectionCriteriaQuery->execute(
+            $command->getId(),
+            [
+                [
```
willy-ahva, line 35: Should we have a VO with named constructor for criterion and/or criteria? IMHO yes.

willy-ahva (approved):

RiffVentura (approved):

components/catalogs/back/src/Application/Handler/CreateCatalogHandler.php
masacc: I was thinking about it for validation also...

tseho: I think we can go back to this one after we handled submission and validation, here it's hardcoded, validation is not necessary yet.

PR #17233, opened 2022-06-17
CXP-1233: Internal API endpoint to update list of criterion by catalod id

masacc:

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc


components/catalogs/back/src/Infrastructure/Validation/CriteriaJsonValidator.php
```diff
@@ -0,0 +1,74 @@
...
+ */
+class CriteriaJsonValidator extends ConstraintValidator
+{
+    public function validate(mixed $value, Constraint $constraint): void
+    {
+        if (!$constraint instanceof CriteriaJson) {
+            throw new UnexpectedTypeException($constraint, CriteriaJson::class);
+        }
+
+        $jsonSchemaValidator = new Validator();
```
tseho, line 23: Why do you use the JsonSchemaValidator ? Any particular reason ?

tseho: Why do you need to do this ? This is not a public API, it will only be called from our frontend.
If it's not a public API, and if the front is sufficiently tested to call it correctly, I don't care if it crash when somebody meddled with the payload.
We never validate the structure when it's the internal API, I think it's a waste of time and it will add more stuff to maintain.

tseho: Moreover, if you validate the structure and never validates the VO, what's the point of the VO ?

You do `$this->denormalizer->denormalize($criteria, Criterion::class . '[]', 'internal'),` to instanciate the VOs but they are never validated, and are directly normalized behind.

Don't get me wrong, I'm aware you wanted to be "ceinture-bretelles' but the feature is not out yet, we don't know how far we will go, it's better to keep things as simple as possible.

components/catalogs/back/src/Infrastructure/Validation/CriteriaJsonValidator.php
```diff
@@ -0,0 +1,74 @@
...
+                ],
+                'operator' => [
+                    'type' => 'string',
+                    'enum' => ['=', '!='],
+                ],
+                'value' => [
+                    'type' => 'boolean',
+                ],
+            ],
+            'required' => ['field', 'operator', 'value'],
```
RiffVentura, line 71: value can be optional

components/catalogs/back/tests/Integration/Infrastructure/Controller/Internal/SaveCriteriaActionTest.php
```diff
@@ -0,0 +1,154 @@
...
+
+/**
+ * @copyright 2022 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ *
+ * @covers \Akeneo\Catalogs\Infrastructure\Controller\Internal\GetCatalogAction
+ */
+class SaveCriteriaActionTest extends IntegrationTestCase
+{
+    private ?KernelBrowser $client;
```
RiffVentura, line 24: Not really needed here, since client is instantiated every time

RiffVentura (approved):

components/catalogs/back/src/Infrastructure/Validation/CriteriaJsonValidator.php
masacc: > You do $this->denormalizer->denormalize($criteria, Criterion::class . '[]', 'internal'), to instanciate the VOs but they are never validated, and are directly normalized behind.

Yes because denormalization can fail before validating them, that's why I chose to validate the JSON body before denormalization, and then no need to re-validate the VO.

> Moreover, if you validate the structure and never validate the VO, what's the point of the VO

The goal is to pass objects instead of arrays from class to class.

> This is not a public API, it will only be called from our frontend. If it's not a public API, and if the front is sufficiently tested to call it correctly, I don't care if it crash when somebody meddled with the payload.

OK, I rather want to be strict on the backend side but if you think it's sufficient on the frontend side, I will remove it. Indeed it will be simpler to maintain.

So to sum up: we save directly the body in the `product_selection_criteria` field without any normalization/denormalization process and without validation. Is it correct?

tseho: > The goal is to pass objects instead of arrays from class to class.
It was extremely valuable when we had no static analysis tool.
With phpstan & psalm, if you describre your array structure AND you validate it on entry, there is no reason to be afraid, it's really hard to f**k it up.

> So to sum up: we save directly the body in the product_selection_criteria field without any normalization/denormalization process and without validation. Is it correct?

I would have done it like this to keep things simple:

```
$payload = \json_decode((string) $request->getContent(), true, 512, JSON_THROW_ON_ERROR);

$violations = $this->validator->validate($payload, [
    new Criteria(), // Instead of using JSONSchema validator, you can combine Assert/All & Assert/Collection to validate the array structure
]);

if ($violations->count() > 0) {
    return new JsonResponse(
        [
            'errors' => $this->normalizer->normalize($violations), // If ConstraintViolationListInterface is not normalized correctly, we can our own normalizer and specify the format
            'message' => 'Criteria are not valid.',
        ],
        Response::HTTP_UNPROCESSABLE_ENTITY
    );
}

$this->updateCatalogProductSelectionCriteriaQuery->execute(
    $catalogId,
    $payload,
);

return new JsonResponse(null, 204);
        ```

tseho (approved):

PR #17245, opened 2022-06-20
CXP-1235: GET /catalogs/id/product-identifiers

tseho:

You can checkout the code commit by commit, they are independent.
- The correct value for the status criterion is "enabled", not "status".
- I've cleanup some left-over in controller tests
- I've added the new controller for product identifiers

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/components/catalogs/back/src/Infrastructure/Controller/Public/GetProductIdentifiersAction.php
+++ b/components/catalogs/back/src/Infrastructure/Controller/Public/GetProductIdentifiersAction.php
@@ -1,0 +1,146 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Akeneo\Catalogs\Infrastructure\Controller\Public;
+
+use Akeneo\Catalogs\Infrastructure\Security\DenyAccessUnlessGrantedTrait;
+use Akeneo\Catalogs\Infrastructure\Security\GetCurrentUsernameTrait;
+use Akeneo\Catalogs\ServiceAPI\Messenger\QueryBus;
+use Akeneo\Catalogs\ServiceAPI\Model\Catalog;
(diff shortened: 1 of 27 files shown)


willy-ahva (approved):

components/catalogs/back/src/Infrastructure/Controller/Public/GetProductIdentifiersAction.php
```diff
@@ -0,0 +1,146 @@
...
+            throw new BadRequestHttpException();
+        }
+
+        return [$searchAfter, $limit];
+    }
+
+    /**
+     * @return array<string>
+     */
+    private function getProductUuids(Catalog $catalog, ?string $searchAfter, int $limit): array
```
masacc, line 88:
```suggestion
    private function getProductIdentifiers(Catalog $catalog, ?string $searchAfter, int $limit): array
```

components/catalogs/back/src/Infrastructure/Controller/Public/GetProductIdentifiersAction.php
```diff
@@ -0,0 +1,146 @@
...
+        }
+    }
+
+    /**
+     * @param array<string> $uuids
+     * @return array<array-key, mixed>
+     */
+    private function paginate(Catalog $catalog, array $uuids, ?string $searchAfter, int $limit): array
+    {
+        $last = \end($uuids);
```
masacc, line 111: `$uuids` seems to rather be idenfitiers so I don't think you can you use it like this for search after. Maybe a call to a query like this:
```
$last = $this->findProductIdentifierQuery->execute(\end($uuids));
```

guillaumefer-akeneo (approved):

PR #17249, opened 2022-06-20
Pullup 6.0 -> master

mmetayer:

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc


jmleroux (approved):

components/catalogs/back/psalm.xml
```diff
@@ -7,6 +7,7 @@
         <directory name="src" />
         <ignoreFiles>
             <directory name="../../../vendor" />
+            <directory name="../../../../../../vendor" />
```
mmetayer, line 10: needed when the psalm analysis is run in EE; there's most probably a better fix, but I couldn't find one
cc @tseho

mmetayer: Edit: didn't work :/

jmleroux: IMHO, there is no need to run linter on CE for EE CI.

nmarniesse (approved):

PR #17321, opened 2022-06-27
Update Composer dependencies for 5.0

jmleroux:
Should fix a few dependabot alerts.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -104,7 +104,6 @@
         "symfony/messenger": "^4.4",
         "symfony/monolog-bundle": "^3.4",
         "symfony/options-resolver": "4.4.25",
-        "symfony/orm-pack": "^1.0",
         "symfony/polyfill-apcu": "1.12.0",
         "symfony/process": "4.4.25",
         "symfony/property-access": "4.4.25",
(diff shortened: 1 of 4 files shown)


composer.json
```diff
@@ -104,7 +104,6 @@
         "symfony/messenger": "^4.4",
         "symfony/monolog-bundle": "^3.4",
         "symfony/options-resolver": "4.4.25",
-        "symfony/orm-pack": "^1.0",
```
jmleroux, line 107: Automatically removed by `composer update`

tests/legacy/features/Context/DataGridContext.php
```diff
@@ -508,7 +508,7 @@ public function theRowsShouldBeSortedBy($naturally, $order, $columnName)
     public function iClickOnTheActionOfTheRowWhichContains($actionName, $element)
     {
         //Wait for the JS action to be linked to the button because it will do the default action of the row otherwise
-        $this->getSession()->wait(6000, false);
```
jmleroux, line 511: This was a big, the second parameter must be a string.
And the updated Selenium2Driver is not compatible anymore with a bool: https://github.com/minkphp/MinkSelenium2Driver/blob/v1.6.0/src/Selenium2Driver.php#L981 . If using a `false` (bool), it results in an invalid JS condition

Because it is 'false' by default, I just removed it.

niarcotyk (approved):

ValentinMumble (approved):

nmarniesse (approved):

StevenVAIDIE (approved):

PR #17335, opened 2022-06-28
Cxp 1242 front

tseho:

:warning: To check the validation, I've added a dumb constraint in the back that always reject.

https://user-images.githubusercontent.com/1421130/176221556-cd662ddc-c5b6-40ee-a712-8bb8e545c1f9.mp4

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc


components/catalogs/front/src/components/CatalogEdit/hooks/useSaveCatalog.ts
```diff
@@ -0,0 +1,50 @@
...
+                    throw new Error(`${response.status} ${response.statusText}`);
+            }
+        }
+
+        return [true, null] as Result;
+    }, []);
+
+    const mutation = useMutation(request, {
+        onSuccess: () => {
+            queryClient.invalidateQueries('catalog');
```
RiffVentura, line 45: Should specify catalog id

components/catalogs/front/src/components/ProductSelection/criteria/useCriteriaRegistry.ts
```diff
@@ -0,0 +1,39 @@
...
+    label: string;
+    factory: () => AnyCriterionState;
+};
+
+type Result = {
+    system: Factory[];
+    getCriterionByField: (field: string) => Promise<AnyCriterion>;
+};
+
+export const useCriteriaRegistry = (): Result => {
```
masacc, line 16: why this file is not placed in a `hooks` directory?

components/catalogs/front/src/components/CatalogEdit/hooks/useSaveCatalog.ts
tseho: As discussed IRL, we will keep the non-optimized syntax for clarity.

components/catalogs/front/src/hooks/useCatalog.ts
```diff
@@ -15,13 +15,5 @@ type Result = {
 };
 
 export const useCatalog = (catalogId: string): Result => {
```
masacc, line 17: There is another hook named the same (`components/catalogs/front/src/components/CatalogEdit/hooks/useCatalog.ts`) and that seems to do the same. Why?

masacc (approved):

components/catalogs/front/src/components/ProductSelection/criteria/useCriteriaRegistry.ts
tseho: I hesitated, I preferred to keep it there because it's the registry of Criteria.
It will be easier for new comer to find out how to register a new one.

Screenshot_2022-07-04_10-59-22

Is that ok ?

components/catalogs/front/src/hooks/useCatalog.ts
tseho: There is a subtle difference between those two.

 `components/catalogs/front/src/components/CatalogEdit/hooks/useCatalog.ts`:
 ```
 type Data = {
    id: string;
    name: string;
    enabled: boolean;
    product_selection_criteria: AnyCriterionState[];
};
export const useCatalog = (catalogId: string): Result => {
    return useCatalogQuery(catalogId);
};
 ```
 
 vs 
 
 `components/catalogs/front/src/hooks/useCatalog.ts`
 ```
 type Data = {
    id: string;
    name: string;
    enabled: boolean;
    owner_username: string;
};
 export const useCatalog = (catalogId: string): Result => {
    return useCatalogQuery(catalogId);
};
```

Both are calling the same query, they share the same cache.
But, they return a different subset of the API response.
The internal one, in CatalogEdit, returns the criteria while the one in the `src/hooks/useCatalog.ts` is publicly exposed and only return id, name, enabled & owner.

Having only one hook `useCatalog` would mean that we expose publicly the list of criteria.

components/catalogs/front/src/components/ProductSelection/criteria/useCriteriaRegistry.ts
masacc: I don't find the rule very clear when or not to place a hook in a `/hooks` directory. But if it's ok for the rest of the team too, go for it.

tseho: Your point of being consistent is valid, I will move it to hooks.

RiffVentura (approved):

PR #17341, opened 2022-06-29
CXP-1225 Redirect to catalog edit page

RiffVentura:
Added a public endpoint that redirects to the catalog edit page for the app

There was a need for fetching connected app by its related user identifier(username)  so `FindOneConnectedAppByUserIdentifierQuery` was added
`FindOneConnectedAppByUserIdQuery` was replaced by the new query since it's used only once


src/Akeneo/Connectivity/Connection/back/Infrastructure/Apps/Controller/Public/ConfigureCatalogAction.php
```diff
@@ -0,0 +1,85 @@
...
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\HttpKernel\Exception\AccessDeniedHttpException;
+use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
+use Symfony\Component\Routing\RouterInterface;
+
+/**
+ * @copyright 2022 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+final class ConfigureCatalogAction
```
masacc, line 22: As it does not really configure a catalog and it only redirects when calling this endpoint, I think we could name it `RedirectToEditCatalogAction`

src/Akeneo/Connectivity/Connection/back/tests/Unit/spec/Infrastructure/Apps/Controller/Public/ConfigureCatalogActionSpec.php
```diff
@@ -0,0 +1,210 @@
...
+    ) {
+        $this->beConstructedWith(
+            $router,
+            $security,
+            $catalogQueryBus,
+            $findOneConnectedAppByUserIdentifierQuery,
+        );
+    }
+
+    public function it_is_an_configure_catalog_action(): void
```
masacc, line 40:
```suggestion
    public function it_is_a_configure_catalog_action(): void
```

masacc (approved):

tseho:
Just to remember, we have to do something about that:

Screenshot_2022-07-06_16-56-41

tseho (approved):

PR #17345, opened 2022-06-29
CPM-674: Add uuid on webhook messages

pierallard:
As identifier will be optional, we need to create webhook messages with uuid instead of identifiers.


src/Akeneo/Pim/Enrichment/Component/Product/Webhook/ProductCreatedAndUpdatedEventDataBuilder.php
```diff
@@ -50,6 +50,7 @@ public function supports(BulkEventInterface $event): bool
 
     public function build(BulkEventInterface $bulkEvent, Context $context): EventDataCollection
     {
+        // TODO Change this once GetConnectorProduct is done by Uuids
```
Balamung, line 53: Can you create a CPM ticket so we can make the change on post-adr ? (The connector products are changed in the endpoints API, so it should be merged there soon)

Balamung (approved):

PR #17368, opened 2022-07-01
CXP-1241: Be able to add a criterion family

willy-ahva:

https://user-images.githubusercontent.com/1421130/177799254-ff4d02fc-a894-484e-b403-ea7deb202366.mp4

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc


components/catalogs/front/src/components/ProductSelection/hooks/useFamiliesByCodes.ts
```diff
@@ -0,0 +1,65 @@
...
+
+type Error = string | null;
+type Result = {
+    isLoading: boolean;
+    isError: boolean;
+    data: Family[] | undefined;
+    error: Error;
+};
+
+const LIMIT = 20;
```
RiffVentura, line 17: Seems a bit low. Is there a reason for it?

RiffVentura: Isn't it better then to not reuse the `useInfiniteFamilies` hook ? (We don't need pagination here, nor search)
I feel like it should have a dedicated hook and endpoint. (since use cases are different)

tseho: It started as a separate hook with dedicated API calls.
But sadly, we have to prepared to have worst case scenario.

If you have a customer with thousands of families, and he selected 300 of them in this field.
Do you fetch the 300 in one query ? If that's the case, you have to put no limit on the endpoint...

Each time the PIM had an endpoint without pagination AND without limits, it was working with dev environments with low amount of data but it always come back to bite us in the *** when it's not working anymore for some and it became a SLA.

With the current solution, 99% of users will pick less than 20 families and will have only API call.
For the 1% that will do the unexpected and pick hundreds of families, it will work over multiple API calls.

RiffVentura: Fair point :+1:

RiffVentura (approved):

components/catalogs/front/src/components/ProductSelection/hooks/useFamiliesByCodes.ts
tseho: And regarding the question on why did I reuse the other hook.
I would have to reproduce a similar behavior in the current one => fetch 20, has more ? fetch 20 more. has more ? etc ...

The hook was not finished and was already 200+ lines without handling everything, I was reinventing the wheel so I decided to reuse the infinite one.

components/catalogs/front/src/components/ProductSelection/criteria/FamilyCriterion/FamilyCriterion.tsx
```diff
@@ -0,0 +1,57 @@
...
+    const [showFamilies, setShowFamilies] = useState<boolean>(false);
+
+    useEffect(() => {
+        setShowFamilies([Operator.IN_LIST, Operator.NOT_IN_LIST].includes(state.operator));
+    }, [state.operator]);
+
+    return (
+        <List.Row>
+            <List.TitleCell width={150}>
+                {translate('akeneo_catalogs.product_selection.criteria.status.label')}
```
RiffVentura, line 36:
```suggestion
                {translate('akeneo_catalogs.product_selection.criteria.family.label')}
```

PR #17369, opened 2022-07-01
OCT-80: dependency cruiser

tseho:

`depcruise` does static analysis of dependencies & support custom rules.
It comes with default rules like
- `no-non-package-json` => you cannot uses dependencies not declared in package.json
- `not-to-deprecated` => warning when using a deprecated dependency

You can also write custom rules.
For example, the following one forbid to use internal files from other major components, it can only use modules exported in the `index.ts` of others.
```
        {
            name: 'no-to-other-components-internals',
            severity: 'error',
            from: {
                path: '^src/components/([^/]+)/.+',
            },
            to: {
                path: '^src/components/([^/]+)/(?!index).+',
                pathNot: '^src/components/$1',
            },
        },
```

Screenshot_2022-07-01_16-03-28

:warning: This PR, as it is, still have 50 dependency errors to be fixed before having a green CI, it's not fixed yet to not pollute the discussion around it.

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/components/catalogs/Makefile
+++ b/components/catalogs/Makefile
@@ -119,6 +119,7 @@ catalogs-lint-front: yarn-policies
 	$(CATALOGS_YARN) workspace @akeneo-pim-community/catalogs prettier --check
 	$(CATALOGS_YARN) workspace @akeneo-pim-community/catalogs eslint
 	$(CATALOGS_YARN) workspace @akeneo-pim-community/catalogs tsc --noEmit --strict
+	$(CATALOGS_YARN) workspace @akeneo-pim-community/catalogs depcruise
 
 .PHONY: catalogs-unit-front
 catalogs-unit-front: yarn-policies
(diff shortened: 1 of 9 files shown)


masacc (approved):
I see what this new tool brings to us but I'm starting to wonder if we cannot rationalize all the tools that we are using for static/unit/integration/... tests in front and back because we have a lot of them. When onboarding someone new it's a lot to learn and understand.

RiffVentura (approved):

tseho:
> rationalize all the tools

Do you think we should remove some of them ? Which one do you have in mind ?
Actually, IMO, the one which bring the less value in my opinion is phpinsights.
For all the other ones, I understand your point about the challenges when onboarding newcomers.
But at the same time, automating what can be statically analyzed save us time and reduce confrontation during reviews.

For' some developers, they will feel more confident in their PR when asking for a review if they successfully passed all the tests.
On top of having static analysis, it will really guide them to have rules like coverage.
=> Regarding coverage, I'm not talking about 100% dumb coverage, I'm talking about smarter rules like we started in Connectivity, eg: Each query should have a least one integration test.

PR #17414, opened 2022-07-08
CPM-680: remove usage of identifiers in DQI and ProductIndexer

pierallard:
This PR does several things:
- Remove usage of Identifiers in DQI Score computing
- This removal implies the temporary query we did for DQI (id to uuid) is not used anymore so we removed it
- Remove usage of identifiers in the ElasticsearchProductProjection
- Insert an equivalent temporary query (id to uuid, "SqlFindProductUuids")
- Adds the 2 methods "indexFromProductUuid" and "indexFromProductUuids" in the ProductIndexer
- Set @deprecated for the 2 old ones.

Next step is to check where the 2 old ones @deprecated are used, and use the new ones instead, and use the temporary query if needed, at an upper level.


cemorin (approved):

Balamung:
Did you remove GetProductUuidsFromProductIdentifiersQuery because it was no longer used ? I'm looking at maybe using it in my PR to convert identifiers to UUIDs in the ProductController for example (and other places where we currently use identifiers, and I thought it'd be cleaner than injecting the doctrine_connection everywhere to add the method :)

pierallard:
> Did you remove GetProductUuidsFromProductIdentifiersQuery because it was no longer used ? I'm looking at maybe using it in my PR to convert identifiers to UUIDs in the ProductController for example (and other places where we currently use identifiers, and I thought it'd be cleaner than injecting the doctrine_connection everywhere to add the method :)

Yes, but this is the one from DQI. You should not use it elsewhere !

Balamung (approved):

src/Akeneo/Pim/Automation/DataQualityInsights/back/Infrastructure/Elasticsearch/GetDataQualityInsightsPropertiesForProductProjection.php
```diff
@@ -20,31 +20,28 @@ final class GetDataQualityInsightsPropertiesForProductProjection implements GetA
...
+        Assert::isInstanceOf($productUuidCollection, ProductUuidCollection::class);
+        $productScores = $this->getProductScoresQuery->byProductUuidCollection($productUuidCollection);
+        $productKeyIndicators = $this->getProductsKeyIndicators->compute($productUuidCollection);
 
         $additionalProperties = [];
-        foreach ($productIdentifierIds as $productIdentifier => $productId) {
-            $index = (string)$productId;
-            $additionalProperties[$productIdentifier] = [
+        foreach ($productUuidCollection as $productUuid) {
+            $index = (string) $productUuid;
```
mmetayer, line 43:
```suggestion
            $index = $productUuid->toString();
```

mmetayer (approved):
Nice and clean

PR #17443, opened 2022-07-12
PIM-10528: Fix escaped special characters in page titles

mmetayer:
Page titles are set through the [pim/page-title](https://github.com/akeneo/pim-community-dev/blob/master/src/Akeneo/Platform/Bundle/UIBundle/Resources/public/js/page-title.js) component in javascript, we don't need to escape special characters here (as opposite as if they were included in a HTML template)
Also, in some pages, labels were escaped twice (eg `&` would become `&amp;amp;`)

before:
before

after:
after

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/src/Akeneo/Platform/Bundle/UIBundle/Resources/public/js/controller/mass-edit.js
+++ b/src/Akeneo/Platform/Bundle/UIBundle/Resources/public/js/controller/mass-edit.js
@@ -6,10 +6,9 @@ define([
   'oro/translator',
   'pim/controller/front',
   'pim/form-builder',
-  'pim/page-title',
   'routing',
   'pim/analytics',
-], function ($, _, __, BaseController, FormBuilder, PageTitle, Routing, analytics) {
+], function ($, _, __, BaseController, FormBuilder, Routing, analytics) {
   const ACTION_PRODUCT_GRID = 'product-edit';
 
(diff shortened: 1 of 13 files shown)


src/Akeneo/Platform/Bundle/UIBundle/Resources/public/js/controller/mass-edit.js
```diff
@@ -6,10 +6,9 @@ define([
   'oro/translator',
   'pim/controller/front',
   'pim/form-builder',
-  'pim/page-title',
```
mmetayer, line 9: not used in this component

src/Akeneo/Platform/Bundle/UIBundle/Resources/public/js/controller/product/index.js
```diff
@@ -5,7 +5,6 @@ define([
   'pim/form-builder',
   'pim/user-context',
   'oro/mediator',
-  'pim/page-title',
```
mmetayer, line 8: not used

FSAkeneo (approved):

ghoussard (approved):

PR #17464, opened 2022-07-15
RAB: Fix remote storage bug of break session

ValentinMumble:

In this PR: 
- We fixed some behat tests than does not pass only in CE (non SAAS) =>
  - Quick export is local or none depending on version in order to make the tests/legacy/features/pim/enrichment/product/mass-action/quick-export/quick_export_with_media.feature:29 pass
  - We now only exclude local file on local storage instead of do not export the whole files on locale storage because media are not in locale => it break the tests
- We made icecat fixture by default to none storage in order to not have the import now

- [x] Tests
- [x] Migration & Installer

### DIFF
--- a/src/Akeneo/Platform/Bundle/ImportExportBundle/Infrastructure/Step/UploadStep.php
+++ b/src/Akeneo/Platform/Bundle/ImportExportBundle/Infrastructure/Step/UploadStep.php
@@ -52,7 +52,7 @@ final class UploadStep extends AbstractStep
             throw new \LogicException('malformed job parameters, missing storage configuration');
         }
 
-        if ('local' === $jobParameters[self::STORAGE_KEY]['type'] || 'none' === $jobParameters[self::STORAGE_KEY]['type']) {
+        if ('none' === $jobParameters[self::STORAGE_KEY]['type']) {
             return;
         }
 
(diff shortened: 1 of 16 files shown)


StevenVAIDIE (approved):

src/Akeneo/Platform/Bundle/ImportExportBundle/Infrastructure/Step/UploadStep.php
```diff
@@ -52,7 +52,7 @@ protected function doExecute(StepExecution $stepExecution)
             throw new \LogicException('malformed job parameters, missing storage configuration');
         }
 
-        if ('local' === $jobParameters[self::STORAGE_KEY]['type'] || 'none' === $jobParameters[self::STORAGE_KEY]['type']) {
+        if ('none' === $jobParameters[self::STORAGE_KEY]['type']) {
```
ghoussard, line 55: If we don't remove it
```suggestion
        if (NoneStorage::type === $jobParameters[self::STORAGE_KEY]['type']) {
```

ghoussard, line 55: If we don't remove it
```suggestion
        if (NoneStorage::type === $jobParameters[self::STORAGE_KEY]['type']) {
```

ghoussard (approved):

MadeWilson (approved):

PR #17468, opened 2022-07-18
CPM-629: Fix ZddMigration command

pierallard:
This PR fixes the command
- The "notices" are dropped from datadog and we can't see migrations, we replaced by "info"
- The command raised a 500 when the table does not exist, now it raises a warning
- The logged messages were too hard to find, we introduced a new format with "name - action - message" and add more context.

### DIFF
--- a/src/Akeneo/Platform/Bundle/InstallerBundle/Command/MigrateZddCommand.php
+++ b/src/Akeneo/Platform/Bundle/InstallerBundle/Command/MigrateZddCommand.php
@@ -49,7 +49,10 @@ final class MigrateZddCommand extends Command
             $this->logger->warning('%s - skip - Table pim_one_time_task does not exist', [
                 'action' => 'skip',
             ]);
+
+            return Command::SUCCESS;
         }
+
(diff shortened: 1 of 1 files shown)


Balamung (approved):

cemorin (approved):

src/Akeneo/Platform/Bundle/InstallerBundle/Command/MigrateZddCommand.php
```diff
@@ -45,35 +45,46 @@ protected function configure()
 
     protected function execute(InputInterface $input, OutputInterface $output): int
     {
-        $this->logger->notice('pim:zdd-migration:migrate start');
+        if (!$this->tableExists('pim_one_time_task')) {
+            $this->logger->warning('%s - skip - Table pim_one_time_task does not exist', [
+                'action' => 'skip',
+            ]);
```
nmarniesse, line 51: If `pim_one_time_task` does not exist, we don't exit?

nmarniesse (approved):

PR #17486, opened 2022-07-19
TIP-1597: do not display jobs of features that are not activated yet

ahocquard:

## Issue
If a feature is not enabled yet, it should not be visible in the process tracker. For example, with GE with feature flags, there is a bug for the rule job:  it is visible because the rule cronjob is executed (same infra as serenity). 

Also, confirmed with PM: if a feature is enabled due to an upgrade, you should see the new jobs executed **since** the upgrade, but not the old one when the feature was deactivated.

## Potential solution

We could either:
1. prevent the execution of the job in the command itself (execute rule for example),  which impact every commands
2. or make it invisible when executing the job

As a reminder, executing a job of a deactivated feature is not necessary useless. It can help to prepare data in case of an upgrade for example, etc (deactivated feature != dead code never run). For the rules, for example, it just run the job on 0 rules, so no business impact.

## Choice and reason
 
I chose the second solution, as cronjob will be transformed as jobs soon, and I didn't know the exact shape of the future commands to execute cronjobs.
Maybe the provided solution is not the best solution, but I know it's the one that does not impact teammates. We could change it later on after "cronjob as job" transformation done by @jmleroux .

Note: I completely rewrote the test to be agnostic of the jobs of the PIM and remove the use of the SQL catalog.

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/src/Akeneo/Tool/Bundle/BatchBundle/DependencyInjection/AkeneoBatchExtension.php
+++ b/src/Akeneo/Tool/Bundle/BatchBundle/DependencyInjection/AkeneoBatchExtension.php
@@ -40,5 +40,9 @@ class AkeneoBatchExtension extends Extension
                 ->getDefinition('akeneo_batch.mail_notifier')
                 ->addTag('akeneo_batch.notifier');
         }
+
+        if ('test' === $container->getParameter('kernel.environment')) {
+            $loader->load('test/jobs.yml');
+        }
     }
 }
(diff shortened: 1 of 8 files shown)


StevenVAIDIE (approved):
:+1:

ValentinMumble (approved):

PR #17491, opened 2022-07-20
RAB-849: Add active button state on DSM button

ValentinMumble:
- Adding a new optional prop on the Button component to force its active state.
- Fixing issue with border width, setting the style as none would remove it completely, resulting in the ghost version being 2px wider than its non-ghost version

Before:
Screenshot 2022-07-20 at 11 06 40

After:
image


StevenVAIDIE (approved):
Nice catch

ghoussard (approved):

PR #17517, opened 2022-07-25
CPM-591: Add uuid in product imports

cemorin:

Allow uuid in import and find product by uuid when provided

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/src/Akeneo/Pim/Enrichment/Component/Product/Connector/Processor/Denormalizer/FindProductToImport.php
+++ b/src/Akeneo/Pim/Enrichment/Component/Product/Connector/Processor/Denormalizer/FindProductToImport.php
@@ -31,7 +31,7 @@ class FindProductToImport
     public function fromFlatData(
         string $productIdentifierCode,
         string $familyCode,
-        ?string $uuid,
+        ?string $uuid = null,
     ): ProductInterface {
         $product = null;
         if (null !== $uuid) {
--- a/src/Akeneo/Pim/Enrichment/Component/Product/Connector/Processor/Denormalizer/FindProductToImport.php
+++ b/src/Akeneo/Pim/Enrichment/Component/Product/Connector/Processor/Denormalizer/FindProductToImport.php
@@ -7,6 +7,8 @@ namespace Akeneo\Pim\Enrichment\Component\Product\Connector\Processor\Denormaliz
 use Akeneo\Pim\Enrichment\Component\Product\Builder\ProductBuilderInterface;
 use Akeneo\Pim\Enrichment\Component\Product\Model\ProductInterface;
 use Akeneo\Tool\Component\StorageUtils\Repository\IdentifiableObjectRepositoryInterface;
+use Ramsey\Uuid\Uuid;
+use Webmozart\Assert\Assert;
 
 /**
  * Find the product to import depending the data given by the reader
@@ -17,10 +19,6 @@ use Akeneo\Tool\Component\StorageUtils\Repository\IdentifiableObjectRepositoryIn
  */
 class FindProductToImport
 {
-    /**
-     * @param IdentifiableObjectRepositoryInterface $productRepository
-     * @param ProductBuilderInterface               $productBuilder
-     */
     public function __construct(
         private IdentifiableObjectRepositoryInterface $productRepository,
         private ProductBuilderInterface $productBuilder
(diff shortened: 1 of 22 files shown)


src/Akeneo/Pim/Enrichment/Component/Product/Connector/Processor/Denormalizer/FindProductToImport.php
```diff
@@ -17,37 +19,29 @@
...
-    /**
-     * @param IdentifiableObjectRepositoryInterface $productRepository
-     * @param ProductBuilderInterface               $productBuilder
-     */
     public function __construct(
-        IdentifiableObjectRepositoryInterface $productRepository,
-        ProductBuilderInterface $productBuilder
+        private IdentifiableObjectRepositoryInterface $productRepository,
+        private ProductBuilderInterface $productBuilder
     ) {
```
cemorin, line 25: Use another productRepository to use directly `find` method below

src/Akeneo/Pim/Enrichment/Component/Product/Connector/Processor/Denormalizer/FindProductToImport.php
```diff
@@ -17,37 +19,29 @@
...
-     * @param string $productIdentifierCode
-     * @param string $familyCode
-     *
-     * @return ProductInterface
      */
     public function fromFlatData(
         string $productIdentifierCode,
-        string $familyCode
+        string $familyCode,
+        ?string $uuid,
```
cemorin, line 34:
```suggestion
        ?string $uuid = null,
```

tests/back/Pim/Enrichment/Specification/Component/Product/Connector/Processor/Denormalizer/ProductProcessorSpec.php
```diff
@@ -1098,7 +1207,7 @@ function it_does_not_convert_a_variant_product_to_simple_product_when_the_job_pa
             'family' => 'clothing',
             'enabled' => true,
         ])->shouldBeCalled()->willReturn($filteredItem);
-        $productToImport->fromFlatData('my_sku', 'clothing')->willReturn($product);
+        $productToImport->fromFlatData('my_sku', 'clothing', null)->willReturn($product);
```
pierallard, line 1210: Don't forget to remove all these nulls

pierallard (approved):

src/Akeneo/Pim/Enrichment/Component/Product/Connector/Processor/Denormalizer/FindProductToImport.php
```diff
@@ -17,37 +19,29 @@
...
     ): ProductInterface {
-        $product = $this->productRepository->findOneByIdentifier($productIdentifierCode);
+        $product = null;
+        if (null !== $uuid) {
+            Assert::methodExists($this->productRepository, 'findOneByUuid');
```
SamirBoulil, line 38: I'm not sure why we need to make that check :/, the productrepository always as the findOneByUuid no ? it should be part of the ProductRepositoryInterface ?

cemorin: It's a tricky situation, this class does not use `ProductRepositoryInterface` but `IdentifiableObjectRepositoryInterface` which handles various types of objects.
And the injection used for this productRepository here is used both for product and productModels (`DelegatingProductRepository.php`).
It's not the best solution but validated for now with @pierallard

src/Akeneo/Pim/Enrichment/Component/Product/Connector/Processor/Denormalizer/FindProductToImport.php
SamirBoulil: Maybe if we "Refactor > Change method signature" we could avoid this = null and put the null method call in the codebase instead of in the function signatures

Balamung (approved):

src/Akeneo/Pim/Enrichment/Component/Product/Connector/Processor/Denormalizer/FindProductToImport.php
cemorin: So send the null directly from the place that calls `fromFlatData` ? Because the null is also handled there.
This change was in order to avoid changing all the existing tests by having to add this null parameter to all of them

PR #17524, opened 2022-07-26
PIM-10551: Add filtering on locale specific attributes when fetching values

Balamung:

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/CHANGELOG-5.0.md
+++ b/CHANGELOG-5.0.md
@@ -1,5 +1,8 @@
 # 5.0.x
 
+## Bug fixes
+- PIM-10551: Add filtering on locale specific attributes when fetching values
+
 # 5.0.104 (2022-07-11)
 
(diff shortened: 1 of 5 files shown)


ahocquard (commented):
Is the removal of a locale specific attribute handled in a background process?

Because:
- the versioning is wrong now
- the updated date of the product is wrong
- webhook are not triggered
- etc

Filtering values at read is just  a part of the problem actually.

ahocquard (approved):
After discussion with @pierrejolly and @SamirBoulil , it's GTM because it fixes the SLA. The whole usecase is not handled properly, but it is work as designed, and it's just badly designed.

pierrejolly (approved):

PR #17533, opened 2022-07-27
CXP-1238: Catalog Completeness criterion Back

RiffVentura:
- Added paginated endpoint to get channels with locales
- Made explicit/strict product selection criteria payload validation
- Updated internal PQB call  to use channels and locale as context options for filters


masacc (approved):

components/catalogs/Makefile
```diff
@@ -66,8 +66,8 @@ catalogs-unit-back:
 	APP_ENV=test $(PHP_RUN) vendor/bin/phpunit \
 		--configuration $(CATALOGS_PATH)/back/ \
 		--bootstrap $(CATALOGS_BOOTSTRAP_PATH) \
-		--testsuite Catalogs_Unit \
-		--log-junit var/tests/phpunit/phpunit_catalogs_unit.xml
+		--log-junit var/tests/phpunit/phpunit_catalogs_unit.xml \
+		--testsuite Catalogs_Unit ${O}
```
tseho, line 70:
```suggestion
		--testsuite Catalogs_Unit ${O}
```

I'm honestly against it.
Having arguments like `${0}` in the makefile make it really hard to understand and maintain.
Either we stop using the makefile and use a better alternative, or we could also create shortcuts.
I wouldn't mind something like `components/catalogs/bin/phpunit` with predefined args.

components/catalogs/back/src/Infrastructure/Validation/UpdateCatalogPayloadIsValidValidator.php
```diff
@@ -0,0 +1,214 @@
...
+                    'value' => [
+                        new Assert\Type('array'),
+                        new Assert\All(new Assert\Type('string')),
+                    ],
+                ],
+                'allowMissingFields' => false,
+                'allowExtraFields' => false,
+            ]),
+        ];
+    }
```
tseho, line 145:
```suggestion
    private function getFamilyConstraints(): array
    {
        return [
            new Assert\Collection([
                'fields' => [
                    'field' => [
                        new Assert\IdenticalTo('family'),
                    ],
                    'operator' => [
                        new Assert\Type('string'),
                        new Assert\Choice(['EMPTY', 'NOT EMPTY', 'IN', 'NOT IN']),
                    ],
                    'value' => [
                        new Assert\Type('array'),
                        new Assert\All(new Assert\Type('string')),
                    ],
                ],
                'allowMissingFields' => false,
                'allowExtraFields' => false,
            ]),
        ];
    }
```

Can you extract it to a separate Constraint class ? It could extends Compound.

This way, in 

```
                                $constraints = match ($criterion['field'] ?? null) {
                                    'enabled' => $this->getEnabledConstraints(),
                                    'family' => $this->getFamilyConstraints(),
                                    'completeness' => $this->getCompletenessConstraints(),
                                    default => []
                                };
```
We can do: 
```
                                $constraints = match ($criterion['field'] ?? null) {
                                    'enabled' => new EnabledFieldIsValid(),
                                    'family' => new FamilyFieldIsValid(),
                                    'completeness' => new CompletenessFieldIsValid(),
                                    default => []
                                };
```
And clearly separate the code.

The current version will not scale when we will support +20 types of fields.

src/Akeneo/Connectivity/Connection/back/Application/Settings/Command/CreateConnectionHandler.php
```diff
@@ -52,7 +52,7 @@ public function handle(CreateConnectionCommand $command): ConnectionWithCredenti
             throw new ConstraintViolationListException($violations);
         }
 
-        $user = $this->createUser->execute($command->code(), $command->label(), ' ');
+        $user = $this->createUser->execute($command->code(), $command->label(), ' ', [$command->getName()]);
```
tseho, line 55: It is related ?

components/catalogs/back/tests/Integration/Infrastructure/Persistence/GetChannelsQueryTest.php
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    public function testItGetsChannelByCode(): void
+    {
+        //Already existing as part of minimal catalog: ecommerce with en_US
+        $this->createChannel('tablet', ['en_US', 'de_DE']);
+        $this->createChannel('mobile', ['en_US', 'fr_FR']);
+
+        $result = $this->query->execute(1, 20, 'mobile');
+        $page2 = $this->query->execute(2, 20, 'mobile');
```
tseho, line 90:
```suggestion
```

components/catalogs/back/tests/Integration/Infrastructure/Persistence/GetChannelsQueryTest.php
```diff
@@ -0,0 +1,132 @@
...
+                'label'=> '[mobile]',
+                'locales' => [
+                    ['code' => 'en_US', 'label' => 'English (United States)'],
+                    ['code' => 'fr_FR', 'label' => 'French (France)'],
+                ],
+            ]
+        ];
+
+        self::assertEquals($expected, $result);
+        self::assertEquals([], $page2);
```
tseho, line 104:
```suggestion
```

components/catalogs/back/tests/Integration/Infrastructure/Persistence/GetChannelsQueryTest.php
```diff
@@ -0,0 +1,132 @@
...
+            'locales' => $locales,
+            'currencies' => ['USD'],
+            'category_tree' => 'master',
+        ]);
+
+        $violations = $this->validator->validate($channel);
+        self::assertSame(0, $violations->count(), (string) $violations);
+
+        $this->channelSaver->save($channel);
+    }
```
tseho, line 131: Can you move it to `components/catalogs/back/tests/Integration/IntegrationTestCase.php` ?

Maybe I'm wrong to share the factories but it will avoid to duplicate a lot of code in tests.
Also, when services API will be released for those, it will be easier to change it at one place without touching all tests.

In the long term strategy, I would like to have each factory in a trait included in IntegrationTestCase, so we don't end with a god class of 2000 lines.

components/catalogs/back/tests/Integration/Infrastructure/Persistence/GetProductIdentifiersQueryTest.php
```diff
@@ -64,6 +67,41 @@ public function testItThrowsWhenTheSearchAfterIsInvalid(): void
         $this->query->execute('db1079b6-f397-4a6a-bae4-8658e64ad47c', 'invalid_search_after');
     }
 
+    public function testItGetsPaginatedProductIdentifiersFromCatalogCriteriaHavingAScopeAndAChannel(): void
+    {
+        $this->purgeDataAndLoadTechnicalCatalog();
```
tseho, line 72: Can you avoid using the technical catalog ?

I would rather that we rely only on minimal and force us to explicitly setup the data in the test.
Looking at the test here, I don't know why it's working with completeness > 25 ? How many required attributes there is in the ecommerce channel in technical ?

Before spending time on a rewrite, what is needed in this test to use minimal ? Do you think it's too dirty to create the attributes, attributes groups, etc ?

components/catalogs/back/tests/Integration/IntegrationTestCase.php
```diff
@@ -112,15 +121,18 @@ protected function getAuthenticatedInternalApiClient(string $username = 'admin')
 
     protected function assertViolationsListContains(
         ConstraintViolationListInterface $violations,
-        string $expectedMessage
+        string $expectedMessage,
+        ?string $expectedPropertyPath = null,
```
tseho, line 125: Do you think that we should have a separate one for the property path case ?

Something like `protected function assertViolationsListContainsAtPath(`

components/catalogs/back/tests/Integration/IntegrationTestCase.php
```diff
@@ -51,6 +54,14 @@ protected function purgeDataAndLoadMinimalCatalog(): void
...
     }
 
+    protected function purgeDataAndLoadTechnicalCatalog(): void
+    {
+        $catalog = self::getContainer()->get('akeneo_integration_tests.catalogs');
+        $configuration = $catalog->useTechnicalCatalog();
+        $fixturesLoader = self::getContainer()->get('akeneo_integration_tests.loader.fixtures_loader');
+        $fixturesLoader->load($configuration);
+    }
+
```
tseho, line 64:
```suggestion
```

:fire:

components/catalogs/back/src/Infrastructure/Validation/UpdateCatalogPayloadIsValidValidator.php
```diff
@@ -0,0 +1,98 @@
...
+                                if (!\is_array($criterion)) {
+                                    return;
+                                }
+
+                                $constraints = match ($criterion['field'] ?? null) {
+                                    'completeness' => new CompletenessFieldIsValid($this->getChannelLocalesQuery),
+                                    'enabled' => new EnabledFieldIsValid(),
+                                    'family' => new FamilyFieldIsValid(),
+                                    default => null
+                                };
```
tseho, line 75:
```suggestion
                                $constraints = match ($criterion['field'] ?? null) {
                                    'completeness' => new CompletenessFieldIsValid($this->getChannelLocalesQuery),
                                    'enabled' => new EnabledFieldIsValid(),
                                    'family' => new FamilyFieldIsValid(),
                                    default => null
                                };
```

What do you think about extracting this part to a new private method ?
It could be clearer when adding new entries to it.

tseho (approved):

guillaumefer-akeneo (approved):

components/catalogs/back/src/Infrastructure/Validation/UpdateCatalogPayloadIsValidValidator.php
masacc: I agree, I'll make the change

PR #17542, opened 2022-07-28
CPM-636: Add Create product by uuid

pierallard:
This PR adds CREATE products endpoint
There are 3 "crappy" things:
- We have to get identifiers from associated products uuids
- We can't create a product by defining its UUID

Cards are added in the backlog, endpoint works.


src/Akeneo/Pim/Enrichment/Bundle/Controller/ExternalApi/CreateProductByUuidController.php
```diff
@@ -0,0 +1,246 @@
...
+            $this->throwDocumentedHttpException(sprintf('The %s attribute does not exist in your PIM.', $e->getAttributeCode()), $e);
+        } catch (\InvalidArgumentException $e) {
+            $this->throwDocumentedHttpException($e->getMessage(), $e);
+        } catch (ViolationsException $e) {
+            $this->throwDocumentedHttpException($e->violations()->get(0)->getMessage(), $e);
+        } catch (LegacyViolationsException $e) {
+            $this->throwDocumentedHttpException($e->violations()->get(0)->getMessage(), $e);
+        } catch (InvalidPropertyTypeException $e) {
+            $this->throwDocumentedHttpException($e->getMessage(), $e);
+        }
```
nmarniesse, line 113:
```suggestion
        } catch (\InvalidArgumentException|InvalidPropertyTypeException $e) {
            $this->throwDocumentedHttpException($e->getMessage(), $e);
        } catch (ViolationsException|LegacyViolationsException $e) {
            $this->throwDocumentedHttpException($e->violations()->get(0)->getMessage(), $e);
        }
```

src/Akeneo/Pim/Enrichment/Bundle/Controller/ExternalApi/CreateProductByUuidController.php
```diff
@@ -0,0 +1,246 @@
...
+                'identifier'
+            );
+        }
+
+        if (array_key_exists('variant_group', $data)) {
+            throw new DocumentedHttpException(
+                Documentation::URL_DOCUMENTATION . 'products-with-variants.html',
+                'Property "variant_group" does not exist anymore. Check the link below to understand why.'
+            );
+        }
```
nmarniesse, line 97:
```suggestion
```

It's a new endpoint, so no historic on the format.  
And this message was added in 2017... Maybe it can be removed on the other endpoints now

src/Akeneo/Pim/Enrichment/Bundle/Controller/ExternalApi/CreateProductByUuidController.php
```diff
@@ -0,0 +1,246 @@
...
+
+        try {
+            $this->duplicateValueChecker->check($data);
+        } catch (InvalidPropertyTypeException $e) {
+            $this->eventDispatcher->dispatch(new TechnicalErrorEvent($e));
+            $this->throwDocumentedHttpException($e->getMessage(), $e);
+        }
+
+        $data = $this->orderData($data);
```
nmarniesse, line 78: What is the problem if we don't order?

src/Akeneo/Pim/Enrichment/Bundle/Controller/ExternalApi/CreateProductByUuidController.php
```diff
@@ -0,0 +1,246 @@
...
+            return $data;
+        }
+
+        return ['parent' => $data['parent']] + $data;
+    }
+
+    private function getProductIdentifier(array $data): string
+    {
+        return $data['values'][$this->attributeRepository->getIdentifierCode()][0]['data'];
+    }
```
nmarniesse, line 169:
```suggestion
    private function getProductIdentifier(array $data): ?string
    {
        return $data['values'][$this->attributeRepository->getIdentifierCode()][0]['data'] ?? null;
    }
```

src/Akeneo/Pim/Enrichment/Bundle/Controller/ExternalApi/CreateProductByUuidController.php
pierallard: I personally prefer list exception by exception, I find it more readable

src/Akeneo/Pim/Enrichment/Bundle/Controller/ExternalApi/CreateProductByUuidController.php
pierallard: I copy pasted the naming ; in fact, this method only formats "parent" property that is not the format expected from the API.

tests/back/Pim/Enrichment/EndToEnd/Product/Product/ExternalApi/CreateProductWithUuidEndToEnd.php
```diff
@@ -0,0 +1,1160 @@
...
+        $this->assertSame(Response::HTTP_CREATED, $response->getStatusCode());
+
+        $product = $this->get('pim_catalog.repository.product')->findOneByIdentifier('foo');
+        $standardizedProduct = $this->get('pim_standard_format_serializer')->normalize($product, 'standard');
+
+        $this->assertNotSame('2014-06-14T13:12:50+02:00', $standardizedProduct['created']);
+        $this->assertNotSame('2014-06-14T13:12:50+02:00', $standardizedProduct['updated']);
+    }
+
+    public function testProductCreationWithIdenticalIdentifiers()
```
pierallard, line 783: drop

Balamung (approved):

cemorin (approved):

src/Akeneo/Pim/Enrichment/Bundle/Controller/ExternalApi/ServiceAPIProductUpdater.php
```diff
@@ -0,0 +1,191 @@
...
+
+                $this->throwDocumentedHttpException(
+                    sprintf('Property "categories" expects a valid category code. The category does not exist, "%s" given.', $categoryCodes),
+                    $e
+                );
+            }
+
+            $message = $e->violations()->get(0)->getMessage();
+            $matches = [];
+            if (preg_match('/^Property "associations" expects a valid product identifier. The product does not exist, "(?P<identifier>.*)" given.$/', $message, $matches)) {
```
nmarniesse, line 102: We should avoid preg_match. I tried to find a solution but I don't understand how it's working:
- we give a uuid in entry, but where is it validated? AssociationFieldSetter works only with skus
- if the product with uuid does not exist, how could we find its sku? And vice versa

src/Akeneo/Pim/Enrichment/Bundle/Controller/ExternalApi/ServiceAPIProductUpdater.php
```diff
@@ -0,0 +1,191 @@
...
+            $previousException
+        );
+    }
+
+    private function throwViolationExceptionAndReplaceIdentifiersByUuids(ConstraintViolationListInterface $violations): void
+    {
+        $newViolations = new ConstraintViolationList();
+        foreach ($violations as $violation) {
+            $messageTemplate = $violation->getMessageTemplate();
+            if ($messageTemplate === 'pim_catalog.constraint.quantified_associations.products_do_not_exist') {
```
nmarniesse, line 149: Not a fan. Same remark as above for associations.

An other solution would be to change ReplaceAssociatedProducts to accept uuids (or add new user intent to handle this), then add a validator on this user intent. So in this new validator we could format the message as we want

src/Akeneo/Pim/Enrichment/Bundle/Controller/ExternalApi/ServiceAPIProductUpdater.php
pierallard: User sets a uuid in the body, but I replace them by identifiers in the controller private methods. So the Service API receives identifiers, but the message should not talk about identifiers but uuids, so I have to rephrase the message.
This should be fixed when we'll be able to manage SetAssociations with Uuid in Service API. Céline is aware of this.

src/Akeneo/Pim/Enrichment/Bundle/Controller/ExternalApi/ServiceAPIProductUpdater.php
pierallard: I'm not a fan too, that's the same response, it's temporary until we can associate products through UUID.

PR #17544, opened 2022-07-28
CPM-691: Remove triggers introduced during UUID migration + Clean versioning

SamirBoulil:

This PR introduces 2 migrations:
- A Doctrine migration to drop all triggers during the UUID migration project
- A ZDD migration which empties the `pim_versioning_version.resource_uuid` for versions that are not product related.
    - When running the command `pim:zdd-migration:migrate` we can see the migration updates successfully
Screenshot 2022-07-29 at 16 30 47

- [x] Tests

### DIFF
--- a/src/Akeneo/Pim/Enrichment/Bundle/Command/ZddMigrations/V20220729171405CleanVersioningResourceUuidColumnForNonProductVersions.php
+++ b/src/Akeneo/Pim/Enrichment/Bundle/Command/ZddMigrations/V20220729171405CleanVersioningResourceUuidColumnForNonProductVersions.php
@@ -98,7 +98,7 @@ SQL;
         $allResourceNames = $this->connection->fetchFirstColumn($sql);
         $nextResourceNameToProcess = current(array_filter(
             $allResourceNames,
-            static fn(string $resourceName) => $resourceName !== Product::class
+            static fn (string $resourceName) => $resourceName !== Product::class
         ));
 
         return false !== $nextResourceNameToProcess ? $nextResourceNameToProcess : null;
--- a/src/Akeneo/Pim/Enrichment/Bundle/Command/ZddMigrations/V20220729171405CleanVersioningResourceUuidColumnForNonProductVersions.php
+++ b/src/Akeneo/Pim/Enrichment/Bundle/Command/ZddMigrations/V20220729171405CleanVersioningResourceUuidColumnForNonProductVersions.php
@@ -47,9 +47,18 @@ SQL;
     private function cleanVersioningResourceUuid(): void
     {
         $resourceNameToProcess = $this->findNextResourceNameToProcess();
+        $totalVersionsCleaned = 0;
         while (null !== $resourceNameToProcess) {
             $versionIds = $this->getVersionIdsToCleanForResourceName($resourceNameToProcess);
             $this->cleanVersions($versionIds);
+            $totalVersionsCleaned += \count($versionIds);
+            $this->logger->debug(
+                sprintf(
+                    'Cleaned %d versions from their resource_uuid for resource name %s',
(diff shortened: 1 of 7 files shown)


src/Akeneo/Pim/Enrichment/Bundle/Command/ZddMigrations/V20220729171405CleanVersioningResourceUuidColumnForNonProductVersions.php
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    private function getVersionIdsToClean(): array
+    {
+        $sql = <<<SQL
+SELECT id
+FROM pim_versioning_version
+WHERE 
+	resource_uuid IS NOT NULL
+	AND resource_name != :resource_name
```
SamirBoulil, line 44: I can remove the parameter if you want here and put the RAW product class name here by escaping anti-slashes 4 times: '\\\\'

src/Akeneo/Pim/Enrichment/Bundle/Command/ZddMigrations/V20220729171405CleanVersioningResourceUuidColumnForNonProductVersions.php
```diff
@@ -0,0 +1,105 @@
...
+        while (null !== $resourceNameToProcess) {
+            $versionIds = $this->getVersionIdsToCleanForResourceName($resourceNameToProcess);
+            $this->cleanVersions($versionIds);
+            $resourceNameToProcess = $this->findNextResourceNameToProcess();
+        }
+    }
+
+    private function findNextResourceNameToProcess(): ?string
+    {
+        // This query takes ~40 secs to execute on a big catalog (4M versions)
```
SamirBoulil, line 59: I think that's alright, I didn't want to keep track of the resource name processed in the class etc.

upgrades/schema/Version_7_0_20220728121643_remove_uuid_triggers.php
```diff
@@ -0,0 +1,66 @@
...
+        'pim_catalog_completeness_uuid_update',
+        'pim_dqi_product_criteria_evaluation_uuid_insert',
+        'pim_dqi_product_criteria_evaluation_uuid_update',
+        'pim_dqi_product_score_uuid_insert',
+        'pim_dqi_product_score_uuid_update',
+        'pim_versioning_version_uuid_insert',
+        'pim_versioning_version_uuid_update',
+        'pim_comment_comment_uuid_insert',
+        'pim_comment_comment_uuid_update',
+        'pimee_workflow_product_draft_uuid_insert',
```
SamirBoulil, line 40: I think it's safe to do it in the CE migration, it does not make the migration fail if the trigger does not exist.

src/Akeneo/Pim/Enrichment/Bundle/Command/ZddMigrations/V20220729171405CleanVersioningResourceUuidColumnForNonProductVersions.php
ahocquard:
```
SELECT resource_name
FROM pim_versioning_version
WHERE 
	resource_uuid IS NOT NULL
LIMIT 1;
```

Should be enough in that case. It will return the first resource_name.

SamirBoulil: Ah nah because if the one that is returned is the Product::class, I'll stop looping 😬. I prefer to fetch all of them each time. It's zdd, let's call it a compromise :) ?

PR #17550, opened 2022-07-29
CXP-1253: Added notifications to user when  catalog is saved

RiffVentura:
- Catalog save returns a boolean that indicates save success
 - Added notification inside connectivity to notify when the catalog is saved
 - Updated the fake catalog edit container to have some feedback on save
 
 Need feedback on: If catalogs cannot be saved due to user error notification is a warning

### DIFF
--- a/components/catalogs/front/src/FakeCatalogEditContainer.tsx
+++ b/components/catalogs/front/src/FakeCatalogEditContainer.tsx
@@ -26,16 +26,16 @@ type Props = {};
 const FakeCatalogEditContainer: FC<PropsWithChildren<Props>> = () => {
     const {id} = useParams<{id: string}>();
     const [form, save, isDirty] = useCatalogForm(id);
-    const deps = useDependenciesContext();
+    const {notify} = useDependenciesContext();
 
     const saveHandler = async () => {
         const isSaveSuccessful = await save();
 
-        if (deps.notify) {
+        if (notify) {
             if (isSaveSuccessful) {
-                deps.notify(NotificationLevel.SUCCESS, 'Catalog is saved');
+                notify(NotificationLevel.SUCCESS, 'Catalog is saved');
             } else {
-                deps.notify(NotificationLevel.ERROR, 'Catalog have errors');
+                notify(NotificationLevel.ERROR, 'Catalog have errors');
             }
         }
     };
--- a/components/catalogs/front/src/FakeCatalogEditContainer.tsx
+++ b/components/catalogs/front/src/FakeCatalogEditContainer.tsx
@@ -1,4 +1,4 @@
-import React, {FC, PropsWithChildren} from 'react';
+import React, {FC, PropsWithChildren, useState} from 'react';
 import {useParams} from 'react-router';
 import {CatalogEdit, useCatalogForm} from './components/CatalogEdit';
 import {Button} from 'akeneo-design-system';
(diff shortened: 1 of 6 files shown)


components/catalogs/front/src/FakeCatalogEditContainer.tsx
```diff
@@ -20,11 +20,31 @@ const DirtyWarning = styled.div`
...
 type Props = {};
 
 const FakeCatalogEditContainer: FC<PropsWithChildren<Props>> = () => {
     const {id} = useParams<{id: string}>();
     const [form, save, isDirty] = useCatalogForm(id);
+    const [isSuccess, setSuccess] = useState<boolean | null>(null);
+
+    const saveHandler = async () => {
+        const isSaveSuccessful = await save();
+        setSuccess(isSaveSuccessful);
```
masacc, line 46: what do you think of adding a `setTimeout()`-like timer to make it disappear? do you take a look at other CRA to see how they fake the notification system (like they did for the translate hook)?

src/Akeneo/Connectivity/Connection/front/src/connect/components/ConnectedApp/Catalog/ConnectedAppCatalogContainer.tsx
```diff
@@ -17,16 +18,33 @@ type Props = {
...
     const [form, save, isDirty] = useCatalogForm(catalog.id);
 
+    const handleSave = async () => {
+        try {
+            const success = await save();
+
+            const message = success
+                ? 'akeneo_connectivity.connection.connect.connected_apps.edit.catalogs.edit.flash.success'
+                : 'akeneo_connectivity.connection.connect.connected_apps.edit.catalogs.edit.flash.warning';
+            notify(success ? NotificationLevel.SUCCESS : NotificationLevel.WARNING, translate(message));
```
masacc, line 36: I think an error you can use `NotificationLevel.ERROR` here too. Warning is to say "that's not perfect, it can fail in the future, but it works for now", here it did not work.

masacc (approved):

components/catalogs/front/src/FakeCatalogEditContainer.tsx
masacc: I found that in the dependencies of the microfrontend, there is a notification system. You can use it like this :
```
import {NotificationLevel, useDependenciesContext} from '@akeneo-pim-community/shared';

...


    const deps = useDependenciesContext();

    const saveHandler = async () => {
        const isSaveSuccessful = await save();

        if (deps.notify) {
            if (isSaveSuccessful) {
                deps.notify(NotificationLevel.SUCCESS, 'Catalog is saved');
            } else {
                deps.notify(NotificationLevel.ERROR, 'Catalog have errors');
            }
        }
    };
```

components/catalogs/front/src/FakeCatalogEditContainer.tsx
```diff
@@ -25,6 +26,19 @@ type Props = {};
 const FakeCatalogEditContainer: FC<PropsWithChildren<Props>> = () => {
     const {id} = useParams<{id: string}>();
     const [form, save, isDirty] = useCatalogForm(id);
+    const deps = useDependenciesContext();
```
tseho, line 29:
```suggestion
    const {notify} = useDependenciesContext();
```

tseho: Why can't you do this ?

src/Akeneo/Connectivity/Connection/front/src/connect/components/ConnectedApp/Catalog/ConnectedAppCatalogContainer.tsx
tseho: Errors during save are always "ERROR" in the PIM.

tseho (approved):

components/catalogs/front/src/FakeCatalogEditContainer.tsx
masacc: yes I can, bad copy paste

PR #17569, opened 2022-08-02
CPM-698: Add ValueObject to allow user to call UpsertProductCommand with uuid or identifier

cemorin:
We want to be able to create or update a product with its uuid

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/src/Akeneo/Pim/Enrichment/Bundle/Controller/InternalApi/Product/UpdateProductController.php
+++ b/src/Akeneo/Pim/Enrichment/Bundle/Controller/InternalApi/Product/UpdateProductController.php
@@ -17,6 +17,7 @@ use Akeneo\Pim\Enrichment\Product\API\Command\Exception\LegacyViolationsExceptio
 use Akeneo\Pim\Enrichment\Product\API\Command\Exception\ViolationsException;
 use Akeneo\Pim\Enrichment\Product\API\Command\UpsertProductCommand;
 use Akeneo\Pim\Enrichment\Product\API\Query\GetUserIntentsFromStandardFormat;
+use Akeneo\Pim\Enrichment\Product\API\ValueObject\ProductIdentifier;
 use Akeneo\Pim\Enrichment\Product\Domain\Model\ViolationCode;
 use Akeneo\Tool\Component\StorageUtils\Saver\SaverInterface;
 use Akeneo\Tool\Component\StorageUtils\Updater\ObjectUpdaterInterface;
@@ -177,7 +178,7 @@ final class UpdateProductController
         $userId = $this->userContext->getUser()?->getId();
         $command = UpsertProductCommand::createFromCollection(
             $userId,
-            $product->getIdentifier() ?? '',
+            ProductIdentifier::fromSku($product->getIdentifier()) ?? '',
             $userIntents
         );
         $this->commandMessageBus->dispatch($command);
--- a/src/Akeneo/Pim/Enrichment/Bundle/Controller/InternalApi/Product/UpdateProductController.php
+++ b/src/Akeneo/Pim/Enrichment/Bundle/Controller/InternalApi/Product/UpdateProductController.php
@@ -19,6 +19,7 @@ use Akeneo\Pim\Enrichment\Product\API\Command\UpsertProductCommand;
 use Akeneo\Pim\Enrichment\Product\API\Query\GetUserIntentsFromStandardFormat;
 use Akeneo\Pim\Enrichment\Product\API\ValueObject\ProductIdentifier;
 use Akeneo\Pim\Enrichment\Product\Domain\Model\ViolationCode;
+use Akeneo\Pim\Structure\Component\Repository\AttributeRepositoryInterface;
 use Akeneo\Tool\Component\StorageUtils\Saver\SaverInterface;
 use Akeneo\Tool\Component\StorageUtils\Updater\ObjectUpdaterInterface;
 use Akeneo\UserManagement\Bundle\Context\UserContext;
@@ -58,6 +59,7 @@ final class UpdateProductController
         protected AttributeFilterInterface $productAttributeFilter,
         private MessageBusInterface $commandMessageBus,
         private MessageBusInterface $queryMessageBus,
+        private AttributeRepositoryInterface $attributeRepository,
     ) {
     }
 
@@ -176,9 +178,10 @@ final class UpdateProductController
         $userIntents = $handledStamp->getResult();
 
         $userId = $this->userContext->getUser()?->getId();
+        $identifierCode = $this->attributeRepository->getIdentifierCode();
         $command = UpsertProductCommand::createFromCollection(
             $userId,
-            ProductIdentifier::fromSku($product->getIdentifier()) ?? '',
+            ProductIdentifier::fromAttributeCodeAndIdentifier($identifierCode, $product->getIdentifier()) ?? '',
             $userIntents
         );
         $this->commandMessageBus->dispatch($command);
--- a/src/Akeneo/Pim/Enrichment/Component/Product/Model/AbstractProduct.php
+++ b/src/Akeneo/Pim/Enrichment/Component/Product/Model/AbstractProduct.php
@@ -98,6 +98,7 @@ abstract class AbstractProduct implements ProductInterface
 
     public function setUuid(UuidInterface $uuid): ProductInterface
     {
+        Assert::same($uuid->getVersion(), 4);
         $this->uuid = $uuid;
 
         return $this;
--- a/src/Akeneo/Pim/Enrichment/Component/Product/Model/AbstractProduct.php
+++ b/src/Akeneo/Pim/Enrichment/Component/Product/Model/AbstractProduct.php
@@ -96,14 +96,6 @@ abstract class AbstractProduct implements ProductInterface
         return $this->uuid;
     }
 
-    public function setUuid(UuidInterface $uuid): ProductInterface
-    {
-        Assert::same($uuid->getVersion(), 4);
-        $this->uuid = $uuid;
-
-        return $this;
-    }
-
     /**
      * {@inheritdoc}
      */
--- a/src/Akeneo/Pim/Enrichment/Component/Product/Model/AbstractProduct.php
+++ b/src/Akeneo/Pim/Enrichment/Component/Product/Model/AbstractProduct.php
@@ -96,6 +96,13 @@ abstract class AbstractProduct implements ProductInterface
         return $this->uuid;
     }
 
+    public function setUuid(UuidInterface $uuid): ProductInterface
+    {
+        $this->uuid = $uuid;
+
+        return $this;
+    }
+
     /**
      * {@inheritdoc}
      */
--- a/src/Akeneo/Pim/Enrichment/Product/back/API/Command/UpsertProductCommand.php
+++ b/src/Akeneo/Pim/Enrichment/Product/back/API/Command/UpsertProductCommand.php
@@ -28,6 +28,12 @@ use Webmozart\Assert\Assert;
 final class UpsertProductCommand
 {
     /**
+     * @param ProductIdentifier|ProductUuid|string|null $identifierOrUuid
+     * The product can now be created or updated by :
+     * - a string: will be deprecated, related to the only attribute identifier
+     * - a ProductIdentifier: a pair attribute of the identifier code + the attribute code
+     * - a ProductUuid: the uuid of the existing product or the one you want to assign to the future product
+     * - null: the product will have a random uuid and no identifier
      * @param ValueUserIntent[] $valueUserIntents
      */
     public function __construct(
--- a/src/Akeneo/Pim/Enrichment/Product/back/API/Command/UpsertProductCommand.php
+++ b/src/Akeneo/Pim/Enrichment/Product/back/API/Command/UpsertProductCommand.php
@@ -15,6 +15,8 @@ use Akeneo\Pim\Enrichment\Product\API\Command\UserIntent\QuantifiedAssociation\Q
 use Akeneo\Pim\Enrichment\Product\API\Command\UserIntent\SetEnabled;
 use Akeneo\Pim\Enrichment\Product\API\Command\UserIntent\UserIntent;
 use Akeneo\Pim\Enrichment\Product\API\Command\UserIntent\ValueUserIntent;
+use Akeneo\Pim\Enrichment\Product\API\ValueObject\ProductIdentifier;
+use Akeneo\Pim\Enrichment\Product\API\ValueObject\ProductUuid;
 use Webmozart\Assert\Assert;
 
 /**
@@ -30,7 +32,7 @@ final class UpsertProductCommand
      */
     public function __construct(
         private int $userId,
-        private string $productIdentifier,
+        private ProductIdentifier | ProductUuid | string | null $identifierOrUuid = null,
         private ?FamilyUserIntent $familyUserIntent = null,
         private ?CategoryUserIntent $categoryUserIntent = null,
         private ?ParentUserIntent $parentUserIntent = null,
@@ -46,7 +48,7 @@ final class UpsertProductCommand
     /**
      * @param UserIntent[] $userIntents
      */
-    public static function createFromCollection(int $userId, string $productIdentifier, array $userIntents): self
+    public static function createFromCollection(int $userId, ProductIdentifier | ProductUuid | string | null $identifierOrUuid = null, array $userIntents = []): self
     {
         $valueUserIntents = [];
         $categoryUserIntent = null;
@@ -83,7 +85,7 @@ final class UpsertProductCommand
 
         return new self(
             userId: $userId,
-            productIdentifier: $productIdentifier,
+            identifierOrUuid: $identifierOrUuid,
             familyUserIntent: $familyUserIntent,
             categoryUserIntent: $categoryUserIntent,
             parentUserIntent: $parentUserIntent,
@@ -104,9 +106,9 @@ final class UpsertProductCommand
         return $this->userId;
     }
 
-    public function productIdentifier(): string
+    public function identifierOrUuid(): ProductIdentifier | ProductUuid | string | null
     {
-        return $this->productIdentifier;
+        return $this->identifierOrUuid;
     }
 
     public function familyUserIntent(): ?FamilyUserIntent
--- a/src/Akeneo/Pim/Enrichment/Product/back/API/ValueObject/ProductIdentifier.php
+++ b/src/Akeneo/Pim/Enrichment/Product/back/API/ValueObject/ProductIdentifier.php
@@ -1,0 +1,40 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Akeneo\Pim\Enrichment\Product\API\ValueObject;
+
+use Webmozart\Assert\Assert;
+
+/**
+ * @copyright 2022 Akeneo SAS (https://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+class ProductIdentifier
+{
+    private function __construct(private string $attributeCode, private string $identifier)
+    {
+        Assert::notEmpty($this->attributeCode);
+        Assert::notEmpty($this->identifier);
+    }
+
+    public static function fromAttributeCodeAndIdentifier(string $attributeCode, string $identifier): ProductIdentifier
+    {
+        return new ProductIdentifier($attributeCode, $identifier);
+    }
+
+    public static function fromSku(string $sku): ProductIdentifier
+    {
+        return new ProductIdentifier('sku', $sku);
+    }
+
+    public function attributeCode(): string
+    {
+        return $this->attributeCode;
+    }
+
+    public function identifier(): string
+    {
+        return $this->identifier;
+    }
+}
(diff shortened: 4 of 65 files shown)


src/Akeneo/Pim/Enrichment/Product/back/API/ValueObject/ProductIdentifier.php
```diff
@@ -0,0 +1,40 @@
...
+        Assert::notEmpty($this->attributeCode);
+        Assert::notEmpty($this->identifier);
+    }
+
+    public static function fromAttributeCodeAndIdentifier(string $attributeCode, string $identifier): ProductIdentifier
+    {
+        return new ProductIdentifier($attributeCode, $identifier);
+    }
+
+    public static function fromSku(string $sku): ProductIdentifier
```
cemorin, line 26: remove

src/Akeneo/Pim/Enrichment/Product/back/Application/UpsertProductHandler.php
```diff
@@ -50,12 +54,48 @@ public function __invoke(UpsertProductCommand $command): void
 
         $this->checkConsistencyWithConnectedUser($command->userId());
 
-        $product = $this->productRepository->findOneByIdentifier($command->productIdentifier());
+        // Temporary: we retrieve identifier value from the valueUserIntents to build the product
+        $identifier = \array_filter(
```
cemorin, line 58: remove

src/Akeneo/Pim/Enrichment/Product/back/Application/UpsertProductHandler.php
```diff
@@ -50,12 +54,48 @@ public function __invoke(UpsertProductCommand $command): void
...
+
+        if ($command->identifierOrUuid() instanceof ProductUuid) {
+            // TODO: should we retrieve product by identifier if product is null here?
+            $product = $this->productRepository->find($command->identifierOrUuid()->uuid());
+        } elseif ($command->identifierOrUuid() instanceof ProductIdentifier) {
+            $product = $this->productRepository->findOneByIdentifier($command->identifierOrUuid()->identifier());
+        } elseif (\is_string($command->identifierOrUuid())) {
+            $product = $this->productRepository->findOneByIdentifier($command->identifierOrUuid());
+        } else {
+            if (null === $identifier?->value()) {
```
cemorin, line 71: remove

src/Akeneo/Pim/Enrichment/Product/back/Application/UpsertProductHandler.php
```diff
@@ -50,12 +54,48 @@ public function __invoke(UpsertProductCommand $command): void
...
+
         $isCreation = false;
         if (null === $product) {
             $isCreation = true;
-            $product = $this->productBuilder->createProduct($command->productIdentifier());
+
+            if ($command->identifierOrUuid() instanceof ProductIdentifier) {
+                $product = $this->productBuilder->createProduct($command->identifierOrUuid()->identifier());
+            } elseif ($command->identifierOrUuid() instanceof ProductUuid) {
+                $product = $this->productBuilder->createProduct($identifier->value(), null, $command->identifierOrUuid()->uuid());
```
cemorin, line 85: this->productBuilder->createProduct(null, null, $command->identifierOrUuid()->uuid());

src/Akeneo/Pim/Enrichment/Product/back/Application/UpsertProductHandler.php
```diff
@@ -50,12 +54,48 @@ public function __invoke(UpsertProductCommand $command): void
...
+                $product->setUuid($command->identifierOrUuid()->uuid());
```
cemorin, line 86: remove

src/Akeneo/Pim/Enrichment/Product/back/Application/UpsertProductHandler.php
```diff
@@ -50,12 +54,48 @@ public function __invoke(UpsertProductCommand $command): void
...
+            } elseif (\is_string($command->identifierOrUuid())) {
+                $product = $this->productBuilder->createProduct($command->identifierOrUuid());
+            } else {
+                if (null === $identifier?->value()) {
```
cemorin, line 90: remove

src/Akeneo/Pim/Enrichment/Component/Product/Model/AbstractProduct.php
```diff
@@ -94,6 +94,13 @@ public function getUuid(): UuidInterface
         return $this->uuid;
     }
 
+    public function setUuid(UuidInterface $uuid): ProductInterface
+    {
+        $this->uuid = $uuid;
```
pierallard, line 99: Assert that $uuid version is 4 please

src/Akeneo/Pim/Enrichment/Product/back/API/Command/UpsertProductCommand.php
```diff
@@ -30,7 +32,7 @@ final class UpsertProductCommand
      */
     public function __construct(
         private int $userId,
-        private string $productIdentifier,
+        private ProductIdentifier | ProductUuid | string | null $identifierOrUuid = null,
```
pierallard, line 35: Please add documentation to explain that a product can be retrived/created by
- a string : will be deprecated, related to the only attribute identifier
- a ProductIdentifier: a pair attribute of type identifier + a string
- a ProductUuid

src/Akeneo/Pim/Enrichment/Product/back/Domain/Query/GetCategoryCodes.php
```diff
@@ -22,4 +23,15 @@ interface GetCategoryCodes
      *  }
      */
     public function fromProductIdentifiers(array $productIdentifiers): array;
```
pierallard, line 25: Will this method be deprecated ? If yes, please add @deprecated in it

src/Akeneo/Pim/Enrichment/Product/back/Domain/Query/GetNonViewableCategoryCodes.php
```diff
@@ -22,4 +23,15 @@ interface GetNonViewableCategoryCodes
      *  }
      */
     public function fromProductIdentifiers(array $productIdentifiers, int $userId): array;
```
pierallard, line 25: Will this method be deprecated ? If yes, please add @deprecated in it

src/Akeneo/Pim/Enrichment/Product/back/Infrastructure/Validation/IsUserOwnerOfTheProductValidator.php
```diff
@@ -32,13 +34,17 @@ public function validate($command, Constraint $constraint): void
...
 
-        try {
-            $productIdentifier = ProductIdentifier::fromString($command->productIdentifier());
-        } catch (\InvalidArgumentException) {
-            return;
+        $productCategoryCodes = null;
+        if ($command->identifierOrUuid() instanceof ProductIdentifierValueObject) {
+            $productCategoryCodes = $this->getProductCategoryCodesFromIdentifier($command->identifierOrUuid()->identifier());
+        } elseif ($command->identifierOrUuid() instanceof ProductUuid) {
+            $productCategoryCodes = $this->getCategoryCodes->fromProductUuids([
```
pierallard, line 41: For less maintenance, maybe it will be better to only call methods xxxFromProductUuids, then have an adapter from identifier to uuid in this file.

src/Akeneo/Pim/Enrichment/Product/back/Infrastructure/Validation/IsUserOwnerOfTheProductValidator.php
```diff
@@ -50,4 +56,17 @@ public function validate($command, Constraint $constraint): void
                 ->addViolation();
         }
     }
+
+    /**
+     * @return mixed|string[]|void|null
```
pierallard, line 61: Result is very complex for a simple function like this IMHO

src/Akeneo/Pim/Enrichment/Product/back/Infrastructure/Validation/ShouldStayOwnerOfTheProductValidator.php
```diff
@@ -46,14 +48,38 @@ public function validate($categoryUserIntent, Constraint $constraint): void
         Assert::isInstanceOf($command, UpsertProductCommand::class);
 
         if ($categoryUserIntent instanceof SetCategories) {
-            $nonViewableCategoryCodes = $this->getNonViewableCategoryCodes->fromProductIdentifiers([
-                ProductIdentifier::fromString($command->productIdentifier()),
-            ], $command->userId())[$command->productIdentifier()] ?? [];
+            $nonViewableCategoryCodes = [];
+            if ($command->identifierOrUuid() instanceof ProductIdentifierValueObject) {
+                $nonViewableCategoryCodes = $this->getNonViewableCategoryCodes->fromProductIdentifiers([
```
pierallard, line 53: Same remark than above

src/Akeneo/Pim/Enrichment/Product/back/Test/Integration/Handler/UpsertProductIntegration.php
```diff
@@ -88,7 +91,45 @@ public function it_creates_an_empty_product(): void
...
+
+        $this->clearDoctrineUoW();
+        $product = $this->productRepository->find($uuid);
+        Assert::assertNotNull($product);
+        Assert::assertSame('identifier', $product->getIdentifier());
+        Assert::assertSame($uuid->toString(), $product->getUuid()->toString());
+    }
+
+    /** @test */
+    public function it_creates_a_product_without_identifier_or_uuid_as_param()
```
pierallard, line 124:
```suggestion
    public function it_creates_a_product_without_identifier_nor_uuid_as_param()
```

src/Akeneo/Pim/Enrichment/Product/back/API/Command/UpsertProductCommand.php
```diff
@@ -46,7 +54,7 @@ public function __construct(
     /**
      * @param UserIntent[] $userIntents
      */
-    public static function createFromCollection(int $userId, string $productIdentifier, array $userIntents): self
+    public static function createFromCollection(int $userId, ProductIdentifier | ProductUuid | string | null $identifierOrUuid = null, array $userIntents = []): self
     {
         $valueUserIntents = [];
```
StevenVAIDIE, line 59: Should we made a :

```
    if (is_string($identifierOrUuid) {
         trigger_deprecation('Akeneo\Pim\Enrichment\Product\API', 'master', 'The string $identifierOrUuid is deprecated, use "ProductIdentifier" instead.');
    }
```

In order to handle this deprecation ?

StevenVAIDIE (commented):
:heavy_check_mark: ok about tailored import impact :clap:

src/Akeneo/Pim/Enrichment/Product/back/API/Command/UpsertProductCommand.php
```diff
@@ -46,7 +54,7 @@ public function __construct(
     /**
      * @param UserIntent[] $userIntents
      */
-    public static function createFromCollection(int $userId, string $productIdentifier, array $userIntents): self
+    public static function createFromCollection(int $userId, ProductIdentifier | ProductUuid | string | null $identifierOrUuid = null, array $userIntents = []): self
```
mmetayer, line 57: idea: could we: 
```php
    public static function createFromCollection(
        int $userId, 
        ProductIdentifier | string | null $productIdentifier = null,
        ProductUuid | null $productUuid = null,
        array $userIntents = []
    ): self {
        if (null !== $productIdentifier && null !== $productUuid) {
            throw new \InvalidArgumentException('only accepts one value for productIdentifier or productUUid');
        }
        ....
        return new self(
            userId: $userId,
            identifierOrUuid: $productIdentifier ?? $productUuid, 
            familyUserIntent: $familyUserIntent,
            categoryUserIntent: $categoryUserIntent,
            ...
        );
    }
```
This would allow keeping the same param name in the named constructor, and would decrease the amount of code in this PR (avoid to replace `productIdentifier` by `identifierOrUuid` every time it's called)

nmarniesse: I'm not fan having 2 separate parameters because the contract looks like less clear:
- what happen if we have both?
- what happen if we have none? do we create a new product? But how we can retrieve it by its uuid?

We discussed about that with Raccoons and Popeye and we agreed to do like that.

src/Akeneo/Pim/Enrichment/Component/Product/Builder/ProductBuilder.php
```diff
@@ -52,10 +52,14 @@ public function __construct(
     /**
      * {@inheritdoc}
      */
-    public function createProduct($identifier = null, $familyCode = null)
+    public function createProduct($identifier = null, $familyCode = null, $uuid = null)
     {
         $product = new $this->productClass();
 
+        if (null !== $uuid && $product->getUuid() !== $uuid->toString()) {
```
SamirBoulil, line 59:
```suggestion
        if (null !== $uuid) {
```
I'd always do it to keep the code simpler

src/Akeneo/Pim/Enrichment/Bundle/Controller/InternalApi/Product/UpdateProductController.php
```diff
@@ -175,9 +178,10 @@ private function updateProduct(ProductInterface $product, array $data): void
         $userIntents = $handledStamp->getResult();
 
         $userId = $this->userContext->getUser()?->getId();
+        $identifierCode = $this->attributeRepository->getIdentifierCode();
         $command = UpsertProductCommand::createFromCollection(
             $userId,
-            $product->getIdentifier() ?? '',
+            ProductIdentifier::fromAttributeCodeAndIdentifier($identifierCode, $product->getIdentifier()) ?? '',
```
nmarniesse, line 184:
```suggestion
            ProductUuid::fromUuid($product->getUuid()),
```

Using uuid is always better than using the identifier. And we don't need the AttributeRepository anymore

src/Akeneo/Pim/Enrichment/Component/Product/Model/AbstractProduct.php
```diff
@@ -94,6 +95,14 @@ public function getUuid(): UuidInterface
         return $this->uuid;
     }
 
+    public function setUuid(UuidInterface $uuid): ProductInterface
+    {
+        Assert::same($uuid->getVersion(), 4);
```
SamirBoulil, line 100: This check is an invariant (rule) of UuidInterface in my opinion not the AbstractProduct

src/Akeneo/Pim/Enrichment/Component/Product/Model/AbstractProduct.php
```diff
@@ -94,6 +95,14 @@ public function getUuid(): UuidInterface
         return $this->uuid;
     }
 
+    public function setUuid(UuidInterface $uuid): ProductInterface
```
nmarniesse, line 98: Please no. The uuid of a product must never change. We should find another way. 

I understand we have to do that in the case we want to create a product with a uuid. It it possible to add an optional parameter in the constructor instead?  
Maybe in this PR I would not allow to create a product with a uuid (so in the handler throw an exception if the product uuid does not exist). And allow to do that in another PR. My concern about that: what happen if we create a product manually with a uuid that already exist in the DB? The old product is lost? It can't happen with the service API but it can with the builder/updater/saver process

src/Akeneo/Pim/Enrichment/Product/back/API/Command/UpsertProductCommand.php
```diff
@@ -104,9 +112,9 @@ public function userId(): int
         return $this->userId;
     }
 
-    public function productIdentifier(): string
+    public function identifierOrUuid(): ProductIdentifier | ProductUuid | string | null
```
nmarniesse, line 115:
```suggestion
    public function identifierOrUuid(): ProductIdentifier | ProductUuid | null
```

If a string is provided at construct time, we can return a ProductIdentifier. Less conditions to make in the handler/validators/...

src/Akeneo/Pim/Enrichment/Product/back/API/Command/UpsertProductCommand.php
SamirBoulil: Discussed it a bit with Mathio IRL, (to me the clearer way is to have 2 named constructors `createWithIdentifier` and `createWithUuid` for instance).

We find Pierre M's version clearer for external user, someone who does not know about this command shouldn't have to guess if he needs to set the Identifier, the uuid or both. It should be clearer (even though it's possible to do it this way with named constructor)

src/Akeneo/Pim/Enrichment/Product/back/API/Command/UpsertProductCommand.php
```diff
@@ -26,11 +28,17 @@
...
+     * - a string: will be deprecated, related to the only attribute identifier
+     * - a ProductIdentifier: a pair attribute of the identifier code + the attribute code
+     * - a ProductUuid: the uuid of the existing product or the one you want to assign to the future product
+     * - null: the product will have a random uuid and no identifier
      * @param ValueUserIntent[] $valueUserIntents
      */
     public function __construct(
         private int $userId,
-        private string $productIdentifier,
+        private ProductIdentifier | ProductUuid | string | null $identifierOrUuid = null,
```
nmarniesse, line 41:
```suggestion
        private ProductIdentifier | ProductUuid | null $identifierOrUuid = null,
```

See my comment below to transform string in a ProductIdentifier

src/Akeneo/Pim/Enrichment/Product/back/API/ValueObject/ProductIdentifier.php
```diff
@@ -0,0 +1,35 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Akeneo\Pim\Enrichment\Product\API\ValueObject;
```
SamirBoulil, line 5: **Not in your PR**, but this namespace raises questions, ValueObject is a pattern used in the Domain layer to express concept of the domain. It is wierd to see this pattern used in an API :/

src/Akeneo/Pim/Enrichment/Product/back/Infrastructure/Query/SqlGetProductUuids.php
```diff
@@ -0,0 +1,42 @@
...
+
+        return false === $uuid ? null : Uuid::fromString($uuid);
+    }
+
+    public function fromIdentifiers(array $identifiers): ?array
+    {
+        $uuids = [];
+        foreach ($identifiers as $identifier)
+        {
+            $uuids[] = $this->fromIdentifier($identifier);
```
nmarniesse, line 37: - 1 query by identifier is bad. You can remember in production a simple query is between 5-10ms with the latency. So with 100 identifiers, it could cost 1 second to fetch all uuids! We must fetch all in one query.
- how do you know which uuid belong to which identifier? It would be probably better to return an associative array with identifiers as keys and uuids as values

src/Akeneo/Pim/Enrichment/Product/back/Test/Acceptance/Context/ProductContext.php
```diff
@@ -87,7 +87,7 @@ public function theUserUpsertsAProductWithTheIdentifier(string $username, string
      */
     public function anUnknownUserTriesToUpsertAProductWithTheIdentifier(string $identifier): void
     {
-        $command = new UpsertProductCommand(userId: -10, productIdentifier: $identifier);
+        $command = new UpsertProductCommand(userId: -10, identifierOrUuid: $identifier);
```
nmarniesse, line 90: It demonstrate we should use named argument careful. Here it's useless

nmarniesse (commented):
IMO this PR does too much thing. But ok with that if you arrive to review and merge it. Good effort!  

The setUuid on the Product is problematic for me.

src/Akeneo/Pim/Enrichment/Product/back/Application/UpsertProductHandler.php
```diff
@@ -50,11 +52,31 @@ public function __invoke(UpsertProductCommand $command): void
 
         $this->checkConsistencyWithConnectedUser($command->userId());
 
-        $product = $this->productRepository->findOneByIdentifier($command->productIdentifier());
+        $product = null;
+        if ($command->identifierOrUuid() instanceof ProductUuid) {
+            $product = $this->productRepository->find($command->identifierOrUuid()->uuid());
+            if (null !== $product) {
+                Assert::isInstanceOf($product, ProductInterface::class);
+            }
```
cemorin, line 60:
```suggestion
            Assert::nullOrIsInstanceOf($product, ProductInterface::class);

```

src/Akeneo/Pim/Enrichment/Product/back/Infrastructure/Validation/IsUserOwnerOfTheProductValidator.php
```diff
@@ -32,13 +35,16 @@ public function validate($command, Constraint $constraint): void
         Assert::isInstanceOf($command, UpsertProductCommand::class);
         Assert::isInstanceOf($constraint, IsUserOwnerOfTheProduct::class);
```
cemorin, line 37: don't check this if creation

PR #17571, opened 2022-08-02
RAB-934: migration add automation column in job_instance

MadeWilson:

add a new column to store automation configuration in job_instance table

- [ ] Tests
- [ ] Migration & Installer

### DIFF
--- a/upgrades/schema/Version_7_0_20220802151250_add_automation_column_in_job_instance.php
+++ b/upgrades/schema/Version_7_0_20220802151250_add_automation_column_in_job_instance.php
@@ -20,7 +20,10 @@ final class Version_7_0_20220802151250_add_automation_column_in_job_instance ext
 {
     public function up(Schema $schema): void
     {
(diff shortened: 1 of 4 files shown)


grena:
So we should add a second indexed column `scheduled`, right?

upgrades/schema/Version_7_0_20220802151250_add_automation_column_in_job_instance.php
```diff
@@ -0,0 +1,33 @@
...
+namespace Pim\Upgrade\Schema;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+
+final class Version_7_0_20220802151250_add_automation_column_in_job_instance extends AbstractMigration
+{
+    public function up(Schema $schema): void
+    {
+        $this->addSql(<<<SQL
```
StevenVAIDIE, line 23: The migration should be idempotent, we should check if column exist before adding column

ghoussard:
> So we should add a second indexed column `scheduled`, right?

It is not under `automation` key like in the Spike ?

grena:
Closed in favor of #17616

PR #17591, opened 2022-08-04
GRF-217: InternalAPI to standard format

etiennekerbastard:

Implementation of services for conversion of data from InternalAPI to standard format + tests

- [x] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc


src/Akeneo/Category/back/Application/Converter/FieldsRequirementChecker.php
```diff
@@ -0,0 +1,70 @@
...
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+class FieldsRequirementChecker
+{
+    /**
+     * Check if all required fields are present.
+     * @param array<string> $item
+     * @param array<string, mixed> $fields
+     */
+    public function checkFieldsPresence(array $item, array $fields): void
```
antoinetrouve, line 21:
```suggestion
    public function checkFieldsExist(array $item, array $fields): void
```

src/Akeneo/Category/back/Infrastructure/Converter/InternalAPI/InternalAPIToStd.php
```diff
@@ -0,0 +1,32 @@
...
+namespace Akeneo\Category\Infrastructure\Converter\InternalAPI;
+
+use Akeneo\Category\Application\Converter\ConverterInterface;
+use Akeneo\Category\Application\Converter\FieldsRequirementChecker;
+
+/**
+ * @copyright 2022 Akeneo SAS (https://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+class InternalAPIToStd implements ConverterInterface
```
antoinetrouve, line 14:
```suggestion
class InternalApiToStd implements ConverterInterface
```

src/Akeneo/Category/back/Infrastructure/Converter/InternalAPI/InternalAPIToStd.php
```diff
@@ -0,0 +1,32 @@
...
+{
+    public function __construct(private FieldsRequirementChecker $fieldsRequirementChecker)
```
antoinetrouve, line 16:
```suggestion
    public function __construct(private FieldsRequirementChecker $fieldsChecker)
```

src/Akeneo/Category/back/Infrastructure/Converter/InternalAPI/InternalAPIToStd.php
```diff
@@ -0,0 +1,32 @@
...
+    {
+    }
+
+    public function convert(array $data): array
+    {
+        $this->fieldsRequirementChecker->checkFieldsPresence(['code', 'labels'], $data['properties']);
```
antoinetrouve, line 22: It could be a thing like this to be more readable : 
```suggestion
        $this->fieldsChecker->checkExist(['code', 'labels'], $data['properties']);
```

antoinetrouve (approved):

src/Akeneo/Category/back/Infrastructure/Converter/InternalAPI/InternalAPIToStd.php
```diff
@@ -0,0 +1,32 @@
...
+        $this->fieldsRequirementChecker->checkFieldsNotEmpty(['code'], $data['properties']);
+
+        $convertedData = [];
+        $convertedData['code'] = $data['properties']['code'];
+        $convertedData['parent'] = null;
```
olivierpontier, line 27: Be careful, a category without a parent (set to NULL) means the category is a root. Will this property be use during the update of the category?

etiennekerbastard: Indeed, I had been conflicted about this as well. 

Currently [standard format for updating a category](https://github.com/akeneo/pim-enterprise-dev/blob/master/internal_doc/standard_format.md#category) has a key "parent" which represents the code of the category's parent or if the category has no parent (root category) it is set to null.

So I had initially implemented it so that "parent" value would be equal to the value passed by the HTTP request. 

Problem is: in the HTTP request the parent information is never included so in default format I ended up having it defaulting to null.

I can put it back like I initially implemented it but we will have to remember to add the parent information in the HTTP request. What do you think ?

src/Akeneo/Category/back/Infrastructure/Converter/InternalAPI/InternalAPIToStd.php
etiennekerbastard: In this class' namespace we use InternalAPI

src/Akeneo/Category/back/Infrastructure/Converter/InternalAPI/InternalAPIToStd.php
etiennekerbastard: I like it better my way :) Also your suggestion contradicts another one of your suggestions above (checkFieldsExist)

src/Akeneo/Category/back/Infrastructure/Converter/InternalAPI/InternalAPIToStd.php
olivierpontier: Thanks for your answer. Are we agree that the `parent` property is not covered by the use case of the Category edition, so far? The set of the parent will be covered during the creation of a category or the reorganization of the tree structure. So, could we ignore the `parent` property until we are treating those new use cases?

etiennekerbastard: Yes, setting the parent would only happen at category creation or when moving categories around in the tree. I think that we can leave it defaulting to null in standard format for now as it will not be read by the service converting std format to user intents in the update end point.

etiennekerbastard: Or just leave it out entirely until we start dealing with category creation or moving categories around, at a later stage, if that's what you meant when you said "ignore the parent property", I'm not sure :) let's have a quick call maybe and make sure we're on the same page ?

olivierpontier: After a quick call, we decided we will ignore the parent property until we start dealing with category creation or moving categories around, at a later stage

olivierpontier (approved):

PR #17594, opened 2022-08-05
PIM-10571: Fix infinite scroll of attribute group selector in family edit form

ManuelActisCassa:
Before fix, not all attribute groups were shown in the infinite scroll of attribute group selector in family edit form because empty family groups were not returned by the item fetcher, thus not returning as many options as results per page, which made the function think there were no more options. At the end, this caused the "more" condition to not be triggered.

The fix allows to filter the loaded items while preserving all the items found before filtering, making it possible for the function to know that there still are options to look for.

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/src/Akeneo/Platform/Bundle/UIBundle/Resources/public/js/form/common/add-select/select.js
+++ b/src/Akeneo/Platform/Bundle/UIBundle/Resources/public/js/form/common/add-select/select.js
@@ -214,15 +214,17 @@ define([
      * @return {Object}
      */
     prepareChoices: function (items) {
-      return _.chain(
-        _.sortBy(items, function (item) {
-          return item.sort_order;
-        })
-      )
-        .map(function (item) {
-          return ChoicesFormatter.formatOne(item);
-        })
-        .value();
+      return 0 === Object.keys(items).length
+        ? items
+        : _.chain(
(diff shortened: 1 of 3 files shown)


mmetayer (approved):

pierallard (approved):

src/Akeneo/Platform/Bundle/UIBundle/Resources/public/js/form/common/add-select/select.js
```diff
@@ -214,15 +214,17 @@ define([
...
-      return _.chain(
-        _.sortBy(items, function (item) {
-          return item.sort_order;
-        })
-      )
-        .map(function (item) {
-          return ChoicesFormatter.formatOne(item);
-        })
-        .value();
+      return 0 === Object.keys(items).length
```
ManuelActisCassa, line 217: Early return when items is an empty object else this method return an empty array.

mmetayer:
green build: https://app.circleci.com/pipelines/github/akeneo/pim-community-dev/38887/workflows/497e42f7-0d87-4261-80b7-09ede26bb5b1

PR #17616, opened 2022-08-09
RAB-945: migration DB

MadeWilson:
add a new column to store automation configuration in job_instance table, and an indexed column to check if a job_instance is automated or not. Since we have client with more than 1000+ entry in the table scheduled boolean column is indexed.

- [x] Tests
- [x] Migration & Installer

### DIFF
--- a/upgrades/schema/Version_7_0_20220802151250_add_automation_column_in_job_instance.php
+++ b/upgrades/schema/Version_7_0_20220802151250_add_automation_column_in_job_instance.php
@@ -20,7 +20,10 @@ final class Version_7_0_20220802151250_add_automation_column_in_job_instance ext
 {
     public function up(Schema $schema): void
     {
-        $this->addSql('ALTER TABLE akeneo_batch_job_execution ADD COLUMN automation JSON DEFAULT NULL');
+        $this->addSql(<<<SQL
+        ALTER TABLE akeneo_batch_job_instance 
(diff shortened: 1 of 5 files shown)


upgrades/schema/Version_7_0_20220802151250_add_automation_column_in_job_instance.php
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    private function isColumnAlreadyCreated(string $columnName): bool
+    {
+        $sql = <<<SQL
+            SHOW COLUMNS FROM akeneo_batch_job_instance LIKE :columnName;
+        SQL;
+        $connection = $this->container->get('database_connection');
+        $result = $connection->executeQuery($sql, ['columnName' => $columnName])->fetch();
+        if (!empty($result)) {
```
grena, line 73: You can directly return `!empty($result)` if you want

grena (approved):
:+1: 

As there is a DB migration, don't forget to give some information about metrics: 
- what's the max number of job instances in production

upgrades/schema/Version_7_0_20220802151250_add_automation_column_in_job_instance.php
MadeWilson: your god damn right !

ghoussard (approved):

ghoussard:
refresh

PR #17639, opened 2022-08-10
CXP-1245: allow Julia to add a criterion based on an attribute

tseho:

This PR only includes the possibility to search and add attributes criteria.
The first attribute is not done yet, it's for the next PR.

https://user-images.githubusercontent.com/1421130/183946655-1c9d5d4c-7728-4a34-9d0e-bcda01370372.mp4

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/components/catalogs/back/src/Infrastructure/Controller/Internal/GetAttributeAction.php
+++ b/components/catalogs/back/src/Infrastructure/Controller/Internal/GetAttributeAction.php
@@ -1,0 +1,39 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Akeneo\Catalogs\Infrastructure\Controller\Internal;
+
+use Akeneo\Catalogs\Infrastructure\Persistence\FindOneAttributeByCodeQuery;
+use Symfony\Component\HttpFoundation\JsonResponse;
+use Symfony\Component\HttpFoundation\RedirectResponse;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
+
+/**
+ * @copyright 2022 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+class GetAttributeAction
+{
+    public function __construct(
+        private FindOneAttributeByCodeQuery $findOneAttributeByCode,
+    ) {
+    }
+
+    public function __invoke(Request $request, string $code): Response
+    {
+        if (!$request->isXmlHttpRequest()) {
+            return new RedirectResponse('/');
+        }
+
+        $attribute = $this->findOneAttributeByCode->execute($code);
+
+        if (null === $attribute) {
+            throw new NotFoundHttpException(\sprintf('$attribute "%s" does not exist.', $code));
+        }
+
+        return new JsonResponse($attribute);
+    }
+}
--- a/components/catalogs/back/src/Infrastructure/Controller/Internal/GetAttributeAction.php
+++ b/components/catalogs/back/src/Infrastructure/Controller/Internal/GetAttributeAction.php
@@ -4,7 +4,7 @@ declare(strict_types=1);
 
 namespace Akeneo\Catalogs\Infrastructure\Controller\Internal;
 
-use Akeneo\Catalogs\Infrastructure\Persistence\FindOneAttributeByCodeQuery;
+use Akeneo\Catalogs\Application\Persistence\FindOneAttributeByCodeQueryInterface;
 use Symfony\Component\HttpFoundation\JsonResponse;
 use Symfony\Component\HttpFoundation\RedirectResponse;
 use Symfony\Component\HttpFoundation\Request;
@@ -18,7 +18,7 @@ use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
 class GetAttributeAction
 {
     public function __construct(
-        private FindOneAttributeByCodeQuery $findOneAttributeByCode,
+        private FindOneAttributeByCodeQueryInterface $findOneAttributeByCode,
     ) {
     }
 
--- a/components/catalogs/back/src/Infrastructure/Persistence/FindOneAttributeByCodeQuery.php
+++ b/components/catalogs/back/src/Infrastructure/Persistence/FindOneAttributeByCodeQuery.php
@@ -25,7 +25,7 @@ final class FindOneAttributeByCodeQuery
         /** @var AttributeInterface|null $attribute */
         $attribute = $this->repository->findOneByIdentifier($code);
 
-        if (!$attribute) {
+        if (null === $attribute) {
             return null;
         }
 
--- a/components/catalogs/back/src/Infrastructure/Persistence/FindOneAttributeByCodeQuery.php
+++ b/components/catalogs/back/src/Infrastructure/Persistence/FindOneAttributeByCodeQuery.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Akeneo\Catalogs\Infrastructure\Persistence;
 
+use Akeneo\Catalogs\Application\Persistence\FindOneAttributeByCodeQueryInterface;
 use Akeneo\Pim\Structure\Component\Model\AttributeInterface;
 use Akeneo\Pim\Structure\Component\Repository\AttributeRepositoryInterface;
 
@@ -11,7 +12,7 @@ use Akeneo\Pim\Structure\Component\Repository\AttributeRepositoryInterface;
  * @copyright 2022 Akeneo SAS (http://www.akeneo.com)
  * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
  */
-final class FindOneAttributeByCodeQuery
+final class FindOneAttributeByCodeQuery implements FindOneAttributeByCodeQueryInterface
 {
     public function __construct(private AttributeRepositoryInterface $repository)
     {
--- a/components/catalogs/back/src/Infrastructure/Persistence/FindOneAttributeByCodeQuery.php
+++ b/components/catalogs/back/src/Infrastructure/Persistence/FindOneAttributeByCodeQuery.php
@@ -1,0 +1,40 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Akeneo\Catalogs\Infrastructure\Persistence;
+
+use Akeneo\Pim\Structure\Component\Model\AttributeInterface;
+use Akeneo\Pim\Structure\Component\Repository\AttributeRepositoryInterface;
+
+/**
+ * @copyright 2022 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+final class FindOneAttributeByCodeQuery
+{
+    public function __construct(private AttributeRepositoryInterface $repository)
+    {
+    }
+
+    /**
+     * @return array{code: string, label: string, type: string, scopable: bool, localizable: bool}
+     */
+    public function execute(string $code): ?array
+    {
+        /** @var AttributeInterface|null $attribute */
+        $attribute = $this->repository->findOneByIdentifier($code);
+
+        if (!$attribute) {
+            return null;
+        }
+
+        return [
+            'code' => $attribute->getCode(),
+            'label' => $attribute->getLabel(),
+            'type' => $attribute->getType(),
+            'scopable' => $attribute->isScopable(),
+            'localizable' => $attribute->isLocalizable(),
+        ];
+    }
+}
--- a/components/catalogs/back/src/Infrastructure/Persistence/SearchAttributesQuery.php
+++ b/components/catalogs/back/src/Infrastructure/Persistence/SearchAttributesQuery.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Akeneo\Catalogs\Infrastructure\Persistence;
 
+use Akeneo\Catalogs\Application\Persistence\SearchAttributesQueryInterface;
 use Akeneo\Pim\Structure\Component\Model\AttributeInterface;
 use Akeneo\Tool\Component\StorageUtils\Repository\SearchableRepositoryInterface;
 
@@ -11,7 +12,7 @@ use Akeneo\Tool\Component\StorageUtils\Repository\SearchableRepositoryInterface;
  * @copyright 2022 Akeneo SAS (http://www.akeneo.com)
  * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
  */
-class SearchAttributesQuery
+final class SearchAttributesQuery implements SearchAttributesQueryInterface
 {
     private const ALLOWED_TYPES = [
         'pim_catalog_text',
--- a/components/catalogs/back/src/Infrastructure/Persistence/SearchAttributesQuery.php
+++ b/components/catalogs/back/src/Infrastructure/Persistence/SearchAttributesQuery.php
@@ -1,0 +1,50 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Akeneo\Catalogs\Infrastructure\Persistence;
+
+use Akeneo\Pim\Structure\Component\Model\AttributeInterface;
+use Akeneo\Tool\Component\StorageUtils\Repository\SearchableRepositoryInterface;
+
+/**
+ * @copyright 2022 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+class SearchAttributesQuery
+{
+    private const ALLOWED_TYPES = [
+        'pim_catalog_text',
+    ];
+
+    public function __construct(
+        private SearchableRepositoryInterface $searchableAttributeRepository,
+    ) {
+    }
+
+    /**
+     * @return array<array{code: string, label: string, type: string, scopable: bool, localizable: bool}>
+     */
+    public function execute(?string $search = null, int $page = 1, int $limit = 20): array
+    {
+        $attributes = $this->searchableAttributeRepository->findBySearch(
+            $search,
+            [
+                'limit' => $limit,
+                'page' => $page,
+                'types' => self::ALLOWED_TYPES,
+            ],
+        );
+
+        return \array_map(
+            static fn (AttributeInterface $attribute) => [
+                'code' => $attribute->getCode(),
+                'label' => $attribute->getLabel(),
+                'type' => $attribute->getType(),
+                'scopable' => $attribute->isScopable(),
+                'localizable' => $attribute->isLocalizable(),
+            ],
+            $attributes
+        );
+    }
+}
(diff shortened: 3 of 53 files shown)


components/catalogs/back/src/Infrastructure/Persistence/FindOneAttributeByCodeQuery.php
```diff
@@ -0,0 +1,40 @@
...
+
+    /**
+     * @return array{code: string, label: string, type: string, scopable: bool, localizable: bool}
+     */
+    public function execute(string $code): ?array
+    {
+        /** @var AttributeInterface|null $attribute */
+        $attribute = $this->repository->findOneByIdentifier($code);
+
+        if (!$attribute) {
```
masacc, line 28:
```suggestion
        if (null === $attribute) {
```
I rather want to be explicit

components/catalogs/back/src/Infrastructure/Persistence/SearchAttributesQuery.php
```diff
@@ -0,0 +1,50 @@
...
+        );
+
+        return \array_map(
+            static fn (AttributeInterface $attribute) => [
+                'code' => $attribute->getCode(),
+                'label' => $attribute->getLabel(),
+                'type' => $attribute->getType(),
+                'scopable' => $attribute->isScopable(),
+                'localizable' => $attribute->isLocalizable(),
+            ],
```
masacc, line 46: Should we think about creating our normalizers? (I think about it for Channel and Locale)

components/catalogs/back/src/Infrastructure/Symfony/Resources/config/routing/internal.yml
```diff
@@ -21,3 +21,13 @@ akeneo_catalogs_internal_get_families:
...
     controller: Akeneo\Catalogs\Infrastructure\Controller\Internal\GetFamiliesAction
     methods: [ GET ]
+
+akeneo_catalogs_internal_get_attributes:
+    path: '/rest/catalogs/attributes'
+    controller: Akeneo\Catalogs\Infrastructure\Controller\Internal\GetAttributesAction
+    methods: [ GET ]
+
+akeneo_catalogs_internal_get_attribute:
+    path: '/rest/catalogs/attributes/{code}'
```
masacc, line 31: I use to write more precise routes first

components/catalogs/back/src/Infrastructure/Symfony/Resources/config/services.yml
```diff
@@ -75,3 +75,10 @@ services:
     Akeneo\Catalogs\Infrastructure\Persistence\GetFamiliesByCodeQuery:
         arguments:
             - '@pim_enrich.repository.family.search'
+
+    Akeneo\Catalogs\Infrastructure\Persistence\SearchAttributesQuery:
+        arguments:
+            - '@pim_enrich.repository.attribute.search'
+
+    Akeneo\Pim\Structure\Component\Repository\AttributeRepositoryInterface:
+        alias: 'pim_catalog.repository.attribute'
```
masacc, line 84: not sure this should be in catalogs

components/catalogs/back/tests/Integration/Infrastructure/Controller/Internal/GetAttributeActionTest.php
```diff
@@ -0,0 +1,54 @@
...
+declare(strict_types=1);
+
+namespace Akeneo\Catalogs\Test\Integration\Infrastructure\Controller\Internal;
+
+use Akeneo\Catalogs\Test\Integration\IntegrationTestCase;
+use Doctrine\DBAL\Connection;
+use PHPUnit\Framework\Assert;
+
+/**
+ * @covers \Akeneo\Catalogs\Infrastructure\Controller\Internal\GetAttributeAction
```
masacc, line 12: why do you need to specify this?

components/catalogs/back/tests/Integration/Infrastructure/Persistence/SearchAttributesQueryTest.php
```diff
@@ -0,0 +1,95 @@
...
+use Akeneo\Catalogs\Infrastructure\Persistence\SearchAttributesQuery;
+use Akeneo\Catalogs\Test\Integration\IntegrationTestCase;
+
+/**
+ * @copyright 2022 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ *
+ * @covers \Akeneo\Catalogs\Infrastructure\Persistence\SearchAttributesQuery
+ */
+class SearchAttributesQueryTest extends IntegrationTestCase
```
masacc, line 16: You don't test the search feature in this file. Is it expected?

components/catalogs/back/tests/Integration/IntegrationTestCase.php
```diff
@@ -213,4 +214,11 @@ protected function enableCatalog(string $id): void
             ]
         );
     }
+
+    protected function createAttribute(array $data): void
+    {
+        $attribute = self::getContainer()->get('pim_catalog.factory.attribute')->create();
+        self::getContainer()->get('pim_catalog.updater.attribute')->update($attribute, $data);
+        self::getContainer()->get('pim_catalog.saver.attribute')->save($attribute);
+    }
```
masacc, line 223: I made a more complicated one in my [PR](https://github.com/akeneo/pim-community-dev/pull/17533/files#diff-52ed2af996b4a37534b351b5e6ad90a32409fbbc93f1dbaa1eeb540026009542R276). I prefer your because I think it's enough but I inspire me from what we've done before. Your thoughts ?
```    protected function createAttribute(array $attributeData): void
    {
        /** @var SimpleFactoryInterface $attributeFactory */
        $attributeFactory = self::getContainer()->get('pim_catalog.factory.attribute');
        /** @var AttributeInterface $attribute */
        $attribute = $attributeFactory->create();

        /** @var ObjectUpdaterInterface $attributeUpdater */
        $attributeUpdater = self::getContainer()->get('pim_catalog.updater.attribute');
        $attributeUpdater->update($attribute, $attributeData);

        /** @var ValidatorInterface $validator */
        $validator = self::getContainer()->get('validator');
        $violations = $validator->validate($attribute);
        self::assertSame(0, $violations->count(), (string) $violations);

        /** @var AttributeSaver $attributeSaver */
        $attributeSaver = self::getContainer()->get('pim_catalog.saver.attribute');
        $attributeSaver->save($attribute);
    }
```

components/catalogs/front/src/components/ProductSelection/hooks/useFindCriterionByField.ts
```diff
@@ -2,18 +2,48 @@ import {useCallback} from 'react';
...
+                    ['attribute', field],
+                    async () => {
+                        const response = await fetch(`/rest/catalogs/attributes/${field}`, {
+                            headers: {
+                                'X-Requested-With': 'XMLHttpRequest',
+                            },
+                        });
+
+                        return await response.json();
+                    },
```
masacc, line 35: is it a bad practice to call another hook (`useAttribute`) in this hook?

components/catalogs/front/src/components/ProductSelection/hooks/useInfiniteAttributeCriterionFactories.unit.ts
```diff
@@ -0,0 +1,193 @@
...
+
+import {act, renderHook} from '@testing-library/react-hooks';
+import fetchMock from 'jest-fetch-mock';
+import {mocked} from 'ts-jest/utils';
+import {Attribute} from '../models/Attribute';
+import {ReactQueryWrapper} from '../../../../tests/ReactQueryWrapper';
+import {useInfiniteAttributeCriterionFactories} from './useInfiniteAttributeCriterionFactories';
+import {useFindAttributeCriterionByType} from './useFindAttributeCriterionByType';
+
+test('it fetch attributes & paginate criterion factories', async () => {
```
masacc, line 11:
```suggestion
test('it fetches attributes & paginates criterion factories', async () => {
```

components/catalogs/front/src/components/ProductSelection/hooks/useInfiniteAttributeCriterionFactories.unit.ts
```diff
@@ -0,0 +1,193 @@
...
+                factory: expect.any(Function),
+            },
+        ],
+        error: null,
+        hasNextPage: false,
+        fetchNextPage: expect.any(Function),
+    });
+});
+
+test('it return a custom factory with the attribute code as field', async () => {
```
masacc, line 101:
```suggestion
test('it returns a custom factory with the attribute code as field', async () => {
```

components/catalogs/back/src/Infrastructure/Persistence/SearchAttributesQuery.php
tseho: I had the same reflection in mind when I did it. I dunno, we should probably do it.

components/catalogs/back/src/Infrastructure/Symfony/Resources/config/services.yml
tseho: The alias or using directly this service ?

components/catalogs/back/tests/Integration/Infrastructure/Controller/Internal/GetAttributeActionTest.php
tseho: I don't need to, but I want to.
It tells phpunit that this test will cover the class `GetAttributeAction` and only this one. Other classes, like the query, should not be considered covered by this test, even if they are executed.

Why ? because if we don't do this, the query could have coverage of 100% because it's used in the controller. When looking at the coverage report, we could think, "wow, this query is already covered, no need to test it". <= It's what happened with the families queries.
And then, 2 months later, you change the controller and it uses another query.
Your first query is still not explicitly tested and it's not covered anymore.

components/catalogs/back/tests/Integration/Infrastructure/Persistence/SearchAttributesQueryTest.php
tseho: Good catch, forgot to do it.

components/catalogs/back/tests/Integration/IntegrationTestCase.php
tseho: I'm always in favor for the simple.
Hopefully, one day, we will have a nice ServiceAPI on this.

components/catalogs/back/src/Infrastructure/Symfony/Resources/config/services.yml
masacc: the alias

components/catalogs/back/tests/Integration/Infrastructure/Controller/Internal/GetAttributeActionTest.php
masacc: ok good idea

components/catalogs/back/src/Infrastructure/Persistence/SearchAttributesQuery.php
tseho: Ok, here is the issue.
I think that our queries, since there is no real ServiceAPI yet, serve as an anti-corruption layer.
Here, we normalize the doctrine entity returned into our standard/domain representation of this object.

Moving the normalization outside the query, in a dedicated normalizer for example, would really leak this doctrine entity in several place in our code.

What's interesting is that we should think how we want to handle those anti-corruption layers.

Meanwhile, I can't think of a simple & better way of doing what we do.
It does not worry me, because the queries are integration-tested correctly, we can safely change the services that are called inside with confidence.

components/catalogs/front/src/components/ProductSelection/hooks/useFindCriterionByField.ts
tseho: No, it's a "bad practice", it's technically impossible :D

In React, a Component (or a hook), must always call the same number of hooks, in the same order during its initialization.
You cannot defer a hook or conditionally call it.

```
if (code !== undefined) {
 const attribute = useAttribute(code);
}
```
:arrow_up: :red_circle: this is forbidden !

This is the exact situation I find myself here. I only want to fetch an attribute if the field did not match known ones.
So, no hook for me !
Yes, I have to duplicate the fetch, and that's bad.

I though about separating the fetch methods from the hooks to reuse, I don't know if it's worth it. Thoughts ?

masacc: Yes this was what I think about when I asked the question but I saw that you use `client.fetchQuery` that will use the same cache as in `useAttribute` so it's ok for me.

components/catalogs/back/src/Infrastructure/Persistence/SearchAttributesQuery.php
masacc: ah yes I remember we talked about it before my holidays ;-)

tseho: So, ok to stay as it is for now ?

masacc: yep

components/catalogs/back/src/Infrastructure/Controller/Internal/GetAttributeAction.php
```diff
@@ -0,0 +1,39 @@
...
+use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
+
+/**
+ * @copyright 2022 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+class GetAttributeAction
+{
+    public function __construct(
+        private FindOneAttributeByCodeQuery $findOneAttributeByCode,
```
RiffVentura, line 21: Shouldn't it be an interface? (same comment for other action)

components/catalogs/back/src/Infrastructure/Persistence/SearchAttributesQuery.php
```diff
@@ -0,0 +1,50 @@
...
+namespace Akeneo\Catalogs\Infrastructure\Persistence;
+
+use Akeneo\Pim\Structure\Component\Model\AttributeInterface;
+use Akeneo\Tool\Component\StorageUtils\Repository\SearchableRepositoryInterface;
+
+/**
+ * @copyright 2022 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+class SearchAttributesQuery
```
RiffVentura, line 14:
```suggestion
final class SearchAttributesQuery
```

components/catalogs/back/tests/Integration/Infrastructure/Controller/Internal/GetAttributeActionTest.php
```diff
@@ -0,0 +1,54 @@
...
+                'HTTP_X-Requested-With' => 'XMLHttpRequest',
+            ],
+        );
+
+        $response = $client->getResponse();
+        Assert::assertEquals(200, $response->getStatusCode());
+
+        $attribute = \json_decode($response->getContent(), true);
+        Assert::assertSame('name', $attribute['code']);
+        Assert::assertArrayHasKey('code', $attribute);
```
RiffVentura, line 48: Is it on purpose to query keys individually instead of a whole object comparison like it is done in the query integration test?

components/catalogs/front/src/components/ProductSelection/hooks/useAttribute.ts
```diff
@@ -0,0 +1,28 @@
...
+            const response = await fetch(`/rest/catalogs/attributes/${code}`, {
+                headers: {
+                    'X-Requested-With': 'XMLHttpRequest',
+                },
+            });
+
+            return await response.json();
+        },
+        {
+            staleTime: 60,
```
RiffVentura, line 25: Do you think we should also specify `staleTime`  in future queries? Or is it specific to attributes ?

components/catalogs/front/src/components/ProductSelection/hooks/useAttribute.ts
```diff
@@ -0,0 +1,28 @@
+import {useQuery} from 'react-query';
+import {Attribute} from '../models/Attribute';
+
+type Error = string | null;
+type Result = {
```
RiffVentura, line 5: I've been wondering if we should make a more robust  Result type for the queries.
Since useQuery assures us of data is undefined on loading or error
But when loading is false, typescript request to check data as well anyway
So I had something like that in mind:
```typescript
type SuccessResult = {
  isLoading: false;
  isError: false;
  data: SomeType;
  error: null
};

type LoadingResult = {
  isLoading: true;
  isError: false;
  data: undefined;
  error: null
};
type ErrorResult = {
  isLoading: false;
  isError: true;
  data: undefined;
  error: Error
};
type Result = SuccessResult |  LoadingResult | ErrorResult
```
This would help type inference I believe
What do you think ?

RiffVentura (approved):

components/catalogs/front/src/components/ProductSelection/hooks/useAttribute.ts
tseho: Already tried. We cannot. It's incompatible with react-query internal types.

components/catalogs/front/src/components/ProductSelection/hooks/useAttribute.ts
tseho: IMHO, it should be case by case.

Here, it was valuable.
First, an attribute does not change often.
The risks of having it changing by another user, during the cache duration, are small ... and I'm not sure it would affect us, you cannot move an attribute from non-scopable to scopable for example, you have to remove it first.

Second, I did it here because this query is called 2 times, from 2 places, in a really short burst when you add an attribute criterion.
The performance benefice was immediate and not hypothetical.

components/catalogs/back/tests/Integration/Infrastructure/Controller/Internal/GetAttributeActionTest.php
tseho: Yes.
For me, it's 2 different layers with different expectations.

On the query/persistence layer, I want to be stricter on what's returned, I want to know when it changes and when it breaks.
Especially since it behave like a anti-corruption layer with external services.

On the controller layer, meh, I have looser expectations, I already know that the persistence layer is correctly tested, I just want to make sure that I'm returning the structure expected by the frontend.

Anyway, I will slightly improve this test to reduce the number of asserts.

components/catalogs/back/src/Infrastructure/Controller/Internal/GetAttributeAction.php
tseho: I will update it because we done it in other places (not with families tho).

I think it's useless as it is, because in the hexagonal pattern, you put interfaces into your layers to create ports & adapters.
Like, you are writing business code into your application layer, you only refer to the interfaces of the persistence to prevent yourself to couple your business code to the infra code.

Here, we have an infrastructure class (the query) used only in infrastructure (the controller), we don't use it in the application.

What's the point ?

And I'm not sure that that the Dependency Inversion Principle, we are using concrete implementations of other modules, only our concrete implem, in the same layer of the same module...

Anyway, will update it to stay consistent.

components/catalogs/front/src/components/ProductSelection/hooks/useAttribute.ts
RiffVentura: bummer :disappointed:

components/catalogs/back/src/Infrastructure/Controller/Internal/GetAttributeAction.php
RiffVentura: My point was  about consistency with others
I agree that it brings nothing here (Makes me wonder if we should refrain from doing it  now while it is manageable to revert)

masacc (approved):

PR #17658, opened 2022-08-12
CXP-1236 Added category criterion (Back)

RiffVentura:
- Added endpoint to fetch categories by code
- Added endpoint to fetch categories that roots
- Added endpoint to fetch category children
- Updated coupling detector config with dependencies used by the queries


components/catalogs/back/tests/Integration/IntegrationTestCase.php
```diff
@@ -213,4 +214,23 @@ protected function enableCatalog(string $id): void
             ]
         );
     }
+
+    protected function createCategory(array $data = []): CategoryInterface
+    {
+        $category = self::getContainer()->get('pim_catalog.factory.category')->create();
+        self::getContainer()->get('pim_catalog.updater.category')->update($category, $data);
+        self::getContainer()->get('validator')->validate($category);
+        self::getContainer()->get('pim_catalog.saver.category')->save($category);
```
masacc, line 223: I like to add `/** @var */` to see which coupling we have (even in the tests).

And also, do you think the validation step is mandatory here?

components/catalogs/back/tests/Integration/Infrastructure/Controller/Internal/GetCategoriesActionTest.php
```diff
@@ -0,0 +1,85 @@
...
+namespace Akeneo\Catalogs\Test\Integration\Infrastructure\Controller\Internal;
+
+use Akeneo\Catalogs\Test\Integration\IntegrationTestCase;
+use PHPUnit\Framework\Assert;
+
+/**
+ * @copyright 2022 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+class GetCategoriesActionTest extends IntegrationTestCase
```
masacc, line 14: add `@covers` comment like explained in this discussion: https://github.com/akeneo/pim-community-dev/pull/17639#discussion_r943318774

masacc (approved):

components/catalogs/back/tests/Integration/IntegrationTestCase.php
tseho: > And also, do you think the validation step is mandatory here?
Agree with Thomas, the validator is not required.

guillaumefer-akeneo (approved):

components/catalogs/back/tests/Integration/Infrastructure/Controller/Internal/GetCategoriesActionTest.php
```diff
@@ -0,0 +1,85 @@
...
+                'HTTP_X-Requested-With' => 'XMLHttpRequest',
+            ],
+        );
+
+        $response = $client->getResponse();
+        Assert::assertEquals(200, $response->getStatusCode());
+
+        $categories = \json_decode($response->getContent(), true);
+
+        Assert::assertEmpty($categories, 'No categories should be returned');
```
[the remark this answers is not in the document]
tseho: Having an endpoint `/categories` that can be fitlered with `/categories?codes=` is ok.

But when you call `/categories`, having it returns 200 OK with an empty array, because it calls the method `$this->categoryRepository->getCategoriesByCodes([]);`, it's not ok.

I agree with masacc's conclusion, the default behavior, without query parameters, is not implemented because we don't need it, and that's ok for an internal endpoint, but it should throw.

components/catalogs/back/src/Infrastructure/Persistence/GetCategoryTreeRootsQuery.php
```diff
@@ -0,0 +1,54 @@
...
+use Akeneo\UserManagement\Bundle\Context\UserContext;
+
+/**
+ * @copyright 2022 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+class GetCategoryTreeRootsQuery implements GetCategoryTreeRootsQueryInterface
+{
+    public function __construct(
+        private UserContext $userContext,
```
tseho, line 19: I think that we should not use this dependency, it's a really legacy one in the back that creates "magic" in the code.
Why magic ? Because I don't expect a query in the persistence layer to returns different results based on a hidden context I have no control over.

Instead, we should do something like in the controller:
`$locale = $request->get('locale', 'en_US');`

Then we call the endpoint with the user locale already available in the front, and pass it to the query.

components/catalogs/back/.php_cd.php
```diff
@@ -69,6 +69,12 @@
             'Akeneo\Tool\Bundle\ElasticsearchBundle\Client',
             'Akeneo\Tool\Component\StorageUtils\Cursor\CursorFactoryInterface',
             'Symfony\Component\OptionsResolver',
+
+            'Akeneo\UserManagement\Bundle\Context\UserContext',
```
tseho, line 73:
```suggestion
```
see the other comment about `UserContext` for the reason.

components/catalogs/back/src/Infrastructure/Persistence/GetCategoryTreeRootsQuery.php
```diff
@@ -0,0 +1,54 @@
...
+    private function normalizeCategoryTree(CategoryTree $categoryTree, string $locale): array
+    {
+        /** @var array{id: int, code: string, labels: array<string, string>} $normalizedTree */
+        $normalizedTree = $categoryTree->normalize();
+
+        $normalizedTree['isLeaf'] = false;
+        $normalizedTree['label'] = $normalizedTree['labels'][$locale] ?? "[{$normalizedTree['code']}]";
+        unset($normalizedTree['labels']);
+
+        return $normalizedTree;
```
tseho, line 52:
```suggestion
        return [
          'id' => $categoryTree->id,
          'code' => $categoryTree->code,
          'label' => $categoryTree->labels[$locale] ?? "[{$categoryTree->code}]",
          'isLeaf' => false,
        ];
```
It can really simplify your code and we will have less surprises if the CategoryTree::normalize evolves.

Also, do you really need to expose the id since you have the code ?

components/catalogs/back/src/Infrastructure/Symfony/Resources/config/routing/internal.yml
```diff
@@ -21,3 +21,20 @@ akeneo_catalogs_internal_get_families:
...
+
+akeneo_catalogs_internal_get_categories:
+    path: '/rest/catalogs/categories'
+    controller: Akeneo\Catalogs\Infrastructure\Controller\Internal\GetCategoriesAction
+    methods: [ GET ]
+
+akeneo_catalogs_internal_get_category_tree_roots:
+    path: '/rest/catalogs/categories/tree-roots'
+    controller: Akeneo\Catalogs\Infrastructure\Controller\Internal\GetCategoryTreeRootsAction
+    methods: [ GET ]
```
tseho, line 33: I will invoke the REST expert ... @masacc !

I think that here, it should be only one endpoint.
You should be able to do `/rest/catalogs/categories?root=true` or something like this.

components/catalogs/back/tests/Integration/Infrastructure/Controller/Internal/GetCategoriesActionTest.php
```diff
@@ -0,0 +1,85 @@
...
+            ],
+        );
+
+        $response = $client->getResponse();
+        Assert::assertEquals(200, $response->getStatusCode());
+
+        $categories = \json_decode($response->getContent(), true);
+
+        $expectedTshirtCategory = [
+            'id' => $tshirtCategory->getId(),
```
tseho, line 46:
```suggestion
            'id' => $tshirtCategory->getId(),
```

Feels like a red flag. Same question as above, do we need to expose the id ?
If the only reason is because we need it to call `/rest/catalogs/categories/{categoryId}/children`, then, we could also change this other route to `/rest/catalogs/categories/{code}/children`.

components/catalogs/back/tests/Integration/IntegrationTestCase.php
```diff
@@ -213,4 +214,23 @@ protected function enableCatalog(string $id): void
...
+    }
+
+    protected function getCategory(string $code): ?CategoryInterface
+    {
+        $repository = self::getContainer()->get('pim_catalog.repository.category');
+
+        $categories = $repository->getCategoriesByCodes([$code]);
+
+        return \count($categories) > 0 ? $categories[0] : null;
+    }
```
tseho, line 235: btw, If you remove the id from our category model (and queries), you can burn this method, it's only used because you need the correct id to assert a query result.

tseho (commented):
Great job on this difficult PR, I've only a few comments.

components/catalogs/back/src/Infrastructure/Symfony/Resources/config/routing/internal.yml
masacc: Yes you are right

masacc: maybe `isRoot` to be consistent with `isLeaf`

tseho: both are valid for me, no preference.

components/catalogs/back/tests/Integration/Infrastructure/Controller/Internal/GetCategoriesActionTest.php
RiffVentura: You are right ! CategoryRepository can only fetch children through parent Id
But yes I could fetch the parent category first to get Id in the children query beforehand (assuming Category Service API would have a similar interface)
Will change!

components/catalogs/back/src/Infrastructure/Symfony/Resources/config/routing/internal.yml
RiffVentura: How do you think the endpoint should behave with both `isRoot` and `codes` query parameters specified?

tseho: Should throw a 400

components/catalogs/back/src/Infrastructure/Validation/CatalogPayload/CategoriesFieldIsValid.php
```diff
@@ -0,0 +1,70 @@
...
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ *
+ * @psalm-suppress PropertyNotSetInConstructor
+ */
+class CategoriesFieldIsValid extends Compound
+{
+    public function __construct(private GetCategoriesByCodeQueryInterface $getCategoriesByCodeQuery)
+    {
+        parent::__construct();
+    }
```
tseho, line 24: A constraint should not have dependencies.
Instead, you should move your callback to a dedicated constraint with its own validator.

see 
https://github.com/akeneo/pim-community-dev/pull/17651/files#diff-beb87611e5fbff6ad25740cc7cd453a17d2e50ee3b142adf5486548307602cd8R59
https://github.com/akeneo/pim-community-dev/pull/17651/files#diff-3e6795361f5df7568e54810472c8b2a8b35f16d0edd1cc1247d71d4830a8d1ec
https://github.com/akeneo/pim-community-dev/pull/17651/files#diff-d9f68a99555b81a0d504dec9a15a0e27f8f5318ac3a75cbe4408e7aa1dde2780

tseho: Can you also move the constraints to the same directory defined as in CXP-1245 ?

components/catalogs/back/src/Infrastructure/Validation/UpdateCatalogPayloadIsValidValidator.php
```diff
@@ -22,8 +24,10 @@
  */
 class UpdateCatalogPayloadIsValidValidator extends ConstraintValidator
 {
-    public function __construct(private GetChannelLocalesQueryInterface $getChannelLocalesQuery)
-    {
+    public function __construct(
+        private GetChannelLocalesQueryInterface $getChannelLocalesQuery,
+        private GetCategoriesByCodeQueryInterface $getCategoriesByCodeQuery,
+    ) {
```
tseho, line 30: Same, pls check what we did in CXP-1245

components/catalogs/back/tests/Integration/Infrastructure/Controller/Internal/GetCategoriesActionTest.php
```diff
@@ -0,0 +1,99 @@
...
+            'isLeaf' => true,
+        ];
+
+        $expectedShoesCategory = [
+            'code' => 'shoes',
+            'label' => 'Shoes',
+            'isLeaf' => true,
+        ];
+
+        Assert::assertEquals([$expectedShoesCategory, $expectedTshirtCategory], $categories);
```
tseho, line 59:
```suggestion
        Assert::assertEquals([
          [
            'code' => 'shoes',
            'label' => 'Shoes',
            'isLeaf' => true,
          ],
          [
            'code' => 'tshirt',
            'label' => 'T-shirt',
            'isLeaf' => true,
          ]
        ], $categories);
```
Not important but if you don't need these variables, you can simplify interpretation of your test.

components/catalogs/back/src/Application/Persistence/GetCategoriesByCodeQueryInterface.php
```diff
@@ -0,0 +1,18 @@
...
+ * @copyright 2022 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+interface GetCategoriesByCodeQueryInterface
+{
+    /**
+     * @param array<string> $categoryCodes
+     * @return array<array-key, array{code: string, label: string, isLeaf: bool}>
+     */
+    public function execute(array $categoryCodes, string $locale): array;
```
tseho, line 17:
```suggestion
    public function execute(array $categoryCodes, string $locale = 'en_US'): array;
```

components/catalogs/back/src/Infrastructure/Persistence/GetCategoriesByCodeQuery.php
```diff
@@ -0,0 +1,53 @@
...
+class GetCategoriesByCodeQuery implements GetCategoriesByCodeQueryInterface
+{
+    public function __construct(private CategoryRepositoryInterface $categoryRepository)
+    {
+    }
+
+    /**
+     * @inheritDoc
+     */
+    public function execute(array $categoryCodes, string $locale): array
```
tseho, line 25:
```suggestion
    public function execute(array $categoryCodes, string $locale = 'en_US'): array
```

components/catalogs/back/src/Infrastructure/Validation/CatalogPayload/CategoriesFieldIsValid.php
```diff
@@ -0,0 +1,70 @@
...
+                        ],
+                    ],
+                    'allowMissingFields' => false,
+                    'allowExtraFields' => false,
+                ]),
+                new Assert\Callback(function (array $criterion, ExecutionContextInterface $context): void {
+                    /** @var array<array-key, string> $categoryCodes */
+                    $categoryCodes = $criterion['value'];
+
+                    $existingCategories = $this->getCategoriesByCodeQuery->execute($categoryCodes, 'en_US');
```
tseho, line 56:
```suggestion
                    $existingCategories = $this->getCategoriesByCodeQuery->execute($categoryCodes);
```

tseho: I think that the locale should be optional on the query.
First, the query can have the default value in one place instead of everywhere it's used.
Then, It's not breaking if you don't pass it or don't have `en_US` in your PIM because we fallback on the `[code]` label.

components/catalogs/back/src/Infrastructure/Controller/Internal/GetCategoriesAction.php
```diff
@@ -0,0 +1,59 @@
...
+ */
+class GetCategoriesAction
+{
+    public function __construct(
+        private GetCategoryTreeRootsQueryInterface $getCategoryTreeRootsQuery,
+        private GetCategoriesByCodeQueryInterface $getCategoriesByCodeQuery
+    ) {
+    }
+
+    public function __invoke(Request $request): Response
```
tseho, line 27: No pagination ?

tseho: Ok, I saw that you control the amount of codes requested in the front, 20 by 20.
Do you think that the controller should throw if you request more codes than that ?

components/catalogs/back/src/Infrastructure/Controller/Internal/GetCategoryChildrenAction.php
```diff
@@ -0,0 +1,37 @@
...
+ * @copyright 2022 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+class GetCategoryChildrenAction
+{
+    public function __construct(private GetCategoryChildrenQueryInterface $getCategoryChildrenQuery)
+    {
+    }
+
+    public function __invoke(Request $request, string $categoryCode): Response
```
tseho, line 24: Same here. No pagination ?

components/catalogs/back/src/Infrastructure/Controller/Internal/GetCategoriesAction.php
RiffVentura: Honestly, I don't know,
It would be a simpler fix but not a clean one
CategoryRepository had no paginated query and Rule Engine implementation made no use of it (unless to manually paginate codes)

Moreover, how would children be displayed with pagination?

What do you think should be done?

components/catalogs/back/src/Infrastructure/Symfony/Resources/translations/validators.en_US.yml
```diff
@@ -17,3 +17,5 @@ akeneo_catalogs:
                     empty: 'This value must not be empty.'
                 completeness:
                     value: 'Completeness value must be between 0 and 100 percent.'
+                category:
+                    value: 'Some of the selected categories do not exist or have been selected. Please remove these categories.'
```
tseho, line 21: What do you mean by "or have been selected" ?

tseho: Did Megane validated this message ?

components/catalogs/back/src/Infrastructure/Validation/ProductSelection/CategoriesCriterionContainsValidCategoriesValidator.php
```diff
@@ -0,0 +1,49 @@
...
+            throw new UnexpectedValueException($value, 'array');
+        }
+
+        /** @var array<array-key, string> $categoriesCodes */
+        $categoriesCodes = $value['value'];
+
+        $existingCategories = $this->getCategoriesByCodeQuery->execute($categoriesCodes);
+        $existingCategoryCodes = \array_column($existingCategories, 'code');
+
+        $nonExistingCategoryCodes = \array_diff($categoriesCodes, $existingCategoryCodes);
```
tseho, line 40: If you don't need the diff, you can simply do a count !== count of the 2 lists.

components/catalogs/back/src/Infrastructure/Validation/ProductSelection/SystemCriterion/CategoriesCriterion.php
```diff
@@ -0,0 +1,61 @@
...
+    {
+        return [
+            new Assert\Sequentially([
+                new Assert\Collection([
+                    'fields' => [
+                        'field' => [
+                            new Assert\IdenticalTo('categories'),
+                        ],
+                        'operator' => [
+                            new Assert\Type('string'),
```
tseho, line 38:
```suggestion
```

Optional, but I found out that it's not needed. Assert\Choice is really strict.

components/catalogs/back/tests/Integration/Infrastructure/Controller/Internal/GetCategoryChildrenActionTest.php
```diff
@@ -0,0 +1,72 @@
...
+            'code' => 'child2',
+            'parent' => 'parent_category',
+            'labels' => ['en_US' => 'Child 2 category']
+        ]);
+
+        $this->createCategory(['code' => 'grand_child', 'parent' => 'child1']);
+
+        $client->request(
+            'GET',
+            "/rest/catalogs/categories/parent_category/children",
```
tseho, line 46: It seems that we don't have proper code style linting in the tests, you are mixing `'` and `"`.
You don't need to change, we will take a look on the linter later to find out why.

tseho (approved):
Great job !

components/catalogs/back/src/Infrastructure/Symfony/Resources/translations/validators.en_US.yml
RiffVentura: Not yet
Indeed it does not make any sense

RiffVentura: Updated with validated message
```
The following categories do not exist anymore: category1, category2, [...]. Please remove them from the criterion value.
```

PR #17660, opened 2022-08-12
DPE-33-1 : Create Deprecation Dependencies with only PHPStan config file

niarcotyk:
**Purpose**: 
- PR to implement an extension to PHPStan that allows running a dependency check and update
- This extension issues deprecation warnings on code, which uses properties/functions/methods/classes that are annotated as @deprecated.
**Use**:
- Can be launched in a CI pipeline
Screenshot from 2022-08-12 16-11-31


ValentinMumble (approved):

niarcotyk:
Close because the developers went through another way

PR #17663, opened 2022-08-16
CXP 1237 Added category criterion (Front)

RiffVentura:
Added new system criterion: Category
It displays a category tree inspired by Rule Engine but does not use any external dependency except DSM

### DIFF
--- a/components/catalogs/front/src/components/ProductSelection/criteria/CategoryCriterion/CategoryCriterion.integration.tsx
+++ b/components/catalogs/front/src/components/ProductSelection/criteria/CategoryCriterion/CategoryCriterion.integration.tsx
@@ -1,0 +1,502 @@
+import React from 'react';
+import {fireEvent, render, screen, waitFor, within} from '@testing-library/react';
+import {ThemeProvider} from 'styled-components';
+import {pimTheme} from 'akeneo-design-system';
+import {Operator} from '../../models/Operator';
+import {ReactQueryWrapper} from '../../../../../tests/ReactQueryWrapper';
+import fetchMock from 'jest-fetch-mock';
+import {CategoryCriterion} from './CategoryCriterion';
+
+jest.mock('../../hooks/useOperatorTranslator');
+
+const changeOperatorValueTo = (operator: string) => {
+    const operatorSelect = screen.getByTestId('operator');
+    fireEvent.click(within(operatorSelect).getByRole('textbox'));
+    fireEvent.click(screen.getByText(operator));
+};
+
+beforeEach(() => {
+    fetchMock.resetMocks();
+});
+
+test('it renders the selected categories', async () => {
+    const categories = [
+        {
+            id: 1,
+            code: 'catA',
+            label: '[catA]',
+            isLeaf: false,
+        },
+        {
+            id: 43,
+            code: 'catB',
+            label: '[catB]',
+            isLeaf: true,
+        },
+    ];
+
+    fetchMock.mockResponses(
+        //useCategories with catA and catB
+        JSON.stringify(categories),
+        //useCategories with an empty array
+        JSON.stringify([])
+    );
+
+    render(
+        <ThemeProvider theme={pimTheme}>
+            <ReactQueryWrapper>
+                <CategoryCriterion
+                    state={{field: 'categories', operator: Operator.IN_LIST, value: ['catA', 'catB']}}
+                    onChange={jest.fn()}
+                    onRemove={jest.fn()}
+                    errors={{
+                        field: null,
+                        operator: null,
+                        value: null,
+                    }}
+                />
+            </ReactQueryWrapper>
+        </ThemeProvider>
+    );
+
+    expect(screen.getByText('akeneo_catalogs.product_selection.criteria.category.label')).toBeInTheDocument();
+    expect(screen.getByText(Operator.IN_LIST)).toBeInTheDocument();
+
+    expect(await screen.findByText('[catA]')).toBeInTheDocument();
+    expect(screen.getByText('[catB]')).toBeInTheDocument();
+});
+
+test('it renders inputs with validation errors', () => {
+    fetchMock.mockResponseOnce(JSON.stringify([]));
+
+    render(
+        <ThemeProvider theme={pimTheme}>
+            <ReactQueryWrapper>
+                <CategoryCriterion
+                    state={{field: 'categories', operator: Operator.IN_LIST, value: ['catA', 'catB']}}
+                    onChange={jest.fn()}
+                    onRemove={jest.fn()}
+                    errors={{
+                        field: null,
+                        operator: 'Invalid operator.',
+                        value: 'Invalid value.',
+                    }}
+                />
+            </ReactQueryWrapper>
+        </ThemeProvider>
+    );
+
+    expect(screen.getByText('Invalid operator.')).toBeInTheDocument();
+    expect(screen.getByText('Invalid value.')).toBeInTheDocument();
+});
+
+test('it calls onRemove when criterion is removed', () => {
+    fetchMock.mockResponseOnce(JSON.stringify([]));
+
+    const onRemove = jest.fn();
+
+    render(
+        <ThemeProvider theme={pimTheme}>
+            <ReactQueryWrapper>
+                <CategoryCriterion
+                    state={{field: 'categories', operator: Operator.IN_LIST, value: ['catA', 'catB']}}
+                    onChange={jest.fn()}
+                    onRemove={onRemove}
+                    errors={{
+                        field: null,
+                        operator: null,
+                        value: null,
+                    }}
+                />
+            </ReactQueryWrapper>
+        </ThemeProvider>
+    );
+
+    fireEvent.click(screen.getByTitle('akeneo_catalogs.product_selection.action.remove'));
+
+    expect(onRemove).toHaveBeenCalled();
+});
+
+test('it calls onChange when the operator changes', () => {
+    const categories = [
+        {
+            id: 1,
+            code: 'catA',
+            label: '[catA]',
+            isLeaf: false,
+        },
+        {
+            id: 43,
+            code: 'catB',
+            label: '[catB]',
+            isLeaf: true,
+        },
+    ];
+
+    fetchMock.mockResponses(
+        //useCategories with catA and catB
+        JSON.stringify(categories),
+        //useCategories with an empty array
+        JSON.stringify([])
+    );
+
+    const onChange = jest.fn();
+
+    render(
+        <ThemeProvider theme={pimTheme}>
+            <ReactQueryWrapper>
+                <CategoryCriterion
+                    state={{field: 'categories', operator: Operator.IN_LIST, value: ['catA', 'catB']}}
+                    onChange={onChange}
+                    onRemove={jest.fn()}
+                    errors={{
+                        field: null,
+                        operator: null,
+                        value: null,
+                    }}
+                />
+            </ReactQueryWrapper>
+        </ThemeProvider>
+    );
+
+    changeOperatorValueTo(Operator.IN_CHILDREN_LIST);
+
+    expect(onChange).toHaveBeenCalledWith({
+        field: 'categories',
+        operator: Operator.IN_CHILDREN_LIST,
+        value: ['catA', 'catB'],
+    });
+});
+
+test('it hides value field and resets selected value when operator value is unclassified', async () => {
+    const categories = [
+        {
+            id: 1,
+            code: 'catA',
+            label: '[catA]',
+            isLeaf: false,
+        },
+        {
+            id: 43,
+            code: 'catB',
+            label: '[catB]',
+            isLeaf: true,
+        },
+    ];
+
+    fetchMock.mockResponses(
+        //useCategories with catA and catB
+        JSON.stringify(categories),
+        //useCategories with an empty array
+        JSON.stringify([])
+    );
+
+    const onChange = jest.fn();
+
+    render(
+        <ThemeProvider theme={pimTheme}>
+            <ReactQueryWrapper>
+                <CategoryCriterion
+                    state={{field: 'categories', operator: Operator.IN_LIST, value: ['catA', 'catB']}}
+                    onChange={onChange}
+                    onRemove={jest.fn()}
+                    errors={{
+                        field: null,
+                        operator: null,
+                        value: null,
+                    }}
+                />
+            </ReactQueryWrapper>
+        </ThemeProvider>
+    );
+
+    expect(await screen.findByText('[catA]')).toBeInTheDocument();
+
+    changeOperatorValueTo(Operator.UNCLASSIFIED);
+
+    expect(onChange).toHaveBeenCalledWith({
+        field: 'categories',
+        operator: Operator.UNCLASSIFIED,
+        value: [],
+    });
+});
+
+test('it calls onChange with remaining items when removing category from selection ', async () => {
+    const categories = [
+        {
+            id: 1,
+            code: 'catA',
+            label: '[catA]',
+            isLeaf: false,
+        },
+        {
+            id: 43,
+            code: 'catB',
+            label: '[catB]',
+            isLeaf: true,
+        },
+        {
+            id: 67,
+            code: 'catC',
+            label: '[catC]',
+            isLeaf: false,
+        },
+    ];
+
+    fetchMock.mockResponses(
+        //useCategories with catA, catB, and catC
+        JSON.stringify(categories),
+        //useCategories with an empty array
+        JSON.stringify([])
+    );
+
+    const onChange = jest.fn();
+
+    render(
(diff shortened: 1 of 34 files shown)


guillaumefer-akeneo (approved):

components/catalogs/front/src/components/ProductSelection/criteria/CategoryCriterion/CategoryCriterion.integration.tsx
```diff
@@ -0,0 +1,502 @@
...
+
+const changeOperatorValueTo = (operator: string) => {
+    const operatorSelect = screen.getByTestId('operator');
+    fireEvent.click(within(operatorSelect).getByRole('textbox'));
+    fireEvent.click(screen.getByText(operator));
+};
+
+beforeEach(() => {
+    fetchMock.resetMocks();
+});
```
masacc, line 20: in CXP-1222, it has been added to `setup-integration.ts` : https://github.com/akeneo/pim-community-dev/pull/17664/files#diff-908fa8c7d70bd94a58103d8581ff789e5e18804103420f5ffe6c984c36df369d

components/catalogs/front/src/components/ProductSelection/criteria/CategoryCriterion/CategoryCriterion.tsx
```diff
@@ -0,0 +1,73 @@
...
+import {Operator} from '../../models/Operator';
+import {CriterionModule} from '../../models/Criterion';
+import {CategoryCriterionState} from './types';
+import {useTranslate} from '@akeneo-pim-community/shared';
+import {CategoryOperatorInput} from './CategoryOperatorInput';
+import {CategorySelectInput} from './CategorySelectInput';
+
+const Fields = styled.div`
+    display: flex;
+    gap: 20px;
```
masacc, line 13: don't forget:
```suggestion
    gap: 20px;
    flex-wrap: wrap;
    flex-grow: 1;
```

components/catalogs/front/src/components/ProductSelection/criteria/CategoryCriterion/CategoryCriterion.tsx
```diff
@@ -0,0 +1,73 @@
...
+            <List.RemoveCell>
+                <IconButton
+                    ghost='borderless'
+                    level='tertiary'
+                    icon={<CloseIcon />}
+                    title={translate('akeneo_catalogs.product_selection.action.remove')}
+                    onClick={onRemove}
+                />
+            </List.RemoveCell>
+            {errorHelpers.length > 0 && <List.RowHelpers>{errorHelpers}</List.RowHelpers>}
```
masacc, line 68: to replace by what has been done for `CompletenessCriterion` : https://github.com/akeneo/pim-community-dev/pull/17664/files#diff-0537dd656c55a456af6b9dd5dce45e2de21175e8a2f1c616b15e6ae1349b1c3aR64

components/catalogs/front/src/components/ProductSelection/hooks/useCategories.ts
```diff
@@ -0,0 +1,27 @@
...
+        const response = await fetch(`/rest/catalogs/categories?codes=${_codes}`, {
+            headers: {
+                'X-Requested-With': 'XMLHttpRequest',
+            },
+        });
+
+        return await response.json();
+    }, [codes]);
+
+    return useQuery<Data, ResultError, Data>(['categories', codes], fetchCategories);
```
masacc, line 26: why don't you do it like in the other hooks? Is it just for readability or is there another reason?

components/catalogs/front/src/components/ProductSelection/hooks/useCategories.ts
```diff
@@ -15,6 +16,7 @@ type Result = {
 };
 
 export const useCategories = ({codes = [], isRoot = false}: QueryParams): Result => {
+    const locale = useUserContext().get('uiLocale');
     return useQuery<Data, ResultError, Data>(['categories', {codes, isRoot}], async () => {
```
masacc, line 20: add locale in cache key

components/catalogs/front/src/components/ProductSelection/criteria/CategoryCriterion/CategoryCriterion.integration.tsx
```diff
@@ -0,0 +1,419 @@
...
+            isLeaf: false,
+        },
+        {
+            code: 'catB',
+            label: '[catB]',
+            isLeaf: true,
+        },
+    ];
+
+    fetchMock.mockResponseOnce(JSON.stringify(categories));
```
tseho, line 32: When working on the CXP-1245 with Guillaume & Thomas, we started having issues with using the simple `mockResponseOnce` of fetchMock.
One big issue of this, it's that if your code change and the API calls are not executed in the same order, your integration test is not passing anymore, and this is an issue. we don't want to have fragile tests.

What we did instead, is to use this syntax:

```
    fetchMock.mockResponse(req => {
        switch (req.url) {
            case '/rest/catalogs/attributes/name_localizable_scopable':
                return Promise.resolve(
                    JSON.stringify({
                        label: 'Name',
                        code: 'name',
                        type: 'pim_catalog_text',
                        scopable: true,
                        localizable: true,
                    })
                );
```
This way, mocked response are connected to a specific url, not the order they are called.

tseho: Yes, we haven't updated it in the other tests but we plan to, for consistency.

components/catalogs/front/src/components/ProductSelection/criteria/CategoryCriterion/CategoryCriterion.tsx
```diff
@@ -0,0 +1,73 @@
...
+`;
+
+const CategoryCriterion: FC<CriterionModule<CategoryCriterionState>> = ({state, errors, onChange, onRemove}) => {
+    const translate = useTranslate();
+    const hasError = Object.values(errors).filter(n => n).length > 0;
+    const [showCategories, setShowCategories] = useState<boolean>(false);
+
+    useEffect(() => {
+        setShowCategories(Operator.UNCLASSIFIED !== state.operator);
+    }, [state.operator]);
```
tseho, line 36: the useEffect is useless here.
To understand more where useEffect should be used and why, and understand why this particular usage is not recommended, you can checkout the new react documentation about it.
https://beta.reactjs.org/learn/you-might-not-need-an-effect#updating-state-based-on-props-or-state

```suggestion
    const showCategories = Operator.UNCLASSIFIED !== state.operator;
```

Yes, the same mistake was done in FamilyCriterion but I already planned to fix it, see https://github.com/akeneo/pim-community-dev/pull/17694/files

components/catalogs/front/src/components/ProductSelection/criteria/CategoryCriterion/CategoryOperatorInput.tsx
```diff
@@ -0,0 +1,65 @@
...
+        <SelectInput
+            emptyResultLabel=''
+            openLabel=''
+            value={state.operator}
+            onChange={handleChange}
+            clearable={false}
+            invalid={isInvalid}
+            data-testid='operator'
+        >
+            <SelectInput.Option value={Operator.IN_LIST} title={translateOperator(Operator.IN_LIST)}>
```
tseho, line 37: It's not blocking, but we followed your advice in CXP-1245
https://github.com/akeneo/pim-community-dev/pull/17651/files#diff-2f000887acd38a74ed90fd094b8d14e5b50b03370d17958639cda7cce5ac8890

components/catalogs/front/src/components/ProductSelection/criteria/CategoryCriterion/CategoryTreeSelector.tsx
```diff
@@ -0,0 +1,78 @@
...
+        if (selectedTree !== null || categoryTreeRoots === undefined) {
+            return;
+        }
+
+        if (categoryTreeRoots.length === 0) {
+            throw new Error('No tree root found');
+        }
+
+        onChange(categoryTreeRoots[0]);
+    }, [selectedTree, categoryTreeRoots, onChange]);
```
tseho, line 41: What's the purpose of this useEffect ?

components/catalogs/front/src/components/ProductSelection/hooks/useCategories.ts
```diff
@@ -0,0 +1,43 @@
...
+type Data = Category[];
+type ResultError = Error | null;
+type Result = {
+    isLoading: boolean;
+    isError: boolean;
+    data: Data | undefined;
+    error: ResultError;
+};
+
+export const useCategories = ({codes = [], isRoot = false}: QueryParams): Result => {
```
tseho, line 18: It worries me that we don't have pagination here.

components/catalogs/front/src/components/ProductSelection/hooks/useCategoryTreeRoots.ts
```diff
@@ -0,0 +1,33 @@
...
+    isError: boolean;
+    data: Category[] | undefined;
+    error: ResultError;
+};
+
+export const useCategoryTreeRoots = (): Result => {
+    const queryClient = useQueryClient();
+    const categoryResult = useCategories({isRoot: true});
+
+    const categoryTreeRoots = categoryResult.data;
```
tseho, line 18:
```suggestion
    const {data: categoryTreeRoots} = useCategories({isRoot: true});
```

components/catalogs/front/tests/ReactQueryWrapper.tsx
```diff
@@ -1,10 +1,26 @@
...
-const ReactQueryWrapper: FC = ({children}) => (
-    <QueryClientProvider client={new QueryClient()}>
-        {children}
-    </QueryClientProvider>
-);
+setLogger({
+    log: console.log,
+    warn: console.warn,
+    error: () => {/* no logging output in console on error for tests */},
+});
```
tseho, line 8: Why ?

components/catalogs/front/tests/ReactQueryWrapper.tsx
```diff
@@ -1,10 +1,26 @@
...
+
+const ReactQueryWrapper: FC = ({children}) => {
+    const queryClient = new QueryClient({
+        defaultOptions: {
+            queries: {
+                retry: false,
+            },
+        }
+    });
```
tseho, line 17: Which failing query forced you to do this ?

components/catalogs/front/src/components/ProductSelection/criteria/CategoryCriterion/CategoryTreeSelector.tsx
RiffVentura: Initialize selected tree with the first tree found
I probably should have moved it into a custom hook with a proper name

components/catalogs/front/tests/ReactQueryWrapper.tsx
RiffVentura: useCategories with codes and isRoot is forbidden so the query throws before it fetches
However, when testing the behavior, the query is retried and appears to be loading indefinitely.
In the [docs](https://tanstack.com/query/v4/docs/guides/testing#turn-off-retries), this was suggested as a fix
I initially wanted to use this client only in the test case, but I figured this was better as it would prevent wasting time to figure out again why it is happening.
On the same note: falling queries pollute the output that's why I've turned off logging on error

components/catalogs/front/tests/ReactQueryWrapper.tsx
RiffVentura: see the comment below

components/catalogs/front/tests/ReactQueryWrapper.tsx
tseho: >  However, when testing the behavior, the query is retried and appears to be loading indefinitely.

We should avoid having different behaviors between the production and the tests.
You can test it without waiting indefinitely, see https://github.com/akeneo/pim-community-dev/pull/17706

>  On the same note: falling queries pollute the output that's why I've turned off logging on error

Please do not disable logging during the tests.
I've tested locally when enabling it again, on top of the useCategories error logged, there is also 2 errors in the CategoryCriterion integration that are muted.
If I'm not mistaken, those errors are shown because there is missing mocks to API calls.

If you want to mute a specific error during one test where you know it's throwing, you should do like we already do in similar tests.
see https://github.com/akeneo/pim-community-dev/blob/master/components/catalogs/front/src/components/ErrorBoundary/ErrorBoundary.integration.tsx#L24-L27

components/catalogs/front/tests/ReactQueryWrapper.tsx
tseho: https://github.com/akeneo/pim-community-dev/pull/17663#discussion_r951404077

components/catalogs/front/src/components/ProductSelection/criteria/CategoryCriterion/CategoryTreeSelector.tsx
tseho: Having a subcomponent filling the default value of a `useState` is an anti-pattern.
If you need to select automatically the default tree root, you should change CategoryTreeSelector so that it requires a selected tree in its props and ask the parent to fetch the find the first one if there is none selected before rendering the subcomponent.

Thanks to react query and the option `staleTime: 60,`, both components can call the same query in a short time while producing only one API call.

tseho: And yes, you can move it to a new hook.
Then, you can replace
```
const [selectedTree, setSelectedTree] = useState<Category | null>(null);
```
by 
```
const [selectedTree, setSelectedTree] = useSelectedTree();
```

components/catalogs/front/src/components/ProductSelection/hooks/useCategories.ts
```diff
@@ -0,0 +1,43 @@
...
+    const locale = useUserContext().get('uiLocale');
```
tseho, line 19:
```suggestion
    const locale = useUserContext().get('catalogLocale');
```

To translate the data, it's `catalogLocale`, not `uiLocale`.

components/catalogs/front/src/components/ProductSelection/criteria/CategoryCriterion/CategorySelectInput.tsx
```diff
@@ -0,0 +1,66 @@
...
+            {isOpen && selectedTree && (
+                <TreeDropdown verticalPosition='down' onClose={() => setOpen(false)} parentRef={dropdownParentRef}>
+                    <CategoryTreeSelector selectedTree={selectedTree} onChange={setSelectedTree} />
+                    {
+                        <CategorySelector
+                            root={selectedTree}
+                            selectedCategories={selectedCategories}
+                            onCategorySelect={handleCategorySelection}
+                        />
+                    }
```
tseho, line 59: Why the `{ }` around it ?

components/catalogs/front/src/components/ProductSelection/hooks/useCategories.ts
```diff
@@ -0,0 +1,49 @@
...
+            const response = await fetch('/rest/catalogs/categories?' + queryParameters, {
+                headers: {
+                    'X-Requested-With': 'XMLHttpRequest',
+                },
+            });
+
+            return await response.json();
+        },
+        {
+            staleTime: 60,
```
tseho, line 46:
```suggestion
            staleTime: 60,
```
Btw, I was mistaken, it's ms, not seconds.

And i've added better default values in https://github.com/akeneo/pim-community-dev/pull/17731/files#diff-aa18996c1675030f63cf13d1c41f6bc8fb736a5c6f14813eb6405183e8fd20d0

components/catalogs/front/src/components/ProductSelection/hooks/useCategoryChildren.ts
```diff
@@ -0,0 +1,48 @@
...
+                'X-Requested-With': 'XMLHttpRequest',
+            },
+        });
+
+        const responseJson = await response.json();
+
+        if (response.ok) {
+            const children = responseJson as Category[];
+            children.forEach(child => {
+                queryClient.setQueryData(['categories', [child.code]], [child]);
```
tseho, line 30: which query are you pre-caching ?
useCategories uses `['categories', {codes, isRoot, locale}],`

guillaumefer-akeneo (approved):

components/catalogs/front/src/components/ProductSelection/hooks/useCategoryChildren.ts
RiffVentura: I forgot to delete it as I moved it out of the fetch function into useEffect below. I will remove it

PR #17681, opened 2022-08-18
CXP-1245: Proposal on validation refactoring

tseho:

Basically, I moved from
```
            new Assert\Sequentially([
                new AttributeTextCriterionStructure(),
                new AttributeTextCriterionValues(),
            ]),
```
to 
```
            new Assert\Sequentially([
                new Assert\Collection([
                    'fields' => [
                        'field' => [
                            new Assert\Type('string'),
                            new Assert\NotBlank(),
                        ],
                        'operator' => [
                            new Assert\Type('string'),
                            new Assert\NotBlank(),
                            new Assert\Choice(['=', '!=', 'CONTAINS', 'DOES NOT CONTAIN', 'STARTS WITH', 'EMPTY', 'NOT EMPTY']),
                        ],
                        'value' => [
                            new Assert\Type('string'),
                        ],
                        'scope' => [
                            new Assert\Type('string'),
                        ],
                        'locale' => [
                            new Assert\Type('string'),
                        ],
                    ],
                    'allowMissingFields' => false,
                    'allowExtraFields' => false,
                ]),
                new CriterionOperatorsRequireEmptyValue([
                    'EMPTY',
                    'NOT_EMPTY',
                ]),
                new AttributeCriterionContainsValidScope(),
                new AttributeCriterionContainsValidLocale(),
            ]),
```

Where everything from the previous `AttributeTextCriterionValues` is now reusable.

Names of the new constraints are not final, I'm open to better suggestions :)

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc


tseho:
`CriterionOperatorsWithEmptyValue` => `CriterionOperatorsRequireEmptyValue` ?
`AttributeCriterionScope` => `AttributeCriterionContainsValidScope` ?
`AttributeCriterionLocale` => `AttributeCriterionContainsValidLocale` ?

components/catalogs/back/src/Infrastructure/Validation/ProductSelection/CriterionOperatorsWithEmptyValueValidator.php
```diff
@@ -0,0 +1,65 @@
...
+        }
+    }
+
+    private function isEmpty(mixed $value): bool
+    {
+        if (null === $value) {
+            return true;
+        }
+
+        if (is_string($value) && '' === $value) {
```
masacc, line 59: In `NotBlankValidator`, Symfony uses this condition:
```
if (false === $value || (empty($value) && '0' != $value))
```
What do you think about it?

components/catalogs/back/src/Infrastructure/Validation/ProductSelection/CriterionOperatorsWithEmptyValueValidator.php
```diff
@@ -0,0 +1,65 @@
...
+    {
+        /** @var AttributeCriterion $value */
+
+        if (!$constraint instanceof CriterionOperatorsWithEmptyValue) {
+            throw new UnexpectedTypeException($constraint, CriterionOperatorsWithEmptyValue::class);
+        }
+
+        if (0 === \count($constraint->operators)) {
+            throw new \LogicException('The operators option should not be empty');
+        }
```
masacc, line 36: What about validation if there is only operators that need non empty value?

components/catalogs/back/src/Infrastructure/Validation/ProductSelection/CriterionOperatorsWithEmptyValueValidator.php
tseho: Yes, I've looked into `NotBlankValidator`. I don't want to use `empty` and for boolean, I'd rather wait for the boolean attribute to check what's the expected behavior.

components/catalogs/back/src/Infrastructure/Validation/ProductSelection/CriterionOperatorsWithEmptyValueValidator.php
tseho: In this case, you don't need this validator.
If all operators require NOT NULL, you can enforce it on the `Structure` constraint, with a `NotBlank` on values.
If all operators require NULL, you can also enforice it on the `Structure` constraint.

masacc: yes good catch

masacc (approved):

guillaumefer-akeneo (approved):

PR #17683, opened 2022-08-18
PIM-10584: add migration & update fixtures to fix volume flow conversion

grena:
Fix some conversion for the Volume Flow units.

**Migration performances**
We only update one entry, we don't update the table structure, so it should be OK.

- [x] Tests
- [x] Migration & Installer
- [ ] PM Validation (Story)


ghoussard (approved):

upgrades/test_schema/Version_7_0_20220818093018_fix_volume_flow_conversion_Integration.php
```diff
@@ -0,0 +1,93 @@
...
+    protected function setUp(): void
+    {
+        parent::setUp();
+        $this->connection = $this->get('database_connection');
+    }
+
+    /** @test */
+    public function it_fixes_the_incorrect_volume_flow_conversion(): void
+    {
+        self::assertEquals('mul', $this->getConversionOperatorFor('LITER_PER_MINUTE'));
```
StevenVAIDIE, line 30: As you fixed the minimal catalog you should update the unit as before first

StevenVAIDIE (approved):

PR #17686, opened 2022-08-18
CXP-1247: add AttributeBooleanCriterion

guillaumefer-akeneo:
image

- [x] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc


components/catalogs/front/src/components/ProductSelection/criteria/AttributeYesNoCriterion/AttributeYesNoCriterion.integration.tsx
```diff
@@ -0,0 +1,390 @@
...
+            // useInfiniteChannels
+            case '/rest/catalogs/channels?page=1&limit=20':
+                return Promise.resolve(JSON.stringify([channelEcommerce, channelPrint]));
+            default:
+                throw Error(req.url);
+        }
+    });
+});
+
+test('it renders the scopable and localizable text attribute criterion', async () => {
```
tseho, line 83:
```suggestion
test('it renders the scopable and localizable yes/no attribute criterion', async () => {
```

Also, should we call it boolean instead of yes/no ?
Internally, it's `pim_catalog_boolean` :thinking:

components/catalogs/front/src/components/ProductSelection/criteria/AttributeYesNoCriterion/AttributeYesNoCriterion.integration.tsx
```diff
@@ -0,0 +1,390 @@
...
+    render(
+        <ThemeProvider theme={pimTheme}>
+            <ReactQueryWrapper>
+                <AttributeYesNoCriterion
+                    state={{
+                        field: 'name_localizable_scopable',
```
tseho, line 89:
```suggestion
                        field: 'name_localizable_scopable',
```

It's not essential, but I would have picked another name for the attributes.
Having a boolean attribute called `name` or `name_localizable_scopable` does not really make sense.

tseho (approved):
You forgot to launch the CI, I can already tell you that it's going to be red, you forgot to update `SearchAttributesQueryTest`.
It's ok, it's something that can be easily overlooked.

Don't worry too much about my comments, it's small things.
It's a great PR you did there :clap:

components/catalogs/front/src/components/ProductSelection/criteria/AttributeYesNoCriterion/factory.ts
```diff
@@ -0,0 +1,16 @@
...
+
+export default (state?: Partial<AttributeYesNoCriterionState>): AttributeYesNoCriterionState => {
+    if (!state?.field) {
+        throw Error('You need to specific the attribute code when calling the attribute criterion factory');
+    }
+
+    return {
+        field: state.field,
+        operator: state?.operator ?? Operator.EQUALS,
+        value: state?.value ?? null,
```
RiffVentura, line 12: Either the default operator is empty or the default value is a boolean
```suggestion
        value: state?.value ?? false,
```

RiffVentura (approved):
:+1:

components/catalogs/front/src/components/ProductSelection/criteria/AttributeYesNoCriterion/AttributeYesNoCriterion.integration.tsx
guillaumefer-akeneo: Good question, it's the same things for the AttribueMeasurementCriterion with `pim_catalog_metric`.

tseho: Ah ah, it's the other way around for the other. `metric` is a typo, It should have been called measurement from the start.
it is now, the code was refactored to be moved to MeasurementBundle. Only the attribute type was never renamed because of the impact.

tseho: It raise an interesting point.
We need to update the Ubiquitous Language of the Catalogs feature.

components/catalogs/front/src/components/ProductSelection/criteria/AttributeYesNoCriterion/factory.ts
tseho: I would say that true should be the default value, not false.
But @meganelepalud-akeneo should decide.

components/catalogs/front/src/components/ProductSelection/criteria/AttributeYesNoCriterion/AttributeYesNoCriterion.integration.tsx
guillaumefer-akeneo: Renaming done :heavy_check_mark: => AttributeBooleanCriterion

components/catalogs/front/src/components/ProductSelection/criteria/AttributeYesNoCriterion/AttributeYesNoCriterion.integration.tsx
tseho: I would say that it's a matter of taste but I tend to be careful about having tests that make sense.

If you read your test aloud, it says, "I have an attribute `name`, I want to find all products where the value is `true`".

In AttributeSimpleSelectCriterion, I always named my attribute `clothing_size` with the options `xs, s, m, l, xl`.
In AttributeNumberCriterion, I named it `number_battery_cells`.

It's not essential as I said, but it make more sense and the test is easier to read, IMO.

tseho: You are not only writing tests to check that your code is working, you are also writing the intent behind the behavior.
If somebody else goes back to this piece of code in several years, it will really confuse them why it's done like this.

Can you rename it to an attribute code that makes sense for a boolean ?

guillaumefer-akeneo: It's renamed : display_color => Display Color :)

PR #17731, opened 2022-08-23
CXP-1244: Adds AttributeSimpleSelectCriterion

tseho:

https://user-images.githubusercontent.com/1421130/186208113-3ed2deac-1559-4e1a-84ba-c39db4e5b014.mp4

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/components/catalogs/back/src/Infrastructure/Validation/ProductSelection/AttributeCriterionContainsValidAttributeOptionsValidator.php
+++ b/components/catalogs/back/src/Infrastructure/Validation/ProductSelection/AttributeCriterionContainsValidAttributeOptionsValidator.php
@@ -52,13 +52,12 @@ final class AttributeCriterionContainsValidAttributeOptionsValidator extends Con
             return;
         }
 
-        $page = 1;
-        $limit = 50;
+        $chunks = \array_chunk($options, 50);
 
-        while (\count($slice = \array_slice($options, ($page - 1) * $limit, $limit)) > 0) {
-            $page++;
+        foreach ($chunks as $codes) {
+            $existingCodes = $this->getAttributeOptionsByCodeQuery->execute($value['field'], $codes);
 
-            if (\count($this->getAttributeOptionsByCodeQuery->execute($value['field'], $slice)) != \count($slice)) {
+            if (\count($existingCodes) !== \count($codes)) {
                 $this->context
                     ->buildViolation('akeneo_catalogs.validation.product_selection.criteria.attribute_option.unknown')
                     ->atPath('[value]')
--- a/components/catalogs/back/src/Infrastructure/Validation/ProductSelection/AttributeCriterionContainsValidAttributeOptionsValidator.php
+++ b/components/catalogs/back/src/Infrastructure/Validation/ProductSelection/AttributeCriterionContainsValidAttributeOptionsValidator.php
@@ -1,0 +1,71 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Akeneo\Catalogs\Infrastructure\Validation\ProductSelection;
+
+use Akeneo\Catalogs\Application\Persistence\FindOneAttributeByCodeQueryInterface;
+use Akeneo\Catalogs\Application\Persistence\GetAttributeOptionsByCodeQueryInterface;
+use Symfony\Component\Validator\Constraint;
+use Symfony\Component\Validator\ConstraintValidator;
+use Symfony\Component\Validator\Exception\UnexpectedTypeException;
+use Symfony\Component\Validator\Exception\UnexpectedValueException;
+
+/**
+ * @copyright 2022 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ *
+ * @psalm-suppress PropertyNotSetInConstructor
+ *
+ * @phpstan-type AttributeCriterion array{field: string, value: array<string>}
+ * @phpstan-type Attribute array{code: string, label: string, type: string, scopable: bool, localizable: bool}
+ */
+final class AttributeCriterionContainsValidAttributeOptionsValidator extends ConstraintValidator
+{
+    public function __construct(
+        private FindOneAttributeByCodeQueryInterface $findOneAttributeByCodeQuery,
+        private GetAttributeOptionsByCodeQueryInterface $getAttributeOptionsByCodeQuery,
+    ) {
+    }
+
+    public function validate($value, Constraint $constraint): void
+    {
+        if (!$constraint instanceof AttributeCriterionContainsValidAttributeOptions) {
+            throw new UnexpectedTypeException($constraint, AttributeCriterionContainsValidAttributeOptions::class);
+        }
+
+        if (!\is_array($value)) {
+            throw new UnexpectedValueException($value, 'array');
+        }
+
+        /** @var AttributeCriterion $value */
+
+        $attribute = $this->findOneAttributeByCodeQuery->execute($value['field']);
+
+        if (null === $attribute) {
+            throw new \LogicException('Attribute not found');
+        }
+
+        $options = $value['value'];
+
+        if ([] === $options) {
+            return;
+        }
+
+        $page = 1;
+        $limit = 50;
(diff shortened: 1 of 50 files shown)


components/catalogs/back/src/Infrastructure/Validation/ProductSelection/AttributeCriterionContainsValidAttributeOptionsValidator.php
```diff
@@ -0,0 +1,71 @@
...
+
+        $attribute = $this->findOneAttributeByCodeQuery->execute($value['field']);
+
+        if (null === $attribute) {
+            throw new \LogicException('Attribute not found');
+        }
+
+        $options = $value['value'];
+
+        if ([] === $options) {
```
RiffVentura, line 51: Could be omitted as the slice of an empty array is empty

tseho: Yes, but it's unnecessary to go further if empty.
Moreover, it's the slice implementation is replaced by something else later on, like an iterator without the need for pagination, the query will be then executed even if options is empty.

components/catalogs/front/src/components/ProductSelection/criteria/AttributeSimpleSelectCriterion/AttributeSimpleSelectOperatorInput.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+const AttributeSimpleSelectOperatorInput: FC<Props> = ({state, onChange, isInvalid}) => {
+    const translateOperator = useOperatorTranslator();
+
+    const handleChange = useCallback(
+        (newValue: string) => {
+            const operator = newValue as AttributeSimpleSelectCriterionOperator;
+
+            onChange({
+                ...state,
```
RiffVentura, line 21: I think it should reset locale and scope too

components/catalogs/front/src/components/ProductSelection/hooks/useAttributeOptionsByCodes.ts
```diff
@@ -0,0 +1,63 @@
...
+                ...cache,
+                ...newOptions,
+            }));
+        }
+    }
+
+    return {
+        isLoading: isLoading,
+        isError: isError,
+        data: Object.values(cache),
```
RiffVentura, line 60: Strange to return the whole cache. I would expect to have only options supplied in arguments

components/catalogs/front/src/components/ProductSelection/hooks/useFindAttributeCriterionByType.unit.ts
```diff
@@ -2,16 +2,19 @@ jest.unmock('./useFindAttributeCriterionByType');
 
 import {renderHook} from '@testing-library/react-hooks';
 import {useFindAttributeCriterionByType} from './useFindAttributeCriterionByType';
+import {AnyAttributeCriterion} from '../models/Criterion';
+import AttributeTextCriterion from '../criteria/AttributeTextCriterion';
+import AttributeSimpleSelectCriterion from '../criteria/AttributeSimpleSelectCriterion';
 
-const types: string[] = ['pim_catalog_text'];
+const criterions: [string, AnyAttributeCriterion][] = [
```
RiffVentura, line 9:
```suggestion
const criteria: [string, AnyAttributeCriterion][] = [
```

components/catalogs/front/src/components/ProductSelection/hooks/useInfiniteAttributeOptions.ts
```diff
@@ -0,0 +1,100 @@
...
+            const _search = search || pageParam?.search || '';
+            const _codes = codes.join(',');
+
+            const params = new URLSearchParams({
+                locale: locale,
+                codes: _codes,
+                search: _search,
+                page: _page.toString(),
+                limit: limit.toString(),
+            }).toString();
```
RiffVentura, line 51: could be inlined

```suggestion
            const params = new URLSearchParams({
                locale: locale,
                codes: codes.join(','),
                search: search || pageParam?.search || '',
                page: (pageParam?.number || 1).toString(),
                limit: limit.toString(),
            }).toString();
```

components/catalogs/front/tests/mockFetchResponses.ts
```diff
@@ -0,0 +1,22 @@
+import fetchMock from 'jest-fetch-mock';
+
+type Response = {
+    url: string;
+    /* eslint-disable-next-line @typescript-eslint/no-explicit-any */
+    json: any;
+};
+
+export const mockFetchResponses = (responses: Response[]): void => {
```
RiffVentura, line 9: :heart_eyes:

RiffVentura (approved):
:+1:

components/catalogs/front/src/components/ProductSelection/criteria/AttributeSimpleSelectCriterion/AttributeSimpleSelectCriterion.tsx
```diff
@@ -0,0 +1,82 @@
...
+                    {showValueInput && (
+                        <CriterionField>
+                            <AttributeSimpleSelectValueInput
+                                state={state}
+                                onChange={onChange}
+                                isInvalid={!!errors.value}
+                            />
+                        </CriterionField>
+                    )}
+                    {attribute?.scopable && (
```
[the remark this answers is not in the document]
tseho: Why ?
If the attribute is localizable, you select "IS NOT EMPTY", you hide the locale input, do you mean "values for ALL locales are not empty" ?
On the other side, if you say "IS NOT EMPTY" and select "English", you really mean "when the English value is not empty".
Which makes a lot more sense from Julia's perspective.

components/catalogs/front/src/components/ProductSelection/criteria/AttributeSimpleSelectCriterion/AttributeSimpleSelectOperatorInput.tsx
tseho: Same logic from the comment as above, locale & scope should not be affected by the operator, IMO.

components/catalogs/front/src/components/ProductSelection/criteria/AttributeSimpleSelectCriterion/AttributeSimpleSelectCriterion.tsx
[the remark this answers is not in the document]
RiffVentura: I didn't know that level of granularity was possible/allowed. This does mean that in case Julia wants a non empty value regardless of locale or channel, she must setup same criterion for each combination of locale/channel.

components/catalogs/back/src/Infrastructure/Validation/ProductSelection/AttributeCriterionContainsValidAttributeOptionsValidator.php
```diff
@@ -0,0 +1,71 @@
...
+            return;
+        }
+
+        $page = 1;
+        $limit = 50;
+
+        while (\count($slice = \array_slice($options, ($page - 1) * $limit, $limit)) > 0) {
```
RiffVentura, line 58: I think `array_chunk` would fit better here unless there are memory limitations I'm not aware of

tseho: No memory limitations, you are right, I should use array_chunk instead, it will be better.

components/catalogs/front/src/index.tsx
```diff
@@ -36,7 +36,14 @@ const routes = {
 
 DangerousMicrofrontendAutomaticAuthenticator.enable('admin', 'admin');
 
-const client = new QueryClient();
+const client = new QueryClient({
```
tseho, line 39: do the same in connectivity

guillaumefer-akeneo (approved):

PR #17781, opened 2022-08-29
RAB-1003: Save ids instead of username & labels in automation parameters

grena:

- add a way to retrieve all visible user groups (all groups but the system "All" group)
- change the way we upsert user, it's now possible to give `group_ids` instead of group labels (as user group label can be modified later on)

- [x] Tests

### DIFF
--- a/src/Akeneo/UserManagement/Bundle/Doctrine/ORM/Repository/GroupRepository.php
+++ b/src/Akeneo/UserManagement/Bundle/Doctrine/ORM/Repository/GroupRepository.php
@@ -85,6 +85,6 @@ class GroupRepository extends EntityRepository implements GroupRepositoryInterfa
 
     public function findOneById(int $id): ?GroupInterface
     {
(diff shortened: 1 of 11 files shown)


src/Akeneo/UserManagement/Bundle/Doctrine/ORM/Repository/GroupRepository.php
```diff
@@ -82,4 +82,9 @@ public function getUserQueryBuilder(GroupInterface $group)
             ->where('groups = :group')
             ->setParameter('group', $group);
     }
+
+    public function findOneById(int $id): ?GroupInterface
```
MadeWilson, line 86: same has $groupeRepository->find($id)

MadeWilson (approved):

ValentinMumble (approved):

StevenVAIDIE (approved):

PR #17783, opened 2022-08-29
CPM-653: do not sort on _id in ES cursors

cemorin:

In ES 7.4, sorting or aggregating on the '_id' field is deprecated, and [will be disallowed in future versions](https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-id-field.html)
This PR replaces the default sort on '_id' in ES cursors by a sort on 'id' (which is exactly the same value, but is part of the document body in the Enrichment and Workflow indexes).

***Note***
There used to be some very generic cursors in the `Akeneo\Tool\Elasticsearch` namespace, which theoretically could be used to iterate over any ES index. The cursors are not used anywhere in the PIM anymore (Enrichment and workflow now have their own implementations, (as well as AssetManager and ReferenceEntity), and other indexes don't rely on cursors to iterate over the index, so I decided to delete them

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc


Balamung (approved):

src/Akeneo/Tool/Bundle/ElasticsearchBundle/Cursor/Cursor.php
```diff
@@ -115,7 +115,7 @@ protected function getNextIdentifiers(array $esQuery, int $size = null)
             return [];
         }
 
-        $sort = ['_id' => 'asc'];
+        $sort = ['id' => 'asc'];
```
cemorin, line 118: Duplicate this class in Enrichment and use it with `id` only in enrichment

antoinearrive (approved):

src/Akeneo/Pim/Automation/DataQualityInsights/back/Infrastructure/Elasticsearch/Query/GetUpdatedEntityIdsQuery.php
```diff
@@ -60,7 +60,7 @@ public function since(\DateTimeImmutable $updatedSince, int $bulkSize): \Generat
         $searchQuery = [
             '_source' => ['id'],
             'size' => $bulkSize,
-            'sort' => ['identifier' => 'asc', '_id' => 'asc'],
+            'sort' => ['identifier' => 'asc', 'id' => 'asc'],
```
docteurklein, line 63: can there be a case where sorting just by `identifier` is not enough?

src/Akeneo/Tool/Bundle/ElasticsearchBundle/Cursor/FromSizeCursor.php
```diff
@@ -1,120 +0,0 @@
...
- * Bounded cursor to iterate over items where a start and a limit are defined.
- * Internally, this is implemented with the from/size pagination.
- * {@see https://www.elastic.co/guide/en/elasticsearch/reference/5.x/search-request-from-size.html}
- *
- * @author    Julien Janvier <jjanvier@akeneo.com>
- * @author    Marie Bochu <marie.bochu@akeneo.com>
- * @copyright 2017 Akeneo SAS (http://www.akeneo.com)
- * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
- */
-class FromSizeCursor extends AbstractCursor implements CursorInterface
```
ahocquard, line 19: for my culture, is it because we always use cursor that return identifiers now?

ahocquard (approved):

src/Akeneo/Pim/Automation/DataQualityInsights/back/Infrastructure/Elasticsearch/Query/GetUpdatedEntityIdsQuery.php
mmetayer: the primary sort is on 'identifier', and identifiers are soon going to be optional. The additional sort on 'id' allows having a deterministic sort order, thus a proper use of search_after

PR #17784, opened 2022-08-29
CXP-1248: Adds AttributeMultiSelectCriterion

tseho:

https://user-images.githubusercontent.com/1421130/187211913-72e53e68-4b4b-4e96-8d1c-ea96e30fe38b.mp4

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/components/catalogs/back/tests/Integration/Infrastructure/Validation/ProductSelection/AttributeCriterion/AttributeMultiSelectCriterionTest.php
+++ b/components/catalogs/back/tests/Integration/Infrastructure/Validation/ProductSelection/AttributeCriterion/AttributeMultiSelectCriterionTest.php
@@ -1,0 +1,382 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Akeneo\Catalogs\Test\Integration\Infrastructure\Validation\ProductSelection\AttributeCriterion;
+
+use Akeneo\Catalogs\Infrastructure\Validation\ProductSelection\AttributeCriterion\AttributeMultiSelectCriterion;
+use Symfony\Component\Validator\Validator\ValidatorInterface;
+
+/**
+ * @copyright 2022 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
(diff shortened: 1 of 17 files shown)


guillaumefer-akeneo (approved):

components/catalogs/back/tests/Integration/Infrastructure/Validation/ProductSelection/AttributeCriterion/AttributeMultiSelectCriterionTest.php
```diff
@@ -0,0 +1,382 @@
...
+                'attribute' => [
+                    'code' => 'materials',
+                    'type' => 'pim_catalog_multiselect',
+                    'options' => ['Cotton', 'Leather', 'Polyester', 'Wool'],
+                    'scopable' => true,
+                    'localizable' => true,
+                ],
+                'criterion' => [
+                    'field' => 'materials',
+                    'operator' => 'EMPTY',
```
willy-ahva, line 51: I would replace it with `Operator::IS_EMPTY`

willy-ahva (approved):

components/catalogs/back/tests/Integration/Infrastructure/Validation/ProductSelection/AttributeCriterion/AttributeMultiSelectCriterionTest.php
tseho: I'm conflicted, I did it initially and removed it.

Here, we are validating API input.
You have 2 situations:
- You use the constant. If for some reason, the value behind this constant change in the back, the test is still passing because it uses the constant, not the literal value.
- You use the literal value. If for some reason, the value behind this constant change in the back, the test is now failing.

I prefer the second scenario because the front expect the current contract to be respected.
I want my tests to light up like a christmas tree if the contract is broken by one side.
With the first scenario, if only the backend constant changes, all tests, in both frontend and backend are green but your application is broken because they expect a different contract in the API.

This is why I've used the literal, not the constant.

PR #17790, opened 2022-08-29
OCT-63: Accelerate integration tests with transactions⚡

tseho:

Test Transaction

**CI Catalogs before:**
Screenshot_2022-08-29_18-17-36

**CI Catalogs after:**
Screenshot_2022-08-29_18-23-41

**Akeneo laptop + Catalogs before:**
Screenshot_2022-08-31_11-31-26

**Akeneo laptop + Catalogs after:**
Screenshot_2022-08-31_11-35-24

**Akeneo laptop + Connectivity before:**
Screenshot_2022-08-30_12-15-38

**Akeneo laptop + Connectivity after:**
Screenshot_2022-08-30_12-15-52

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/components/catalogs/Makefile
+++ b/components/catalogs/Makefile
@@ -66,6 +66,7 @@ catalogs-unit-back:
 		--configuration $(CATALOGS_PATH)/back/ \
 		--bootstrap $(CATALOGS_BOOTSTRAP_PATH) \
 		--testsuite Catalogs_Unit \
+		--order-by random \
 		--log-junit var/tests/phpunit/phpunit_catalogs_unit.xml
 
 .PHONY: catalogs-integration-back
@@ -73,7 +74,9 @@ catalogs-integration-back:
 	APP_ENV=test $(PHP_RUN) vendor/bin/phpunit \
 		--configuration $(CATALOGS_PATH)/back/ \
 		--bootstrap $(CATALOGS_BOOTSTRAP_PATH) \
-		--testsuite Catalogs_Integration ${O}
+		--testsuite Catalogs_Integration \
+		--order-by random \
+		--log-junit var/tests/phpunit/phpunit_catalogs_integration.xml
 
 .PHONY: catalogs-acceptance-back
 catalogs-acceptance-back:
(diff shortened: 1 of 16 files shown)


masacc (approved):

guillaumefer-akeneo:
Why the prefix `EXPERIMENTAL` ?

components/catalogs/Makefile
```diff
@@ -66,14 +66,17 @@ catalogs-unit-back:
 		--configuration $(CATALOGS_PATH)/back/ \
 		--bootstrap $(CATALOGS_BOOTSTRAP_PATH) \
 		--testsuite Catalogs_Unit \
+		--order-by random \
```
guillaumefer-akeneo, line 69: For my information, why `random` ?

masacc: It's a strategy that is recommended when you play your test that executes tests in a random order instead of playing them always in the same order. Why is it recommended? because if you made an error in your setup/teardown you may have undetectable false positives.

tseho:
> Why the prefix EXPERIMENTAL ?

Because I don't have enough feedback on the effects.
If another team start using it but it break randomly their CI, they could blame us.
By calling it `EXPERIMENTAL`, it's obvious that it's "use at your own risk".

components/catalogs/Makefile
guillaumefer-akeneo: Crystal clear, thank you both for your replies.

guillaumefer-akeneo (approved):

tests/back/Integration/IntegrationTestsBundle/Loader/FixturesLoader.php
```diff
@@ -389,7 +397,22 @@ protected function dumpDatabase($filepath): void
      */
     protected function restoreDatabase($filepath): void
     {
-        $this->dbConnection->exec(file_get_contents($filepath));
+        try {
+            $this->dbConnection->exec(file_get_contents($filepath));
+        } catch (\Doctrine\DBAL\Exception $e) {
+            // The database is not empty, an integrity constraint violation was thrown on an unique key.
```
ahocquard, line 403:
```suggestion
           // There can be previous data left in database due to the execution of the previous test. 
           
           //   - When the previous test could failed early. If this previous test used a transaction in the same process as the current test, the previous transaction is therefore not rollback.
           //   - When the previous test does not use transaction or inserted data in another connection/subprocess (like a job), 
            // In that case, an integrity constraint violation is thrown on an unique key. The solution is to delete all data in the database and retry.
```

ahocquard (approved):

PR #17797, opened 2022-08-30
CPM-715: Manage products_uuid in not quantified associated product

pierallard:


cemorin (approved):

src/Akeneo/Pim/Enrichment/Component/Product/Updater/Setter/AssociationFieldSetter.php
```diff
@@ -136,6 +127,40 @@ private function updateAssociatedProducts(
...
+                $this->removeAssociatedProduct($owner, $associatedProduct, $associationType);
+            } else {
+                unset($productUuids[$index]);
+            }
+        }
+
+        foreach ($productUuids as $productUuid) {
+            $associatedProduct = null;
+            if (Uuid::isValid($productUuid)) {
+                $associatedProduct = $this->productRepository->find(Uuid::fromString($productUuid));
```
fchoquet, line 149: Not sure how big can `$productUuids` be and if it's a problem with real data, but there is a potential N+1 issue here (the DB is hit for every product)

It would be more efficient to load all the required associated products in one single SQL query before the loop (I haven't used doctrine recently but I think it supports an array of Ids and return an assoc with the id as key)

PR #17803, opened 2022-08-31
CXP-1271: GetProductsQuery

tseho:

The new query GetProductsQuery takes a catalog id, a search after, and a limit and returns normalized products where values are already filtered depending on the channels, locales & currencies also configured in your catalog.

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/components/catalogs/back/tests/Integration/Infrastructure/Persistence/GetProductUuidsQueryTest.php
+++ b/components/catalogs/back/tests/Integration/Infrastructure/Persistence/GetProductUuidsQueryTest.php
@@ -1,0 +1,169 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Akeneo\Catalogs\Test\Integration\Infrastructure\Persistence;
+
+use Akeneo\Catalogs\Application\Persistence\GetProductUuidsQueryInterface;
+use Akeneo\Catalogs\Domain\Operator;
+use Akeneo\Catalogs\Test\Integration\IntegrationTestCase;
+use Akeneo\Pim\Enrichment\Product\API\Command\UserIntent\SetEnabled;
+use Akeneo\Pim\Enrichment\Product\API\Command\UserIntent\SetTextValue;
+use Doctrine\DBAL\Connection;
+
+/**
+ * @copyright 2022 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+class GetProductUuidsQueryTest extends IntegrationTestCase
+{
+    private ?GetProductUuidsQueryInterface $query;
+    private ?Connection $connection;
+
+    protected function setUp(): void
+    {
+        parent::setUp();
+
+        $this->purgeDataAndLoadMinimalCatalog();
+
+        $this->query = self::getContainer()->get(GetProductUuidsQueryInterface::class);
+        $this->connection = self::getContainer()->get(Connection::class);
+    }
+
+    public function testItGetsMatchingProductsUuids(): void
+    {
+        $this->createUser('owner');
+        $this->logAs('owner');
+
+        $this->createCatalog('db1079b6-f397-4a6a-bae4-8658e64ad47c', 'Store US', 'owner');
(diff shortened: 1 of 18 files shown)


components/catalogs/back/tests/Integration/Infrastructure/Persistence/GetProductUuidsQueryTest.php
```diff
@@ -0,0 +1,169 @@
...
+        $this->createCatalog('db1079b6-f397-4a6a-bae4-8658e64ad47c', 'Store US', 'owner');
+        $this->enableCatalog('db1079b6-f397-4a6a-bae4-8658e64ad47c');
+        $this->setCatalogProductSelection('db1079b6-f397-4a6a-bae4-8658e64ad47c', [
+            [
+                'field' => 'enabled',
+                'operator' => Operator::EQUALS,
+                'value' => true,
+            ],
+        ]);
+        $this->createProduct('blue', [new SetEnabled(true)]);
```
guillaumefer-akeneo, line 47:
```suggestion
        $this->createProduct('tshirt-blue', [new SetEnabled(true)]);
```
And next

guillaumefer-akeneo (approved):

upgrades/schema/Version_7_0_20220830000000_add_catalog_product_values_filters.php
```diff
@@ -0,0 +1,37 @@
...
+        if ($schema->getTable('akeneo_catalog')->hasColumn('product_product_filter_channel')) {
+            $this->disableMigrationWarning();
+            return;
+        }
+
+        $this->addSql(
+            <<<SQL
+            ALTER TABLE akeneo_catalog
+            ADD product_value_filters JSON NOT NULL DEFAULT (JSON_OBJECT()) AFTER product_selection_criteria,
+            ALGORITHM=INPLACE, LOCK=NONE;
```
masacc, line 23: can you explain me these 2 options please ?

masacc (approved):

upgrades/schema/Version_7_0_20220830000000_add_catalog_product_values_filters.php
tseho: ezgif com-gif-maker(4)

I'm kidding, it's a good question.
When you alter a table and a column, you have 2 algorithms available in mysql.
`COPY` & `INPLACE`.
When `COPY` is used, mysql create a new version of the table with all the columns (including new ones), then copy the rows from the old to the new one. In most cases, it's fast.
When `INPLACE` is used, mysql rebuild the table by adding the new columns in place. It's slower.

But here is the tricky part: when `COPY` is used, mysql needs to `LOCK` the existing table while creating the new one and copying the data.
If the migration takes time but your application is still running, your queries on the table being migrated are locked. You can even timeout.
Previously, in the PIM, we didn't really care about those issues because the PIM was always turned off before the migrations.
Nowadays, we run ZeroDownTime deployments, migrations are executed when the PIM is still running.

With `ALGORITHM=INPLACE, LOCK=NONE`, we tell mysql to not lock the table, to not lock any query still trying to read/write. The migration is slower but the PIM is not interrupted.

masacc: OK, thank you for the clear explanation. I have 2 more questions:
1- what happens if we try to insert something in the table with data in the new column and the migration is still running? If nothing is locked, it can crash, can't it?
2- Should we use these options for every migration from now on?

willy-ahva: This answer could be in our Stackoverflow :)

tseho: >  1- what happens if we try to insert something in the table with data in the new column and the migration is still running? If nothing is locked, it can crash, can't it?

If the column doesn't exists yet, it will surely crash.

>  2- Should we use these options for every migration from now on?

That's kinda the status-quo but it's not official and not everybody does it.

PR #17821, opened 2022-09-01
CXP-1271: whistelist properties of normalized products

tseho:

In EE, products have metadatas. We don't want those so I added a whitelist of properties.

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/components/catalogs/back/src/Infrastructure/Persistence/GetProductsQuery.php
+++ b/components/catalogs/back/src/Infrastructure/Persistence/GetProductsQuery.php
@@ -136,10 +136,10 @@ class GetProductsQuery implements GetProductsQueryInterface
      */
     private function filterNormalizedProperties(array $products, array $whitelist): array
     {
-        $keys = array_flip($whitelist);
+        $keys = \array_flip($whitelist);
 
         foreach ($products as &$product) {
-            $product = array_intersect_key($product, $keys);
+            $product = \array_intersect_key($product, $keys);
         }
 
         return $products;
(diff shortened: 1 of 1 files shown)


components/catalogs/back/src/Infrastructure/Persistence/GetProductsQuery.php
```diff
@@ -114,6 +129,22 @@ private function filterCurrencies(array $products, ?array $currencies = null): a
...
+     * @param array<Product> $products
+     * @param array<string> $whitelist
+     * @return array<Product>
+     */
+    private function filterNormalizedProperties(array $products, array $whitelist): array
+    {
+        $keys = array_flip($whitelist);
+
+        foreach ($products as &$product) {
+            $product = array_intersect_key($product, $keys);
```
masacc, line 142:
```suggestion
        $keys = \array_flip($whitelist);

        foreach ($products as &$product) {
            $product = \array_intersect_key($product, $keys);
```

masacc (approved):

willy-ahva (approved):
Could you add a test?

tseho:
> Could you add a test?

Which test do you have in mind ?
It only break in EE conditions.

willy-ahva:
> > Could you add a test?
> 
> Which test do you have in mind ? It only break in EE conditions.

Maybe a test that checks the white list is well applied

tseho:
> Maybe a test that checks the white list is well applied

I understood that part, but how ? the class is already covered by integration tests. So we know that it didn't break the expected behavior.
Do you suggest to write unit tests on the class that will mock both GetConnectorProduct & its normalizer ?

tseho:
I'm merging because we need to unlock the CI, I will gladly continue this discussion and add tests if necessary.

PR #17822, opened 2022-09-01
OCT-117: Automatically launch database tests only when a migration is added

tseho:
**Introduction**

Running all integration tests on the existing migrations adds around **5min** to the CI workflow.
When we initially configured our workflow, we decided that we were going to run those tests only on demand.
It worked well for some time, but last week, I merged a broken migration without running those tests because I forgot to launch them. :boom: 

We need to prevent it from happening again.
But instead of running those tests in all workflows, we agreed it was time to implement `path-filtering`.

When using `path-filtering`, we can rely on the `git diff` between the latest commit and **master** to find which files are affected by the PR.

With the following mapping:
```
upgrades/.* run-database-tests true
```
Any PR affecting the `upgrades/` directory, either by adding or modifying a migration, will have the pipeline parameter `run-database-tests` at `true`.

We can then use this variable in our workflows to run the tests on the database only when needed.

---

You can check this other PR with a fake migration, to see the impact on our CI:
https://github.com/akeneo/pim-community-dev/pull/17882/files
https://app.circleci.com/pipelines/github/akeneo/pim-community-dev/40554/workflows/ed90a2ec-34e2-456d-a704-2ec4531ac3ab

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/.circleci/config/parameters.yml
+++ b/.circleci/config/parameters.yml
@@ -1,0 +1,4 @@
+parameters:
+    run-database-tests:
+        type: boolean
+        default: false
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,13 +4,13 @@ setup: true
 
 orbs:
     continuation: circleci/continuation@0.2.0
-    git-shallow-clone: guitarrapc/git-shallow-clone@2.4.0
+    path-filtering: circleci/path-filtering@0.1.3
 
 jobs:
     merge_config:
         executor: continuation/default
         steps:
-            - git-shallow-clone/checkout
+            - checkout
             - run:
                 name: Install yq (v4)
                 command: |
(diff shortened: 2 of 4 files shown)


masacc (approved):

guillaumefer-akeneo (approved):

willy-ahva (approved):

.circleci/config/parameters.yml
```diff
@@ -0,0 +1,4 @@
+parameters:
+    run-database-tests:
```
ahocquard, line 2:
```suggestion
    upgrade-directory-changed:
```
In my opinion, as for events with listener, it's the job/step decision to decide what is launched according to what is changed.
So I would just use the parameter as for an event: "this directory changed, deal with it".

.circleci/config/parameters.yml
```diff
@@ -0,0 +1,4 @@
+parameters:
```
ahocquard, line 1: I'm wondering if the name `parameters.yml` is not too generic. My original goal when splitting the CI was to split by responsibility, not by technical purpose.

what do you think about "path_filter_parameters.yml" to make it less generic?

ahocquard (approved):
This PR opens a lot of possibilities, thanks!

.circleci/config/parameters.yml
tseho: We have similar generic files in EE https://github.com/akeneo/pim-enterprise-dev/tree/master/.circleci/jobs
I'm not sure we should have one file by global parameter.

I understand that most files were intended to be split by responsibility and this is kinda special, given it's a shared technical configuration.

.circleci/config/parameters.yml
tseho: Damn, I merged too fast and forgot to handle this feedback.

tseho: If you want to change the name of the parameter, leave me an :+1: emoji and I will open a new PR.

PR #17824, opened 2022-09-01
CPM-695: Add uuid to old product endpoints

fchoquet:

Add uuid to the response of `api/rest/v1/products` endpoints

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/src/Akeneo/Connectivity/Connection/back/tests/CatalogBuilder/Enrichment/ProductLoader.php
+++ b/src/Akeneo/Connectivity/Connection/back/tests/CatalogBuilder/Enrichment/ProductLoader.php
@@ -10,6 +10,7 @@ use Akeneo\Tool\Bundle\ElasticsearchBundle\Client;
 use Akeneo\Tool\Component\StorageUtils\Saver\SaverInterface;
 use Akeneo\Tool\Component\StorageUtils\Updater\ObjectUpdaterInterface;
 use PHPUnit\Framework\Assert;
+use Ramsey\Uuid\Uuid;
 use Symfony\Component\Messenger\TraceableMessageBus;
 use Symfony\Component\Validator\Validator\ValidatorInterface;
 
@@ -42,16 +43,21 @@ class ProductLoader
         $this->messageBus = $messageBus;
     }
 
-    public function create($identifier, array $data): ProductInterface
+    public function createWithUuid($uuid, $identifier, array $data): ProductInterface
     {
-        $family = isset($data['family']) ? $data['family'] : null;
+        $family = $data['family'] ?? null;
 
-        $product = $this->builder->createProduct($identifier, $family);
+        $product = $this->builder->createProduct($identifier, $family, $uuid);
         $this->update($product, $data);
 
         $this->messageBus->reset();
 
         return $product;
+    }
+
+    public function create($identifier, array $data): ProductInterface
+    {
+        return $this->createWithUuid(Uuid::uuid4()->toString(), $identifier, $data);
     }
 
     public function update(ProductInterface $product, array $data): void
--- a/src/Akeneo/Pim/Enrichment/Bundle/Controller/ExternalApi/ProductController.php
+++ b/src/Akeneo/Pim/Enrichment/Bundle/Controller/ExternalApi/ProductController.php
@@ -312,6 +312,7 @@ class ProductController
         }
 
         $data = $this->getDecodedContent($request->getContent());
+
         $violations = $this->validator->validate($data, new PayloadFormat());
         if (0 < $violations->count()) {
             $firstViolation = $violations->get(0);
--- a/tests/back/Pim/Enrichment/EndToEnd/Product/Product/ExternalApi/AbstractProductTestCase.php
+++ b/tests/back/Pim/Enrichment/EndToEnd/Product/Product/ExternalApi/AbstractProductTestCase.php
@@ -4,14 +4,20 @@ namespace AkeneoTest\Pim\Enrichment\EndToEnd\Product\Product\ExternalApi;
 
 use Akeneo\Pim\Enrichment\Component\Product\Model\ProductInterface;
 use Akeneo\Pim\Enrichment\Component\Product\Model\ProductModelInterface;
+use Akeneo\Pim\Enrichment\Component\Product\Repository\ProductRepositoryInterface;
+use Akeneo\Pim\Enrichment\Component\Product\Validator\UniqueValuesSet;
 use Akeneo\Pim\Enrichment\Product\API\Command\UpsertProductCommand;
 use Akeneo\Pim\Enrichment\Product\API\Command\UserIntent\UserIntent;
 use Akeneo\Pim\Enrichment\Product\API\ValueObject\ProductUuid;
+use Akeneo\Test\IntegrationTestsBundle\Helper\AuthenticatorHelper;
 use Akeneo\Tool\Bundle\ApiBundle\tests\integration\ApiTestCase;
+use Akeneo\Tool\Bundle\ElasticsearchBundle\Client as EsClient;
+use Akeneo\Tool\Component\StorageUtils\Cache\EntityManagerClearerInterface;
 use AkeneoTest\Pim\Enrichment\Integration\Normalizer\NormalizedProductCleaner;
 use PHPUnit\Framework\Assert;
 use Ramsey\Uuid\Uuid;
 use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\Messenger\MessageBusInterface;
 
 /**
  * @author    Marie Bochu <marie.bochu@akeneo.com>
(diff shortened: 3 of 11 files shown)


tests/back/Pim/Enrichment/EndToEnd/Product/Product/ExternalApi/AbstractProductTestCase.php
```diff
@@ -21,6 +24,26 @@ abstract class AbstractProductTestCase extends ApiTestCase
...
+        $this->get('pim_enrich.product.message_bus')->dispatch($command);
+
+        return $this->get('pim_catalog.repository.product')->findOneByIdentifier($identifier);
+    }
+
+    /**
+     * @param UserIntent[] $userIntents
+     * @deprecated Use createProductWithUuid instead and pass the uuid explicitly
+     */
     protected function createProduct(string $identifier, array $userIntents = []): ProductInterface
```
fchoquet, line 47: I tried to change the implementation of `createProduct` to call `createProductWithUuid` internally and reduce duplication, but it cased a "The value for attribute sku is being updated multiple times" error to happen in tests where we explicitly set the sku attribute. 
Changing it would cause a cascade of changes that will probably cause this PR to grow too big so I gave up this idea.

tests/back/Pim/Enrichment/EndToEnd/Product/Product/ExternalApi/AbstractProductTestCase.php
```diff
@@ -37,7 +60,22 @@ protected function createProduct(string $identifier, array $userIntents = []): P
     /**
      * @param UserIntent[] $userIntents
      */
-    protected function createVariantProduct(string $identifier, array $userIntents = []) : ProductInterface
+    protected function createVariantProductWithUuid(string $uuid, string $identifier, array $userIntents = []): ProductInterface
+    {
+        $this->createProductWithUuid($uuid, $identifier, $userIntents);
+
+        $this->getContainer()->get('pim_catalog.validator.unique_value_set')->reset();
```
pierallard, line 67: Just a little tip for next developers ; instead of writing `$this->getContainer()->get('pim_catalog.validator.unique_value_set')->reset()`, you should write
`$this->getUniqueValueSetValidator()->reset()`
and add a private function
```
private function getUniqueValueSetValidator(): ValidatorInterface 
{
  return $this->getContainer()->get('pim_catalog.validator.unique_value_set');
}
```

Why doing this ? Because PHPStorm is has no idea the method "reset" you're using comes from a ValidatorInterface. The day we will update this method or any method of this class, the refactor will be a lot easier for the next developer.
(Same for all the ->get() you added)

2nd remark : I think the ->getContainer() is useless (not sure)

tests/back/Pim/Enrichment/EndToEnd/Product/Product/ExternalApi/ListProducts/ListProductWithCompletenessEndToEnd.php
```diff
@@ -27,15 +27,15 @@ public function setUp(): void
         parent::setUp();
 
         // product complete, whatever the scope
-        $this->createProduct('product_complete', [
+        $this->createProductWithUuid('aaf518b2-f91e-40f1-a53a-78ce5e81a6f9', 'product_complete', [
```
pierallard, line 30: As uuids are a bit unreadable, I suggest you add constants in the header of the file when you use them for the same product.
For example, here you should have a private const PRODUCT_COMPLETE_UUID = '...', or a list of product uuids like
PRODUCT_UUIDS = [ 'product_complete' => '...', ... ]
Once this is done, the readability of the uuids should be a lot increased, because you know what you're talking about, instead of looking about what uuid it refers with a search in the file.

tests/back/Pim/Enrichment/EndToEnd/Product/Product/ExternalApi/ListProducts/SuccessListProductEndToEnd.php
```diff
@@ -33,23 +33,23 @@ protected function setUp(): void
         parent::setUp();
 
         // no locale, no scope, 1 category
-        $this->createProduct('simple', [
+        $this->createProductWithUuid('4cfd82f2-def8-4869-8008-eabdf658f57c', 'simple', [
```
pierallard, line 36: Same remark

tests/back/Pim/Enrichment/EndToEnd/Product/Product/ExternalApi/AbstractProductTestCase.php
fchoquet: I took this code from the `createVariantProduct` method without thinking too much about it. But yeah you're perfectly right, I'm going to refactor 👍

tests/back/Pim/Enrichment/EndToEnd/Product/Product/ExternalApi/ListProducts/ListProductWithCompletenessEndToEnd.php
fchoquet: Not sure here. I usually avoid injecting variables in expected json structures, otherwise what you are actually testing can quickly become obfuscated by the syntax. But I can make this change.

src/Akeneo/Connectivity/Connection/back/tests/CatalogBuilder/Enrichment/ProductLoader.php
```diff
@@ -42,18 +43,26 @@ public function __construct(
...
         return $product;
     }
 
+    /**
+     * @deprecated Use createWithUuid instead and pass the uuid explicitly
+     */
+    public function create($identifier, array $data): ProductInterface
+    {
+        return $this->createWithUuid(Uuid::uuid4()->toString(), $identifier, $data);
+    }
```
mmetayer, line 64: I wouldn't mark it as deprecated, as it's a perfectly legitimate use case to create a product without specifying its uuid (e.g from the UI).

src/Akeneo/Pim/Enrichment/Bundle/Controller/ExternalApi/ProductController.php
```diff
@@ -312,6 +312,13 @@ public function partialUpdateAction(Request $request, $code): Response
         }
 
         $data = $this->getDecodedContent($request->getContent());
+
+        // We are not supposed to update the uuid, so there should be no uuid in the payload.
+        // But on the other end an implicit rule in our APIs states that we should be able to call
+        // the PATCH endpoint with the response returned by the corresponding GET endpoint (which contains the uuid)
+        // For this reason we simply ignore the passed uuid without triggering an "unexpected attribute" violation
+        unset($data['uuid']);
```
mmetayer, line 320: Perfectly valid to do that here, but it could also be directly set in the ProductUpdater, which is already configured to ignore properties such as `created`, `updated`, `completenesses`, etc.
The CE definition if ProductUpdater is already configured to ignore the uuid, but the EE definitions are not

tests/back/Pim/Enrichment/EndToEnd/Product/Product/ExternalApi/AbstractProductTestCase.php
```diff
@@ -21,6 +24,26 @@ abstract class AbstractProductTestCase extends ApiTestCase
     /**
      * @param UserIntent[] $userIntents
      */
+    protected function createProductWithUuid(string $uuid, string $identifier, array $userIntents = []): ProductInterface
```
mmetayer, line 27: maybe here we could omit the identifier argument, and force the caller to include the SetIdentifierValue intent in the userIntents argument (it will ease the transition once the identifier value is rendered optional)

tests/back/Pim/Enrichment/EndToEnd/Product/Product/ExternalApi/AbstractProductTestCase.php
```diff
@@ -21,6 +24,26 @@ abstract class AbstractProductTestCase extends ApiTestCase
...
+        $command = UpsertProductCommand::createWithUuid(
+            $this->getUserId('admin'),
+            ProductUuid::fromUuid(Uuid::fromString($uuid)),
+            // Because in the underlying function uuid and identifier are mutually exclusive we pass the latest as a user intent
+            array_merge($userIntents, [new SetIdentifierValue('sku', $identifier)])
+        );
+
+        $this->get('pim_enrich.product.message_bus')->dispatch($command);
+
+        return $this->get('pim_catalog.repository.product')->findOneByIdentifier($identifier);
```
mmetayer, line 40:
```suggestion
        return $this->get('pim_catalog.repository.product')->find($uuid);
```

tests/back/Pim/Enrichment/EndToEnd/Product/Product/ExternalApi/AbstractProductTestCase.php
mmetayer: anyway I believe this method is not useful (nor is the original `createProductVariant()` one), because a variant product is a product.
the only difference being the refreshing of the index, the clear of the cache and the reset of the uniquevalueset validator, I would move these 3 calls at the end of the `createProduct()` method, just before the retuen statement

mmetayer (approved):
a few minor comments, but GTM

cemorin (approved):

tseho (approved):

PR #17861, opened 2022-09-06
RAB-1038: Fix exception when we try to override a file with a move

ghoussard:

https://akeneo.atlassian.net/jira/software/projects/RAB/boards/282?selectedIssue=RAB-1038

**Issue :** it is not related to media but to all files (we didn't see it for data files because her file name are often dynamic).
When we launch 2 same exports consecutively, we encounter the following error : FlySystem try to move the created tmp file to his final destination but there is already a file at this location (added by the previous export) so it fails. 

Also, this issue is related to the usage of SFTP protocol :
Indeed, move operation does not exists on SFTP, it perform a rename instead. That means we cannot move file into a path that already exists, it explicitly indicated as an error by the [SFTP specification](https://datatracker.ietf.org/doc/html/draft-ietf-secsh-filexfer-02#section-6.5) : `It is an error if there already exists a file with the name specified by newpath.`

This change fixes the issue but seems to decrease significantly transfer performance. 😕 

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc


src/Akeneo/Platform/Bundle/ImportExportBundle/Infrastructure/StorageClient/TransferFile.php
```diff
@@ -31,8 +31,21 @@ public function transfer(
...
+    /**
+     * @param resource $stream
+     */
+    private function writeOnDestinationFilesystem(StorageClientInterface $destinationFilesystem, string $destinationFilePath, $stream): void
+    {
         $temporaryDestinationFilePath = $this->getTemporaryDestinationFilePath($destinationFilePath);
         $destinationFilesystem->writeStream($temporaryDestinationFilePath, $stream);
+
+        if ($destinationFilesystem->fileExists($destinationFilePath)) {
+            $destinationFilesystem->delete($destinationFilePath);
```
axelhms, line 46: We need to check with Felix but maybe we could use a timestamp in the file already exists from a previous job ? For example if `bag.jpg` already exists, we could create a `bag-07092022.jpg` 

Not sure about the impact tho

axelhms (approved):

src/Akeneo/Platform/Bundle/ImportExportBundle/Infrastructure/StorageClient/TransferFile.php
ghoussard: It could be a technical solution but it will breaks the business aspect : user cannot choose his filename 😕

ghoussard:
Refresh

PR #17883, opened 2022-09-08
CXP-1270: GET /catalogs/{id}/products

masacc:
image

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/components/catalogs/back/src/ServiceAPI/Query/GetProductsQuery.php
+++ b/components/catalogs/back/src/ServiceAPI/Query/GetProductsQuery.php
@@ -1,0 +1,76 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Akeneo\Catalogs\ServiceAPI\Query;
+
+use Symfony\Component\Validator\Constraints as Assert;
+
+/**
+ * @copyright 2022 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ *
+ * @phpstan-type ProductValue array{
+ *      scope: string|null,
+ *      locale: string|null,
+ *      data: mixed,
+ * }
+ *
+ * @phpstan-type Product array{
+ *      uuid: string,
+ *      enabled: boolean,
+ *      family: string,
+ *      categories: array<string>,
+ *      groups: array<string>,
+ *      parent: string|null,
+ *      values: array<string, ProductValue>,
+ *      associations: array<string, array{groups: array<string>, products: array<string>, product_models: array<string>}>,
+ *      quantified_associations: array<string, array{products: array<string>, product_models: array<string>}>,
+ *      created: string,
+ *      updated: string,
+ * }
+ *
+ * @implements QueryInterface<array<Product>>
+ */
+class GetProductsQuery implements QueryInterface
+{
+    public function __construct(
+        #[Assert\NotBlank]
+        #[Assert\Uuid]
+        private string $catalogId,
+        #[Assert\Uuid]
+        private ?string $searchAfter = null,
+        #[Assert\Range(min: 1, max: 1000)]
+        private int $limit = 100,
+        #[Assert\DateTime(\DateTimeInterface::ATOM, 'ISO 8601 format is required.')]
+        private ?string $updatedAfter = null,
+        #[Assert\DateTime(\DateTimeInterface::ATOM, 'ISO 8601 format is required.')]
+        private ?string $updatedBefore = null,
+    ) {
+    }
+
+    public function getCatalogId(): string
+    {
+        return $this->catalogId;
+    }
+
+    public function getSearchAfter(): ?string
+    {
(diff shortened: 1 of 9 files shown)


components/catalogs/back/src/ServiceAPI/Query/GetProductsQuery.php
```diff
@@ -0,0 +1,76 @@
...
+ *      family: string,
+ *      categories: array<string>,
+ *      groups: array<string>,
+ *      parent: string|null,
+ *      values: array<string, ProductValue>,
+ *      associations: array<string, array{groups: array<string>, products: array<string>, product_models: array<string>}>,
+ *      quantified_associations: array<string, array{products: array<string>, product_models: array<string>}>,
+ *      created: string,
+ *      updated: string,
+ * }
```
masacc, line 31: Should we use a model like Catalog endpoints instead of this array?

guillaumefer-akeneo (approved):

components/catalogs/back/src/ServiceAPI/Query/GetProductsQuery.php
willy-ahva: As discussed yes we should but let's do it later.

components/catalogs/back/tests/Integration/Infrastructure/Controller/Public/GetProductIdentifiersActionTest.php
```diff
@@ -11,6 +11,9 @@
 /**
  * @copyright 2022 Akeneo SAS (http://www.akeneo.com)
  * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ *
+ * @covers \Akeneo\Catalogs\Infrastructure\Controller\Public\GetProductIdentifiersAction
+ * @covers \Akeneo\Catalogs\Application\Handler\GetProductIdentifiersHandler
```
willy-ahva, line 16: Quentin told me in another PR: "@covers tells phpunit, "when you run this test, only compute the code coverage for this classes. If any other class is executed, do NOT count them as covered."

This is really valuable when you have integration tests that execute of lot of services but you didn't intend to cover random classes.

Why it's valuable ? Because if you compute the total coverage, you could be misled. "Oh, this really important piece of business code is already covered by tests, we are all good !"


So you'll understand that GetProductIdentifiersHandler is not needed here.

components/catalogs/back/tests/Integration/Infrastructure/Controller/Public/GetProductUuidsActionTest.php
```diff
@@ -12,6 +12,9 @@
 /**
  * @copyright 2022 Akeneo SAS (http://www.akeneo.com)
  * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ *
+ * @covers \Akeneo\Catalogs\Infrastructure\Controller\Public\GetProductUuidsAction
+ * @covers \Akeneo\Catalogs\Application\Handler\GetProductUuidsHandler
```
willy-ahva, line 17: GetProductUuidsHandler is not needed here.

components/catalogs/back/tests/Integration/Infrastructure/Controller/Public/GetProductsActionTest.php
```diff
@@ -0,0 +1,145 @@
...
+use PHPUnit\Framework\Assert;
+use Ramsey\Uuid\Uuid;
+use Symfony\Bundle\FrameworkBundle\KernelBrowser;
+
+/**
+ * @copyright 2022 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ *
+ * @covers \Akeneo\Catalogs\Infrastructure\Controller\Public\GetProductsAction
+ * @covers \Akeneo\Catalogs\Application\Handler\GetProductsHandler
```
willy-ahva, line 18: * @covers \Akeneo\Catalogs\Application\Handler\GetProductsHandler
 i not needed here

willy-ahva (approved):

components/catalogs/back/tests/Integration/Infrastructure/Controller/Public/GetProductIdentifiersActionTest.php
masacc: Totally agree... but we've done like this in all other integration tests that covers Action (except the 3 ones I updated here). And as it is an integration test and as the business logic should be mostly in handlers and the action should behave like a "passe-plat", I think we can reasonably consider adding `@cover` on handler in those tests. What do you think about it?

willy-ahva: Let's do it, i'm in!

RiffVentura (approved):

PR #17887, opened 2022-09-09
CPM-522: Fix coupling detector

pierallard:
**Unused coupling ?**

When a developer adds a dependency in a namespace, he have to add it in a dependency list.
A big job was done to plan how to remove (or decrease) these dependencies.
The issue is that the already-fixed dependencies are not always removed from the dependency list.
Having this dependency add some dirt and does not help us on focusing on a removal of a dependency.
This PR adds a step in the coupling detector step, to ensure there are no unused dependency on any context.

This PR
- Avoid to list all directories except "Test" one on Enrichment one
- Adds a step to check unused coupling
- Removes unused coupling

### DIFF
--- a/src/Akeneo/Connectivity/Connection/back/tests/.php_cd.php
+++ b/src/Akeneo/Connectivity/Connection/back/tests/.php_cd.php
@@ -142,7 +142,6 @@ $rules = [
             'Akeneo\UserManagement\Component\Repository\UserRepositoryInterface',
             'Akeneo\UserManagement\Component\Repository\GroupRepositoryInterface',
 
-            'Akeneo\Platform\Bundle\FeatureFlagBundle\FeatureFlag',
             'Akeneo\Platform\Bundle\FrameworkBundle\Service\PimUrl',
             'Akeneo\Platform\Component\EventQueue\BulkEventInterface',
             'Akeneo\Platform\Component\EventQueue\EventInterface',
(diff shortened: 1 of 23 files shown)


BitOne (approved):

src/Akeneo/Connectivity/Connection/back/tests/.php_cd.php
```diff
@@ -261,7 +257,7 @@
             'Symfony\Component\Validator\ConstraintValidator',
             'Symfony\Component\Validator\Exception\UnexpectedTypeException',
         ]
-    )->in('Akeneo\Connectiettings/Command/CreateConnectionHandler.php vity\Connection\Application\Settings'),
```
tseho, line 264: wtf

AymericPlanche: without any notice/warning of phpcd itself ?

tseho: https://github.com/akeneo/php-coupling-detector/blob/8141a08421c114f23ad8ec4ee307dea18f643377/src/Domain/Rule.php#L73

It only test if the analyzed class has an namespace starting with the string passed in `in()`.

Not sure if this is phpcd fault here, it should have been at least detected during the code review of the change.

cemorin (approved):

tseho (approved):

PR #17890, opened 2022-09-09
CXP-1255: remove pagination on locales filter

guillaumefer-akeneo:

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc


components/catalogs/front/src/hooks/useLocales.ts
```diff
@@ -1,5 +1,5 @@
 import {useQuery} from 'react-query';
-import {Locale} from '../models/Locale';
+import {Locale} from '../components/ProductSelection/models/Locale';
```
masacc, line 2: you should not move this `useLocales` hook in this directory for 2 reasons:
- it uses a model in a more specific namespace
- I think in this directory we should only have "public" hooks that are meant to be included in some external front components as a catalog "vendor". Locale is not part of the catalog business.

tseho: They have to share this hook between 2 majors components.
We have 2 solutions:
- we copy/paste the hook/model in both
- we move it to a shared namespace

Given the fact that locales are not really specific to either business, I think it's ok to move it here.

RiffVentura: I'm leaning more towards moving the model and hook higher (we could also not export what we deem unnecessary through an index file)

RiffVentura (approved):

components/catalogs/front/src/hooks/useLocales.ts
guillaumefer-akeneo: Ok, I moved the Locale on the higher/shared models directory.

masacc (approved):

components/catalogs/front/src/hooks/useLocales.ts
willy-ahva: The Locale model must be moved to `src/model/` you'll see a channel.d.ts in the same folder.

willy-ahva (approved):

willy-ahva (approved):

PR #17897, opened 2022-09-12
OCT-123: Forbid migrations without integration test

tseho:

Recently, we merged a migration without integration tests, because it was really simple.
Adding a column to an existing table, what could go wrong ? right ?

Of course, we broke something.

Here, instead of leaving the decision of creating an integration test on a new migration to the appreciation of its author, we create a simple rule with phpstan that will prevent you from doing the same mistake.

I've included a baseline of all existing errors reported by phpstan to only apply this rule to new migrations.

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc


masacc (approved):

upgrades/baseline.neon
```diff
@@ -0,0 +1,376 @@
+parameters:
+	ignoreErrors:
+		-
+			message: "#^Migration must be covered by an integration test$#"
```
ahocquard, line 4: could you remove migration up to 5.0 included instead and keep only 6.0/7.0 exception?

make-file/test.mk
```diff
@@ -40,6 +41,10 @@ lint-back:
 	# Cache was created with debug enabled, removing it allows a faster one to be created for upcoming tests
 	$(DOCKER_COMPOSE) run --rm php rm -rf var/cache/dev
 
+.PHONY: migration-lint-back
```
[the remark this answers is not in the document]
tseho: I've used phpstan level 5, it will also check for other things:

```
basic checks, unknown classes, unknown functions, unknown methods called on $this, wrong number of arguments passed to those methods and functions, always undefined variables
possibly undefined variables, unknown magic methods and properties on classes with __call and __get
unknown methods checked on all expressions (not just $this), validating PHPDocs
return types, types assigned to properties
basic dead code checking - always false instanceof and other type checks, dead else branches, unreachable code after return; etc.
checking types of arguments passed to methods and functions
```

upgrades/baseline.neon
tseho: Do you mean "delete all the migrations up to 5.0" or "exclude migrations up to 5.0 from phpstan" ?

ahocquard: delete all the migrations up to 5.0 included

tseho: Ok chef

willy-ahva (approved):

PR #17898, opened 2022-09-12
Use APP_DEBUG env var for test Kernel

jmleroux:
This PR allows you to run integration tests without using the cache container.
It can be useful when working on services DI definition by launching the integration test. Without this PR, you have to clear the cache each time you modify the DI, which is a real pain... :cry: 

You can activate the debug several ways:

By adding the env var in the Docker command line:

```bash
APP_ENV=test docker-compose run --rm -e APP_DEBUG=1 php vendor/bin/phpunit src/Akeneo/..../MyTestIntegration.php
```

Or by adding the APP_DEBUG in a local test env file named `.env.test.local`

### DIFF
--- a/tests/back/Integration/TestCase.php
+++ b/tests/back/Integration/TestCase.php
@@ -37,7 +37,7 @@ abstract class TestCase extends KernelTestCase
      */
     protected function setUp(): void
     {
-        $this->testKernel = static::bootKernel(['debug' => false]);
+        $this->testKernel = static::bootKernel(['debug' => (bool)($_SERVER['APP_DEBUG'] ?? false)]);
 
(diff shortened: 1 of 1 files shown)


ahocquard:
1. By activating this variable, a test suite will leak. At the end, you will have a memory limit issue. So, even if it's deactivated by default, I'm wondering if it worths the time saved. You can easily lose several hours on a memory leak due to the activation of this variable.

2. You are not in the same condition as your CI, which was the original intent of our test DX.

3. I'm afraid no people except you will use it, because it's complex to setup and not documented

4. There are other test cases, meaning this is only partially implemented

niarcotyk (approved):

ahocquard (approved):

PR #17904, opened 2022-09-12
CPM-631: Be able to have a family without sku

cemorin:

Be able to have a family without sku

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc


src/Akeneo/Pim/Structure/Component/Updater/FamilyUpdater.php
```diff
@@ -68,6 +68,18 @@ public function update($family, array $data, array $options = [])
...
         }
 
+        // For imports, sku are automatically added to the attribute_requirements and data is directly sent to this update
+        // to keep this "automatic" behavior, we add sku on new families
+        if (null === $family->getCreated()) {
+            $channels = $this->channelRepository->findAll();
+            /** @var ChannelInterface $channel */
+            foreach ($channels as $channel) {
+                $channelCode = $channel->getCode();
+                $identifierCode = $this->attributeRepository->getIdentifierCode();
```
fchoquet, line 78: This does not take any parameter from the foreach loop, so it can be moved outside the loop (even if the `getIndentifierCode` function has an internal cache mechanism, this class is not supposed to know it)

src/Akeneo/Pim/Structure/Component/Updater/FamilyUpdater.php
```diff
@@ -68,6 +68,18 @@ public function update($family, array $data, array $options = [])
...
+                $data['attribute_requirements'][$channelCode] = \array_unique(\array_merge([$identifierCode], $data['attribute_requirements'][$channelCode] ?? []));
```
fchoquet, line 79: It's a good idea to use `array_values` with `array_unique` because `array_unique` turns indexed arrays into associative ones.  It can sometimes lead to hard to interpret bugs.

https://www.tehplayground.com/lRpOKh2ikB5F8ElC

fchoquet (commented):
From what I understand of the context it looks good (I would validate if I felt legitimate for that)
I've found a couple of very minor things.

pierallard (approved):

Beninho (approved):

PR #17909, opened 2022-09-12
OCT-121: Fix first failing test with empty cache

masacc:

In integration tests, we should use decorated service to override ones we need. As we can't use a global `config_test.yaml` file because it would impact all the PIM, we declare a dedicated Kernel class to our integration tests and manually decorate services in it.

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/components/catalogs/back/phpunit.xml.dist
+++ b/components/catalogs/back/phpunit.xml.dist
@@ -19,7 +19,7 @@
         <ini name="intl.error_level" value="0"/>
         <ini name="memory_limit" value="-1"/>
         <ini name="zend.enable_gc" value="1"/>
-        <server name="KERNEL_CLASS" value="Kernel" force="true" />
+        <server name="KERNEL_CLASS" value="Akeneo\Catalogs\Test\Integration\Kernel" force="true" />
     </php>
 
     <groups>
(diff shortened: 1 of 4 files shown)


tseho (commented):
Why didn't you go with the try/catch version proposed in the ticket ?
IMO, I'd rather not having to maintain our own kernel in integrations tests.

see https://github.com/akeneo/pim-community-dev/pull/17912/files

masacc:
> Why didn't you go with the try/catch version proposed in the ticket ? 

Because, instead of replaying the setup without really knowing what happens, I prefer to implement the right way to override services as it is advised by Symfony. Moreover, we are not sure that one day this way to override service will not throw another error or have another unknown side effect.

> IMO, I'd rather not having to maintain our own kernel in integrations tests.

IMO, this is not a big deal because there is this empty `build()` function in the Kernel class that is meant to be used to do this kind of thing.

willy-ahva:
> Why didn't you go with the try/catch version proposed in the ticket ? IMO, I'd rather not having to maintain our own kernel in integrations tests.
> 
> see https://github.com/akeneo/pim-community-dev/pull/17912/files

The Thomas version seems less hackish. Up to you guys!

guillaumefer-akeneo (approved):

RiffVentura (approved):

masacc:
After discussions with @tseho , we agree to use its solution (PR #17912) to avoid having a specific Kernel class to maintain because we do not have a clear vision of the short-term evolution of codebase repositories.

PR #17912, opened 2022-09-12
OCT-121: catch cold boot of the kernel

tseho:

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/components/catalogs/back/tests/Integration/IntegrationTestCase.php
+++ b/components/catalogs/back/tests/Integration/IntegrationTestCase.php
@@ -48,6 +48,20 @@ abstract class IntegrationTestCase extends WebTestCase
         static::bootKernel(['environment' => 'test', 'debug' => false]);
 
         $this->clock = new Clock();
+
+        try {
+            $this->overrideServices();
+        } catch (\Symfony\Component\DependencyInjection\Exception\InvalidArgumentException $e) {
+            static::bootKernel(['environment' => 'test', 'debug' => false]);
+            $this->overrideServices();
+        }
+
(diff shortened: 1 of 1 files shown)


components/catalogs/back/tests/Integration/IntegrationTestCase.php
```diff
@@ -48,6 +49,27 @@ protected function setUp(): void
...
 
         $this->clock = new Clock();
+
+        // When the container cache does not exist (var/cache is empty for exemple),
+        // Symfony build the container when the kernel is booted for the first time.
+        // During the first build of the container, some services, like the listeners, are initialized.
+        // Trying to override a service already initialized is forbidden.
+        // The alternative solution of overidding services in the configuration is not appliable to us, doing so
+        // would affect ALL tests of all contexts in the PIM.
+        // Instead, the dirty but working solution is to catcn this error, reboot the kernel and retry.
```
masacc, line 59:
```suggestion
        // When the container cache does not exist (var/cache is empty for example),
        // Symfony builds the container when the kernel is booted for the first time.
        // During the first build of the container, some services, like the listeners, are initialized.
        // Trying to override a service already initialized is forbidden.
        // The alternative solution of overriding services in the configuration is not appliable to us, doing so
        // would affect ALL tests of all contexts in the PIM.
        // Instead, the dirty but working solution is to catch this error, 
        // reboot the kernel (the cache already exists now so the services will not be initialized) and retry.
```

(I fixed 2-3 typos and add the between parenthesis precision in the last part)

masacc (approved):

PR #17915, opened 2022-09-13
OCT-125: Add acceptance test for /products, /product-identifiers and /product-uuids catalog endpoints

masacc:

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/components/catalogs/back/tests/Acceptance/ApiContext.php
+++ b/components/catalogs/back/tests/Acceptance/ApiContext.php
@@ -61,8 +61,9 @@ class ApiContext implements Context
 
     /**
      * @Given an existing catalog
+     * @Given a disabled catalog
      */
-    public function anExistingCatalog(): void
+    public function aDisabledExistingCatalog(): void
     {
         $connectedAppUserIdentifier = $this->getConnectedApp()->getUsername();
         $this->authentication->logAs($connectedAppUserIdentifier);
@@ -102,25 +103,9 @@ class ApiContext implements Context
     }
 
     /**
-     * @Given the catalog is enabled
+     * @Given an enabled catalog sets up with a product selection criteria
      */
-    public function theCatalogIsEnabled(): void
-    {
-        $connectedAppUserIdentifier = $this->getConnectedApp()->getUsername();
-        $this->authentication->logAs($connectedAppUserIdentifier);
-
-        $this->upsertCatalogQuery->execute(
-            'db1079b6-f397-4a6a-bae4-8658e64ad47c',
-            'Store US',
-            $connectedAppUserIdentifier,
-            true,
-        );
-    }
-
-    /**
-     * @Given a catalog sets up with a product selection
-     */
-    public function aCatalogSetsUpWithAProductSelection(): void
+    public function anEnabledCatalogSetsUpWithAProductSelectionCriteria(): void
     {
         $connectedAppUserIdentifier = $this->getConnectedApp()->getUsername();
         $this->authentication->logAs($connectedAppUserIdentifier);
@@ -374,9 +359,9 @@ class ApiContext implements Context
     }
 
     /**
-     * @When the external application retrieves the product identifiers using the API
+     * @When the external application retrieves the product's identifiers using the API
      */
-    public function theExternalApplicationRetrievesTheProductIdentifiersUsingTheApi(): void
+    public function theExternalApplicationRetrievesTheProductsIdentifiersUsingTheApi(): void
     {
         $this->authentication->logAs($this->getConnectedApp()->getUsername());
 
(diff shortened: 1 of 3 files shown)


components/catalogs/features/api.feature
```diff
@@ -34,3 +34,47 @@ Feature:
...
+    And the catalog is enabled
+    And the catalog has product selection criteria
+    And the following products exist:
+      | uuid                                 | identifier    | enabled |
+      | 21a28f70-9cc8-4470-904f-aeda52764f73 | t-shirt blue  | 0       |
+      | 62071b85-67af-44dd-8db1-9bc1dab393e7 | t-shirt green | 1       |
+    When the external application retrieves the product identifiers using the API
+    Then the response should contain the following product identifiers:
+      | identifier    |
+      | t-shirt green |
```
willy-ahva, line 50:
```suggestion
    Given a catalog set up with a product selection
    When the external application retrieves the product identifiers using the API
    Then the response should contain only the product identifiers from the selection
```
As we are trying to describe the general behavior of the feature, I wonder if something more generic could be better.
I think we don't mind what is inside in the details, as soon as it is simple to read and understand quickly.

RiffVentura (approved):

components/catalogs/back/tests/Acceptance/ApiContext.php
```diff
@@ -283,4 +372,109 @@ public function theCatalogShouldBeUpdatedInThePim(): void
...
+    /**
+     * @Then the response should contain only the product uuids from the selection
+     */
+    public function theResponseShouldContainOnlyTheProductUuidsFromTheSelection(): void
+    {
+        $payload = \json_decode($this->response->getContent(), true, 512, JSON_THROW_ON_ERROR);
+
+        Assert::assertCount(2, $payload['_embedded']['items']);
+
+        $expectedUuids = [
```
willy-ahva, line 436: I wonder if a stateful context could be better. When you do Given a catalog sets up with a product selection, you create products and keep them in $this->expectedProductUuids.
Keep it simple like this for now, but please add a comment like:
`// Comes from the step aCatalogSetsUpWithAProductSelection`

If needed will make it stateful later.

willy-ahva (approved):

components/catalogs/features/api.feature
```diff
@@ -34,3 +34,21 @@ Feature:
     Given several existing catalogs
     When the external application retrieves the catalogs using the API
     Then the response should contain catalogs details
+
+  @database
+  Scenario: Get products identifiers
+    Given a catalog sets up with a product selection
```
tseho, line 40: We want to expose in the acceptance tests what is the purpose of our feature, how it's going to be used and what is the behavior from a business/product POV.

The first lot is clear, "I can create, update, delete a catalog."
The second step, the one you did, is simple too: "I can fetch products from a catalog".
But we have a missing piece, really important from the product POV.

You need to enable the catalog. And what happens if it's not enabled ?
So, I would be more explicit on the state of the catalog:

```suggestion
    Given an enabled catalog with product selection criteria
```

This way, you can also create 3 more tests.
```
Given a disabled catalog
When the external application retrieves the product identifiers using the API
Then the response should contain an empty list
```

Why additional scenarios ? IMO, this is a really critical decision that should be clearly documented.
In the future, If the response behavior changes when a catalog is disabled, you not only change a few integration tests, you also change the acceptance tests => "Yes, we aware that we are radically changing the designed behavior and how users will use it".

masacc: I hesitated to explicitly mention the "enable" state of the catalog. I have not because I focused on the main scenarios but I agree with you and I'm going to add these new ones.

tseho: Otherwise, this is a really good PR :clap:

components/catalogs/back/tests/Acceptance/ApiContext.php
```diff
@@ -24,34 +31,48 @@ class ApiContext implements Context
...
+    }
+
+    private function getConnectedAppClient(): KernelBrowser
+    {
+        return $this->client ??= $this->authentication->createAuthenticatedClient($this->getConnectedApp());
+    }
+
+    /**
+     * @Given an existing catalog
+     * @Given a disabled catalog
```
masacc, line 64: @willy-ahva @tseho  What do you think of having 2 `@Given` annotations for the same function ? It makes sense to me because it's the same logic but it is used in 2 different contexts. However, I'm not sure it's a good practice...

tseho: I think it's not critical but should be avoided.

components/catalogs/features/api.feature
```diff
@@ -34,3 +34,39 @@ Feature:
     Given several existing catalogs
     When the external application retrieves the catalogs using the API
     Then the response should contain catalogs details
+
+  @database
+  Scenario: Get product's identifiers with an enabled catalog
```
tseho, line 39:
```suggestion
  Scenario: Get product's identifiers from an enabled catalog
```
WDYT ?

PR #17942, opened 2022-09-16
GRF-349 fix DSM InputTextArea : richtext : selected dropdown items in…

FSAkeneo:
… toolbar where styled link links in PIM (purple)

When embedded in PIM,
an InputTextArea from DSM havingprop isRichText set to true
had some purple texts in the wysiwyg toolbar UI because the style of the link in PIM was applied.

A dedicated specific selector prevents the color of element being set by the PIM above.

- [ ] PM Validation (Story)

### DIFF
--- a/front-packages/akeneo-design-system/src/components/Input/TextAreaInput/TextAreaInput.tsx
+++ b/front-packages/akeneo-design-system/src/components/Input/TextAreaInput/TextAreaInput.tsx
@@ -29,6 +29,10 @@ const CommonStyle = css<{readOnly: boolean; invalid: boolean} & AkeneoThemedProp
   &:focus-within {
     box-shadow: 0 0 0 2px ${getColor('blue', 40)};
   }
+
+  a.rdw-dropdown-selectedtext > span {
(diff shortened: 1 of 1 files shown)


front-packages/akeneo-design-system/src/components/Input/TextAreaInput/TextAreaInput.tsx
```diff
@@ -29,6 +29,10 @@ const CommonStyle = css<{readOnly: boolean; invalid: boolean} & AkeneoThemedProp
   &:focus-within {
     box-shadow: 0 0 0 2px ${getColor('blue', 40)};
   }
+
+  a.rdw-dropdown-selectedtext > span {
+    color: ${({readOnly}) => (readOnly ? getColor('grey', 100) : getColor('grey', 140))};
```
olivierpontier, line 34: I am not sure the readOnly condition is useful because the rich text toolbar is hidden when the component is in ReadOnly state

olivierpontier (approved):

amigaire (approved):

PR #17962, opened 2022-09-20
PIM-10634: Fix media filter values normalizer to be case insensitive

cemorin:

Add normalizer to the media type filter in datagrid.

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc


tests/back/Pim/Enrichment/Integration/PQB/ProductQueryBuilderIntegration.php
```diff
@@ -140,6 +159,7 @@ protected function setUp(): void
                 ),
                 new SetImageValue('a_localizable_image', null, 'en_US', $this->getFileInfoKey($this->getFixturePath('akeneo.jpg'))),
                 new SetImageValue('a_localizable_image', null, 'fr_FR', $this->getFileInfoKey($this->getFixturePath('akeneo.jpg'))),
+                new SetImageValue('an_image', null, null, $this->getFileInfoKey($this->getFixturePath('akeneo.jpg'))),
```
fchoquet, line 162: Just to confirm it works both ways, we could add a capital letter in the file name (for instance the file name is `Akeneo.jpg` and we search for `aKENEO`

fchoquet (commented):
Looks good to me as long as we accept that there is a different behavior for products indexed before and after this change (until we run the migration)

cemorin:
Seen with @mmetayer, we only need an update of the mapping to make it work

tests/back/Pim/Enrichment/Integration/PQB/ProductQueryBuilderIntegration.php
cemorin: All the fixture files in the PIM are in lowercase, isn't the new test enough? (I updated it)

fchoquet: Well, under the hood, the way it works is:

```
(index time) "AKeno.jpg" --> normalized to [akeneo] token
                                               |
                                           both match!
                                               |
(query time) "akeNEO"  --> normalized to  [akeneo] token
```

So, if the file name is already in its normalized form then the index time normalization is not really tested. That's maybe not a problem with the current implementation because there are other guarantees that the same normalizer is used both at index and query time, but in an ideal world it should be tested.

cemorin: I added a `Akeneo.jpg` fixture plus the test :+1:

fchoquet: Nice! 👍 

But I realise that there is another situation to test if you want to cover all the cases in the jira ticket: multiple words 😬 

Maybe you can rename your new picture `Akeneo Logo.jpg` and add a test to cover search of "akeneo logo"

cemorin: done :)

src/Akeneo/Tool/Bundle/ElasticsearchBundle/Command/UpdateMappingIndexCommand.php
```diff
@@ -89,6 +89,7 @@ public function execute(InputInterface $input, OutputInterface $output)
         }
 
         $io->success("All the indices listed above have been migrated");
+        return 0;
```
mmetayer, line 92: missing blank line

mmetayer (approved):

antoinearrive (approved):

pierallard (approved):

PR #17987, opened 2022-09-22
CXP-1299: fix experimental transaction in job test

tseho:

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/components/catalogs/back/tests/Integration/Infrastructure/Job/DisableCatalogsOnAttributeOptionRemovalTaskletTest.php
+++ b/components/catalogs/back/tests/Integration/Infrastructure/Job/DisableCatalogsOnAttributeOptionRemovalTaskletTest.php
@@ -6,20 +6,15 @@ namespace Akeneo\Catalogs\Test\Integration\Infrastructure\Job;
 
 use Akeneo\Catalogs\Domain\Operator;
 use Akeneo\Catalogs\Test\Integration\IntegrationTestCase;
-use Akeneo\Test\IntegrationTestsBundle\Launcher\JobLauncher;
 
 class DisableCatalogsOnAttributeOptionRemovalTaskletTest extends IntegrationTestCase
 {
-    private ?JobLauncher $jobLauncher;
-
     protected function setUp(): void
     {
         parent::setUp();
 
+        $this->disableExperimentalTestDatabase();
         $this->purgeDataAndLoadMinimalCatalog();
-
-        $this->jobLauncher = self::getContainer()->get('akeneo_integration_tests.launcher.job_launcher');
-        $this->jobLauncher->flushJobQueue();
     }
 
     public function testItDisablesCatalogOnAttributeOptionRemoval(): void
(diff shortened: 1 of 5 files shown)


guillaumefer-akeneo (approved):

components/catalogs/back/tests/Integration/IntegrationTestCase.php
```diff
@@ -382,17 +387,6 @@ protected function enableCurrency(string $code): void
         self::getContainer()->get('pim_catalog.saver.currency')->save($currency);
     }
 
-    protected function assertCatalogIsDisabled(string $id): void
```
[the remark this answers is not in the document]
tseho: I've explained the idea behind this specific change in the slack thread.

willy-ahva (approved):

components/catalogs/back/tests/Integration/Infrastructure/Job/DisableCatalogsOnAttributeOptionRemovalTaskletTest.php
```diff
@@ -6,20 +6,15 @@
...
 
 class DisableCatalogsOnAttributeOptionRemovalTaskletTest extends IntegrationTestCase
 {
-    private ?JobLauncher $jobLauncher;
-
     protected function setUp(): void
     {
         parent::setUp();
 
+        $this->disableExperimentalTestDatabase();
```
masacc, line 16: Shouldn't you re-enable it at `teardown()` if "enabled" is the default status?

tseho: For the next test ?
no need, during the setUp() phase, we always boot a new kernel, all services are re-initialized. So the following test will have `ExperimentalTransactionHelper` with the default value coming from the environment variable.

masacc: OK, I proposed that because I was not sure about some side-effects so it was "ceinture-bretelles" style...

masacc (approved):

components/catalogs/back/tests/Integration/Infrastructure/Job/DisableCatalogsOnAttributeOptionRemovalTaskletTest.php
tseho: And if you ask, "Why are we explicitly doing rollback of the transaction in `teardown()` if it's throwed away by the next test ?"
For a simple reason.
If the rollback cannot happen because there was an autocommit during the test, your test will fail and you will know that there is an issue.

tseho:
FYI, I'm merging in your branch but the CI is still red, your test is not working in EE because `disable_catalogs_on_attribute_option_removal` seems to be unknown in EE.

PR #17988, opened 2022-09-22
[migration uuid] Add option to remove the algorithm inplace

pierallard:
The UUID migration was designed to be SAAS-compliant, so every ALTER TABLE methods use ALGORITHM INPLACE and LOCK NONE.
Today, some users want to run this migration with the server down.
We hope to speed up the migration by removing this inplace algorithm.

We added a new option '--lock-tables', set to true on the doctrine migrations.

### DIFF
--- a/src/Akeneo/Pim/Enrichment/Bundle/Command/MigrateToUuid/MigrateToUuidCommand.php
+++ b/src/Akeneo/Pim/Enrichment/Bundle/Command/MigrateToUuid/MigrateToUuidCommand.php
@@ -60,6 +60,7 @@ class MigrateToUuidCommand extends Command
         $this->addOption('dry-run', 'd', InputOption::VALUE_NEGATABLE, 'dry run', false);
         $this->addOption('with-stats', 's', InputOption::VALUE_NEGATABLE, 'Display stats (be careful the command is way too slow)', false);
         $this->addOption('wait-for-dqi', 'w', InputOption::VALUE_NEGATABLE, 'Wait for DQI job before starting', true);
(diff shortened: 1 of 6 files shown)


src/Akeneo/Pim/Enrichment/Bundle/Command/MigrateToUuid/MigrateToUuidCommand.php
```diff
@@ -60,6 +60,7 @@ protected function configure()
         $this->addOption('dry-run', 'd', InputOption::VALUE_NEGATABLE, 'dry run', false);
         $this->addOption('with-stats', 's', InputOption::VALUE_NEGATABLE, 'Display stats (be careful the command is way too slow)', false);
         $this->addOption('wait-for-dqi', 'w', InputOption::VALUE_NEGATABLE, 'Wait for DQI job before starting', true);
+        $this->addOption('algorithm-inplace', 'a', InputOption::VALUE_NEGATABLE, 'Use ALGORITHM=INPLACE for table migrations. This option does not lock the tables during migration.', true);
```
nmarniesse, line 63: I find it's strange to have an option true by default:
- if you run the command with `--algorithm-inplace` it uses the algorithm inplace (normal)
- but if you run the command without  `--algorithm-inplace` it's also uses the inplace
- you have to add `--algorithm-inplace=false` to disable it, it's not very intuitive

Considering the command should use the algorithm inplace by default (because SaaS is priority), is it better to have an option "--disable-inplace-algorithm" or "--lock-tables"?

I'm chipoting

nmarniesse (approved):

mmetayer (approved):

PR #18003, opened 2022-09-23
PH-268: Allow to purge  job executions older than 1 hour

jmleroux:
### The issue
When running the purge job execution, we first create a job execution.
If we set the purge limit to 0, all executions are purged, included the purge itself.
At the end of the job, we try to update the execution status of this purge, but the execution does not exist anymore.

### The solution
When running the purge as a job, we will have a new hard limit to one hour and we will not purge job executions of the last hour.
But we  will keep the possibility to purge all with a CLI command. Do do that, the CLI command will not launch a job anymore, but use the purger directly.

### DIFF
--- a/src/Akeneo/Tool/Bundle/BatchBundle/Persistence/Sql/GetJobExecutionIds.php
+++ b/src/Akeneo/Tool/Bundle/BatchBundle/Persistence/Sql/GetJobExecutionIds.php
@@ -7,9 +7,10 @@ namespace Akeneo\Tool\Bundle\BatchBundle\Persistence\Sql;
 use Akeneo\Tool\Component\Batch\Job\BatchStatus;
 use DateTime;
 use Doctrine\DBAL\Connection;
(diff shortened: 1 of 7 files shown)


src/Akeneo/Tool/Bundle/BatchBundle/Persistence/Sql/GetJobExecutionIds.php
```diff
@@ -7,24 +7,22 @@
 use Akeneo\Tool\Component\Batch\Job\BatchStatus;
 use DateTime;
 use Doctrine\DBAL\Connection;
-use Doctrine\DBAL\Driver\ResultStatement;
+use Doctrine\DBAL\Result;
```
jmleroux, line 10: To be able to use `fetchOne` nistead of the deprecated `fetchColumn` method

ahocquard (approved):
Perfect, thanks !

PR #18026, opened 2022-09-27
RAB-833: Job ServiceAPI

axelhms:

In this PR, I've added a service api to retrieve job instances.

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/src/Akeneo/Platform/Job/back/ServiceApi/JobInstance/FindJobInstance.php
+++ b/src/Akeneo/Platform/Job/back/ServiceApi/JobInstance/FindJobInstance.php
@@ -33,7 +33,7 @@ class FindJobInstance implements FindJobInstanceInterface
         $sql = <<<SQL
         SELECT
             job_instance.code,
-            job_instance.label,
+            job_instance.label
         FROM akeneo_batch_job_instance job_instance
         %s
         %s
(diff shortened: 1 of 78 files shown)


ghoussard (approved):

src/Akeneo/Platform/Job/back/ServiceApi/JobInstance/FindJobInstance.php
```diff
@@ -0,0 +1,103 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Akeneo\Platform\Job\ServiceApi\JobInstance;
```
ValentinMumble, line 5: I believe we should put the implementation in the Infrastructure folder, usually we also name it according to what implem it is, like `SqlFindJobInstance` but I'm not sure what is the rule for new `ServiceApi`s? What do you think @team?
```suggestion
namespace Akeneo\Platform\Job\Infrastructure\Query\JobInstance;
```

ValentinMumble (approved):

src/Akeneo/Platform/Job/back/ServiceApi/JobInstance/FindJobInstance.php
ghoussard: Agree with you. ServiceAPI only exposes contracts, implementation details should be hide in Infrastructure 👍🏻
I think this is designed like that in others Service API

ValentinMerlet (approved):
:pray:

PR #18047, opened 2022-09-29
CXP-1288: Completed placeholder with an illustration and text

RiffVentura:
Completed source placeholder using DSM components
image

### DIFF
--- a/components/catalogs/front/src/components/ProductMapping/components/SourcePlaceholder.tsx
+++ b/components/catalogs/front/src/components/ProductMapping/components/SourcePlaceholder.tsx
@@ -1,15 +1,49 @@
 import React, {FC} from 'react';
-import {SectionTitle} from 'akeneo-design-system';
+import styled from 'styled-components';
+import {SectionTitle, AttributesIllustration, Link, getColor, getFontSize} from 'akeneo-design-system';
 import {useTranslate} from '@akeneo-pim-community/shared';
+
+const Placeholder = styled.div`
+    display: flex;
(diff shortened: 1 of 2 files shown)


components/catalogs/front/src/components/ProductMapping/components/SourcePlaceholder.tsx
```diff
@@ -1,15 +1,49 @@
 import React, {FC} from 'react';
-import {SectionTitle} from 'akeneo-design-system';
+import styled from 'styled-components';
+import {SectionTitle, AttributesIllustration, Link, getColor, getFontSize} from 'akeneo-design-system';
 import {useTranslate} from '@akeneo-pim-community/shared';
 
+const Placeholder = styled.div`
```
ValentinMumble, line 6: Just for my knowledge, why do we need a different placeholder than the one of the DSM?
Doesn't it break the homogeneity principle of using the same visual components across the PIM?
If the placeholder component inside the DSM has wrong proportions or size values, for it needs to be changed everywhere, but we shouldn't need and definitely do not want multiple variations of the same component...

RiffVentura: You are right! I guess I was too hasty, I remade the component using DSM

masacc (approved):

masacc:
maybe you could update the description of the PR to match the new design using DSM Placeholder element.

guillaumefer-akeneo (approved):

PR #18058, opened 2022-10-03
RAB-1029: Migrate quantified association in PEF to uuid

StevenVAIDIE:

In order to prepare optional identifier, we need to handle the PEF with no identifier. In this PR, I migrate the quantified association management to uuid instead of identifier. Now the InternalAPI use uuid for quantified associations.

To do it I needed to :
- modify the standard format when we update the product from products to products
- I modified the normalize function of quantified link => now we returned the uuid and the identifier, then the normalizer are in charge to remove the uuid or the identifier depending on the context (InternalAPI or ExternalAPI)
- I changed the front to handle the uuid instead of the identifier

- [x] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc


nestorAkeneo:

src/Akeneo/Pim/Enrichment/Bundle/Resources/fonts/installed-fonts.json
```diff
@@ -0,0 +1,14 @@
+{
```
StevenVAIDIE, line 1: :eyes:

PR #18068, opened 2022-10-04
CPM-746 : Add PHPStan

Beninho:
- Add PHPStan
- Add rule for strict type declaration in PHP-CS-Fixer
- Update make file recipes


pierallard (approved):

Beninho:
The CI failed on PHPStan. We (chipmunks squad) have to talk about it.
Check conditions before call function with typed parameters OR call function with generic parameter and check types in functions.

Example : 
```IdentifierGeneratorId::fromString(strval($result['uuid'])),```
and
```public static function fromString(string $id): self```

OR
```IdentifierGeneratorId::fromString($result['uuid']),```
and
```public static function fromString(mixed $id): self```

PR #18070, opened 2022-10-04
RAB-1021: Add CreateJob ServiceApi

pierrejolly:

This PR add a new service API to create Job

### DIFF
--- a/src/Akeneo/Platform/Job/back/Application/CreateJobInstanceHandler.php
+++ b/src/Akeneo/Platform/Job/back/Application/CreateJobInstanceHandler.php
@@ -62,7 +62,7 @@ final class CreateJobInstanceHandler implements CreateJobInstanceHandlerInterfac
         return  $this->jobRegistry->get($jobInstance->getJobName());
     }
 
-    private function validateJob(JobInterface $job, JobParameters $jobParameters, JobInstance $jobInstance)
+    private function validateJob(JobInterface $job, JobParameters $jobParameters, JobInstance $jobInstance): void
     {
         $jobParametersViolations = $this->jobParametersValidator->validate($job, $jobParameters);
         if (0 < $jobParametersViolations->count()) {
--- a/src/Akeneo/Platform/Job/back/Application/CreateJobInstanceHandler.php
+++ b/src/Akeneo/Platform/Job/back/Application/CreateJobInstanceHandler.php
@@ -4,11 +4,10 @@ declare(strict_types=1);
 
 namespace Akeneo\Platform\Job\Application;
 
-use Akeneo\Platform\Job\ServiceApi\JobInstance\CreateJobInstanceCommand;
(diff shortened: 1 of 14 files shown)


src/Akeneo/Platform/Job/back/Application/CreateJobInstanceHandler.php
```diff
@@ -0,0 +1,64 @@
...
+
+        $jobParametersViolations = $this->jobParametersValidator->validate($job, $jobParameters);
+        if (0 < $jobParametersViolations->count()) {
+            throw new InvalidJobException($jobInstance->getCode(), $job->getName(), $jobParametersViolations);
+        }
+
+        $jobInstanceViolations = $this->validator->validate($jobInstance);
+        if (0 < $jobInstanceViolations->count()) {
+            throw new InvalidJobException($jobInstance->getCode(), $job->getName(), $jobInstanceViolations);
+        }
```
ghoussard, line 60: This method is too huge for me, we should directly see what the handler do when we read the `handle` function 🤔 
Something like that should be more readable :
```suggestion
        $jobInstance = $this->createJobInstance($command);
        $job = $this->getJob($jobInstance);
        $this->validateJob($jobInstance, $job);
```

ghoussard (approved):

src/Akeneo/Platform/Job/back/Application/CreateJobInstanceHandler.php
pierrejolly: Fixed 👍

src/Akeneo/Tool/Bundle/BatchBundle/Command/CreateJobInstanceCommand.php
```diff
@@ -94,17 +60,18 @@ protected function execute(InputInterface $input, OutputInterface $output): int
...
-        $jobInstance->setConnector($connector);
-        $jobInstance->setJobName($jobName);
-        $jobInstance->setCode($code);
-        $jobInstance->setLabel($label);
-        $jobInstance->setRawParameters($rawConfig);
-
-        /** @var JobInterface */
-        $job = $this->jobRegistry->get($jobInstance->getJobName());
-        if (null === $job) {
+        $command = new CreateJobInstanceCqrsCommand(
```
axelhms, line 63: What does Cqrs mean ?

axelhms (approved):

src/Akeneo/Tool/Bundle/BatchBundle/Command/CreateJobInstanceCommand.php
pierrejolly: It's short for "Command and Query Responsibility Segregation", it's the pattern we use. Usually we don't add this prefix, but in this case we are in another type of Command class (the one in Symfony CLI), so that seems like a good way to tell them apart.

PR #18075, opened 2022-10-04
CXP-1307: Catalog aggregate

RiffVentura:
Added Catalog aggregate for internal uses with its own query that hydrates it

### DIFF
--- a/components/catalogs/back/src/Domain/Catalog/Catalog.php
+++ b/components/catalogs/back/src/Domain/Catalog/Catalog.php
@@ -2,7 +2,7 @@
 
 declare(strict_types=1);
 
-namespace Akeneo\Catalogs\Domain;
+namespace Akeneo\Catalogs\Domain\Catalog;
 
 /**
  * @copyright 2022 Akeneo SAS (http://www.akeneo.com)
@@ -15,6 +15,12 @@ namespace Akeneo\Catalogs\Domain;
  *      scope?: string,
  *      locale?: string,
  * }
+ *
+ * @phpstan-type ProductValueFilters array{
+ *      channels?: array<string>,
+ *      locales?: array<string>,
+ *      currencies?: array<string>,
+ * }
  */
 final class Catalog
 {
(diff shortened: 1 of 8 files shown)


guillaumefer-akeneo (approved):

masacc (approved):

components/catalogs/back/src/Domain/Catalog/Catalog.php
```diff
@@ -0,0 +1,76 @@
...
+ * }
+ */
+final class Catalog
+{
+    /**
+     * @param array<array-key, ProductSelectionCriterion> $productSelectionCriteria
+     * @param ProductValueFilters $productValueFilters
+     */
+    public function __construct(
+        private string $id,
```
willy-ahva, line 32: read-only properties in PHP8.1 will be fantastic for this kind of use case.

components/catalogs/back/src/Infrastructure/Persistence/Catalog/GetCatalogQuery.php
```diff
@@ -0,0 +1,79 @@
...
+ * @phpstan-import-type ProductSelectionCriterion from Catalog
+ * @phpstan-import-type ProductValueFilters from Catalog
+ */
+final class GetCatalogQuery implements GetCatalogQueryInterface
+{
+    public function __construct(private Connection $connection)
+    {
+    }
+
+    public function execute(string $id): Catalog
```
willy-ahva, line 34:
```suggestion
    public function execute(string $catalogId): Catalog
```
As it's not detailed in the query name `GetCatalogById` I would have named the argument $catalogId, because there is an owner id too in the aggregate.

components/catalogs/back/src/Infrastructure/Persistence/Catalog/GetCatalogQuery.php
```diff
@@ -0,0 +1,79 @@
...
+        WHERE catalog.id = :id
+        SQL;
+
+        /** @var DatabaseCatalog|false $row */
+        $row = $this->connection->executeQuery($query, [
+            'id' => Uuid::fromString($id)->getBytes(),
+        ])->fetchAssociative();
+
+        if (!$row) {
+            throw new \LogicException('Catalog not found.');
```
willy-ahva, line 55: WDYT about having a real Domain exception? We can do it later.

willy-ahva (approved):

components/catalogs/back/src/Infrastructure/Persistence/Catalog/GetCatalogQuery.php
tseho: I think it's better to have a domain exception but I also agree that we should do it later.

tseho (approved):

PR #18083, opened 2022-10-05
CXP-1313: Move Catalog DTO & QueryInterface

tseho:

Having query interfaces in Domain is a reasonable option.

But I think this is not the way to go for us.
- We already have all persistence interfaces in Application/Persistence. => As usual, if we put similar patterns in different directories without an obvious explanation on the case-by-case situation, it doesn't scale well. How do you guess which query belong to Domain and the other to Application ? How do you trace the line ?
- We cannot autowire interfaces that are in Domain. The Domain directory is excluded on purpose https://github.com/akeneo/pim-community-dev/blob/master/components/catalogs/back/src/Infrastructure/Symfony/Resources/config/services.yml#L17 => Why ? If you don't, Symfony tries to convert all classes in it into services, including DTOs. It will complains about unknown constructor arguments. That's why Symfony always recommend to exclude directories containing DTOs from the autowiring feature. 

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/components/catalogs/back/src/Application/Persistence/Catalog/GetCatalogQueryInterface.php
+++ b/components/catalogs/back/src/Application/Persistence/Catalog/GetCatalogQueryInterface.php
@@ -2,7 +2,9 @@
 
 declare(strict_types=1);
 
-namespace Akeneo\Catalogs\Domain\Catalog;
+namespace Akeneo\Catalogs\Application\Persistence\Catalog;
+
+use Akeneo\Catalogs\Domain\Catalog;
 
 /**
  * @copyright 2022 Akeneo SAS (http://www.akeneo.com)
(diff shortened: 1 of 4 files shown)


RiffVentura (approved):

components/catalogs/back/src/Application/Persistence/Catalog/GetCatalogQueryInterface.php
```diff
@@ -2,7 +2,9 @@
 
 declare(strict_types=1);
 
-namespace Akeneo\Catalogs\Domain\Catalog;
+namespace Akeneo\Catalogs\Application\Persistence\Catalog;
+
+use Akeneo\Catalogs\Domain\Catalog;
```
eoliveauakeneo, line 7:
```suggestion
use Akeneo\Catalogs\Domain\Catalogs;
```

willy-ahva (approved):

eoliveauakeneo (approved):

PR #18163, opened 2022-10-19
GRF-371: Display category templates column and hide it where there is no template

BryanS29-akeneo:

Clean legacy codein `CategoryTreeController` .

Add `Category templates column` in `CategoryTreeDataGrid` in order to display the template label if the category tree had one. Hide the column if no one of the category trees have a template associated.

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc


jeremyBeaucousinAkeneo (approved):

src/Akeneo/Category/back/Domain/Model/CategoryTree.php
```diff
@@ -60,6 +60,17 @@ public function getCode(): Code
         return $this->code;
     }
 
+    public function getLabel(string $localeCode): string
+    {
+        $label = $this->labels->getTranslation($localeCode);
+
+        if (!$label) {
```
antoinetrouve, line 67: If i understand here, if there is not translation your are returning the category Tree code, otherwise your return a label. In this function your return 2 different information (even if both are string type). This is weird no ? This is a specific case ?

antoinetrouve (approved):

src/Akeneo/Category/front/src/feature/components/datagrids/CategoryTreeDataGrid.tsx
```diff
@@ -132,6 +152,14 @@ const CategoryTreesDataGrid: FC<Props> = ({trees, refreshCategoryTrees}) => {
               <Table.HeaderCell>
                 {translate('pim_enrich.entity.category.content.tree_list.columns.number_of_categories')}
               </Table.HeaderCell>
+              {featureFlags.isEnabled('enriched_category') &&
+                (isGranted('pim_enrich_product_category_template') ||
+                  isGranted('pim_enrich_product_category_edit_attributes')) &&
```
olivierpontier, line 157: Maybe we can factorize this condition by using a variable like `canManageTemplate` or something like that, it would ease the readability

src/Akeneo/Category/back/Domain/Model/CategoryTree.php
BryanS29-akeneo: In the PIM when we display an information that doesn't have a label, we display its code instead like this `[the_code]`

olivierpontier (approved):

PR #18167, opened 2022-10-19
RAB-1137: Upgrade openspout/openspout to v4

ghoussard:

- Upgrade openspout/openspout from v3 to v4
- Make some changes related to API changes : 
  - add factories to handle Reader/Writer with Options instantiation 
  - replace EntityFactory calls by static constructor on entites

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc


src/Akeneo/Tool/Component/Connector/Reader/File/SpoutReaderFactory.php
```diff
@@ -0,0 +1,47 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Akeneo\Tool\Component\Connector\Reader\File;
```
ghoussard, line 5: I know infrastructure code should be under Bundle instead of Component but if I do it, I break the rule that Component code should not call Bundle code 😕 
Component depended on Spout before this PR, so I let it like that

StevenVAIDIE (approved):

pierrejolly (approved):

PR #18184, opened 2022-10-21
RAB-1109: Add phpstan-deprecation

StevenVAIDIE:

In this PR: 
- I fixed the deprecation that rector does not see (getUsername and MASTER_REQUEST)
- I added a check into the CI to validate that we didn't use anymore deprecated function/class (there is some todo because there is dedicated ticket)
- I fixed some error that phpstan seen (for example function that should return something and return nothing or sprintf with wrong argument number)

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc


src/Akeneo/Connectivity/Connection/back/tests/EndToEnd/Apps/Internal/RefreshConnectedAppActionEndToEnd.php
```diff
@@ -7,6 +7,7 @@
 use Akeneo\Connectivity\Connection\back\tests\EndToEnd\WebTestCase;
 use Akeneo\Connectivity\Connection\Infrastructure\Marketplace\WebMarketplaceApi;
 use Akeneo\Connectivity\Connection\Tests\CatalogBuilder\ConnectedAppLoader;
+use Akeneo\Connectivity\Connection\Tests\CatalogBuilder\ConnectionLoader;
```
willy-ahva, line 10: I don't think it's needed

src/Akeneo/Tool/Component/StorageUtils/Cache/EntityManagerClearerInterface.php
```diff
@@ -14,7 +14,7 @@ interface EntityManagerClearerInterface
     /**
      * Clears the internal cache
      *
-     * @return EntityManagerClearerInterface
+     * @return void
      */
     public function clear();
```
pierrejolly, line 19:
```suggestion
    public function clear(): void;
```

pierrejolly (approved):

MadeWilson (approved):

PR #18196, opened 2022-10-21
PIM-10683: Fix performance regression in CSV and XLSX product and pro…

ahocquard:
…duct model export

A cache query is not used inside the repository. It means that for each value, it triggers a SQL query. It slows down the CSV  and XLSX export probably by a factor between x2 and x5 depending of the catalog (tested on real catalog, I did a x3).

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/src/Akeneo/Pim/Enrichment/Component/Product/Connector/Processor/Normalization/ProductProcessor.php
+++ b/src/Akeneo/Pim/Enrichment/Component/Product/Connector/Processor/Normalization/ProductProcessor.php
@@ -32,7 +32,7 @@ class ProductProcessor implements ItemProcessorInterface, StepExecutionAwareInte
         protected IdentifiableObjectRepositoryInterface $channelRepository,
         protected AttributeRepositoryInterface $attributeRepository,
         protected FillMissingValuesInterface $fillMissingProductModelValues,
-        // TODO: remove nullability after merge into master
+        // TODO: pull up master => remove nullability
         private ?GetAttributes $getAttributes = null,
         private ?GetProductsWithQualityScoresInterface $getProductsWithQualityScores = null,
     ) {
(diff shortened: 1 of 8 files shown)


src/Akeneo/Pim/Enrichment/Component/Product/Connector/Processor/Normalization/ProductProcessor.php
```diff
@@ -55,7 +46,7 @@ public function process($product): array
         $parameters = $this->stepExecution->getJobParameters();
         $structure = $parameters->get('filters')['structure'];
         $channel = $this->channelRepository->findOneByIdentifier($structure['scope']);
-
+        $jobLocales = $this->stepExecution->getJobParameters()->get('filters')['structure']['locales'];
```
StevenVAIDIE, line 49: Can be moved to filterLocaleSpecificAttributeCodes directly

tests/back/Pim/Enrichment/Specification/Component/Product/Connector/Processor/Normalization/ProductProcessorSpec.php
```diff
@@ -268,4 +275,19 @@ public function it_processes_product_with_filter_on_quality_score(
 
         $this->process($product)->shouldBeLike($normalizedProductWithQualityScores);
     }
+
+    private function createAttribute(string $code): Attribute {
+        return new Attribute(
+            '$code',
```
StevenVAIDIE, line 281:
```suggestion
            $code,
```

StevenVAIDIE (approved):
Should we add a Spec ? For me it's should not be tested with phpspec but in case of you forget to add it

ahocquard:
@StevenVAIDIE it's an implementation change already covered by some tests (integration test)

ValentinMumble (approved):

PR #18204, opened 2022-10-24
GRF-492: Attribute tab should be in first place

BryanS29-akeneo:

"Attribute" tab in first place instead of "Property"
Refacto on switching tabs.

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/src/Akeneo/Category/front/src/feature/pages/CategoryEditPage.tsx
+++ b/src/Akeneo/Category/front/src/feature/pages/CategoryEditPage.tsx
@@ -36,10 +36,12 @@ type Params = {
   categoryId: string;
 };
 
-const propertyTabName = '#pim_enrich-category-tab-property';
-const attributeTabName = '#pim_enrich-category-tab-attribute';
-const historyTabName = '#pim_enrich-category-tab-history';
-const permissionTabName = '#pim_enrich-category-tab-permission';
+enum Tabs {
+  ATTRIBUTE = '#pim_enrich-category-tab-attribute',
+  PROPERTY = '#pim_enrich-category-tab-property',
+  HISTORY = '#pim_enrich-category-tab-history',
+  PERMISSION = '#pim_enrich-category-tab-permission',
+}
 
 const CategoryEditPage: FC = () => {
   const {categoryId} = useParams<Params>();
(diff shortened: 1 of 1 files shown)


src/Akeneo/Category/front/src/feature/pages/CategoryEditPage.tsx
```diff
@@ -137,11 +150,10 @@ const CategoryEditPage: FC = () => {
...
 
   useEffect(() => {
-    if (activeTab === attributeTabName && !isGranted('pim_enrich_product_category_edit_attributes')) {
-      setActiveTab(propertyTabName);
-      switchTo(propertyTabName);
+    if (activeTab === Tabs.ATTRIBUTE && !isGranted('pim_enrich_product_category_edit_attributes')) {
+      handleSwitchTo(Tabs.PROPERTY);
     }
-  }, [activeTab]);
+  }, [activeTab, handleSwitchTo, isGranted]);
```
olivierpontier, line 156: Could you double-check the need to define `handleSwitchTo` and `isGranted` as key of this useEffect, hook please?
In my opinion, it is not necessary here, the `activeTab` state should be enough. Furthermore, it could trigger unexpected rerender.

src/Akeneo/Category/front/src/feature/pages/CategoryEditPage.tsx
```diff
@@ -62,11 +64,22 @@ const CategoryEditPage: FC = () => {
...
   const [isCurrent, switchTo] = useTabBar(activeTab);
   const [secondaryActionIsOpen, openSecondaryAction, closeSecondaryAction] = useBooleanState(false);
   const [isDeleteCategoryModalOpen, openDeleteCategoryModal, closeDeleteCategoryModal] = useBooleanState();
 
+  const handleSwitchTo = useCallback(
+    (tab: string) => {
+      setActiveTab(tab);
+      switchTo(tab);
+    },
+    [setActiveTab, switchTo]
```
olivierpontier, line 80: The `setActiveTab` and `switchTo` functions don't seem to be stateful. IMHO, they are initialized at the first execution and we do not need to define them in the key of this `useCallback`. It should avoid unexpected rerender.

```suggestion
    []
```

olivierpontier (approved):

src/Akeneo/Category/front/src/feature/pages/CategoryEditPage.tsx
BryanS29-akeneo: If I don't define them here, I have a message from phpstorm: `ESLint: React Hook useEffect has missing dependencies: 'handleSwitchTo' and 'isGranted'. Either include them or remove the dependency array.(react-hooks/exhaustive-deps)`. It's not critical but that's why I added them. Should I ignore it?

src/Akeneo/Category/front/src/feature/pages/CategoryEditPage.tsx
BryanS29-akeneo: Same as above

antoinetrouve (approved):

jeremyBeaucousinAkeneo (approved):

src/Akeneo/Category/front/src/feature/pages/CategoryEditPage.tsx
olivierpontier: We should ignore it. This warning message from ESLint is interesting but not pertinent to every case.
In our situation, the risk is it could produce unexpected rerender

PR #18206, opened 2022-10-24
PH-278: Refactor CRON job akeneo:connectivity-connection:openid-keys:create

niarcotyk:
Add a job to create or update a public/private openId key couple


jmleroux (approved):

niarcotyk:
Problem

PR #18228, opened 2022-10-25
PH-269: Extract `akeneo:connectivity-audit:update-data` logic

jmleroux:
Extract `akeneo:connectivity-audit:update-data` logic into a dedicated service.
This will allow to reuse this logic easilly, like in scheduled jobs for UCS.

I also add a few logs for observability.

### DIFF
--- a/src/Akeneo/Connectivity/Connection/back/Infrastructure/Audit/UpdateAuditData.php
+++ b/src/Akeneo/Connectivity/Connection/back/Infrastructure/Audit/UpdateAuditData.php
@@ -1,0 +1,86 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Akeneo\Connectivity\Connection\Infrastructure\Audit;
+
+use Akeneo\Connectivity\Connection\Application\Audit\Command\UpdateDataSourceProductEventCountCommand;
+use Akeneo\Connectivity\Connection\Application\Audit\Command\UpdateDataSourceProductEventCountHandler;
+use Akeneo\Connectivity\Connection\Domain\Audit\Persistence\PurgeAuditProductQueryInterface;
+use Akeneo\Connectivity\Connection\Domain\ValueObject\HourlyInterval;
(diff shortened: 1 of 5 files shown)


RiffVentura (approved):

src/Akeneo/Connectivity/Connection/back/Infrastructure/Audit/UpdateAuditData.php
```diff
@@ -0,0 +1,86 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Akeneo\Connectivity\Connection\Infrastructure\Audit;
```
masacc, line 5: just an open question as I don't know exactly how it will be used in UCS: isn't the best time to extract it as a kind of ServiceApi if it's done to be used elsewhere? ping @tseho

jmleroux: yes, ping @tseho :smile:

tseho: No ServiceAPI.

If I remember the context correctly, we are moving work from symfony commands + cronjobs to scheduled jobs.
I got a preview of the other PRs started by Anne-Laure.

It means that the code currently located in symfony commands must be moved to a generic class, so we can call it from a `Step`.

We don't want those newly extracted classes to be called from anybody, it's not a ServiceAPI.

tseho (approved):

src/Akeneo/Connectivity/Connection/back/Infrastructure/Audit/UpdateAuditData.php
jmleroux: > It means that the code currently located in symfony commands must be moved to a generic class, so we can call it from a Step.

:+1:

PR #18241, opened 2022-10-26
CPM-767: Manage associations with uuids

pierallard:
Originally, this PR only fixes
- The associate screen from the PEF
- The assocaite screen from the mass actions

BUT
To do this, we need to change the way the data are sent to the front and the data are sent to the back, to only use uuids. So ; the Standard Format has changed, to now have "product_uuids" key in the associations instead of "products", which contained identifiers.
We changed everywhere the associations are used, but we needed to keep the old format in 2 particular cases
- The import/export : we need to have identifiers
- The old API endpoint : same same.

To be able to still manage identifiers, we use a $context value in the AssociationNormalizer, to force the usage of "products" identifiers instead of "product_uuids".


src/Akeneo/Pim/Enrichment/Bundle/Resources/config/validation/product.yml
```diff
@@ -20,8 +20,8 @@ Akeneo\Pim\Enrichment\Component\Product\Model\Product:
         uuid:
             - Akeneo\Pim\Enrichment\Component\Product\Validator\Constraints\IsUuid4: ~
         identifier:
-            - Akeneo\Pim\Enrichment\Component\Product\Validator\Constraints\NotBlank:
-                attributeCode: 'identifier'
+#            - Akeneo\Pim\Enrichment\Component\Product\Validator\Constraints\NotBlank:
+#                attributeCode: 'identifier'
```
mmetayer, line 24: to revert

src/Akeneo/Pim/Enrichment/Bundle/Storage/Sql/Product/GetAssociatedProductUuidsByProductFromDB.php
```diff
@@ -16,15 +16,15 @@ public function __construct(private Connection $connection)
...
     {
         $sql = <<<SQL
-SELECT DISTINCT(p.identifier) as code
+SELECT DISTINCT(BIN_TO_UUID(p.uuid)) as uuid
 FROM pim_catalog_association a
     INNER JOIN pim_catalog_association_product ap ON a.id = ap.association_id
     INNER JOIN pim_catalog_product p ON p.uuid = ap.product_uuid
 WHERE a.owner_uuid = UUID_TO_BIN(:ownerUuid) AND a.association_type_id = :associationTypeId
-ORDER BY p.identifier ASC;
+ORDER BY uuid ASC;
```
mmetayer, line 27: I think we can remove this ORDER BY clause (it might slow down the query and we sort in the normalizer anyway)

cemorin (approved):

Beninho (approved):

Balamung (approved):

PR #18245, opened 2022-10-26
PIM-10674: Add possibility to send a loading method on Card component

StevenVAIDIE:

In this PR I modified the Card component in order to :
- Have the possibility to give a ref, for example to detect if the component is in the user screen
- Add a props called `loading` that indicates how the browser should load the card image

In order to ease the review you can `Hide whitespace`
Screenshot from 2022-10-26 16-35-07

- [x] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/front-packages/akeneo-design-system/src/components/Card/Card.tsx
+++ b/front-packages/akeneo-design-system/src/components/Card/Card.tsx
@@ -1,4 +1,4 @@
-import React, {isValidElement, ReactElement, ReactNode, MouseEvent} from 'react';
+import React, {isValidElement, ReactElement, ReactNode, MouseEvent, forwardRef, Ref} from 'react';
 import styled, {css} from 'styled-components';
 import {Checkbox, Link, LinkProps, Image} from '../../components';
 import {AkeneoThemedProps, getColor, getFontSize} from '../../theme';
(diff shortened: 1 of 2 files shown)


front-packages/akeneo-design-system/src/components/Card/Card.tsx
```diff
@@ -189,80 +194,89 @@ type CardProps = Override<
...
-
-  React.Children.forEach(children, child => {
-    if (typeof child === 'string') {
-      texts.push(child);
-    } else {
-      if (isValidElement<LinkProps>(child) && Link === child.type) {
-        linkProps = {...child.props, href: disabled ? undefined : child.props.href};
-      } else if (isValidElement<BadgeContainerProps>(child) && BadgeContainer === child.type) {
-        nonLabelChildren.push(React.cloneElement(child, {key: child.key, stacked}));
+const CardComponent = forwardRef<HTMLButtonElement, CardProps>(
```
ghoussard, line 197: I'm not against the idea to use a `Component` suffix but this is not done on other dsm components 🤔

ghoussard (approved):

front-packages/akeneo-design-system/src/components/Card/Card.tsx
StevenVAIDIE: The exported component is always Card not CardComponent. I do it because i need to have a forward ref into a compound component (not already done)

PR #18263, opened 2022-10-27
OCT-164: Add 'id' to connection_error ES index

RiffVentura:
Currently, our logs are polluted with deprecation messages such as:
```
Loading the fielddata on the _id field is deprecated and will be removed in future versions. If you require sorting or aggregating on this field you should also include the id in the body of your documents, and map this field as a keyword field that has [doc_values] enabled
```
In order to mitigate it, another mapped field is required to be used instead of `_id`
This PR only adds the new field to the list of mapped fields
[Another PR](https://github.com/akeneo/pim-community-dev/pull/18266) will update queries that use `_id` on the `connection_error` index

### DIFF
--- a/src/Akeneo/Connectivity/Connection/back/tests/.php_cd.php
+++ b/src/Akeneo/Connectivity/Connection/back/tests/.php_cd.php
@@ -33,6 +33,7 @@ $rules = [
             'Akeneo\Pim\Enrichment\Component\Product\Message\ProductUpdated',
 
             'Webmozart\Assert\Assert',
+            'Ramsey\Uuid\Uuid',
         ]
     )->in('Akeneo\Connectivity\Connection\Domain'),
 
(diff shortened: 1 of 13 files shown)


eoliveauakeneo (approved):

src/Akeneo/Connectivity/Connection/back/tests/.php_cd.php
```diff
@@ -33,6 +33,7 @@
             'Akeneo\Pim\Enrichment\Component\Product\Message\ProductUpdated',
 
             'Webmozart\Assert\Assert',
+            'Ramsey\Uuid\Uuid',
```
guillaumefer-akeneo, line 36: :thinking: I wonder if we can add this class in our domain, shouldn't we make an ACL ?

guillaumefer-akeneo (approved):

willy-ahva (approved):
Good job

src/Akeneo/Connectivity/Connection/back/tests/.php_cd.php
willy-ahva: You're right @guillaumefer-akeneo , it's part of one of our exceptions you use directly in the code without ACL. We should but we don't.

PR #18294, opened 2022-10-31
OCT-157 GET test-apps endpoint

RiffVentura:
Added endpoint to fetch tests-apps for internal use

I had to add a blank endpoint: **GET** `/api/rest/v1/test-apps/{clientId}` in order to use HAL paginator as it requires a route for a single item in the list


tseho:
We don't need the HAL paginator, it's a private endpoint.
Returning the simple list of tests apps as documented in the ticket was enough.
Can you remove the unnecessary endpoint ?

tseho (approved):

PR #18324, opened 2022-11-03
CPM-783: Add front for edit

pierallard:


components/identifier-generator/front/src/feature/controllers/Edit.tsx
```diff
@@ -1,23 +1,44 @@
...
-  // Temporary
-  const generator = {
-    code: identifierGeneratorCode,
-    target: 'sku',
-    labels: {en_US: 'a label'},
-    conditions: [],
-    structure: [{type: PROPERTY_NAMES.FREE_TEXT, string: 'AKN'}],
-    delimiter: null,
-  };
+  if (error) {
```
cemorin, line 15: Inside a useMemo would be better than rerendering a lot in here

cemorin (approved):

Beninho (approved):

PR #18359, opened 2022-11-07
CPM-801: Add mutations

pierallard:
- Add mutations
- Put queries in hooks
- use correct fetch mocks

### DIFF
--- a/components/identifier-generator/front/src/feature/hooks/useCreateIdentifierGenerator.ts
+++ b/components/identifier-generator/front/src/feature/hooks/useCreateIdentifierGenerator.ts
@@ -1,0 +1,31 @@
+import {IdentifierGenerator} from '../models';
+import {InvalidIdentifierGenerator, ServerError} from '../errors';
+import {useMutation, useQueryClient} from 'react-query';
+import {useRouter} from '@akeneo-pim-community/shared';
+
+// eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
+const useCreateIdentifierGenerator = () => {
+  const router = useRouter();
+  const queryClient = useQueryClient();
+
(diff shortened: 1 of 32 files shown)


components/identifier-generator/front/src/feature/hooks/useCreateIdentifierGenerator.ts
```diff
@@ -0,0 +1,34 @@
...
+        const data = await response.json();
+        throw new InvalidIdentifierGenerator(data);
+      }
+
+      if (response.status !== 201) {
+        throw new ServerError();
+      }
+    },
+    {
+      onSuccess: () => queryClient.invalidateQueries('getGeneratorList'),
```
cemorin, line 29: Are both of the `onSuccess` triggered? (the one from the mutate and this one)?

cemorin: And you can handle it all here directly. I feel like it avoids having two places handling the response from this call (just a though/preference, not a real recommandation)

cemorin:
:fire: :fire: :fire:

components/identifier-generator/front/src/feature/hooks/useCreateIdentifierGenerator.ts
pierallard: This is on purpose.
The goal of the useCreateIdentifierGenerator is to Create generator and assure that the queries are invalidated.
The goal of the onSuccess in the component is just to display notifications or redirect for end user.
Yes, the 2 onSuccess are called (you can check the doc)

Balamung (approved):

Beninho (approved):

PR #18362, opened 2022-11-08
CPM-797 : compare uuid instead of identifier

Beninho:
Fix 2-way association for product without sku


Balamung (approved):

src/Akeneo/Pim/Enrichment/Bundle/Doctrine/ORM/Updater/TwoWayAssociationUpdater.php
```diff
@@ -100,4 +103,9 @@ public function removeInversedAssociation(
         $em = $this->registry->getManager();
         $em->persist($associatedEntity);
     }
+
+    private function hasSameProductUuid(ProductInterface $product1, ProductInterface $product2): bool
```
cemorin, line 107: Maybe add a test in the specs for this use case?

cemorin (approved):

PR #18376, opened 2022-11-09
GRF-558 : Refactoring pim_enrich_categorytree_children route

jeremyBeaucousinAkeneo:

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/src/Akeneo/Pim/Enrichment/Bundle/Controller/InternalApi/CategoryTreeController.php
+++ b/src/Akeneo/Pim/Enrichment/Bundle/Controller/InternalApi/CategoryTreeController.php
@@ -1,0 +1,165 @@
+<?php
+
+namespace Akeneo\Pim\Enrichment\Bundle\Controller\InternalApi;
+
+use Akeneo\Category\Domain\Model\Classification\CategoryTree;
+use Akeneo\Category\Domain\Query\GetCategoryInterface;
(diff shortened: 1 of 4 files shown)


src/Akeneo/Category/back/Domain/Model/Enrichment/Category.php
```diff
@@ -134,4 +135,10 @@ public function normalize(): array
             'permissions' => $this->getPermissions()?->normalize(),
         ];
     }
+
+    public function getLabel(): string
```
olivierpontier, line 139: the locale should be passed as argument

src/Akeneo/Category/back/Domain/ValueObject/CategoryCollection.php
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    /**
+     * Retrieve an Category by his identifier.
+     *
+     * @param string $identifier format expected : 'code|uuid' (example : title|69e251b3-b876-48b5-9c09-92f54bfb528d)
+     */
+    public function getCategoryByIdentifier(string $identifier): ?Category
+    {
+        $category = array_filter(
```
olivierpontier, line 47: the name lets us think that we retrieve only one element which is not true. Use plural like $categories for example

PR #18387, opened 2022-11-09
CPM-741: Upgrade doctrine/dbal

mmetayer:

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc


upgrades/schema/Version_6_0_20210101000000_fix_json_columns.php
```diff
@@ -0,0 +1,58 @@
...
+namespace Pim\Upgrade\Schema;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+
+/**
+ * @copyright 2022 Akeneo SAS (https://www.akeneo.com)
+ * @license   https://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+final class Version_6_0_20210101000000_fix_json_columns extends AbstractMigration
```
mmetayer, line 14: The naming is on purpose, we want this migration to be executed as soon as possible

upgrades/test_schema/Version_6_0_20210101000000_fix_json_columns_Integration.php
```diff
@@ -0,0 +1,99 @@
...
+
+    protected function getConfiguration()
+    {
+        return $this->catalog->useMinimalCatalog();
+    }
+
+    public function it_fixes_json_columns_provider()
+    {
+        return [
+            ['akeneo_batch_job_execution', 'raw_parameters'],
```
fchoquet, line 28: I created a data provider in case we want to test several columns, but I don't know which other columns are impacted! I can remove it if we decide to test just one case.

StevenVAIDIE: If you want here are the table impacted https://akeneo.slack.com/archives/C04A0TB9L1J/p1667932977271369

upgrades/test_schema/Version_6_0_20210101000000_fix_json_columns_Integration.php
```diff
@@ -0,0 +1,102 @@
...
+
+    /**
+     * @dataProvider it_fixes_json_columns_provider
+     */
+    public function test_it_fixes_json_columns($table, $column)
+    {
+        $this->assertOnlyCommentChanged($table, $column, function() use ($table, $column) {
+            $this->corruptJsonColumn($table, $column);
+        });
+
```
StevenVAIDIE, line 40:
```suggestion
        $comment = $this->assertOnlyCommentChanged($table, $column, function() use ($table, $column) {
            $this->corruptJsonColumn($table, $column);
        });

        $this->assertEquals('(DC2Type:json_array)', $comment);
```

StevenVAIDIE (approved):

mmetayer (commented):
I can't GTM since I opened the PR, but I would if I could :smile:

upgrades/test_schema/Version_6_0_20210101000000_fix_json_columns_Integration.php
fchoquet: I only kept the CE tables in this test, but I tested the compatibility of impacted EE tables locally. I think it would be overkill to duplicate this test in the EE because all the impacted tables have one of these 2 structures anyway and we cover both here:

`json DEFAULT NULL` 
or 
`json NOT NULL`

fchoquet:
> I can't GTM since I opened the PR, but I would if I could
@mmetayer I can GTM because I didn't open the PR, so I'm going to use your GTM to validate my change :p

fchoquet (approved):

PR #18390, opened 2022-11-10
PIM-10718: Categories with empty labels throw 500 error

BryanS29-akeneo:

Labels are not mandatory in the PIM, we can have a LabelCollection like this : 
```
[
   'en_US' => null,
   'de_DE' => null,
   'fr_FR' => null,
]
```

We can remove assert validation in the constructor of LabelCollection.

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/src/Akeneo/Category/back/Domain/ValueObject/LabelCollection.php
+++ b/src/Akeneo/Category/back/Domain/ValueObject/LabelCollection.php
@@ -2,8 +2,6 @@
 
 namespace Akeneo\Category\Domain\ValueObject;
 
-use Webmozart\Assert\Assert;
-
 /**
  * @copyright 2022 Akeneo SAS (https://www.akeneo.com)
  * @license   https://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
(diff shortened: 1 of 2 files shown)


src/Akeneo/Category/back/Domain/ValueObject/LabelCollection.php
```diff
@@ -18,8 +16,6 @@ final class LabelCollection implements \IteratorAggregate
      */
     private function __construct(private ?array $translatedLabels)
     {
-        Assert::allString($translatedLabels);
```
olivierpontier, line 21: Could we use the nullOrString assertion here? https://github.com/webmozarts/assert#nullable-assertions

olivierpontier (approved):

src/Akeneo/Category/back/Domain/ValueObject/LabelCollection.php
BryanS29-akeneo: No because it receives an array, all I can do is put those asserts:
```
Assert::nullOrIsArray($translatedLabels);
Assert::allStringNotEmpty(array_keys($translatedLabels));
```

PR #18394, opened 2022-11-10
CPM-803: Generate identifier on single save

pierallard:


components/identifier-generator/back/src/Domain/Model/Property/AutoNumber.php
```diff
@@ -68,4 +68,10 @@ public function digitsMin(): int
     {
         return $this->digitsMin;
     }
+
+    public function generate(): string
```
mmetayer, line 72: I wouldn't put tese methods in the domain object (which is just a value object), but rather use an application handler

components/identifier-generator/back/src/Infrastructure/Symfony/Resources/config/event_subscribers.yml
```diff
@@ -4,3 +4,10 @@ services:
       - '@database_connection'
     tags:
       - { name: kernel.event_subscriber }
+
+  Akeneo\Pim\Automation\IdentifierGenerator\Infrastructure\Subscriber\SetIdentifiersSubscriber:
+    arguments:
+      - '@Akeneo\Pim\Automation\IdentifierGenerator\Domain\Repository\IdentifierGeneratorRepository'
+      - '@validator'
```
mmetayer, line 11:
```suggestion
      - '@pim_catalog.validator.product'
```

components/identifier-generator/back/src/Infrastructure/Subscriber/SetIdentifiersSubscriber.php
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    private function setGeneratedIdentifier(
+        IdentifierGenerator $identifierGenerator,
+        ProductInterface $product
+    ): void {
+        $newIdentifier = $identifierGenerator->generate();
+        $value = ScalarValue::value($identifierGenerator->target()->asString(), $newIdentifier);
+        $product->addValue($value);
+        // TODO This seems not working as I don't have an issue with duplicate identifiers.
```
mmetayer, line 76: see previous comment

components/identifier-generator/back/src/Infrastructure/Subscriber/SetIdentifiersSubscriber.php
```diff
@@ -0,0 +1,96 @@
...
+                }
+            }
+        }
+    }
+
+    private function setGeneratedIdentifier(
+        IdentifierGenerator $identifierGenerator,
+        ProductInterface $product
+    ): void {
+        $newIdentifier = $identifierGenerator->generate();
```
mmetayer, line 73: IMO this should be returned by a query/handler, the generator is not aware of other parameters (e.g auto_number, mappings, etc...)

components/identifier-generator/back/src/Infrastructure/Subscriber/SetIdentifiersSubscriber.php
```diff
@@ -0,0 +1,96 @@
...
+        $value = ScalarValue::value($identifierGenerator->target()->asString(), $newIdentifier);
+        $product->addValue($value);
```
mmetayer, line 75: also: `$product->setIdentifier($newIdentifier)` for now ( that'll probably change with multiple identifiers)

components/identifier-generator/back/src/Domain/Model/IdentifierGenerator.php
```diff
@@ -101,4 +101,11 @@ public function normalize(): array
             'delimiter' => $this->delimiter?->asString(),
         ];
     }
+
+    public function match(ProductProjection $productProjection): bool
```
antoinearrive, line 105: just for understand, why this method is not on components/identifier-generator/back/src/Domain/Model/ProductProjection.php with name nullOrEmpty for example

components/identifier-generator/back/src/Infrastructure/Subscriber/SetIdentifiersSubscriber.php
```diff
@@ -0,0 +1,116 @@
...
+            return;
+        }
+
+        $this->generateIdentifier($object);
+    }
+
+    public function setIdentifiers(GenericEvent $event): void
+    {
+        if (!\is_array($event->getSubject())) return;
+        foreach ($event->getSubject() as $subject) {
```
antoinearrive, line 61: possible use of Assert::allIsInstanceOf ?

antoinearrive (approved):

components/identifier-generator/back/src/Domain/Model/IdentifierGenerator.php
pierallard: Because for now, we only check that the identifier is empty. In the future, we will check a lot of other things which are IdentifierGenerator related, not the subscriber.

components/identifier-generator/back/src/Infrastructure/Subscriber/SetIdentifiersSubscriber.php
pierallard: Because Assert is here to "assert" (i.e. throw if this is not the case). Here, a Symfony Subscriber listen for a lot of events, and this event can have a subject which is something else than a Product (an Attribute for example), and for these case we will simply do nothing.

cemorin (approved):

components/identifier-generator/back/src/Domain/Model/IdentifierGenerator.php
Beninho: I don't find a best way to call this function "match", so GTM !

Beninho (approved):

PR #18410, opened 2022-11-14
PIM-10684 : fix an issue when Elasticsearch index is empty

niarcotyk:
fix an issue when Elasticsearch index is empty

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/src/Akeneo/Pim/Enrichment/Bundle/Command/IndexProductModelCommand.php
+++ b/src/Akeneo/Pim/Enrichment/Bundle/Command/IndexProductModelCommand.php
@@ -157,6 +157,7 @@ class IndexProductModelCommand extends Command
         $indexedCount = 0;
 
         $progressBar->start();
+
         foreach ($chunkedCodes as $codes) {
             $treatedBachSize = $this->batchEsStateHandler->bulkExecute($codes, $codesEsHandler);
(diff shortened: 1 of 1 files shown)


src/Akeneo/Pim/Enrichment/Bundle/Command/IndexProductModelCommand.php
```diff
@@ -269,6 +270,11 @@ private function getDiffProductModelCodes(int $batchSize): iterable
                 'size' => $batchSize
             ]);
 
+            if (count($results['hits']['hits']) === 0) {
```
docteurklein, line 273: isn't there a `$results['hits']['count']` or something?
if it's zero, we should in theory just `break`, no?
why `yield array_column($rows, 'code')` ?

niarcotyk: I thought that if it was 0 then it would send the same data as when you run the --all command. The break is that when testing locally on a catalog of 50 product models, the table returned had 2 dimensions (first with MySQL codes and the second with product model codes already indexed and it returned the same error as before the fix)

PR #18454, opened 2022-11-18
CPM-809: Only validate identifier values

pierallard:


antoinearrive (approved):

cemorin (approved):

components/identifier-generator/back/src/Infrastructure/Subscriber/SetIdentifiersSubscriber.php
```diff
@@ -127,4 +152,41 @@ private function getIdentifierGenerators(): array
...
+    /**
+     * Returns Symfony constraints defined here:
+     * src/Akeneo/Pim/Enrichment/Bundle/Resources/config/validation/product.yml
+     *
+     * @return Constraint[]
+     */
+    private function getProductConstraints(ProductInterface $product): array
+    {
+        $metadata = $this->metadataFactory->getMetadataFor($product);
+        Assert::isInstanceOf($metadata, ClassMetadataInterface::class);
```
mmetayer, line 165: that might lead to uncaught exception, this preventing to save of the product. Maybe add a general try/catch somewhere in the `setGeneratorIdentifier` method and cast it to `UnableToSetIdentifierException` ?

components/identifier-generator/back/tests/Specification/Infrastructure/Subscriber/SetIdentifiersSubscriberSpec.php
```diff
@@ -59,33 +68,99 @@ public function it_should_listen_to_pre_save_events(): void
...
+        PropertyMetadataInterface $productPropertyMetadata,
     ): void {
         $product->getValue('sku')->shouldBeCalled()->willReturn(null);
         $identifierGeneratorRepository->getAll()->shouldBeCalled()->willReturn([$this->getIdentifierGenerator()]);
         $value = ScalarValue::value('sku', 'AKN');
         $product->addValue($value)->shouldBeCalled();
         $product->setIdentifier('AKN')->shouldBeCalled();
-        $validator->validate($product)->shouldBeCalled()->willReturn(new ConstraintViolationList());
+
+
```
mmetayer, line 85: :fire:

mmetayer (approved):
GTM after handling comment about exception catching. We'll also have to add integration tests later

components/identifier-generator/back/src/Infrastructure/Subscriber/SetIdentifiersSubscriber.php
pierallard: Don't agree with you ; this is completely internal to Symfony. A MetadataFactory from a class HAS TO return a ClassMetadataInterface. If it's not the case, I prefer to fail with a 500 and a datadog error and failing tests instead of failing silently and not know the root cause.

PR #18460, opened 2022-11-18
[wait until 15/12/22] CPM-817: Add subscriber to fill prefixes

pierallard:


components/identifier-generator/back/src/Infrastructure/Query/SqlUpdateIdentifierPrefixesQuery.php
```diff
@@ -0,0 +1,115 @@
...
+                    ];
+                }
+            }
+        }
+
+        if (\count($newPrefixes) === 0) {
+            return;
+        }
+
+        $placeholders = implode(',', array_fill(0, \count($newPrefixes), '(UUID_TO_BIN(?),?,?,?)'));
```
mmetayer, line 69: probably more efficient if you let the uuid as is (`(?, ?, ?, ?)`) and use $product->getUuid()->getBytes() in the $newPrefixes

components/identifier-generator/back/src/Infrastructure/Query/SqlUpdateIdentifierPrefixesQuery.php
```diff
@@ -0,0 +1,115 @@
...
+
+        $valuesStr = \implode(
+            ',',
+            array_map(
+                fn (array $value): string => \sprintf('(UUID_TO_BIN("%s"), %d, "%s", %d)', ...$value),
+                $newPrefixes
+            )
+        );
```
mmetayer, line 77: to remove

mmetayer (approved):

components/identifier-generator/back/src/Infrastructure/Query/SqlUpdateIdentifierPrefixesQuery.php
pierallard: We can't put directly bytes here :cry:

components/identifier-generator/back/tests/Specification/Domain/Model/ProductIdentifierSpec.php
```diff
@@ -0,0 +1,50 @@
...
+ */
+class ProductIdentifierSpec extends ObjectBehavior
+{
+    function it_is_a_product_identifier()
+    {
+        $this->beConstructedWith('AKN-123');
+        $this->shouldHaveType(ProductIdentifier::class);
+    }
+
+    function it_generates_psimple_refixes()
```
cemorin, line 22:
```suggestion
    function it_generates_simple_prefixes()
```

cemorin (approved):

components/identifier-generator/back/src/Infrastructure/Query/SqlUpdateIdentifierPrefixesQuery.php
mmetayer: even if you use `ParameterType::BINARY` as 3rd argument of `bindValue()` ?

components/identifier-generator/back/src/Infrastructure/Query/SqlUpdateIdentifierPrefixesQuery.php
```diff
@@ -0,0 +1,129 @@
...
+    /**
+     * @return AttributeInterface[]
+     */
+    private function getIdentifierAttributes(): array
+    {
+        if (null === $this->identifierAttributesCache) {
+            $this->identifierAttributesCache = [$this->attributeRepository->getIdentifier()];
+        }
+
+        return $this->identifierAttributesCache;
```
mmetayer, line 127: we could also simply store identifier attribute ids indexed by their code (no need to hydrate the whole attribute IMO). That being said, at this point the identifier is probably already registered in the UoW, so we can do that later in another card (or not)

components/identifier-generator/back/tests/EndToEnd/Infrastructure/Query/SqlUpdateIdentifierPrefixesQueryEndToEnd.php
```diff
@@ -0,0 +1,156 @@
...
+        if (null !== $identifier) {
+            $userIntents = [new SetIdentifierValue('sku', $identifier)];
+        }
+
+        $command = UpsertProductCommand::createWithUuid(
+            $this->admin->getId(),
+            ProductUuid::fromUuid($uuid),
+            $userIntents
+        );
+        ($this->getUpsertProductHandler())($command);
```
mmetayer, line 85: in production code we use the message bus instead of directly calling the handler, we should do that in the tests as well

components/identifier-generator/back/tests/EndToEnd/Infrastructure/Query/SqlUpdateIdentifierPrefixesQueryEndToEnd.php
```diff
@@ -0,0 +1,156 @@
...
+        $productFromDatabase = $this->createProduct('my_identifier_123');
+        Assert::assertEquals($this->getPrefixes($productFromDatabase->getUuid(), $this->getIdentifierId()), [
+            'my_identifier_' => '123',
+            'my_identifier_1' => '23',
+            'my_identifier_12' => '3',
+        ]);
+    }
+
+    /** @test */
+    public function it_should_update_prefixes_on_update(): void
```
mmetayer, line 51: we could also test the fact that prefixes are deleted if the new identifier does not contain any digit

components/identifier-generator/back/tests/EndToEnd/Infrastructure/Query/SqlUpdateIdentifierPrefixesQueryEndToEnd.php
```diff
@@ -0,0 +1,156 @@
...
+    {
+        $productFromDatabase = $this->createProduct('my_identifier_123');
+        $this->setIdentifier($productFromDatabase, 'my_new_identifier_234');
```
mmetayer, line 54: would also worth testing the result when the identifier has several number parts (e.g `my_new_identifier_1234_5678`)

mmetayer (approved):
Noice

components/identifier-generator/back/src/Infrastructure/Query/SqlUpdateIdentifierPrefixesQuery.php
pierallard: The best way should be to have a cached query to get the list of "id => code" for each attributes. This seems overkill as we currently have only 1 identifier, and should not be a big deal to have it in memory. What do you think ?

mmetayer: Yes I agree. We can let it like that for now, and re-think about it when we implement the multiple identifiers feature

PR #18490, opened 2022-11-23
Merge fpm and php containers

fchoquet:
See the corresponding EE PR for explanations: https://github.com/akeneo/pim-enterprise-dev/pull/15607

### DIFF
--- a/.circleci/jobs/build_and_deploy.yml
+++ b/.circleci/jobs/build_and_deploy.yml
@@ -180,9 +180,13 @@ jobs:
                   steps:
                       - run:
                             name: Save the CE image as a tar
+                            # TODO: Restore previous version after the first pass of the nightly build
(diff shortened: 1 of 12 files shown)


.circleci/jobs/build_and_deploy.yml
```diff
@@ -180,9 +180,13 @@ jobs:
                   steps:
                       - run:
                             name: Save the CE image as a tar
+                            # TODO: Restore previous version after the first pass of the nightly build
```
fchoquet, line 183: It seems hacky, but the CI doesn't build the image by default. It gets it from the nightly build. This is a smart optimisation that works fine 99% or times but doesn't allow modification of the docker image.

I'll merge a follow up PR to clean this hack. 

Please let me know if this is not the right approach

PR #18504, opened 2022-11-24
GRF-599 fix blank page on categories when manipulating channels list before it is fetched

amigaire:

When the PIM is build and launched for the 1st time, or when we delete the project's cache, the ChannelSelector component receives an empty list of channels. It try to access the "code" property of every channel of this empty list which resulted in an error and a blank page.

Now we do not display the component if the channels list is empty

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/src/Akeneo/Category/front/src/feature/components/EditAttributesForm.tsx
+++ b/src/Akeneo/Category/front/src/feature/components/EditAttributesForm.tsx
@@ -62,6 +62,7 @@ export const EditAttributesForm = ({attributeValues, template, onAttributeValueC
   const [locale, setLocale] = useState(catalogLocale);
   const catalogChannel = userContext.get('catalogScope');
   const [channel, setChannel] = useState(catalogChannel);
+  const channelList = useMemo(() => Object.values(channels), [channels]);
 
   const handleChannelChange = (value: string): void => {
     setChannel(value);
@@ -140,21 +141,16 @@ export const EditAttributesForm = ({attributeValues, template, onAttributeValueC
     );
   });
 
-  const channelList = Object.values(channels);
-
   return (
     <FormContainer>
       <SectionTitle>
         <SectionTitle.Title>{translate('akeneo.category.attributes')}</SectionTitle.Title>
         <SectionTitle.Spacer />
-        {
-          channelList.length > 0 &&
-          <ChannelSelector
-            value={channel}
-            values={channelList}
-            onChange={handleChannelChange}
-          />
-        }
+        <ChannelSelector
+          value={channel}
+          values={channelList}
+          onChange={handleChannelChange}
+        />
         <LocaleSelector
           value={locale}
           values={Object.values(locales)}
--- a/src/Akeneo/Category/front/src/feature/components/EditAttributesForm.tsx
+++ b/src/Akeneo/Category/front/src/feature/components/EditAttributesForm.tsx
@@ -140,16 +140,21 @@ export const EditAttributesForm = ({attributeValues, template, onAttributeValueC
     );
   });
 
+  const channelList = Object.values(channels);
+
   return (
     <FormContainer>
       <SectionTitle>
         <SectionTitle.Title>{translate('akeneo.category.attributes')}</SectionTitle.Title>
         <SectionTitle.Spacer />
-        <ChannelSelector
+        {
+          channelList.length > 0 &&
+          <ChannelSelector
             value={channel}
-            values={Object.values(channels)}
+            values={channelList}
             onChange={handleChannelChange}
-        />
+          />
+        }
         <LocaleSelector
           value={locale}
           values={Object.values(locales)}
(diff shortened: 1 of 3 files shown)


BryanS29-akeneo (approved):

src/Akeneo/Category/front/src/feature/components/EditAttributesForm.tsx
```diff
@@ -140,16 +140,21 @@ export const EditAttributesForm = ({attributeValues, template, onAttributeValueC
...
+  const channelList = Object.values(channels);
+
   return (
     <FormContainer>
       <SectionTitle>
         <SectionTitle.Title>{translate('Attributes')}</SectionTitle.Title>
         <SectionTitle.Spacer />
-        <ChannelSelector
+        {
+          channelList.length > 0 &&
```
olivierpontier, line 151: What is the purpose of testing if the list of channels is empty or not here? The ChannelSelector component should be resilient enough to handle this case, should not?

src/Akeneo/Category/front/src/feature/components/channel/ChannelSelector.tsx
```diff
@@ -38,7 +38,7 @@ const ChannelSelector = ({value, values, onChange}: ChannelSelectorProps) => {
   return (
     <Dropdown>
       <SwitcherButton inline onClick={() => {open();}} label={translate('pim_common.channel')}>
-        <ChannelSpan code={`[${selectedChannel.code}]`} label={selectedChannel.labels[catalogLocale]} />
+        <ChannelSpan code={`[${selectedChannel?.code}]`} label={selectedChannel?.labels[catalogLocale]} />
```
olivierpontier, line 41: A more resilient way to handle the case there is no selected value could be
```suggestion
        <span>{selectedChannel && (selectedChannel.label ?? `[${selectedChannel.code}]`}</span>
```

Furthermore, you can use the helper [getLabel](https://github.com/akeneo/pim-community-dev/blob/master/front-packages/shared/src/models/label-collection.ts) in the shared library to display the label

olivierpontier:
The issue looks to come from this [line](https://github.com/akeneo/pim-community-dev/pull/18504/files#diff-a81e95941d47c75eb560b638dde1cdd3dab6e3f41e893ce6259eecfac0eb3d24R34)
Here we are expecting that the list of values is never empty. This should throw our error.

A solution could be to fully accept that the value possibly does not exist in the list of values and the component should be resilient enough to handle it.

```
const selectedChannel: Channel|undefined = values.find(channel => channel.code === value);

```

Could you add some tests to cover this use case please?

src/Akeneo/Category/front/src/feature/components/EditAttributesForm.tsx
amigaire: Now ChannelSelector manage empty channels list

src/Akeneo/Category/front/src/feature/components/channel/ChannelSelector.tsx
amigaire: Thx for the tip, it has been improved

src/Akeneo/Category/front/src/feature/components/EditAttributesForm.tsx
```diff
@@ -140,15 +140,17 @@ export const EditAttributesForm = ({attributeValues, template, onAttributeValueC
     );
   });
 
+  const channelList = Object.values(channels);
```
olivierpontier, line 143: Here the channelList will have a new reference in memory at each render of the component. As you are using channelList as a props, it means ChannelSelector will be rerendered each time even if the values have not changed.
To avoid this case you can use the hook [useMemo](https://reactjs.org/docs/hooks-reference.html#usememo)

```suggestion
  const channelList = useMemo(() => Object.values(channels), [channelList]);
```

src/Akeneo/Category/front/src/feature/components/channel/ChannelSelector.tsx
```diff
@@ -1,25 +1,10 @@
...
-        </span>
-      );
-    }
-);
-
-type ChannelProps = {
-  code: string;
-  label: string;
-};
+import React from 'react';
```
olivierpontier, line 7: A common practice, we write the React import in the first place at the top of the list

src/Akeneo/Category/front/src/feature/components/channel/ChannelSelector.tsx
```diff
@@ -38,7 +23,7 @@ const ChannelSelector = ({value, values, onChange}: ChannelSelectorProps) => {
   return (
     <Dropdown>
       <SwitcherButton inline onClick={() => {open();}} label={translate('pim_common.channel')}>
-        <ChannelSpan code={`[${selectedChannel.code}]`} label={selectedChannel.labels[catalogLocale]} />
+        <span>{selectedChannel && getLabel(selectedChannel.labels, catalogLocale, `[${selectedChannel.code}]`)}</span>
```
olivierpontier, line 26: The function `getLabel` already handles wrapping the fallback between brackets

```suggestion
        <span>{selectedChannel && getLabel(selectedChannel.labels, catalogLocale, selectedChannel.code)}</span>
```

antoinetrouve (approved):

src/Akeneo/Category/front/src/feature/components/channel/ChannelSelector.tsx
amigaire: Indeed, thx

olivierpontier (approved):

PR #18514, opened 2022-11-25
PIM-10736: dynamic batch size for indexation

ahocquard:
**Context**

It is possible to have Out of memory issues due to the size of the products or product models to index. This is mainly during an import/PATCH of product models, as it can impact several thousands products.

Our initial batch size of 1000 is quite big for some heavy products. 

For example, it's possible to have 40KB of data in raw values in average per product. For a batch size of 1000 products to index, it means 40 MB of raw values to index:
- after `json_decode`, it can take up to 10x times this memory in PHP, due to the various structures (array, object, etc). 
- any manipulation on raw values to put in another variable could copy the array
- variables are purged after function execution (except for cyclic dependencies). If there are several function manipulating the raw values in the stack, the initial 400MB will be duplicated several times.

And in real life, it's the case. Hydration of raw values, creation of Value Collection, removal of non existing values, etc. 

Therefore, there is an Out of Memory and de-synchronization between ES and Mysql.

**Solution**

Do not choose an arbitrary batch size, but adapt in function of the size of the products. It will batch per chunk of products where the cumulated size of raw values does not exceed a part of the memory. Of course, there are some other content in the products (categories, associations), but it's marginal in their total size.

For example, if the maximum memory is 512MB, the maximum size of all raw values in a chunk is 8MB (see explanation above why it's much smaller than the initial 512MB).

Getting the size of the products is quite fast (30ms for 14000 products) and should not slow down the indexation in a significative way. The good news is that many users will have an increased batch size also this way.

**Note**
This could be used by completeness calculation later on also.

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/src/Akeneo/Pim/Enrichment/Bundle/Elasticsearch/Indexer/PhpMemoryLimit.php
+++ b/src/Akeneo/Pim/Enrichment/Bundle/Elasticsearch/Indexer/PhpMemoryLimit.php
@@ -12,7 +12,6 @@ class PhpMemoryLimit
         $limit = ini_get('memory_limit');
 
         return $this->asBytes($limit);
-
     }
 
     public function asBytes(string $limit): int
@@ -43,5 +42,4 @@ class PhpMemoryLimit
 
         return (int) $limit;
     }
-
 }
--- a/src/Akeneo/Pim/Enrichment/Bundle/Elasticsearch/Indexer/PhpMemoryLimit.php
+++ b/src/Akeneo/Pim/Enrichment/Bundle/Elasticsearch/Indexer/PhpMemoryLimit.php
@@ -1,0 +1,47 @@
+<?php
+
+namespace Akeneo\Pim\Enrichment\Bundle\Elasticsearch\Indexer;
+
+class PhpMemoryLimit
+{
+    const MAXIMUM_MEMORY_IN_BYTES = 1024*1024*1024*1024; // 512TB
+    const DEFAULT_MEMORY_IN_BYTES_WHEN_NOT_SET = 128*1024*1024; // 128MB
+
+    public function asBytesFromPHPConfig(): int
+    {
+        $limit = ini_get('memory_limit');
+
+        return $this->asBytes($limit);
+
+    }
+
+    public function asBytes(string $limit): int
+    {
+        if (!is_string($limit)) {
+            return self::DEFAULT_MEMORY_IN_BYTES_WHEN_NOT_SET;
+        }
+
+        $lastCharacter = substr($limit, -1);
+        if ($lastCharacter === 'K') {
+            return  (int) substr($limit, 0, -1) * 1024;
+        }
+        if ($lastCharacter === 'M') {
+            return  (int) substr($limit, 0, -1) * 1024 * 1024;
+        }
+        if ($lastCharacter === 'G') {
+            return  (int) substr($limit, 0, -1) * 1024 * 1024 * 1024;
+        }
+
+        if ((int) $limit === -1 || '' === trim($limit)) {
+            return self::MAXIMUM_MEMORY_IN_BYTES;
+        }
+
+        if ((int) $limit < -1 || (int) $limit === 0) {
+            return self::DEFAULT_MEMORY_IN_BYTES_WHEN_NOT_SET;
+        }
+
+
+        return (int) $limit;
+    }
+
+}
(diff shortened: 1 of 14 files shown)


src/Akeneo/Pim/Enrichment/Bundle/Storage/Sql/Product/ChunkProductUuids.php
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    /**
+     * Naïve implementation to chunk elements by groups that does not exceed a given size for the raw values.
+     * The main purpose is to adapt the size of the batch dynamically to not reach the PHP memory limit.
+     *
+     * It is not optimized to balance the size between the groups, as the solution would be more complex.
+     * @see https://en.wikipedia.org/wiki/Bin_packing_problem
+     */
+    public function byRawValuesSize(array $productUuids, int $maxSizeInBytesPerChunk)
```
ahocquard, line 26: return + document exected format

src/Akeneo/Pim/Enrichment/Bundle/Elasticsearch/Indexer/PhpMemoryLimit.php
```diff
@@ -0,0 +1,45 @@
+<?php
+
+namespace Akeneo\Pim\Enrichment\Bundle\Elasticsearch\Indexer;
+
+class PhpMemoryLimit
```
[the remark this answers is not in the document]
mmetayer: also, as discussed during the mob review, this class could also be turned into a non-service, with 2 static constructors, e.g:
```php
$memoryLimit = PhpMemoryLimit::fromPhpConfig()->asBytes();
$otherLimit = PhpMemoryLimit::fromValue('512M')->asBytes();
```

I'm not 100% sure it should be done though, so I'll let you make the call

mmetayer (approved):

src/Akeneo/Pim/Enrichment/Bundle/Elasticsearch/Indexer/ProductIndexer.php
```diff
@@ -48,7 +62,8 @@ public function indexFromProductUuids(array $productUuids, array $options = []):
 
         $indexRefresh = $options['index_refresh'] ?? Refresh::disable();
 
-        $chunks = array_chunk($productUuids, self::BATCH_SIZE);
+        $maxMemoryPerChunk = (int) floor($this->phpMemoryLimit->asBytesFromPHPConfig()/ self::MEMORY_RATIO);
+        $chunks =$this->chunkProductUuids->byRawValuesSize($productUuids, $maxMemoryPerChunk);
```
Beninho, line 66:
```suggestion
        $chunks = $this->chunkProductUuids->byRawValuesSize($productUuids, $maxMemoryPerChunk);
```

src/Akeneo/Pim/Enrichment/Bundle/Storage/Sql/ProductModel/ChunkProductModelCodes.php
```diff
@@ -0,0 +1,68 @@
...
+
+namespace Akeneo\Pim\Enrichment\Bundle\Storage\Sql\ProductModel;
+
+use Doctrine\DBAL\Connection;
+
+/**
+ * @copyright 2022 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+class ChunkProductModelCodes
```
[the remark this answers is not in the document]
Beninho: maybe define an abstract class with byRawValuesSize method ?

tests/back/Pim/Enrichment/Specification/Bundle/Elasticsearch/Indexer/PhpMemoryLimitSpec.php
```diff
@@ -0,0 +1,70 @@
...
+
+    function it_returns_memory_when_set_with_megabytes()
+    {
+        $this->asBytes('10M')->shouldReturn(10485760);
+    }
+
+    function it_returns_memory_when_set_with_gigabytes()
+    {
+        $this->asBytes('10G')->shouldReturn(10737418240);
+    }
```
Beninho, line 69: add extra tests with extra chars like '10H' or 'AAA'

Beninho (approved):

PR #18517, opened 2022-11-28
CXP 1291: Select scope and locale

RiffVentura:
Added locale and channel selection as source parameters
For the reuse purposes some hooks have been moved to a more general namespace

image


components/catalogs/front/src/hooks/useChannelLocales.ts
```diff
@@ -1,5 +1,5 @@
 import {useQuery} from 'react-query';
-import {Locale} from '../models/Locale';
+import {Locale} from '../components/ProductSelection/models/Locale';
```
masacc, line 2:
```suggestion
import {Locale} from '../models/Locale';
```

masacc: And I think in `components/catalogs/front/src/components/ProductSelection/hooks/useLocales.ts` there could also be a mismatch in the path of `models/Local`

masacc (approved):

components/catalogs/front/src/hooks/useChannelLocales.ts
RiffVentura: nice catch ! fixed

willy-ahva (approved):

PR #18525, opened 2022-11-28
CXP-1354: Display errors messages

eoliveauakeneo:

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/components/catalogs/front/src/components/CatalogEdit/hooks/useCatalogForm.ts
+++ b/components/catalogs/front/src/components/CatalogEdit/hooks/useCatalogForm.ts
@@ -68,16 +68,14 @@ export const useCatalogForm = (id: string): Result => {
                         dispatch(action);
                         break;
                     case CatalogFormActions.SET_PRODUCT_SELECTION_CRITERIA:
-                        setErrors(errors.filter(
-                            (error) => !error.propertyPath.startsWith('[product_selection_criteria]'))
+                        setErrors(
+                            errors.filter(error => !error.propertyPath.startsWith('[product_selection_criteria]'))
                         );
                         setDirty(true);
                         dispatch(action);
                         break;
                     case CatalogFormActions.SET_PRODUCT_MAPPING:
-                        setErrors(errors.filter(
-                            (error) => !error.propertyPath.startsWith('[product_mapping]'))
-                        );
+                        setErrors(errors.filter(error => !error.propertyPath.startsWith('[product_mapping]')));
                         setDirty(true);
                         dispatch(action);
                         break;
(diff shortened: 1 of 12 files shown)


components/catalogs/front/src/components/CatalogEdit/hooks/useCatalogForm.ts
```diff
@@ -68,9 +68,16 @@ export const useCatalogForm = (id: string): Result => {
                         dispatch(action);
                         break;
                     case CatalogFormActions.SET_PRODUCT_SELECTION_CRITERIA:
-                        if (Object.keys(values.product_selection_criteria).length > Object.keys(action.value).length) {
```
eoliveauakeneo, line 71: @tseho after discussing it with @masacc I removed this control.

On "SET_PRODUCT_MAPPING" we can't have this kind of check and we will remove all the mapping errors on any changes.

The code is easier if the 2 cases have the same behavior, could we have your take on this ?

components/catalogs/front/src/components/ProductMapping/components/SelectAttributeDropdown.tsx
```diff
@@ -97,6 +97,11 @@ export const SelectAttributeDropdown: FC<Props> = ({code, onChange, error}) => {
                         </Dropdown.Overlay>
                     )}
                 </Dropdown>
+                {undefined !== error && (
+                    <Helper inline level="error">
+                        {error}
```
masacc, line 102: There will never be a translation of the error message with that way of doing things. Is it ok?

masacc (approved):

components/catalogs/front/src/components/CatalogEdit/hooks/useCatalogForm.ts
RiffVentura: I'm wondering if the error setting should be handled here at all:
Correct me if I'm wrong but If there are errors on both product selection and product mapping, dispatch will erase errors on the other tab

eoliveauakeneo: I added a filter to keep the errors corresponding to other tabs.

components/catalogs/front/src/components/ProductMapping/components/SelectAttributeDropdown.tsx
eoliveauakeneo: The error translation is handled php side, some were missing for the mapping and I added them like we discussed on slack.

RiffVentura (approved):

PR #18561, opened 2022-12-01
OCT-176: Product Mapping : on save, validate product mapping with product mapping schema requirements

masacc:

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc


components/catalogs/back/src/Infrastructure/Validation/ProductMapping/ProductMappingRespectsSchemaValidator.php
```diff
@@ -0,0 +1,139 @@
...
+ * @phpstan-type ProductMappingSchema array{
+ *      properties: array<string, array{type: string}>
+ * }
+ * @phpstan-import-type ProductMapping from Catalog
+ *
+ * @psalm-suppress PropertyNotSetInConstructor
+ */
+final class ProductMappingRespectsSchemaValidator extends ConstraintValidator
+{
+    private const TYPE_MAPPING = [
```
masacc, line 27:
```suggestion
    private const ALLOWED_TYPE_ASSOCIATION = [
```

components/catalogs/back/src/Infrastructure/Controller/Internal/GetCatalogErrorsAction.php
```diff
@@ -35,20 +35,22 @@ public function __invoke(Request $request, string $catalogId): Response
...
-            'product_mapping' => $catalogDomain->getProductMapping(),
+            'enabled' => $catalog->isEnabled(),
+            'product_selection_criteria' => $catalog->getProductSelectionCriteria(),
+            'product_value_filters' => $catalog->getProductValueFilters(),
+            'product_mapping' => $catalog->getProductMapping(),
         ];
 
         $violations = $this->validator->validate($catalogNormalized, [
-            new CatalogUpdatePayload(),
+            new CatalogUpdatePayload(
```
masacc, line 51: transform ternary into a setter in the constraint

guillaumefer-akeneo (approved):

components/catalogs/back/src/Infrastructure/Controller/Internal/GetCatalogErrorsAction.php
tseho: I don't think that the constraint would embed a "value".
What do you think of having `'product_mapping_schema' => ...,` in `$catalogNormalized` ?

In a validator, you can always access all the values using the context, so the `product_mapping` validator will be able to check `product_mapping_schema`.

WDYT ?

RiffVentura: :exploding_head:  neat !

masacc: OK, let's go with your solution. Not a big fan but none of them totally satisfies me. This one is at least more readable.

masacc: I have reworked as you proposed but I'm clearly not enthusiastic about this... I made a draft PR to see how it looks like: https://github.com/akeneo/pim-community-dev/pull/18579

I think this is more complicated to understand, especially in `UpdateCatalogAction.php`.

WDYT?

RiffVentura (approved):

tseho (approved):

PR #18584, opened 2022-12-05
GRF-646 : Add parent code in the returned json

jeremyBeaucousinAkeneo:

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc


src/Akeneo/Category/back/Infrastructure/Controller/ExternalApi/ListCategoriesController.php
```diff
@@ -100,6 +102,11 @@ public function __invoke(Request $request): JsonResponse|Response
                 $categoryApi->setPosition(($this->getPosition)($category));
             }
 
+            $parentId = $categoryApi->getParentId();
+            if ($parentId) {
+                $parentCategory = $this->getCategory->byId($parentId);
+                $categoryApi->setParentCode($parentCategory->getCode());
```
BryanS29-akeneo, line 108: getCode in the domain model returns a value object, you have to cast it as a string :
`(string) $parentCategory->getCode()`

BryanS29-akeneo (approved):

src/Akeneo/Category/back/Infrastructure/Controller/ExternalApi/ListCategoriesController.php
```diff
@@ -100,6 +102,11 @@ public function __invoke(Request $request): JsonResponse|Response
                 $categoryApi->setPosition(($this->getPosition)($category));
             }
 
+            $parentId = $categoryApi->getParentId();
+            if ($parentId) {
+                $parentCategory = $this->getCategory->byId($parentId);
```
amigaire, line 107: We are doing one call per category here, it does not seem efficient. Could we get the parents' codes in one query instead ?
It could be done with a JOIN in the query retrieving all categories, or one query with all the parent category ids

jeremyBeaucousinAkeneo: I tought about that, for a first implementation it seems sufficient, especially with pagination.

But I agree with you, I'll do the join, it go in the same way than the [GRF-647](https://akeneo.atlassian.net/browse/GRF-647)

etiennekerbastard (approved):

PR #18607, opened 2022-12-07
RAB-1148: Update NodeJS version

StevenVAIDIE:

In this PR, I upgraded the NodeJS version used to build the PIM to 18.

To do this I needed to : 
- Upgrade the version of webpack to 5.0 (https://github.com/webpack/webpack/releases/tag/v5.61.0) because webpack use md4 in webpack4. The other solution is to use `openssl-legacy-provider` (https://github.com/webpack/webpack/issues/14532) but we didn't really solve the problem we just hide it
- Upgrade the import-loader, export-loader because the version we used does not handle webpack5 => https://github.com/webpack-contrib/exports-loader/releases/tag/v1.0.0 & https://github.com/webpack-contrib/imports-loader/releases/tag/v1.0.0
- Migrated the webpack configuration to 5.0:
    - We cannot send custom argument => we now pass through environment variable
    - `loaders` config key have been removed we should use `use` => [enabled --openssl-legacy-provider on microfrontend](https://webpack.js.org/migrate/5/#update-outdated-options)
    - `maxModules` does not exist anymore => https://github.com/webpack/webpack/issues/11477
    - TerserPlugin remove some keys: cache, sourceMap => https://github.com/webpack-contrib/terser-webpack-plugin/blob/master/CHANGELOG.md#500-2020-10-14
    - `moduleIds: 'hashed'` doesn't exist anymore it's now `deterministic` => https://webpack.js.org/blog/2020-10-10-webpack-5-release/#deterministic-chunk-module-ids-and-export-names
    - WatchIgnorePlugin expect an object with path key in constructor => https://github.com/nestjs/docs.nestjs.com/issues/1554
    - loader option should not be string => https://github.com/webpack-contrib/expose-loader#using-configuration
    - We need to explicit polyfill "path", webpack 5 does no longer include a polyfill for this Node.js variable => https://webpack.js.org/migrate/5/#run-a-single-build-and-follow-advice
- We cannot use anymore timeZoneName with dateStyle => I just remove it 
- I enabled --openssl-legacy-provider on microfrontend, we need to update react-script to 5.0 is will do it in another PR

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc


front-packages/measurement/src/feature/context/unsaved-changes-context.ts
```diff
@@ -7,7 +7,7 @@ type UnsavedChangesContextValue = {
 
 const UnsavedChangesContext = createContext<UnsavedChangesContextValue>({
   hasUnsavedChanges: false,
-  setHasUnsavedChanges: () => {},
+  setHasUnsavedChanges: (_newValue: boolean) => {},
```
StevenVAIDIE, line 10: It fix: Types of property 'setHasUnsavedChanges' are incompatible.

webpack.config.js
```diff
@@ -240,12 +240,15 @@ const webpackConfig = {
     new webpack.ProvidePlugin({_: 'underscore', Backbone: 'backbone', $: 'jquery', jQuery: 'jquery'}),
 
     // Ignore these directories when webpack watches for changes
-    new webpack.WatchIgnorePlugin([
-      /node_modules\/(?!@akeneo)/,
-      path.resolve(rootDir, './app'),
```
StevenVAIDIE, line 245: app directory does not exist anymore

webpack.config.js
```diff
@@ -34,20 +32,18 @@ const webpackConfig = {
...
+          filename: 'vendor.min.js',
+          chunks: 'all'
         },
         main: {
           filename: 'main.min.js'
         }
       }
     },
-    moduleIds: 'hashed',
+    moduleIds: 'deterministic',
```
ValentinMumble, line 44: 🤷

ValentinMumble (approved):

ghoussard (approved):

PR #18626, opened 2022-12-08
GRF-654: Replace mock values by the real finder service

BryanS29-akeneo:

It's mainly about puttting the category id in parameter of the converter from standard format to user intent. We need it in `PermissionUserIntentFactory` to get the existing permissions for this category.

There is a lot of files changed because of the adaptation in tests, linter and phpstan (real format send by the front concerning permissions).

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc


src/Akeneo/Category/back/Domain/ValueObject/PermissionCollection.php
```diff
@@ -13,80 +13,103 @@ final class PermissionCollection
...
      */
-    public function addPermission(string $type, array $userGroupIds): self
+    public function addUserGroupsToPermission(string $permission, array $newUserGroups): self
     {
-        if (array_key_exists($type, $this->permissions)) {
-            foreach ($userGroupIds as $userGroupId) {
-                if (!in_array($userGroupId, $this->permissions[$type])) {
-                    $this->permissions[$type][] = $userGroupId;
+        if (array_key_exists($permission, $this->permissions)) {
+            $existingUserGroupsIds = array_map(fn ($existingUserGroup) => $existingUserGroup['id'], $this->permissions[$permission]);
```
antoinetrouve, line 37: There is a way that the key 'id" doesn't exist ? Maybe we have to verify if this key is not exist ?

src/Akeneo/Category/back/Domain/ValueObject/PermissionCollection.php
```diff
@@ -13,80 +13,103 @@ final class PermissionCollection
...
-    public function removePermission(string $type, array $userGroupIds): self
+    public function removeUserGroupsFromPermission(string $permission, array $userGroupsToRemove): self
     {
-        if (array_key_exists($type, $this->permissions)) {
-            foreach ($userGroupIds as $userGroupId) {
-                if (($key = array_search($userGroupId, $this->permissions[$type])) !== false) {
-                    $this->removedPermissions[$type][] = $userGroupId;
-                    unset($this->permissions[$type][$key]);
+        if (array_key_exists($permission, $this->permissions)) {
+            $existingUserGroupsIds = array_map(fn ($existingUserGroup) => $existingUserGroup['id'], $this->permissions[$permission]);
```
antoinetrouve, line 57: Just a little refacto. You use thiis map function sevrale times.
```suggestion
            $existingUserGroupsIds = $this->getUserGroupIds();
```

src/Akeneo/Category/back/Infrastructure/Controller/InternalApi/UpdateCategoryController.php
```diff
@@ -90,10 +89,9 @@ public function __invoke(Request $request, int $id): Response
                 Response::HTTP_BAD_REQUEST,
             );
         }
+
         $category = $this->getCategory->byId($id);
-        if ($category === null) {
-            throw new NotFoundHttpException('Category not found');
-        }
+        $category = $this->findCategoryAdditionalPropertiesRegistry->forCategory($category);
```
antoinetrouve, line 94: We doesn't manage the not found exception anymore?

BryanS29-akeneo: phpstan throwed me an error like category is never null at this state, that's why I removed the NotFound exception.

antoinetrouve: Je ne sais pas trop côté Phpstan mais le `$this->getCategory->byId()` peut renvoyer null dans la signature de sa fonction.

antoinetrouve: L.93 je ferais plutôt un $category->getId() plutot que id. Là du coup tu n'aurras pas besoin de gérer le not found. Par contre pour le premier appel à `$this->getCategory->byId()` (l.49) il faut gérer le null je pense.

src/Akeneo/Category/back/Infrastructure/Registry/FindCategoryAdditionalPropertiesRegistry.php
```diff
@@ -12,15 +12,15 @@ class FindCategoryAdditionalPropertiesRegistry implements FindCategoryAdditional
...
+        private readonly iterable $additionalPropertiesFinder,
     ) {
     }
 
     public function forCategory(Category $category): Category
     {
         foreach ($this->additionalPropertiesFinder as $finder) {
             if ($finder->isSupportedAdditionalProperties() === true) {
-                $category = $finder->execute($category);
+                return $finder->execute($category);
```
amigaire, line 23: It will only allow to get one additional property of the category (i.e. Permission).
If we need to get another additional property, it won't be possible because of the return. (let's imagine we want also to add a property LinkedProductsList)

PR #18634, opened 2022-12-09
PIM-10753: Fix 500 error in the API when patching a product metric with a mathematical notation

pierrejolly:

We have to invalidate value in hexadecimal or power exponent format. Actually, there is no function to do that in php (`ctype_digit` only works with integer and we can receive float here), so we have have to invalidate it with a regex.

see https://akeneo.atlassian.net/browse/PIM-10753

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc


ValentinMumble (approved):
🍖 ?

ghoussard:
> meat_on_bone ?

Done ✅

PR #18635, opened 2022-12-09
RAB-1124: Update monolog/monolog

ghoussard:

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc


src/Akeneo/Tool/Bundle/BatchBundle/Monolog/Handler/BatchLogHandler.php
```diff
@@ -41,6 +42,19 @@ public function __construct(
         $this->filePermission = $filePermission;
         $this->useLocking = $useLocking;
         $this->logDir = $logDir;
+
+        if (($phpMemoryLimit = Utils::expandIniShorthandBytes(ini_get('memory_limit'))) !== false) {
```
StevenVAIDIE, line 46: I think it's better to call the parent::_contruct instead of copying the monolog logic on your side. Here we bypass the constructor of Monolog :nose: :boom: 

For me something like that is better: 
```
    public function __construct(
        int $level,
        bool $bubble,
        ?int $filePermission,
        bool $useLocking,
        string $logDir
    ) {
        $this->logDir = $logDir;
        $url = $this->getRealPath($this->generateLogFilename());

        parent::__construct($url, $level, $bubble, $filePermission, $useLocking);
    }
```

StevenVAIDIE: And in setSubDirectory I think it's better to call the $this->close() before change the url. We didn't have the error for now but if one log is written before the setSubDirectory the stream will not be close and even if the url have change it will not change the written path

src/Akeneo/Tool/Bundle/BatchBundle/Monolog/Handler/BatchLogHandler.php
```diff
@@ -41,6 +42,9 @@ public function __construct(
         $this->filePermission = $filePermission;
         $this->useLocking = $useLocking;
```
StevenVAIDIE, line 43: You can remove the following lines: 
```
        $this->useLocking = $useLocking;
        $this->level = $level;
        $this->bubble = $bubble;
        $this->stream = null;
        $this->url = null;
```

StevenVAIDIE (approved):

PR #18645, opened 2022-12-12
RAB-1235: Add storage type badge & better error message

ValentinMumble:
- User-friendly error message when we cannot upload file (we cannot use exception message as they are formatted differently for each provider)
- Remove dirty hack that skipped upload/download steps for old jobs, as the feature is now released for more than 90 days
- Add placeholders on password & secret inputs
- Add storage type badge next to file path & import button when having a remote storage configured

Screenshot 2022-12-09 at 20 20 56
Screenshot 2022-12-09 at 20 21 20
Screenshot 2022-12-09 at 20 21 32


MadeWilson (approved):

ghoussard (approved):
The badge seems not to be vertically aligned with "File path" on screenshots 🤔

axelhms (approved):

PR #18651, opened 2022-12-12
PH-430: Store database install time

jmleroux:
Add a `pim_configuration` entry to store some installation data. Currently only the time of installation.
This data will be then used to check if a PIM is installed instead of the current test which only test for `oro_user` table existence. Because this table can exists even if there were issues in the install process.
The date is set only at the end of the full database installation and is then more strict than the mere presence of a table.

I also add a migration to initialize this date for existing PIMs. I use the `oro_user` table creation date to be consistent with the current test.

### DIFF
--- a/src/Akeneo/Platform/Bundle/InstallerBundle/Command/DatabaseCommand.php
+++ b/src/Akeneo/Platform/Bundle/InstallerBundle/Command/DatabaseCommand.php
@@ -5,6 +5,7 @@ namespace Akeneo\Platform\Bundle\InstallerBundle\Command;
 use Akeneo\Platform\Bundle\InstallerBundle\Event\InstallerEvent;
 use Akeneo\Platform\Bundle\InstallerBundle\Event\InstallerEvents;
 use Akeneo\Platform\Bundle\InstallerBundle\FixtureLoader\FixtureJobLoader;
+use Akeneo\Platform\Bundle\InstallerBundle\Persistence\Query\InstallTimeQuery;
 use Akeneo\Tool\Bundle\ElasticsearchBundle\ClientRegistry;
 use Akeneo\Tool\Component\Console\CommandExecutor;
 use Doctrine\DBAL\Connection;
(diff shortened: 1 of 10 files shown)


upgrades/schema/Version_7_0_20221212160000_add_database_install_time.php
```diff
@@ -0,0 +1,83 @@
...
+        return 'Initialize the database install time if not present';
+    }
+
+    public function up(Schema $schema): void
+    {
+        if ($this->installDataExists()) {
+            $this->addSql('SELECT 1');
+
+            return;
+        }
```
ghoussard, line 28: You can use skipIf here
```suggestion
        $this->skipIf($this->installDataExists(), 'Install data already exists');
```

ghoussard (approved):

src/Akeneo/Platform/Bundle/InstallerBundle/Persistence/Query/InstallTimeQuery.php
```diff
@@ -0,0 +1,33 @@
...
+namespace Akeneo\Platform\Bundle\InstallerBundle\Persistence\Query;
+
+use Doctrine\DBAL\Connection;
+
+/**
+ * @author JM Leroux <jean-marie.leroux@akeneo.com>
+ * @copyright 2022 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+final class InstallTimeQuery
```
nmarniesse, line 14: If the intent is to store install data (= not only the install time), maybe it can be renamed to `InstallDataQuery`?

nmarniesse (approved):

src/Akeneo/Platform/Bundle/InstallerBundle/Command/DatabaseCommand.php
```diff
@@ -184,6 +174,8 @@ protected function execute(InputInterface $input, OutputInterface $output): int
             );
         }
 
+        $this->installTimeQuery->setInstallTime();
```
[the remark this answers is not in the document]
jmleroux: Maybe. Probably. :)

PR #18656, opened 2022-12-12
CPM-854: Add Enabled condition

pierallard:


components/identifier-generator/front/src/feature/tabs/SelectionTab.tsx
```diff
@@ -1,46 +1,75 @@
...
+    conditions.map(condition => ({
+      id: uuid(),
+      ...condition,
+    }))
+  );
+
+  const removeIds: (conditionsWithId: ConditionsWithIdentifier) => Conditions = conditionsWithId => {
+    return conditionsWithId.map(conditionWithId => {
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
+      const {id, ...condition} = conditionWithId;
```
antoinearrive, line 31: why get an ID and not use it?

components/identifier-generator/front/src/feature/models/conditions/conditions.ts
```diff
@@ -0,0 +1,14 @@
+import {Enabled} from './enabled';
+
+type Condition = {
+  type: CONDITION_NAMES;
+} & Enabled;
```
antoinearrive, line 5: just to know, I don't know this syntax {}&, does it add the enabled object if it is not null?

antoinearrive (approved):

components/identifier-generator/front/src/feature/tabs/SelectionTab.tsx
pierallard: The notation 
```const {id, ...condition} = conditionWithId```
will "extract" the `id` property from `conditionWithId`.
This is the "beautiful way" to remove a key from an object in JS.

File: components/identifier-generator/front/src/feature/models/conditions/conditions.ts (hunk 2 above)
Comment (pierallard) [reply components/identifier-generator/front/src/feature/models/conditions/conditions.ts:5]: These are not objects, these are types.
If I declare `type AB = { a: string, b: string }` and `type BC = { a: string, b: string}`, so `type ABC = AB & AC` will declare a ` {a: string, b: string, c: string }`.
Here, I'm saying "A `Condition` has a `type` property (in the list of allowed names), and can be an `Enabled` condition.
In the future, we will have something like this
```
type Condition = {
  type: CONDITION_NAMES;
} & (Enabled | FamilyCondition | SimpleSelectCondition | AnotherCondition);
```

antoinearrive: ok thanks for the explanation

Beninho (approved):

mmetayer (approved):

cemorin (approved):

PR #18663, opened 2022-12-13
CPM-855: Add "Add condition" button

pierallard:


Beninho (approved):

components/identifier-generator/front/src/feature/tabs/conditions/AddConditionButton.tsx
```diff
@@ -0,0 +1,116 @@
...
+    items: [
+      {
+        code: CONDITION_NAMES.ENABLED,
+        defaultValue: {type: CONDITION_NAMES.ENABLED},
+      },
+    ],
+  },
+];
+
+const AddConditionButton: React.FC<AddConditionButtonProps> = ({onAddCondition}) => {
```
cemorin, line 30: Looks a lot like `AddPropertyButton`, can't it be a mutualized component? With an `items` property.
If the only difference is the code type, maybe it can be just considered as a string

cemorin (approved):

PR #18675, opened 2022-12-13
CPM-880: Fix double save on new identifier

pierallard:


mmetayer (approved):

pierallard:
@antoinearrive I asked myself and wasn't able to reproduce it by tests (it's a context data manipulation, and it's mocked)

PR #18690, opened 2022-12-14
PH-430: Use Application::find() to run the command in CommandExecutor

jmleroux:
The CommandExecutor should only run a _Command_ and not a full _Application_.
With this PR, we find the command into the application and run it without having to trick exception handler or autoexit.

Also, I come to this PR while working on upper stack and having issues with logger, because the logger is kind of _reset_ when using Application::run. I didn't manage to find the exact inner reason, but just running the command canceled this side effect.

Last but not least, using the find method is the recommended way in the docs: https://symfony.com/doc/current/console/calling_commands.html


StevenVAIDIE (approved):

jmleroux:
check frozen

jmleroux:
check frozen

jmleroux:
check frozen

jmleroux:
check frozen

PR #18692, opened 2022-12-14
RAB-785: Add possibility to check if we are in a context of initialization

StevenVAIDIE:

In this PR, I added the possibility to check that we are in the context of an installation or not. By doing this we can now check if the user can execute or not an export/import with local storage.

I renamed job_automation_local_storage by import_export_local_storage because it's not really on job_automation.
I activate this new flag when we load the fixture
I activate this new flag when we need to use the local storage
I replaced local by none storage when we didn't need the generated file

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc


MadeWilson (approved):

src/Akeneo/Platform/Bundle/InstallerBundle/Command/DatabaseCommand.php
```diff
@@ -110,6 +112,7 @@ protected function initialize(InputInterface $input, OutputInterface $output)
      */
     protected function execute(InputInterface $input, OutputInterface $output): int
     {
+        $this->installContextFeatureFlag->enable();
```
ahocquard, line 115: This is not a feature flag. Will it:
- allow to activate a feature between edition or not?
- or release progressively a feature?

If not, this is a context of execution.

tests/back/Integration/IntegrationTestsBundle/Loader/FixturesLoader.php
```diff
@@ -134,6 +92,7 @@ public function __destruct()
 
     public function load(Configuration $configuration): void
     {
+        $this->featureFlags->enable('import_export_local_storage');
```
ahocquard, line 95: I'm a little bit concerned about this "feature flag". Is it really a feature flag? A feature flag should not change across your runtime (it is immutable once fetched).

Should we not use DI instead, to force the use of local storage? For example, it is the case for ACL annotation cache, bucket, etc.

src/Akeneo/Platform/Bundle/InstallerBundle/Command/DatabaseCommand.php
StevenVAIDIE: Yes you have right it's a context of execution not a feature flag

tests/back/Integration/IntegrationTestsBundle/Loader/FixturesLoader.php
StevenVAIDIE: We don't want to force the usage of something we want to have the possibility to do it when we are not in SAAS. The only problem is that we use the import local storage for fixture that why we try to allow it also when we load the fixtures

axelhms (approved):

PR #18708, opened 2022-12-15
CPM-861: Update match method for family

pierallard:


components/identifier-generator/back/src/Domain/Model/ProductProjection.php
```diff
@@ -11,8 +11,9 @@
 final class ProductProjection
 {
     public function __construct(
-        private ?string $identifier,
-        private bool $enabled,
+        private readonly ?string $identifier,
+        private readonly bool $enabled,
+        private readonly ?string $familyCode,
```
Beninho, line 16: Why do not set a default value to null or empty string for familyCode ?

Beninho (approved):

components/identifier-generator/back/src/Domain/Model/ProductProjection.php
pierallard: If I put default value to null, I have to put next parameters with a default value too, and I think they will not have default value.

cemorin (approved):

PR #18722, opened 2022-12-16
CXP-1357 : Disable Catalogs when Channels associated with mapping are…

hogren:
… removed

image

image

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/components/catalogs/back/src/Infrastructure/Job/DisableCatalogsOnChannelRemovalTasklet.php
+++ b/components/catalogs/back/src/Infrastructure/Job/DisableCatalogsOnChannelRemovalTasklet.php
@@ -6,6 +6,7 @@ namespace Akeneo\Catalogs\Infrastructure\Job;
 
 use Akeneo\Catalogs\Application\Persistence\Catalog\DisableCatalogQueryInterface;
 use Akeneo\Catalogs\Application\Persistence\Catalog\GetCatalogIdsUsingChannelsAsFilterQueryInterface;
+use Akeneo\Catalogs\Application\Persistence\Catalog\GetCatalogIdsUsingChannelsInMappingQueryInterface;
 use Akeneo\Catalogs\Application\Service\DispatchInvalidCatalogDisabledEventInterface;
 use Akeneo\Tool\Component\Batch\Model\StepExecution;
 use Akeneo\Tool\Component\Connector\Step\TaskletInterface;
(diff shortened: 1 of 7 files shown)


components/catalogs/back/src/Infrastructure/Persistence/Catalog/GetCatalogIdsUsingChannelsInMappingQuery.php
```diff
@@ -0,0 +1,43 @@
...
+    public function __construct(
+        private readonly Connection $connection,
+    ) {
+    }
+
+    /**
+     * {@inheritdoc}
+     * @throws \Doctrine\DBAL\Driver\Exception
+     * @throws \Doctrine\DBAL\Exception
+     * @throws \JsonException
```
masacc, line 25: we usually don't add phpdoc for errors that we have no control over or we don't explicitly throw in the function. So I think you can remove these ones.

masacc (approved):

guillaumefer-akeneo (approved):

components/catalogs/back/src/Infrastructure/Job/DisableCatalogsOnChannelRemovalTasklet.php
```diff
@@ -35,7 +37,12 @@ public function execute(): void
         /** @var string[] $channelCodes */
         $channelCodes = $this->stepExecution->getJobParameters()->get('channel_codes');
 
-        $catalogsIds = $this->getCatalogIdsUsingChannelsAsFilterQuery->execute($channelCodes);
+        $catalogsIds = \array_unique(
+            \array_merge(
+                $this->getCatalogIdsUsingChannelsAsFilterQuery->execute($channelCodes),
+                $this->getCatalogIdsUsingChannelsInMappingQuery->execute($channelCodes)
```
RiffVentura, line 43: I'm wondering if it makes sense to combine both of these queries

RiffVentura (approved):

components/catalogs/back/src/Infrastructure/Job/DisableCatalogsOnChannelRemovalTasklet.php
hogren: @RiffVentura  You think about create a more big query for the two purposes ?
I asked @masacc about that. The debate is not closed. I choose this solution for the maintainability and reusability. The requests are very small and are executed on very small tables, there is no impacts in performance IMO.

eoliveauakeneo (approved):

components/catalogs/back/src/Infrastructure/Job/DisableCatalogsOnChannelRemovalTasklet.php
RiffVentura: Yes, combine both WHERE statements.
But for now, let's keep it simple and maintainable as you did :smile:

PR #18729, opened 2022-12-16
CXP-1348: Endpoint GET mapped-product/uuid

guillaumefer-akeneo:

- [x] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/components/catalogs/back/src/Application/Handler/GetMappedProductHandler.php
+++ b/components/catalogs/back/src/Application/Handler/GetMappedProductHandler.php
@@ -62,6 +62,7 @@ final class GetMappedProductHandler
         $productMappingSchema = $this->getProductMappingSchemaQuery->execute($catalog->getId());
         $productMapping = $catalog->getProductMapping();
 
+        /** @var RawProduct $product */
         return $this->productMapper->getMappedProduct($product, $productMappingSchema, $productMapping);
     }
 }
--- a/components/catalogs/back/src/Application/Handler/GetMappedProductHandler.php
+++ b/components/catalogs/back/src/Application/Handler/GetMappedProductHandler.php
@@ -7,6 +7,7 @@ namespace Akeneo\Catalogs\Application\Handler;
 use Akeneo\Catalogs\Application\Exception\CatalogNotFoundException;
 use Akeneo\Catalogs\Application\Persistence\Catalog\GetCatalogQueryInterface;
 use Akeneo\Catalogs\Application\Persistence\Catalog\Product\GetRawProductQueryInterface;
+use Akeneo\Catalogs\Application\Persistence\ProductMappingSchema\GetProductMappingSchemaQueryInterface;
 use Akeneo\Catalogs\Application\Service\ProductMapperInterface;
 use Akeneo\Catalogs\ServiceAPI\Exception\CatalogDisabledException;
 use Akeneo\Catalogs\ServiceAPI\Exception\CatalogNotFoundException as ServiceApiCatalogNotFoundException;
@@ -27,6 +28,7 @@ final class GetMappedProductHandler
         private GetCatalogQueryInterface $getCatalogQuery,
         private GetRawProductQueryInterface $getRawProductQuery,
         private ProductMapperInterface $productMapper,
+        private GetProductMappingSchemaQueryInterface $getProductMappingSchemaQuery,
     ) {
     }
 
@@ -51,6 +53,9 @@ final class GetMappedProductHandler
             throw new ServiceApiProductNotFoundException();
         }
 
-        return $this->productMapper->getMappedProduct($product, $catalog);
+        $productMappingSchema = $this->getProductMappingSchemaQuery->execute($catalog->getId());
+        $productMapping = $catalog->getProductMapping();
+
+        return $this->productMapper->getMappedProduct($product, $productMappingSchema, $productMapping);
     }
 }
--- a/components/catalogs/back/src/Application/Handler/GetMappedProductHandler.php
+++ b/components/catalogs/back/src/Application/Handler/GetMappedProductHandler.php
@@ -7,11 +7,12 @@ namespace Akeneo\Catalogs\Application\Handler;
 use Akeneo\Catalogs\Application\Exception\CatalogNotFoundException;
 use Akeneo\Catalogs\Application\Persistence\Catalog\GetCatalogQueryInterface;
 use Akeneo\Catalogs\Application\Persistence\Catalog\Product\GetRawProductQueryInterface;
+use Akeneo\Catalogs\Application\Persistence\Catalog\Product\IsProductBelongingToCatalogQueryInterface;
 use Akeneo\Catalogs\Application\Persistence\ProductMappingSchema\GetProductMappingSchemaQueryInterface;
 use Akeneo\Catalogs\Application\Service\ProductMapperInterface;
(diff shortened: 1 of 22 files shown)


components/catalogs/back/tests/Integration/Infrastructure/Controller/Public/GetMappedProductActionTest.php
```diff
@@ -0,0 +1,351 @@
...
+                'source' => 'description',
+                'scope' => 'print',
+                'locale' => null,
+            ],
+            'size_label' => [
+                'source' => 'size',
+                'scope' => null,
+                'locale' => null,
+            ],
+        ]);
```
RiffVentura, line 305: Integration test case has https://github.com/akeneo/pim-community-dev/blob/34b146c88c9c512cd8f43c5cb8173010f0e8e392/components/catalogs/back/tests/Integration/IntegrationTestCase.php#L360 method that you could use simply with named parameters like [this](https://github.com/akeneo/pim-community-dev/blob/59d434726f0ba0657f2430e6f789c29c015108ba/components/catalogs/back/tests/Integration/Infrastructure/Controller/Public/GetProductIdentifiersActionTest.php#L65)

components/catalogs/back/src/Infrastructure/Service/MapProducts.php
```diff
@@ -0,0 +1,104 @@
...
+ * @phpstan-import-type RawProduct from GetRawProductsQueryInterface
+ */
+class MapProducts implements MapProductsInterface
+{
+    public function __construct(
+        private readonly CatalogsMappingStorageInterface $catalogsMappingStorage,
+    ) {
+    }
+
+    public function __invoke(array $products, Catalog $catalog): array
```
RiffVentura, line 29: It would be better two have 2 public methods for single and collection mapping, this would avoid zero index access which is awkward (the actual function that does the mapping could be private and take in product mapping and its schema as arguments, and be used inside public methods). What do you think?

guillaumefer-akeneo: It's updated.

components/catalogs/back/src/Application/Service/ProductMapperInterface.php
```diff
@@ -0,0 +1,33 @@
...
+     * @return array<MappedProduct>
+     */
+    public function getMappedProducts(array $products, Catalog $catalog): array;
+
+    /**
+     * @param RawProduct $product
+     *
+     * @return MappedProduct
+     */
+    public function getMappedProduct(array $product, Catalog $catalog): array;
```
tseho, line 32:
```suggestion
    /**
     * @param RawProduct $product
     *
     * @return MappedProduct
     */
    public function getMappedProduct(array $product, Catalog $catalog): array;
```
I would rather have only one method and do the loop outside.

components/catalogs/back/src/Infrastructure/Service/ProductMapper.php
```diff
@@ -0,0 +1,126 @@
...
+        );
+    }
+
+    public function getMappedProduct(array $product, Catalog $catalog): array
+    {
+        $productMappingSchema = $this->getProductMappingSchema($catalog->getId());
+        $productMapping = $catalog->getProductMapping();
+
+        return $this->mapProduct($product, $productMappingSchema, $productMapping);
+    }
```
tseho, line 48:
```suggestion
    public function getMappedProduct(array $product, Catalog $catalog): array
    {
        $productMappingSchema = $this->getProductMappingSchema($catalog->getId());
        $productMapping = $catalog->getProductMapping();

        return $this->mapProduct($product, $productMappingSchema, $productMapping);
    }
```

With the previous suggestion, I was telling you that we should loop externally.
It raise one problem, the mapping schema will be loaded several times.

To prevent this, we have 2 options:
- we have a LRU cache in the class (stateful)
- we change the signature of the mapper to take the schema & the mapping as arguments instead of the Catalog.

eoliveauakeneo (approved):

components/catalogs/back/src/Application/Persistence/Catalog/Product/GetRawProductQueryInterface.php
```diff
@@ -0,0 +1,28 @@
+<?php
+declare(strict_types=1);
+
+namespace Akeneo\Catalogs\Application\Persistence\Catalog\Product;
+
+/**
+ * @copyright 2022 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ *
+ * @phpstan-type RawProduct array{
```
masacc, line 10: can you import it from `components/catalogs/back/src/Application/Persistence/Catalog/Product/GetRawProductsQueryInterface.php` ?

components/catalogs/back/src/Application/Persistence/Catalog/Product/GetRawProductQueryInterface.php
```diff
@@ -0,0 +1,28 @@
+<?php
+declare(strict_types=1);
```
masacc, line 2: missing blank line

components/catalogs/back/tests/Integration/Infrastructure/Controller/Public/GetMappedProductActionTest.php
```diff
@@ -0,0 +1,355 @@
...
+
+        $this->commandBus = self::getContainer()->get(CommandBus::class);
+
+        $this->disableExperimentalTestDatabase();
+        $this->purgeDataAndLoadMinimalCatalog();
+
+        $this->createUser('admin', ['IT support'], ['ROLE_ADMINISTRATOR']);
+    }
+
+    public function testItGetMappedProductByCatalogIdAndProductUuid(): void
```
masacc, line 40:
```suggestion
    public function testItGetsMappedProductByCatalogIdAndProductUuid(): void
```
:smile:

masacc (approved):

components/catalogs/back/src/Application/Persistence/Catalog/Product/GetRawProductQueryInterface.php
RiffVentura: I think it should be other way around: Single Entity defines it and Multi Entity should reference it

masacc: I agree

components/catalogs/back/src/Application/Handler/GetMappedProductHandler.php
```diff
@@ -0,0 +1,68 @@
...
+
+        if (!$catalog->isEnabled()) {
+            throw new CatalogDisabledException();
+        }
+
+        if (!$this->isProductBelongingToCatalogQuery->execute($catalog, $query->getProductUuid())) {
+            throw new ProductNotFoundException();
+        }
+
+        $product = $this->getRawProductQuery->execute($query->getProductUuid());
```
RiffVentura, line 56: This should be wrapped with the try-catch to prevent product selection errors: check the catalog validity-> disable catalog and send the appropriate event

RiffVentura: see https://github.com/akeneo/pim-community-dev/pull/18747/files

RiffVentura (approved):

PR #18737, opened 2022-12-20
CPM-762: Fix memory leak on pricemask

pierallard:
When computing the masks of a price value, if the number of currencies is big (> 20), the mask generation creates a memory error because the number of masks is 2^19, and it takes a lot of memory.

A first solution was done to improve the computation of the mask and avoid creation of big arrays (see https://github.com/akeneo/pim-community-dev/commit/3166742a438db5da1ab5db32c5fe93fbc53f7553 ).

This solution will remove from the currencies list the currencies which are not active anymore. When you had in the past 20 active currencies, you can save a product with these currencies. But if you deactive a currency, the value remains in the product.
This PR removes the non-active currencies.


mmetayer:
replaced by #18738

PR #18748, opened 2022-12-21
CXP-1177: Update legal text in the authentication step

eoliveauakeneo:

Today users agree with the app privacy when connecting an app that requests authentication scopes. 

When these scopes are updated, we re-display the authentication step to warn our users that the app needs more information. They shouldn’t have to agree to the app privacy policy again. 

This is why, when the app updates its authentication scopes, we hide the agreement checkbox and only display the existing subtext as in the screenshot

image

### DIFF
--- a/src/Akeneo/Connectivity/Connection/back/Infrastructure/Apps/Controller/Internal/GetWizardDataAction.php
+++ b/src/Akeneo/Connectivity/Connection/back/Infrastructure/Apps/Controller/Internal/GetWizardDataAction.php
@@ -55,7 +55,7 @@ final class GetWizardDataAction
             throw new NotFoundHttpException("Invalid app identifier");
         }
 
-        [$oldAuthorizationScopeMessages, $newAuthorizationScopeMessages] = $this->getAuthorizationScopes($app->getId(), $appAuthorization);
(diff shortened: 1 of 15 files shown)


src/Akeneo/Connectivity/Connection/back/Infrastructure/Apps/Controller/Internal/GetWizardDataAction.php
```diff
@@ -86,7 +87,10 @@ private function getAuthorizationScopes(string $appId, AppAuthorization $appAuth
         $oldAuthorizationScopeMessages = $isFirstConnection ? null : $this->scopeMapperRegistry->getMessages($originalScopes);
         $newAuthorizationScopeMessages = $this->scopeMapperRegistry->getMessages($newScopes);
 
-        return [$oldAuthorizationScopeMessages, $newAuthorizationScopeMessages];
+//        dump($oldAuthorizationScopeMessages);
+//        dd($newAuthorizationScopeMessages);
```
guillaumefer-akeneo, line 91: To remove

guillaumefer-akeneo (approved):

RiffVentura (approved):

masacc (approved):

PR #18774, opened 2022-12-27
OCT-196 : Fix the width of catalog header to display the save button on small resolution

hogren:
With a window width below 1380px, buttons at the top right corner began to be masked. We need to horizontally scroll to use them.
image

I couldn't use the `max-width` solution proposed by @eoliveauakeneo, because the value to enter is different for each page.
I found a solution with the `overflow` rule.
However, there were two problems with it. 
The first was a truncated button on several pages. 
image
I adapted for this the header button list construction.

The last but not least, the user drop down menu was hidden. I tried to use `overflow-y: visible` in combination with `overflow-x: hidden`. But, as we see in this stackoverflow post, this does not work :
https://stackoverflow.com/questions/6421966/css-overflow-x-visible-and-overflow-y-hidden-causing-scrollbar-issue
So I used a workaround and I place a comment to explain that. I am not proud about that.
Another solution would be to accept a `min-height` for the header… But I think it is not acceptable.

In the end, we can reduce the width to approximately 1000px without problem, except the inaccessibility of the breadcrumb.
image

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/src/Akeneo/Connectivity/Connection/front/src/common/components/PageHeader.tsx
+++ b/src/Akeneo/Connectivity/Connection/front/src/common/components/PageHeader.tsx
@@ -51,6 +51,10 @@ const IllustrationContainer = styled.div`
     }
 `;
 
+const MainContainer = styled.div`
+    overflow-x: hidden;
+`
+
 export const PageHeader = ({
     children: title,
     breadcrumb,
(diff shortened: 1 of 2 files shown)


masacc (approved):

hogren:
@akeneo/squad-octopus : thanks to @eoliveauakeneo, there is already a css trick in the code

PR #18782, opened 2022-12-29
OCT-202: add placeholder on the source selector in mapping

hogren:
image

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/components/catalogs/front/src/components/ProductMapping/components/SelectAttributeDropdown.tsx
+++ b/components/catalogs/front/src/components/ProductMapping/components/SelectAttributeDropdown.tsx
@@ -14,10 +14,9 @@ type Props = {
     code: string;
     onChange: (value: Attribute) => void;
     error: string | undefined;
-    placeholder: string | undefined;
 };
 
-export const SelectAttributeDropdown: FC<Props> = ({code, onChange, error, placeholder}) => {
+export const SelectAttributeDropdown: FC<Props> = ({code, onChange, error}) => {
     const translate = useTranslate();
     const [isOpen, setIsOpen] = useState<boolean>(false);
     const [search, setSearch] = useState<string>('');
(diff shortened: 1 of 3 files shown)


components/catalogs/front/src/components/ProductMapping/components/SelectAttributeDropdown.tsx
```diff
@@ -46,7 +47,8 @@ export const SelectAttributeDropdown: FC<Props> = ({code, onChange, error}) => {
                         onMouseDown={openDropdown}
                         emptyResultLabel={translate('akeneo_catalogs.common.select.no_matches')}
                         openLabel={translate('akeneo_catalogs.common.select.open')}
-                        value={attribute?.label ?? (code.length > 0 ? `[${code}]` : '')}
+                        value={attribute?.label ?? (code.length > 0 ? `[${code}]` : null)}
+                        placeholder={placeholder ?? ''}
```
masacc, line 51: Why using a props for that? Can it be hardcoded?

eoliveauakeneo (approved):

components/catalogs/front/src/components/ProductMapping/components/SelectAttributeDropdown.tsx
hogren: If we want to reuse the component, the placeholder need to be changed. For me, the message must be associated with the application component and not the design component. Are you agree ? @masacc @RiffVentura

tseho: SelectAttributeDropdown is not a design component and is not reusable.
It's why it exists under the directory `src/components/ProductMapping/components/`.
Its only purpose it to be used in ProductMapping, not anywhere else.
If you try to use it somewhere else, you will see that `make catalogs-lint-front` will forbid you from doing so.

tseho: I know it can be seen as an extreme interpretation of Single Responsiblity Principle of SOLID.
> A software module has one, and only one, reason to change.

And yes, from a technical point-of-view, everything can be build as reusable components.

But resisting the urge of doing so is really important for the maintainability of our project.
If you missed it, we already have a component doing something similar in the ProductSelection.
When you think about having a piece of code reusable in different functional contexts, you always have to ask yourself, "Am I 100% sure that this piece will match the needs of multiple functional contexts and multiple business needs for the 10 next years ?"
If you cannot answer yes to this question, doing like we do is the way to go.

hogren: @tseho : I understand this principle. I adapted the code.
cc @masacc @RiffVentura

masacc (approved):

PR #18805, opened 2023-01-03
CPM 901 : Add pills on tabs

Beninho:

When there is a validation error,
The error message as well as a pill must be displayed in the correct tab.

Screenshot from 2023-01-03 15-12-02

- [X] Tests

### DIFF
--- a/components/identifier-generator/front/src/feature/components/TabValidationErrors.tsx
+++ b/components/identifier-generator/front/src/feature/components/TabValidationErrors.tsx
@@ -1,0 +1,34 @@
+import React, {useMemo} from 'react';
+import {Styled} from './Styled';
+import {Violation} from '../validators';
+
+type Props = {
+  errors: Violation[];
+};
+
+const TabValidationErrors: React.FC<Props> = ({errors}) => {
+  const displayedErrors = useMemo(
+    () => errors?.map(({message}) => message).filter((value, index, self) => self.indexOf(value) === index),
+    [errors]
+  );
+
+  return (
+    <>
+      {displayedErrors?.length > 0 && (
+        <Styled.MainErrorHelper level="error">
+          {displayedErrors.length > 1 ? (
+            <Styled.ErrorList>
+              {displayedErrors.map(message => (
+                <li key={message}>{message}</li>
+              ))}
+            </Styled.ErrorList>
+          ) : (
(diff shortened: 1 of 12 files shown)


components/identifier-generator/front/src/feature/components/TabValidationErrors.tsx
```diff
@@ -0,0 +1,34 @@
...
+  const displayedErrors = useMemo(
+    () => errors?.map(({message}) => message).filter((value, index, self) => self.indexOf(value) === index),
+    [errors]
+  );
+
+  return (
+    <>
+      {displayedErrors?.length > 0 && (
+        <Styled.MainErrorHelper level="error">
+          {displayedErrors.length > 1 ? (
```
pierallard, line 19: I'm not a fan of having a dedicated Styled Element just for simple/multiple.
If you want to do this, please use :only-child or something else directly in the CSS.

components/identifier-generator/front/src/feature/pages/CreateOrEditGeneratorPage.tsx
```diff
@@ -43,13 +43,15 @@ const CreateOrEditGeneratorPage: React.FC<CreateOrEditGeneratorProps> = ({
...
-  const hasStructureErrors = useMemo(
-    () =>
-      validationErrors.filter(item => item?.path?.includes('structure') || item?.path?.includes('delimiter'))?.length >
-      0,
+  const getErrorByPathname = useCallback(
+    (pathNames: string[]) => validationErrors.filter(item => pathNames.some(v => item?.path?.includes(v))),
     [validationErrors]
   );
 
+  const generalValidationErrors = getErrorByPathname(['labels']);
```
pierallard, line 51: byPathNames*

cemorin (approved):

components/identifier-generator/front/src/feature/components/TabValidationErrors.tsx
cemorin: It does not only handle specific style, the multiple is specific to `ul` and the simple is a `p`. The style should be applied only if it's in this condition no?
Or the style could be inside `MainErrorHelper` and contain both `ul` and `p` styles because I don't think this styled component could be reusable for now

PR #18820, opened 2023-01-04
OCT-203: refactor ProductSchemaValidatorTest

tseho:

Instead of having all tests in the same file, which will not scale with the upcoming changes to the meta schema, I propose to use a DataProvider which will load the schemas to test from a directory.

If you want to add a valid schema to test, add it to `ProductSchema/valid`.
If you want to add an invalid schema to test, add it to `ProductSchema/invalid`.

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/components/catalogs/back/tests/Integration/Infrastructure/Validation/ProductSchema/invalid/0.0.2-with-invalid-uuid-extra-fields.json
+++ b/components/catalogs/back/tests/Integration/Infrastructure/Validation/ProductSchema/invalid/0.0.2-with-invalid-uuid-extra-fields.json
@@ -1,9 +1,0 @@
-{
-  "$schema": "https://api.akeneo.com/mapping/product/0.0.2/schema",
-  "properties": {
-    "uuid": {
-      "type": "boolean",
-      "title": "Description"
-    }
-  }
-}
--- a/components/catalogs/back/tests/Integration/Infrastructure/Validation/ProductSchema/invalid/0.0.2-with-invalid-uuid-extra-fields.json
+++ b/components/catalogs/back/tests/Integration/Infrastructure/Validation/ProductSchema/invalid/0.0.2-with-invalid-uuid-extra-fields.json
@@ -1,0 +1,9 @@
+{
+  "$schema": "https://api.akeneo.com/mapping/product/0.0.2/schema",
+  "properties": {
+    "uuid": {
+      "type": "boolean",
+      "title": "Description"
+    }
+  }
+}
--- a/components/catalogs/back/tests/Integration/Infrastructure/Validation/ProductSchema/invalid/0.0.2-with-invalid-uuid-type.json
+++ b/components/catalogs/back/tests/Integration/Infrastructure/Validation/ProductSchema/invalid/0.0.2-with-invalid-uuid-type.json
@@ -1,5 +1,6 @@
 {
   "$schema": "https://api.akeneo.com/mapping/product/0.0.2/schema",
+  "description": "The data must must match the const value at #/properties/uuid/type",
   "properties": {
     "uuid": {
       "type": "boolean"
--- a/components/catalogs/back/tests/Integration/Infrastructure/Validation/ProductSchema/invalid/0.0.2-with-invalid-uuid-type.json
+++ b/components/catalogs/back/tests/Integration/Infrastructure/Validation/ProductSchema/invalid/0.0.2-with-invalid-uuid-type.json
@@ -1,0 +1,8 @@
+{
+  "$schema": "https://api.akeneo.com/mapping/product/0.0.2/schema",
+  "properties": {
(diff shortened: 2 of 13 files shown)


hogren (approved):
What do you think about separate tests cases files in subsubdir for the schema version and explicitly indicate that the version is for the schema, with "0.0.1_schema" (or something like that) for a more obvious reading ?
Ex: `ProductSchema/invalid/0.0.1-with-invalid-type-number.json` => `ProductSchema/invalid/0.0.1_schema/with-invalid-type-number.json`

tseho:
I though about it but I went for simplicity.
Having 2 directories PER version, `invalid/0.0.1` & `valid/0.0.1`, will create a really annoying directory structure.
And for valid tests, you would only have one file per directory.

I expect us to have around 10 versions at end of Q1. It would mean that our tests would be separated into 20 directories instead of 2. That's extreme coming from the current situation where everything was in only one file.

I'm open to see what the others think about it.

RiffVentura:
I'm ok with two folder approach
however, there might be a lot of copy-paste if we want to be exhaustive: since changes are incremental do we test that no regression is made?

tseho:
> however, there might be a lot of copy-paste if we want to be exhaustive: since changes are incremental do we test that no regression is made?

For valid JSON Schema, the same test will grow each time we add something new in it.
For invalid scenarios, yes, there could be some copy/paste between versions.
But we cannot exclude that we officially release the feature, we drop old version and maintain only the last one as a `1.0.0`.
If we do this, the last version should be perfectly tested anyway.

components/catalogs/back/tests/Integration/Infrastructure/Validation/ProductSchema/invalid/0.0.2-with-invalid-uuid-extra-fields.json
```diff
@@ -0,0 +1,9 @@
+{
+  "$schema": "https://api.akeneo.com/mapping/product/0.0.2/schema",
+  "properties": {
+    "uuid": {
+      "type": "boolean",
+      "title": "Description"
+    }
```
masacc, line 7:
```suggestion
    "uuid": {
      "type": "string",
      "extra_field": "some extra data"
    }
```

components/catalogs/back/tests/Integration/Infrastructure/Validation/ProductSchemaValidatorTest.php
```diff
@@ -55,120 +55,24 @@ public function testItRejectsTheSchema(string $schema): void
...
+     * @return array<string, array{schema: string}>
+     */
+    private function readFilesFromDirectory(string $directory): array
+    {
+        $files = scandir($directory);
+        $files = array_filter($files, fn ($file) => !str_starts_with($file, '.'));
+
+        return array_combine($files, array_map(fn ($file) => [
+            'schema' => file_get_contents($directory . '/' . $file),
+        ], $files));
```
masacc, line 76:
```suggestion
        $files = \scandir($directory);
        $files = \array_filter($files, fn ($file) => !\str_starts_with($file, '.'));

        return \array_combine($files, \array_map(fn ($file) => [
            'schema' => \file_get_contents($directory . '/' . $file),
        ], $files));
```

components/catalogs/back/tests/Integration/Infrastructure/Validation/ProductSchema/invalid/0.0.2-with-invalid-uuid-extra-fields.json
tseho: I want to remove this scenario. I don't see the value. What do you think ?

components/catalogs/back/tests/Integration/Infrastructure/Validation/ProductSchema/invalid/0.0.2-with-invalid-uuid-type.json
```diff
@@ -0,0 +1,9 @@
+{
+  "$schema": "https://api.akeneo.com/mapping/product/0.0.2/schema",
+  "description": "The data must must match the const value at #/properties/uuid/type",
```
masacc, line 3:
```suggestion
  "description": "The data must match the const value at #/properties/uuid/type",
```

masacc (approved):

components/catalogs/back/tests/Integration/Infrastructure/Validation/ProductSchema/invalid/0.0.2-with-invalid-uuid-type.json
tseho: I cannot, this is word for word the error outputed by the JSON Schema validator library.

PR #18832, opened 2023-01-05
add the --complete argument to avoid the deprecation notice

lasdou:

Recently the pim:installer:db command was raising a deprecation notice when running the doctrine:schema:update see following screenshot

Capture d’écran 2023-01-02 à 14 57 06

adding the --complete argument solves it

here is the output of the doctrine:schema:update command help. about the complete argument :

```
Description:
  Executes (or dumps) the SQL needed to update the database schema to match the current mapping metadata

Usage:
  doctrine:schema:update [options]
    Options:
      --em=EM           Name of the entity manager to operate on
      --complete        If defined, all assets of the database which are not relevant to the current metadata will be dropped.
      --dump-sql        Dumps the generated SQL statements to the screen (does not execute them).
    -f, --force           Causes the generated SQL statements to be physically executed against your database.
  -h, --help            Display help for the given command. When no command is given display help for the list command
  -q, --quiet           Do not output any message
  -V, --version         Display this application version
      --ansi|--no-ansi  Force (or disable --no-ansi) ANSI output
  -n, --no-interaction  Do not ask any interactive question
  -e, --env=ENV         The Environment name. [default: "prod"]
      --no-debug        Switch off debug mode.
    -v|vv|vvv, --verbose  Increase the verbosity of messages: 1 for normal output, 2 for more verbose output and 3 for debug

Help:
  The doctrine:schema:update command generates the SQL needed to
  synchronize the database schema with the current mapping metadata of the
  default entity manager.
  
  For example, if you add metadata for a new column to an entity, this command
  would generate and output the SQL needed to add the new column to the database:
  
  doctrine:schema:update --dump-sql
  
  Alternatively, you can execute the generated queries:
  
  doctrine:schema:update --force
  
  If both options are specified, the queries are output and then executed:
  
  doctrine:schema:update --dump-sql --force
  
  Finally, be aware that if the --complete option is passed, this
  task will drop all database assets (e.g. tables, etc) that are *not* described
  by the current metadata. In other words, without this option, this task leaves
  untouched any "extra" tables that exist in the database, but which aren't
  described by any metadata. Not passing that option is deprecated.
  
  Hint: If you have a database with tables that should not be managed
  by the ORM, you can use a DBAL functionality to filter the tables and sequences down
  on a global level:
  
      $config->setFilterSchemaAssetsExpression($regexp);
```

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc


nmarniesse (approved):

BitOne (approved):

lasdou:
up !

jmleroux:
Closed in favor of #19590

PR #18834, opened 2023-01-06
GRF-674: Add filter on empty value when export categories

antoinetrouve:
Via the public api : `api/rest/v1/categories?with_enriched_attributes=true`.

After discussion: We have to not saving empty data in db. 

**Before:** 

GRF-674-export-with-empty-value

**After:** 
GRF-674-exportnon-empty-value

- [X] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc


BryanS29-akeneo (approved):

src/Akeneo/Category/back/Infrastructure/Controller/ExternalApi/ListCategoriesController.php
```diff
@@ -103,4 +105,41 @@ public function __invoke(Request $request): JsonResponse|Response
...
+     * @return array{
+     *     code: string,
+     *     parent: string|null,
+     *     updated: string,
+     *     labels: array<string, string>|null,
+     *     position: int|null,
+     *     values: array<string, array<string, mixed>>|null
+     * } only return normalized category with enriched value data.
+     */
+    private function filterEmptyValues(array $normalizedCategory): array
```
olivierpontier, line 128: It should the responsibility of the ExternalApiCategory::normalize method to check if values are empty or not before returning them. It should also simplify the way to cover this function by unit tests

antoinetrouve: Are we sure that at each time we normalize we want to filter empty value ? the issue is only about the export categories end point and not the normalization process that's why I choose to put this logic int the controller.

olivierpontier: From my point of view, the ExternalApiCategory object is used to format the response for the API and has no other use. So filtering empty values could be one of those responsibilities. This would hide the manipulation of the "value_collection" structure within this class. It doesn't make sense to me in the controller.
It could be interesting to share this topic with the rest of the team in order to get feedback on the separation of concerns in general. It could help to design future solutions. What do you think?

antoinetrouve: As per discussion this behavior has not o be exist. So we decide to filter the value at the beginning of the saving process (before setting the applier).

jeremyBeaucousinAkeneo (approved):

PR #18835, opened 2023-01-06
CXP-1378: Added textarea to the source list

RiffVentura:
Added pim_catalog_textarea attributes to the list of possible attributes when the mapping target expects a string


masacc (approved):

components/catalogs/back/tests/Integration/Infrastructure/Validation/ProductMapping/AttributeSource/AttributeTextareaSourceTest.php
```diff
@@ -0,0 +1,211 @@
...
+                'expectedMessage' => 'This value should be of type string.',
+            ],
+            'invalid scope' => [
+                'attribute' => [
+                    'code' => 'name',
+                    'type' => 'pim_catalog_textarea',
+                    'group' => 'other',
+                    'scopable' => true,
+                    'localizable' => true,
+                ],
```
hogren, line 141: I am ok that a test need less abstraction and factorization than other code sections. But I just think that factorize this table which is copied/pasted many times can be useful to reduce the code length, and avoid errors.

hogren (approved):

components/catalogs/back/tests/Integration/Infrastructure/Validation/ProductMapping/AttributeSource/AttributeTextareaSourceTest.php
masacc: I think it's more readable if everything is here, copy/paste is not a bad practice in tests

PR #18844, opened 2023-01-09
CPM-890: Add ScopeSelector and its hook

cemorin:

Scope selector will be used in the simple select condition line

Screenshot from 2023-01-09 13-24-02

- [x] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/components/identifier-generator/front/src/feature/components/ScopeSelector.tsx
+++ b/components/identifier-generator/front/src/feature/components/ScopeSelector.tsx
@@ -1,0 +1,49 @@
+import React from 'react';
+import {Helper, SelectInput, SkeletonPlaceholder, Table} from 'akeneo-design-system';
+import {LabelCollection, useTranslate, useUserContext} from '@akeneo-pim-community/shared';
+import {useGetScopes} from '../hooks/useGetScopes';
+import {Unauthorized} from '../errors';
+
+type ScopeSelectorProps = {
+  value: string | null;
+  onChange: (code: string) => void;
+};
+
+const ScopeSelector: React.FC<ScopeSelectorProps> = ({value, onChange}) => {
+  const translate = useTranslate();
+  const currentCatalogLocale = useUserContext().get('catalogLocale');
+  const {data: options, isLoading, error} = useGetScopes();
+
+  const getLabel = (labels: LabelCollection, code: string) => labels[currentCatalogLocale] || `[${code}]`;
+
+  if (error) {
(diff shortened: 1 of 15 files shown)


components/identifier-generator/front/src/feature/components/ScopeSelector.tsx
```diff
@@ -0,0 +1,49 @@
+import React from 'react';
+import {Helper, SelectInput, SkeletonPlaceholder, Table} from 'akeneo-design-system';
+import {LabelCollection, useTranslate, useUserContext} from '@akeneo-pim-community/shared';
+import {useGetScopes} from '../hooks/useGetScopes';
```
pierallard, line 4: import in index.ts

components/identifier-generator/front/src/feature/components/ScopeSelector.tsx
```diff
@@ -0,0 +1,49 @@
...
+import {Unauthorized} from '../errors';
+
+type ScopeSelectorProps = {
+  value: string | null;
```
pierallard, line 8: Can you declare a "Scope" in Model folder ? (Which is a string)

components/identifier-generator/front/src/feature/tests/fixtures/scopes.ts
```diff
@@ -0,0 +1,28 @@
...
+  {'id':90, 'code':'fr_FR', 'label':'French (France)', 'region':'France', 'language':'French'}
+];
+const channel = {
+  currencies:['USD', 'EUR'],
+  locales:locales,
+  category_tree:'master',
+  conversion_units:[],
+  meta: {}
+};
+export const mockedScopes = [
```
pierallard, line 13: I'm not fan of exporting data from tests/ folder... But I don't have any other solutions than copy/paste. If you have a better idea, I'm "preneur"

pierallard (approved):

components/identifier-generator/front/src/feature/tests/fixtures/scopes.ts
cemorin: It does not look like there's a global good practice about this, so whatever you'd prefer!
Would you prefer `src/feature/fixtures` ? (or even `__fixtures__`)

cemorin: As seen together, I exported the fixtures as default so that it's not automatically importer anywhere else

Beninho (approved):

PR #18848, opened 2023-01-09
CPM-896: Add "match" method for simple select

pierallard:


components/identifier-generator/back/src/Infrastructure/Subscriber/SetIdentifiersSubscriber.php
```diff
@@ -198,4 +204,22 @@ private function updatePropertyPath(
...
+        foreach ($product->getValues() as $value) {
+            Assert::isInstanceOf($value, ValueInterface::class);
+            $key = join('_', [
+                $value->getLocaleCode() ?? '<all_locales>',
+                $value->getScopeCode() ?? '<all_channels>',
+            ]);
+            $result[$value->getAttributeCode()][$key] = $value->getData();
+        }
+
+        return $result;
```
mmetayer, line 223:
```suggestion
        return \array_map(
            static fn (ValueInterface $value) => $value->getData(),
            $product->getValues()->toArray()
        );
```

components/identifier-generator/back/src/Domain/Model/ProductProjection.php
```diff
@@ -31,4 +41,11 @@ public function familyCode(): ?string
     {
         return $this->familyCode;
     }
+
+    public function value(string $attributeCode, ?string $locale = null, ?string $scope = null): mixed
+    {
+        $key = \join('_', [$locale ?? '<all_locales>', $scope ?? '<all_channels>']);
+
+        return $this->productValues[$attributeCode][$key] ?? null;
+    }
```
mmetayer, line 50:
```suggestion
    public function value(string $attributeCode, ?string $localeCode, ?string $channelCode): mixed
    {
        $key = \sprintf('%s-%s-%s', $attributeCode, $channelCode ?? '<all_channels>', $localeCode ?? '<all_locales>');

        return $this->productValues[$key] ?? null;
    }
```

mmetayer (approved):

Beninho (approved):

PR #18849, opened 2023-01-09
First prototype of a solution to avoid user id problem with docker

hogren:
This PR is just for prototyping a solution to avoid the problem we have when we are not the UID 1000 on our OS.

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,10 +1,13 @@
-DOCKER_COMPOSE = docker-compose
-NODE_RUN = $(DOCKER_COMPOSE) run -u node --rm -e YARN_REGISTRY -e PUPPETEER_SKIP_CHROMIUM_DOWNLOAD=1 -e PUPPETEER_EXECUTABLE_PATH=/usr/bin/google-chrome node
+DEV_UID ?= $(shell id -u || echo 1000)
+DEV_GID ?= $(shell id -g)
+DOCKER_COMPOSE = DEV_UID=${DEV_UID} DEV_GID=${DEV_GID} docker-compose
+NODE_RUN = $(DOCKER_COMPOSE) run --user "$(DEV_UID):$(DEV_GID)" --rm -e YARN_REGISTRY -e PUPPETEER_SKIP_CHROMIUM_DOWNLOAD=1 -e PUPPETEER_EXECUTABLE_PATH=/usr/bin/google-chrome node
 YARN_RUN = $(NODE_RUN) yarn
-PHP_RUN = $(DOCKER_COMPOSE) run --rm php php
-PHP_EXEC = $(DOCKER_COMPOSE) exec -u www-data fpm php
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -1,11 +1,14 @@
...
-NODE_RUN = $(DOCKER_COMPOSE) run -u node --rm -e YARN_REGISTRY -e PUPPETEER_SKIP_CHROMIUM_DOWNLOAD=1 -e PUPPETEER_EXECUTABLE_PATH=/usr/bin/google-chrome node
+DEV_UID ?= $(shell id -u || echo 1000)
+DEV_GID ?= $(shell id -g)
+DOCKER_COMPOSE = DEV_UID=${DEV_UID} DEV_GID=${DEV_GID} docker-compose
+NODE_RUN = $(DOCKER_COMPOSE) run --user "$(DEV_UID):$(DEV_GID)" --rm -e YARN_REGISTRY -e PUPPETEER_SKIP_CHROMIUM_DOWNLOAD=1 -e PUPPETEER_EXECUTABLE_PATH=/usr/bin/google-chrome node
 YARN_RUN = $(NODE_RUN) yarn
-PHP_RUN = $(DOCKER_COMPOSE) run --rm php php
-PHP_EXEC = $(DOCKER_COMPOSE) exec -u www-data fpm php
+PHP_RUN = $(DOCKER_COMPOSE) run --user "$(DEV_UID):$(DEV_GID)" --rm php php
+PHP_EXEC = $(DOCKER_COMPOSE) exec --user "$(DEV_UID):$(DEV_GID)" -u www-data fpm php
```
damien-carcel, line 7:
```suggestion
PHP_EXEC = $(DOCKER_COMPOSE) exec --user "$(DEV_UID):$(DEV_GID)" fpm php
```

damien-carcel: --user and -u are the same

docker-compose.yml
```diff
@@ -35,20 +34,20 @@ services:
       BLACKFIRE_CLIENT_TOKEN: '${BLACKFIRE_CLIENT_TOKEN:-client_token}'
     volumes:
       - './:/srv/pim'
+      - './var/logs:/var/log'
+      - './var:/var/run/php'
     working_dir: '/srv/pim'
-    command: 'php-fpm -F'
+    command: 'php-fpm -F -g /tmp/php8.1-fpm.pid'
+    user: ${DEV_UID}:${DEV_GID}
```
damien-carcel, line 41: To remove. PHP FPM should always been run as root. The main process only will be root, then it will spawn the workers with the `www-data` user.
Which is why we chose on Shared Catalogs to set `www-data` as we like during the build of the Docker image.

PR #18882, opened 2023-01-12
 PIM-10793: pim_api_access_token and pim_api_refresh_token tabels are never purged

RiffVentura:
`pim_api_access_token` and `pim_api_refresh_token` tables are never purged which leads to a considerable number of expired tokens that are never deleted (approx. 230M  for the worst-case scenarios). Deleting such a huge number of rows leads MySQL to generate a huge binary log. This log could fill up the disk space, preventing MySQL to finish the deletion, and keeping the `pim_api_access_token` table locked
The library that manages these tokens for both of these tables provides us with a command (`fos:oauth-server:clean`) to [delete all expired tokens](https://github.com/FriendsOfSymfony/FOSOAuthServerBundle/blob/master/Command/CleanCommand.php#L66)
However, its [implementation](https://github.com/FriendsOfSymfony/FOSOAuthServerBundle/blob/master/Entity/TokenManager.php#L86) is too simple as it attempts to delete all tokens all at once.

In this PR, we created a command to delete 50 million rows in 500 batches of 100,000. In the second [PR EE](https://github.com/akeneo/pim-enterprise-dev/pull/16181), the command is executed every day at 0:42. The number of lines will decrease day after day until it reaches equilibrium.

I've experimented with different batches sizes:
Given a table with 5 million entries, the command  deletes 1M rows
Note: average time per batch is based on the average of the first 10 batch times. Each batch test is executed 3 times to iron out anomalies
Note 2: Tests were made on a CircleCI deployed environment
|Batch size| Average time per batch| Average time to delete 1M rows|
|--------------|-------------------------------|-------------------------------------------------|
|100 000 | 16 281 ms | 162 817 ms|
|50 000 | 4910 ms | 100 825 ms|
|25 000 | 3 428 ms | 134 777 ms|
|10 000 | 759 ms | 97 673 ms|
|5 000 | 687 ms | 102 170 ms|

No locking of the table was observed during the execution of a 100 000 batch and  API endpoints (authentication or product endpoints) remained functional. furthermore, it appears InnoDB implementation [does not lock the table in case of a delete query BUT locks the range scanned](https://dev.mysql.com/doc/refman/8.0/en/glossary.html#glos_next_key_lock)

[PR EE](https://github.com/akeneo/pim-enterprise-dev/pull/16181)


src/Akeneo/Tool/Bundle/ApiBundle/Doctrine/DBAL/DeleteExpiredAccessTokenQuery.php
```diff
@@ -0,0 +1,43 @@
...
+use Doctrine\DBAL\Connection;
+use Doctrine\DBAL\ParameterType;
+
+/**
+ * @copyright 2023 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+class DeleteExpiredAccessTokenQuery
+{
+    private const DEFAULT_BATCH_SIZE = 100_000;
```
ahocquard, line 16: Could you reduce it to 10k instead please?
- First, it will be less pressure on Mysql to flush.
- second thing, it's more efficient based on your benchmark, which is not a surprise.
- third thing, it will prevent to have too long gaps lock on the table

Thanks

src/Akeneo/Tool/Bundle/ApiBundle/Doctrine/DBAL/DeleteExpiredAccessTokenQuery.php
```diff
@@ -0,0 +1,43 @@
...
+        $statement = $this->connection->prepare(<<<SQL
+            DELETE FROM pim_api_access_token
+            WHERE expires_at < :now_timestamp
+            LIMIT :row_count;
+        SQL);
+
+        $statement->bindValue('row_count', $batchSize, ParameterType::INTEGER);
+        $statement->bindValue('now_timestamp', $nowTimestamp, ParameterType::INTEGER);
+
+        for ($i = 0; $i < self::NUMBER_OF_LOOP; $i++) {
```
ahocquard, line 36: what is the purpose of limiting it to 50m rows? Why do you not loop until no more rows to delete?

tseho: Because they ran a test on a duplicated customer environment, with 230M rows in each table and it was taking more than 30h to run.
Flushing 50M rows maximum per execution (per day), makes that we can guarantee that this task always run under 24h.
For customers having 200M+ tokens in the table, yes, it will be completely purged over multiple days during the first runs, not all at once.

ahocquard: In that case, the code should explain why this arbitrary choice is made in my opinion.

tseho: good point. cc @guillaumefer-akeneo

guillaumefer-akeneo: Is it enough to explain this in a comment ?
Or should the names of the 2 constants (`DEFAULT_BATCH_SIZE`, `NUMBER_OF_LOOP`) reflect the name and if so, I'm open to suggestions ?

src/Akeneo/Tool/Bundle/ApiBundle/tests/integration/Handler/DeleteExpiredTokensHandlerIntegration.php
```diff
@@ -0,0 +1,116 @@
...
+
+        $this->connection->insert('pim_api_refresh_token', [
+            'client' => $clientId,
+            'user' => $userId,
+            'token' => 'invalid_refresh_token',
+            'expires_at' => $expiredTimestamp,
+        ]);
+
+        $this->assertTrue($this->accessTokenExists('invalid_access_token'));
+        $this->assertTrue($this->refreshTokenExists('invalid_refresh_token'));
```
masacc, line 61: I would have named them `expired_*_token` as they are not invalid.

upgrades/test_schema/Version_8_0_20230124095459_add_scheduled_job_delete_expired_tokens_Integration.php
```diff
@@ -0,0 +1,64 @@
...
+
+    public function test_it_adds_an_instance_if_not_present()
+    {
+        $this->deleteJobInstance('delete_expired_tokens');
+        $this->assertFalse($this->jobInstanceExists('delete_expired_tokens'));
+        $this->reExecuteMigration(self::MIGRATION_LABEL);
+        $this->assertTrue($this->jobInstanceExists('delete_expired_tokens'));
+    }
+
+    public function test_it_does_not_adds_an_instance_if_present()
```
masacc, line 33:
```suggestion
    public function test_it_does_not_add_an_instance_if_present()
```

masacc (approved):

src/Akeneo/Tool/Bundle/ApiBundle/Doctrine/DBAL/DeleteExpiredAccessTokenQuery.php
ahocquard: Totally valid in a comment, as you explain the why, not what :+1:

upgrades/schema/Version_8_0_20230124095459_add_scheduled_job_delete_expired_tokens.php
```diff
@@ -0,0 +1,45 @@
...
+    {
+        $sql = <<<SQL
+            INSERT INTO akeneo_batch_job_instance 
+                (`code`, `label`, `job_name`, `status`, `connector`, `raw_parameters`, `type`)
+            VALUES
+            (
+                :code,
+                :label,
+             
+                :code,
```
hogren, line 26:
```suggestion
                :label,
                :code,
```

hogren (approved):

PR #18908, opened 2023-01-17
Run tests against an upgraded dump in CI

fchoquet:
Run tests against a dump upgraded with migrations instead of doctrine schema. The goal is to run tests against something closer to what we actually have in production.

WIP


fchoquet:
Closing as the POC is done.
See EE corresponding PR for more details: https://github.com/akeneo/pim-enterprise-dev/pull/16178

PR #18871, opened 2023-01-11
GRF-681 : Add warning before category deletion

jeremyBeaucousinAkeneo:
**Description (for Contributor and Core Developer)**

**Definition Of Done (for Core Developer only)**

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Changelog (maintenance bug fixes)
- [ ] Tech Doc

PR #18867, opened 2023-01-11
GRF-339: Apply category permissions on children after saving

BryanS29-akeneo:
**Description (for Contributor and Core Developer)**

- Add a new `CategoryEditedEvent` dispatched in the `UpdateCategoryController`
- Creation of a new sql query to retrieve category children ids

**Definition Of Done (for Core Developer only)**

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Changelog (maintenance bug fixes)
- [ ] Tech Doc

PR #18910, opened 2023-01-17
GRF-682 : Connect front to service to get child categories count

jeremyBeaucousinAkeneo:
Before enriched categories we use to have this modal :
image

The text was always displayed:
> All the categories in that tree will be deleted, and the products will be removed from all those categories.
We add a condition with three use cases :
- No associate category but with products : 
image
- With categories, we don't count product (generic message) : 
image
- With no category or product associated : 
image
 

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc


src/Akeneo/Category/back/Infrastructure/Controller/InternalApi/CategoryController.php
```diff
@@ -86,4 +88,13 @@ public function getAction(string $identifier): JsonResponse
 
         return new JsonResponse($normalizedCategory);
     }
+
+    public function listChildrens(int $id): JsonResponse
```
olivierpontier, line 92: See review [here](https://github.com/akeneo/pim-community-dev/pull/18871/files)

src/Akeneo/Category/front/src/feature/CategoriesApp.tsx
```diff
@@ -17,7 +17,9 @@ const CategoriesApp: FC<Props> = ({setCanLeavePage}) => {
     <Router basename="/enrich/product-category-tree">
       <Switch>
         <Route path="/:treeId/tree">
-          <CategoriesTreePage />
+          <EditCategoryProvider setCanLeavePage={setCanLeavePage}>
```
olivierpontier, line 20: Why do we use the context of EditCategory page with the pages CategoriesTree and CategoryIndex?
Is it a way to factorize the use of a context property over those pages? Maybe creating a dedicated one would help to separate the concerns

src/Akeneo/Category/front/src/feature/components/datagrids/DeleteCategoryModal.tsx
```diff
@@ -1,24 +1,54 @@
...
+          categoriesNumber: categories.length,
+        });
+      } else if (numberOfProducts) {
+        warning = translate('pim_enrich.entity.category.category_tree_deletion.warning_products', {
+          name: categoryLabel,
+        });
+      }
+      setWarning(warning);
+    })
+  }, [numberOfProducts, categoryLabel, translate, countCategoryChildren, categoryId]);
```
olivierpontier, line 44:
```suggestion
  }, [countCategoryChildren]);
```

src/Akeneo/Category/front/src/feature/components/datagrids/DeleteCategoryModal.tsx
```diff
@@ -1,24 +1,54 @@
...
+      if (categories && categories.length > 0) {
+        warning = translate('pim_enrich.entity.category.category_tree_deletion.warning_categories_number', {
+          categoriesNumber: categories.length,
+        });
+      } else if (numberOfProducts) {
+        warning = translate('pim_enrich.entity.category.category_tree_deletion.warning_products', {
+          name: categoryLabel,
+        });
+      }
+      setWarning(warning);
```
olivierpontier, line 42: As the function countCategoryChildren is asynchronous we should check if the component is still mounted before updating the state

src/Akeneo/Category/front/src/feature/CategoriesApp.tsx
jeremyBeaucousinAkeneo: Because the query client provider was embedded into it, I will isolate it elsewhere.

src/Akeneo/Category/front/src/feature/components/datagrids/DeleteCategoryModal.tsx
```diff
@@ -1,24 +1,58 @@
...
+  categoryId,
+}) => {
+  const isMounted = useIsMounted();
   const translate = useTranslate();
+  const featureFlags = useFeatureFlags();
+
+  const [warning, setWarning] = useState('');
+  const countCategoryChildren = useCountCategoryChildren();
+
+  useEffect(() => {
```
pchasle, line 31: Generaly `useEffect` is not recommended for this kind of state manipulation, see: 
- https://beta.reactjs.org/learn/you-might-not-need-an-effect
- or https://www.youtube.com/watch?v=HPoC-k7Rxwo

Not necessarely an issue here if it work well, and I don't have any quick solution to avoid it without going a little bit deeper in the code myself.

But something to keep in mind and/or discuss together about useEffect usage.

src/Akeneo/Category/back/Infrastructure/Controller/InternalApi/CategoryController.php
```diff
@@ -18,10 +20,11 @@
...
     public function __construct(
-        protected CategoryRepositoryInterface $repository,
-        protected CategoryExtension $twigExtension,
-        protected NormalizerInterface $normalizer,
-        protected CollectionFilterInterface $collectionFilter,
+        protected CategoryRepositoryInterface     $repository,
+        protected CategoryExtension               $twigExtension,
+        protected NormalizerInterface             $normalizer,
+        protected CollectionFilterInterface       $collectionFilter,
+        protected GetCategoryChildrenIds          $getCategoryChildrenIds,
```
amigaire, line 27:
```suggestion
        protected GetCategoryChildrenIdsInterface          $getCategoryChildrenIds,
```

src/Akeneo/Category/back/Infrastructure/Controller/InternalApi/CategoryController.php
```diff
@@ -2,6 +2,8 @@
 
 namespace Akeneo\Category\Infrastructure\Controller\InternalApi;
 
+use Akeneo\Category\Application\Query\GetCategoryChildrenIds;
```
amigaire, line 5:
```suggestion
```

src/Akeneo/Category/back/Infrastructure/Controller/InternalApi/CategoryController.php
```diff
@@ -86,4 +89,11 @@ public function getAction(string $identifier): JsonResponse
 
         return new JsonResponse($normalizedCategory);
     }
+
+    public function listChildrens(int $id): JsonResponse
+    {
+        $categoriesId = ($this->getCategoryChildrenIds)($id);
+
+        return new JsonResponse($categoriesId);
```
amigaire, line 97:
```suggestion
        $categoryIds = ($this->getCategoryChildrenIds)($id);

        return new JsonResponse($categoryIds);
```

src/Akeneo/Category/front/src/feature/hooks/useCountCategoryChildren.ts
```diff
@@ -0,0 +1,31 @@
...
+        id: categoryId,
+      });
+      return fetch(url).then(response => {
+        return response.json();
+      });
+    },
+    [router]
+  );
+
+  const mutation = useMutation<number[], any, number>(fetchCategories);
```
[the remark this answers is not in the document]
jeremyBeaucousinAkeneo: The useQuery did not behave has expected (it keeps values in memory), it does not recall the API when I change the category on the page. This is why I use mutation to make the call on demand and recall on each selection.

src/Akeneo/Category/back/Infrastructure/Symfony/Resources/config/controllers.yml
```diff
@@ -26,6 +26,7 @@ services:
             - '@pim_enrich.twig.category_extension'
             - '@pim_internal_api_serializer'
             - '@pim_catalog.filter.chained'
+            - '@Akeneo\Category\Application\Query\GetCategoryChildrenIds'
```
amigaire, line 29: I do not see the new introduce Class in this PR

src/Akeneo/Category/front/src/feature/hooks/useCountCategoryChildren.ts
```diff
@@ -0,0 +1,31 @@
...
+};
+
+export type CountCategoryChildren = (categoryId: number) => CountCategoryChildren;
+
+export const useCountCategoryChildren = () => {
+  const router = useRouter();
+
+  const fetchCategories = useCallback(
+    async (categoryId: number): Promise<number[]> => {
+      const url = router.generate('pim_enrich_category_rest_list_childrens', {
```
amigaire, line 18:
```suggestion
      const url = router.generate('pim_enrich_category_rest_list_children', {
```

src/Akeneo/Category/back/Infrastructure/Symfony/Resources/config/routing/internal_api/category.yml
```diff
@@ -12,3 +12,8 @@ pim_enrich_category_rest_get:
     path: /rest/{identifier}
     defaults: { _controller: pim_enrich.controller.rest.category:getAction }
     methods: [GET]
+
+pim_enrich_category_rest_list_childrens:
+    path: /rest/{id}/childrens
+    defaults: { _controller: pim_enrich.controller.rest.category:listChildrens }
```
amigaire, line 18:
```suggestion
pim_enrich_category_rest_list_children:
    path: /rest/{id}/children
    defaults: { _controller: pim_enrich.controller.rest.category:listChildren }
```

src/Akeneo/Category/front/src/feature/components/datagrids/CategoryTreeDataGrid.tsx
```diff
@@ -250,7 +250,13 @@ const CategoryTreesDataGrid: FC<Props> = ({trees, refreshCategoryTrees}) => {
               categoryLabel={categoryTreeToDelete.label}
               closeModal={closeConfirmationModal}
               deleteCategory={deleteCategoryTree}
-              message={'pim_enrich.entity.category.category_tree_deletion.confirmation'}
+              message={
+                featureFlags.isEnabled('enriched_category')
+                  ? 'pim_enrich.entity.category.category_tree_deletion.confirmation_question'
+                  : 'pim_enrich.entity.category.category_deletion.confirmation'
```
amigaire, line 256:
```suggestion
                  : 'pim_enrich.entity.category.category_tree_deletion.confirmation'
```

Keep code as it was without enriched categories

src/Akeneo/Category/front/src/feature/hooks/useCountCategoryChildren.ts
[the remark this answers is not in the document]
pchasle: Yes, it's the normal behavior for `useQuery`. You will want to parameterized the query key with the categoryId, see https://tanstack.com/query/latest/docs/react/guides/query-keys#if-your-query-function-depends-on-a-variable-include-it-in-your-query-key
You will benefit from the cache and it will still be invalidated between different categories.

Something like that:
```ts
export const useCountCategoryChildren = (categoryId: number) => {
  const router = useRouter();

  // using memo to not recompute the router.generate, not necessary for the whole callback
  const url = useMemo(
    () => router.generate('pim_enrich_category_rest_list_childrens', { id: categoryId }),
    [categoryId]
  );

  // can count directly instead here and avoid the logic in the component code
  const countCategoryChildren = async (categoryId: number): Promise<number> => fetch(url).then(response => response.json()).length;

  return useQuery(['countCategoryChildren', categoryId], () => countCategoryChildren(categoryId));
};
```

Then on `DeleteCategoryModal.tsx` you can use something like that and not have any useEffect:
```tsx
  const {data: categoryChildrenCount, isLoading} = useCountCategoryChildren(categoryId);

  // it's totally fine to assign the warning value proceduraly like that for each rerender (but it can be wrapped in a useMemo for some optimisation at the drawback of making the code more complex to read)
  
  let warning = '';
  if (featureFlags.isEnabled('enriched_category') && !isLoading) {
    if (categoryChildrenCount > 0) {
      warning = translate('pim_enrich.entity.category.category_tree_deletion.warning_categories_number', {
        categoriesNumber: categories.length,
      });
    } else if (numberOfProducts) {
      warning = translate('pim_enrich.entity.category.category_tree_deletion.warning_products', {
        name: categoryLabel,
      });
    }
  }
```

Just wrote the code here and didn't try it, so take it as an example.

src/Akeneo/Category/front/src/feature/components/datagrids/DeleteCategoryModal.tsx
```diff
@@ -1,24 +1,58 @@
...
+            categoriesNumber: categories.length,
+          });
+        } else if (numberOfProducts) {
+          warning = translate('pim_enrich.entity.category.category_tree_deletion.warning_products', {
+            name: categoryLabel,
+          });
+        }
+        setWarning(warning);
+      });
+    }
```
amigaire, line 47: This seems odd, if you have counted category's children and there are at least 1 child, you display the number of children impacted.
Else, you display the number of products impacted, if there is.
If there are no children and no products, you have an empty modal's text.
An empty text doesn't seem to be the original behavior (before enriched categories).

src/Akeneo/Category/front/src/feature/hooks/useCountCategoryChildren.ts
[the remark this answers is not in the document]
jeremyBeaucousinAkeneo: Thanks for your explanation, I saw what I missed when I tried, I am learning about using useQuery.

Thanks a lot.

src/Akeneo/Category/back/Infrastructure/Controller/InternalApi/CategoryController.php
jeremyBeaucousinAkeneo: The interface GetCategoryChildrenIds is the interface of the class GetCategoryChildrenIdsSql.
Do you suggest that I rename it to be more clear?

src/Akeneo/Category/back/Infrastructure/Controller/InternalApi/CategoryController.php
jeremyBeaucousinAkeneo: I have to delete this line which import a class which does not exists :
use Akeneo\Category\Application\Query\GetCategoryChildrenIdsInterface;

src/Akeneo/Category/back/Infrastructure/Symfony/Resources/config/controllers.yml
jeremyBeaucousinAkeneo: This class was created in the [GRF-339](https://akeneo.atlassian.net/browse/GRF-339). This is the associate PR where we can see the file https://github.com/akeneo/pim-community-dev/pull/18867/files

src/Akeneo/Category/front/src/feature/components/datagrids/DeleteCategoryModal.tsx
jeremyBeaucousinAkeneo: Before enriched categories we use to have this modal :
image

The text was always displayed:
> All the categories in that tree will be deleted, and the products will be removed from all those categories.
We add a condition with three use cases :
- No associate category but with products : 
image
- With categories, we don't count product (generic message) : 
image
- With no category or product associated : 
image
 
I hope it's more clear for you.

amigaire: If it is confirmed, it is ok for me. I think I misunderstood the difference between the text and the warning content (red rectangle with text)

src/Akeneo/Category/back/Infrastructure/Controller/InternalApi/CategoryController.php
amigaire: I would create an interface then and use it in the construct method instead of directly use the class

src/Akeneo/Category/front/src/feature/components/datagrids/DeleteCategoryModal.tsx
amigaire: I would like to hear more about this case with you @pchasle , as front and React are not my strongest skill

src/Akeneo/Category/back/Infrastructure/Controller/InternalApi/CategoryController.php
amigaire: No it's okay, keep it that way

pchasle (approved):

src/Akeneo/Category/front/src/feature/hooks/useCountCategoryChildren.ts
```diff
@@ -0,0 +1,20 @@
+import {useMemo} from 'react';
+import {useRouter} from '@akeneo-pim-community/shared';
+import {useQuery} from 'react-query';
+
+type Result = {
+  data: number | undefined;
+  isLoading: boolean;
+};
+
+export const useCountCategoryChildren = (categoryId: number): Result => {
```
antoinetrouve, line 10: Do we have a test on this new hook ?

jeremyBeaucousinAkeneo: Testing a Hook that does not transform data does not seam relevant to me (too mush time to create a mock juste for an API call).

I think It'd be better to test the rendering on the component. @pchasle what do you think ?

pchasle: I agree

PR #18914, opened 2023-01-17
CPM-895: Call backend for Add Condition Button

pierallard:


cemorin (approved):

components/identifier-generator/front/src/feature/hooks/useGetConditionItems.ts
```diff
@@ -0,0 +1,123 @@
...
+const useGetConditionItems: (
+  isOpen: boolean,
+  conditions: Conditions,
+  limit?: number
+) => {
+  conditionItems: ItemsGroup[];
+  handleNextPage: () => void;
+  searchValue: string;
+  setSearchValue: (searchValue: string) => void;
+} = (isOpen, conditions, limit = 20) => {
```
Beninho, line 49: Why do you set limit at 20 here if you check value after and set at default value ?

Beninho (approved):

PR #18929, opened 2023-01-19
CXP-1376: mapping - accept number as target with number source

hogren:
image

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/components/catalogs/back/src/Application/Mapping/ValueExtractor/Extractor/Number/NumberFromNumberAttributeValueExtractor.php
+++ b/components/catalogs/back/src/Application/Mapping/ValueExtractor/Extractor/Number/NumberFromNumberAttributeValueExtractor.php
@@ -22,8 +22,16 @@ final class NumberFromNumberAttributeValueExtractor implements NumberValueExtrac
         /** @var mixed $value */
         $value = $product['raw_values'][$code][$scope][$locale] ?? null;
 
-        // @todo check if value is "floatable" or "intable" (i.e if it's a string that can be parsed)
-        return \is_numeric($value) ? (float) $value : null;
+        if (\is_numeric($value)) {
+            $intValue = (int) $value;
+            if ($intValue == $value) {
+                return $intValue;
+            }
+
+            return (float) $value;
+        }
+
+        return null;
     }
 
     public function getSupportedSourceType(): string
--- a/components/catalogs/back/src/Infrastructure/Command/CatalogFixtureCommand.php
+++ b/components/catalogs/back/src/Infrastructure/Command/CatalogFixtureCommand.php
@@ -130,6 +130,11 @@ class CatalogFixtureCommand extends Command
                     'scope' => null,
                     'locale' => null,
                 ],
+                'zoom' => [
+                    'source' => null,
+                    'scope' => null,
+                    'locale' => null,
+                ],
             ];
 
             $this->setCatalogProductMapping($catalogWithMappingId, $productMapping);
@@ -207,6 +212,10 @@ class CatalogFixtureCommand extends Command
             },
             "is_released": {
               "type": "boolean"
+            },
+            "zoom": {
+              "type": "number",
+              "title": "Optical Zoom"
             }
           }
         }
--- a/components/catalogs/back/src/Infrastructure/Validation/ProductMapping/AttributeSource/AttributeNumberSource.php
+++ b/components/catalogs/back/src/Infrastructure/Validation/ProductMapping/AttributeSource/AttributeNumberSource.php
@@ -1,0 +1,53 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Akeneo\Catalogs\Infrastructure\Validation\ProductMapping\AttributeSource;
+
+use Akeneo\Catalogs\Infrastructure\Validation\ProductMapping\AttributeSourceContainsValidLocale;
+use Akeneo\Catalogs\Infrastructure\Validation\ProductMapping\AttributeSourceContainsValidScope;
+use Symfony\Component\Validator\Constraint;
+use Symfony\Component\Validator\Constraints as Assert;
+use Symfony\Component\Validator\Constraints\Compound;
+
+/**
+ * @copyright 2023 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ *
+ * @psalm-suppress PropertyNotSetInConstructor
+ */
+final class AttributeNumberSource extends Compound
+{
+    /**
+     * @param array<array-key, mixed> $options
+     *
+     * @return array<array-key, Constraint>
+     */
+    protected function getConstraints(array $options = []): array
+    {
+        return [
+            new Assert\Sequentially([
+                new Assert\Collection([
+                    'fields' => [
+                        'source' => [
+                            new Assert\Type('string'),
+                            new Assert\NotBlank(),
+                        ],
+                        'scope' => [
+                            new Assert\Type('string'),
+                            new Assert\NotBlank(allowNull: true),
+                        ],
+                        'locale' => [
+                            new Assert\Type('string'),
+                            new Assert\NotBlank(allowNull: true),
+                        ],
+                    ],
+                    'allowMissingFields' => false,
+                    'allowExtraFields' => false,
+                ]),
+                new AttributeSourceContainsValidScope(),
+                new AttributeSourceContainsValidLocale(),
+            ]),
+        ];
+    }
+}
--- a/components/catalogs/back/tests/Integration/Infrastructure/Validation/ProductMapping/AttributeSource/AttributeNumberSourceTest.php
+++ b/components/catalogs/back/tests/Integration/Infrastructure/Validation/ProductMapping/AttributeSource/AttributeNumberSourceTest.php
@@ -1,0 +1,162 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Akeneo\Catalogs\Test\Integration\Infrastructure\Validation\ProductMapping\AttributeSource;
+
+use Akeneo\Catalogs\Infrastructure\Validation\ProductMapping\AttributeSource\AttributeNumberSource;
+use Symfony\Component\Validator\Validator\ValidatorInterface;
+
+/**
+ * @copyright 2022 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ *
+ * @covers \Akeneo\Catalogs\Infrastructure\Validation\ProductMapping\AttributeSource\AttributeNumberSource
+ */
+class AttributeNumberSourceTest extends AbstractAttributeSourceTest
+{
+    private ?ValidatorInterface $validator;
+
+    protected function setUp(): void
+    {
+        parent::setUp();
+        $this->validator = self::getContainer()->get(ValidatorInterface::class);
+    }
+
+    /**
+     * @dataProvider validDataProvider
+     * @param array<array-key, array{source: string|null, scope:string|null, locale: string|null}> $source
+     */
+    public function testItReturnsNoViolation(array $attribute, array $source): void
+    {
+        $this->createAttribute($attribute);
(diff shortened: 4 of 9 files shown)


components/catalogs/back/src/Infrastructure/Service/ProductMapper.php
```diff
@@ -58,11 +61,20 @@ public function getMappedProduct(
...
         return $product['raw_values'][$attributeCode][$scope][$locale] ?? '';
     }
+
+    private function castValue(string $type, mixed $value): mixed
+    {
+        return match ($type) {
+            'string' => (string) $value,
+            default => $value,
+        };
+    }
```
hogren, line 78: It's a temp method, waiting for the registry solution.

masacc: I don't think the cast should be in the `attributeValueExtractorRegistry`. Here in a dedicated method is ok for me as long as it is a simple switch/match. If it implies more logic, we could create some service(s).

components/catalogs/back/src/Infrastructure/Validation/ProductMapping/AttributeSource/AttributeNumberSource.php
```diff
@@ -0,0 +1,49 @@
...
+                        'source' => [
+                            new Assert\Type('string'),
+                            new Assert\NotBlank(),
+                        ],
+                        'scope' => [
+                            new Assert\IsNull(),
+                        ],
+                        'locale' => [
+                            new Assert\IsNull(),
+                        ],
```
masacc, line 41: number attributes are scopable and localizable

components/catalogs/back/src/Infrastructure/Validation/ProductMapping/ProductMappingRespectsSchemaValidator.php
```diff
@@ -26,7 +26,7 @@
  */
 final class ProductMappingRespectsSchemaValidator extends ConstraintValidator
 {
-    public function __construct(
+        public function __construct(
```
masacc, line 29: nope

components/catalogs/back/tests/Integration/Infrastructure/Validation/ProductMapping/AttributeSource/AttributeNumberSourceTest.php
```diff
@@ -0,0 +1,77 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Akeneo\Catalogs\Test\Integration\Infrastructure\Validation\ProductMapping\AttributeSource;
+
```
masacc, line 6: rm blank line

components/catalogs/back/src/Application/Service/TargetTypeConverter.php
```diff
@@ -15,7 +15,11 @@ final class TargetTypeConverter
     private const PIM_ATTRIBUTE_TYPES = [
         'string' => [
             'pim_catalog_text',
+            'pim_catalog_number'
```
tseho, line 18:
```suggestion
            'pim_catalog_number',
```

components/catalogs/back/src/Application/Service/TargetTypeConverter.php
```diff
@@ -15,7 +15,11 @@ final class TargetTypeConverter
...
         ],
+        'number' => [
+            'pim_catalog_number'
+        ]
```
tseho, line 22:
```suggestion
        'number' => [
            'pim_catalog_number',
        ],
```

We need to enforce this on the CI

components/catalogs/back/tests/Integration/Infrastructure/Validation/ProductMapping/AttributeSource/AttributeNumberSourceTest.php
hogren: @masacc all other files around have the same first lines. Are you sure ? If yes, which blank line is the one that you want that I delete (l4 or l6) ?

components/catalogs/back/src/Infrastructure/Validation/ProductMapping/AttributeSource/AttributeNumberSource.php
hogren: @masacc I changed that and I adapted tests.

components/catalogs/back/tests/Integration/Infrastructure/Validation/ProductMapping/AttributeSource/AttributeNumberSourceTest.php
masacc: that's ok, nothing to remove. I read it too fast I guess

masacc (approved):

components/catalogs/back/src/Application/Service/TargetTypeConverter.php
```diff
@@ -15,9 +15,13 @@ final class TargetTypeConverter
     private const PIM_ATTRIBUTE_TYPES = [
         'string' => [
             'pim_catalog_text',
+            'pim_catalog_number',
             'pim_catalog_textarea',
             'pim_catalog_simpleselect',
         ],
+        'number' => [
+            'pim_catalog_number',
+        ],
```
masacc, line 24: this one is for accepting pim_catalog_number as source for number target (not related to your PR)

components/catalogs/back/src/Application/Service/TargetTypeConverter.php
```diff
@@ -15,9 +15,13 @@ final class TargetTypeConverter
     private const PIM_ATTRIBUTE_TYPES = [
         'string' => [
             'pim_catalog_text',
+            'pim_catalog_number',
```
masacc, line 18: there is a unit test on this class to update if you add a line here

components/catalogs/back/src/Infrastructure/Validation/ProductMapping/AttributeSource/AttributeNumberSource.php
```diff
@@ -0,0 +1,63 @@
...
+                            ]),
+                        ],
+                        'locale' => [
+                            new Assert\AtLeastOneOf([
+                                new Assert\IsNull(),
+                                new Assert\Sequentially([
+                                    new Assert\Type('string'),
+                                    new Assert\NotBlank(),
+                                ]),
+                            ]),
```
masacc, line 52:
```suggestion
                            new Assert\Type('string'),
                            new Assert\NotBlank(allowNull: true),
```

components/catalogs/back/src/Infrastructure/Validation/ProductMapping/AttributeSource/AttributeNumberSource.php
```diff
@@ -0,0 +1,63 @@
...
+                            new Assert\NotBlank(),
+                        ],
+                        'scope' => [
+                            new Assert\AtLeastOneOf([
+                                new Assert\IsNull(),
+                                new Assert\Sequentially([
+                                    new Assert\Type('string'),
+                                    new Assert\NotBlank(),
+                                ]),
+                            ]),
```
masacc, line 43:
```suggestion
                            new Assert\Type('string'),
                            new Assert\NotBlank(allowNull: true),
```

components/catalogs/back/src/Infrastructure/Command/CatalogFixtureCommand.php
```diff
@@ -120,6 +120,11 @@ protected function execute(InputInterface $input, OutputInterface $output): int
                     'scope' => null,
                     'locale' => 'en_US',
                 ],
+                'zoom' => [
+                    'source' => 'optical_zoom',
+                    'scope' => null,
+                    'locale' => null,
+                ],
```
masacc, line 127: not necessary, there are already some string targets where we can map pim_catalog_number attributes.

However, you must add this use case in the acceptance tests for mapping

components/catalogs/back/src/Application/Service/TargetTypeConverter.php
masacc: Ok, I checked the backlog and there is no ticket for "pim_catalog_number -> string", only this one in which PR title is different from the ticket title which is different from the ticket description. Perhaps you can split into 2 tickets the 2 use cases like we did for others :
- pim_catalog_number -> string
- pim_catalog_number -> number

tseho: If it's not on `Available attribute types for each target type` notion page, we shouldn't support it.

masacc: Both are in it but "pim_catalog_number -> string" is not in MVP. You know what to do @hogren ;-)

masacc:
You forgot a few things in this PR :

1. in the extractor, you must transform the raw value according to this :

- raw value is int -> return int
- raw value is float -> return float
- raw value is_string and is_numeric -> if it matches an int, cast to int, else cast to float and return

2. add a target with a type number in the fixtures

masacc (approved):

components/catalogs/back/tests/Integration/Infrastructure/Validation/ProductMapping/AttributeSource/AttributeNumberSourceTest.php
```diff
@@ -0,0 +1,162 @@
...
+                    'scope' => 'ecommerce',
+                    'locale' => 'en_US',
+                ],
+            ],
+        ];
+    }
+    public function invalidDataProvider(): array
+    {
+        return [
+            'with a blank scope' => [
```
RiffVentura, line 115: There are some cases missing:
not a string locale, channel
extra fields,
missing fields

components/catalogs/back/src/Application/Mapping/ValueExtractor/Extractor/Number/NumberFromNumberAttributeValueExtractor.php
```diff
@@ -22,8 +22,16 @@ public function extract(
         /** @var mixed $value */
         $value = $product['raw_values'][$code][$scope][$locale] ?? null;
 
-        // @todo check if value is "floatable" or "intable" (i.e if it's a string that can be parsed)
-        return \is_numeric($value) ? (float) $value : null;
+        if (\is_numeric($value)) {
```
RiffVentura, line 25: feel free to ignore but I found this way in my price collection PR
```PHP
        if (!\is_scalar($value) || !\is_numeric($value)) {
            return null;
        }

        if (\is_float($value) || \is_int($value)) {
            return $value;
        }

        if ($value === (string) (int) $value) {
            return (int) $value;
        }

        if ($value === (string) (float) $value) {
            return (float) $value;
        }

        return null;
```

RiffVentura (approved):

components/catalogs/back/src/Application/Mapping/ValueExtractor/Extractor/Number/NumberFromNumberAttributeValueExtractor.php
masacc: I'm not sure we should try to return a value if it's a scalar. In my case, my thought was: we should return a number so :
- if the value is a number (int or float), no problem
- if it's a string that represents a number, ok then maybe a wrong cast, let's do the job and re-cast it in the right type
- if it's something else, uh-oh I don't think the value makes sense, we should not return it

Regarding the "how" you check it: I don't like the "double cast", but I don't like my `==` too. So I don't know which one is the best. Maybe you could tell me why you choose this way?

RiffVentura: phpstan doesn't like mixed type so the scalar check is mainly to narrow value type

the double conversion was mainly to avoid `==` .
I've realized now my implementation is flawed when it comes to detecting these values `56.00` and '56.30`

masacc: OK I think I will stay as it is

PR #18946, opened 2023-01-23
CXP-1408: Adds feature flag to Catalogs

tseho:

New feature flag `catalogs` available only in SaaS by default.

This PR should be reviewed commit by commit.

I've changed the format of `akeneo_feature_flag.yml`, using the feature as the array key.
This way, we can override it in different Symfony environments.

Example:
By default, `catalogs` is enabled only in SaaS (Serenity, Trial, Growth).
We can override this configuration in `config/packages/dev/` to enable it by default when working locally.

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/.circleci/jobs/install_front_dependencies.yml
+++ b/.circleci/jobs/install_front_dependencies.yml
@@ -26,7 +26,7 @@ jobs:
                                 name: Build front shared libraries
                                 command: |
                                     [ -d ~/project/front-packages/akeneo-design-system/lib ] || docker-compose run -u node --rm node yarn dsm:build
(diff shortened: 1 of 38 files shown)


RiffVentura:
I'm wondering if flags should be applied to ServiceApi  handlers as well :thinking:

masacc (approved):

masacc:
Is there a need to declare the feature flag as a variable in root `.env` like we did before ?

RiffVentura (approved):

masacc (approved):

RiffVentura (approved):

PR #18956, opened 2023-01-25
CPM-900 + CPM-919 + CPM-903 + CPM-872 + CPM-906 = CPM-4500

pierallard:
Please read it commit by commit, the name of the commit is the matching card.

### DIFF
--- a/components/identifier-generator/front/src/feature/tabs/structure/preview/AutoNumberPreview.tsx
+++ b/components/identifier-generator/front/src/feature/tabs/structure/preview/AutoNumberPreview.tsx
@@ -9,12 +9,9 @@ type AutoNumberPreviewProps = {
 const AutoNumberPreview: React.FC<AutoNumberPreviewProps> = ({property}) => {
   const {digitsMin, numberMin} = property;
 
-  const getFormattedNumber = useMemo(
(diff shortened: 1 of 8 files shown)


components/identifier-generator/front/src/feature/tabs/structure/preview/AutoNumberPreview.tsx
```diff
@@ -9,12 +9,9 @@ type AutoNumberPreviewProps = {
 const AutoNumberPreview: React.FC<AutoNumberPreviewProps> = ({property}) => {
   const {digitsMin, numberMin} = property;
 
-  const getFormattedNumber = useMemo(
-    () => (i: number) => String(Math.max(i, numberMin)).padStart(digitsMin, '0'),
-    [digitsMin, numberMin]
-  );
+  const getFormattedNumber = useMemo(() => (i: number) => String(i).padStart(digitsMin, '0'), [digitsMin]);
```
cemorin, line 12: Since all used params are available, it doesn't have to be a function.

```suggestion
  const formattedNumber = useMemo(() => String(numberMin).padStart(digitsMin, '0'), [digitsMin, numberMin]);
```

And below it would be
```
  return <Preview.Highlight>{formattedNumber}</Preview.Highlight>;
```

cemorin (approved):

mmetayer (approved):

PR #18970, opened 2023-01-26
CPM-912: Update delimiter to options migration

pierallard:
Add migration from delimiter to JSON options


mmetayer (approved):
Since the target column migration is not merged in master yet, we could also have used it to simplify the "isPreviousDatabaseVersion" functions (only one migration, and only 2 possible schema states)

pierallard:
@mmetayer Totally agree with you.
As this means update this branch then the branch after, may I create a refactor card to do it in 1 another PR ?

mmetayer:
of course!

cemorin (approved):

PR #18981, opened 2023-01-27
CPM-914: Upper and lower cases

pierallard:


components/identifier-generator/back/src/Application/Generate/GenerateIdentifierHandler.php
```diff
@@ -35,14 +36,33 @@ public function __invoke(
         GenerateIdentifierCommand $command,
     ): string {
         $identifierGenerator = $command->getIdentifierGenerator();
+        $transformedDelimiter = $identifierGenerator->delimiter()?->asString() ?? '';
+        switch ($identifierGenerator->textTransformation()->normalize()) {
+            case TextTransformation::UPPERCASE: $transformedDelimiter = \strtoupper($transformedDelimiter);
```
mmetayer, line 41: rather `\mb_strtoupper`? (to check with the db queries)

components/identifier-generator/back/tests/Specification/Application/Generate/GenerateIdentifierHandlerSpec.php
```diff
@@ -36,57 +36,75 @@ public function let(
...
+
+        $getNextIdentifierQuery
+            ->fromPrefix($identifierGenerator, 'aKn-a', 0)
+            ->shouldBeCalled()
+            ->willReturn(43);
+
+        $this->__invoke($generateIdentifierCommand)->shouldReturn('aKn-a43');
+    }
+
+    public function it_should_generate_an_identifier_with_delimiter_and_uppercase(
```
mmetayer, line 70: we also might wanna add acceptance/e2e tests

mmetayer (approved):

components/identifier-generator/back/tests/Specification/Application/Generate/GenerateIdentifierHandlerSpec.php
pierallard: done!

components/identifier-generator/back/src/Application/Generate/GenerateIdentifierHandler.php
```diff
@@ -35,14 +36,33 @@ public function __invoke(
         GenerateIdentifierCommand $command,
     ): string {
         $identifierGenerator = $command->getIdentifierGenerator();
+        $transformedDelimiter = $identifierGenerator->delimiter()?->asString() ?? '';
+        switch ($identifierGenerator->textTransformation()->normalize()) {
```
antoinearrive, line 40: why transform the delimiter then the properties, rather than transform the entire result before the return ?

antoinearrive (approved):

components/identifier-generator/back/src/Application/Generate/GenerateIdentifierHandler.php
pierallard: For the computing of the auto-number, we need to have the current prefix.
If you don't transform the delimiter, and the delimiter is "a", we need to tell the auto number generator that the prefix is "AKNA", not "AKNa".

Beninho (approved):

PR #18983, opened 2023-01-27
CXP-1402: mapping - accept date as source for string-datetime target

hogren:

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/components/catalogs/back/src/Infrastructure/Validation/ProductMapping/AttributeSource/AttributeDateSource.php
+++ b/components/catalogs/back/src/Infrastructure/Validation/ProductMapping/AttributeSource/AttributeDateSource.php
@@ -1,0 +1,53 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Akeneo\Catalogs\Infrastructure\Validation\ProductMapping\AttributeSource;
+
+use Akeneo\Catalogs\Infrastructure\Validation\ProductMapping\AttributeSourceContainsValidLocale;
+use Akeneo\Catalogs\Infrastructure\Validation\ProductMapping\AttributeSourceContainsValidScope;
+use Symfony\Component\Validator\Constraint;
+use Symfony\Component\Validator\Constraints as Assert;
+use Symfony\Component\Validator\Constraints\Compound;
+
+/**
+ * @copyright 2023 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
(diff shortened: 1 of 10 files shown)


guillaumefer-akeneo (approved):

masacc (commented):
You have to be absolutely sure that what you get from `raw_values` is a date in ISO8601 format in `StringDateTimeFromDateAttributeValueExtractor`. 

For now the only check that is done is that the value is a string :
```
return \is_string($value) ? $value : null;
```

components/catalogs/back/src/Infrastructure/Validation/ProductMapping/AttributeSource/AttributeDateSource.php
```diff
@@ -0,0 +1,63 @@
...
+                            new Assert\NotBlank(),
+                        ],
+                        'scope' => [
+                            new Assert\AtLeastOneOf([
+                                new Assert\IsNull(),
+                                new Assert\Sequentially([
+                                    new Assert\Type('string'),
+                                    new Assert\NotBlank(),
+                                ]),
+                            ]),
```
masacc, line 43:
```suggestion
                            new Assert\Type('string'),
                            new Assert\NotBlank(allowNull: true),
```

hogren: Thanks for the tip ! @masacc

components/catalogs/back/src/Application/Service/TargetTypeConverter.php
```diff
@@ -18,6 +18,9 @@ final class TargetTypeConverter
             'pim_catalog_textarea',
```
masacc, line 18: I moved this file in `Application/Mapping/TargetTypeConverter.php` (Sercice -> Mapping). If you have any conflict, be aware of that. Same for its associated TU.

components/catalogs/back/tests/Unit/Service/TargetTypeConverterTest.php
```diff
@@ -47,6 +47,13 @@ public function validConversionProvider(): array
                     'pim_catalog_simpleselect',
                 ],
             ],
+            'date' => [
```
masacc, line 50:
```suggestion
            'string+date-time' => [
```

masacc (approved):

masacc (approved):

PR #18987, opened 2023-01-27
CXP-1392: Constraint - string - format (regexp)

guillaumefer-akeneo:
image

- [x] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc


components/catalogs/back/src/Infrastructure/Symfony/Resources/meta-schemas/product-0.0.5.json
```diff
@@ -86,6 +86,10 @@
         "maxLength": {
           "type": "integer",
           "minimum": 0
+        },
+        "pattern": {
+          "type": "string",
+          "format": "regex"
```
masacc, line 92: as said this morning, should be in another meta schema

components/catalogs/back/src/Infrastructure/Validation/ProductMappingSchemaValidator.php
```diff
@@ -100,4 +106,21 @@ private function getMetaSchemaLocalPath(string $id): ?string
...
+    {
+        /** @var array{properties: array<string, array<string, string>>} $schema */
+        $schema = \json_decode(\json_encode($schema, JSON_THROW_ON_ERROR) ?: '{}', true, 512, JSON_THROW_ON_ERROR);
+
+        foreach ($schema['properties'] as $property) {
+            if (!isset($property['pattern'])) {
+                continue;
+            }
+
+            if (@\preg_match(\sprintf('/%s/', \preg_quote($property['pattern'], '/')), '') === false) {
```
masacc, line 120: are you sure it works all times?

for example does the "pattern" include the first and last characters? Can you have a problem by using `/` as the first and last characters (I used to use `~` because we don't find it in url for example) ?

Moreover I read this in the `preg_match` documentation :

> Erreurs / Exceptions
>
> Si le masque regex passé ne compile pas à une regex valide, une E_WARNING est émise.

I don't know if it can have some impacts here (I think it's ok with the `@`)

tseho: I assisted to the test-first approach of this part, by Guillaume & Hélène.
They tested it correctly, I'm confident in the code presented.

tseho: > Can you have a problem by using / as the first and last characters

That's where `preg_quote($pattern, '/')` is useful.

masacc: oh yes good catch

components/catalogs/back/src/Infrastructure/Symfony/Resources/meta-schemas/product-0.0.7.json
```diff
@@ -0,0 +1,104 @@
...
+          "minimum": 0
+        },
+        "maxLength": {
+          "type": "integer",
+          "minimum": 0
+        },
+        "pattern": {
+          "type": "string",
+          "format": "regex"
+        }
```
masacc, line 71: Can we use the `dependentSchemas` to improve that and say the `pattern` property can only be used if the `type` is `string` like we did for the `format` property?

masacc (approved):

masacc:
Don't forget to do the PR in pim-api-docs to add this new version of the schema

components/catalogs/back/src/Infrastructure/Symfony/Resources/meta-schemas/product-0.0.7.json
eoliveauakeneo: Done

components/catalogs/back/tests/Integration/Infrastructure/Validation/ProductSchema/valid/0.0.6.json
```diff
@@ -31,6 +31,11 @@
         "thumbnail": {
             "type": "string",
             "format": "uri"
+        },
+        "meta_description": {
+            "type": "string",
+            "minLength": 50,
+            "maxLength": 160
         }
```
hogren, line 39: Why it is in this PR ?

components/catalogs/front/src/components/ProductMapping/ProductMapping.integration.tsx
```diff
@@ -18,6 +18,9 @@ const PRODUCT_MAPPING_SCHEMA = {
             type: 'string',
             title: 'Product name',
             description: 'This is the name of the product shown on the ecommerce',
+            minLength: 3,
+            maxLength: 50,
+            pattern: '[a-zA-Z].',
```
hogren, line 23: I think it not very coherent use case, even if it works.
cc @masacc

hogren (approved):

components/catalogs/back/tests/Integration/Infrastructure/Validation/ProductSchema/valid/0.0.6.json
masacc: because I've done (and merged) the 0.0.6 before the 0.0.5 was merged and I forgot this in my valid version of the schema in test

components/catalogs/front/src/components/ProductMapping/ProductMapping.integration.tsx
masacc: Yes, but the question is more : does it add some value to the test?

components/catalogs/back/src/Infrastructure/Validation/ProductMappingSchemaValidator.php
```diff
@@ -73,6 +73,12 @@ public function validate($value, Constraint $constraint): void
 
             $this->logger->debug('A Product Mapping Schema validation failed', $errors);
         }
+
+        if ($this->containsInvalidRegexes($value)) {
+            $this->context
+                ->buildViolation('You must provide a schema with valid regexes.')
```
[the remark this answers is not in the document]
tseho: agree but value vs cost here.

RiffVentura (approved):

components/catalogs/front/src/components/ProductMapping/ProductMapping.integration.tsx
eoliveauakeneo: So do I change the test with a more complex pattern or leave it like that ?

PR #19015, opened 2023-02-01
CPM-940: Create Condition MultiSelect

Beninho:
**Description (for Contributor and Core Developer)**

New MultiSelect, copy of simpleSelect for the moment

**Definition Of Done (for Core Developer only)**

- [X] Tests

PR #19017, opened 2023-02-01
CPM-938: update validator to be compatible with multi_select type

Beninho:

Be able to create/update a selection by MultiSelectOption

- Update validator
- Update AttributeShouldHaveType to make it compatible with a list of types
- add tests

- [X] Tests


components/identifier-generator/back/src/Infrastructure/Validation/SimpleOrMultiSelectShouldBeValid.php
```diff
@@ -10,7 +10,7 @@
  * @copyright 2022 Akeneo SAS (https://www.akeneo.com)
  * @license   https://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
  */
-final class SimpleSelectShouldBeValid extends Constraint
+final class SimpleOrMultiSelectShouldBeValid extends Constraint
```
mmetayer, line 13: As in #19015 I'd really prefer if you created a new Validator, even if it means copy-pasting the SimpleSelectShouldBeValidValidator for now (e.g: what happens if we add a new operator only for multi select conditions in the future?)

mmetayer (approved):
seen IRL with @pierallard: personally I would still prefer having separate validators, but this implementation is OK

PR #19025, opened 2023-02-01
Edit link not visible with Wysiwyg field

gauquier:
Fixed edit link not visible with Wysiwyg field

This pull request addresses the issue of the edit link not being visible when using the Wysiwyg field. The solution involves updating the CSS code to properly display the edit link when using the Wysiwyg field, ensuring that it is clearly visible to users. 

image003

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc


anthonyDevSf (approved):

nocturnalfrog (approved):

Username070:
Just to back these changes, I would like to add a reference to this MDN: https://developer.mozilla.org/en-US/docs/Web/API/Event/originalTarget.

It states that `originalTarget`
> "Non-standard: This feature is non-standard and is not on a standards track. Do not use it on production sites facing the Web: it will not work for every user. There may also be large incompatibilities between implementations and the behavior may change in the future."

PR #19074, opened 2023-02-07
CPM 929: Set delimiter not nullable

Beninho:

Set delimiter not nullable

- [X] Tests


components/identifier-generator/back/src/Application/Generate/GenerateIdentifierHandler.php
```diff
@@ -36,29 +36,29 @@ public function __invoke(
         GenerateIdentifierCommand $command,
     ): string {
         $identifierGenerator = $command->getIdentifierGenerator();
-        $transformedDelimiter = $identifierGenerator->delimiter()?->asString() ?? '';
+        $transformedDelimiter = $identifierGenerator->delimiter()->asString();
```
mmetayer, line 39: careful, currently `Delimiter::asString()` can return null, you'll get an error when calling mb_strtoxxx functions

mmetayer (approved):

pierallard (approved):

PR #19079, opened 2023-02-07
GRF-724: Fix attribute value to be clearable & remove persistence of null value

pchasle:
- [x] Fix attribute values to be clearable while enriching categories
- [x] Do not persist empty attribute values in the `pim_catalog_category` table
  - Fix REST API to not return empty attribute values

### DIFF
--- a/src/Akeneo/Category/back/Domain/UserIntent/Factory/ValueUserIntentFactory.php
+++ b/src/Akeneo/Category/back/Domain/UserIntent/Factory/ValueUserIntentFactory.php
@@ -38,7 +38,7 @@ final class ValueUserIntentFactory implements UserIntentFactory
     }
 
     /**
-     * @param array<string, AttributeValueApi> $data
+     * @phpstan-param array<string, AttributeValueApi> $data
      *
      * @return array|UserIntent[]
      */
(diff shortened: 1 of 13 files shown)


src/Akeneo/Category/back/Domain/UserIntent/Factory/ValueUserIntentFactory.php
```diff
@@ -60,7 +60,7 @@ public function create(string $fieldName, int $categoryId, mixed $data): array
                 }
 
                 if (is_string($value['data']) && $this->isDataStringEmpty($value['data'])) {
-                    continue;
+                    $value['data'] = null;
```
pchasle, line 63: We don't skip empty value to keep the intent to "clear" the value.
Frontend can push `null` directly but check for empty string and override it with `null`

src/Akeneo/Category/back/Domain/UserIntent/Factory/ValueUserIntentFactory.php
```diff
@@ -38,7 +38,7 @@ public function getSupportedFieldNames(): array
     }
 
     /**
-     * @param array<string, AttributeValueApi> $data
+     * @phpstan-param array<string, AttributeValueApi> $data
```
pchasle, line 41: I'm using the phpstan prefix since it reference a phpstan "virtual" type `AttributeValueApi`. PHPStorm doesn't mind but some IDE do & report an error.
Static analysis work the same for phpstan & phpstorm.

BryanS29-akeneo (approved):

PR #19135, opened 2023-02-10
PH-799: DatabaseCommand should not write on output

jmleroux:
We should use the logger instead of direct console output.

Console output is always detected as `error` on Datadog.

Advantages for using the logger:
- more correct log levels
- real logs with all the context information
- possibility to not ingest if not needed
- etc.

Cons:
- not displayed by default in the dev console, but it can be easily displayed by using a higher verbosity with the `-vv` parameter
 
In brief, more log control (and in the end maybe a lower billing :wink: )


src/Akeneo/Platform/Bundle/InstallerBundle/Command/DatabaseCommand.php
```diff
@@ -224,14 +228,12 @@ protected function loadFixturesStep(InputInterface $input, OutputInterface $outp
                 ]),
                 InstallerEvents::PRE_LOAD_FIXTURE
             );
-            if ($input->getOption('verbose')) {
```
jmleroux, line 227: No need to test verbosity with the logger, it's part of the logger logic.

Beninho (approved):

StevenVAIDIE (approved):

PR #19141, opened 2023-02-10
CXP-1390: avoid 'any' in typescript

tseho:

First, really useful link when you are struggling with weird typescript stuff: https://www.typescriptlang.org/docs/handbook/utility-types.html

Then, instead of going into `[key: string]: any;` territory...
I'm proposing another version, which is more strict and will prevent us to have errors due to `any`.
I've commented the code in the PR because it's not obvious at first sight.

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/components/catalogs/front/src/components/ProductMapping/components/RequirementsCollapse.tsx
+++ b/components/catalogs/front/src/components/ProductMapping/components/RequirementsCollapse.tsx
@@ -4,12 +4,17 @@ import {useTranslate} from '@akeneo-pim-community/shared';
 import {Target} from '../models/Target';
 import styled from 'styled-components';
 
+const ACCEPTED_CONSTRAINTS = ['minLength', 'maxLength'] as const;
+
 type Props = {
-    selectedTarget: Target;
+    target: Target;
 };
 
+type TargetWithOnlyConstraints = Pick<Target, typeof ACCEPTED_CONSTRAINTS[number]>;
+type TargetWith<K extends keyof TargetWithOnlyConstraints> = Required<Pick<TargetWithOnlyConstraints, K>>;
+
 type Constraint = {
-    key: string;
+    key: keyof TargetWithOnlyConstraints;
     value: string | number;
 };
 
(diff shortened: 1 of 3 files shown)


components/catalogs/front/src/components/ProductMapping/components/RequirementsCollapse.tsx
```diff
@@ -4,43 +4,39 @@ import {useTranslate} from '@akeneo-pim-community/shared';
 import {Target} from '../models/Target';
 import styled from 'styled-components';
 
+const ACCEPTED_CONSTRAINTS = ['minLength', 'maxLength'] as const;
```
tseho, line 7: `as const` => https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html#const-assertions
basically, it says to typescript keep it like it is, do not try to deduce that as `string[]`.

components/catalogs/front/src/components/ProductMapping/components/RequirementsCollapse.tsx
```diff
@@ -4,43 +4,39 @@ import {useTranslate} from '@akeneo-pim-community/shared';
...
+
 type Props = {
-    selectedTarget: Target;
+    target: Target;
 };
 
+type TargetWithOnlyConstraints = Pick<Target, typeof ACCEPTED_CONSTRAINTS[number]>;
```
tseho, line 13: `Pick` => https://www.typescriptlang.org/docs/handbook/utility-types.html#picktype-keys
It create a new type from `Target` while keeping only the keys of `ACCEPTED_CONSTRAINTS`.
Basically,

```
{
    code: string;
    label: string;
    type: string;
    format: string | null;
    description?: string;
    minLength?: number;
    maxLength?: number;
};
```
becomes
```
{
    minLength?: number;
    maxLength?: number;
};
```

`typeof ACCEPTED_CONSTRAINTS[number]` is a weird typescript trick, it converts `['minLength', 'maxLength']` to `'minLength' | 'maxLength'`.

components/catalogs/front/src/components/ProductMapping/components/RequirementsCollapse.tsx
```diff
@@ -4,43 +4,39 @@ import {useTranslate} from '@akeneo-pim-community/shared';
...
+type TargetWith<K extends keyof TargetWithOnlyConstraints> = Required<Pick<TargetWithOnlyConstraints, K>>;
```
tseho, line 14: When doing, `TargetWith<'minLength'>`,
```
{
    minLength?: number;
    maxLength?: number;
};
```
becomes
```
{
    minLength: number;
};
```
You can notice that the property is not undefined-able anymore.
Typescript is happy.

components/catalogs/front/src/components/ProductMapping/components/RequirementsCollapse.tsx
```diff
@@ -4,43 +4,39 @@ import {useTranslate} from '@akeneo-pim-community/shared';
...
+
 type Constraint = {
-    key: string;
+    key: keyof TargetWithOnlyConstraints;
```
tseho, line 17: Easy improvement, instead of `string`, with `keyof TargetWithOnlyConstraints`, you are only allowed to create an instance of `Constraint` when the key is one of `ACCEPTED_CONSTRAINTS`.

eoliveauakeneo (approved):

helenegzlz (approved):

PR #19156, opened 2023-02-13
DatabaseCommand should not write on output

jmleroux:
We should use the logger instead of direct console output.

Console output is always detected as error on Datadog.

Advantages for using the logger:

- more correct log levels
- real logs with all the context information
- possibility to not ingest if not needed
- etc.

Cons:

- not displayed by default in the dev console, but it can be easily displayed by using a higher verbosity with the -vv parameter

In brief, more log control (and in the end maybe a lower billing :wink: )

### DIFF
--- a/src/Akeneo/Platform/Bundle/InstallerBundle/Command/DatabaseCommand.php
+++ b/src/Akeneo/Platform/Bundle/InstallerBundle/Command/DatabaseCommand.php
@@ -206,12 +206,7 @@ class DatabaseCommand extends Command
             $input->setOption('fixtures', self::LOAD_BASE);
         }
 
-        $this->logger->info(
-            sprintf(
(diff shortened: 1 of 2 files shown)


mmetayer (approved):

Beninho (approved):

src/Akeneo/Platform/Bundle/InstallerBundle/Command/DatabaseCommand.php
```diff
@@ -201,7 +206,7 @@ protected function loadFixturesStep(InputInterface $input, OutputInterface $outp
             $input->setOption('fixtures', self::LOAD_BASE);
         }
 
-        $output->writeln(
+        $this->logger->info(
```
nmarniesse, line 209: Remove <info> in the message?

nmarniesse (approved):

PR #19180, opened 2023-02-15
PIM-10831: Prevent association picker from sending too many requests

mmetayer:

***Previous behavior of the product and product model picker:***

There were several flaws with the current product picker in the association grid:
- each time a product/model is selected, a request was sent to the backend to get the product information (in order to render the items in the basket), BUT the request included ALL the basket items, not only the one you just selected)
- so, for instance, when selecting all the elements of a page (25), 25 requests would be sent to the backend:
    - get products?ids=my_first_product
    - get products?ids=my_first_product,my_second_product
    - get products?ids=my_first_product,my_second_product,my_third_product
    - etc...
- Moreover the event listeners were not unregistered, so each time you opened the product picker, the number of requests per basket item would be incremented (e.g, if you opened it, then closed and reopened it, 2 requests per basket item would be sent - in the previous example that would be 50 requests to select a page, then 75, then 100...)

=> This could quickly result in a failure of the backend server, having to deal with too many requests at once (plus having to retrieve, hydrate and normalize products/product models is not the cheapest operation in the PIM...)
associations

***The fix***

Actually, there is no need to send a request to get product or model information, as the datagrid already contains all the information we need to render the basket: image, label, document type (product or product model), id, ...
In this PR I try to use this piece of information in order to render the basket, then pass it to the main form (association form or quantified association form) so that it can handle it.

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/src/Akeneo/Platform/Bundle/UIBundle/Resources/public/js/product/form/product-and-product-model-picker.js
+++ b/src/Akeneo/Platform/Bundle/UIBundle/Resources/public/js/product/form/product-and-product-model-picker.js
@@ -10,88 +10,150 @@
  * @copyright 2017 Akeneo SAS (http://www.akeneo.com)
  * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
  */
-define([
-  'jquery',
-  'oro/translator',
-  'pim/common/item-picker',
-  'pim/fetcher-registry',
-  'pim/media-url-generator',
-], function ($, __, ItemPicker, FetcherRegistry, MediaUrlGenerator) {
+define(['underscore', 'pim/common/item-picker', 'pim/media-url-generator'], function (
+  _,
+  ItemPicker,
+  MediaUrlGenerator
+) {
   return ItemPicker.extend({
+    items: [],
+
+    initialize: function () {
+      this.items = [];
+
+      return ItemPicker.prototype.initialize.apply(this, arguments);
+    },
+
     /**
      * {@inheritdoc}
      */
     selectModel: function (model) {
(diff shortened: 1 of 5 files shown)


Beninho (approved):

cemorin (approved):

src/Akeneo/Platform/Bundle/UIBundle/Resources/public/js/product/form/product-and-product-model-picker.js
```diff
@@ -10,88 +10,150 @@
...
-    },
+    imagePathMethod: item => MediaUrlGenerator.getMediaShowUrl(item.image?.filePath, 'thumbnail_small'),
 
     /**
      * {@inheritdoc}
      */
-    labelMethod: function (item) {
-      return item.meta.label[this.getLocale()];
-    },
+    labelMethod: item => item.label,
```
StevenVAIDIE, line 152: Maybe there is an impact on the label displayed (I think before we display the label of the selected locale, now it will be displayed the label on the UI locale) => to revalidate

phaseinducer (approved):

StevenVAIDIE (approved):
Nice job ! I already try to refactor this code and I didn't success due to the itemCode ! :clap:

mmetayer:
@StevenVAIDIE I created a ticket in our backlog to take your suggestions into account: https://akeneo.atlassian.net/browse/CPM-977, but for now I'll merge as is (the ticket is urgent and was already validated support-side)

PR #19207, opened 2023-02-16
PIM-10844: Filter empty attribute option labels

mmetayer:

When updating an attribute option from the attribute edit form UI, empty labels were inserted anyway in the `pim_catalog_attribute_option_value` table (with value = NULL). This happened only in the UI, because it still uses a Symfony form, as opposed to import and API (which use the repo/factory/updater/saver stack)
This led to inconsistencies in the API, because such null labels were returned, unlike for other PIM entities

In this PR:
- I make sure null attribute option labels are not returned anymore
- and that null labels are not inserted anymore when using the UI

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/src/Akeneo/Pim/Structure/Bundle/Form/Type/AttributeOptionType.php
+++ b/src/Akeneo/Pim/Structure/Bundle/Form/Type/AttributeOptionType.php
@@ -2,6 +2,7 @@
 
 namespace Akeneo\Pim\Structure\Bundle\Form\Type;
 
+use Akeneo\Pim\Structure\Component\Model\AttributeOptionValueInterface;
 use Symfony\Component\Form\AbstractType;
 use Symfony\Component\Form\Extension\Core\Type\CollectionType;
 use Symfony\Component\Form\Extension\Core\Type\HiddenType;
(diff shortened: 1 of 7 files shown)


src/Akeneo/Pim/Structure/Bundle/Form/Type/AttributeOptionType.php
```diff
@@ -76,6 +77,8 @@ protected function addFieldOptionValues(FormBuilderInterface $builder)
                 'allow_add'    => true,
                 'allow_delete' => true,
                 'by_reference' => false,
+                'delete_empty' =>
+                    static fn (AttributeOptionValueInterface $optionValue = null) => null === $optionValue?->getValue()
```
mmetayer, line 81: the ideal solution would be to use the repo/updater/saver stack instead of a symfony form, but that would imply too much refacto for a maintenance (especially on the frontend side), so instead I remove empty labels from the collection

Beninho (approved):

cemorin (approved):

mmetayer:
green build: https://app.circleci.com/pipelines/github/akeneo/pim-community-dev/48552/workflows/8efef56a-4dec-4a37-b76a-cd047253f7b2

PR #19251, opened 2023-02-22
JEL-239: Create User System for tagged commands

LaurentPetard:
While benchmarking a simple Symfony command execution in the PIM, we detected that a significant number of undesirable services (as Doctrine ORM) were called and took a significant amount of time. At the origin of this is a legacy listener that automatically authenticate an admin system user to every Akeneo/PIM Symfony commands (command names that started by `pim` or `akeneo`)  https://github.com/akeneo/pim-community-dev/blob/master/src/Akeneo/UserManagement/Bundle/EventListener/CreateUserSystemListener.php

In our case we need to be able to run a command as fast as possible, so we don't want to trigger this listener. And in a more global case, this authentication is useless for most of the PIM commands. It's mandatory only for process that involve permissions.

So we decided to trigger this listener from a new dedicated tag. To take minimum risks, we added this tag to all commands, except for those that we are sure they don't need this authentication. Then each team can remove the tag for commands they are in charge.

### DIFF
--- a/src/Akeneo/UserManagement/back/Infrastructure/Cli/EventListener/AuthenticateCommandAsAdminUser.php
+++ b/src/Akeneo/UserManagement/back/Infrastructure/Cli/EventListener/AuthenticateCommandAsAdminUser.php
@@ -1,53 +1,33 @@
 <?php
 
-namespace Akeneo\UserManagement\Bundle\EventListener;
+namespace Akeneo\UserManagement\Infrastructure\Cli\EventListener;
 
 use Akeneo\Tool\Component\StorageUtils\Factory\SimpleFactoryInterface;
 use Akeneo\UserManagement\Bundle\Security\SystemUserToken;
 use Akeneo\UserManagement\Component\Model\UserInterface;
-use Doctrine\DBAL\DBALException;
+use Akeneo\UserManagement\Infrastructure\Cli\Registry\AuthenticatedAsAdminCommandRegistry;
+use Doctrine\DBAL\Exception as DBALException;
 use Doctrine\Persistence\ObjectRepository;
 use Symfony\Component\Console\Event\ConsoleCommandEvent;
 use Symfony\Component\Security\Core\Authentication\Token\Storage\TokenStorageInterface;
 
 /**
- * Create a user system if token is null on CLI command "pim" or "akeneo"
+ * Create a user system for the commands that need to be authenticated as admin user before executing
  * This listener is called before each command.
  *
  * @author    Marie Bochu <marie.bochu@akeneo.com>
  * @copyright 2017 Akeneo SAS (http://www.akeneo.com)
  * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
  */
-class CreateUserSystemListener
+class AuthenticateCommandAsAdminUser
 {
-    /** @var TokenStorageInterface */
-    private $tokenStorage;
-
-    /** @var ObjectRepository */
-    private $groupRepository;
-
-    /** @var ObjectRepository */
-    private $roleRepository;
-
-    /** @var SimpleFactoryInterface */
-    private $userFactory;
-
-    /**
-     * @param TokenStorageInterface  $tokenStorage
-     * @param ObjectRepository       $groupRepository
-     * @param ObjectRepository       $roleRepository
-     * @param SimpleFactoryInterface $userFactory
-     */
     public function __construct(
-        TokenStorageInterface $tokenStorage,
-        ObjectRepository $groupRepository,
-        ObjectRepository $roleRepository,
-        SimpleFactoryInterface $userFactory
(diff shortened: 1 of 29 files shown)


src/Akeneo/UserManagement/Bundle/EventListener/CreateUserSystemListener.php
```diff
@@ -20,42 +21,27 @@
...
     ) {
-        $this->tokenStorage = $tokenStorage;
-        $this->groupRepository = $groupRepository;
-        $this->roleRepository = $roleRepository;
-        $this->userFactory = $userFactory;
+    }
+
+    public function registerCommand(Command $command): void
+    {
+        $this->commandsThatNeedUserSystem[] = $command->getName();
```
LaurentPetard, line 36: Not good if the command is not implemented yet and configured as "lazy"

src/Akeneo/UserManagement/back/Infrastructure/Cli/Registry/AuthenticatedAsAdminCommandRegistry.php
```diff
@@ -0,0 +1,24 @@
...
+ * @copyright 2023 Akeneo SAS (https://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+final class AuthenticatedAsAdminCommandRegistry
+{
+    private array $commandsAuthenticatedAsAdminUser = [];
+
+    public function registerCommand(string $commandName): void
+    {
+        $this->commandsAuthenticatedAsAdminUser[] = $commandName;
```
nmarniesse, line 17:
```suggestion
        $this->commandsAuthenticatedAsAdminUser[$commandName] = $commandName;
```

Could be indexed in order to replace the `\in_array` below by `\array_key_exists` which is a lot quicker

src/Akeneo/UserManagement/back/Infrastructure/Cli/EventListener/AuthenticateCommandAsAdminUser.php
```diff
@@ -1,61 +1,41 @@
...
- * Create a user system if token is null on CLI command "pim" or "akeneo"
+ * Create a user system for the commands that need to be authenticated as admin user before executing
  * This listener is called before each command.
  *
  * @author    Marie Bochu <marie.bochu@akeneo.com>
  * @copyright 2017 Akeneo SAS (http://www.akeneo.com)
  * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
  */
-class CreateUserSystemListener
+class AuthenticateCommandAsAdminUser
```
nmarniesse, line 22:
```suggestion
class AuthenticateCommandAsAdminUserListener
```

src/Akeneo/UserManagement/back/Infrastructure/Cli/Registry/AuthenticatedAsAdminCommandRegistry.php
```diff
@@ -0,0 +1,24 @@
...
+    }
+
+    public function isCommandAuthenticatedAsAdminUser(string $commandName): bool
```
nmarniesse, line 20:
```suggestion
    public function isCommandNeededAuthenticationAsAdminUser(string $commandName): bool
```

nmarniesse (approved):

src/Akeneo/UserManagement/back/Infrastructure/Cli/Registry/AuthenticatedAsAdminCommandRegistry.php
LaurentPetard: We choose `akeneo.command.authenticated_as_admin_user` as tag name and `AuthenticatedAsAdminCommandRegistry` for the name of the registry. So I think that it's weird to add the verb `need` here. The name of the registry indicates that it registers "commands authenticated as admin user", not "commands that need to be authenticated as admin user".

nmarniesse: If in my code I call the method `isCommandAuthenticatedAsAdminUser` and it returns true, I assume the command had already the authentication, that is not true. `isCommandNeededAuthenticationAsAdminUser` is also not clear as you explain.  
Maybe `isCommandRegisteredToBeAuthenticatedAsAdminUser`? Or simply `isCommandRegistered`?

LaurentPetard: Simple is better. OK for `isCommandRegistered`

ahocquard (approved):
Could we add phpunit test (without Akeneo TestCase class) to just init the container and check that every commands have a static name defined? Just be sure to not load any command.

RiffVentura (approved):

PR #19260, opened 2023-02-23
OCT-224: Fixed no parameters message display and channel selection error

RiffVentura:
- fixed empty parameters placeholder display
- fixed a bug where a new target is poorly initialized

image

### DIFF
--- a/components/catalogs/front/src/components/ProductMapping/components/SourcePanel.tsx
+++ b/components/catalogs/front/src/components/ProductMapping/components/SourcePanel.tsx
@@ -49,6 +49,8 @@ export const SourcePanel: FC<Props> = ({target, source, onChange, errors}) => {
         shouldDisplayLocale ||
         shouldDisplayChannel ||
         shouldDisplayChannelLocale ||
+        shouldDisplayCurrency ||
+        shouldDisplayChannelCurrency ||
         shouldDisplayTranslationValue
     );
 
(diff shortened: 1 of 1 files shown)


components/catalogs/front/src/components/ProductMapping/components/SourcePanel.tsx
```diff
@@ -62,7 +64,7 @@ export const SourcePanel: FC<Props> = ({target, source, onChange, errors}) => {
             if (shouldDisplayTranslationValue) {
                 source = {...source, parameters: {...source.parameters, label_locale: null}};
             }
-            if (shouldDisplayCurrency) {
+            if (attribute.type === 'pim_catalog_price_collection') {
```
hogren, line 67: Can you explain please ?

hogren (approved):

components/catalogs/front/src/components/ProductMapping/components/SourcePanel.tsx
RiffVentura:
```typescript
const shouldDisplayCurrency = source !== null && !attribute?.scopable && attribute?.type === 'pim_catalog_price_collection';
````
evaluates to undefined before any attribute is selected
The point of `initSource` is to initialize Source value on attribute selection.
Due to it being a useCallback and it being called before source onChange, when function value is evaluated, `shouldDisplayCurrency` is still undefined and the source is now missing parameters property which breaks JSX later.

But it shouldn't use the flag in the first place as it would only work only for non-scopable attribute
The solution is to use attribute type as the source of truth (which semantically makes the most sense IMO)

hogren: :+1: thank you @RiffVentura

tseho (approved):

PR #19270, opened 2023-02-24
CXP-1409: Map multi-select attributes (string[])

masacc:

image

image

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc


guillaumefer-akeneo (approved):

components/catalogs/front/src/components/ProductMapping/ProductMapping.tsx
```diff
@@ -44,12 +45,7 @@ export const ProductMapping: FC<Props> = ({productMapping, productMappingSchema,
             if (productMappingSchema === undefined) {
                 return;
             }
-            const target: Target = {
-                code: targetCode,
-                label: productMappingSchema.properties[targetCode]?.title ?? targetCode,
-                type: productMappingSchema.properties[targetCode].type,
-                format: productMappingSchema.properties[targetCode].format ?? null,
-            };
+            const target = createTargetFromProductMappingSchema(targetCode, productMappingSchema);
```
RiffVentura, line 48: I would move the requirements below too since it relates to the target initialization as well

RiffVentura (approved):
:+1:

components/catalogs/front/src/components/ProductMapping/utils/createTargetSourceAssociationsFromProductMapping.ts
```diff
@@ -1,7 +1,7 @@
 import {ProductMapping as ProductMappingType} from '../models/ProductMapping';
 import {Source} from '../models/Source';
 
-export const createTargetsFromProductMapping = (mapping: ProductMappingType): [string, Source][] => {
+export const createTargetSourceAssociationsFromProductMapping = (mapping: ProductMappingType): [string, Source][] => {
```
RiffVentura, line 4: is still necessary since the backend should already sort the target list :thinking:

masacc: Yes you are right, we can replace it only by an `Object.entries`. Good point !

PR #19281, opened 2023-02-27
PIM-10820: Partially revert [PIM-10350] to fix case sensitivity on op…

cemorin:
…tions import

Due to this PR: https://github.com/akeneo/pim-community-dev/pull/16751/files
When importing a simple or multi select option with the wrong case, it updated the option case.
This issue it resolved was to not add a new version in product history.
It has been validated to partially revert this previous fix, and keep the versioning with both cases. 

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/src/Akeneo/Pim/Enrichment/Component/Product/Factory/NonExistentValuesFilter/NonExistentSimpleSelectValuesFilter.php
+++ b/src/Akeneo/Pim/Enrichment/Component/Product/Factory/NonExistentValuesFilter/NonExistentSimpleSelectValuesFilter.php
@@ -25,17 +25,16 @@ class NonExistentSimpleSelectValuesFilter implements NonExistentValuesFilter
             return $onGoingFilteredRawValues;
         }
 
-        $optionCodes = $this->getExistingOptionCodes($selectValues);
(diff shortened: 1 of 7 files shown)


mmetayer (approved):
TODO: Add a class description for these 2 services (the implem is efficient, but not easy to understand, we should describe what exactly are these services supposed to do)

src/Akeneo/Pim/Enrichment/Component/Product/Factory/NonExistentValuesFilter/NonExistentSimpleSelectValuesFilter.php
```diff
@@ -92,7 +84,7 @@ private function getOptionCodes(array $selectValues): array
 
         $uniqueOptionCodes = [];
         foreach ($optionCodes as $attributeCode => $optionCodeForThisAttribute) {
-            $uniqueOptionCodes[$attributeCode] = \array_unique($optionCodeForThisAttribute);
+            $uniqueOptionCodes[$attributeCode] = array_unique($optionCodeForThisAttribute);
```
antoinearrive, line 87: why remove \ ?

antoinearrive (approved):

PR #19285, opened 2023-02-27
CXP-1439: 3 component split of Source Panel

RiffVentura:
Refactoring of Source Panel into 3 main components: Requirements, Source Selection, and Source Parameters

### DIFF
--- a/components/catalogs/front/src/components/ProductMapping/components/SourceParameters/SourceParameters.tsx
+++ b/components/catalogs/front/src/components/ProductMapping/components/SourceParameters/SourceParameters.tsx
@@ -1,5 +1,4 @@
 import React, {FC} from 'react';
-import {SectionTitle, Tag} from 'akeneo-design-system';
 import {Source} from '../../models/Source';
 import {SourceErrors} from '../../models/SourceErrors';
 import {SelectLabelLocaleDropdown} from './SelectLabelLocaleDropdown';
@@ -8,29 +7,19 @@ import {SelectChannelCurrencyDropdown} from './SelectChannelCurrenciesDropdown';
 import {useTranslate} from '@akeneo-pim-community/shared';
 import styled from 'styled-components';
 import {useAttribute} from '../../../../hooks/useAttribute';
+import {SourceSectionTitle} from '../SourceSectionTitle';
 
 const Information = styled.p`
     font-style: italic;
     margin-top: 10px;
 `;
 
-const SectionSubtitle = () => {
-    const translate = useTranslate();
-    return (
-        <SectionTitle>
-            <Tag tint='purple'>2</Tag>
-            <SectionTitle.Title level='secondary'>
-                {translate('akeneo_catalogs.product_mapping.source.parameters.title')}
-            </SectionTitle.Title>
-        </SectionTitle>
-    );
-};
-
 type Props = {
     source: Source | null;
     errors: SourceErrors | null;
     onChange: (value: Source) => void;
 };
+
 export const SourceParameters: FC<Props> = ({source, errors, onChange}) => {
     const translate = useTranslate();
     const {data: attribute} = useAttribute(source?.source ?? '');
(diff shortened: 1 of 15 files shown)


components/catalogs/front/src/components/ProductMapping/components/SourceParameters/SourceParameters.tsx
```diff
@@ -0,0 +1,73 @@
...
+                    {translate('akeneo_catalogs.product_mapping.source.parameters.no_parameters_message')}
+                </Information>
+            </>
+        );
+    }
+
+    return (
+        <>
+            <SectionSubtitle />
+            {undefined !== source.parameters.label_locale && (
```
helenegzlz, line 52: I am probably just quibbling, but in some lines above, the comparisons between two values ex: "attribute === undefined" is not in Yoda's style but here it is. Maybe we can harmonize it? Or is there a reason?

helenegzlz (approved):

eoliveauakeneo (approved):

components/catalogs/front/src/components/ProductMapping/components/SourceParameters/SourceParameters.tsx
RiffVentura: done

components/catalogs/front/src/components/ProductMapping/components/SourceSelection/SourceSelection.tsx
```diff
@@ -0,0 +1,46 @@
+import React, {FC} from 'react';
+import {Source} from '../../models/Source';
+import {Target} from '../../models/Target';
+import {SourceErrors} from '../../models/SourceErrors';
+import {useAttribute} from '../../../../hooks/useAttribute';
+import {Attribute} from '../../../../models/Attribute';
+import {SectionTitle, Tag} from 'akeneo-design-system';
```
masacc, line 7:
```suggestion
import {SectionTitle as SectionSubtitle, Tag} from 'akeneo-design-system';
```
to be consistent with `SourceParameters` ?
WDYT ? Is it over-engineered ?

components/catalogs/front/src/components/ProductMapping/components/SourceSelection/SourceSettings.tsx
```diff
@@ -0,0 +1,52 @@
...
+    errors: SourceErrors | null;
+    onChange: (value: Source) => void;
+};
+
+export const SourceSettings: FC<Props> = ({source, attribute, errors, onChange}) => {
+    const onChangeMiddleware = useCallback(
+        source => {
+            if (
+                (attribute.type === 'pim_catalog_simpleselect' || attribute.type === 'pim_catalog_multiselect') &&
+                (undefined === source.parameters?.label_locale || null === source.parameters?.label_locale)
```
masacc, line 21: is this condition useful?

masacc (approved):

components/catalogs/front/src/components/ProductMapping/components/SourceSelection/SourceSettings.tsx
RiffVentura: Even if we assume that the source is properly initialized, from function POV you are still dealing with a source instance where parameters are optional. So I would argue that yes
It can be avoided if Source type-hinting is more clever (e.g union types on the criteria in product selection) but that should be done in another PR

components/catalogs/front/src/components/ProductMapping/components/SourceSelection/SourceSelection.tsx
RiffVentura: SectionSubtitle in SourceParameters is a component that contains multiple components (that uses SectionTitle)
Here SectionTitle is used to display a single string so it would be strange to alias it that way don't you think?

masacc: Yes that's ok for me, just wondering if we can have something cleaner to avoid confusion with the 2 levels of title. What do you think about extracting the SectionSubtitle as a standalone component with 2 props (number and title) and using it in both places?

RiffVentura: done [here](https://github.com/akeneo/pim-community-dev/pull/19285/commits/21e67b26c71a06e9ccb45859f0efe2a7e82da40d)

PR #19296, opened 2023-02-28
PIM-10861: Fix calcuation of completeness channel ratio on Activity Dashboard

LaurentPetard:
The completeness ratio of a channel on the Activity dashboard is not well calculated. It should be the percent of products that have 100% completeness on all locales on that channel.

I reworked the ES query to do the calculation with aggregations.

Example for 4 products, with 2 products fully enriched on all locales
Before:
PIM-10861-before

After:
PIM-10861-after

### DIFF
--- a/src/Akeneo/Pim/Enrichment/Bundle/Storage/ElasticsearchAndSql/FollowUp/GetCompletenessPerChannelAndLocale.php
+++ b/src/Akeneo/Pim/Enrichment/Bundle/Storage/ElasticsearchAndSql/FollowUp/GetCompletenessPerChannelAndLocale.php
@@ -41,14 +41,12 @@ class GetCompletenessPerChannelAndLocale implements GetCompletenessPerChannelAnd
     {
         $categoriesCodeAndLocalesByChannel = $this->getCategoriesCodesAndLocalesByChannel();
 
-        $totalProductsByChannel = $this->countTotalProductsInCategoriesByChannel($categoriesCodeAndLocalesByChannel);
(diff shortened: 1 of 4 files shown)


nmarniesse (approved):

jpa44 (approved):

src/Akeneo/Pim/Enrichment/Bundle/Storage/ElasticsearchAndSql/FollowUp/GetCompletenessPerChannelAndLocale.php
```diff
@@ -126,22 +124,47 @@ private function getCategoriesCodesAndLocalesByChannel(): array
...
     {
         if (empty($categoriesCodeAndLocalesByChannels)) {
             return [];
         }
 
         $body = [];
-
         foreach ($categoriesCodeAndLocalesByChannels as $categoriesCodeAndLocalesByChannel) {
+            $aggregations = [
+                'all_locales' => [
```
mmetayer, line 146: wouldn't it be more consistent with "<all_locales>"? 
also there's a (very very unlikely) case a custom `all_locales` locale was created by import?
Not sure this is possible though, so non blocking

mmetayer (approved):

src/Akeneo/Pim/Enrichment/Bundle/Storage/ElasticsearchAndSql/FollowUp/GetCompletenessPerChannelAndLocale.php
LaurentPetard: the character `<` is forbidden as aggregation name, I choose `--` instead, as it's not allowed for locale codes

PR #19329, opened 2023-03-02
RAB-1287: Move delete attribute groups error into a dedicated subscriber 

StevenVAIDIE:

In this PR, 

- I move all check before deletion into a dedicated subscriber, by this way we will have those check in mass delete 
- I also call on confirm on DoubleCheckModalDeletion when user hit Enter on confirmation input
- I fixed some translations that are not plurialized

- [x] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/src/Akeneo/Pim/Structure/Bundle/Resources/translations/jsmessages.en_US.yml
+++ b/src/Akeneo/Pim/Structure/Bundle/Resources/translations/jsmessages.en_US.yml
@@ -286,8 +286,8 @@ pim_enrich.entity.attribute_group:
         confirm: '{1}Are you sure you want to delete this attribute group?|]1, Inf[Are you sure you want to delete these {{ count }} attribute groups?'
         confirmation_phrase: 'Please type "{{ confirmation_word }}" to confirm:'
         confirmation_word: delete
-        attribute_warning: '{1}Deleting these attribute groups will impact {{ number_of_attribute }} child attribute. You will need to move this child attribute to a new attribute group.]1, Inf[Deleting these attribute groups will impact {{ number_of_attribute }} child attributes. You will need to move these child attributes to a new attribute group.'
-        select_attribute_group: '{1}Please select the attribute group to move the child attribute to:]1, Inf[Please select the attribute group to move the {{ number_of_attribute }} child attribute to:'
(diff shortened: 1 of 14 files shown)


src/Akeneo/Pim/Structure/Bundle/Resources/translations/jsmessages.en_US.yml
```diff
@@ -283,11 +283,11 @@ pim_enrich.entity.attribute_group:
...
-        confirm: '{1}Are you sure you want to delete this attribute group ?|]1, Inf[Are you sure you want to delete these {{ count }} attribute groups ?'
-        confirmation_phrase: 'Please type "{{ confirmation_word }}" to confirm.'
+        confirm: '{1}Are you sure you want to delete this attribute group?|]1, Inf[Are you sure you want to delete these {{ count }} attribute groups?'
+        confirmation_phrase: 'Please type "{{ confirmation_word }}" to confirm:'
         confirmation_word: delete
-        attribute_warning: 'Deleting these attribute groups will impact {{ number_of_attribute }} child attributes. You will need to move these child attributes to a new attribute group.'
-        select_attribute_group: 'Please select the attribute group to move the {{ number_of_attribute }} child attributes to:'
+        attribute_warning: '{1}Deleting these attribute groups will impact {{ number_of_attribute }} child attribute. You will need to move this child attribute to a new attribute group.]1, Inf[Deleting these attribute groups will impact {{ number_of_attribute }} child attributes. You will need to move these child attributes to a new attribute group.'
```
MadeWilson, line 289: reading this I wonder if the plural shouldn't be on child instead of attribute

MadeWilson (approved):

src/Akeneo/Pim/Structure/Bundle/Resources/translations/jsmessages.en_US.yml
StevenVAIDIE: For me child attributes is good enough but my English is approximative @ValentinMumble ?

ValentinMumble: After researching, it looks like it depends on the noun used as an adjective ahah https://english.stackexchange.com/questions/468019/children-as-adjective-as-opposed-to-child
> We don't say teethache, teethpaste, teeth decay, mice trap, micehole. So it definitely doesn't apply to irregular plurals in general
> Google ngram search of "women soldiers" and "women doctors" shows the "women" form far more popular throughout history than the corresponding"woman" forms.
I liked the example of a shoe store -> we don't say a shoes store even though we usually buy multiple shoes
But I don't really know which apply more in our case 🤷 , we need a native english speaker here

ValentinMumble (approved):

PR #19331, opened 2023-03-02
CPM-986: Fix cache on get families

pierallard:
There is a cache on queries with react-query
When we did a call for having the families from another component, then wanted to have the same call but updating the "limit", the cache was telling is that we already having this result, which is not the case.
We added the limit to the parameters able to change and calling a new call.


mmetayer (approved):

Beninho (approved):

cemorin (approved):
MULTI GTM

PR #19360, opened 2023-03-06
JEL-239: Add test to check that all Symfony commands have a default name

LaurentPetard:


src/Akeneo/Tool/Bundle/MessengerBundle/tests/integration/CheckSymfonyCommandsDefaultNameIntegration.php
```diff
@@ -0,0 +1,31 @@
...
+use PHPUnit\Framework\Assert;
+use Symfony\Bundle\FrameworkBundle\Test\KernelTestCase;
+
+/**
+ * @copyright 2023 Akeneo SAS (https://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+final class CheckSymfonyCommandsDefaultNameIntegration extends KernelTestCase
+{
+    public function testAllSymfonyCommandsHaveADefaultName(): void
```
ahocquard, line 16: could you explain the why please in a comment

as a detail, we mainly use snake case for tests now

ahocquard (approved):

src/Akeneo/Tool/Bundle/MessengerBundle/tests/integration/CheckSymfonyCommandsDefaultNameIntegration.php
```diff
@@ -0,0 +1,31 @@
...
+    {
+        $container = static::getContainer();
+
+        // The parameter 'console.command.ids' contains the commands with deprecated registration (i.e. without defined static $defaultName)
```
nmarniesse, line 20: Something bothers me: are we sure this implem is future-proof? I mean it looks like an internal implem in the symfony console component, could it change in a next version?

src/Akeneo/Tool/Bundle/MessengerBundle/tests/integration/CheckSymfonyCommandsDefaultNameIntegration.php
LaurentPetard: Good to know for the snake case usage in integration tests. I think that a global communication about that would be nice because I'm surely not the only one that didn't know this changing ^^

src/Akeneo/Tool/Bundle/MessengerBundle/tests/integration/CheckSymfonyCommandsDefaultNameIntegration.php
LaurentPetard: Yes it bothers me too, and of course it's impossible to be sure that it's future-proof, but the risk is low. The test will fail if they rename the parameter or use it for other stuff, so we will be able to adapt it.

The only other way I found to list commands from the container was to get every Akeneo services with the method `getServiceIds` and load them to check if it's a command or not. But it's not better, because it implements every services, including the lazy ones, and it's not future-proof because the method `getServiceIds` is not declared in `ContainerInterface`

nmarniesse: Ok let's merge like that, it will not be a hard work to refactor later if needed

nmarniesse (approved):

PR #19405, opened 2023-03-10
AST-527: Fix memoize callback issue with DateInput component

olivierpontier:

Ensure the right reference of the onChange handler is called when the DateInput is rerendered.

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc


juliensnz (approved):

olivierpontier:
> It might be useful to add this rules for hooks so that this won't happen again? https://reactjs.org/docs/hooks-rules.html#eslint-plugin What do you think? (maybe not in this PR)

Good point! I checked in the PIM. It looks to have already been set in akeneo-design-system (and other yarn workspaces): https://github.com/akeneo/pim-community-dev/blob/a94bf7feb207222be6d4d8a4b5fc9b885120a296/front-packages/akeneo-design-system/.eslintrc.json#L23

PhpStorm can display those warnings, but sometimes it is not relevant. 

Maybe we have to take the habit of testing this kind of use case with the rerender function. It looks like an interesting topic for the front guild, does it?

PR #19406, opened 2023-03-10
RAB-1300: Define a limit for attribute groups creation

ghoussard:

In this PR, 

We added a limit of attribute groups, now user cannot anymore create an attribute group when it reach 1000 attribute groups. It can continue to edit attribute group even if he as more than 1000 attribute groups.

Actually only one client have more than 1000 attribute groups (half have no attribute inside). The client have been advert. We didn't handle correctly the import due to the fact that attribute group are created by batch (unitary validated before). We cannot create more that 1000 + attribute group batch size (100)

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc


src/Akeneo/Platform/Bundle/UIBundle/Resources/config/form_extensions/attribute_group/index.yml
```diff
@@ -1,47 +0,0 @@
-extensions:
```
ghoussard, line 1: Not used anymore

StevenVAIDIE (approved):

src/Akeneo/Platform/Bundle/UIBundle/Resources/config/form_extensions/attribute_group/index.yml
ValentinMumble: 🔥

ValentinMumble (approved):

mmetayer (commented):
some tests are missing:
- unit tests for the new validator
- being able to update an existing attribute group even if the limit is reached
- what happens for import or internal API? (best way to test that would be an acceptance test)

pierallard (approved):
We usually don't put const in Validator (your limit for example), but as a parameter in the .yml validator.
But that's OK for me, it's just a detail.

mmetayer (approved):

PR #19425, opened 2023-03-14
PIM-10889 : Update the category updated at on update category

antoinetrouve:

**What's the problem ?**

The Category updated date is not updated when update category label. The user cannot retrieve the category by its updated date.
See issue [here](https://akeneo.atlassian.net/jira/software/c/projects/PIM/boards/9?modal=detail&selectedIssue=PIM-10889&quickFilter=650).  

Under the hood, the updated date is not really managed by the domain model (but the SQL query only) that's why we encountered some unexpected behavior regarding this date.

**The solution**
After a design meeting.
As a maintenance domain and because the problem impact the public API, we decided to implement a quick fix. This solution stay ISO with [the updated date implementation](https://github.com/akeneo/pim-community-dev/blob/master/src/Akeneo/Category/back/Infrastructure/Storage/Save/Query/UpsertCategoryBaseSql.php#L115) and can be quickly refacto.

We manage the updated date with a specific [sql query services](https://github.com/akeneo/pim-community-dev/blob/bd1f1ce073c7c494f8b0adf9e19b479b8d1f04b7/src/Akeneo/Category/back/Infrastructure/Storage/Sql/Update/UpdateCategoryUpdatedDateSql.php) to handle the updated date when the user update Category's label.

**Result**
Before to update de_DE: 
Screenshot from 2023-03-14 12-14-46

After to update label de_DE: 
Screenshot from 2023-03-14 12-16-24

**What's next ?**

Griffin's squad need to design and provide a better solution to handle Category updated date as a functional matters.
See [here](https://akeneo.atlassian.net/jira/software/projects/GRF/boards/301?selectedIssue=GRF-781) to provide a global solution.

- [x] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc


src/Akeneo/Category/back/tests/Integration/Infrastructure/Storage/Sql/Update/UpdateCategoryUpdatedDateSqlIntegration.php
```diff
@@ -0,0 +1,64 @@
...
+        $previousCategory = $this->get(GetCategoryInterface::class)->byCode('socks');
+
+        sleep(1);
+        $this->get(UpdateCategoryUpdatedDate::class)->execute('socks');
+
+        /** @var Category $updatedCategory */
+        $updatedCategory = $this->get(GetCategoryInterface::class)->byCode('socks');
+
+        $this->assertEquals((string)$updatedCategory->getCode(), (string)$previousCategory->getCode());
+        $this->assertNotEqualsWithDelta(
```
etiennekerbastard, line 40: This might turn out to be a flaky test: If the `$this->get(UpdateCategoryUpdatedDate::class)->execute('socks');` does not happen instantly the delta will be superior to 1. You could just compare the two dates and verify than one is more recent than the other.

antoinetrouve: After discuss with you, We decided to use Master category from the functional catalog to avoird using sleep function. The I reworks the assert test using assertGreaterThan to be more human speaking.

jeremyBeaucousinAkeneo (approved):

BryanS29-akeneo (approved):

PR #19429, opened 2023-03-15
GRF-776 : Create ZDD migration to replace empty category labels by NULL

jeremyBeaucousinAkeneo:
The initial migration was identified with execution time risk on deployment.

It was removed to be reimplemented differently.

Three different implementations have been tested with 1 million lines.

- Global update (initial implementation) :
```php
$this->connection->executeQuery(<<<SQL
    UPDATE pim_catalog_category_translation
    SET label=NULL
    WHERE label = '';
SQL);
```
The migration took 8,87 seconds : 
```
Start migration with 1000000 lines
End migration
 8.874019441 seconds

Time: 01:51.588, Memory: 84.50 MB

```
- With limit:
```php
$this->connection->transactional(function () {
  $limit = 1000;
  $sql = <<<SQL
     SELECT id
     FROM pim_catalog_category_translation
     WHERE label = ''
     LIMIT :limit; 
  SQL;
   while (true) {
            $rows = $this->connection->fetchAllAssociative($sql,
                [
                    'limit' => $limit,
                ],
                [
                    'limit' => \PDO::PARAM_INT,
                ]
            );
            if (empty($rows)) {
                return;
            }
            $ids = array_map(fn ($row) => $row['id'], $rows);
            $this->updateCategoryTranslation($ids);
        }
    });
```
The migration took 63,43 seconds : 
```
Start migration with 1000000 lines
End migration
 63.435844639 seconds

Time: 01:40.883, Memory: 84.50 MB

```
- By package of 1000 (greater than):
```php
$this->connection->transactional(function () {
      $formerId = 0;
      $sql = <<<SQL
         SELECT id
         FROM pim_catalog_category_translation
         WHERE label = ''
         AND id > :limit; 
      SQL;
      while (true) {
          $rows = $this->connection->fetchAllAssociative($sql,
              [
                  'limit' => $formerId,
              ],
              [
                  'limit' => \PDO::PARAM_INT,
              ]
          );
          if (empty($rows)) {
              return;
          }
          $formerId = (int)end($rows)['id'] + 1000;
          $ids = array_map(fn ($row) => $row['id'], $rows);
          $this->updateCategoryTranslation($ids);
      }
  });

```
The migration took 16.17 seconds : 
```
Start migration with 100000 lines
End migration
 16.176794722 seconds

Time: 01:40.883, Memory: 84.50 MB

```

The migration has also been run on a cloned instance from srnt-fotronic which has more than 20k categories.
The migration took 0ms : [datadog](https://app.datadoghq.eu/logs?query=kube_namespace%3Asrnt-pimci-long-duplic-srnt-fotronic-671666&additional_filters=%5B%7B%7D%5D&cols=host%2Cservice&index=%2A&messageDisplay=inline&stream_sort=time%2Cdesc&viz=stream&from_ts=1679071500000&to_ts=1679071800000&live=false)

We decided to keep the implementation as it was with the test results.

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc


src/Akeneo/Category/back/Infrastructure/Migration/V20230313SetEmptyCategoryLabelsToNullZddMigration.php
```diff
@@ -0,0 +1,45 @@
...
+            UPDATE pim_catalog_category_translation
+            SET label=NULL
+            WHERE label = '';
+        SQL);
+    }
+
+    public function migrateNotZdd(): void
+    {
+        // Do nothing
+    }
```
etiennekerbastard, line 39: Why do we need this?

jeremyBeaucousinAkeneo: I reuse the interface that made us implement this signature.

This is the documentation on the interface :
 ```php
/**
     * As any change in a ZDD Migration have to be reflected in the database schema,
     * we need to provide a way to migrate also without ZDD.
     * It's useful for users running the PIM out of Saas.
*/
public function migrateNotZdd(): void;
```

I understand there are several mechanisms to execute migrations, maybe @ghoussard  could explain us why.

Comment (ghoussard) [reply src/Akeneo/Category/back/Infrastructure/Migration/V20230313SetEmptyCategoryLabelsToNullZddMigration.php:39]: Hello, we added this method during the materialization of v7. As you know, ZDD migration was introduced during the previous year and are ran outside the classic migration workflow. That means when you're a Flex user and you want to upgrade, you have to run the classic migration workflow, so you're missing ZDD migration. That was a major issue which made the PIM v7 not working.
To resolve it, we decided to create some doctrine migration (see https://github.com/akeneo/pim-community-dev/pull/19019) which directly calls ZDD migration.
And why a `notZddMigration` method ? It's because chipmunks ZDD migrations contains a lot of logging and we don't want to have these logs when we run the classic migration workflow. So if you don't have logging, you can directly call `$this->migrateZdd()` inside it, however you have to re-implement it without logging.
 Hope it's clearer for you. If not, we can discuss IRL 👍🏻

Comment (etiennekerbastard) [reply src/Akeneo/Category/back/Infrastructure/Migration/V20230313SetEmptyCategoryLabelsToNullZddMigration.php:39]: @ghoussard Cristal clear! Thanks :)

**Review by etiennekerbastard — APPROVED** [review]

File: src/Akeneo/Category/back/Infrastructure/Migration/V20230313SetEmptyCategoryLabelsToNullZddMigration.php [hunk 2]
```diff
@@ -0,0 +1,45 @@
...
+    /**
+     * @throws Exception
+     */
+    public function migrate(): void
+    {
+        $this->connection->executeQuery(<<<SQL
+            UPDATE pim_catalog_category_translation
+            SET label=NULL
+            WHERE label = '';
+        SQL);
```
amigaire, line 33: This may take a long time, creating a lock on the table.
Should we batch it in multiple transactions?

jeremyBeaucousinAkeneo: Good point, we create this migration in ZDD for that reason, for clients that have a huge amount of category that could be safer.

I'll do it.

jeremyBeaucousinAkeneo: New implementations have been tested without relevant results.

pchasle (approved):

amigaire (approved):

PR #19436, opened 2023-03-15
CPM-995: Reorder identifier generators

mmetayer:

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/components/identifier-generator/front/src/feature/pages/ListPage.tsx
+++ b/components/identifier-generator/front/src/feature/pages/ListPage.tsx
@@ -1,5 +1,5 @@
 import React, {useCallback, useEffect, useMemo, useState} from 'react';
-import {LocaleCode, PageContent, useSecurity, useTranslate, useUserContext} from '@akeneo-pim-community/shared';
+import {PageContent, useRouter, useSecurity, useTranslate, useUserContext} from '@akeneo-pim-community/shared';
 import {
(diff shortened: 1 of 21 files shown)


components/identifier-generator/front/src/feature/pages/ListPage.tsx
```diff
@@ -13,20 +13,173 @@ import {
...
+type ListTableProps = {
+  isErrored: boolean;
+  isLoading: boolean;
+  isManageIdentifierGeneratorAclGranted: boolean;
+  generators: IdentifierGenerator[];
+  onDelete: (code: string) => () => void;
+  identifierAttributes: FlattenAttribute[];
+  filtered: boolean;
+};
+const ListTable: React.FC<ListTableProps> = ({
```
Beninho, line 32: I think it would be clearer to separate in another file.
ListTableProps and ListTable

Beninho (approved):

components/identifier-generator/front/src/feature/pages/ListPage.tsx
mmetayer: that would imply exporting the component, and we don't want to expose it in the rest of the application

cemorin (approved):

PR #19438, opened 2023-03-15
SP-490: add a service API to retrieve a list of families or a single Family

rouliane:
This PR adds a new public service API in the structure context to get a list of families (or a single one too), that returns basic generic Family read models. We needed this service API for Supplier Portal needs. For now the read model only has what we needed in our side (code and labels), but it should be easy to add new properties if needed.

As discussed with Mathias, there is not LRU cache implementation for now, we thought it would not be needed at this point.

### DIFF
--- a/src/Akeneo/Pim/Structure/Bundle/Query/PublicApi/Family/Sql/SqlGetFamilies.php
+++ b/src/Akeneo/Pim/Structure/Bundle/Query/PublicApi/Family/Sql/SqlGetFamilies.php
@@ -1,0 +1,54 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Akeneo\Pim\Structure\Bundle\Query\PublicApi\Family\Sql;
+
+use Akeneo\Pim\Structure\Component\Query\PublicApi\Family\Family;
+use Akeneo\Pim\Structure\Component\Query\PublicApi\Family\GetFamilies;
+use Doctrine\DBAL\Connection;
+
+class SqlGetFamilies implements GetFamilies
+{
+    public function __construct(private readonly Connection $connection)
+    {
+    }
+
+    public function byCodes(array $familyCodes): array
+    {
+        if (empty($familyCodes)) {
+            return [];
+        }
+
+        $sql = <<<SQL
+SELECT
+   family.code AS code,
+   COALESCE(JSON_OBJECTAGG(trans.locale, trans.label), JSON_ARRAY()) AS labels
+FROM pim_catalog_family family
+INNER JOIN pim_catalog_family_translation trans ON family.id = trans.foreign_key
+WHERE family.code IN (:familyCodes)
+GROUP BY family.code
+SQL;
+        $rows = $this->connection->executeQuery(
+            $sql,
+            [
+                'familyCodes' => $familyCodes,
+            ],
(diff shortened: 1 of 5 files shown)


src/Akeneo/Pim/Structure/Bundle/Query/PublicApi/Family/Sql/SqlGetFamilies.php
```diff
@@ -0,0 +1,58 @@
...
+        }
+
+        return $families;
+    }
+
+    public function byCode(string $familyCode): ?Family
+    {
+        $byCodes = $this->byCodes([$familyCode]);
+
+        if ([] === $byCodes) {
```
phaseinducer, line 52:
```suggestion
        if (empty($byCodes)) {
```
?

src/Akeneo/Pim/Structure/Bundle/Query/PublicApi/Family/Sql/SqlGetFamilies.php
```diff
@@ -0,0 +1,58 @@
...
+            $sql,
+            [
+                'familyCodes' => $familyCodes,
+            ],
+            ['familyCodes' => Connection::PARAM_STR_ARRAY]
+        )->fetchAllAssociative();
+
+        $families = [];
+        foreach ($rows as $row) {
+            $families[$row['code']] = new Family($row['code'], json_decode($row['labels'], true));
```
phaseinducer, line 42: Indexing by codes shouldn't be hidden (it doesn't appear in the interface). Why not using a phpstan return type?

kpAkeneo (approved):

src/Akeneo/Pim/Structure/Component/Query/PublicApi/Family/GetFamilies.php
```diff
@@ -0,0 +1,13 @@
...
+
+declare(strict_types=1);
+
+namespace Akeneo\Pim\Structure\Component\Query\PublicApi\Family;
+
+interface GetFamilies
+{
+    public function byCode(string $familyCode): ?Family;
+
+    /** @return Family[] */
```
mmetayer, line 11:
```suggestion
    /** @return array<string, Family> */
```

src/Akeneo/Pim/Structure/Bundle/Query/PublicApi/Family/Sql/SqlGetFamilies.php
```diff
@@ -0,0 +1,58 @@
...
+
+    public function byCode(string $familyCode): ?Family
+    {
+        $byCodes = $this->byCodes([$familyCode]);
+
+        if ([] === $byCodes) {
+            return null;
+        }
+
+        return array_pop($byCodes);
```
mmetayer, line 56:
```suggestion
        return $byCodes[$familyCode] ?? null;
```

src/Akeneo/Pim/Structure/Bundle/Query/PublicApi/Family/Sql/SqlGetFamilies.php
rouliane: It's not something we need really, I don't know why I wrote it this way, should I simply remove it instead ?

mmetayer (approved):

PR #19473, opened 2023-03-22
RAB-1360: Always display other in replacement attribute groups

ValentinMumble:
- Always display other in replacement attribute groups


pierallard (approved):

ValentinMumble:
w

PR #19524, opened 2023-03-29
CXP-1454: Map status (boolean)

helenegzlz:
The Status system attribute is suggested for the boolean source.
image
The Status is not suggested when the attribute is not a boolean, string in this example.
image

- [x] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc


components/catalogs/back/src/Application/Mapping/ValueExtractor/Extractor/ValueExtractorInterface.php
```diff
@@ -29,6 +29,8 @@ interface ValueExtractorInterface
     public const SOURCE_TYPE_CATEGORIES = 'categories';
     public const SOURCE_TYPE_FAMILY = 'family';
 
+    public const SOURCE_TYPE_STATUS = 'status';
+
```
masacc, line 33:
```suggestion
    public const SOURCE_TYPE_STATUS = 'status';

```
rm blank line

components/catalogs/back/src/Infrastructure/Command/CatalogFixtureCommand.php
```diff
@@ -293,6 +298,10 @@ private function getProductMappingSchemaRaw(): string
                 "type": "string"
               },
               "title": "Categories"
+            },
+            "enabled": {
+              "type": "boolean",
+              "title": "Is enabled"
```
masacc, line 304: there is already a boolean target type in these fixtures, I don't think it's necessary to add another one

components/catalogs/back/tests/Integration/Infrastructure/Validation/ProductMapping/AttributeSource/SystemAttributeStatusSourceTest.php
```diff
@@ -0,0 +1,139 @@
...
+    {
+        parent::setUp();
+        $this->createAttribute([
+            'code' => 'status',
+            'type' => 'status',
+            'scopable' => false,
+            'localizable' => false,
+            'attribute_group_code' => 'system',
+            'attribute_group_label' => 'System',
+        ]);
```
masacc, line 28: you can't create an attribute with code `status`, it's a reserved word from Akeneo PIM (maybe it's only a front check, it's weird it didn't complain). Moreover, you don't want to check on an attribute here but on the product status property so you don't need to create an attribute.

components/catalogs/front/src/components/ProductMapping/components/SourceSelection/SourceSettings.tsx
```diff
@@ -41,7 +41,7 @@ export const SourceSettings: FC<Props> = ({source, attribute, errors, onChange})
             }
 
             if (
-                (source.source === 'categories' || source.source === 'family') &&
+                (source.source === 'categories' || source.source === 'family' || source.source === 'status') &&
```
masacc, line 44:
```suggestion
                (['categories', 'family', 'status'].incluedes(source.source) &&
```

masacc (approved):

components/catalogs/back/tests/Integration/Infrastructure/Validation/ProductMapping/AttributeSource/SystemAttributeStatusSourceTest.php
helenegzlz: It's removed now :)

PR #19531, opened 2023-03-29
CPM-1018: Add exceptions on Repositories

pierallard:


components/identifier-generator/back/tests/Acceptance/Context/DatabaseContext.php
```diff
@@ -53,7 +54,17 @@ public function theIdentifierGenerator(string $generatorCode): void
      */
     public function theIdentifierShouldNotBeCreated(): void
     {
-        Assert::null($this->generatorRepository->get(BaseCreateOrUpdateIdentifierGenerator::DEFAULT_IDENTIFIER_GENERATOR_CODE));
+        try {
+            $this->generatorRepository->get(BaseCreateOrUpdateIdentifierGenerator::DEFAULT_IDENTIFIER_GENERATOR_CODE);
+
+            throw new \InvalidArgumentException(
+                \sprintf(
+                    'An identifier generator with code "%s" was created.',
```
antoinearrive, line 62:
```suggestion
                    'An identifier generator with code "%s" was not created.',
```
 ?

antoinearrive (approved):

components/identifier-generator/back/tests/Acceptance/Context/DatabaseContext.php
pierallard: No ; it the "get" is OK, it means an identifier was created, and should not !

PR #19545, opened 2023-04-03
RAB-1330: DSM component: tiles

MadeWilson:

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc


front-packages/akeneo-design-system/src/components/Tiles/Tiles.tsx
```diff
@@ -6,30 +6,47 @@ import {Key, Override} from '../../';
 
 type Size = 'small' | 'big';
```
ValentinMumble, line 7:
```suggestion
type Size = 'small' | 'big' | 'inline';
```
I believe Stéphane said that the inline was supposed to be on the same existing `size` prop, as we cannot have a big & inline tile so size="big" with inline="true" would make no sense

ValentinMumble (approved):

StevenVAIDIE (commented):
How we are going to add the help icon ? 

Actually the TileContainer is not a flex component so the icon and the title are under each other but in the case of inline tile we should have in the same line the title and the icon

Screenshot from 2023-04-05 13-45-18

The text-align should not be center and we should have a fixed width in order to have a space between the title and the help icon

front-packages/akeneo-design-system/src/components/Tiles/Tiles.tsx
```diff
@@ -86,32 +109,39 @@ type TilesProps = {
   children?: ReactNode;
 
   /**
-   * The size can be 'small' (by default) or 'big'
+   * The size can be 'small' (by default), 'big' or 'inline'
    */
   size?: Size;
 };
 
 type TileProps = Override<
```
StevenVAIDIE, line 117: If it's only possible to do not have an icon in inline size we should report it in the type, something like that

```
type TileProps = Override<
  React.HTMLAttributes<HTMLDivElement>,
  (
    | {
      icon: React.ReactElement<IconProps>;
      size?: 'big' | 'small';
    }
    | {
      size?: 'inline';
      icon: undefined;
    }
  ) & {
    selected?: boolean;
    onClick?: () => void;
  }
>;
```

ghoussard (approved):

PR #19559, opened 2023-04-05
OCT-235: Display the default value in the target table

eoliveauakeneo:

(!) I created a TargetSourceAssociation folder to organize its sub-components  

image


components/catalogs/front/src/components/ProductMapping/components/TargetSourceAssociation/TargetSourceAssociation.tsx
```diff
@@ -41,19 +42,25 @@ export const TargetSourceAssociation: FC<Props> = memo(
                     {targetLabel ?? targetCode}
                     {isRequired && <RequiredPill level='warning' data-testid='required-pill' />}
                 </TargetCell>
-                {(null === source || null === source.source) && (
+                {null === source.source && (undefined === source.default || null === source.default) && (
```
RiffVentura, line 45:
```suggestion
                {null === source.source && undefined === source.default && (
```
since empty values are removed from the source so it can only be undefined or am I wrong?

RiffVentura (approved):

guillaumefer-akeneo (approved):

PR #19601, opened 2023-04-14
PIM-10938: Fix getNextObject when use clicks on variant during sequential edit

cemorin:

When a user has a product model in a sequential edit mode, and clicks on a variant of this product, the `getNextObject` returns the first element of the objectSet.

Now if the id of the product is not in the subjetSet, we check if it has a linked parent in the `objectSet`

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc


pierallard (approved):
:viandejs:

PR #19609, opened 2023-04-17
JEL-299: Dispatch ProductsWereCreatedOrUpdated events

LaurentPetard:
The `ProductWasUpdated` and `ProductWasCreated` were already implemented. New public event `ProductsWereCreatedOrUpdated` that contains a list of both classes

We dispatch it in a queue on save/bulkSave (for now it's the only way to handle legacy way and new way)

Add DQI consumer/handler in order to recompute product scores when a product is created/updated

The feature-flag `FLAG_DATA_QUALITY_INSIGHTS_UCS_EVENT_ENABLED` is temporary, until we implement and merge all the events needed to replace the current DQI job. No message will be dispatched until we remove it (or activate it)

EE: https://github.com/akeneo/pim-enterprise-dev/pull/17742


.env
```diff
@@ -39,13 +39,15 @@ PUBSUB_SUBSCRIPTION_WEBHOOK=
 PUBSUB_SUBSCRIPTION_SCHEDULED_JOB=
 PUBSUB_SUBSCRIPTION_PAUSED_JOB=
 PUBSUB_SUBSCRIPTION_DQI_LAUNCH_PRODUCT_EVALUATIONS=
+PUBSUB_SUBSCRIPTION_DQI_PRODUCT_SCORE_COMPUTE=dqi-product-score-compute
```
LaurentPetard, line 42: What do you think of `PUBSUB_SUBSCRIPTION_DQI_EVALUATE_PRODUCT_ON_UPSERT` instead ? 
It's more accurate than "score compute" (this wording is not used in DQI)

nmarniesse (approved):

Beninho (approved):

PR #19628, opened 2023-04-19
JEL-230: Add laziness to services to solve performance issue

LaurentPetard:
Some services can be costly when they are instantiated  (they trigger by cascade a lot of stuff)
So when they are used as dependency in a subscriber/listener that react to generic or common events, they can degrade the performances event if the subscriber/listener do nothing.

By doing so we can save 71% of the execution time of an empty Symfony command console. To give an idea of the impact on performances, on my laptop I went from executing 1.5 process per second to 5.2 process per second, which is huge.

Here the flame graphs of phpspy traces of the execution of looping on an empty Symfony command console:
Before:
traces-local-loop-master
After:
traces-local-loop-fix

### DIFF
--- a/src/Akeneo/UserManagement/back/Infrastructure/Cli/AuthenticateAdminUser.php
+++ b/src/Akeneo/UserManagement/back/Infrastructure/Cli/AuthenticateAdminUser.php
@@ -1,0 +1,55 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Akeneo\UserManagement\Infrastructure\Cli;
+
+use Akeneo\Tool\Component\StorageUtils\Factory\SimpleFactoryInterface;
+use Akeneo\UserManagement\Bundle\Security\SystemUserToken;
+use Akeneo\UserManagement\Component\Model\UserInterface;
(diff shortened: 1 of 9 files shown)


src/Akeneo/UserManagement/back/Infrastructure/Cli/AuthenticateAdminUser.php
```diff
@@ -0,0 +1,55 @@
...
+
+/**
+ * @copyright 2023 Akeneo SAS (https://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ *
+ * This service was initially a part of AuthenticateCommandAsAdminUserListener
+ * Its main purpose is to isolate the authentication details, to be able to make this service lazy
+ * So it will be instantiated only if needed (and so its dependencies)
+ */
+class AuthenticateAdminUser
```
LaurentPetard, line 22: The dependencies below can't be simply made lazy. So the solution is to deport them in a dedicated lazy service

LaurentPetard:
Moved to https://github.com/akeneo/pim-community-dev/pull/19877

PR #19650, opened 2023-04-24
GRF-829 : Fix Web API permissions are not taken into account for categories endpoints

jeremyBeaucousinAkeneo:
…gories endpoints

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc


src/Akeneo/Category/back/tests/EndToEnd/ExternalApi/CategoryAuthorizationIntegration.php
```diff
@@ -0,0 +1,298 @@
+<?php
+
+namespace Akeneo\Test\Category\EndToEnd\ExternalApi;
+
+use Akeneo\Test\Integration\Configuration;
+use Akeneo\Tool\Bundle\ApiBundle\Stream\StreamResourceResponse;
+use Akeneo\Tool\Bundle\ApiBundle\tests\integration\ApiTestCase;
```
amigaire, line 7: This class is used for EndToEnd tests in the whole project.
And the tests in this class seems to be more "EndToEnd tests" rather than "Integration tests".

I suggest to move this test class to the EndToEnd tests part and rename the class to end with "EndToEnd".
And then check the tests are still running and on success.

src/Akeneo/Category/back/Infrastructure/Controller/ExternalApi/ListCategoriesController.php
```diff
@@ -34,6 +37,11 @@ public function __construct(
      */
     public function __invoke(Request $request): JsonResponse|Response
     {
+        if ($this->securityFacade->isGranted('pim_api_category_list') === false) {
+            // even if this is a read endpoint, the user must be granted edition rights
+            // as this should only be used for the purpose of updating a category from the UI
```
amigaire, line 42: The comment does not match the value of the ACL:

> the user must be granted edition rights
VS

> 'pim_api_category_list'

When you have the right to list the categories, you can't edit a category, but you can see the list of category trees.

I suggest to correct the comment

src/Akeneo/Category/back/Infrastructure/Controller/ExternalApi/DownloadCategoryMediaFileController.php
```diff
@@ -29,6 +32,12 @@ public function __construct(
      */
     public function __invoke(string $code)
     {
+        if ($this->securityFacade->isGranted('pim_api_category_list') === false) {
+            // even if this is a read endpoint, the user must be granted edition rights
+            // as this should only be used for the purpose of updating a category from the UI
```
amigaire, line 37: The comment does not match the value of the ACL:

the user must be granted edition rights
VS

'pim_api_category_list'

When you have the right to list the categories, you can't edit a category, but you can see the list of category trees.

I suggest to correct the comment

src/Akeneo/Category/back/Infrastructure/Controller/ExternalApi/CategoryController.php
```diff
@@ -235,4 +247,13 @@ protected function validateCodeConsistency($code, array $data)
             throw new UnprocessableEntityHttpException(sprintf('The code "%s" provided in the request body must match the code "%s" provided in the url.', $data['code'], $code));
         }
     }
+
+    private function checkAclRights()
+    {
+        if ($this->securityFacade->isGranted('pim_api_category_edit') === false) {
+            // even if this is a read endpoint, the user must be granted edition rights
+            // as this should only be used for the purpose of updating a category from the UI
```
amigaire, line 255: The comment does not match the value of the ACL:

the user must be granted edition rights
VS

'pim_api_category_list'

When you have the right to list the categories, you can't edit a category, but you can see the list of category trees.

I suggest to correct the comment

amigaire (approved):

src/Akeneo/Category/back/Infrastructure/Controller/ExternalApi/CategoryController.php
pchasle: Yes, the comments are probably not necessary, `if isGranted('pim_api_category_edit') === false then throw new AccessDeniedException()` is explicit enough.

src/Akeneo/Category/back/Infrastructure/Controller/ExternalApi/CategoryController.php
```diff
@@ -78,6 +84,7 @@ public function __construct(
      */
     public function createAction(Request $request)
     {
+        $this->checkAclRights();
```
pchasle, line 87: Wait a minute ... the `@AclAncestor("pim_api_category_edit")` annotation should already do the ACL check, but it look like it's not working :thinking:

pchasle: Ok, I think that you need a specific config file to have the ACL annotation work when it's not an entity (see https://github.com/akeneo/pim-community-dev/blob/master/src/Oro/Bundle/SecurityBundle/Resources/doc/implementation.md)

I will recommend to remove all the `@AclAncestor` of the methods you updated @jeremyBeaucousinAkeneo and keep the check using the securityFacade service.

src/Akeneo/Category/back/tests/EndToEnd/ExternalApi/CategoryAuthorizationIntegration.php
jeremyBeaucousinAkeneo: I agree with you, concerning renaming it EndToEnd test. But when you say the whole project I do not get you.
I'm already in a namespace Akeneo\Test\Category\EndToEnd\ExternalApi, this class test is only covering the external routes "/api/rest*".

Where do you propose to move it ?

amigaire (approved):

pchasle (approved):

PR #19654, opened 2023-04-25
PIM-10949: Remove duplicated variant product

Beninho:

In order to fix data if exists duplicated variant product

- [X] Tests
- [X] Migration & Installer
- [ ] PM Validation (Story)
- [X] Changelog (maintenance bug fixes)
- [ ] Tech Doc

### DIFF
--- a/upgrades/schema/Version_8_0_20230425090000_remove_duplicated_variant_product.php
+++ b/upgrades/schema/Version_8_0_20230425090000_remove_duplicated_variant_product.php
@@ -1,0 +1,55 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Pim\Upgrade\Schema;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+
+/**
+ * @copyright 2023 Akeneo SAS (http://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+final class Version_8_0_20230425090000_remove_duplicated_variant_product extends AbstractMigration
+{
+    public function getDescription(): string
+    {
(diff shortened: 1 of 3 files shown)


upgrades/schema/Version_8_0_20230425090000_remove_duplicated_variant_product.php
```diff
@@ -0,0 +1,55 @@
...
+
+    public function down(Schema $schema): void
+    {
+        $this->throwIrreversibleMigrationException();
+    }
+
+    private function haveDuplicatedVariantProduct(): bool
+    {
+        $sql = <<<SQL
+SELECT   COUNT(*) AS nbr_doublon, product_model_id, raw_values
```
cemorin, line 45:
```suggestion
SELECT   COUNT(*) AS nbr_duplicate, product_model_id, raw_values
```
It's chipotage

cemorin (approved):

upgrades/schema/Version_8_0_20230425090000_remove_duplicated_variant_product.php
```diff
@@ -0,0 +1,55 @@
...
+            'No duplicated variant product'
+        );
+
+        $this->connection->executeStatement(<<<SQL
+DELETE t1
+FROM pim_catalog_product AS t1, pim_catalog_product AS t2
+WHERE t1.id > t2.id
+  AND t1.product_model_id = t2.product_model_id
+  AND t1.raw_values = t2.raw_values;
+SQL);
```
mmetayer, line 34: there's no index on raw_values, this query will take too long to execute, plus duplicate variant products doesn't necessarily means that the raw_values are exactly the same

mmetayer (changes requested):
the pb is too complex to be executed in a doctrine migration, let's try to find another way to fix it

PR #19675, opened 2023-05-01
GRF-828: fix category tree deletion

pchasle:
Fix **SLA PIM-10956 Deletion of category once template activated throwing 500**

Use case 1:
1. Activate template
1. Deactivate template
1. Re-Activate template
1. Delete Category Tree → 500 Error on Foreign Key constraint

Use case 2:
1. Activate template
1. Remove children category of the tree → 500 Error

### Causes

Since deactivating a template doesn’t delete it (soft delete) there is 2 templates linked to the category tree, so trying to delete the template doesn’t work until the deactivated one is removed too.
Once the job that cleanup categories values & remove the deactived template is done the category can be removed so this is not critical since user can just wait & retry.

The issue was that the category deletion code didn't took deactivated template into account.

### Suggested fix
Simplify the template deletion when a category tree is removed and just delete all templates, attributes and associations between category tree and templates before.

- Remove `RemoveCategoryTreeTemplateSubscriber` that listen to `PRE_REMOVE`
- Remove `RemoveTemplateAndAttributesAfterCategoryTreeDeletionSubscriber` that listen to `POST_REMOVE`
- Create a dedicated Command+Handler to delete a category tree and the associated templates


jeremyBeaucousinAkeneo (approved):

etiennekerbastard (approved):

mmetayer (approved):
approved for the enrichment part (BTW I guess this controller should be moved to the Category namespace)

PR #19697, opened 2023-05-09
JEL: Refacto message with a single transport

LaurentPetard:
The consumers daemons will be launched specifying the consumer name ("queue" in Symfony Messenger word)
Ex: `./bin/console messenger:consume akeneo_message --queues=dqi_launch_product_and_product_model_evaluations_consumer`


LaurentPetard:
Change branch

PR #19702, opened 2023-05-10
RAB: Allow override Makefile env variables

ValentinMumble:
RAB: Allow override Makefile env variables so that yarn can be run locally for instance:

`YARN_RUN=yarn make front`

Using the native yarn greatly improves performances on Mac machines

YARN_RUN env var can also be exported in a bash config file to avoid using it in every command, in a .zsh_rc or .bash_rc:

`export YARN_RUN=yarn`


Makefile
```diff
@@ -1,8 +1,8 @@
-DOCKER_COMPOSE = docker-compose
-NODE_RUN = $(DOCKER_COMPOSE) run -u node --rm -e YARN_REGISTRY -e PUPPETEER_SKIP_CHROMIUM_DOWNLOAD=1 -e PUPPETEER_EXECUTABLE_PATH=/usr/bin/google-chrome node
-YARN_RUN = $(NODE_RUN) yarn
-PHP_RUN = $(DOCKER_COMPOSE) run --rm php php
-PHP_EXEC = $(DOCKER_COMPOSE) exec -u www-data httpd php
+DOCKER_COMPOSE ?= docker-compose
```
StevenVAIDIE, line 1: Same here

StevenVAIDIE (approved):

ghoussard (approved):

PR #19718, opened 2023-05-15
fix stateful UcsMiddleware injecting wrong tenant id in messages

tseho:

In the previous version of `UcsMiddleware`, the property `$this->pimTenantId`, from `APP_TENANT_ID` env var, was used for adding the tenant id into messages.
In long running processes, this variable was supposed to stay null after each envelope.
That was not the case because after the first message is consumed, it becomes stateful because of the line
```
$this->pimTenantId = $stamp->pimTenantId();
```

With the new version, the middleware never become stateful in long-running processes and always use the stamp as a source of truth.

Since the webhook feature has no visible impact on the PIM (it's a background process that does not change PIM data), it took a lot of time to be noticed.

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/src/Akeneo/Tool/Bundle/MessengerBundle/Middleware/UcsMiddleware.php
+++ b/src/Akeneo/Tool/Bundle/MessengerBundle/Middleware/UcsMiddleware.php
@@ -22,23 +22,16 @@ class UcsMiddleware implements MiddlewareInterface
 
     public function handle(Envelope $envelope, StackInterface $stack): Envelope
     {
-        // No tenant ID in the env, but existing in the received message
-        // We are in a daemon long-running process
-        if (null === $this->pimTenantId && null !== $envelope->last(TenantIdStamp::class)) {
-            /** @var TenantIdStamp $stamp */
-            $stamp = $envelope->last(TenantIdStamp::class);
-            $this->pimTenantId = $stamp->pimTenantId();
+        // We always try to use the tenantid from the stamp, if there is any, in case of long-running process.
+        // If there is none, we fallback on the tenantid coming from the env variables.
+        $tenantId = $envelope->last(TenantIdStamp::class)?->pimTenantId() ?: $this->pimTenantId;
+
+        if ($tenantId && null === $envelope->last(TenantIdStamp::class)) {
+            $envelope = $envelope->with(new TenantIdStamp($tenantId));
         }
 
-        // Tenant ID exists in this env
-        // We are in a contextualized process
-        if ($this->pimTenantId && null === $envelope->last(TenantIdStamp::class)) {
-            $envelope = $envelope->with(new TenantIdStamp($this->pimTenantId));
-        }
-
-        // Enrich the message with the tenant ID
-        if ($this->pimTenantId && $envelope->getMessage() instanceof TenantAwareInterface) {
-            $envelope->getMessage()->setTenantId($this->pimTenantId);
+        if ($tenantId && $envelope->getMessage() instanceof TenantAwareInterface) {
+            $envelope->getMessage()->setTenantId($tenantId);
         }
 
         return $stack->next()->handle($envelope, $stack);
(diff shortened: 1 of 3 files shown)


src/Akeneo/Tool/Bundle/MessengerBundle/Middleware/UcsMiddleware.php
```diff
@@ -16,29 +17,31 @@
...
-        // We are in a daemon long-running process
-        if (null === $this->pimTenantId && null !== $envelope->last(TenantIdStamp::class)) {
-            /** @var TenantIdStamp $stamp */
-            $stamp = $envelope->last(TenantIdStamp::class);
-            $this->pimTenantId = $stamp->pimTenantId();
+        // We always try to use the tenantid from the stamp, if there is any, in case of long-running process.
+        // If there is none, we fallback on the tenantid coming from the env variables.
+        $tenantId = $envelope->last(TenantIdStamp::class)?->pimTenantId() ?: $this->pimTenantId;
+
+        if (null === $tenantId) {
```
ahocquard, line 32: could it be "" also?

src/Akeneo/Tool/Bundle/MessengerBundle/Middleware/UcsMiddleware.php
```diff
@@ -16,29 +17,31 @@
...
+                'A message of type "%s" is consumed without a tenant id available',
+                get_class($envelope->getMessage()),
+            ));
         }
 
-        // Tenant ID exists in this env
-        // We are in a contextualized process
-        if ($this->pimTenantId && null === $envelope->last(TenantIdStamp::class)) {
-            $envelope = $envelope->with(new TenantIdStamp($this->pimTenantId));
+        if ($tenantId && null === $envelope->last(TenantIdStamp::class)) {
```
ahocquard, line 39:
```suggestion
        if (null === $envelope->last(TenantIdStamp::class)) {
```

src/Akeneo/Tool/Bundle/MessengerBundle/Middleware/UcsMiddleware.php
```diff
@@ -16,29 +17,31 @@
...
+            $envelope = $envelope->with(new TenantIdStamp($tenantId));
         }
 
-        // Enrich the message with the tenant ID
-        if ($this->pimTenantId && $envelope->getMessage() instanceof TenantAwareInterface) {
-            $envelope->getMessage()->setTenantId($this->pimTenantId);
+        if ($tenantId && $envelope->getMessage() instanceof TenantAwareInterface) {
```
ahocquard, line 43:
```suggestion
        if ($envelope->getMessage() instanceof TenantAwareInterface) {
```

ahocquard (approved):

src/Akeneo/Tool/Bundle/MessengerBundle/Middleware/UcsMiddleware.php
tseho: `$tenantId` can be null.

src/Akeneo/Tool/Bundle/MessengerBundle/Middleware/UcsMiddleware.php
```diff
@@ -16,29 +17,31 @@
...
+        // We always try to use the tenantid from the stamp, if there is any, in case of long-running process.
+        // If there is none, we fallback on the tenantid coming from the env variables.
+        $tenantId = $envelope->last(TenantIdStamp::class)?->pimTenantId() ?: $this->pimTenantId;
+
+        if (null === $tenantId) {
+            $this->logger->warning(sprintf(
+                'A message of type "%s" is consumed without a tenant id available',
+                get_class($envelope->getMessage()),
+            ));
         }
```
jmleroux, line 37: I am still thinking about valid use case when we could not need the tenant ID.
We may simply have to crash if no tenant ID is resolved.

jmleroux: Ah... Unless for Serenity backward compatibility...

jmleroux (approved):

docteurklein (approved):

src/Akeneo/Tool/Bundle/MessengerBundle/Middleware/UcsMiddleware.php
docteurklein: early return maybe, but not before calling `stack->next()`

PR #19737, opened 2023-05-17
CPM-1064

Beninho:

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc


upgrades/schema/Version_8_0_20230412073848_add_main_attribute_column.php
```diff
@@ -0,0 +1,38 @@
...
+final class Version_8_0_20230412073848_add_main_attribute_column extends AbstractMigration
+{
+    public function getDescription(): string
+    {
+        return 'Adds the is_main_identifier column';
+    }
+
+    public function up(Schema $schema): void
+    {
+        $this->skipIf(
```
mmetayer, line 23: we don't want to skip the migration if the column already exists, or else doctrine will try to run it again later. Here we just want to mark it as done (an early return is enough)

upgrades/schema/Version_8_0_20230412073848_add_main_attribute_column.php
```diff
@@ -0,0 +1,38 @@
...
+            $schema->getTable('pim_catalog_attribute')->hasColumn('main_identifier'),
+            'main_identifier column already exists in pim_catalog_attribute'
+        );
+
+        $this->addSql('ALTER TABLE pim_catalog_attribute ADD main_identifier TINYINT(1) NOT NULL DEFAULT FALSE;');
+        $this->addSql(
+            'UPDATE pim_catalog_attribute SET main_identifier = TRUE WHERE code = \'sku\';'
```
mmetayer, line 30: instead of hard coded sku, we should use:
- the only attribute with type 'pim_catalog_identifier'
- if there are several (although unlikely), use the first one

upgrades/test_schema/Version_8_0_20230412073848_add_main_attribute_column_Integration.php
```diff
@@ -0,0 +1,63 @@
...
+            $this->connection->executeStatement(
+                <<<SQL
+            ALTER TABLE pim_catalog_attribute DROP COLUMN main_identifier
+            SQL
+            );
+        }
+
+        Assert::assertFalse($this->mainIdentifierColumnExists());
+        $this->reExecuteMigration(self::MIGRATION_NAME);
+        Assert::assertTrue($this->mainIdentifierColumnExists());
```
mmetayer, line 49: please also add an assertion checking that one trow has the "main_identifier" set to true

antoinearrive (approved):

mmetayer (commented):
Please add a title to the PR :wink:

mmetayer (commented):
Please add a title to the PR :wink:

Balamung (approved):

mmetayer (approved):

PR #19738, opened 2023-05-17
PIM-10978 Make sure to change updated date on edited category

amigaire:

An anticorruption layer has been introduced in the Pim\Enrichment part of the code to be able to use the service to update the category "updated" column during categories import.
We implemented an ACL because the code for the import of categories will be reworked in the future (i.e. it will be part of the new Bounded Context Category, hence no ACL needed anymore) and because a service api is not a desired solution: this updating service is very specific to Category BC and should not be callable outside of the BC. 

### DIFF
--- a/src/Akeneo/Pim/Enrichment/Component/Category/CategoryUpdater.php
+++ b/src/Akeneo/Pim/Enrichment/Component/Category/CategoryUpdater.php
@@ -2,6 +2,7 @@
 
 namespace Akeneo\Pim\Enrichment\Component\Category;
 
+use Akeneo\Pim\Enrichment\Product\Infrastructure\AntiCorruptionLayer\ACLUpdateCategoryUpdatedDate;
 use Akeneo\Tool\Component\Localization\Model\TranslatableInterface;
 use Akeneo\Tool\Component\Localization\TranslatableUpdater;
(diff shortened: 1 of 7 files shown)


src/Akeneo/Pim/Enrichment/Component/Category/CategoryUpdater.php
```diff
@@ -40,6 +34,8 @@ public function update($category, array $data, array $options = [])
 
         if (isset($data['labels']) && $category instanceof TranslatableInterface) {
             $this->translatableUpdater->update($category, $data['labels']);
+            // PIM-10978: category's date of update must be changed when translations are updated
+            $this->updateCategoryUpdatedDate->execute($category->getCode());
```
mmetayer, line 38: This directly updates the category in the DB, which is not the purpose of this service. Here, if the category is invalid, and the changes are rolled back, the update date will still be updated in the DB

mmetayer (commented):
This should already be handled by [this subscriber](https://github.com/akeneo/pim-community-dev/blob/master/src/Akeneo/Pim/Enrichment/Bundle/EventSubscriber/Category/OnSave/SetUpdatedPropertyOnTranslationUpdateSubscriber.php). A regression may have been introduced, I'd suggest investigating this class

PR #19792, opened 2023-05-29
GRF: Cleanup user group controller & query

pchasle:
Linked to this issue https://akeneo.atlassian.net/browse/GRF-752 and this SDS https://akeneo.atlassian.net/browse/SDS-28363

- Removed the default value from the user group query as it's already defined on the SQL query.
- Define the user group limit value of `1000` at the controller level since this is an issue with this route/ui part only and should not impact the bounded context ServiceApi.

### DIFF
--- a/src/Akeneo/UserManagement/back/Infrastructure/Storage/SqlFindUserGroups.php
+++ b/src/Akeneo/UserManagement/back/Infrastructure/Storage/SqlFindUserGroups.php
@@ -61,16 +61,16 @@ class SqlFindUserGroups implements FindUserGroups
         $sqlWherePart = empty($sqlWhereParts) ? '' : 'AND '.implode(' AND ', $sqlWhereParts);
 
         return <<<SQL
-            SELECT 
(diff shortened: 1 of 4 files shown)


src/Akeneo/UserManagement/back/Infrastructure/Storage/SqlFindUserGroups.php
```diff
@@ -61,16 +61,16 @@ private function buildQuery(
...
+            SELECT
                 oag.id,
-                oag.name, 
-                oag.type, 
+                oag.name,
+                oag.type,
                 oag.default_permissions
             FROM oro_access_group oag
             WHERE oag.type = 'default'
-            ${sqlWherePart}
```
pchasle, line 71: `${var}` string interpolation deprecated since PHP 8.2

jeremyBeaucousinAkeneo (approved):

amigaire (approved):

PR #19828, opened 2023-06-06
RAB-1425: Resume import job

MadeWilson:

We made every import job able to resume at the position where they are paused.
To do so, we needed to backup the import file at the pause and to recover it at the resume.

**Backuper**

As the `JobFileBackuper` is already present since the pause export PR, we refactored it a bit to add the `recover` method and to not need to store the backup file path. As every backup file is stored in job execution directory in the archivist, only the local file path is enough to retrieve it. And this local file path does not change between a pause and a resume because it is saved in the current state (export) or in the job parameters (import).

**StatefulInterface**

We got rid of `PauseableReaderInterface` and `PausableWriterInterface` because they leaded to a design mess. Every database writer is stateless, that means we needed to dummy implement `getState` and `rewindToState` on these writers. So, we created a StatefulInterface which expose a method to set the state and get the current state from a step element. Every reader (database and file) implements it and every file writer implements it. We updated the script to detect not pausable item step according to this design change.

**Rewind to state**

As we don't have a `rewindToState` method anymore, we made the choice to implement this logic inside the initialize method.

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc


src/Akeneo/Tool/Bundle/BatchBundle/EventListener/BackupImportFileOnPauseSubscriber.php
```diff
@@ -0,0 +1,44 @@
...
+        $stepExecution = $event->getStepExecution();
+
+        if ($stepExecution->getJobExecution()->getJobInstance()->getType() !== 'import') {
+            return;
+        }
+
+        $fileToBackUp = new JobFileLocation($stepExecution->getJobParameters()->get('storage')['file_path'], true);
+
+        $backupFilePath = 'paused_job/step/' . $stepExecution->getId();
+        $this->filesystemOperator->write($backupFilePath, file_get_contents($fileToBackUp->path()));
```
ghoussard, line 42: IMO this logic should be encapsulated in a service which is responsible to backup and restore the file. Then we can just call it both event listeners.
Also, we should re-discuss where we store this backuped file. Here it is stored in the jobs storage, in the export PR it is stored in the archive storage 🤔

ghoussard, line 42: IMO this logic should be encapsulated in a service which is responsible to backup and restore the file. Then we can just call it both event listeners.
Also, we should re-discuss where we store this backuped file of if we really want to backup it. If I remember well, we said "Don't backup the file, it's already stored on a persisted storage (jobs storage in this case), just get it and re-validate it" 🤔

src/Akeneo/Tool/Component/Batch/Event/EventInterface.php
```diff
@@ -27,6 +27,8 @@ interface EventInterface
     const STEP_EXECUTION_ERRORED = 'akeneo_batch.step_execution_errored';
     const STEP_EXECUTION_COMPLETED = 'akeneo_batch.step_execution_completed';
     const INVALID_ITEM = 'akeneo_batch.invalid_item';
+    const BEFORE_STEP_EXECUTION_PAUSED = 'akeneo_batch.step_execution_paused';
```
ghoussard, line 30:
```suggestion
    const BEFORE_STEP_EXECUTION_PAUSED = 'akeneo_batch.before_step_execution_paused';
```

src/Akeneo/Tool/Component/Batch/Item/PausableReaderInterface.php
```diff
@@ -12,4 +12,6 @@
 interface PausableReaderInterface
 {
     public function getState(): array;
+
+    public function rewindToState(int $key): void;
```
ghoussard, line 16: Rewind from a key is an implementation detail, we should use the state as param
```suggestion
    public function rewindToState(array $state): void;
```

ghoussard (commented):
Could you schedule a meeting with the team to mob-review this PR please?

src/Akeneo/Tool/Bundle/BatchBundle/EventListener/BackupImportFileOnPauseSubscriber.php
ghoussard: Seen IRL : keep backup

src/Akeneo/Tool/Component/Batch/Job/Job.php
```diff
@@ -205,6 +206,10 @@ protected function doExecute(JobExecution $jobExecution)
                 continue;
             }
 
+            if (null !== $stepExecution && $stepExecution->getStatus()->getValue() === BatchStatus::PAUSED) {
+                $stepExecution->setExecutionContext(new ExecutionContext());
+            }
```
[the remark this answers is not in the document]
ghoussard: Initialize it in getStepExecution

src/Akeneo/Tool/Bundle/BatchBundle/EventListener/MountFileOnResumeSubscriber.php
```diff
@@ -0,0 +1,47 @@
...
+    ) {
+    }
+    public static function getSubscribedEvents(): array
+    {
+        return [
+            EventInterface::BEFORE_STEP_EXECUTION_RESUME => 'beforeStepExecutionResume'
+        ];
+    }
+
+    public function beforeStepExecutionResume(StepExecutionEvent $event): void
```
ghoussard, line 31: TODO: se brancher si le hook qui telecharge les fichiers depuis un remote storage

ghoussard: This hook is now a step so if we don't download the file twice when we resume the job

jeremyBeaucousinAkeneo (approved):

src/Akeneo/Tool/Component/Batch/Step/AbstractStep.php
```diff
@@ -71,6 +68,10 @@ final public function execute(StepExecution $stepExecution)
     {
         $this->dispatchStepExecutionEvent(EventInterface::BEFORE_STEP_EXECUTION, $stepExecution);
 
+        if ($stepExecution->getStatus()->getValue() === BatchStatus::PAUSED) {
+            $this->dispatchStepExecutionEvent(EventInterface::BEFORE_STEP_EXECUTION_RESUME, $stepExecution);
```
pchasle, line 72: Any reason for dispatching and event to handle the backup/recover in a separate listener and not do it there directly?

pchasle (approved):

src/Akeneo/Tool/Component/Batch/Step/AbstractStep.php
MadeWilson: Most of the logic not directly linked to the step execution is handled throught listener in this class, this was in a sens of continuity with what has been already written.

PR #19831, opened 2023-06-07
RAB-1416: Save current state when an export job is paused

marinfazioakeneo:

In order to pause export jobs, we implemented `PauseableReaderInterface` on database readers and `PauseableWriterInterface` on file writers. That means :
- We save the last item read for each readers ;
- We backup exported file and/or buffer and we save the backup path for each writer.
  - When a file writer writes multiple files such as medias or splitted XLSX files, it is handled by design because they are saved in the job execution directory in the archive storage. We just save references of these files to avoid rewriting it when we resume.

We created a service which is responsible to backup a file in the archive directory of the job execution.

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc


amigaire (approved):

src/Akeneo/Tool/Component/Connector/Writer/File/ExportedFileBackuper.php
```diff
@@ -0,0 +1,38 @@
...
+
+use Akeneo\Tool\Component\Batch\Model\JobExecution;
+use League\Flysystem\FilesystemOperator;
+
+class ExportedFileBackuper
+{
+    private const BACKUP_DIR = 'backup';
+
+    public function __construct(
+        private FilesystemOperator $filesystemOperator,
```
mmetayer, line 15: a FilesystemWriter would be enough here IMO

mmetayer (approved):

pierrejolly (approved):

PR #19837, opened 2023-06-08
PH-1083: :bug: Scheduled job lock must not auto release

jmleroux:

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/src/Akeneo/Tool/Component/Connector/Step/CheckJobUniqueStep.php
+++ b/src/Akeneo/Tool/Component/Connector/Step/CheckJobUniqueStep.php
@@ -45,7 +45,7 @@ class CheckJobUniqueStep extends AbstractStep
         Assert::notEmpty($jobCode, 'The job code must not be empty');
 
         $lockIdentifier = sprintf('scheduled-job-%s', $jobCode);
-        $lock = $this->lockFactory->createLock($lockIdentifier, static::LOCK_TTL_IN_SECONDS);
+        $lock = $this->lockFactory->createLock($lockIdentifier, static::LOCK_TTL_IN_SECONDS, false);
 
         if (!$lock->acquire()) {
             $jobExecution = $stepExecution->getJobExecution();


src/Akeneo/Tool/Bundle/VersioningBundle/Command/PurgeCommand.php
```diff
@@ -136,10 +136,7 @@ public function interact(InputInterface $input, OutputInterface $output)
      */
     protected function execute(InputInterface $input, OutputInterface $output): int
     {
-        $noDebug = $input->getOption('no-debug');
```
jmleroux, line 139: this option does not exists

docteurklein: I think it does, (default behavior of symfony commands), and reflects `--no-debug` which is always present for every command

src/Akeneo/Platform/Bundle/ImportExportBundle/Repository/InternalApi/JobExecutionRepository.php
```diff
@@ -50,4 +52,22 @@ public function createDatagridQueryBuilder()
 
         return $qb;
     }
+
+    public function isOtherJobExecutionRunning(JobExecution $jobExecution): bool
+    {
+        $sql = <<< SQL
+        SELECT EXISTS(
```
ahocquard, line 59: if the rule fail for any reason (OOM kill, etc), the next rule will never be executed, right? That's why we initially chose a lock.

jmleroux: Indeed but:
- we probably don't want to have several false "STARTED" job execution and we may think about something to fix the crashed ones
- this PR is a hotfix that can be improved in the future. Maybe reintroduce a better lock or find another way.

jmleroux: My first tries on a customized Lock were not su soccesful and we need something to fix the current situation.

ahocquard (approved):
I GTM for fixing stuff quickly. But to rework to be:
- robust if a job crashes
- atomic

PR #19885, opened 2023-06-16
GRF-845: edit template properties form

pchasle:
image


src/Akeneo/Category/front/src/feature/components/datagrids/CategoryTreeDataGrid.tsx
```diff
@@ -14,9 +15,27 @@ import styled from 'styled-components';
 import {NoResults} from './NoResults';
 import {DeleteCategoryModal} from './DeleteCategoryModal';
 import {deleteCategory} from '../../infrastructure';
-import {createTemplate} from '../templates/createTemplate';
 import {useCountCategoryTreesChildren} from '../../hooks';
 
+const createTemplate = async (categoryTree: CategoryTreeModel, catalogLocale: string, router: Router) => {
```
jeremyBeaucousinAkeneo, line 20: Why reimport it at template level ?

I know this is the only way to create a template for now, but I'm more used to creating a hook for a backend call.

pchasle: This function was in `components/templates/createTemplate.tsx` but it's not a component and should probably be a custom hooks.

Also it's only used in this file that is `components/datagrids/CategoryTreeDataGrid.tsx`, so a completely different part of application / folder structure.

I moved it here to make it private to this component for now. It will be probably removed / rewritten next week with the new "load predefined attributes" feature.

jeremyBeaucousinAkeneo (approved):

src/Akeneo/Category/front/src/feature/components/datagrids/CategoryTreeDataGrid.tsx
jeremyBeaucousinAkeneo: OK for me thanks ;).

PR #19893, opened 2023-06-19
PIM-11050: On "locale" DIV, missing the indication of the country

jpa44:
On DQI locale filter the display was not the same on the other locale filter in the PIM
i fixed by changing the locale language to the locale label in the locale switcher template

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/src/Akeneo/Platform/Bundle/UIBundle/Resources/public/templates/product/locale-switcher.html
+++ b/src/Akeneo/Platform/Bundle/UIBundle/Resources/public/templates/product/locale-switcher.html
@@ -1,13 +1,13 @@
 <a class="AknActionButton <% if (displayInline) { %> AknActionButton--big AknActionButton--noLeftBorder<% } else { %>AknActionButton--withoutBorder<% } %>" data-toggle="dropdown">
     <% if (!displayInline && displayLabel) { %><%- label %>: <% } %>
-    <span class="AknActionButton-highlight"><%= i18n.getFlag(currentLocale.code, false) %> <%- currentLocale.language %></span>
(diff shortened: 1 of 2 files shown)


src/Akeneo/Platform/Bundle/UIBundle/Resources/public/templates/product/locale-switcher.html
```diff
@@ -1,13 +1,13 @@
 <a class="AknActionButton <% if (displayInline) { %> AknActionButton--big AknActionButton--noLeftBorder<% } else { %>AknActionButton--withoutBorder<% } %>" data-toggle="dropdown">
     <% if (!displayInline && displayLabel) { %><%- label %>: <% } %>
-    <span class="AknActionButton-highlight"><%= i18n.getFlag(currentLocale.code, false) %> <%- currentLocale.language %></span>
+    <span class="AknActionButton-highlight"><%= i18n.getFlag(currentLocale.code, false) %> <%- currentLocale.label %></span>
```
nmarniesse, line 3: In the grid, there is a different code: `<%- getDisplayName(currentLocale) %>`
Did you try it? Which one is better? If both are working I would prefer having the same code everywhere

https://github.com/akeneo/pim-community-dev/blob/master/src/Akeneo/Platform/Bundle/UIBundle/Resources/public/templates/product/grid/locale-switcher.html#L3

jpa44: Yes i tried to use this function but it throw undefined error
it seem that product grid locale switcher is not used

nmarniesse (approved):
GTM once the changelog is updated

PR #19894, opened 2023-06-19
GROWTH-165: Set feature flags on route_complete event

BryanS29-akeneo:

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/src/Akeneo/Platform/Bundle/UIBundle/Resources/public/js/feature-flags.ts
+++ b/src/Akeneo/Platform/Bundle/UIBundle/Resources/public/js/feature-flags.ts
@@ -1,6 +1,7 @@
 import jQuery from 'jquery';
 
 const routing = require('routing');
+const Mediator = require('oro/mediator');
 
 class FeatureFlags {
   static features: {[feature: string]: boolean};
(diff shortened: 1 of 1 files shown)


src/Akeneo/Platform/Bundle/UIBundle/tests/front/unit/feature-flags.unit.ts
```diff
@@ -0,0 +1,48 @@
...
+
+    test('it should throw an error if FeatureFlags is already initialized', async () => {
+      FeatureFlags.features = {};
+
+      await expect(FeatureFlags.initialize()).rejects.toThrowError('FeatureFlags is already initialized.');
+    });
+  });
+
+  describe('isEnabled', () => {
+    test('it should throw an error if FeatureFlags is not initialized', () => {
```
fchoquet, line 35: A feature flag mechanism should never throw an exception or trigger an error. When the flag can't be resolved it should fall back to a default value.

fchoquet (commented):
See my remark. I know this is not directly related to your change and that you are just adding tests to an existing behavior, but I think it's important to fix it.

(also, I don't see the big picture, so maybe these exceptions are catched)

PR #19900, opened 2023-06-20
CPM-1083: Use IdentifierValue in SetIdentifierSubscriber

cemorin:

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc


components/identifier-generator/back/src/Infrastructure/Subscriber/SetIdentifiersSubscriber.php
```diff
@@ -109,8 +110,8 @@ private function setGeneratedIdentifier(
         $command = GenerateIdentifierCommand::fromIdentifierGenerator($identifierGenerator, $productProjection);
         $newIdentifier = ($this->generateIdentifierCommandHandler)($command);
 
-        $value = ScalarValue::value($identifierGenerator->target()->asString(), $newIdentifier);
-        Assert::isInstanceOf($value, ScalarValue::class);
+        $value = IdentifierValue::value($identifierGenerator->target()->asString(), true, $newIdentifier);
```
mmetayer, line 113: would be best to have the real value of `$isMainIdentifier` here (as of now it's indeed always `true`). Can you add a TODO and create a card plz?

mmetayer (approved):

Beninho (approved):

PR #19906, opened 2023-06-21
PIM-11039: Fix export with duplicated labels

mmetayer:

When exporting products or models with labels, the column presenter was used to generate column names for both the header row and "data" rows, which in some cases could lead to some inconsistencies, because "data" rows don't necessarily have values for each column. As a result, some values could be "out of bounds" (at the end of the row, in a column without header)
Now the column presenter returns a mapping `[original_column_name => translated_column_name]`, which is then used to prepare proper data rows

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/src/Akeneo/Tool/Component/Connector/Writer/File/FlatItemBufferFlusher.php
+++ b/src/Akeneo/Tool/Component/Connector/Writer/File/FlatItemBufferFlusher.php
@@ -84,22 +84,21 @@ class FlatItemBufferFlusher implements StepExecutionAwareInterface
         $headers = $this->sortHeaders($buffer->getHeaders());
         $headers = $this->columnPresenter->present($headers, $this->stepExecution->getJobParameters()->all());
 
-        $hollowItem = array_fill_keys($headers, '');
-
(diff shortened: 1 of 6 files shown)


src/Akeneo/Tool/Component/Connector/Writer/File/FlatItemBufferFlusher.php
```diff
@@ -153,14 +151,14 @@ protected function writeIntoSeveralFiles(
...
-            );
-
-            $incompleteItem = array_combine($incompleteKeys, $incompleteItem);
-
-            $item = array_replace($hollowItem, $incompleteItem);
+            $item = array_fill_keys($headers, '');
+            foreach ($incompleteItem as $key => $value) {
+                $headerName = $headers[$key] ?? null;
+                if (null === $headerName) {
+                    continue;
```
mmetayer, line 158: this should NEVER happen, as headers contain every column name

Beninho (approved):

PR #19928, opened 2023-06-22
CPM-1100: Update queries with link to identifier product column

pierallard:
This PR prepares the moment when we will drop the `identifier` column from `pim_catalog_product` table.
Now, the identifier can be found in the `pim_catalog_product_unique_data` table, in the `raw_data` column, where the attribute is the main one.

We updated 2 types of queries:
- The ones where we don't need products without identifiers (in these cases, a INNER JOIN is done)
- The ones where we need NULL value as identifier if product exists but has no identifier (in these cases, a LEFT JOIN is done).

We manually removed the `identifier` column, then try to execute all the integration/endtoend tests. Majority was green, except some ones with validation or product save.


mmetayer (approved):

src/Akeneo/Pim/Enrichment/Bundle/Doctrine/ORM/Repository/ProductRepository.php
```diff
@@ -23,19 +24,29 @@ class ProductRepository extends EntityRepository implements
...
+            SELECT BIN_TO_UUID(product_uuid)
+            FROM pim_catalog_product_unique_data
+            WHERE attribute_id = :attributeId
+            AND raw_data IN (:identifiers)
+            SQL,
+            [
+                'identifiers' => $identifiers,
+                'attributeId' => $this->getMainIdentifierId(),
+            ], [
+                'identifiers' => Connection::PARAM_STR_ARRAY,
```
mmetayer, line 45:
```suggestion
                'identifiers' => Connection::PARAM_STR_ARRAY,
                'attributeId' => ParameterType::INTEGER,
```
(probably better for using the index)

cemorin (approved):

mmetayer (approved):

Balamung (approved):

Beninho (approved):

PR #19948, opened 2023-06-26
RAB-1456: DatabaseCommand should not rewrite pim_configuration table

ghoussard:

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc


src/Akeneo/Platform/Installer/back/src/Infrastructure/Symfony/Resources/config/services.yml
```diff
@@ -21,22 +21,29 @@ services:
...
+      - { name: akeneo.installer.user_configuration_resetter }
 
   Akeneo\Platform\Installer\Infrastructure\UserConfigurationResetter\UserUiLocaleResetter:
     arguments:
       - '@database_connection'
     tags:
-      - {name: akeneo.installer.user_configuration_resetter}
+      - { name: akeneo.installer.user_configuration_resetter }
+
+  akeneo.installer.doctrine.pim_configuration_schema_asset_filter:
```
StevenVAIDIE, line 44: What is the propose of this ? Who delete this table ?

ghoussard: We run `doctrine:schema:update` with `--complete` option so this command will remove any not mapped asset in the database (pim_configuration included). This is a way to declare filter which exclude some asset like it is done in the [doctrine bundle](https://github.com/doctrine/DoctrineBundle/blob/2.10.x/DependencyInjection/Compiler/WellKnownSchemaFilterPass.php).

StevenVAIDIE: I will preferred having it in the doctrine.yml configuration (cf: https://symfony.com/bundles/DoctrineMigrationsBundle/current/index.html#manual-tables) but problem is that we should have to declare it in EE and CE.

Problem with regex is that we didn't have a real way to do good regex, for example your regex will filter `pim_configuration_another_name` table if we want to do it we should use the following regex: ```^(?!pim_configuration$)\w+$```

Another solution is to copy the service named BlacklistSchemaAssetFilter (because it's deprecated) in order to check it with strict comparison. As you want

StevenVAIDIE (approved):

pierrejolly (approved):

PR #19957, opened 2023-06-27
JEL-299: add mechanism to ack messages before product score computing

nmarniesse:

Because computing product score can be long, we ack the message to not reach the limit of 10 minutes for the execution time.
The execution time can change a lot, depending on the number of attributes in the product, the number of locales, channels, ...

If the limit of 10 minutes is reached, the message is put back in the queue. So new score computing will be triggered for nothing.

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc


src/Akeneo/Pim/Automation/DataQualityInsights/tests/back/EndToEnd/ComputeProductScoreOnProductCreateOrUpdateEndToEnd.php
```diff
@@ -0,0 +1,69 @@
...
+
+use Akeneo\Pim\Automation\DataQualityInsights\Domain\ValueObject\ProductUuid;
+use Akeneo\Pim\Enrichment\Component\Product\Model\ProductInterface;
+use Ramsey\Uuid\Uuid;
+
+/**
+ * @copyright 2023 Akeneo SAS (https://www.akeneo.com)
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+final class ComputeProductScoreOnProductCreateOrUpdateEndToEnd extends MessengerTestCase
```
LaurentPetard, line 15: In my opinion it's more an integration test on Messenger than a "end to end" test.

src/Akeneo/Pim/Enrichment/Product/back/Application/UpsertProductHandler.php
```diff
@@ -77,10 +79,11 @@ public function __invoke(UpsertProductCommand $command): void
         $isUpdate = $product->isDirty();
         $this->productSaver->save($product);
 
+        $date = $this->clock->now();
```
LaurentPetard, line 82: Why not use `$product->getUpdated()` and `$product->getCreated()` to get the exact date ?

LaurentPetard (approved):

src/Akeneo/Pim/Enrichment/Product/back/Application/UpsertProductHandler.php
nmarniesse: I found some reasons to not rely on getCreated/getUpdated:
- the dates are updated by the save process, that is a bad design. (we have subscribers that do the stuff). If the dates are not updated anymore, we break the feature. I preferred to be totally independent by handling our own date
- weird in spec tests, because the getCreated should return null then return a date. after the save. It's a smell that something is wrong

nmarniesse: Maybe I'm wrong, we can change if you prefer

LaurentPetard: My concern is that there's a risk that the date in the event may be different from the one saved in database. But it should not exceed one second, so it's maybe acceptable.

PR #19960, opened 2023-06-27
PIM-11067: fix display banner when user does not have system configur…

marinfazioakeneo:
…ation acl

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/front-packages/shared/src/dependencies/system-configuration.ts
+++ b/front-packages/shared/src/dependencies/system-configuration.ts
@@ -17,8 +17,6 @@ const systemConfiguration: SystemConfiguration = {
 
       return Promise.resolve();
     } catch (error) {
-      systemConfigurationData['sandbox_banner'] = true;
-
       return Promise.resolve();
     }
   },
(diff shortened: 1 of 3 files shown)


front-packages/shared/src/dependencies/system-configuration.ts
```diff
@@ -17,8 +17,6 @@ const systemConfiguration: SystemConfiguration = {
 
       return Promise.resolve();
     } catch (error) {
-      systemConfigurationData['sandbox_banner'] = true;
```
StevenVAIDIE, line 20: Can you revert this change ? In this way we will display the banner by default if there is another problem

StevenVAIDIE (approved):
Do not forget to change the CHANGELOG.md by adding a new line

front-packages/shared/src/dependencies/system-configuration.ts
marinfazioakeneo: done

ValentinMumble (approved):

PR #19978, opened 2023-06-29
PIM-11011: Fix versioning on common attribute values update using no …

cemorin:
…real time imports

When evaluating changeset between two versions, the operator `+` was merging left into right (taking stale data).
`array_merge` merges right into. (which is the newer version in this case)

`$modification + $oldSnapshot `
will result in:

```
description = "Quelles belles chaussures"
```

`array_merge`
will result in:
```
description = Nouvelles belles chaussures
```

So old version was re-applied and there was a difference between the newer version and the old, so the pending version was not removed.

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -104,6 +104,7 @@
 - PIM-11071: Fix potential divided by zero in completeness calculation
 - PIM-11073: Fix search with underscore on attribute grid
 - PIM-11035: Modify the command to fit with UCS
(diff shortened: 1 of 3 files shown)


pierallard (approved):
Quel talent

Beninho (approved):

cemorin:
> Quel talent

:clap: :clap: @pierallard

PR #19983, opened 2023-06-30
CPM-1067: Add endpoint to switch main identifier

pierallard:
This PR introduce
- A new endpoint for the end users
- A Command to switch main identifier
- A Handler to execute the Query
- A Validator to be sure the command is executable

### DIFF
--- a/src/Akeneo/Pim/Structure/Bundle/Application/SwitchMainIdentifier/SwitchMainIdentifierValidator.php
+++ b/src/Akeneo/Pim/Structure/Bundle/Application/SwitchMainIdentifier/SwitchMainIdentifierValidator.php
@@ -1,0 +1,69 @@
+<?php
+
+namespace Akeneo\Pim\Structure\Bundle\Application\SwitchMainIdentifier;
+
+use Akeneo\Pim\Structure\Component\AttributeTypes;
+use Akeneo\Pim\Structure\Component\Model\Attribute;
+use Akeneo\Pim\Structure\Component\Repository\AttributeRepositoryInterface;
+use Webmozart\Assert\Assert;
+
+/**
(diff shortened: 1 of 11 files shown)


src/Akeneo/Pim/Structure/Bundle/Application/SwitchMainIdentifier/SwitchMainIdentifierValidator.php
```diff
@@ -0,0 +1,69 @@
...
+                '%s attribute is not an identifier',
+                $this->newMainIdentifier
+            ));
+        }
+    }
+
+    private function validateAttributeIsNotMainIdentifier(): void
+    {
+        Assert::isInstanceOf($this->newMainIdentifier, Attribute::class);
+        if ($this->newMainIdentifier->isMainIdentifier()) {
```
Beninho, line 56: Shouldn't we do nothing if identifier is already the main?

Beninho (approved):

antoinearrive (approved):

src/Akeneo/Pim/Structure/Bundle/Application/SwitchMainIdentifier/SwitchMainIdentifierValidator.php
pierallard: It's ~20 lines above this one

pierallard: No because it will update "update" field even if we do nothing, and will only update main_identifier from 1 to 0, which will lead to "no main identifier"

mmetayer (approved):

PR #19999, opened 2023-07-04
PIM-11088: Catch ElasticSearch exception

Beninho:

When Elasticsearch is down, `/analytics/collect_data` return a **500 internal server error**.
And Navigation bar is not shown
PHP ElasticSearch Client throw _NoNodesAvailableException_

- [X] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [X] Changelog (maintenance bug fixes)
- [ ] Tech Doc

### DIFF
--- a/front-packages/shared/src/components/Navigation/PimNavigation.tsx
+++ b/front-packages/shared/src/components/Navigation/PimNavigation.tsx
@@ -54,9 +54,11 @@ const PimNavigation: FC<Props> = ({entries, activeEntryCode, activeSubEntryCode,
 
   useEffect(() => {
     fetch(router.generate('pim_analytics_data_collect')).then(response => {
-      response.json().then(data => {
-        setPimVersion(data);
-      });
+      if (response.ok) {
+        response.json().then(data => {
+          setPimVersion(data);
+        });
+      }
     });
   }, []);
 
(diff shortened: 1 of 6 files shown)


src/Akeneo/Platform/Bundle/CatalogVolumeMonitoringBundle/Persistence/Query/Elasticsearch/CountProducts.php
```diff
@@ -40,8 +37,13 @@ public function fetch(): CountVolume
...
             ],
         ];
-        $result = $this->client->count($query);
-
-        return new CountVolume((int)$result['count'], self::VOLUME_NAME);
+        try {
+            $result = $this->client->count($query);
+            $count = (int)$result['count'];
+        } catch (NoNodesAvailableException $exception) {
+            $count = -1;
```
nmarniesse, line 44: IMO we should not return a value, but a custom exception that explains clearly that we can't get the result.
Below the `CountVolume` class should not accept a negative value as it's not a valid count.

nmarniesse (commented):
Catching the exception hides the problem. We will not be aware of problem in the future, if for example ES is down for a long time.
For me:
- either we let the back like it is, and we handle the problem in front (but it can be hard depending on the legacy we have in front)
- or we add some `$logger->error('...')` to be notified

ahocquard:
I would make the front resilient to any issue on this endpoint to not block the menu, instead of fixing one by one all the potential issues.

cemorin (approved):

nmarniesse (approved):

mmetayer (approved):

PR #20002, opened 2023-07-04
CPM-1154: Add withNumber prop to ProgressIndicator

cemorin:

image

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc


antoinearrive (approved):

StevenVAIDIE (commented):
In my opinion there is not sense to have this variation, the designer should decide if we always want to have number or no.

cemorin:
> In my opinion there is not sense to have this variation, the designer should decide if we always want to have number or no.

I'll ask :+1:

front-packages/akeneo-design-system/src/components/ProgressIndicator/ProgressIndicator.tsx
```diff
@@ -80,11 +91,12 @@ type StepProps = Override<
      * The label of the step.
      */
     children?: ReactNode;
+    index?: number;
```
StevenVAIDIE, line 94: In order to be explicit about the fact that this props is calculated
```suggestion
    /**
     * @private
     */
    index?: number;
```

StevenVAIDIE (approved):
A lot more better like this ! 

Strange that the visual test pass even if the style have changed but anyway

Beninho (approved):

pierallard (approved):

PR #20012, opened 2023-07-06
SAAS-882 Allow using non default auth in-memory cache implementation …

gmagniez:
### Allow using non default auth in-memory cache implementation for GooglePubSub client.

Will allow adding for exemple an APCu cache to share the access_tokens between FPM processes when using short-lived credentials.

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -108,6 +108,7 @@
         "openspout/openspout": "^4.9.0",
         "opis/json-schema": "^2.3",
         "phpseclib/phpseclib": "^3.0",
+        "psr/cache": "^1.0.0",
         "psr/event-dispatcher": "^1.0.0",
         "ramsey/uuid": "4.7.1",
         "ramsey/uuid-doctrine": "^1.8",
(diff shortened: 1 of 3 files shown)


nestorAkeneo:

src/Akeneo/Tool/Bundle/MessengerBundle/Transport/GooglePubSub/PubSubClientFactory.php
```diff
@@ -14,21 +15,20 @@
...
+    private array $baseConfig = [];
 
-    public function __construct(string $keyFilePath)
+    public function __construct(string $keyFilePath = null, CacheItemPoolInterface $authCache = null)
     {
         if (!empty($keyFilePath)) {
-            $this->keyFilePath = $keyFilePath;
+            $this->baseConfig['keyFilePath'] = $keyFilePath;
+        }
+        if (!is_null($authCache)) {
```
ahocquard, line 25:
```
$authCache !== null
```

`is_null` is barely used in PHP world (even if it probably compiles the same way as an instruction instead of a function call)

composer.json
```diff
@@ -108,6 +108,7 @@
         "openspout/openspout": "^4.9.0",
         "opis/json-schema": "^2.3",
         "phpseclib/phpseclib": "^3.0",
+        "psr/cache": "^1.0.0",
```
ahocquard, line 111: `composer.lock` should have a change

gmagniez: I've upgraded the dependency to 2.0.0 version as it was already pulled by another package and regenerated the composer.lock

PR #20026, opened 2023-07-07
CPM-1066: Identifier attribute creation: scopable, localizable and unique fields are readonly

mmetayer:
Before:

before

After:

fixed


antoinearrive (approved):

Beninho (approved):

mmetayer:
Does not work, as configurations are merged instead of being overridden

PR #20059, opened 2023-07-12
CPM-1153: Add disabled prop to Tile component

cemorin:

image

- [ ] Tests
- [ ] Migration & Installer
- [ ] PM Validation (Story)
- [ ] Tech Doc


front-packages/akeneo-design-system/src/components/Tiles/Tiles.unit.tsx
```diff
@@ -4,13 +4,19 @@ import {render, screen, fireEvent} from '../../storybook/test-util';
 import {AssetCollectionIcon} from '../../icons';
 
 test('it renders tile correctly', () => {
```
StevenVAIDIE, line 6:
```suggestion
test('it trigger onClick when title is clicked', () => {
```

antoinearrive (approved):

front-packages/akeneo-design-system/src/components/Tiles/Tiles.tsx
```diff
@@ -58,12 +59,22 @@ const TileContainer = styled.div<
           border: 1px solid ${getColor('grey', 80)};
         `}
   &:hover {
```
StevenVAIDIE, line 61: In my opinion more easier to understand that all css hover style is applied only when is not disabled instead of saying which css style is applied on hover when disabled (we agree it totally do the same things)
 
```suggestion
    ${({disabled}) =>
      !disabled &&
      css`
        &:hover {
          border: 2px solid ${getColor('blue', 100)};
          color: ${getColor('blue', 100)};
          margin: 0;
          background: ${getColor('blue', 10)};
        }
      `}
  }
```

StevenVAIDIE (approved):
:+1:

Beninho (approved):

PR #20068, opened 2023-07-13
Avoid hydrating entities when counting from ES cursors

phaseinducer:
**Orignal issue**
When trying to count products for a Supplier Portal use case, we noticed that the product repo injected in ES cursors is called to hydrate found entities even just for counting.
In EE, the repo is decorated and also [applies permissions by side effect](https://github.com/akeneo/pim-enterprise-dev/blob/master/src/Akeneo/Pim/Permission/Bundle/Persistence/ORM/EntityWithValue/ProductRepository.php#L248).

**Why it's useless**
- Permissions are [already applied in the ES query by the PQB](https://github.com/akeneo/pim-enterprise-dev/blob/master/src/Akeneo/Pim/Permission/Bundle/Persistence/ORM/EntityWithValue/ProductQueryBuilderFactory.php#L68-L70).
- Other operations done when calling `rewind()` from `count()` concern only pagination and sorting, which are irrelevant when counting.

**Proposed solution**
Stop relying on the call to `rewind()` to populate the count result by a very hidden and implicit side effect (3 nested method calls needed), and dedicate a call to the ES client for counting.

Bonus: counting entities should be more efficient.

### DIFF
--- a/src/Akeneo/Pim/Enrichment/Bundle/Elasticsearch/AbstractCursor.php
+++ b/src/Akeneo/Pim/Enrichment/Bundle/Elasticsearch/AbstractCursor.php
@@ -23,6 +23,7 @@ abstract class AbstractCursor implements CursorInterface
     protected Client $esClient;
     protected ProductRepositoryInterface $productRepository;
     protected ProductModelRepositoryInterface $productModelRepository;
+    protected array $esQuery;
     protected ?array $items = null;
     protected ?int $count = null;
     protected int $position = 0;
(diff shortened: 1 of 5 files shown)


src/Akeneo/Pim/Enrichment/Bundle/Elasticsearch/AbstractCursor.php
```diff
@@ -68,8 +69,11 @@ public function valid(): bool
      */
     public function count(): int
     {
-        if (null === $this->items) {
-            $this->rewind();
+        if (null === $this->count) {
+            $esQuery = $this->esQuery + ['track_total_hits' => true];
```
mmetayer, line 73:
```suggestion
            $esQuery = \array_replace($this->esQuery, ['track_total_hits' => true]);
```
(in case `track_total_hits` is already set to `false`, though it's very unlikely)

mmetayer (approved):

PR #20099, opened 2023-07-20
CPM-1080: Forbid deletion of main identifier (front)

pierallard:
image (5)


src/Akeneo/Platform/Bundle/UIBundle/Resources/public/js/attribute/form/delete/DeleteModal.tsx
```diff
@@ -19,6 +21,17 @@ const Highlight = styled.span`
   font-weight: bold;
 `;
 
+const useMainIdentifierCode = () => {
```
cemorin, line 24: Just a suggestion, we're starting to use it at several places.
It could be useful to have it in a dedicated file (using react-hook maybe?)

cemorin (approved):

src/Akeneo/Platform/Bundle/UIBundle/Resources/public/js/attribute/form/delete/DeleteModal.tsx
pierallard: Answered IRL

PR #20289, opened 2023-09-26
PIM-11200: Fix versinoning on table attributes

pierallard:
# How to reproduce

How to reproduce the bug
- Create a product with a table attribute
- Save a table value in this product, with the order coming from the UI (still the same)
- Update this product table value through API, by changing the column order
- Update this product again in any other value
- A new version is created, including the table value

# The issue

The issue comes from the VersionBuilder. This VersionBuilder creates "changeset" (the changes of the new product regarding the previous version of the product.
For each value, we compare strings to strings (for attributes and properties). If `old !== new`, then the value is considered as updated.
As the string `"[{a: "a", b:"b"}]"` differs from `"[{b:"b", a: "a"}]"`, a new version is created.

# The fix

We use 
- `\json_decode` to decode the string. If the string was previously a string, no changed are made. In the case of a table attribute, it will create an array of associative arrays.
- We use `!=` instead of `!==`. Indeed, with triple equals, 2 arrays are not the same if the key order is not the same. With double equals, the key order is not important.

### DIFF
--- a/src/Akeneo/Tool/Bundle/VersioningBundle/Builder/VersionBuilder.php
+++ b/src/Akeneo/Tool/Bundle/VersioningBundle/Builder/VersionBuilder.php
@@ -184,8 +184,14 @@ class VersionBuilder
 
     private function hasValueChanged($old, $new): bool
     {
-        if (null !== $hasChanged = $this->hasLegacyDateChanged($old, $new)) {
-            return $hasChanged;
+        $hasLegacyDateChanged = $this->hasLegacyDateChanged($old, $new);
+        if (null !== $hasLegacyDateChanged) {
+            return $hasLegacyDateChanged;
+        }
+
+        $hasTableAttributeChanged = $this->hasTableAttributeChanged($old, $new);
+        if (null !== $hasTableAttributeChanged) {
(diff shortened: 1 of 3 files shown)


src/Akeneo/Tool/Bundle/VersioningBundle/Builder/VersionBuilder.php
```diff
@@ -184,11 +184,12 @@ function ($item) {
...
     private function hasValueChanged($old, $new): bool
     {
-        if (null !== $hasChanged = $this->hasLegacyDateChanged($old, $new)) {
+        $hasChanged = $this->hasLegacyDateChanged($old, $new);
+        if (null !== $hasChanged) {
             return $hasChanged;
         }
 
-        return $old !== $new;
+        return \json_decode($old, true) != \json_decode($new, true);
```
mmetayer, line 192:
```bash
php > var_dump(json_decode('', true) == json_decode('0', true));
bool(true)
```
:/

mmetayer (approved):

Beninho (approved):

mmetayer:
Don't forget the CHANGELOG :slightly_smiling_face:

PR #20333, opened 2023-10-31
OCT-214: add possibility to customize node UID and GID when building node image

hogren:


docker/Dockerfile_node
```diff
@@ -1,7 +1,10 @@
 FROM debian:bullseye-slim
 
-RUN groupadd --gid 1000 node \
-    && useradd --uid 1000 --gid node --shell /bin/bash --create-home node
+ARG DEV_UID=1000
+ARG DEV_GID=1000
+
+RUN groupadd --gid ${DEV_GID} node \
+    && useradd --uid ${DEV_UID} --gid node --shell /bin/bash --create-home node
```
damien-carcel, line 7: Don't you need to modify a Makefile or something to allow to change those args?

PR #20351, opened 2023-11-23
PIM-10985: Mark skipped migrations as executed

pierallard:
# The issue

When coding migrations, we usually used "skipIf" method in case of "the state of the PIM is already migrated". This method skips the migration, that's OK, but it did not add it as "executed". Each time we will try to re-execute migrations, we will try to re-execute theses ones.
This is not a good practice ! "skipIf" means "skip and retry later".

# The solution

Don't use "skipIf" ! Return is OK, it will mark migration as executed.

### DIFF
--- a/CHANGELOG-6.0.md
+++ b/CHANGELOG-6.0.md
@@ -1,5 +1,7 @@
 # 6.0.x
 
+- PIM-10985: Mark skipped migrations as executed
+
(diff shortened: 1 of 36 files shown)


Beninho (approved):

StevenVAIDIE (approved):
Really nice ! I was always afraid that this thing would break between our fingers !

PR #20383, opened 2024-01-09
PIM-11342: Fix infinite loop in Re-compute DQI products scores

LaurentPetard:
There were a infinite loop in the job that re-compute the score of all the products. In the SQL query, the condition on the UUID was comparing an UUID as string with an UUID as binary.

The script is not tested, and it would need too much effort for a SLA. Especially as this job is only executed manually, and should be fully reworked.

### DIFF
--- a/CHANGELOG-7.0.md
+++ b/CHANGELOG-7.0.md
@@ -1,5 +1,9 @@
 # 7.0.x
 
+## Bug fixes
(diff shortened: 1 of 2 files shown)


jpa44 (approved):

artggd (approved):
Nice

PR #20501, opened 2025-02-21
[Backport] PIM-11936: Add reference entity collection filter validator and limit

ValentinMumble:
See: https://github.com/akeneo/pim-enterprise-dev/pull/27664

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -86,7 +86,7 @@
         "doctrine/event-manager": "^1.1.1",
         "doctrine/instantiator": "^1.4.0",
         "doctrine/migrations": "^3.1.4",
-        "doctrine/orm": "^2.9.0",
+        "doctrine/orm": "2.20.0",
         "doctrine/persistence": "^2.2.3",
         "dompdf/dompdf": "2.0.3",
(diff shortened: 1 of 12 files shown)


composer.json
```diff
@@ -86,7 +86,7 @@
         "doctrine/event-manager": "^1.1.1",
         "doctrine/instantiator": "^1.4.0",
         "doctrine/migrations": "^3.1.4",
-        "doctrine/orm": "^2.9.0",
+        "doctrine/orm": "2.20.0",
```
ValentinMumble, line 89: I had to fix doctrine's version because on version `2.20.2` a behat fails on `Uncaught PHP Exception Doctrine\ORM\ORMInvalidArgumentException: "Entity Akeneo\Pim\Enrichment\Component\Product\Model\ProductModelAssociation@7037 is not managed. An entity is managed if its fetched from the database or registered as new through EntityManager#persist`
The issue is present since December 2024 at least, see: https://github.com/akeneo/pim-enterprise-dev/commits/7.0/

olivierpontier (approved):

phaseinducer (approved):

PR #20632, opened 2026-06-02
Community Edition Tech Refresh - MySQL 8.4 LTS

BitOne:
## Summary

- Upgrades MySQL from 8.0.34 to 8.4 LTS, the next long-term support release
- Updates version requirements to accept the 8.4.x series (>= 8.4.0, < 9.0.0)

## Test plan

- [x] CI passes with MySQL 8.4
- [x] Application starts and connects to MySQL 8.4
- [x] Run the full test suite against MySQL 8.4


dmirtillo:
[mysql-8.4-pr20632-discussion.patch](https://github.com/user-attachments/files/28595577/mysql-8.4-pr20632-discussion.patch)

Thanks for driving the MySQL 8.4 LTS refresh — it's much needed and we are testing this in our internal dev environments (Akeneo CE 7.0, `v7.0.84`) and it works well. While rolling it out we hit a few gaps **beyond** what this PR covers and wanted to share them for discussion. The Doctrine platform/types/config work here is applied unchanged; everything below is additive.

## 1. Root-cause `GREATEST()` fix (the substance of #20624)

This PR makes the Doctrine layer *tolerant* of `DATETIME(6)` (`...000000`) values, but the ES projections still *produce* them: `GREATEST()` mixing a `DATETIME` with an integer `0` fallback (`COALESCE(x, 0)`) gets promoted to `DATETIME(6)` on 8.4. We fix the source so every `GREATEST()` argument stays a `DATETIME` (`COALESCE(..., 0)` → `COALESCE(..., <row>.updated)`).

**Question:** should this PR also absorb #20624? Tolerating *and* not producing `DATETIME(6)` are complementary.

## 2. Raw `createFromFormat('Y-m-d H:i:s')` sites that bypass the Doctrine types

A few code paths read a datetime straight from a DB row and parse it strictly, so `MySQL84Platform` never sees them — they still throw `Expected format: Y-m-d H:i:s` if their column is promoted to `DATETIME(6)`. We hardened them with the same `.u` rule as `MySQL84Platform::getDateTimeFormatString()`:

- `JobExecutionRowHydrator.php` (DB `start_time`)
- `Category::fromDatabase()` (DB `updated`)
- `GetDelayUntilNextRequest.php` (DB `updated`)
- `ApplyProductSearchQueryParametersToPQB.php` (2 sites — these parse API filter input rather than DB output, so lower risk; included for consistency, happy to drop)

**Question:** would a small shared helper (or exposing the platform's format rule) be preferable to inlining the `str_contains` check at each site?

## 3. `docker-compose.yml` — auth flag removed in 8.4

The dev compose passed `--default-authentication-plugin=mysql_native_password`, which is *removed* in 8.4 (server won't start) → replaced with `--authentication-policy=caching_sha2_password`. This PR's compose doesn't carry that flag, so it wasn't an issue here — flagging in case other CE dev compose variants need the same treatment.

---

> Patch is against `v7.0.84` (this PR targets `master`), so line offsets differ — please read by intent. It applies cleanly via both `git apply` and `patch -p1`.

BitOne:
Thank a lot @dmirtillo for your detailed comment.
Some details first:
- we won't update the 7.0 branch to make it compatible with MySQL 8.4, as we are in maintenance mode on this branch. I know it's not what you implied, but I wanted to make this clear. That's why we released [new versions recently](https://github.com/akeneo/pim-community-dev/releases/tag/v2026.3).
- I would prefer not to put any authentication (or any flag for that matter) that are not strictly mandatory, that's why I simply removed the previous one

Great finding on the GREATEST() issue! It's worth building a test here, and add a comment on the PHPDoc of the function otherwise it will stay quite cryptic for future maintainers.

> Question: would a small shared helper (or exposing the platform's format rule) be preferable to inlining the str_contains check at each site?

Yes, definitely, as that would help the readability. We can still use the [Platform file](https://github.com/akeneo/pim-community-dev/pull/20632/changes#diff-ccc3690d98da6d65b74963abd7163aec480b1606ec46fc55b3e0fc97665ed7a0), as this makes it clear it's related to the platform specificities.

And here again, it seems we are missing some tests, as the CI is green on this PR.

I will have a look at what kind of tests we should do.

BitOne:
@dmirtillo, I was able to reproduce the issue on the GREATEST() part.
I added tests on the ES projection integration tests:
- https://github.com/akeneo/pim-community-dev/pull/20632/changes#diff-7068eb18c90886a61459bf1321dfc293bd0f893020b17ae67a0e9483bb9c5e0cR133
- https://github.com/akeneo/pim-community-dev/pull/20632/changes#diff-935d88b35fba3fd9f5ad307f622c58bb6dfd7ce51104cc69ca98a6b6bd04266fR298

And indeed these two tests fail on MySQL 8.4.

But I need your help: I have a hard time reproducing the bugs you mentioned on the following code:
> JobExecutionRowHydrator.php (DB start_time)
> Category::fromDatabase() (DB updated)
> GetDelayUntilNextRequest.php  (DB updated)

I created tests for:
- [JobExecution](https://github.com/akeneo/pim-community-dev/pull/20632/changes#diff-b436a9aab06ac76117ab3e579da583d6ac9c2746def4fdc59a13392848134239R385)
- [Category](https://github.com/akeneo/pim-community-dev/pull/20632/changes#diff-de8765ebb39c7f27bfffb8399799aad97d8d7764f30287b1f983ed71e7d145afR147).
- [Event API Delay](https://github.com/akeneo/pim-community-dev/pull/20632/changes#diff-697aa752d94a52a585fbb0db52ac694379b105e84a78b9d0caff09554d54a57dR34)

But none of them are failing.

Can you tell me what was failing under MySQL 8.4 that made you do these changes?
Thanks!

dmirtillo:
Thanks for reproducing the `GREATEST()` failures and adding the integration tests — and for the question about the other sites, which is what made me dig in. Short answer: **the only real, reproducible failure is `GREATEST()`; the `JobExecutionRowHydrator` / `Category` / `GetDelayUntilNextRequest` paths don't fail on 8.4, so there's nothing to change there.**

Could you tell me which branch or release is this planned to be released on?

## Why those three tests don't fail

Their datetime values come straight from a plain `DATETIME(0)` column:

- `Category` → `category.updated` (`GetCategoriesSql` / `GetCategorySql`)
- Event API delay → `SELECT updated …`, column is `updated DATETIME NOT NULL`
- JobExecution → `start_time` selected directly

MySQL 8.4 does **not** change the precision of `DATETIME(0)` columns — they still return values *without* microseconds, so `createFromFormat('Y-m-d H:i:s', …)` parses fine. No `DATETIME(6)` value ever reaches those hydrators, which is exactly why your tests for them pass. The `MySQL84Platform` you added already covers anything that *does* go through the Doctrine types, so no extra per-site hardening is needed.

## The one thing still missing: the `GREATEST()` root cause

This PR makes the Doctrine layer *tolerant* of `DATETIME(6)`, but the ES projections still *produce* it: `GREATEST()` mixing a `DATETIME` with an integer `0` fallback (`COALESCE(x, 0)`) is promoted to `DATETIME(6)` on 8.4 (`…000000`). The fix is to keep every `GREATEST()` argument a `DATETIME` by using the row's own column as the fallback:

```sql
-- before (promoted to DATETIME(6) on 8.4):
GREATEST(product.updated, COALESCE(sub_product_model.updated, 0), COALESCE(root_product_model.updated, 0))
-- after (stays DATETIME on all versions):
GREATEST(product.updated, COALESCE(sub_product_model.updated, product.updated), COALESCE(root_product_model.updated, product.updated))
```

Semantically identical — a row's own `updated` can never beat itself in `GREATEST()`. This is the substance of #20624; tolerating *and* not producing `DATETIME(6)` are complementary, so it'd be great to absorb it here. It makes your two new ES projection integration tests pass on 8.4.

Patch below applies cleanly on this PR branch (`community-edition-mysql-8.4-upgrade`) via `git apply` / `patch -p1`.

<details>
<summary>📎 GREATEST() root-cause fix (2 files)</summary>

[mysql-8.4-greatest-fix-PR-branch.patch](https://github.com/user-attachments/files/28751487/mysql-8.4-greatest-fix-PR-branch.patch)

```diff
diff --git a/src/Akeneo/Pim/Enrichment/Bundle/Storage/Sql/ElasticsearchProjection/GetElasticsearchProductModelProjection.php b/src/Akeneo/Pim/Enrichment/Bundle/Storage/Sql/ElasticsearchProjection/GetElasticsearchProductModelProjection.php
index 6990c10d..16fa9b04 100644
--- a/src/Akeneo/Pim/Enrichment/Bundle/Storage/Sql/ElasticsearchProjection/GetElasticsearchProductModelProjection.php
+++ b/src/Akeneo/Pim/Enrichment/Bundle/Storage/Sql/ElasticsearchProjection/GetElasticsearchProductModelProjection.php
@@ -98,7 +98,7 @@ WITH
             product_model.code,
             product_model.created,
             root_product_model.code AS parent_code,
-            GREATEST(product_model.updated, COALESCE(root_product_model.updated, 0)) as updated,
+            GREATEST(product_model.updated, COALESCE(root_product_model.updated, product_model.updated)) as updated,
             product_model.updated as entity_updated,
             JSON_MERGE_PATCH(COALESCE(root_product_model.raw_values, '{}'), COALESCE(product_model.raw_values, '{}')) AS raw_values,
             family.code AS family_code,
diff --git a/src/Akeneo/Pim/Enrichment/Bundle/Storage/Sql/ElasticsearchProjection/GetElasticsearchProductProjection.php b/src/Akeneo/Pim/Enrichment/Bundle/Storage/Sql/ElasticsearchProjection/GetElasticsearchProductProjection.php
index aa7b4346..87281e1a 100644
--- a/src/Akeneo/Pim/Enrichment/Bundle/Storage/Sql/ElasticsearchProjection/GetElasticsearchProductProjection.php
+++ b/src/Akeneo/Pim/Enrichment/Bundle/Storage/Sql/ElasticsearchProjection/GetElasticsearchProductProjection.php
@@ -139,7 +139,7 @@ WITH
             JSON_ARRAY(sub_product_model.id, root_product_model.id) AS ancestor_ids,
             JSON_ARRAY(sub_product_model.code, root_product_model.code) AS ancestor_codes,
             product.created AS created_date,
-            GREATEST(product.updated, COALESCE(sub_product_model.updated, 0), COALESCE(root_product_model.updated, 0)) AS updated_date,
+            GREATEST(product.updated, COALESCE(sub_product_model.updated, product.updated), COALESCE(root_product_model.updated, product.updated)) AS updated_date,
             product.updated AS entity_updated_date,
             COALESCE(JSON_KEYS(product.raw_values), JSON_OBJECT()) AS attribute_codes_in_product_raw_values,
             JSON_MERGE_PATCH(
```
</details>

BitOne:
Thanks @dmirtillo, it makes sense. Do you mind testing this PR branch with your use cases? If that works for you, I will merge it, and let people do some beta testing on their end on the main branch.

dmirtillo:
@BitOne Sure, I'm happy to test it!

Do you think you could also implement the patch from my previous comment?
I'm referring to

[mysql-8.4-greatest-fix-PR-branch.patch](https://github.com/user-attachments/files/28751487/mysql-8.4-greatest-fix-PR-branch.patch)

BitOne:
> Do you think you could also implement the patch from my previous comment?

Yes, applied. Thank you!

AdelELHAIBA-Akeneo (approved):

buts26:
Hello everyone,

When can we expect a release containing this fix?

PR #20646, opened 2026-08-22
Reindex reverse side of 2-way association on removal (#20547)

### LINKED ISSUE #20547: Index refresh missing when deleting a product from 2-way association
# Description

When a product is deleted from a 2-way association, only the product where the change has been made has its index refreshed in Elasticsearch.

As a result, only one product appears in the API results when your request has a filter on the date.

e.g.: `/api/rest/v1/products?search={"enabled":[{"operator":"=","value":true}],"updated":[{"operator":">","value":"2025-09-09 07:00:00"}]}&pagination_type=search_after&limit=10&with_count=true`

# Localisation

Akeneo v5.x, v6.x and v7.x

# Steps to reproduce

1. Log into a demo instance with demo products
2. Create a 2-way association
3. Edit product HP KQ245AA (sku: 16672632), go to the association tab
4. Select the newly created 2-way association, add HP Pro Webcam (sku: 3466711)
5. Confirm that both products have been updated in Akeneo and Elasticsearch  

Image

6. Remove HP Pro Webcam from the association of HP KQ245AA product
7. Confirm that both products have been updated in Akeneo
8. Confirm that only HP KQ245AA has been updated in Elasticsearch
Image

# How it should work

Both products HP KQ245AA and HP Pro Webcam should have their indexes refreshed in Elasticsearch after deletion of a 2-way association.

wakqasahmed:
## Summary

Fixes #20547.

When a 2-way association is removed, only the product/product model the change was applied on gets reindexed in Elasticsearch. The reciprocal (other) side of the association is correctly updated in the database, but its Elasticsearch document is never refreshed, so it can go missing from date-filtered API results (e.g. `updated > X`) even though it was in fact modified.

## Root cause

- `TwoWayAssociationUpdater::removeInversedAssociation()` correctly mutates and `persist()`s the reciprocal entity in the database.
- `PersistTwoWayAssociationSubscriber::handlePreSave()` queues entities for Elasticsearch reindexing by walking the *owner's current* association state at `PRE_SAVE` time.
- On **add**, the newly associated product still appears in the owner's association list at `PRE_SAVE` time, so it gets queued and reindexed correctly.
- On **remove**, the disassociated product has already been removed from the owner's association list by the time `PRE_SAVE` fires, so it's never walked, never queued, and never reindexed — even though its DB row did change.

## Fix

Added `TwoWayAssociationsIndexationQueue`, a small shared service between `TwoWayAssociationUpdater` and `PersistTwoWayAssociationSubscriber`. `TwoWayAssociationUpdater::removeInversedAssociation()` now explicitly queues the reciprocal entity (product or product model) for reindexing, in addition to persisting it. The queue is still flushed on `POST_SAVE`, after the whole unit of work — including the reciprocal entity — has been committed to the database, so the reindexer reads fresh data.

## Test plan

- [x] `php -l` on all changed/added PHP files (no `php`/vendor available in this environment to run the full suite)
- [x] Updated `TwoWayAssociationUpdaterSpec` to assert the reciprocal product/product model is queued for reindexing on `removeInversedAssociation()`, and that nothing is queued when the method throws
- [x] Added `TwoWayAssociationsIndexationQueueSpec` covering queue/flush behavior for both products and product models
- [ ] Could not run the PhpSpec/PHPUnit suite locally (no `php` binary / vendor directory in this environment) — please let CI confirm


wakqasahmed:
Hi @micheltag @mmetayer — noticed this PR doesn't have a reviewer assigned yet — it's been about 4 days, CI is green and it's mergeable. Would you (or whoever's best placed) be able to take a look, or point me to who should? Thanks!

BitOne:
Thanks @wakqasahmed for your contribution.

Please run the tests locally before submitting a PR. Also, in this context, I expect to see integration testing, as the issue you described requires a running MySQL and Elasticsearch to really validate the solution.

Thank you in advance.

wakqasahmed:
Thanks @BitOne — fair on both counts. I've now got the full docker stack running locally (MySQL + Elasticsearch + the pubsub emulator, `make pim-test` style with `APP_ENV=test make database`), and I've added the integration test you asked for.

First, running the specs locally caught a genuine mistake of mine that I'd otherwise have sent to CI again: in `PersistTwoWayAssociationSubscriberSpec` I'd called `->reveal()` on the injected collaborators, which isn't a phpspec `Collaborator` method (that's Prophecy's API) — it needed `->getWrappedObject()`. That example was broken:

```
 47  - it reindexes the reciprocal product after a two way association is removed by the updater
      method `Double\ManagerRegistry\ManagerRegistry\P5::reveal()` is not defined.
3 examples (2 passed, 1 broken)
```

Fixed, and the three specs touched by this PR are now green:

```
$ make phpspec F=tests/back/Pim/Enrichment/Specification/Bundle/Doctrine/ORM/Updater/TwoWayAssociationUpdaterSpec.php
15 examples (15 passed)
$ make phpspec F=tests/back/Pim/Enrichment/Specification/Bundle/EventSubscriber/EntityWithAssociations/PersistTwoWayAssociationSubscriberSpec.php
3 examples (3 passed)
$ make phpspec F=tests/back/Pim/Enrichment/Specification/Component/Product/Association/TwoWayAssociationsIndexationQueueSpec.php
5 examples (5 passed)
```

On the integration side, I've added `tests/back/Pim/Enrichment/Integration/Product/Association/TwoWayAssociationIndexationIntegration.php`. Associations aren't part of the product ES document, so asserting on document content wouldn't show anything. Instead the test deletes the reciprocal product's ES document, then saves only the *owner* with a `DissociateProducts` intent — so the document can only reappear if the two-way association update actually queued the reciprocal for reindexation.

Against this branch, that plus the pre-existing `CreateTwoWayAssociationIntegration` both pass:

```
$ APP_ENV=test docker-compose run --rm php php vendor/bin/phpunit -c phpunit.xml.dist \
    --testsuite PIM_Integration_Test \
    --filter '(CreateTwoWayAssociationIntegration|TwoWayAssociationIndexationIntegration)'
PHPUnit 9.6.16 by Sebastian Bergmann and contributors.

......                                                              6 / 6 (100%)

Time: 00:19.802, Memory: 153.00 MB

OK (6 tests, 78 assertions)
```

I also checked the new test genuinely reproduces the reported bug rather than just passing: reverting the `src/` changes on this branch and re-running it fails as expected, which is the behaviour described in #20547.

```
1) ...TwoWayAssociationIndexationIntegration::it_reindexes_the_reciprocal_product_when_the_owner_removes_a_two_way_association
Expected the Elasticsearch document "product_98a15a60-a429-4c9a-924d-15e404eee2a5" to exist
Failed asserting that 0 is identical to 1.

Tests: 2, Assertions: 5, Failures: 1.
```

Restoring `src/` makes it green again. Pushed as 7f809aa. Sorry for the earlier round-trips on CI — I have the local setup working now and will run things here before pushing.

wakqasahmed:
Hi @micheltag @BitOne — circling back on this one — still no reviewer assigned, quiet for a while since the last activity. No pressure, just don't want it to get lost.

PR #20651, opened 2026-08-27
Fix product grid crash on numeric attribute code columns

wakqasahmed:

A purely numeric attribute code (e.g. `385`) used as a product grid column crashed the grid with `Could not found column(s) "385" for sorting`.

`FeatureFlagDatagridFilterListener::featureFlagOnDefaultColumns()` rebuilt the columns config with `offsetUnset()` + `offsetAddToArray()`. `offsetAddToArray()` merges via PHP's `array_merge()`, which renumbers integer-like array keys sequentially — a decimal string key like `"385"` is stored as an int key by PHP, so `array_merge` turned it into `0`. The sorter config a few lines below is built with `offsetSetByPath()`, which assigns directly and keeps the real key, so `SorterExtension` could no longer match the column to its sorter definition and threw.

Switched to `offsetSet()` — direct, key-preserving assignment, the same approach `featureFlagOnSorter()` already uses a few lines below. Behaviour for string-keyed columns is unchanged.

- [x] Tests — added `FeatureFlagDatagridFilterListenerTest`, registered in `phpunit.xml.dist`, covering a numeric-coded column keeping its key alongside normal string-keyed columns and feature-flagged-off removal
- [ ] Migration & Installer — not applicable
- [ ] PM Validation (Story) — not applicable, bug fix
- [ ] Changelog (maintenance bug fixes) — happy to add if requested
- [ ] Tech Doc — not applicable

Could not run PHPUnit in the environment used to write this (no PHP runtime available) — please confirm in CI.


BitOne:
Hello @wakqasahmed,

Please run the tests locally, as shown in previous PR.
Thank you!

wakqasahmed:
Understood @BitOne — I've got the docker setup working locally now. This change is covered by a PHPUnit test rather than a phpSpec one, so the equivalent target is `make phpunit`:

```
$ make phpunit F=src/Oro/Bundle/DataGridBundle/Tests/Unit/EventListener/FeatureFlagDatagridFilterListenerTest.php
APP_ENV=test docker-compose run --rm php php vendor/bin/phpunit -c phpunit.xml.dist src/Oro/Bundle/DataGridBundle/Tests/Unit/EventListener/FeatureFlagDatagridFilterListenerTest.php
PHPUnit 9.6.16 by Sebastian Bergmann and contributors.

..                                                                  2 / 2 (100%)

Time: 00:00.161, Memory: 8.00 MB

OK (2 tests, 4 assertions)
```

Since this PR also registers a new `OroDataGridBundle` testsuite in `phpunit.xml.dist`, I ran it through that entry too, to confirm the registration actually picks the test up:

```
$ APP_ENV=test docker-compose run --rm php php vendor/bin/phpunit -c phpunit.xml.dist --testsuite OroDataGridBundle
OK (2 tests, 4 assertions)
```

Both exited 0. No changes were needed, so the branch is unchanged at e632263.

wakqasahmed:
Hi @micheltag @BitOne — this one seems to have no reviewer assigned, about 4 days in with CI passing. Happy to be pointed at whoever's best placed to look.

wakqasahmed:
Hi @micheltag @BitOne — another follow-up on this one, still unassigned after a while. Let me know if it needs anything from me.

